PR #88, opened 2018-02-14
Clean up the output that goes to stderr.

dlorenc:
Info logging is a bit ugly, so messages that should be user-visible get
printed to stderr directly.

Also add a --quiet flag.


dgageot:
@dlorenc should we close this one?

r2d4:
Closing and tracking in this issue #370

PR #122, opened 2018-03-01
Generate a manifest if one is not provided.

dlorenc:
Note: this also includes the change from #119. I pulled that out into it's own PR because this one will likely take longer to get merged.

I mainly did this as a prototype, but I actually like it now. LMK what you think.

### DIFF
--- a/examples/no-manifest/skaffold.yaml
+++ b/examples/no-manifest/skaffold.yaml
@@ -1,0 +1,47 @@
+apiVersion: skaffold/v1
+kind: Config
+build:
+  # tagPolicy determines how skaffold is going to tag your image
+  # we provide a few strategies here, although you most likely won't need to care!
+  # Valid values for tagPolicy are
+  #
+  # sha256    - tags the image with the checksum of the built image (image id)
+  # gitCommit - tags the image with the git commit of your current repository
+  
+  # tagPolicy: sha256
+
+  # artifacts is a list of the actual images you're going to be building
+  # you can include as many as you want here.
+  artifacts:
+    # The name of the image we're going to build, the full reference to the built
+    # image will be available later in the the deploy section
+  - imageName: changeme # gcr.io/k8s-skaffold/skaffold-example
+    # dockerfilePath: Dockerfile # not required if its just $workspace/Dockerfile
+
+    # workspace is the path to your dockerfile context
+    workspace: examples/getting-started
+
+  # This next section is where you'll put your specific builder configuration.
+  # Here, we're using a local builder, which simply builds using the host docker
+
+  # If you're using minikube, you can skip the push step for even faster build
+  # and deploy cycles.
+  # local: {} # Use this for minikube configuration
+  local: {}
+    # Skaffold defers to your ~/.docker/config for authentication information
+    # If you're using GCR, make sure that you have gcloud and
+    # docker-credentials-helper-gcr configured correctly.
+    #
+    # gcloud components install docker-credential-gcr
+    # docker-credential-gcr configure-docker
+    # skipPush: true
+
+# The deploy section has all the information needed to deploy. Along with build:
+# it is a required section.
+deploy:
+  # We'll be deferring to kubectl for the deployment. The kubectl deployer uses
+  # a client side kubectl apply to apply the manifests to the cluster. You'll
+  # need a kubectl version installed thats compatible with your cluster.
+  kubectl:
+    # Manifests is a list of manifests to deploy
+    manifests:
--- a/pkg/skaffold/deploy/kubectl.go
+++ b/pkg/skaffold/deploy/kubectl.go
@@ -20,8 +20,13 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
+	"text/template"
+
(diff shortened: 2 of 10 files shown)


examples/no-manifest/skaffold.yaml
```diff
@@ -0,0 +1,47 @@
...
+  # artifacts is a list of the actual images you're going to be building
+  # you can include as many as you want here.
+  artifacts:
+    # The name of the image we're going to build, the full reference to the built
+    # image will be available later in the the deploy section
+  - imageName: changeme # gcr.io/k8s-skaffold/skaffold-example
+    # dockerfilePath: Dockerfile # not required if its just $workspace/Dockerfile
+
+    # workspace is the path to your dockerfile context
+    workspace: examples/getting-started
```
r2d4: examples/no-manifest

examples/no-manifest/skaffold.yaml
```diff
@@ -0,0 +1,47 @@
...
+  # gitCommit - tags the image with the git commit of your current repository
+  
+  # tagPolicy: sha256
+
+  # artifacts is a list of the actual images you're going to be building
+  # you can include as many as you want here.
+  artifacts:
+    # The name of the image we're going to build, the full reference to the built
+    # image will be available later in the the deploy section
+  - imageName: changeme # gcr.io/k8s-skaffold/skaffold-example
```
r2d4: you can just put this as `gcr.io/k8s-skaffold/skaffold-example`

examples/no-manifest/skaffold.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: skaffold/v1
+kind: Config
+build:
+  # tagPolicy determines how skaffold is going to tag your image
```
r2d4: you can remove all the comments

pkg/skaffold/deploy/kubectl.go
```diff
@@ -31,6 +36,35 @@ import (
 	"github.com/pkg/errors"
 )
 
+// Slightly modified from kubectl run --dry-run
+var deploymentTemplate = template.Must(template.New("deployment").Parse(`apiVersion: extensions/v1beta1
+kind: Deployment
```
r2d4: you need a service too right?

pkg/skaffold/deploy/kubectl.go
```diff
@@ -44,6 +78,23 @@ func NewKubectlDeployer(cfg *config.DeployConfig) (*KubectlDeployer, error) {
 // Run templates the provided manifests with a simple `find and replace` and
 // runs `kubectl apply` on those manifests
 func (k *KubectlDeployer) Run(out io.Writer, b *build.BuildResult) (*Result, error) {
+
```
r2d4: nit remove newline

examples/no-manifest/skaffold.yaml
r2d4: on second thought, you can leave this like it is and remove the dockerfile and main.go if you want. Then this folder can just be the skaffold.yaml

pkg/skaffold/docker/parse.go
```diff
@@ -108,6 +109,47 @@ func GetDockerfileDependencies(workspace string, r io.Reader) ([]string, error)
...
+		return nil, errors.Wrap(err, "parsing dockerfile")
+	}
+
+	// Check the dockerfile and the base.
+	ports := []string{}
+	for _, value := range res.AST.Children {
+		switch value.Value {
+		case from:
+			base := value.Next.Value
+			if base == "SCRATCH" {
```
r2d4: why couldn't you expose ports in a scratch image?

r2d4: you don't have ca-certs but internal services, etc.

pkg/skaffold/deploy/kubectl.go
dlorenc: You don't *need* one, only if you want to access it over an LB or something. I think we should probably recommend port-forward in any samples.

pkg/skaffold/docker/parse.go
dlorenc: It's about the FROM check - we check your Dockerfile and the base image for ports. a SCRATCH base image won't have any PORTS exposed itself so we don't need to try to check it. We still check your Dockerfile for EXPOSES though.

r2d4:
TIL switching the repository from private to public breaks the existing forks (why? I have no clue).  I had to re-fork and push to the new remote

https://help.github.com/articles/what-happens-to-forks-when-a-repository-is-deleted-or-changes-visibility/#changing-a-private-repository-to-a-public-repository

dlorenc:
> TIL switching the repository from private to public breaks the existing forks (why? I have no clue). I had to re-fork and push to the new remote

Yeah I'll have to recreate that, but I think this one is still working, unless I'm missing something.

pkg/skaffold/deploy/kubectl.go
dlorenc: done.

dlorenc:
> Yeah I'll have to recreate that, but I think this one is still working, unless I'm missing something.

nvm, it wasn't picking up new commits. I see. Opened a new PR: https://github.com/GoogleCloudPlatform/skaffold/pull/141

Issue #94, opened 2018-02-19
Handle optional port-forwarding


viglesiasce:
I started to go down the route of figuring out how to work with Skaffold and use a remote debugger (delve in my case). 

The only thing I really needed from Skaffold was for it to be able to reconfigure a port-forwarding setup to one of the new pods that was created.

I think it could look something like:

```yaml
portForwarding:
  - labelSelector:
      app: my-app
      component: backend
    port: 2345
  - labelSelector:
      app: my-app
      component: frontend
    port: 8080:80
```

Skaffold could pick the most recently deployed pod that matches the selectors, although some more precise detection of the "right pod" may be necessary.

bhack:
I think that debugging it is a very important feature to be covered in dev. 
What is the debugging hipotesis of the Skaffold dev loop? How it is supposed currently to do debugging?

r2d4:
I'm going to close this with #945. There is #969 with some additional ideas around how the feature might work.

PR #141, opened 2018-03-07
Generate a manifest if one is not provided.

dlorenc:
Followup to https://github.com/GoogleCloudPlatform/skaffold/pull/122 after recreating my fork.

### DIFF
--- a/pkg/skaffold/deploy/kubectl.go
+++ b/pkg/skaffold/deploy/kubectl.go
@@ -78,7 +78,6 @@ func NewKubectlDeployer(cfg *config.DeployConfig) (*KubectlDeployer, error) {
 // Run templates the provided manifests with a simple `find and replace` and
 // runs `kubectl apply` on those manifests
 func (k *KubectlDeployer) Run(out io.Writer, b *build.BuildResult) (*Result, error) {
-
 	if len(k.DeployConfig.KubectlDeploy.Manifests) == 0 {
 		if len(b.Builds) != 1 {
 			return nil, errors.New("must specify manifest if using more than one image")
--- a/pkg/skaffold/deploy/kubectl.go
+++ b/pkg/skaffold/deploy/kubectl.go
@@ -20,8 +20,13 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
+	"text/template"
+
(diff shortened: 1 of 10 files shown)


pkg/skaffold/deploy/kubectl.go
```diff
@@ -31,6 +36,35 @@ import (
 	"github.com/pkg/errors"
 )
 
+// Slightly modified from kubectl run --dry-run
+var deploymentTemplate = template.Must(template.New("deployment").Parse(`apiVersion: extensions/v1beta1
```
r2d4: Whats the tutorial for this example if you don't have the ports exposed through a service?
Are we going to make people kubectl expose them?

You need to hit the endpoint to get the log messages

r2d4:
Last nit, everything looks great except I wonder if we need to supply the service too (or change the example)

pkg/skaffold/deploy/kubectl.go
dlorenc: I was thinking port-forward.

r2d4: Isn't the EXPOSE directive suggesting that they actually want the port exposed to some degree? 

I'm not sure if users who are getting started are going to know that they have to run a specific `kubectl port-forward` to access their service

dlorenc: It's a bit redundant I guess - you need to expose things at the pod level and then again later in a service/LB. I can create a service too - but creating a load balancer worries me a bit.

Basically anything that automatically opens up the cluster should be considered a little bit more carefully. Maybe we could prompt or something?

r2d4: Yeah I'm not sure, lets leave it as just a deployment now and then we'll try to get some feedback around if we should go ahead and template the service too.

dlorenc: Yeah maybe we can automate this in something like https://github.com/GoogleCloudPlatform/skaffold/issues/94

The main reason I'm hesitant is that I worry something like this would happen:
https://blog.heptio.com/on-securing-the-kubernetes-dashboard-16b09b1b7aca?gi=7a65c1c530e2

If someone creates something with an LB and forgets about it later.

r2d4: Absolutely.  Maybe we can also solve this from our side with automatic port forwarding #94

PR #143, opened 2018-03-07
Cache Image configs by name

dgageot:
This makes it possible to redeploy code faster by not having to lookup for an image’s config if we already fetched it.

### DIFF
--- a/pkg/skaffold/docker/parse.go
+++ b/pkg/skaffold/docker/parse.go
@@ -23,14 +23,12 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
-
-	"github.com/containers/image/types"
-
-	"github.com/containers/image/manifest"
-
-	"github.com/containers/image/docker"
+	"sync"
 
 	"github.com/GoogleCloudPlatform/skaffold/pkg/skaffold/util"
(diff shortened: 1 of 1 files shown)


pkg/skaffold/docker/parse.go
```diff
@@ -123,7 +121,14 @@ func processBaseImage(value *parser.Node) ([]string, error) {
 	return cfg.Config.OnBuild, nil
 }
 
+var imageToConfigCache sync.Map
+
 func retrieveImageConfig(image string) (*manifest.Schema2Image, error) {
+	cachedCfg, present := imageToConfigCache.Load(image)
```
r2d4: What if the image gets overwritten in the registry?

:latest, someone clobbering existing tag, etc.

dlorenc: Yeah I thought about that a bit, this kind of matches docker build semantics though. It won't pull an image for updates if it's already pulled locally unless you force it to.

dgageot: Exactly

r2d4: That makes sense, LGTM!

PR #147, opened 2018-03-08
docs: split verbose config template from getting-started example

### LINKED ISSUE #107: Separate example config YAML from the reference template
Right now the example's skaffold.yaml https://github.com/GoogleCloudPlatform/skaffold/blob/f321400fb678007d85a1df88f6dc8a0a9ed63f60/examples/getting-started/skaffold.yaml is the file that has the most documentation.

Consider creating a `skaffold.yaml.template` with that documentation, and reducing the example yaml to only the stuff that's uncommented.

This way:
- you can have 1 source of truth for documentation (rather than replicating it in each example)
- have brief examples that just work without requiring user to read every single field's description (less steep learning curve).

r2d4:
fixes #107

### DIFF
--- a/examples/annotated-skaffold.yaml
+++ b/examples/annotated-skaffold.yaml
@@ -1,0 +1,64 @@
+apiVersion: skaffold/v1
+kind: Config
+build:
+  # tagPolicy determines how skaffold is going to tag your image
+  # we provide a few strategies here, although you most likely won't need to care!
+  # Valid values for tagPolicy are
+  #
+  # sha256    - tags the image with the checksum of the built image (image id)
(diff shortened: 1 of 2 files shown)


dlorenc:
This makes sense, but where is the .tmpl used or linked from?

r2d4:
Nowhere, its just a nice verbose config for reference. Maybe it should be .yaml so that gets syntax highlighted

dlorenc:
> Nowhere, its just a nice verbose config for reference. Maybe it should be .yaml so that gets syntax highlighted

Yeah, if we're not actually using it as a template I'd keep it as yaml.

r2d4:
done ptal @dlorenc

PR #153, opened 2018-03-08
allow a tag to be passed into `skaffold run` on the CLI

### LINKED ISSUE #152: allow a tag to be provided on the command line of `skaffold run`
When folks want to make public releases of images/charts/yaml they may want more control over the image tags that are used for the docker images. 

So it might be nice to allow folks to pass that in on the CLI if they want. e.g. 

```
skaffold run --tag 1.2.3
```
From a quick noodle of the code it looks a fairly easy patch, adding an optional tag to the CLI `SkaffoldOptions`. Lemme submit a PR...

jstrachan:
so when creating releases users can have more control over the image tags used
fixes #152

### DIFF
--- a/cmd/skaffold/app/cmd/cmd.go
+++ b/cmd/skaffold/app/cmd/cmd.go
@@ -58,6 +58,7 @@ func NewSkaffoldCommand(out, err io.Writer) *cobra.Command {
 }
 
 func AddRunDevFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&opts.CustomTag, "tag", "t", "", "The optional custom tag to use for images which overrides the current Tagger configuration")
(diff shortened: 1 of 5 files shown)


cmd/skaffold/app/cmd/cmd.go
```diff
@@ -58,6 +58,7 @@ func NewSkaffoldCommand(out, err io.Writer) *cobra.Command {
 }
 
 func AddRunDevFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&opts.CustomTag, "tag", "t", "", "The optional custom tag to use for images which overrides the current Tagger configuration")
```
r2d4: These flags are shared between run and dev, maybe it makes sense just for `skaffold run`

jstrachan: good call - lemme move it to run only flags...

jstrachan:
@r2d4 sounds good to me

PR #159, opened 2018-03-08
dockerfile: no dependency on remote file

### LINKED ISSUE #158: Dockerfile ADD directive is not correctly handled
Hello, I think I found a bug in skaffold dockerfiles watcher.  In my Dockerfile I have ADD directive that tries to copy from a remote url a file:
```
# My Dockerfile
ADD https://repo.percona.com/apt/percona-release_0.1-5.stretch_all.deb /tmp/
```
In the [ADD](https://docs.docker.com/engine/reference/builder/#add) doc is specified that: `The ADD instruction copies new files, directories or remote file URLs from <src> and adds them to the filesystem of the image at the path <dest>.`

This is the error that I get:
```
run: running skaffold steps: running watch: adding deps for artifact: getting dockerfile dependencies: expanding dockerfile paths: getting file info: stat https:/repo.percona.com/apt/percona-release_0.1-5.stretch_all.deb: no such file or director
```

Thank you

r2d4:
fixes #158

### DIFF
--- a/pkg/skaffold/docker/parse.go
+++ b/pkg/skaffold/docker/parse.go
@@ -220,8 +220,13 @@ func processCopy(workspace string, value *parser.Node, paths map[string]struct{}
 		if hasMultiStageFlag(value.Flags) {
 			return nil
 		}
-		dep := path.Join(workspace, src)
-		paths[dep] = struct{}{}
+		if !strings.HasPrefix(src, "http://") && !strings.HasPrefix(src, "https://") {
+			dep := path.Join(workspace, src)
+			paths[dep] = struct{}{}
+		} else {
+			logrus.Debugf("Skipping watch on remote dependency %s", src)
+		}
+
 		value = value.Next
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


pkg/skaffold/docker/parse.go
```diff
@@ -220,8 +220,13 @@ func processCopy(workspace string, value *parser.Node, paths map[string]struct{}
 		if hasMultiStageFlag(value.Flags) {
 			return nil
 		}
-		dep := path.Join(workspace, src)
-		paths[dep] = struct{}{}
+		if !strings.HasPrefix(src, "http://") && !strings.HasPrefix(src, "https://") {
```
dlorenc: Maybe use url.Parse and check the Scheme? instead of checking both of these directly?

r2d4: this is one less step, since i'd have to do this string comparison on the scheme anyways right?

dlorenc: Yeah I meant more of checking whether or not there is a scheme, but that function doesn't really make sense to me now that I look at it. This is fine.

PR #168, opened 2018-03-09
Add options for GCR auth

dekkagaijin:
Add references to installing `docker-credential-{gcr, gcloud}`

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,11 @@ You will need the following components to get started with Skaffold:
 1. docker
 
 1. Docker image registry
-   -  Your docker client should be configured to push to an external docker image repository. If you're using a minikube or Docker for Desktop cluster,
-      you can skip this requirement.
-   -  If you are using Google Container Registry (GCR), run: `gcloud docker -a`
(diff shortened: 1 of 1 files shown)


viglesiasce:
Thanks @dekkagaijin!! Should we just be recommending the `gcloud auth configure-docker`? It seems like that one is the best option. Any gotchas?

dekkagaijin:
> Should we just be recommending the gcloud auth configure-docker? It seems like that one is the best option.

I'd guess that that should work for 99% of users doing local development and pushing to GCR, but  `docker-credential-gcr` exists for those that don't want to take a dependency on `gcloud` (and python).

> Any gotchas?

It's currently in beta, pending GA release for Docker 18.03 (or whichever version thereafter ends up being stable). Fixed the command to reflect that.

viglesiasce:
Awesome. Maybe once that goes GA we can remove the other two and refer out to the GCP docs for "other options".

dekkagaijin:
sgtm, I'll swing around when it lands

PR #171, opened 2018-03-09
Fixed Kubernetes context for build and deploy

### LINKED ISSUE #165: Use the same kubectl context in a skaffold session
I just switched my kubectl context (through side-effect command `kubectl config set-context`) while I was in the middle of a `skaffold dev` session. So I switched from `minikube` to my PROD cluster.

Next thing I know I modified a file in my repo and `skaffold dev` deployed destroyed my PROD.

I think we need to:
1. discover current kubectl context (`kubectl config current-context`) once
2. use this value like `kubectl --context=$VAL [...cmd]` until the process dies.

`skaffold dev` should not be changing its target cluster based on what's configured on kubectl. I think this way is more intuitive. We can support this by modifying logs slightly, such as:

    Deploy starting... (minikube)
    Deploy complete. (minikube)

dgageot:
Memoize the context and pass it to every kubectl and helm command.
Fixes #165

I didn't want to introduce a SkaffoldSession that would hold that kind of information
because right now there's only the kubecontext. WDYT?

### DIFF
--- a/pkg/skaffold/runner/runner.go
+++ b/pkg/skaffold/runner/runner.go
@@ -50,11 +50,17 @@ var kubernetesClient = kubernetes.GetClientset
 
 // NewForConfig returns a new SkaffoldRunner for a SkaffoldConfig
 func NewForConfig(opts *config.SkaffoldOptions, cfg *config.SkaffoldConfig) (*SkaffoldRunner, error) {
-	builder, err := getBuilder(&cfg.Build)
+	kubeContext, err := kubernetes.CurrentContext()
+	if err != nil {
+		return nil, errors.Wrap(err, "getting current cluster context")
+	}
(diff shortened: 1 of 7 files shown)


pkg/skaffold/runner/runner.go
```diff
@@ -50,11 +50,17 @@ var kubernetesClient = kubernetes.GetClientset
 
 // NewForConfig returns a new SkaffoldRunner for a SkaffoldConfig
 func NewForConfig(opts *config.SkaffoldOptions, cfg *config.SkaffoldConfig) (*SkaffoldRunner, error) {
-	builder, err := getBuilder(&cfg.Build)
+	kubeContext, err := kubernetes.CurrentContext()
```
r2d4: Nice, I didn't like that this was buried in the NewLocalBuilder.

Issue #36, opened 2018-02-06
Conditionally rebuild artifacts from watch


r2d4:
Currently, all artifacts are rebuilt after one source file changes, and changes during the build and deploy pipeline are ignored.

dgageot:
Instead of triggering a build on the first change event, we could wait for say 500ms and trigger a build only if no event has been received since.

Then we need to decide if a build should be interrupted in a change event occurs after it's started.

dgageot:
@r2d4 This should be fixed now, right?

PR #173, opened 2018-03-09
[WIP] Improve rebuild

dgageot:
Related to #36

Here's what I think should be done:

 + [X] Trigger a build only after a quiet period to avoid triggering a build while lots of file are being written.
 + [ ] Make sure that changes that happen while we build and deploy are not lost. aka always deploy the last changes.
 + [ ] Handle build cancellation. If I change a file every 2s and the build takes 10s, don't I want to just actually build the last change.


dgageot:
@r2d4 This should be good for a review. It can be improved but it's already way better like this.

pkg/skaffold/runner/runner.go
```diff
@@ -135,32 +135,37 @@ func (r *SkaffoldRunner) Run() error {
...
-		if err != nil {
-			return errors.Wrap(err, "running watch")
-		}
-		if evt.EventType == watch.WatchStop {
+		fmt.Println("WAIT") // ??
+		// Wait for next change
+		evt := <-r.fileChanges
+		fmt.Println(evt.EventType) // ??
+		switch {
+		case evt.EventType == watch.WatchError:
```
r2d4: I'm not sure the watch.Event is the right design.

WatchStart and WatchStop are "meta" events without an artifact
WatchError and WatchChange correspond to a single artifact, although they contain multiple "changed artifacts"

Should we get as granular to return the actual notify events? (create, delete, change) We might need this if we're going to start working on "smarter" context uploads and streaming.

I think I did this initially to cut down on how many channels we needed to pass around (one for top level watch events, and one for individual changes)

Do we expose the file changes channel separately from the watch ready/not ready/canceled channel?

pkg/skaffold/kubernetes/log.go
```diff
@@ -55,33 +55,36 @@ func StreamLogs(out io.Writer, client corev1.CoreV1Interface, image string) erro
...
-				rc, err := getStream(req)
-				if err != nil {
-					return errors.Wrap(err, "setting up container log stream")
-				}
-				defer rc.Close()
-				header := fmt.Sprintf("[%s %s]", p.Name, c.Name)
-				if err := streamRequest(out, header, rc); err != nil {
-					return errors.Wrap(err, "streaming request")
-				}
+			if c.Image != image {
```
r2d4: I think you can rebase these changes now that the other PR is merged.

dgageot:
@r2d4 I've changed quite a few things. If you had time to review, that would be awesome

dgageot:
🕺

PR #190, opened 2018-03-10
Add CI on osx

dgageot:
Let's see if it works...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,12 @@
 language: go
-os: linux
+os:
+  - linux
+  - osx
 go:
   - "1.10.x"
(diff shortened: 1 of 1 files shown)


r2d4:
This is an error related to symlinks.  The temp dir in osx is actually symlinked and that messes up some of the filesystem watching.  Ideally we shouldn't get to this point since we filter out all the symlinks before it gets to "watch", but we should still fix it.

dgageot:
@r2d4 I fixed the symlink issue in https://github.com/GoogleCloudPlatform/skaffold/pull/173/files#diff-cb57a8326c4a77302035078732d760f8R185. This seems to work great now.

PR #197, opened 2018-03-13
Mute logs during builds

### LINKED ISSUE #166: Stop streaming logs while build/deploy
I realized logs are still streamed from the containers while `skaffold dev` is doing image rebuilds and deploys. 

This is reducing the build/deploy output legibility since the frequent log lines (such as errors) coming from the applications themselves are interleaving with the `docker build` output etc.

We probably should consider suppressing the logs from applications during build/deploy.

dgageot:
Fixes #166

### DIFF
--- a/pkg/skaffold/kubernetes/log.go
+++ b/pkg/skaffold/kubernetes/log.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


dgageot:
@r2d4 yeah. I choose to simply mute the logs and let them die and be recreated as before. That decouples the muting from the rest.

PR #215, opened 2018-03-17
[WIP] Replace git with go-git

dgageot:
Replace `git` usage with [go-git](https://github.com/src-d/go-git)


dgageot:
@r2d4 @dlorenc should be good for review! 🐱

Issue #193, opened 2018-03-12
Handle deploy for multiple images(artifacts) in helm values


AMecea:
Hello, I am working on a project that has multiple docker images and I pass those images to helm through chart values. For the first time the `dev` command works well, but when I modify some files that trigger the build for only one artifact the deploy fails with the following warning:
```
WARN[0069] run: deploy step: deploying test: matching build results to chart values: No build present for image1
```
But I was expecting to deploy the new values with the new image. I think that the function [JoinTagsToBuildResult](https://github.com/GoogleCloudPlatform/skaffold/blob/master/pkg/skaffold/deploy/deploy.go#L38) should match the builds with the chart values not the other way.

skaffold.yaml:
```
apiVersion: skaffold/v1
kind: Config
build:
  tagPolicy: sha256
  artifacts:
  - imageName: image1
    dockerfilePath: hack/docker/toolbox/Dockerfile
    workspace: .
  - imageName: image2
    workspace: .
  local: {}
deploy:
  helm:
    releases:
    - name: test
      chartPath: hack/charts/controller
      values:
        image: image1
        conf.toolbox-image: image2
```
skaffold version is `v0.2.0`

Thank you

r2d4:
I'm going to call this something of a bug since its a regression since we've added conditional builds.  Basically, on the first run, everything is built but on the next run, only the changed artifacts are build. The deploy actually needs to know about the latest of _all_ built artifacts so that we can template the case of many builds to one deployment.

AMecea:
With version `0.3.0` half of the issue is solved. But now if I have extra values for the helm chart that are not images and are not specified in artifacts: the deploy fails.

Config example:
```
apiVersion: skaffold/v1
kind: Config
build:
  tagPolicy: sha256
  artifacts:
  - imageName: image1
    dockerfilePath: Dockerfile
  local: {}
deploy:
  helm:
    releases:
    - name: test
      chartPath: hack/charts/controller
      values:
        image: image1
        something: else
```
Error message:
```
Helm release test not installed. Installing...WARN[0001] run: deploy: deploy step: deploying test: matching build results to chart values: No build present for else
```

Do I miss something? How can I include extra chart configs?

Thank you!

booninite:
We're doing it via `setValues` for now, if you're still having the issue @AMecea

priyawadhwa:
I'd suggest trying out a more recent version of skaffold; we've fixed a bunch of bugs since `v0.2.0` :) 

[...]

PR #223, opened 2018-03-18
Remember deployed tags

dgageot:


pkg/skaffold/deploy/helm.go
```diff
@@ -29,16 +29,14 @@ import (
 )
 
 type HelmDeployer struct {
-	*config.DeployConfig
-	kubeContext string
+	*KubectlDeployer
```
r2d4: Why does HelmDeployer embed KubectlDeployer?

pkg/skaffold/deploy/kubectl.go
```diff
@@ -65,6 +65,8 @@ spec:
 
 type KubectlDeployer struct {
 	*config.DeployConfig
+
+	buildTags   *BuildTags
```
r2d4: Should buildTags go on SkaffoldRunner instead?

pkg/skaffold/deploy/helm.go
dgageot: I thought it made sense but I'll revert it

pkg/skaffold/deploy/kubectl.go
dgageot: I can try but I'm not sure it makes sense

dgageot:
@r2d4 Should be better now. I realized I'm not a big fan of passing a `BuildResult` to a `Deployer`.

r2d4:
Fixes #193

pkg/skaffold/runner/runner.go
```diff
@@ -166,24 +167,69 @@ func (r *SkaffoldRunner) dev(artifacts []*config.Artifact) error {
...
 	}
+
 	fmt.Fprintln(r.opts.Output, "Deploy complete in", time.Now().Sub(start))
 
-	return bRes, nil, nil
+	return dRes, nil
+}
+
+func mergeWithPreviousBuilds(builds, previous []build.Build) []build.Build {
+	updatedBuilds := map[string]bool{}
```
r2d4: any issue with just putting map[string]build.Build on SkaffoldRunner instead of the slice?

dgageot: no, no issue. Do you think it would be easier/clearer?

dgageot:
@r2d4 I've rebased the PR

pkg/skaffold/runner/runner.go
r2d4: Nope, might be slightly cleaner, but shouldn't really make much of a difference.

Issue #218, opened 2018-03-17
skaffold is slow to start


dgageot:
Each time skaffold is started, it retrieves the config for every base image used by the artifacts.
Most of the time, those images are local. If there's a way to use the local data, we should do so.

dlorenc:
We should definitely do this when using a local builder. It's not only a performance thing, it's also about correctness. You could have a local image that has the same name but a different configuration from a remote one.

dlorenc:
Though if we're using a cloud builder we need to check the registry.

dgageot:
This is fixed by https://github.com/GoogleCloudPlatform/skaffold/pull/224

PR #224, opened 2018-03-18
Search for image locally

dgageot:
This does the job of fixing #218 on Docker for Desktop but needs more work to work on minikube and other platforms.

### DIFF
--- a/pkg/skaffold/docker/client.go
+++ b/pkg/skaffold/docker/client.go
@@ -24,8 +24,10 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/GoogleCloudPlatform/skaffold/pkg/skaffold/constants"
+	"github.com/GoogleCloudPlatform/skaffold/pkg/skaffold/kubernetes"
 	"github.com/GoogleCloudPlatform/skaffold/pkg/skaffold/util"
 	"github.com/docker/docker/api"
 	"github.com/docker/go-connections/tlsconfig"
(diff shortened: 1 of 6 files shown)


pkg/skaffold/docker/parse.go
```diff
@@ -158,50 +159,73 @@ func processBaseImage(value *parser.Node) ([]string, error) {
...
+	}
+
+	imageCache.Store(image, cfg)
+
+	return cfg, nil
+}
+
+func retrieveLocalImage(image string) ([]byte, error) {
+	// TODO(dgageot): docker api might not be available or it could be minikube
+	api, err := NewImageAPIClient()
```
r2d4: This is tricky since we've made an implicit dependency between getting the docker daemon and the cluster context.

In the future, we should really do this clientside and not talk to the daemon, getting around this issue. But for now we should defer to the daemon for pull.

It makes sense to have a function that does the resolution, so other packages can call it. Essentially moving the switch statement outside local.go into its own function.

```golang
func NewImageAPIClient() (ImageAPIClient, error) {
// get kubectl context
// context switch statement
}
```

This makes the package dependency more explicit and has the question of where it should live. Its maybe not totally correct to have pkg/docker depend on pkg/kubernetes, but I don't think we have a choice by what we've chosen.

Alternatively, we could use a daemonless pull library like https://github.com/containers/image or https://github.com/google/go-containerregistry. But I think that can wait until we run into an import issue.

r2d4: That function could be memoized too, so that contexts dont change midway through a skaffold invocation

dgageot: Thanks @r2d4, I'll try something based on your idea. To be precise, we already use a daemonless pull library. It's just that it's both slow and incorrect compared to reading from the daemon's local storage.

dgageot:
@r2d4 I implemented your suggestion. Can you please TAL? Thanks!

pkg/skaffold/docker/client.go
```diff
@@ -39,18 +41,39 @@ type DockerAPIClient interface {
...
+var (
+	dockerAPIClientOnce sync.Once
+	dockerAPIClient     DockerAPIClient
+	dockerAPIClientErr  error
+)
+
 // NewDockerAPIClient guesses the docker client to use based on current kubernetes context.
-func NewDockerAPIClient(kubeContext string) (DockerAPIClient, error) {
+func NewDockerAPIClient() (DockerAPIClient, error) {
+	dockerAPIClientOnce.Do(func() {
```
r2d4: do both of these have to be wrapped in sync.Once? Only either the NewDockerAPIClient or CurrentContext should need to be memoized. I think just the NewDockerAPIClient would be sufficient, since current context should be quick enough (it just parses the kubeconfig and doesn't talk to the server afaik) whereas the new docker API client getter actually pings the docker endpoint

dgageot:
CurrentContext is used somewhere else and can be used in the future. It is
quick to compute but we don’t want it to change. So we memoize it.
Le mar. 27 mars 2018 à 19:40, Matt Rickard <notifications@github.com> a
écrit :

r2d4:
Yup I guess thats fair. Docker endpoint and kubectl context shouldn't change in a dev cycle

Issue #228, opened 2018-03-19
Color the logs


dgageot:
It would be nice to color the logs to differentiate artifacts.

PR #230, opened 2018-03-19
Store the creation date of the Log Aggregator

### LINKED ISSUE #229: First few lines of log are ignored
The logs that happen between deployment and the actual connection to the logs are lost. That's painful because the first few lines are usually important.

dgageot:
Fixes #229

### DIFF
--- a/pkg/skaffold/kubernetes/log.go
+++ b/pkg/skaffold/kubernetes/log.go
@@ -31,6 +31,26 @@ import (
 	restclient "k8s.io/client-go/rest"
 )
 
+const defaultRetry int = 5
+
+// LogAggregator aggregates the logs for all the deployed pods.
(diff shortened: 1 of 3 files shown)


pkg/skaffold/kubernetes/log.go
```diff
@@ -31,6 +31,26 @@ import (
...
 )
 
+const defaultRetry int = 5
+
+// LogAggregator aggregates the logs for all the deployed pods.
+type LogAggregator struct {
+	Muter
+
+	creationTime time.Time
+	output       io.Writer
```
r2d4: ref #228 We should think about passing some sort of OutputWriter/Logger struct here that does some nice coloring and formatting. It would be nice to reuse the same special writer in our output in run/deploy too.

PR #231, opened 2018-03-19
fix:(env vars) allow environment variables to be used for images

### LINKED ISSUE #184: support environment variable expressions for docker registries?
I'd like to be able to use an environment variable to describe the docker registry to use. e.g. in a `skaffold.yaml` something like:

```yaml
apiVersion: skaffold/v1
kind: Config
build:
  artifacts:
  - imageName: $MY_DOCKER_REGISTRY_SERVICE_HOST:$MY_DOCKER_REGISTRY_SERVICE_PORT/foo/bar
    workspace: .
  local: {}
deploy:
  kubectl:
    manifests:
```
So processing all references to environment variables in the image name before passing it to the underlying docker commands etc.

jstrachan:
so we can refer to docker registries via environment variables, e.g. if the docker registry is running inside a kubernetes cluster so that kubernetes service discovery does not work (as the local docker daemon is not running inside k8s)

fixes #184

### DIFF
--- a/pkg/skaffold/build/local.go
+++ b/pkg/skaffold/build/local.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 
 	"github.com/GoogleCloudPlatform/skaffold/pkg/skaffold/build/tag"
 	"github.com/GoogleCloudPlatform/skaffold/pkg/skaffold/config"
@@ -114,8 +115,9 @@ func (l *LocalBuilder) Build(out io.Writer, tagger tag.Tagger, artifacts []*conf
 		if digest == "" {
 			return nil, fmt.Errorf("digest not found")
 		}
+		imageName := os.ExpandEnv(artifact.ImageName)
 		tag, err := tagger.GenerateFullyQualifiedImageName(&tag.TagOptions{
-			ImageName: artifact.ImageName,
+			ImageName: imageName,
 			Digest:    digest,
 		})
 		if err != nil {
@@ -133,7 +135,7 @@ func (l *LocalBuilder) Build(out io.Writer, tagger tag.Tagger, artifacts []*conf
 			}
 		}
 		res.Builds = append(res.Builds, Build{
-			ImageName: artifact.ImageName,
+			ImageName: imageName,
 			Tag:       tag,
 			Artifact:  artifact,
 		})


jstrachan:
this PR would let [Jenkins X](http://jenkins-x.io/) use skaffold for building the docker images + pushing them - as OOTB in Jenkins X we're using a cluster local docker registry which can only be accessed via the service IP addresses since the docker daemon is not running in a pod, so we can't use kubernetes service discovery - though we can use service environment variables to find the docker registry

dlorenc:
I had a slightly different idea around this that I just started working on with #251.

The idea was to define an EnvironmentVariable tag policy, where users can specify a Go template string which will be executed to generate the image name. 

The main difference I see is that the template approach would also let you add your own custom formatting for the tag portion of the image itself, if you wanted to include a Jenkins release name, etc.

WDYT about that approach?

dlorenc:
Also, congrats on the Jenkins X launch. We're excited to help get it working great with Minikube, Skaffold and GKE.

jstrachan:
@dlorenc a template would be fine too! so long as there's some way to evaluate an env var I'm totally happy ;)

thanks for the kind words too!

dlorenc:
> @dlorenc a TagPolicy is for the image tag though right? The use case for environment variables is for the image name (in particular the docker registry prefix of the image).
> 
> The Tagger just adds the bit after the : in a docker image like: myregistry/myuser/image:1.2.3 right?

Somewhat confusingly the Tagger actually generates the entire image name. The interface is here:
https://github.com/GoogleCloudPlatform/skaffold/blob/master/pkg/skaffold/build/tag/tag.go#L21

And the SHA implementation: https://github.com/GoogleCloudPlatform/skaffold/blob/master/pkg/skaffold/build/tag/sha256.go#L31

An image name comes in as an input, because the build step needs to build it with some name, but the tagger could rename it to anything it wants before a push.

jstrachan:
@dlorenc aha, thanks! so that all sounds good to me then :)

dgageot:
@dlorenc @jstrachan I'm wondering if it doesn't make sense to support env variable placeholders anywhere in a `skaffold.yaml`, like in a compose file: https://docs.docker.com/compose/environment-variables/

jstrachan:
@dgageot that doc looks like you can refer to environment variables in image names right?

@dlorenc if we want to keep environment variable expressions out of the `skaffold.yaml` another option could be to allow the docker registry or image prefix to be passed in as a CLI argument? 

e.g. something like
```yaml
apiVersion: skaffold/v1alpha1
kind: Config
build:
  artifacts:
  - imageName: my-example
    workspace: .
```

then:

``` 
$ skaffold run -f skaffold.yaml --registry $MY_REGISTRY_HOST:$MY_REGISTRY_PORT
```
then it'd overwrite/add the registry prefix on the image name?

dgageot:
@jstrachan exactly! My point is "Why only for image names?". Why not anywhere in the `skaffold.yaml`?

jstrachan:
sounds good to me - happy to use environment variables in other places in the yaml. Could be handy for the image tag, user or image name too really

jstrachan:
I like the idea allowing environment expressions in string values in the YAML.

One possible downside to allowing expressions anywhere is that schema aware YAML editors start to complain about `$FOO` not being a number etc. So I wonder if restricting expressions to just string values in the YAML for now might help avoid future smart YAML editor issues?

e.g. it'd be good to add Skaffold's schema (via json schema generated from the model) to something like this: https://github.com/redhat-developer/vscode-yaml for a nicer editor experience

jordanwilson230:
@dgageot @jstrachan Agreed. I think this is a reasonable request and within scope given Skaffold's focus towards early-on/local development, and the fact that these are often the environments with the most (necessary) variability.

jstrachan:
I've just submitted a PR that lets environment variables be used as an expression anywhere in the YAML file: https://github.com/GoogleCloudPlatform/skaffold/pull/295 as an alternative approach to this PR

dgageot:
I guess we can close this one if #295 is merged

dgageot:
Closing as a duplicate of #295

PR #238, opened 2018-03-19
runner: add watch message when watching for changes

r2d4:
If the container doesn't immediately log, the tool appears to "hang" at "Depoy Complete.". We should communicate the to users that the tool is watching for changes.


dgageot:
LGTM

PR #243, opened 2018-03-20
.dockerignore doesn't work if context is in a parent directory

### LINKED ISSUE #237: .dockerignore doesn't work if context is in a parent directory
E.g. if this is the directory structure:
```
skaffold/
    docker/
        Dockerfile
        .dockerignore
        ignored.txt
    tests/
        run.go
```

Then running a build from `run.go` with workspace/context = `../docker` will result in `ignored.txt` being copied into the image.

This is because the code uses `fileutils.Matches` to identify files specified by .dockerignore. And a pattern of `ignored.txt` does not match `../docker/ignored.txt`.

The solution is to make sure all file paths are relative to the context before passing them to `ApplyDockerIgnore`

geniass:
so dockerignore works with parent contexts

Closes #237

### DIFF
--- a/pkg/skaffold/docker/parse.go
+++ b/pkg/skaffold/docker/parse.go
@@ -101,7 +101,10 @@ func GetDockerfileDependencies(workspace string, r io.Reader) ([]string, error) 
 	// Look for .dockerignore.
 	ignorePath := filepath.Join(workspace, ".dockerignore")
 
-	absWorkspacePath, _ := filepath.Abs(workspace)
+	absWorkspacePath, err := filepath.Abs(workspace)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting workspace absolute path")
+	}
 	var relPaths []string
 	for _, path := range expandedDeps {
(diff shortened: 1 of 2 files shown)


pkg/skaffold/docker/parse.go
```diff
@@ -101,10 +101,27 @@ func GetDockerfileDependencies(workspace string, r io.Reader) ([]string, error)
 	// Look for .dockerignore.
 	ignorePath := filepath.Join(workspace, ".dockerignore")
 
-	filteredDeps, err := util.ApplyDockerIgnore(expandedDeps, ignorePath)
+	absWorkspacePath, _ := filepath.Abs(workspace)
```
dlorenc: We should handle this error. Abs can return errors in a few cases.

geniass: Ah missed that one. Pushed a new commit

geniass:
@dlorenc Added the missing error check, but I noticed that the OSX test randomly failed. I don't have a Mac so I can't replicate it. Any idea what's wrong?

PR #260, opened 2018-03-21
Better logs

dgageot:
@r2d4 @dlorenc I've changed quite a lot how the logs are working. Now, it sets a watch up to listen to new pods. If a pod uses an image we built (we keep the tags), then the logs for *every* container in the pods are printed.

Also, I took the opportunity to use less random colors for the logs. Now: 1 artifact = 1 color. It's not perfect because a pod can use multiple artifacts. But it's getting better!

I've removed the unit tests for the logger because it's not worse maintaining them if you are not happy with the overall design. Could you please review the PR and then I add the tests back?


dgageot:
@r2d4 I've improve the code and fixed a few issues. Can you TAL? Thanks!

pkg/skaffold/kubernetes/log.go
```diff
@@ -28,132 +29,141 @@ import (
...
+				if evt.Type != watch.Added && evt.Type != watch.Modified {
+					continue
+				}
+
+				pod, ok := evt.Object.(*v1.Pod)
+				if !ok {
+					continue
+				}
+
+				color, present := a.podColor(pod)
```
r2d4: Its a little odd that podColor also returns if the pod matches any of the containers we're deploying. Could be a little more explicit here.

pkg/skaffold/kubernetes/log.go
```diff
@@ -28,132 +29,141 @@ import (
...
-	}
+	}()
+
+	return nil
 }
 
-func (a *LogAggregator) SetCreationTime(t time.Time) {
-	a.creationTime = t
+// RegisterImage register an image for which pods need to be logged.
+func (a *LogAggregator) RegisterImage(image string, color int) {
```
r2d4: we register new images in the logaggregator, but do we ever need to unregister them?

pkg/skaffold/kubernetes/log.go
```diff
@@ -28,132 +29,141 @@ import (
...
-// nolint: interfacer
-func (a *LogAggregator) streamLogs(client corev1.CoreV1Interface, image string) error {
-	pods, err := client.Pods("").List(meta_v1.ListOptions{
-		IncludeUninitialized: true,
-	})
-	if err != nil {
-		return errors.Wrap(err, "getting pods")
+// podColor reads the images used by a pod and compares them to the list
+// of registered colors.
+func (a *LogAggregator) podColor(pod *v1.Pod) (color int, found bool) {
```
r2d4: ideally log color wouldn't be so tightly coupled with actually choosing which logs to stream, but I think this is OK for now.

r2d4:
@dgageot feel free to merge once you rebase

dgageot:
@r2d4 It's rebased but I can't merge it

r2d4:
Ok I think it should be fixed now, there was another setting I needed to toggle

PR #276, opened 2018-03-23
Support pushing to docker Hub

### LINKED ISSUE #271: Handle docker auth to DockerHub
We currently don't inject dockerhub credentials for images without a `docker.io` prefix. This is something that gets resolved in the docker CLI, so I think its something we need to handle here. 

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/skaffold-users/fmnS3y3zOno/6gG6rdnaCAAJ

dgageot:
Fix #271

### DIFF
--- a/pkg/skaffold/docker/auth.go
+++ b/pkg/skaffold/docker/auth.go
@@ -17,6 +17,7 @@ limitations under the License.
 package docker
 
 import (
+	"context"
 	"encoding/base64"
 	"encoding/json"
 	"os"
@@ -27,6 +28,7 @@ import (
 	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/registry"
 	"github.com/moby/moby/pkg/homedir"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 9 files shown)


pkg/skaffold/docker/auth.go
```diff
@@ -27,13 +27,14 @@ import (
...
+	"github.com/docker/docker/registry"
 	"github.com/moby/moby/pkg/homedir"
 	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
 )
 
 const (
-	configFileDir = ".docker"
+	configFileDir    = ".docker"
+	dockerHubAuthKey = "ttps://index.docker.io/v1/"
```
tuananh: missing a `h` in the link `ttps://index.docker.io/v1/` ?

dgageot: :-)

dgageot:
@tuananh Thanks for your help. Should be better now

pkg/skaffold/docker/client.go
```diff
@@ -34,23 +34,23 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-type ImageAPIClient interface {
-	client.ImageAPIClient
+type DockerAPIClient interface {
+	client.CommonAPIClient
```
r2d4: We could lock this down to ImageAPIClient + RegistryAPIClient, but I think this is fine.

PR #281, opened 2018-03-23
Add integration tests

r2d4:


r2d4:
I'm working on exposing the logs publicly, but this should be ready now. I'll follow up with some docs to run this locally with a GKE cluster or minikube.

deploy/skaffold/Dockerfile
```diff
@@ -12,19 +12,24 @@
...
+        curl \
+        build-essential \
+        git \ 
+        gcc \
+        python-dev \
+        python-setuptools \
+        apt-transport-https \
+        lsb-release
+
+COPY --from=golang:1.10 /usr/local/go /usr/local/go
```
dlorenc: Go isn't in the ubuntu package manager anywhere?

deploy/skaffold/Dockerfile
```diff
@@ -35,13 +40,28 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}
     tar -zxvf helm-${HELM_VERSION}-linux-amd64.tar.gz && \
     mv linux-amd64/helm /usr/local/bin/helm
 
-FROM gcr.io/distroless/base 
-WORKDIR /root/
+ENV CLOUD_SDK_VERSION 193.0.0
+RUN easy_install -U pip && \
+    pip install -U crcmod && \
```
dlorenc: Do we use gsutil anywhere? We only need crcmod to make some gsutil operations faster.

deploy/skaffold/Dockerfile
```diff
@@ -35,13 +40,28 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}
...
+    export CLOUD_SDK_REPO="cloud-sdk-$(lsb_release -c -s)" && \
+    echo "deb https://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" > /etc/apt/sources.list.d/google-cloud-sdk.list && \
```
dlorenc: I'm not a big fan of using the deb installer for the SDK, it makes extra components weird. Can you use the tarball instead? https://cloud.google.com/sdk/downloads#versioned

integration/run_test.go
```diff
@@ -19,31 +19,58 @@ limitations under the License.
...
 
+var gkeZone = flag.String("gke-zone", "us-central1-a", "gke zone")
+var gkeClusterName = flag.String("gke-cluster-name", "integration-tests", "name of the integration test cluster")
+var gcpProject = flag.String("gcp-project", "k8s-skaffold", "the gcp project where the integration test cluster lives")
+var remote = flag.Bool("remote", false, "if true, run tests on a remote GKE cluster")
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+	if *remote {
+		cmd := exec.Command("gcloud", "container", "clusters", "get-credentials", *gkeClusterName, "--zone", *gkeZone, "--project", *gcpProject)
```
dlorenc: Should we use our exec.Command helper instead?

deploy/skaffold/Dockerfile
r2d4: We can't ensure that we use the same version that way

deploy/skaffold/Dockerfile
r2d4: No, I can remove this. Wasn't sure if we were going to need upload anything once we were inside running the tests

integration/run_test.go
r2d4: The helper is used on the next line below, we don't do anything special for creating the cmd, just running

deploy/skaffold/Dockerfile
dlorenc: Ah, makes sense.

integration/run_test.go
dlorenc: D'oh, I was thinking of the special one we have in minikube for integration tests. Nevermind.

Makefile
```diff
@@ -82,3 +87,16 @@ release: cross
...
+
+.PHONY: integration-in-docker
+integration-in-docker:
+	docker build \
+		-f deploy/skaffold/Dockerfile \
+		-t gcr.io/$(GCP_PROJECT)/skaffold-integration .
+	docker run \
+		-v /var/run/docker.sock:/var/run/docker.sock \
+		-v $(HOME)/.config/gcloud:/root/.config/gcloud \
+		-v $(PWD):/go/src/github.com/GoogleCloudPlatform/skaffold \
```
dlorenc: nit: reuse the variables for GoogleCloudPlatform/skaffold?

deploy/skaffold/Dockerfile
r2d4: Yeah, I'll do that

deploy/skaffold/Dockerfile
r2d4: fixed

Makefile
r2d4: fixed

r2d4:
This passed, re-running with an internal config change for public logs.

r2d4:
@dlorenc I think this is ready to merge.

A note to users, we're working on getting approval to make the logs public, please bear with us. In the meantime, I'll make sure that there are clear directions on how to run the tests locally.

PR #286, opened 2018-03-26
Allows for setting arbitrary values via helm's --set flag. 

### LINKED ISSUE #285: Allow for adding arbitrary values to helm deploy via helm's --set flag
Currently the only values that can be added to a helm deploy with Skaffold are the build container names and tags. Since many helm deployments that are managed via CI use helm's --set flag to set values, it would be best if Skaffold could do the same.

Additionally, supporting environment variables in these values would allow for a more dynamic deployment.

gmorse81:
Support's replacement of environment variables in the values.

fixes #285


gmorse81:
I signed it!

gmorse81:
Sure, i can update the PR. Maybe there is another way to accomplish what i am looking to do... 

for local dev, i am running the mysql image from docker hub. it will load a database from a sql dump if you place the .sql.gz file in a specific directory on the container instance. The easiest way to accomplish this is via a volume mount to the host machine. 

Since the developers i work with may be cloning my project to any location on their local machines, but i need to reliably mount a hostPath which contains the sql dump, i am using the $PWD env var to get the base path of the project. the dev can then put the sql dump file into a directory at $PWD/database-dump and then i can mount that directory. 

so, really all i need is the ability to get the absolute path to the location of the skaffold.yaml file. I don't really need the whole environment.

PR #288, opened 2018-03-27
No template in kubectl manifests

r2d4:
This is a breaking config change. It allows users to not have to template their kubectl manifests for the basic deployer. Simply just leave the image name and it will be replaced. Still not fully correct, but not any worse than what existed before.


r2d4:
We could even take this a step further an replace line by line, only replacing when the key is "image" and the tag matches.

dgageot:
@r2d4 I'm all for this change. I find it easier to use. I have a few questions though:
 + Can this be applied to the help deployer too? If it can, we could get rid of `JoinTagsToBuildResult` function.
 + Why list the image names in the deploy section when they are all listed in the build section?

dgageot:
@r2d4 in fact reading your code more carefully, I realized that the `Images` field in `Manifest` struct is only used by the unit tests! It's unused by the rest of the code than instead seems to get the list of images from the build step. Should we get rid of that piece of confirmation altogether?

r2d4:
Good catch :)

For helm we might be able to do something like this, but that would involve mutating the values.yaml file instead of using the --set runtime flag.  Basically we replace the image with --set key=tag at runtime, but if the user has multiple charts we'll get an error if we try to substitute something thats not present.

Well before when we were replacing by key, this meant that users could use the same key across multiple manifests, and we would only substitute the right one. Now, we can substitute all the artifacts, since theres no ambiguity.

PR #295, opened 2018-03-28
fix:(env vars) allow environment variables to be used in the yaml

### LINKED ISSUE #184: support environment variable expressions for docker registries?

jstrachan:
so we can refer to docker registries via environment variables, e.g. if the docker registry is running inside a kubernetes cluster so that kubernetes service discovery does not work (as the local docker daemon is not running inside k8s)

this approach lets us use environment variable expressions anywhere in the YAML

fixes #184

### DIFF
--- a/pkg/skaffold/config/config.go
+++ b/pkg/skaffold/config/config.go
@@ -19,8 +19,9 @@ package config
 import (
 	"github.com/GoogleCloudPlatform/skaffold/pkg/skaffold/constants"
 
-	yaml "gopkg.in/yaml.v2"
 	"os"
+
+	yaml "gopkg.in/yaml.v2"
 )
 
 // SkaffoldConfig is the top level config object
--- a/pkg/skaffold/config/config.go
+++ b/pkg/skaffold/config/config.go
@@ -20,6 +20,7 @@ import (
 	"github.com/GoogleCloudPlatform/skaffold/pkg/skaffold/constants"
 
 	yaml "gopkg.in/yaml.v2"
+	"os"
 )
 
 // SkaffoldConfig is the top level config object
@@ -125,7 +126,9 @@ var DefaultRunSkaffoldConfig = &SkaffoldConfig{
 // The default config argument provides default values for the config,
 // which can be overridden if present in the config file.
 func Parse(config []byte, defaultConfig *SkaffoldConfig) (*SkaffoldConfig, error) {
-	if err := yaml.Unmarshal(config, defaultConfig); err != nil {
+	// lets expand any environment variable expressions in the config
+	expanded := os.ExpandEnv(string(config))
+	if err := yaml.Unmarshal([]byte(expanded), defaultConfig); err != nil {
 		return nil, err
 	}
 


dgageot:
@jstrachan Can you please run `goimports -w pkg/skaffold/config/config.go` to make sure the imports are sorted in `config.go`?

dgageot:
@r2d4 I think we should acknowledge that users will want to use env variables anywhere, not just for registry names. I know this can lead to very weird config files but I'm ok with that. wdyt?

jstrachan:
@dgageot fixed the imports. 

I agree it should be fine to allow environment expressions anywhere in the YAML. If we hit issues further down the road we could investigate a more YAML-syntax aware replacement strategy if needed.

dgageot:
@jstrachan Can you give a try to the new `envTemplate` tagger? I think it'll solve your use case.

jstrachan:
the `envTemplate` works great thanks!

One thing I hit my head on first time I tried using it was any registry/group in the `envTemplate` gets appended where the image name is if there is already a registry/group in there. e.g.
```yaml 
  tagPolicy:
    envTemplate:
      template: "myregistry/myuser/{{.IMAGE_NAME}}:{{.FOO}}"
  artifacts:
    workspace: .
```
results in...
```
Successfully tagged myregistry/myuser/gcr.io/k8s-skaffold/skaffold-example:123
```
so if you are going to use a registry or group in the `envTemplate` you have to ensure you remove the registry/group in the `imageName` too.

e.g. this works fine:
```
build:
  tagPolicy:
    envTemplate:
      template: "myregistry/myuser/{{.IMAGE_NAME}}:{{.FOO}}"
  artifacts:
  - imageName: changeme
    workspace: .
...
Successfully tagged myregistry/myuser/changeme:123
```
I guess one day the `envTemplate` could be smart enough to recognise how many `/` characters there are in the generated expression and replace either the full expression, the group + image name or just the image name? Not a huge biggie though

mailaneel:
Hi, just curious to know the reason for not getting merged? is there any alternative instead of this fix. The use case I have is to pass env variables in buildArgs ($NPM_TOKEN)

```
    docker:
      buildArgs:
        NPM_TOKEN: ${NPM_TOKEN}
```

scottfuoco:
> Hi, just curious to know the reason for not getting merged? is there any alternative instead of this fix. The use case I have is to pass env variables in buildArgs ($NPM_TOKEN)
> 
> ```
>     docker:
>       buildArgs:
>         NPM_TOKEN: ${NPM_TOKEN}
> ```

Is there a good solution to this?  I have the same use case currently using v0.23.0.  Right now we are leaving the npm token blank in the skaffold.yaml and manually filling it in when we pull down our repo.

PR #307, opened 2018-03-30
integration: mount in application credentials path and env var

r2d4:


dlorenc:
Do you want to merge this as is or make it safe for when that variable isn't set first?

r2d4:
You can still run integration tests locally with `make integration`, this is just for the dockerized version, which only supports GKE right now.

PR #316, opened 2018-04-02
Add an environment variable template tagger.

dlorenc:
There was a bug/documentation issue with the existing tagPolicy struct. it should not have been inline.

Tests are coming, i want to get some feedback on the interface first.

This allows users to specify a golang template string that uses the Environment to generate tags for images.

I also merge in the opts.TagOptions values as well, for multi-image scenarios. This needs to be documented.

### DIFF
--- a/integration/run_test.go
+++ b/integration/run_test.go
@@ -20,6 +20,7 @@ package integration
 
 import (
 	"flag"
+	"fmt"
 	"os"
 	"os/exec"
 	"testing"
@@ -70,6 +71,7 @@ func TestRun(t *testing.T) {
 		extraArgs   []string
 		deployments []testObject
 		pods        []testObject
+		env         map[string]string
 
 		remoteOnly bool
 	}
@@ -106,6 +108,18 @@ func TestRun(t *testing.T) {
 			dir:       "../examples",
 			extraArgs: []string{"-f", "annotated-skaffold.yaml"},
 		},
+		{
+			description: "getting-started envTagger",
+			pods: []testObject{
+				{
+					name:      "getting-started",
+					namespace: "default",
+				},
+			},
+			dir:       "../examples/environment-variables",
+			extraArgs: []string{"-f", "skaffold-env-tag.yaml"},
+			env:       map[string]string{"FOO": "foo"},
+		},
 		// // Don't run this test for now. It takes awhile to download all the
 		// // dependencies
 		// {
@@ -140,6 +154,11 @@ func TestRun(t *testing.T) {
 			args := []string{"run"}
 			args = append(args, testCase.extraArgs...)
 			cmd := exec.Command("skaffold", args...)
+			env := os.Environ()
+			for k, v := range testCase.env {
+				env = append(env, fmt.Sprintf("%s=%s", k, v))
+			}
+			cmd.Env = env
 			cmd.Dir = testCase.dir
 			out, outerr, err := util.RunCommand(cmd, nil)
 			if err != nil {
--- a/pkg/skaffold/build/tag/env_template.go
+++ b/pkg/skaffold/build/tag/env_template.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2018 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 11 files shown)


dlorenc:
cc @jstrachan @dgageot What do you think of this version instead of #295

dgageot:
@dlorenc I do prefer supporting env variables everywhere but I'm ok with whatever unlocks @jstrachan 's use case. 

To be honest, private registries and image urls are one of the worst things in Docker in terms of UX. It's always an issue. No matter what we do, the user will have to change the skaffold.yaml, set a variable or a flag.

I'm all for supporting env variables everywhere (such as what #295 does) because I've seen people using that kind of things over and over again with compose. I also understand that we might want to not give that dangerous power to our users.

jstrachan:
@dlorenc I don't much mind really, so long as I can use environment variables in the image name for docker registry/organisation/image I'm happy ;)

pkg/skaffold/build/tag/env_template.go
```diff
@@ -0,0 +1,67 @@
...
+	envMap := map[string]string{}
+	for _, env := range os.Environ() {
+		kvp := strings.SplitN(env, "=", 2)
+		if len(kvp) != 2 {
+			return "", fmt.Errorf("error parsing environment variables, %s does not contain an =", kvp)
+		}
+		envMap[kvp[0]] = kvp[1]
+	}
+
+	envMap["ImageName"] = opts.ImageName
```
r2d4: How are these discoverable? Can you add it to the annotated example or something? Or start a doc for taggers?

pkg/skaffold/build/tag/env_template.go
```diff
@@ -0,0 +1,67 @@
...
+	envMap["Digest"] = opts.Digest
+
+	logrus.Debugf("Executing template %v with environment %v", c.Template, envMap)
+	if err := c.Template.Execute(&buf, envMap); err != nil {
```
r2d4: I think its a little odd to be using a go-template here, especially since that suggests that theres a template struct somewhere.

I would be in favor of using bash style syntax, but in the interest of reducing complexity, this is fine. We might be able to reuse the docker shell lexer here with minimal effort. Not sure.

pkg/skaffold/runner/runner.go
```diff
@@ -113,6 +113,10 @@ func getDeployer(cfg *config.DeployConfig, kubeContext string) (deploy.Deployer,
 }
 
 func newTaggerForConfig(t config.TagPolicy) (tag.Tagger, error) {
+	logrus.Error(t.EnvTemplateTagger)
```
r2d4: remove

dlorenc:
> I don't think this needs an example, but maybe a short integration test?

Yeah definitely needs some tests. Do people prefer this approach or the other one?

dlorenc:
Added a unit test. I'll do an integration test next.

pkg/skaffold/config/config.go
```diff
@@ -35,15 +35,16 @@ type SkaffoldConfig struct {
 
 // BuildConfig contains all the configuration for the build steps
 type BuildConfig struct {
-	Artifacts []*Artifact `yaml:"artifacts,omitempty"`
-	TagPolicy TagPolicy   `yaml:"tagPolicy,omitempty"`
+	Artifacts []*Artifact `yaml:"artifacts"`
```
dgageot: Please don't remove the `omitempty` on those two lines. It's useful

dlorenc: Ah sorry, merge conflict issue.

dlorenc:
@r2d4 This should be RFAL.

integration/run_test.go
```diff
@@ -140,6 +154,9 @@ func TestRun(t *testing.T) {
 			args := []string{"run"}
 			args = append(args, testCase.extraArgs...)
 			cmd := exec.Command("skaffold", args...)
+			for k, v := range testCase.env {
+				cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", k, v))
```
r2d4: if nil this will use the process's env

if not, it will only use what you pass in. maybe we can just use the Setenv test function we have

dgageot:
I'm going to merge this one and fix the nitpicks in another PR

PR #327, opened 2018-04-03
Read tagPolicy from config

### LINKED ISSUE #326: tagPolicy is not parsed from config file
### Expected behavior
Skaffold should use the tagPolicy in the config file. 
### Actual behavior
tagPolicy is not read properly, so the default value is used.
### Information

Skaffold version: v0.3.0
Operating system: Fedora 27
Content of skaffold.yaml:
```
apiVersion: skaffold/v1alpha2
kind: Config
build:
  tagPolicy:
    sha256: {}
  artifacts:
    workspace: .
  local: {}
deploy:
  kubectl:
    manifests:
    - k8s-*
```

The old syntax still works:
```
build:
  sha256: {}
```

Looks like the fix is changing `,inline` to `tagPolicy` [here](https://github.com/GoogleCloudPlatform/skaffold/blob/44fcb3a04ee4d0994f0d8d85f14b51798aad205f/pkg/skaffold/config/config.go#L38)

geniass:
Closes #326

### DIFF
--- a/pkg/skaffold/config/config.go
+++ b/pkg/skaffold/config/config.go
@@ -35,7 +35,7 @@ type SkaffoldConfig struct {
 // BuildConfig contains all the configuration for the build steps
 type BuildConfig struct {
 	Artifacts []*Artifact `yaml:"artifacts"`
-	TagPolicy TagPolicy   `yaml:",inline"`
+	TagPolicy TagPolicy   `yaml:"tagPolicy"`
 	BuildType `yaml:",inline"`
 }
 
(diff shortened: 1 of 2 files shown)


r2d4:
Yup, we need a much better way of applying defaults. I would love to continue unmarshalling into the default struct, but i think we'll have to just manually apply the default logic after we parse the config. @dlorenc picks up the `inline` error in #316

dgageot:
@geniass I think this was fixed by https://github.com/GoogleCloudPlatform/skaffold/pull/332. Feel free to reopen if it's not the case!

geniass:
Looks like it fixes it. Thanks!

PR #340, opened 2018-04-05
Fix unreliable file change detection

### LINKED ISSUE #313: File change detection is unreliable when coding with IntelliJ/GoLand
When using IntelliJ/Goland, after a few changes, no file changes are detected anymore. Even outside of the IDE. Looks like it breaks silently.

dgageot:
When IntelliJ saves a file, it renames it and recreates it. That's called "Safe save mode" or something similar.
Because the file watcher loses track of a file as soon as it is renamed, we need to add watches on directories too. This way, if a file reappears, we get the notification.

If a file appears, that wasn't tracked in the first place, we just ignore it. Sometimes that's ok, sometimes it's not. In an ideal world, we should match this new file with the dependencies of the Dockerfile.

Fixes #313

### DIFF
--- a/pkg/skaffold/watch/watch.go
+++ b/pkg/skaffold/watch/watch.go
@@ -18,6 +18,7 @@ package watch
 
 import (
 	"context"
+	"path/filepath"
 	"sort"
 	"time"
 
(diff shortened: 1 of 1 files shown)


dgageot:
ping @r2d4 @dlorenc

PR #342, opened 2018-04-05
Don’t hardcode namespace

dgageot:


dgageot:
Sure! I was just making this part of the code look like the rest

PR #394, opened 2018-04-14
add support for multi-stage build in getting-started example

### LINKED ISSUE #393: Multi-stage builds support in getting-started Dockerfile
### Expected behavior
Use multi-stage builds for getting-started example. 

### Information
Currently the Dockerfile used in the getting started examples is in the "single-stage" format. As an optimisation, multi-stage format could be used

zanetworker:
Add multi-stage build format for getting-started Dockerfile, which resolves #393. 

Signed-off-by: zanetworker <adel.zalok.89@gmail.com>

### DIFF
--- a/examples/getting-started/Dockerfile
+++ b/examples/getting-started/Dockerfile
@@ -1,6 +1,8 @@
-FROM golang:1.9.4-alpine3.7
-
+FROM golang:1.9.4-alpine3.7 as builder
 WORKDIR /go/src/github.com/GoogleCloudPlatform/skaffold/examples/getting-started
-CMD ["./app"]
 COPY main.go .
 RUN go build -o app main.go
+
+FROM alpine:latest  
+CMD ["./app"]
+COPY --from=builder /go/src/github.com/GoogleCloudPlatform/skaffold/examples/getting-started/app .


examples/getting-started/Dockerfile
```diff
@@ -1,6 +1,9 @@
-FROM golang:1.9.4-alpine3.7
-
+FROM golang:1.9.4-alpine3.7 as builder
 WORKDIR /go/src/github.com/GoogleCloudPlatform/skaffold/examples/getting-started
-CMD ["./app"]
 COPY main.go .
-RUN go build -o app main.go
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .
```
dgageot: Do we really need `CGO_ENABLED=0 GOOS=linux`?

examples/getting-started/Dockerfile
```diff
@@ -1,6 +1,9 @@
...
+
+FROM alpine:latest  
+WORKDIR /root/
+COPY --from=builder /go/src/github.com/GoogleCloudPlatform/skaffold/examples/getting-started/app .
+CMD ["./app"]
```
dgageot: Can you move `COPY` to the end to maximize caching?

examples/getting-started/Dockerfile
```diff
@@ -1,6 +1,9 @@
...
-
+FROM golang:1.9.4-alpine3.7 as builder
 WORKDIR /go/src/github.com/GoogleCloudPlatform/skaffold/examples/getting-started
-CMD ["./app"]
 COPY main.go .
-RUN go build -o app main.go
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .
+
+FROM alpine:latest  
+WORKDIR /root/
```
dgageot: Is `WORKDIR` needed?

examples/getting-started/Dockerfile
zanetworker: > Preview

Yes that makes sense, will do that.

examples/getting-started/Dockerfile
zanetworker: Not really in this case, just for convenience when debugging. Removed it.

examples/getting-started/Dockerfile
zanetworker: No, removed it.

zanetworker:
Thanks for your review, applied the changes.

PR #402, opened 2018-04-18
Add support remote manifest from kubernetes(replace mode)

sergeylanzman:
This PR adds support replace exist deployment with new image.
You can set up your deployment in cluster(namespace/deployment/deployment_name) in skaffold.yaml
After this skaffold on the build get deployment yaml from a cluster and replace only image URL with a new tag and after this deploy again to cluster.
It's a good case if you have many namespaces and very complicate deployment with a lot of env params generate by namespace. Now you can replace the only code in an existing deployment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 out/
 examples/bazel/bazel-*
 *.new
+.idea/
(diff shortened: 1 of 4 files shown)


sergeylanzman:
@r2d4 thanks for your feedback. I fixed 2 issues

PR #409, opened 2018-04-19
don't watch to vendor folder

sergeylanzman:
Watcher crash with error `To many files` if you have a medium size of vendor folder.
In common case we user don't change code in vendor folder and not needed watch for each file.

P.S. Maybe need add watch_exclude_folders settings, because if you have big folders like node_modules it's huge or watcher crash

### DIFF
--- a/pkg/skaffold/watch/watch.go
+++ b/pkg/skaffold/watch/watch.go
@@ -22,6 +22,8 @@ import (
 	"sort"
 	"time"
 
+	"strings"
+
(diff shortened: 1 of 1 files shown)


dgageot:
@sergeylanzman Thank you for submitting this patch. Some code is supposed to do that already but it turns out it's broken. Here's a PR that should fix it. Can you try it? https://github.com/GoogleContainerTools/skaffold/pull/420

PR #410, opened 2018-04-19
delete hardcoded OSChoice linux

sergeylanzman:
delete hardcoded OSChoice linux
Not need to provide OSChoice,
```
        // If not "", overrides the use of platform.GOARCH when choosing an image or verifying architecture match.
	ArchitectureChoice string
	// If not "", overrides the use of platform.GOOS when choosing an image or verifying OS match.
	OSChoice string
```


r2d4:
@sergeylanzman  we just removed containers/image in favor of https://github.com/google/go-containerregistry which we're also developing

PR #417, opened 2018-04-19
Add kaniko builder

r2d4:
This branched off of #411 which needs to be merged first

PR #434, opened 2018-04-22
Fix kaniko example

carlossg:
And add some extra debug info

It is picking the default local build instead of Kaniko

Follow up after #417


r2d4:
opened https://github.com/GoogleContainerTools/skaffold/issues/435

PR #444, opened 2018-04-23
Revert commits

r2d4:
"Better streaming of docker messages" causes skaffold to panic on gotty.go. I think we can just update our docker/cli dependency here, but it might be better to just revert this.

I believe its this issue
https://github.com/moby/moby/pull/32401

Capture digest at build fails for me on all platforms. I'm not sure why it was passing CI. I tried this on docker 17.09-ce.

### DIFF
--- a/pkg/skaffold/build/bazel.go
+++ b/pkg/skaffold/build/bazel.go
@@ -20,12 +20,12 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


r2d4:
The digest PR works with docker for mac (18.03-ce), but not minikube (17.09-ce)

r2d4:
Digest PR works on 17.09-ce

r2d4:
We're going to revert this until we upgrade the version of docker in minikube. 

ref https://github.com/kubernetes/minikube/issues/2764

Issue #447, opened 2018-04-23
Helptext is shown after all errors


r2d4:
```
$ skaffold run
Starting build...
Error: running skaffold steps: build: build step: running build for artifact: Could not find dockerfile: Dockerfile
Usage:
  skaffold run [flags]

Flags:
  -f, --filename string       Filename or URL to the pipeline file (default "skaffold.yaml")
  -h, --help                  help for run
  -p, --profile stringArray   Activate profiles by name
  -t, --tag string            The optional custom tag to use for images which overrides the current Tagger configuration
      --toot                  Emit a terminal beep after the deploy is complete

Global Flags:
  -v, --verbosity string   Log level (debug, info, warn, error, fatal, panic (default "warning")
```

dgageot:
Fixed with #449

PR #449, opened 2018-04-23
cmd: silence usage text

r2d4:
This supresses usage test from the output when we have an application error OR usage error. I can't find out a clear way to use cobra's arg validation and toggle this.
https://github.com/spf13/cobra/issues/340

I'm a bit undecided about this. On one hand, we shouldn't output message error text on our application error, but on the other hand, for legitimate usage errors, the user needs to now run an additional `skaffold --help`

### DIFF
--- a/cmd/skaffold/app/cmd/cmd.go
+++ b/cmd/skaffold/app/cmd/cmd.go
@@ -48,6 +48,7 @@ func NewSkaffoldCommand(out, err io.Writer) *cobra.Command {
 		if err := SetUpLogs(err, v); err != nil {
 			return err
 		}
+		rootCmd.SilenceUsage = true
 		logrus.Infof("Skaffold %+v", version.Get())
 		return nil
 	}


r2d4:
fixes #447

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -40,12 +40,12 @@ var (
 )
 
 var rootCmd = &cobra.Command{
-	Use:   "skaffold",
-	Short: "A tool that facilitates continuous development for Kubernetes applications.",
+	Use:          "skaffold",
+	Short:        "A tool that facilitates continuous development for Kubernetes applications.",
+	SilenceUsage: true,
```
dgageot: If you move this line to `PersistentPreRunE` function, it will silent usage only after args are successfully checked.

dgageot:
@r2d4 Do you think you can update your PR?

cmd/skaffold/app/cmd/cmd.go
tejal29: or use https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/ctc_lib
Where we silenceUsage Errors in initializers.

Issue #322, opened 2018-04-03
skaffold dev doesnt work with v0.3.0


sirajg:
`skaffold dev` does not work with built in examples 

### Expected behavior
`skaffold dev` should build, deploy and watch for changes.

### Actual behavior
```.../getting-started$ skaffold dev
Starting build...
Found [docker-for-desktop] context, using local docker daemon.
WARN[0000] run: build: build step: running build: read auth configs: error getting credentials - err: exit status 1, out: `no hostname in URL` 
Watching for changes...
```
### Information
This error is seen only in v0.3.0. Same setup with v0.2.0 works.

```
Skaffold version: v.0.3.0
Operating system: mac os
Content of skaffold.yaml: ...

apiVersion: skaffold/v1alpha2
kind: Config
build:
  artifacts:
  - imageName: gcr.io/k8s-skaffold/skaffold-example
    workspace: .
  local: {}
deploy:
  kubectl:
    manifests:
      - k8s-*

```

### Steps to reproduce the behavior

  1. ...
  2. ...

dgageot:
@sirajg Do you have docker installed? If it is, can you show the output of `docker info`? What's your kubernetes context?

sirajg:
```kubectl config get-contexts
CURRENT   NAME                 CLUSTER                      AUTHINFO             NAMESPACE
*         docker-for-desktop   docker-for-desktop-cluster   docker-for-desktop   
          minikube             minikube                     minikube       
```
```docker info
docker info
Containers: 220
 Running: 118
 Paused: 0
 Stopped: 102
Images: 388
Server Version: 18.03.0-ce-rc3
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host ipvlan macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: cfd04396dc68220d1cecbe686a6cc3aa5ce3667c
runc version: 4fc53a81fb7c994640722ac585fa9ca548971871
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.9.87-linuxkit-aufs
Operating System: Docker for Mac
OSType: linux
Architecture: x86_64
[...]

PR #454, opened 2018-04-24
Use docker CLI code to read the config

dgageot:
+ Support non-existing config file
 + Support config in old format


dgageot:
Fixes #322

PR #455, opened 2018-04-24
Support ~ in secret path for Kaniko

dgageot:


dgageot:
@r2d4 I fixed the code based on your feedback!

PR #468, opened 2018-04-25
updating instructions to update skaffold.yaml build artifact image location and name

tariq-islam:
Without instructing users on this step, at least in the intro getting-started example, they'll be pushing to a registry that they do not have access to. This field needs to be updated in order for the push to work.


dgageot:
@tariq-islam That was quick ;-)

PR #469, opened 2018-04-25
Instruction on updating skaffold.yaml with registry location

tariq-islam:
User needs to update the skaffold.yaml file in order to push resulting image to the appropriate registry, vs the default value that's already in there.


tariq-islam:
I've signed it!

PR #470, opened 2018-04-25
updating instructions to update skaffold.yaml registry loc

tariq-islam:
Currently the user will receive an error when the built image is pushed, as the default value is not a registry that users have access to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,8 @@ You will need the following components to get started with Skaffold:
     cd examples/getting-started
     ```
 
+1. In the skaffold.yaml file, update the `build.artifacts.imageName` stanza to the registry where your image will be pushed (`<registry/image:tag>`). For example `gcr.io/<your-project-ID>/your-image`, where `gcr.io/<your-project-ID>` is your GCP container registry location. As mentioned earlier, your docker client should be configured to push to this location.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,6 +97,8 @@ You will need the following components to get started with Skaffold:
     cd examples/getting-started
     ```
 
+1. In the skaffold.yaml file, update the `build.artifacts.imageName` stanza to the registry where your image will be pushed (`<registry/image:tag>`). For example `gcr.io/<your-project-ID>/your-image`, where `gcr.io/<your-project-ID>` is your GCP container registry location. As mentioned earlier, your docker client should be configured to push to this location.
```
r2d4: These instructions are actually for "local clusters" (i.e. minikube or docker for desktop) where you don't actually have to do this step. Maybe we should make the heading or tutorial clearer, since a few users have run into this while trying it out on GKE.

tariq-islam: Given that clarification I'd close this issue. It's more of a UX thing, it's not immediately intuitive to look in skaffold.yaml for the resultant image location/name.

dlorenc:
Going to close this as per the comments in the PR review. This UX issue is definitely something we'd like to tackle. #480 is tracking the overall issue.

PR #476, opened 2018-04-26
Add a Build command

### LINKED ISSUE #425: Make git tagger the default for both dev and run
Right now the default tagger is different for skaffold dev and skaffold run. I think it makes more sense to always tag with git commit, since its more digestible (no pun intended) than a seemingly random checksum.

I've been convinced @dlorenc

dgageot:
Also fixes #425, which means there are no more different defaults in dev and run modes.

### DIFF
--- a/cmd/skaffold/app/cmd/build.go
+++ b/cmd/skaffold/app/cmd/build.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"io"
 
-	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner"
 	"github.com/spf13/cobra"
 )
 
@@ -31,11 +30,20 @@ func NewCmdBuild(out io.Writer) *cobra.Command {
 		Short: "Builds the artifacts",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runSkaffold(out, filename, func(ctx context.Context, r *runner.SkaffoldRunner) error {
-				return r.Build(ctx)
-			})
+			return build(out, filename)
 		},
 	}
 	AddRunDevFlags(cmd)
 	return cmd
+}
+
+func build(out io.Writer, filename string) error {
+	ctx := context.Background()
+
+	runner, err := NewRunner(out, filename)
+	if err != nil {
+		return err
+	}
+
+	return runner.Build(ctx)
 }
--- a/cmd/skaffold/app/cmd/build.go
+++ b/cmd/skaffold/app/cmd/build.go
@@ -1,0 +1,41 @@
+/*
+Copyright 2018 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"context"
+	"io"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner"
+	"github.com/spf13/cobra"
+)
+
+// NewCmdBuild describes the CLI command to build artifacts.
+func NewCmdBuild(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "build",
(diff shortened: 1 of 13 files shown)


pkg/skaffold/runner/runner.go
```diff
@@ -139,15 +139,28 @@ func newTaggerForConfig(t v1alpha2.TagPolicy) (tag.Tagger, error) {
...
 
-// Run runs the skaffold build and deploy pipeline.
-func (r *SkaffoldRunner) Run(ctx context.Context) error {
-	if r.opts.DevMode {
-		if r.opts.Cleanup {
-			return cleanUpOnCtrlC(ctx, r.dev, r.cleanup)
-		}
-		return r.dev(ctx)
+// Build builds the artifacts.
+func (r *SkaffoldRunner) Build(ctx context.Context) error {
```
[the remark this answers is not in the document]
r2d4: Since the builder is embedded in the runner, you can just call build straight from that. Then, our cmd/build would just look like

```
runner := NewForConfig()
runner.Build()
```

Or most likely you can reuse the private r.build here so that the UX looks similar.

pkg/skaffold/runner/runner.go
```diff
@@ -139,15 +139,28 @@ func newTaggerForConfig(t v1alpha2.TagPolicy) (tag.Tagger, error) {
...
+	bRes, err := r.build(ctx, r.config.Build.Artifacts)
+
+	for _, res := range bRes.Builds {
```
r2d4: This could live in r.build and be exposed there. Or if you dont think its appropriate for dev/run mode, we can add it in cmd/

pkg/skaffold/config/util.go
```diff
@@ -27,22 +27,22 @@ import (
...
+	v1alpha2.Version: func(contents []byte, useDefault bool) (util.VersionedConfig, error) {
 		config := new(v1alpha2.SkaffoldConfig)
-		err := config.Parse(contents, useDefault, dev)
+		err := config.Parse(contents, useDefault)
 		return config, err
 	},
 }
 
-func GetConfig(contents []byte, useDefault bool, dev bool) (util.VersionedConfig, error) {
+func GetConfig(contents []byte, useDefault bool) (util.VersionedConfig, error) {
```
r2d4: Does this also mean that we no longer have default differences between dev/run? I'm ok with that if thats the case, but it should be noted.

pkg/skaffold/schema/v1alpha2/config.go
```diff
@@ -178,16 +178,12 @@ func defaultToDockerArtifacts(cfg *SkaffoldConfig) {
...
 	if cfg.Build.TagPolicy != (TagPolicy{}) {
 		return
 	}
 
-	if dev {
-		cfg.Build.TagPolicy = TagPolicy{ShaTagger: &ShaTagger{}}
-	} else {
-		cfg.Build.TagPolicy = TagPolicy{GitTagger: &GitTagger{}}
-	}
+	cfg.Build.TagPolicy = TagPolicy{GitTagger: &GitTagger{}}
```
r2d4: I'm guessing addresses #425 ?

pkg/skaffold/runner/runner.go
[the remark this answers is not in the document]
dgageot: I'd rather try that in another PR. I've tried something similar and failed.

cmd/skaffold/app/cmd/build.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// NewCmdBuild describes the CLI command to build artifacts.
+func NewCmdBuild(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "build",
+		Short: "Builds the artifacts",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runSkaffold(out, filename, func(ctx context.Context, r *runner.SkaffoldRunner) error {
```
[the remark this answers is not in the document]
dgageot: Good point! I'll just return the runner

pkg/skaffold/runner/runner.go
dgageot: `Build` and `build` don't take the same parameters. I'd rather try to refactor that in a separate PR

pkg/skaffold/schema/v1alpha2/config.go
dgageot: Yes! That's in the commit message. I've added that to the PR description. It greatly simplify the code and makes more sense for the users.

dgageot:
@r2d4 I've applied your feedback and would love to refactor the Dev/Build/Run functions on a separate PR. It requires refactoring a lot of the tests too.

PR #487, opened 2018-04-27
fix: allow environment variables to be used in templated helm values

### LINKED ISSUE #486: cannot use environment variables in helm templates
this came up here: 
https://github.com/GoogleContainerTools/skaffold/issues/361#issuecomment-380723998

I think we need a way to pass in template expressions to helm chart values, like we do for `envTemplates` so that we can pass in, say, the local docker registry ip/port or private registry via environment variables like we can do for building the docker image name.

jstrachan:
fixes #486

### DIFF
--- a/pkg/skaffold/deploy/helm.go
+++ b/pkg/skaffold/deploy/helm.go
@@ -17,10 +17,15 @@ limitations under the License.
 package deploy
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
+	"strconv"
+	"strings"
+	"text/template"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/v1alpha2"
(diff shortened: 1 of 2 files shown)


pkg/skaffold/deploy/helm.go
```diff
@@ -144,3 +163,39 @@ func (h *HelmDeployer) deleteRelease(out io.Writer, r v1alpha2.HelmRelease) erro
 	out.Write(stdout)
 	return nil
 }
+
+func (h *HelmDeployer) evaluateTemplates(setValueTemplates map[string]string, r *build.BuildResult) (map[string]string, error) {
+	results := map[string]string{}
+	envMap := map[string]string{}
+	for _, env := range environ() {
```
tejal29: Can you re-factor and re-use some of the code here https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/tag/env_template.go#L53

dgageot:
@jstrachan @afrittoli Could you rebase and re-use the code that was extracted since?

jstrachan:
@dgageot @afrittoli I've recreated this PR reusing the util code for parsing templates and replacing environment variables:
https://github.com/GoogleContainerTools/skaffold/pull/707

so am gonna close this PR now

PR #494, opened 2018-04-28
fix defaults on builds defined in profiles

### LINKED ISSUE #493: default Dockerfile not applied for profiles
### Expected behavior

Defining a `build` in a profile should inherit the default `Dockerfile` and workspace of `.` so that it behaves the same as a regular `build` in the root directory.

e.g. moving

```yaml
apiVersion: skaffold/v1alpha2
kind: Config
build:
  artifacts:
  - workspace: .
    docker: {}
```

to be

```yaml
apiVersion: skaffold/v1alpha2
kind: Config
profiles:
- name: run
  build:
    artifacts:
    - workspace: .
      docker: {}
```

should not change the behaviour. Whereas the latter barfs as it can't find the `Dockerfile`

### Actual behavior

Currently defaults of `.` on workspace and `Dockerfile` for the `dockerfilePath`  are only applied on the root `build` not when applying profiles.

Test case and fix on the way

jstrachan:
fixes #493

### DIFF
--- a/pkg/skaffold/schema/v1alpha2/config.go
+++ b/pkg/skaffold/schema/v1alpha2/config.go
@@ -226,6 +226,10 @@ func (c *SkaffoldConfig) ApplyProfiles(profiles []string) error {
 
 	c.Profiles = nil
 
+	// lets populate any missing default values
+	setDefaultDockerfiles(c)
+	setDefaultWorkspaces(c)
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/skaffold/schema/v1alpha2/config.go
```diff
@@ -226,6 +226,10 @@ func (c *SkaffoldConfig) ApplyProfiles(profiles []string) error {
 
 	c.Profiles = nil
 
+	// lets populate any missing default values
+	setDefaultDockerfiles(c)
```
dgageot: There's more default values set when we read the configuration. Shouldn't we try to reuse the same code to avoid duplication and missing things?

jstrachan: sounds good to me. I was just trying to un-break profiles

dgageot:
@jstrachan https://github.com/GoogleContainerTools/skaffold/pull/533 applies all the defaults to profiles

dgageot:
@jstrachan, #533 was merged. It should fix your issue.

PR #497, opened 2018-04-30
fix: allow an environment variable to default the deploy namespace

### LINKED ISSUE #496: allow the helm deploy to default the namespace value from an environment variable if there is no value in the YAML
in `dev` mode its handy to let developers spin up separate namespaces and work on the code and redeploy them without clashing with other developers. i.e. each developer should probably have a separate deploy namespace.

Right now to do this you need each developer to edit the `skaffold.yaml` file and update the `deploy.helm.namespace` value to something unique to avoid developers overwriting each other.

It would be nice to be able to define an environment variable, say, `SKAFFOLD_DEPLOY_HELM_NAMESPACE` which if there is no explicit namespace in `deploy.helm.namespace` then the `$SKAFFOLD_DEPLOY_HELM_NAMESPACE` is used and if thats not available then use the current namespace.

We may want to use separate namespaces for building as deploying. For example; there may be a shared namespace with, say, secrets for pushing to docker registries, dealing with helm repos or downloading artifacts (say, maven) from a maven proxy server etc. So multiple developers can be building in the same namespace; but its common to want to deploy in separate namespaces.

Thats certainly the way its looking for Jenkins X; for building stuff; use a shared team namespace but use separate Preview Environments to deploy code. 

So it'd make things easier for developers using Jenkins X if we can point the, say, `SKAFFOLD_DEPLOY_HELM_NAMESPACE` environment variable at each individual developers `Development Environment` namespace; so that developers don't have to edit any code to avoid clashing

jstrachan:
lets allow an environment variable to be used to define the helm deploy namespace
fixes #496

### DIFF
--- a/pkg/skaffold/deploy/helm.go
+++ b/pkg/skaffold/deploy/helm.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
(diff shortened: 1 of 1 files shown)


pkg/skaffold/deploy/helm.go
```diff
@@ -106,8 +107,12 @@ func (h *HelmDeployer) deployRelease(out io.Writer, r v1alpha2.HelmRelease, b *b
 		args = append(args, "upgrade", r.Name, r.ChartPath)
 	}
 
-	if r.Namespace != "" {
-		args = append(args, "--namespace", r.Namespace)
+	ns := r.Namespace
```
dgageot: Is there a way to map the cli arguments to env variables? I think there is. It would be much better to handle that once for all when CLI arguments are parsed.

jstrachan: good point yeah; in this case we're merging default values from the CLI / env var overriding any missing values in the `skaffold.yaml`

PR #529, opened 2018-05-08
Add --tiller-namespace support to config

dannydavidson:
I have a cluster I need to deploy to that is using namespacing heavily for isolation.  While I can use an environment var before running skaffold to set the tiller namespace, it'd be nice to make it fully declarative in the config.  

If this has merit I can add some tests, but wanted to get thoughts first.

### DIFF
--- a/pkg/skaffold/deploy/helm.go
+++ b/pkg/skaffold/deploy/helm.go
@@ -68,7 +68,12 @@ func (h *HelmDeployer) Cleanup(ctx context.Context, out io.Writer) error {
 }
 
 func (h *HelmDeployer) args(moreArgs ...string) []string {
-	return append([]string{"--kube-context", h.kubeContext}, moreArgs...)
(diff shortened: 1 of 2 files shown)


dannydavidson:
I signed it!

dlorenc:
I think this definitely makes sense to support, but I'm not sure it makes sense in the skaffold config file like this. Ideally this file would be portable between clusters, and cluster/helm config would live somewhere else.

Is there a helm configuration file we could parse anywhere on the system?

dlorenc:
Maybe we could follow the environment variables here?

https://github.com/kubernetes/helm/issues/2519

dannydavidson:
Yeah the environment variable approach is what I've been using.  It's likely sufficient.  Will close.

michas2:
If you have multiple kube-contexts configured and need a different tiller-namespace for each of them, wouldn't it make sense to use the namespace according to the config?

PR #557, opened 2018-05-16
Simplify configuration versioning

dgageot:
Make the code a bit simpler

### DIFF
--- a/pkg/skaffold/config/util.go
+++ b/pkg/skaffold/config/util.go
@@ -24,30 +24,31 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/v1alpha2"
 )
 
-// Ordered list of all schema versions
-var Versions = []string{v1alpha1.Version, v1alpha2.Version}
+// Versions is an ordered list of all schema versions.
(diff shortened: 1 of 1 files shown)


tejal29:
Please rebase from latest master.

pkg/skaffold/config/util.go
```diff
@@ -24,31 +24,33 @@ import (
...
+		return new(v1alpha2.SkaffoldConfig)
 	},
 }
 
 func GetConfig(contents []byte, useDefault bool) (util.VersionedConfig, error) {
 	for _, version := range Versions {
-		if cfg, err := schemaVersions[version](contents, useDefault); err == nil {
+		cfg := schemaVersions[version]()
+
+		if err := cfg.Parse(contents, useDefault); err == nil {
```
tejal29: this will hide the actual error when there is a Parse error.
Please get rid of err == nil line or
```
if err := cfg.Parse(contents, useDefault); cfg.GetVersion() == version {
   return cfg, err
}

dgageot:
@tejal29 Thanks for the review. This should be good to go

Issue #550, opened 2018-05-15
Automatic docker login for CI


errordeveloper:
I think it'd be very nice to avoid having to run `docker login` when you want to let `skaffold run` build and push your images. Perhaps some kind of convention would be reasonable, e.g. detect commonly used environment variables.

Most CI user would end-up following the instructions vendors provide.

For example, [Travis CI](https://docs.travis-ci.com/user/build-stages/share-docker-image/) suggests to use this:
```
echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
```
So detecting `DOCKER_PASSWORD` and `DOCKER_USERNAME` would work.

[CircleCI 1.0](https://circleci.com/docs/1.0/docker/) suggests this:
```
docker login -u $DOCKER_USER -p $DOCKER_PASS
```
While [CircleCI 2.0](https://circleci.com/docs/2.0/building-docker-images/) is much more Docker-native, and they have per-image auth, they still offer same `docker login` example as an option.

It'd be very convenient to the user if they could keep same secrets in CI config and drop `docker login` + `docker push`, and start using `skaffold run` instead.

dlorenc:
I like this idea. We could implement an "environment variable" keychain in go-containerregistry.

derekperkins:
I would want to use this with other container registries as well, specifically Azure.

`az acr login --name myregistry`

jonjohnsonjr:
@derekperkins for normal authentication (where you control the environment), I'd recommend using a docker credential helper:
https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol
https://github.com/Azure/acr-docker-credential-helper

The go-containerregistry keychain will try to authenticate via whatever cred helpers you have configured in your docker config file.

derekperkins:
Thanks for the suggestion @jonjohnsonjr! It caused a panic first run, but that may have been user error. I'll wait until my login expires and try again to see if it works.

balopat:
Hi All, do we still think this is useful? I would like to collect more data / community support before we raise the priority on this feature. Thanks!

errordeveloper:
I think this would still be very useful indeed.

dgageot:
Hi everyone, I'm going to close this one since we didn't see a lot of demand and we do not plan on working on that subject anytime soon. Feel free to reopen if you really want that feature to be implemented.

CarlQLange:
[...]

PR #558, opened 2018-05-16
[WIP] Run in a container

errordeveloper:
I've had a look at [`deploy/skaffold/Dockerfile`](https://github.com/GoogleContainerTools/skaffold/blob/3246ef8e2028611460aaf911bd38781d9f19764b/deploy/skaffold/Dockerfile), and it uses Ubuntu base image and I couldn't actually build it locally. Also, it seems like that container image is meant for integration tests.

I'd like to see an official slim container that I can use in CI. I'm not sure what's the preferred way to go about it, so I did the simplest thing first – multi-stage build based on Alpine. Please take a look and let me know.

I suppose we will need to consider:
- location on Dockerfile (it's in root at the moment)
- distroless instead of Alpine
- image location in GCR and tagging
- use of GCB (at the moment I'm inclined to Dockerfile, so to keep it simple for contributors to build locally)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,87 @@
+# Copyright 2018 The Skaffold Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM golang:1.10-alpine AS build
+
+RUN apk add --update \
+      curl \
+      git \
+      make \
+      py-pip \
+      python \
+      python-dev \
+      && true
+
+RUN mkdir /out
+RUN mkdir -p /out/etc/apk && cp -r /etc/apk/* /out/etc/apk/
+RUN apk add --no-cache --initdb --root /out \
+    alpine-baselayout \
+    busybox \
+    ca-certificates \
+    coreutils \
+    git \
+    libc6-compat \
+    libgcc \
+    libstdc++ \
+    python \
+    && true
+
+ENV SKAFFOLD $GOPATH/src/github.com/GoogleContainerTools/skaffold
+RUN mkdir -p "$(dirname ${SKAFFOLD})"
+COPY . $SKAFFOLD
+
+WORKDIR $SKAFFOLD
+RUN make \
+    && cp out/skaffold /out/usr/local/bin/skaffold
+
+WORKDIR /out
+
+RUN ln -s /lib /lib64
+
+ENV KUBECTL_VERSION v1.10.6
+RUN curl --silent --location "https://dl.k8s.io/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" --output usr/local/bin/kubectl \
+    && chmod +x usr/local/bin/kubectl
+
+ENV DOCKER_VERSION 18.03.0
+RUN curl --silent --location "https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}-ce.tgz" \
+    | tar xz docker/docker \
+    && mv docker/docker usr/local/bin/docker && rm -rf docker
+
+ENV CLOUD_SDK_VERSION 206.0.0
+RUN curl --silent --location "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${CLOUD_SDK_VERSION}-linux-x86_64.tar.gz" \
+    | tar xz \
+    && pip install --root=/out crcmod==1.7
+
+ENV HELM_VERSION v2.9.1
+RUN curl --silent --location "https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}-linux-amd64.tar.gz" \
+    | tar xz linux-amd64/helm \
+    && mv linux-amd64/helm usr/local/bin/helm
+
+ENV DOCKER_CREDENTIAL_GCR_VERSION 1.5.0
+RUN curl --silent --location "https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v${DOCKER_CREDENTIAL_GCR_VERSION}/docker-credential-gcr_linux_amd64-${DOCKER_CREDENTIAL_GCR_VERSION}.tar.gz" \
+    | tar xz ./docker-credential-gcr \
+    && mv docker-credential-gcr usr/local/bin/docker-credential-gcr
+# TODO: docker-credential-gcr configure-docker
+
+ENV KUSTOMIZE_VERSION 1.0.6
+RUN curl --silent --location "https://github.com/kubernetes-sigs/kustomize/releases/download/v${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64" --output usr/local/bin/kustomize \
+    && chmod +x usr/local/bin/kustomize
+
+ENV BAZEL_VERSION 0.16.1
+RUN curl --silent --location "https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-x86_64" --output usr/local/bin/bazel \
+    && chmod +x usr/local/bin/bazel
+
+FROM scratch
+CMD skaffold
+COPY --from=build  /out /


errordeveloper:
I've tested this on Docker for Mac:
```
docker run -ti --rm \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -v $HOME/.kube:/root/.kube \
  -v $(pwd):$(pwd) -w $(pwd) \
  errordeveloper/skaffold:1c894265d81c618c782500cde2d788fb77f3b416 \
     skaffold dev
```
The build worked, but deploy hits
```
Error: starting logger: Get https://localhost:6443/api/v1/pods?includeUninitialized=true&watch=true: dial tcp 127.0.0.1:6443: connect: connection refused
```
I suppose I could fix it by running in a pod, but I'm not so sure if it makes much sense, as the user experience is kind of convoluted.

My primary use-case was to make it easy to run in Docker-enabled CI, without having to download the binary.

I was able to use `skaffold build` on Docker for Mac this way:
```
docker run -ti --rm \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -v $HOME/.kube:/root/.kube \
  -v $(pwd):$(pwd) -w $(pwd) \
  errordeveloper/skaffold:1c894265d81c618c782500cde2d788fb77f3b416 \
     skaffold build
```

This broadly satisfies my needs, although I've noticed that removing `-v $HOME/.kube:/root/.kube` results in an error, it'd be good to eliminate it.
```
Error: getting skaffold config: getting k8s client: Error creating kubeConfig: invalid configuration: no configuration has been provided
```

I'll try this image in CI now and report back.

errordeveloper:
I've tried using this in CircleCI with a following config:
```
# .circleci/config.yml 
version: 2
jobs:
  build:
    docker:
      - image: errordeveloper/skaffold:1c894265d81c618c782500cde2d788fb77f3b416
    steps:
      - checkout
      - run: skaffold build 
workflows:
  version: 2
  build_and_test:
    jobs:
      - build
```

I got the same error, I didn't expect anything different. The config is quite nice an simple.

I think I could actually download the binary, but that won't make any difference. I'll try to stab-out `kubeconfig` for now, let's see if it will work.

I've opened #559 to discuss the general issue of `kubeconfig` dependency.

errordeveloper:
With a dummy `kubeconfig` added to the image, I've been able to use this config:
```
version: 2
jobs:
  build:
    docker:
      - image: errordeveloper/skaffold:66cc263ef18f107adce245b8fc622a8ea46385f2
    steps:
      - checkout
      - setup_remote_docker: {docker_layer_caching: true}
      - run: skaffold build 
workflows:
  version: 2
  build_and_test:
    jobs:
      - build
```
I still need to add registry auth, but so far I think it's rather nice and simple config, it demonstrates how easy it is to use Skaffold in a Docker-native CI. In Travis CI I'd have to call `docker run`, but it's still better then having to download all the binaries.

Let's figure out how to fix #550 properly and then review how to build the container image. Also #550 would be handy at some point.

r2d4:
Sorry for the delayed response - I just got back from vacation after KubeCon :)

> location on Dockerfile (it's in root at the moment)
root SGTM

> distroless instead of Alpine
I'm inclined to use distroless, but no strong opinions here.

> image location in GCR and tagging
> use of GCB (at the moment I'm inclined to Dockerfile, so to keep it simple for contributors to build locally)
we can follow the example and set GCB as a profile if users want

I'm also inclined to either build the integration test image on top of this one or have them be the same. There's currently nothing extra in the integration test image and theoretically they should really be the same.

The real issue is shipping support for all the builders and deployers. The image is a bit large now because it needs to include the binaries for each builder and deployer (and its actually incorrect right now since it doesn't include bazel). I think anything we publish needs to have all of them, otherwise we'll have to maintain many.

errordeveloper:
> I'm also inclined to either build the integration test image on top of this one or have them be the same. There's currently nothing extra in the integration test image and theoretically they should really be the same.

Yes, that'd be a good idea. I just didn't know whether way the other Dockerfile is done is important in any way or not, so I've not decided to do any refactoring there yet.

> The real issue is shipping support for all the builders and deployers. The image is a bit large now because it needs to include the binaries for each builder and deployer (and its actually incorrect right now since it doesn't include bazel). I think anything we publish needs to have all of them, otherwise we'll have to maintain many.

I see, good point, I didn't realise this. I suppose it'd have to come before we refactor the integration image.

errordeveloper:
@r2d4 I've just rebased this and update it to include latest versions of all builder.

Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+ENV HELM_VERSION v2.9.1
+RUN curl --silent --location "https://storage.googleapis.com/kubernetes-helm/helm-${HELM_VERSION}-linux-amd64.tar.gz" \
+    | tar xz linux-amd64/helm \
+    && mv linux-amd64/helm usr/local/bin/helm
+
+ENV DOCKER_CREDENTIAL_GCR_VERSION 1.5.0
+RUN curl --silent --location "https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v${DOCKER_CREDENTIAL_GCR_VERSION}/docker-credential-gcr_linux_amd64-${DOCKER_CREDENTIAL_GCR_VERSION}.tar.gz" \
+    | tar xz ./docker-credential-gcr \
+    && mv docker-credential-gcr usr/local/bin/docker-credential-gcr
+# TODO: docker-credential-gcr configure-docker
```
errordeveloper: Is this setup step required? Do we know what does it actually do?

Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+
+WORKDIR $SKAFFOLD
+RUN make \
+    && cp out/skaffold /out/usr/local/bin/skaffold
+
+WORKDIR /out
+
+RUN ln -s /lib /lib64
+
+ENV KUBECTL_VERSION v1.10.6
```
errordeveloper: It seems plausible to support latest 1.10, do folks agree?

Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+RUN curl --silent --location "https://dl.k8s.io/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" --output usr/local/bin/kubectl \
+    && chmod +x usr/local/bin/kubectl
+
+ENV DOCKER_VERSION 18.03.0
```
errordeveloper: One constraint I have here, is that it has to support multi-stage build, and I'm happy to include the latest version available (perhaps LTS one), but I don't know if we have other constrains or not...

Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+    && mv docker-credential-gcr usr/local/bin/docker-credential-gcr
+# TODO: docker-credential-gcr configure-docker
+
+ENV KUSTOMIZE_VERSION 1.0.6
+RUN curl --silent --location "https://github.com/kubernetes-sigs/kustomize/releases/download/v${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64" --output usr/local/bin/kustomize \
+    && chmod +x usr/local/bin/kustomize
+
+ENV BAZEL_VERSION 0.16.1
+RUN curl --silent --location "https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-x86_64" --output usr/local/bin/bazel \
+    && chmod +x usr/local/bin/bazel
```
errordeveloper: Looks like it needs glibc... 😿

```
dfc0b8b9e9f8:/go# ldd /usr/local/bin/bazel
        /lib64/ld-linux-x86-64.so.2 (0x7f2384c48000)
        librt.so.1 => /lib64/ld-linux-x86-64.so.2 (0x7f2384c48000)
        libdl.so.2 => /lib64/ld-linux-x86-64.so.2 (0x7f2384c48000)
        libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7f2384c48000)
        libm.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f2384c48000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x7f23848f6000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x7f23846e4000)
        libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f2384c48000)
Error relocating /usr/local/bin/bazel: __realpath_chk: symbol not found
Error relocating /usr/local/bin/bazel: __memcpy_chk: symbol not found
Error relocating /usr/local/bin/bazel: __sprintf_chk: symbol not found
dfc0b8b9e9f8:/go# bazel
Error relocating /usr/local/bin/bazel: __realpath_chk: symbol not found
Error relocating /usr/local/bin/bazel: __memcpy_chk: symbol not found
Error relocating /usr/local/bin/bazel: __sprintf_chk: symbol not found
dfc0b8b9e9f8:/go#
```

We have `libc6-compat` already, and I tried using instructions from https://github.com/winfinit/mongodb-prebuilt/issues/35 without any luck. Will have another go later on.

errordeveloper: Also filed https://github.com/bazelbuild/bazel/issues/5891.

Dockerfile
balopat: `docker-credential-gcr configure-docker` sets up the "gcr.io" repository patterns to use the docker-credential-gcr credential helper for the docker deamon. 
Skaffold needs the gcr credential helper to be able to push to gcr.io repos (https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/docker/auth.go) but as far as I understand we don't depend on the docker daemon to push so this line can probably go away. @r2d4?

r2d4: no, we follow the same auth flow as the CLI to push - read the .docker/config to figure out what credential helpers we need to call

Dockerfile
errordeveloper: For now, I'm gonna compile bazel, as it's relatively easy to do. However, it'd be good to compare how much do we win (in terms of image size) from having alpine vs e.g. ubuntu.

errordeveloper: Also filed https://github.com/bazelbuild/bazel/issues/5909.

errordeveloper:
As discuss on thew call today, I'll look at the existing Dockerfile and see if that works.

balopat:
@errordeveloper so, I'd rename this PR to "make skaffold Docker image smaller", what do you think? I seems that the current gcr.io/k8s-skaffold/skaffold image should be good enough for CI, just a bit on the larger side due to the ubuntu baseimage?

errordeveloper:
Yes, perhaps! But I am tempted to close this for now really.

dgageot:
@errordeveloper Are you ok if I close it, then?

errordeveloper:
Yes, please.

PR #560, opened 2018-05-16
Show errors of kubectl and helm commands

dgageot:


jonjohnsonjr:
This commit message references `helm`, but the change only seems to fix `kubectl` -- am I missing something?

dgageot:
@jonjohnsonjr No, I missed something! I'll push another PR to align helm's behavior. Thanks for noticing!

PR #566, opened 2018-05-18
Skip fully qualified names when replacing images names

dgageot:


dgageot:
@jonjohnsonjr That's a good idea.

pkg/skaffold/deploy/kubectl_test.go
```diff
@@ -412,19 +338,49 @@ spec:
         - containerPort: 80`, manifests.String())
 }
 
-func TestRemoveTag(t *testing.T) {
-	removed := removeTag("host:1234/user/container:tag")
-	if removed != "host:1234/user/container" {
-		t.Errorf("Tag vas not removed from an image name with port: %s ", removed)
+func TestSplitTag(t *testing.T) {
+	var tests = []struct {
```
dlorenc: Can we add a test case for digest?

pkg/skaffold/deploy/kubectl.go
```diff
@@ -339,7 +344,12 @@ func recursiveReplaceImage(i interface{}, replacements map[string]*replacement)
 	}
 }
 
-func removeTag(image string) string {
-	re := regexp.MustCompile(":[^/]+$")
-	return re.ReplaceAllString(image, "")
+func splitTag(image string) (string, string) {
+	re := regexp.MustCompile(`(.*):([^/]+)$`)
+	matches := re.FindStringSubmatch(image)
```
dlorenc: It might be a little more robust to reuse the name parsing utilities here: https://github.com/google/go-containerregistry/blob/master/name/ref.go

PR #540, opened 2018-05-11
Simpler runner code

dgageot:
This is a first step in my master plan to simplify runner's code

Also fixes #559 and #516

PR #581, opened 2018-05-23
Make sure deployed images are built while parsing skaffold config

### LINKED ISSUE #507: Unhelpful error: WARN[0001] run: build: build step: running push: denied: requested access to the resource is denied
- Skaffold v0.5.0.
- Docker for Mac (w/ Kubernetes) is currently running
- kubectl current-context pointing to a GKE cluster.
- docker config.json has `gcloud auth configure-docker` stuff

skaffold.yaml:

```yaml
apiVersion: skaffold/v1alpha2
kind: Config
build:
  artifacts:
  - imageName: vote
    workspace: voting-app/vote/
  - imageName: worker
    workspace: voting-app/worker/
  - imageName: result
    workspace: voting-app/result/
deploy:
  kubectl:
    manifests:
    - ./voting-app/k8s-specifications/*
```

My `image:` fields in the k8s Deployment.yamls are not maching this `imageName:` currently but I still 
 expect the image to be built but with a warning saying "image unused", example:

```yaml
    ...
    spec:
      containers:
      - image: gcr.io/YOUR_PROJECT_ID/vote
        name: vote
```

I see:

- This log, after the first image is built:

```
f846841ed47f: Waiting
0198944a9875: Waiting
9dfa40a0da3b: Waiting
WARN[0001] run: build: build step: running push: denied: requested access to the resource is denied
```

- Full logs (-v=debug):  https://gist.github.com/ahmetb/f41f2868d928834535e377888cd89a07

- 2nd and 3rd images are not being built.

- Shows a misleading "Watching for changes..." while something is broken. Skaffold should exit in this case IMO, with a more helpful error message.

priyawadhwa:
To fix #507 , during the parsing of the skaffold config I added a function to make sure that all deployed images are also built. If not, then skaffold exits with an error.


priyawadhwa:
@dgageot To use the example in the issue, the problem was that skaffold was building an image `vote` but deploying `gcr.io/some-registry/vote` which didn't exist. The error this gave is pretty unclear, and parsing the config was the only way I could think of to return a clearer error when this happens. 

But with #540 skaffold should exit when this happens, we could instead just change the error message if push fails to something like `push failed, perhaps this image wasn't built`?

dgageot:
@priyawadhwa yes, we should be able to just tune the error message

priyawadhwa:
Sounds good, I'll go ahead and close this PR then.

Issue #573, opened 2018-05-21
Set a user agent on GCP API calls


dlorenc:

dlorenc:
Include OS, skaffold version

PR #582, opened 2018-05-23
Add a user agent to GCB calls.

dlorenc:


dlorenc:
Fixes #573

dlorenc:
> I think we should add platform to the useragent too. Feel free to merge after that

Good call.

PR #583, opened 2018-05-23
Add namespace CLI flag to run

### LINKED ISSUE #496: allow the helm deploy to default the namespace value from an environment variable if there is no value in the YAML

afrittoli:
Add a CLI flag "namespace" to the run command to override the
namespace for all helm deployments defined in the pipeline.

Fixes #496

### DIFF
--- a/cmd/skaffold/app/cmd/cmd.go
+++ b/cmd/skaffold/app/cmd/cmd.go
@@ -75,6 +75,7 @@ func AddRunDevFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&filename, "filename", "f", "skaffold.yaml", "Filename or URL to the pipeline file")
 	cmd.Flags().BoolVar(&opts.Notification, "toot", false, "Emit a terminal beep after the deploy is complete")
 	cmd.Flags().StringArrayVarP(&opts.Profiles, "profile", "p", nil, "Activate profiles by name")
+	cmd.Flags().StringVarP(&opts.Namespace, "namespace", "n", "", "Run Helm deployments in the specified namespace")
 }
 
 func AddFixFlags(cmd *cobra.Command) {
--- a/pkg/skaffold/config/options.go
+++ b/pkg/skaffold/config/options.go
@@ -23,4 +23,5 @@ type SkaffoldOptions struct {
 	Notification bool
 	Profiles     []string
 	CustomTag    string
+	Namespace    string
 }
--- a/pkg/skaffold/deploy/helm.go
+++ b/pkg/skaffold/deploy/helm.go
@@ -33,14 +33,16 @@ import (
 type HelmDeployer struct {
 	*v1alpha2.DeployConfig
 	kubeContext string
+	namespace   string
 }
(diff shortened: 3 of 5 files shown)


cmd/skaffold/app/cmd/cmd.go
```diff
@@ -72,6 +72,7 @@ func AddRunDevFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&filename, "filename", "f", "skaffold.yaml", "Filename or URL to the pipeline file")
 	cmd.Flags().BoolVar(&opts.Notification, "toot", false, "Emit a terminal beep after the deploy is complete")
 	cmd.Flags().StringArrayVarP(&opts.Profiles, "profile", "p", nil, "Activate profiles by name")
+	cmd.Flags().StringVarP(&opts.Namespace, "namespace", "n", "", "Run Helm deployments in the specified namespace")
```
dlorenc: Would it make sense to make this part of the deploy interface, and make all deployers support it rather than just helm?

afrittoli: Thanks for reviewing! It sounds like a good idea, however the kubectl does not support namespace today, so I think it should be first added there, which I think should go in a separate PR.
The current approach does not prevent us from extending this flag to all deployer later.

dgageot:
@afrittoli the linter is complaining about `helm.go` though

afrittoli:
Rebased and fixed linter warnings.

dgageot:
@afrittoli sorry, the linter is still complaining. Also when you rebase, it will use the latest linter I just merged. It could produce some more errors. Sorry...

afrittoli:
Heh, sorry, I updated my linter install locally now.
Ineffective assignment was complaining so I just declared the var with no assignment.

PR #592, opened 2018-05-24
kaniko: pin image version to v0.1.0

### LINKED ISSUE #591: Pin Kaniko Image tag
https://github.com/GoogleContainerTools/skaffold/blob/2fb273da349176ccae5dfe4476e06e8da75d3838/pkg/skaffold/constants/constants.go#L41

r2d4:
fixes #591

### DIFF
--- a/pkg/skaffold/constants/constants.go
+++ b/pkg/skaffold/constants/constants.go
@@ -38,5 +38,6 @@ const (
 	DefaultDockerForDesktopContext = "docker-for-desktop"
 	GCSBucketSuffix                = "_cloudbuild"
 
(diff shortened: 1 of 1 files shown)


r2d4:
I manually verified this works

jonjohnsonjr:
Thoughts re: pinning to a digest instead of a tag?

dlorenc:
It's effectively the same for kaniko. We have effectively immutable tags.

r2d4:
Done :)

r2d4:
Ah thats clearer. Didn't know about that syntax.

r2d4:
Updated

PR #599, opened 2018-05-25
Fix Helm example

Eronarn:
I had trouble getting this example to work until I used "image.tag" instead of "image". This is also the value that is checked for in the test for Helm deploys: https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/deploy/helm_test.go#L47

So I think the example is just wrong.

### DIFF
--- a/examples/helm-deployment/skaffold.yaml
+++ b/examples/helm-deployment/skaffold.yaml
@@ -13,7 +13,7 @@ deploy:
       namespace: skaffold
       #valuesFilePath: helm-skaffold-values.yaml
       values:
-        image: skaffold-helm
+        image.tag: skaffold-helm
       #setValues get appended to the helm deploy with --set.  
       #setValues:
         #some.key: someValue


Eronarn:
Though come to think of it, I think the behavior here might be wrong. The default Helm template you get on a `helm create` expects `image.repository` and `image.tag` as values, which it concatenates, rather than a tag containing the name of the repository...

dlorenc:
@Eronarn are you still planning to sign the CLA?

Eronarn:
CLA should be signed for Quantopian now.

Does it make sense to merge this, and then for me to send another PR addressing my prior comment about changing how replacement works to better support the default Helm template?

Eronarn:
Ah, bother - I committed this is the web UI, so it won't have my Quantopian email address, but my personal one. So I think the CLA-bot is going to get confused.

Should I make a new PR or force push over that commit into this PR?

dgageot:
@Eronarn I think a whole new PR is easier

dlorenc:
I think this PR is stale now because of the CLA issue. Please reopen if we want to move this one forward.

PR #606, opened 2018-05-28
Skaffold build output

### LINKED ISSUE #604: `skaffold build` should support --output format flag
`skaffold build --output {{.ImageName}}`

The output type should probably be `Build`. This will make `skaffold build` more usable in programmatic situations (scripts, CI).  We already have a `TemplateFlag` that does this behavior. Check out `skaffold docker deps --output` for a similar construction.

dgageot:
add --output and --quiet to `skaffold build`

Fixes #604

### DIFF
--- a/cmd/skaffold/app/cmd/build.go
+++ b/cmd/skaffold/app/cmd/build.go
@@ -18,13 +18,16 @@ package cmd
 
 import (
 	"context"
-	"fmt"
 	"io"
 
+	"github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/flags"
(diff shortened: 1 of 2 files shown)


cmd/skaffold/app/cmd/build.go
```diff
@@ -53,14 +67,19 @@ func build(out io.Writer, filename string) error {
 		return errors.Wrap(err, "creating runner")
 	}
 
-	bRes, err := runner.Build(ctx, out, runner.Tagger, config.Build.Artifacts)
+	buildOut := out
+	if quietFlag {
```
r2d4: Maybe we can set this at the root level and quiet all commands by setting out = ioutil.Discard

PR #612, opened 2018-05-28
Simpler code that logs containers

dgageot:
@r2d4 I think this change is safe but I'd like your feedback.


dgageot:
@r2d4 I believe a watch event will always be received each time a pod's container is actually started. And from that point, the logs can be attached.

PR #613, opened 2018-05-28
Remove afero

dgageot:
This is the last bit of code using aero to juggle with implicit working directory.


pkg/skaffold/util/util.go
```diff
@@ -66,25 +64,27 @@ func StrSliceContains(sl []string, s string) bool {
 
 // ExpandPathsGlob expands paths according to filepath.Glob patterns
 // Returns a list of unique files that match the glob patterns passed in.
-func ExpandPathsGlob(paths []string) ([]string, error) {
+func ExpandPathsGlob(workingDir string, paths []string) ([]string, error) {
```
r2d4: Was this code just wrong before without the workingDir?

dgageot: No. It was using afero that implicitly sets a root dir

Issue #341, opened 2018-04-05
File Watcher ignores new files.


dgageot:
If a file appears, that wasn't tracked in the first place, we just ignore it. Sometimes that's ok, sometimes it's not.

In an ideal world, we should match this new file with the dependencies of the Dockerfile.

Instead of passing a list of files to the Watcher, we should pass a `FileMatcher` that compares a file to the list of dependencies for a Dockerfile.

dgageot:
Fixed with #620

Issue #287, opened 2018-03-26
Error adding watch for files


garvincasimir:
### Expected behavior
Watch should be added for Dockerfile and other deps
### Actual behavior
Error:
adding watch for C:\Users\garvin\projects\kube-task\job\Dockerfile: The parameter is incorrect

### Information
I looked at the code and the error seems to be coming from here:
https://github.com/GoogleCloudPlatform/skaffold/blob/739a5048151dc610956fa7117075add73bff1795/pkg/skaffold/watch/watch.go#L202

Perhaps this is an issue with http://github.com/rjeczalik/notify and Windows?

Skaffold version: v0.2.0
Operating system: Windows 10 pro (build 16299)
Content of skaffold.yaml: 
```
apiVersion: skaffold/v1alpha1
kind: Config
build:
  artifacts:
  - imageName: localhost:5000/kube-task 
    workspace: C:\Users\garvin\projects\kube-task\job
  local: {}
deploy:
  kubectl:
    manifests:
    - paths:
      - k8s/*
      parameters:
        KUBE-TASK: localhost:5000/kube-task
```
### Steps to reproduce the behavior

  1. Run skaffold dev
  2. Error appears after pod deployment

jguo13:
I also had the same issue trying to run skaffold on windows 7 after running skaffold dev, is there a fix for this?

Anakin-Hao:
Running into the same issue.
My Env:
Win 10 Pro
skaffold 0.3.0

F21:
I also ran into the same problem with Windows 10 (latest version) and skaffold 0.3.0.

However, after building master (7349a0f3bc97926c5a894bb1e0af7c63e254a373), this is no longer happens.

To use master, install Go, delete the old skaffold binary and run `go get -u github.com/GoogleCloudPlatform/skaffold/cmd/skaffold`. Assuming `$GOPATH/bin` is in your PATH, you can run `skaffold` from anywhere.

I believe this might be due to switching to fsnotify a few days ago.

dgageot:
Fixed with #620

PR #620, opened 2018-05-30
New code to watch file changes

### LINKED ISSUE #617: Watch does not recompute dependencies
Current flow:
1. Artifact dependencies are computed (expanded)
2. Those computed dependencies are sent to the watcher
3. Watcher watches that file list and directory list

This doesn't handle new files or directories, and sometimes fails on deleted files. Not sure what should happen here - maybe each file change should be matched against the glob pattern given by the artifact. Either way, it would require the watcher to be given more information.

This is particularly painful when using skaffold on a new project that has a dockerfile that copies the entire repository as the build context

dgageot:
+ This uses the mTime based polling to watch artifacts and deployment files.
 + It now supports adding and deleting files
 + It also supports changing Dockerfiles and Bazel files.
 + Extract the cleanup code to support changing the skaffold.yaml

Fixes #617, #287 and #341

### DIFF
--- a/pkg/skaffold/runner/runner.go
+++ b/pkg/skaffold/runner/runner.go
@@ -20,8 +20,8 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
+	"time"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/tag"
@@ -33,18 +33,18 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/watch"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"golang.org/x/sync/errgroup"
 )
+
+const PollInterval = 500 * time.Millisecond
 
 // SkaffoldRunner is responsible for running the skaffold build and deploy pipeline.
 type SkaffoldRunner struct {
 	build.Builder
 	deploy.Deployer
 	tag.Tagger
-	watch.WatcherFactory
-	build.DependencyMapFactory
 
-	builds []build.Artifact
+	watchFactory watch.Factory
+	builds       []build.Artifact
 }
 
 // NewForConfig returns a new SkaffoldRunner for a SkaffoldConfig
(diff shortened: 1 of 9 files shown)


dgageot:
Currently fails hard on an invalid Bazel `WORKSPACE` file

dgageot:
@dlorenc @mattmoor I've rebased the code. Should be good to review

pkg/skaffold/runner/runner.go
```diff
@@ -167,74 +166,52 @@ func (r *SkaffoldRunner) Run(ctx context.Context, out io.Writer, artifacts []*v1
...
-
-	watcher, err := r.WatcherFactory(depMap.Paths())
-	if err != nil {
-		return nil, errors.Wrap(err, "creating watcher")
-	}
+	imageList := kubernetes.NewImageList()
+	colorPicker := kubernetes.NewColorPicker(artifacts)
+	logger := kubernetes.NewLogAggregator(out, imageList, colorPicker)
 
 	deployDeps, err := r.Dependencies()
```
r2d4: I think we'll have to introduce something similar for deploy dependencies so that they are recalculated during the loop.

pkg/skaffold/watch/files.go
```diff
@@ -0,0 +1,174 @@
...
+		select {
+		case <-ticker.C:
+			files, err := walk(w.dirs)
+			if err != nil {
+				return errors.Wrap(err, "listing files in folders")
+			}
+
+			changesSinceLastTick := computeDiff(previousFiles, files)
+			if len(changesSinceLastTick) > 0 {
+				// fmt.Println("First change detected")
```
r2d4: debug level log?

pkg/skaffold/watch/files.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	return &fileWatcher{
+		dirs:         dirs,
+		files:        files,
+		pollInterval: pollInterval,
+	}, nil
+}
+
+func uniqueDirs(dirs []string) []string {
```
r2d4: nit: maybe move this to util.UniqueStrSlice? Seems useful in a few places

pkg/skaffold/runner/runner.go
r2d4: Right now these are mostly static or unimplemented - except kubectl accepts a glob pattern, which can change if new untracked files are added that satisfy the pattern. r.Dependencies() can act like the DependencyMap did for artifacts.

dgageot: Yes, I’d rather handle that in another PR because it’s a bit more complicated. New manifests could be in a folder that’s not watched.

Issue #35, opened 2018-02-06
Subcommands for build and deploy implementations


r2d4:
Exposing the build and deploy steps as single runnable commands
```
$ skaffold build local
$ skaffold deploy helm --chart-path
``` 

We could autogenerate most, if not all, of the command by iterating over `BuildType` and `DeployType` and then generating flags based on the parameters on the subtypes like `LocalBuild`.  Each command would output a go-template `BuildResult` or `DeployResult`.

dlorenc:
Thanks, added this one to our team project to prioritize it. Feel free to take it if you want to give it a try, though :)

jonjohnsonjr:
Build command was added in https://github.com/GoogleContainerTools/skaffold/pull/476

PR #624, opened 2018-06-01
cmd: add skaffold deploy

r2d4:
I might be able to get rid of some of the redundancy in the build struct - it shouldn't really need anything other than the full image tag. Now that we have go-containerregistry, it becomes a lot easier to separate the parts (tag/context)

### DIFF
--- a/cmd/skaffold/app/cmd/deploy.go
+++ b/cmd/skaffold/app/cmd/deploy.go
@@ -1,0 +1,86 @@
+/*
+Copyright 2018 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


r2d4:
fixes #35

r2d4:
ref #625

cmd/skaffold/app/cmd/deploy.go
```diff
@@ -0,0 +1,86 @@
...
+	cmd := &cobra.Command{
+		Use:   "deploy",
+		Short: "Deploys the artifacts",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runDeploy(out, filename)
+		},
+	}
+	AddRunDevFlags(cmd)
+	cmd.Flags().StringSliceVar(&images, "images", nil, "A list of images to deploy")
```
jonjohnsonjr: Any reason not to make these args instead of a flag?

r2d4: If they are args we can't have any more nested subcommands. Not really planning any now but thats been confusing in the past.

Issue #585, opened 2018-05-23
proposal to make values filed in skaffold.yaml generate an override values.yaml


louis-murray:
Currently skaffold will set override values defined in `setValues` through the `--set` flag of helm. I think it would be better to also have a way to generate an override `values.yaml` to handle this. There's some limitations with using `--set` like overriding contents of an array or other complex/nested overrides. I've also seen it cause issues with long strings that contain double quotes. I think that the `values` field would be a good candidate for this. Currently, it's non-intuitive to be able to only set image related values via the `values` field, noted in #193 .

I think this could also potentially address #500.

I'd also be happy to put in the leg work to add this functionality if you guys think it's a good feature to add :) 

### Proposed behavior
use `values` field in skaffold.yaml to populate override values through an override values.yaml
### Actual behavior
can only populates values through --set via `setValues` or `values`, if it's image related
### Information

- Skaffold version: HEAD
- Operating system: macOS
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1alpha2
kind: Config
build:
  tagPolicy:
    sha256: {}
    gitCommit: {}
  artifacts:
  - imageName: dummy.imagerepo.net
    workspace: server
    docker: {}
  local: {}
deploy:
  helm:
    releases:
    - name: test-release
      chartPath: charts/test
      namespace: testing
<!-- currently it's like -->
      values:
        images.server: dummy.imagerepo.net/test-image
<!-- proposing to change to something like -->
      values:
        images:
         server: dummy.imagerepo.net/test-image
        info:
         - can do this too
```

### Steps to reproduce the behavior

1. create a skaffold.yaml using helm and include `values` field

enriched:
I think that this partially addresses #500, but in order to assign a tag that has been defined via tagPolicy would you need #783 as well?

PR #632, opened 2018-06-05
add option to generate override values.yaml based on data passed into skaffold.yaml

louis-murray:
This is the implementation for #585.

When looking at the code more I saw the reason that `values` only contains images so I made a new field called `overrides` that will contain the contents of the generated override values.yaml.

I think the best course of action to clear up any confusion would be to rename the `values` section to something more image related like `artifacts` and then the `overrides` section I've implemented here would become `values`. I'd be happy to include that change in this PR but didn't want to introduce a breaking change without speaking about it first. Otherwise, it's just something to keep note of for the next major version bump

### DIFF
--- a/cmd/skaffold/app/cmd/deploy.go
+++ b/cmd/skaffold/app/cmd/deploy.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 Google LLC
+Copyright 2018 The Skaffold Authors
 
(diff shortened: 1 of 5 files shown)


louis-murray:
I signed it!

dgageot:
@louis-murray Can you please rebase?

louis-murray:
@dgageot done!

PR #640, opened 2018-06-06
Remove no-manifest code.

### LINKED ISSUE #625: Clean up build struct
Currently the build struct looks like 

```golang
// Build is the result corresponding to each Artifact built.
type Build struct {
	ImageName string
	Tag       string
	Artifact  *v1alpha2.Artifact // The artifact used in the build.
}
```

It should really only need

```golang
type Build struct {
   Tag name.Tag // from go-containerregistry
}
```

We might need a function to match it back up to the corresponding artifact, but I don't think that has to live on the struct itself.

dgageot:
It helps simplify the code and makes the kubectl plugin easier
to evolve.

Also fixes #625


pkg/skaffold/deploy/kubectl.go
```diff
@@ -230,33 +174,6 @@ func (k *KubectlDeployer) readRemoteManifest(name string) ([]byte, error) {
...
-func generateManifest(b build.Build) ([]byte, error) {
-	logrus.Info("No manifests specified. Generating a deployment.")
-
-	dockerfilePath := filepath.Join(b.Artifact.Workspace, b.Artifact.DockerArtifact.DockerfilePath)
-	r, err := os.Open(dockerfilePath)
-	if err != nil {
-		return nil, errors.Wrap(err, "reading dockerfile")
-	}
-
-	ports, err := docker.PortsFromDockerfile(r)
```
dlorenc: You should also be able to delete this function from the docker package: `PortsFromDockerfile` and tests.

dgageot: Thanks @dlorenc I removed it too now

errordeveloper:
Ah, cool – I was too tempted to improve this!

PR #641, opened 2018-06-06
Add kustomize deployer

r2d4:


examples/kustomize/patch.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kustomize-test
+  labels:
+    app: kustomize-test
+spec:
+  replicas: 1
+  selector:
```
dlorenc: Can you clear everything in here except the piece you're trying to patch? I think it's just image, but it would serve as a better example to reduce duplication.

r2d4: Yup will do, I lazy copy-pasta'd

r2d4: done

pkg/skaffold/deploy/kubectl.go
```diff
@@ -97,7 +97,7 @@ func (k *KubectlDeployer) Deploy(ctx context.Context, out io.Writer, builds []bu
 		return errors.Wrap(err, "replacing images in manifests")
 	}
 
-	err = k.kubectl(manifests.reader(), out, "apply", "-f", "-")
+	err = kubectl(manifests.reader(), out, k.kubeContext, "apply", "-f", "-")
```
dlorenc: nit: inline these if/errs?

r2d4: done

PR #651, opened 2018-06-08
Update go-containerregistry

dgageot:
I don't know what I'm doing here. 

@jonjohnsonjr could you please take a look at what I had to change in pkg/skaffold/docker/image.go?
Thanks!


pkg/skaffold/docker/image.go
```diff
@@ -133,10 +132,7 @@ func addTag(ref name.Reference, targetRef name.Reference, auth authn.Authenticat
 		return err
 	}
 
-	wo := remote.WriteOptions{
-		MountPaths: []name.Repository{ref.Context()},
-	}
-	return remote.Write(targetRef, img, auth, t, wo)
+	return remote.Write(targetRef, img, auth, t, remote.WriteOptions{})
```
dlorenc: I think we want to keep MountOptions if we can.

dgageot: MountPaths field was removed. I don't know if something else needs to be done. @jonjohnsonjr @mattmoor ?

jonjohnsonjr: The happy path for mounting is now handled by the library, see https://github.com/google/go-containerregistry/pull/186

tl;dr: remote layer reads retain the source repo and we specialize that case on remote layer writes.

jonjohnsonjr:
LGTM

pkg/skaffold/docker/image.go
dlorenc: Ah, nice. Missed that change in the library somehow.

PR #658, opened 2018-06-08
cloudbuild: increase timeout to 20m

r2d4:
literally buy us some time


r2d4:
The mirror only has :1.8 and :latest, so unless that gets updated more frequently, I like the flexibility of being able upgrade go versions sooner (we require 1.10 now)

dlorenc:
We could always just copy it ourselves to somewhere in gcr.

jonjohnsonjr:
+1 to copying it to GCR. Should be faster and more reliable than going over the internet.

PR #648, opened 2018-06-07
Asciidocs and refdocs tooling

balopat:
This is in context of #629. 

See generated example: 
 * HTML: https://storage.googleapis.com/balint-skaffold/docs/index.html
 * PDF: https://storage.googleapis.com/balint-skaffold/docs/index.pdf

Documentation tooling requirements:

- Written in a format that is renderable on Github 
- Content is accessible through an outline (sidebar preferably)
- Enables easy PR workflows
- Easy to review changes 
- Easy for a contributor to propose changes 

Additional goals: 

- Composable pages to reduce duplication
- Ability to style/brand 
- Satisfy the following layout goals: 

Layout goals: 

- Have a great, searchable single page HTML with a clear outline (the same for the PDF)
- Support Github as a renderer 
  - Have a full Github landing page (just like currently) 
  - The examples would still live in their separate folders as standalone documented examples, with a good Github specific README with table of contents 
- Keep duplication minimal
  - The first chapter of the reference docs would come mostly from the Github landing page 
  - Examples would be included in the reference docs

PR #661, opened 2018-06-08
cloudbuild: fix docs generation + adds testing for GCB triggers

### LINKED ISSUE #262: Use --cache-from in container builder
Container Builder does a full build each time. It would be interesting to conditionally use a `--cache-from` option to speed the build up. See https://cloud.google.com/container-builder/docs/speeding-up-builds

### LINKED ISSUE #668: Refactor skaffold docker image to use multi-stage build

balopat:
#648 broke the build - this fixes it
Fixes #262.
Fixes #668.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ VERSION ?= v$(VERSION_MAJOR).$(VERSION_MINOR).$(VERSION_BUILD)
 GOOS ?= $(shell go env GOOS)
 GOARCH = amd64
 BUILD_DIR ?= ./out
+DOCS_DIR ?= ./docs/generated
 ORG := github.com/GoogleContainerTools
 PROJECT := skaffold
 REPOPATH ?= $(ORG)/$(PROJECT)
@@ -86,7 +87,8 @@ release: cross docs
         		-f deploy/skaffold/Dockerfile \
         		--cache-from gcr.io/$(GCP_PROJECT)/skaffold-builder \
         		-t gcr.io/$(GCP_PROJECT)/skaffold:$(VERSION) .
-	gsutil -m cp $(BUILD_DIR)/$(PROJECT)-* gs://$(RELEASE_BUCKET)/releases/$(VERSION)/
+	gsutil -m cp  gs://$(RELEASE_BUCKET)/releases/$(VERSION)/
+	gsutil -m cp $(DOCS_DIR)/* gs://$(RELEASE_BUCKET)/releases/$(VERSION)/docs/
 	gsutil -m cp $(BUILD_DIR)/$(PROJECT)-* gs://$(RELEASE_BUCKET)/latest/
 
 .PHONY: release-build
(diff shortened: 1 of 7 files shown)


balopat:
@r2d4 this should work, I'm running the last tests. 
We will have to introduce the `$SKAFFOLD_IMAGE` and `$SKAFFOLD_BUCKET` substitutions on the GCB triggers for this to work before merging to master!

balopat:
FYI. added the substitution parameters in both triggers:

image

hack/gcb_testing/simulate_build_trigger.sh
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+IMAGE=$1
+BUCKET=$2
+
+gcloud container builds submit . \
+    --config=deploy/cloudbuild.yaml \
+    --substitutions="_SKAFFOLD_IMAGE=$IMAGE,_SKAFFOLD_BUCKET=$BUCKET,COMMIT_SHA=testing_$(date +%s)"
```
r2d4: I'd rather have this be driven off the makefile rather than cloudbuild substitutions.

balopat:
Substitutions allow for testing the cloudbuild scripts on different projects from k8s-skaffold and the skaffold bucket. Otherwise it's impossible to test cloudbuilds without messing with production data.

hack/gcb_testing/simulate_build_trigger.sh
balopat: I am not sure what you mean - I create a target in the Makefile, but the substitutions would stay anyway - unless we want to test the pipelines only in prod.

Makefile
```diff
@@ -81,9 +81,46 @@ integration: install $(BUILD_DIR)/$(PROJECT)
...
+	gsutil -m cp $(BUILD_DIR)/$(PROJECT)-* gs://$(RELEASE_BUCKET)/releases/$(VERSION)/
+	gsutil -m cp $(BUILD_DIR)/$(PROJECT)-* gs://$(RELEASE_BUCKET)/latest/
+
+.PHONY: release-build
+release-build: cross docs
+	docker build \
+    		-f deploy/skaffold/Dockerfile \
+    		--cache-from gcr.io/$(GCP_PROJECT)/skaffold-builder \
+    		-t gcr.io/$(GCP_PROJECT)/skaffold:$(COMMIT) .
+	gsutil -m cp $(BUILD_DIR)/$(PROJECT)-* gs://$(RELEASE_BUCKET)/builds/$(VERSION)/
```
balopat: The folder should be $(COMMIT)

balopat:
@r2d4 ready to go. 
Additional benefits: 
 * the build is 1 minute faster than usual even though it has the asciidocs 
 * it fails if there is a mismatch between the actual git tag and the Makefile tag :)

balopat:
Added the `_RELEASE_BUCKET` substitution on both triggers as well - also tested the triggers themselves on my project - it would be great if we could create the triggers from code but it seems it's a UI only feature (probably due to the Github auth step).

balopat:
with the latest updates the time for builds is between 5 min 13 sec and 3 min 54 sec on my project. It takes 2 min 30 secs to start to see the logs - that might be the time for spinning up the bigger machine, but still faster than 9 mins.

r2d4:
Maybe squash a few of these commits and merge?

Prasanna4743:
??=BD20$

PR #665, opened 2018-06-09
Add support for skaffold.yml as a default config file fixes #225

### LINKED ISSUE #225: `skaffold dev` should default to `skaffold.ya?ml`
It appears as though with the omission of a skaffold config file, it defaults to `skaffold.yaml`. As a user that likes and uses `*.yml` for file extensions, I feel as though skaffold should pick up `skaffold.yml` by default as well.

In a project full of `*.yml` files, it feels odd to be forced to use `skaffold.yaml` by default.

gtseres:


pkg/skaffold/util/util.go
```diff
@@ -120,7 +120,17 @@ func ReadConfiguration(filename string) ([]byte, error) {
...
-		return ioutil.ReadFile(filename)
+		directory := filepath.Dir(filename)
+		baseName := filepath.Base(filename)
+		if baseName != "skaffold.yaml" {
+			return ioutil.ReadFile(filename)
+		}
+		contents, err := ioutil.ReadFile(filename)
+		if err != nil {
+			// Try reading the skaffold.yml file instead
+			return ioutil.ReadFile(filepath.Join(directory, "skaffold.yml"))
```
r2d4: Thanks! one nit: can you add a logrus.Infof() here that outputs the error from trying to read the skaffold.yaml and maybe another that says that we're trying to fall back to skaffold.yml? Otherwise the error will just say that it couldn't find "skaffold.yml"

Issue #517, opened 2018-05-04
Support compose files


dgageot:
Here's a silly idea: Why don't we support running Skaffold with just a compose file. No skaffold.yaml.

 + The list of artifacts can be guessed from the compose file. They are all of `docker build` type.
 + We would deploy using `kubectl` deployer. Transformation from the compose file to Kubernetes manifests could be done using [Kompose](https://github.com/kubernetes/kompose)
 + Everything else works the same. `skaffold dev` would still watch artifact dependencies and show the aggregated logs.

dgageot:
Duplicate of #727

dlorenc:
This isn't a duplicate?

balopat:
closing in favor of #1195 and #1273.

PR #674, opened 2018-06-12
[WIP] support docker compose files

r2d4:
#517 + updating docker/docker vendor

something fun to play around with. I had some difficultly vendoring kompose (it vendors a 2 year old version of kubernetes/kubernetes 😳) so this just shells out for now.


guillaumerose:
On Docker for Mac/Windows, you can also try to use the Stack object we introduce in Docker for Mac and Windows last year. 
See https://github.com/docker/cli/blob/7ad30360c823c4f7748e1ac7ef98ec21414cb691/kubernetes/compose/v1beta2/stack.go#L19

The client is here: https://github.com/docker/cli/blob/7ad30360c823c4f7748e1ac7ef98ec21414cb691/kubernetes/client/clientset/clientset.go#L16

Currently it's only on Mac and Windows but we want it to be available on Linux pretty soon.

Thanks for this! I'm looking forward to using it.

r2d4:
I'll close this. I'd like to see the feature too.

Issue #671, opened 2018-06-11
Enhance the Kaniko builder


r2d4:
Namely

- [x] Surface some of the flags that have been added since the initial release
- [x] The option to use an existing cluster secret instead of creating one from a keyfile passed in.
- [x] Generate unique pod names
- [x] Configurable namespace for kaniko builder

cc @sharifelgamal @priyawadhwa

JonnyBGod:
Also would be good to be able to use Kaniko by passing just the local context and destination skipping the need for a GC Bucket

dlorenc:
A few other improvements:

- Support a configurable namespace for kaniko
- Improve secret management (specifying existing secrets, etc.)
- Generate unique names for the pod (in case of errors with earlier versions, etc.)

dlorenc:
A few other things left to do. Updated in the top comment to track in one spot.

dlorenc:
> Also would be good to be able to use Kaniko by passing just the local context and destination skipping the need for a GC Bucket

I think that's more of a kaniko limitation than a skaffold issue - would you mind opening an issue in that repo with the use case?

priyawadhwa:
You can pass in a local directory as the build context to kaniko with the `--context` flag. I think skaffold would just have to mount the local context into the pod to be able to use it.

dlorenc:
I think the problem would be first getting the source from the local machine into the cluster and onto the node,although it's true skaffold might be able to do that before starting the kaniko container.

r2d4:
Closing this with #803, which upgrades the version to v0.2.0. We can track more specific feature requests in another issue.

PR #678, opened 2018-06-13
Add support to the kaniko builder for secrets that already exist.

dlorenc:
Specifying secrets via local files is a bit burdensome. This PR adds initial support for secrets that already exist in the cluster (the kaniko-secret). We should probably make the name of this configurable.

### DIFF
--- a/examples/kaniko/skaffold.yaml
+++ b/examples/kaniko/skaffold.yaml
@@ -4,8 +4,8 @@ build:
   artifacts:
   - imageName: gcr.io/k8s-skaffold/skaffold-example
   kaniko:
-    gcsBucket: k8s-skaffold
-    pullSecret: ~/secrets/kaniko/kaniko.json
+    gcsBucket: skaffold-debug
+    pullSecretName: e2esecret
 deploy:
   kubectl:
     manifests:
--- a/examples/kaniko/skaffold.yaml
+++ b/examples/kaniko/skaffold.yaml
@@ -4,7 +4,7 @@ build:
   artifacts:
   - imageName: gcr.io/k8s-skaffold/skaffold-example
   kaniko:
-    gcsBucket: skaffold-debug
+    gcsBucket: skaffold-kaniko
     pullSecretName: e2esecret
 deploy:
   kubectl:
--- a/examples/kaniko/skaffold.yaml
+++ b/examples/kaniko/skaffold.yaml
@@ -6,6 +6,7 @@ build:
   kaniko:
     gcsBucket: skaffold-kaniko
     pullSecretName: e2esecret
+    namespace: default
 deploy:
   kubectl:
     manifests:
--- a/integration/run_test.go
+++ b/integration/run_test.go
@@ -165,7 +165,7 @@ func TestRun(t *testing.T) {
 			args:        []string{"run"},
 			pods: []testObject{
 				{
-					name: "getting-started",
+					name: "getting-started-kaniko",
 				},
 			},
 			dir:        "../examples/kaniko",
--- a/integration/run_test.go
+++ b/integration/run_test.go
@@ -160,6 +160,17 @@ func TestRun(t *testing.T) {
 			},
 			dir: "../examples/kustomize",
 		},
+		{
+			description: "kaniko example",
+			args:        []string{"run"},
+			pods: []testObject{
+				{
+					name: "getting-started",
+				},
+			},
+			dir:        "../examples/kaniko",
+			remoteOnly: true,
+		},
 	}
 
 	for _, testCase := range testCases {
--- a/pkg/skaffold/build/kaniko.go
+++ b/pkg/skaffold/build/kaniko.go
@@ -24,7 +24,6 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/tag"
-	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kaniko"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes"
@@ -68,7 +67,7 @@ func (k *KanikoBuilder) Build(ctx context.Context, out io.Writer, tagger tag.Tag
 			return nil, errors.Wrap(err, "reading secret")
 		}
 
-		_, err = client.CoreV1().Secrets(k.KanikoBuild.Namespace).Create(&v1.Secret{
+		if _, err := client.CoreV1().Secrets(k.KanikoBuild.Namespace).Create(&v1.Secret{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:   k.KanikoBuild.PullSecretName,
 				Labels: map[string]string{"skaffold-kaniko": "skaffold-kaniko"},
@@ -76,8 +75,7 @@ func (k *KanikoBuilder) Build(ctx context.Context, out io.Writer, tagger tag.Tag
 			Data: map[string][]byte{
 				constants.DefaultKanikoSecretName: secretData,
 			},
-		})
-		if err != nil {
+		}); err != nil {
 			logrus.Warnf("creating secret: %s", err)
 		}
 		defer func() {
--- a/pkg/skaffold/build/kaniko.go
+++ b/pkg/skaffold/build/kaniko.go
@@ -76,7 +76,7 @@ func (k *KanikoBuilder) Build(ctx context.Context, out io.Writer, tagger tag.Tag
 				constants.DefaultKanikoSecretName: secretData,
 			},
 		}); err != nil {
-			logrus.Warnf("creating secret: %s", err)
+			return nil, errors.Wrapf(err, "creating secret: %s", err)
 		}
 		defer func() {
 			if err := client.CoreV1().Secrets(k.KanikoBuild.Namespace).Delete(k.KanikoBuild.PullSecretName, &metav1.DeleteOptions{}); err != nil {
--- a/pkg/skaffold/build/kaniko.go
+++ b/pkg/skaffold/build/kaniko.go
@@ -57,7 +57,7 @@ func (k *KanikoBuilder) Build(ctx context.Context, out io.Writer, tagger tag.Tag
 
 	if k.KanikoBuild.PullSecret == "" {
 		logrus.Debug("No pull secret specified. Checking for one in the cluster.")
-		if _, err := client.CoreV1().Secrets("default").Get("kaniko-secret", metav1.GetOptions{}); err != nil {
+		if _, err := client.CoreV1().Secrets(k.KanikoBuild.Namespace).Get(k.KanikoBuild.PullSecretName, metav1.GetOptions{}); err != nil {
 			return nil, errors.Wrap(err, "checking for existing kaniko secret")
 		}
 	} else {
@@ -66,9 +66,9 @@ func (k *KanikoBuilder) Build(ctx context.Context, out io.Writer, tagger tag.Tag
 			return nil, errors.Wrap(err, "reading secret")
 		}
 
-		_, err = client.CoreV1().Secrets("default").Create(&v1.Secret{
+		_, err = client.CoreV1().Secrets(k.KanikoBuild.Namespace).Create(&v1.Secret{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:   "kaniko-secret",
+				Name:   k.KanikoBuild.PullSecretName,
 				Labels: map[string]string{"kaniko": "kaniko"},
 			},
 			Data: map[string][]byte{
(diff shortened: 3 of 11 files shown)


dlorenc:
Ref https://github.com/GoogleContainerTools/skaffold/issues/435

This will make the integration tests a bit easier to run in CI since we won't need to manage secrets.

dlorenc:
Ref https://github.com/GoogleContainerTools/skaffold/issues/671

pkg/skaffold/build/kaniko.go
```diff
@@ -48,28 +48,35 @@ func (k *KanikoBuilder) Build(ctx context.Context, out io.Writer, tagger tag.Tag
 		return nil, errors.Wrap(err, "getting kubernetes client")
 	}
 
-	secretData, err := ioutil.ReadFile(k.KanikoBuild.PullSecret)
-	if err != nil {
-		return nil, errors.Wrap(err, "reading secret")
-	}
+	if k.KanikoBuild.PullSecret == "" {
```
r2d4: I was thinking that this would be implemented by passing something into the config rather than just default to an existing hardcoded one

dlorenc: Yeah I mentioned that in the original title, I'm thinking something like:

pullSecret (this is what we have today, a local file)
pullSecretName (the name of the secret in the cluster)
nameSpace (the NS for everything to run in, including the secret)

Does that make sense?

dlorenc:
Now with an integration test!

dlorenc:
```shell
--- PASS: TestRun (118.10s)
    --- PASS: TestRun/getting-started_example (22.13s)
    --- PASS: TestRun/annotated_getting-started_example (4.01s)
    --- PASS: TestRun/getting-started_envTagger (9.15s)
    --- PASS: TestRun/gcb_builder_example (35.15s)
    --- PASS: TestRun/deploy_kustomize (1.42s)
    --- PASS: TestRun/kaniko_example (46.24s)
```

Nice!

pkg/skaffold/build/kaniko.go
r2d4: Yep SGTM

pkg/skaffold/build/kaniko.go
```diff
@@ -48,28 +48,35 @@ func (k *KanikoBuilder) Build(ctx context.Context, out io.Writer, tagger tag.Tag
...
-	if err != nil {
-		logrus.Warnf("creating secret: %s", err)
-	}
-	defer func() {
-		if err := client.CoreV1().Secrets("default").Delete("kaniko-secret", &metav1.DeleteOptions{}); err != nil {
-			logrus.Warnf("deleting secret")
+		_, err = client.CoreV1().Secrets(k.KanikoBuild.Namespace).Create(&v1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:   k.KanikoBuild.PullSecretName,
+				Labels: map[string]string{"kaniko": "kaniko"},
```
r2d4: nit: I know you didn't add this but maybe we should add a skaffold specific identifier here so we could eventually figure out which ones were initialized by skaffold

pkg/skaffold/kaniko/kaniko.go
```diff
@@ -54,10 +54,11 @@ func RunKanikoBuild(ctx context.Context, out io.Writer, artifact *v1alpha2.Artif
 		return "", errors.Wrap(err, "starting log streamer")
 	}
 	imageDst := fmt.Sprintf("%s:%s", artifact.ImageName, initialTag)
-	p, err := client.CoreV1().Pods("default").Create(&v1.Pod{
+	p, err := client.CoreV1().Pods(cfg.Namespace).Create(&v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:   "kaniko",
-			Labels: map[string]string{"kaniko": "kaniko"},
+			Name:      "kaniko",
+			Labels:    map[string]string{"kaniko": "kaniko"},
```
r2d4: same comment here

pkg/skaffold/schema/v1alpha2/config.go
```diff
@@ -238,17 +241,31 @@ func (c *SkaffoldConfig) setDefaultWorkspaces() {
 	}
 }
 
-func (c *SkaffoldConfig) expandKanikoSecretPath() error {
-	if c.Build.KanikoBuild == nil || c.Build.KanikoBuild.PullSecret == "" {
-		return nil
+func (c *SkaffoldConfig) setDefaultNamespace() {
```
r2d4: maybe setDefaultKanikoNamespace?

pkg/skaffold/schema/v1alpha2/config.go
```diff
@@ -238,17 +241,31 @@ func (c *SkaffoldConfig) setDefaultWorkspaces() {
...
+	}
+	if c.Build.KanikoBuild.Namespace == "" {
+		c.Build.KanikoBuild.Namespace = "default"
 	}
+}
 
-	absPath, err := homedir.Expand(c.Build.KanikoBuild.PullSecret)
-	if err != nil {
-		return fmt.Errorf("unable to expand pullSecret %s", c.Build.KanikoBuild.PullSecret)
+func (c *SkaffoldConfig) setDefaultSecret() error {
```
r2d4: setDefaultKanikoSecret?

pkg/skaffold/schema/v1alpha2/config.go
```diff
@@ -238,17 +241,31 @@ func (c *SkaffoldConfig) setDefaultWorkspaces() {
...
+		if err != nil {
+			return fmt.Errorf("unable to expand pullSecret %s", c.Build.KanikoBuild.PullSecret)
+		}
 
-	c.Build.KanikoBuild.PullSecret = absPath
+		c.Build.KanikoBuild.PullSecret = absPath
+		return nil
+	}
+	if c.Build.KanikoBuild.PullSecretName == "" {
+		c.Build.KanikoBuild.PullSecret = "kaniko-secret"
```
r2d4: maybe add this as a constant somewhere, since its references a few times in different files

pkg/skaffold/schema/v1alpha2/config.go
dlorenc: Done.

pkg/skaffold/schema/v1alpha2/config.go
dlorenc: Done

pkg/skaffold/kaniko/kaniko.go
dlorenc: Done

pkg/skaffold/build/kaniko.go
dlorenc: Done

dlorenc:
RFAL.

examples/kaniko/skaffold.yaml
```diff
@@ -4,8 +4,8 @@ build:
   artifacts:
   - imageName: gcr.io/k8s-skaffold/skaffold-example
   kaniko:
-    gcsBucket: k8s-skaffold
-    pullSecret: ~/secrets/kaniko/kaniko.json
+    gcsBucket: skaffold-debug
```
balopat: great, glad this bucket is being used, I think I created it to test logs from cloudbuild :) 
Maybe we could create a specific one for integration tests? Or maybe an example specific one so that the example itself is a bit more self-descriptive?

dlorenc: Done.

dlorenc:
Anything else here? @r2d4

pkg/skaffold/schema/v1alpha2/config.go
r2d4: Doesn't look like you used the constant here

pkg/skaffold/schema/v1alpha2/config.go
```diff
@@ -238,17 +241,31 @@ func (c *SkaffoldConfig) setDefaultWorkspaces() {
...
 
-func (c *SkaffoldConfig) expandKanikoSecretPath() error {
-	if c.Build.KanikoBuild == nil || c.Build.KanikoBuild.PullSecret == "" {
-		return nil
+func (c *SkaffoldConfig) setDefaultKanikoNamespace() {
+	if c.Build.KanikoBuild == nil {
+		return
+	}
+	if c.Build.KanikoBuild.Namespace == "" {
+		c.Build.KanikoBuild.Namespace = "default"
```
r2d4: This should be resolved from the kubernetes client. The default namespace will be different if you configure your context with a different namespace.

pkg/skaffold/kaniko/kaniko.go
```diff
@@ -54,10 +54,11 @@ func RunKanikoBuild(ctx context.Context, out io.Writer, artifact *v1alpha2.Artif
 		return "", errors.Wrap(err, "starting log streamer")
 	}
 	imageDst := fmt.Sprintf("%s:%s", artifact.ImageName, initialTag)
-	p, err := client.CoreV1().Pods("default").Create(&v1.Pod{
+	p, err := client.CoreV1().Pods(cfg.Namespace).Create(&v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:   "kaniko",
-			Labels: map[string]string{"kaniko": "kaniko"},
+			Name:      "kaniko",
+			Labels:    map[string]string{"skaffold-kaniko": "skaffold-kaniko"},
```
r2d4: can we add some more information about how skaffold invoked kaniko? basically the relevant parts of the kaniko skaffold config

pkg/skaffold/build/kaniko.go
```diff
@@ -55,28 +56,34 @@ func (k *KanikoBuilder) Build(ctx context.Context, out io.Writer, tagger tag.Tag
...
+		if _, err := client.CoreV1().Secrets(k.KanikoBuild.Namespace).Create(&v1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:   k.KanikoBuild.PullSecretName,
+				Labels: map[string]string{"skaffold-kaniko": "skaffold-kaniko"},
+			},
+			Data: map[string][]byte{
+				constants.DefaultKanikoSecretName: secretData,
+			},
+		}); err != nil {
+			logrus.Warnf("creating secret: %s", err)
```
r2d4: I guess these were there before but should we really just warn if we fail to create the secret? This should most likely throw an error.

integration/run_test.go
```diff
@@ -160,6 +160,17 @@ func TestRun(t *testing.T) {
 			},
 			dir: "../examples/kustomize",
 		},
+		{
+			description: "kaniko example",
+			args:        []string{"run"},
+			pods: []testObject{
+				{
+					name: "getting-started",
```
r2d4: we should probably rename the pod to not use the getting-started name

pkg/skaffold/schema/v1alpha2/config.go
dlorenc: I switched this up to just pass "" by default, which seems to be what we do everywhere else: https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/kubernetes/log.go#L71

pkg/skaffold/kaniko/kaniko.go
dlorenc: I don't think there's actually anything else in there worth including. Any suggestions?

```
type KanikoBuild struct {
	GCSBucket      string `yaml:"gcsBucket,omitempty"`
	PullSecret     string `yaml:"pullSecret,omitempty"`
	PullSecretName string `yaml:"pullSecretName,omitempty"`
	Namespace      string `yaml:"namespace,omitempty"`
}
```

r2d4: why not all of it? should be useful for debugging if it doesn't invoke what they'd like

pkg/skaffold/schema/v1alpha2/config.go
r2d4: Thats a little different - "" is all namespaces

dgageot:
@dlorenc It's failing on Kokomo with:
```
run_test.go:194: skaffold: Starting build...
4251
    		 Running [skaffold run]: stdout Starting build...
4252
    		, stderr: time="2018-06-18T19:58:47Z" level=fatal msg="build step: checking for existing kaniko secret: an empty namespace may not be set when a resource name is provided"
4253
    		, err: exit status 1: exit status 1
```

dlorenc:
OK, should be ready for a look again. I refactored some of the kube config handling to avoid an import cycle and provide access to the config itself.

examples/kaniko/skaffold.yaml
```diff
@@ -4,8 +4,9 @@ build:
   artifacts:
   - imageName: gcr.io/k8s-skaffold/skaffold-example
   kaniko:
-    gcsBucket: k8s-skaffold
-    pullSecret: ~/secrets/kaniko/kaniko.json
+    gcsBucket: skaffold-kaniko
+    pullSecretName: e2esecret
+    namespace: default
```
r2d4: I don't think we want to run the integration test in the default namespace

pkg/skaffold/kaniko/kaniko.go
r2d4: SGTM

examples/kaniko/skaffold.yaml
dlorenc: It doesn't run there overall, just the kaniko build. The issue is that the secret needs to be pre-created in the cluster in a fixed location, I could put it in a different namespace if we want, but it can't be dynamically created.

r2d4: It could be dynamically created if we pointed to the secret json itself on the builder. We can merge this now, and if we run into trouble switch it to dynamically created.

dlorenc: Yeah, we could definitely store the secret.json in kokoro and inject it into the builder as well.

PR #682, opened 2018-06-14
Implement packaging for helm deployment

hypnoglow:
**A note for reviewers**: this is a proposal. If you give the green light to it, I will add required tests and documentation.

---

This commit adds a possibility to package a helm chart before deployment
using `helm package` command. This can be useful to set a specific version
for a chart.

---

One of the workflows that skaffold assumes is to deploy a Helm chart that
resides within some directory in the project folder. `Chart.yaml` file in
that folder is basically static file, and `--version` flag in
`helm install ./path/to/chart --version` command actually does nothing,
so does "version:" field in helm section of **skaffold.yaml**.

To set a chart version dynamically, we have to use `helm package` command
setting `--version` (and optionally `--app-version`) flags accordingly.

This commit adds a new section `packaged:` to helm section **skaffold.yaml**
that can be used to configure dynamic chart versions.

### DIFF
--- a/examples/annotated-skaffold.yaml
+++ b/examples/annotated-skaffold.yaml
@@ -106,15 +106,27 @@ deploy:
     # helm releases to deploy.
     # releases:
     # - name: skaffold-helm
-    #  chartPath: skaffold-helm
-    #  valuesFilePath: helm-skaffold-values.yaml
-    #  values:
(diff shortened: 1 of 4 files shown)


r2d4:
This SGTM. Tests and documentation would be much appreciated!

Issue #523, opened 2018-05-06
Support WSL and minikube


F21:
### Expected behavior
Skaffold should work under WSL when used with minikube on windows

### Actual behavior
Running `skaffold dev` in WSL (windows subsystem for linux) with minikube does not work at the moment.

This is because skaffold calls `minikube docker-env --shell none` to get the docker configuration, but the `docker_cert_path` points to a windows path: 

```
DOCKER_HOST=tcp://192.168.1.113:2376
DOCKER_CERT_PATH=C:\Users\user\.minikube\certs
DOCKER_API_VERSION=1.23
```

### Information

```
Skaffold version: version: a0879920e6ba86a997a02b4e28ea1e6b84dbdc46
Operating system: Windows 10 1804 64-bit with WSL
Content of skaffold.yaml: n/a
```

Some possible solutions:
- Detect if the environment is WSL and use `wslpath` to convert `DOCKER_CERT_PATH` into a linux path.
- Use the certificates configured in configuration in the `~/.kube` folder.
- Implement configuration for skaffold (under `~/.skaffold`?)

remoe:
@F21, please try the master.

F21:
@remoe I am not sure why, but master does not appear to be looking for minikube:

```
2018/06/20 09:46:42 Unable to read "/home/user/.docker/config.json", falling back on anonymous: open /home/user/.docker/config.json: no such file or directory
Starting build...
FATA[0009] first build: exiting dev mode because the first build failed: running build for artifact: running build: docker build: Cannot connect to the Docker daemon at tcp://0.0.0.0:2375. Is the docker daemon running?
```
My `minikube.exe` is available in my PATH and I have symlinked `/usr/local/bin/minikube` to `minikube.exe`:

```
$ which minikube
/usr/local/bin/minikube
$ which minikube.exe
/c/Minikube/minikube.exe
```

F21:
My minikube is not installed in WSL. It's installed in Windows, but because I prefer using 1 shell, I made the symlink  so that I can execute minikube from WSL.

remoe:
do you have a wrong docker-env?

It works on my side:

```bash
cat ~/.minikube/docker-env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://<ip>:2376"
export DOCKER_CERT_PATH="/mnt/d/Users/<username>/.minikube/certs"
export DOCKER_API_VERSION="1.35"
```

```bash
[...]

PR #694, opened 2018-06-17
add WSL support

remoe:
add WSL support #523

The DOCKER_CERT_PATH environment will be changed with ```wslpath``` tool.

Here you find a documentation how one can use minikube on windows:
- https://medium.com/@ddebastiani/install-kubernetes-on-windows-wsl-c36f6b2571d2
- https://www.jamessturtevant.com/posts/Running-Kubernetes-Minikube-on-Windows-10-with-WSL/

After this installation one need to correct the kube config with:

```bash
USERNAME=$(id -n -u)
kubectl config set-cluster minikube --certificate-authority=/mnt/c/Users/${USERNAME}/.minikube/ca.crt
kubectl config set-credentials minikube --client-certificate=/mnt/c/Users/${USERNAME}/.minikube/client.crt --client-key=/mnt/c/Users/${USERNAME}/.minikube/client.key
kubectl config set-context minikube --cluster=minikube --user=minikube
```

### DIFF
--- a/pkg/skaffold/docker/client.go
+++ b/pkg/skaffold/docker/client.go
@@ -46,15 +46,10 @@ var (
 	dockerAPIClientOnce sync.Once
 	dockerAPIClient     APIClient
 	dockerAPIClientErr  error
-	procVerDetect       bool
-	isWsl               bool
 )
 
 // NewAPIClient guesses the docker client to use based on current kubernetes context.
 func NewAPIClient() (APIClient, error) {
-	procVerDetect = false
-	isWsl = false
-
 	dockerAPIClientOnce.Do(func() {
 		kubeContext, err := kubernetes.CurrentContext()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/skaffold/docker/client.go
```diff
@@ -43,10 +46,15 @@ var (
 	dockerAPIClientOnce sync.Once
 	dockerAPIClient     APIClient
 	dockerAPIClientErr  error
+	procVerDetect       bool
+	isWsl               bool
```
dlorenc: I'm not a fan of having this at both the package level and set through a function. Could we drop this here and always return the result of detectWsl?

pkg/skaffold/docker/client.go
```diff
@@ -123,8 +131,44 @@ func newMinikubeAPIClient() (APIClient, error) {
...
+			if bytes.Contains(b, []byte("Microsoft")) {
+				isWsl = true
+			}
+		}
+	}
+	return isWsl, nil
+}
+
+func getMiniKubeFilename() (string, error) {
+	const winMiniKubeEnv = "SKAFFOLD_WINDOWS_MINIKUBE"
```
dlorenc: Could you move this constant to the package level?

pkg/skaffold/docker/client.go
```diff
@@ -123,8 +131,44 @@ func newMinikubeAPIClient() (APIClient, error) {
...
+	if found, _ := detectWsl(); found {
+		filename, envExists := os.LookupEnv(winMiniKubeEnv)
```
dlorenc: Would it make more sense to use exec.LookPath with minikube.exe instead of this environment variable to find the path to Minikube?

pkg/skaffold/docker/client.go
remoe: then we don't cache. Ok.

pkg/skaffold/docker/client.go
remoe: good idea. done.

remoe:
the travis build error is not from my code:

```
cmd/skaffold/app/flags/template_test.go:30:22: github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/flags.templateData composite literal uses unkeyed fields (govet)
	data             = &templateData{"test"}
	                    ^
pkg/skaffold/build/container_builder.go:75:10: github.com/GoogleContainerTools/skaffold/pkg/skaffold/build.GoogleCloudBuilder composite literal uses unkeyed fields (govet)
	return &GoogleCloudBuilder{cfg}, nil
```

dlorenc:
Sent https://github.com/GoogleContainerTools/skaffold/pull/698 to fix that.

PR #697, opened 2018-06-18
mv tagPolicy:env example

ahmetb:
examples/environment-variables is far too vague (I actually thought I can use
it to pass env vars to be used as variables in skaffold.yaml). So renaming.


dlorenc:
Not sure why travis started failing on this. I sent a PR to fix that issue.

dlorenc:
Ah, looks like a change in metalinter.

Running  `go get -u github.com/golangci/golangci-lint/cmd/golangci-lint` locally made it so I could repro.

Issue #234, opened 2018-03-19
Add a Windows CI


dgageot:
Can a CI be added on [AppVeyor](https://www.appveyor.com)?

r2d4:
I've added some basic kokoro support (a Google-Hosted Jenkins). It has executors on different platforms (macOS, windows, linux). I don't know too much about it, but I plan to run our integration tests there soon.

dgageot:
Fixed with #702

PR #702, opened 2018-06-19
windows: add appveyor

r2d4:
There are some breaking unit tests, but getting started on this.


r2d4:
Had to skip two tests since they won't work on windows, details in the commits.

r2d4:
Fixes #234

Issue #531, opened 2018-05-08
High CPU usage on Mac OS X


aplex:
### Actual behavior
skaffold process generates about 100% sustained cpu usage when it is started with "skaffold dev" command. The project itself is quite small (4 deployments of third-party services and 1 locally built docker image), and the high cpu usage is present even without any local file changes.

VBoxHeadless process also has a sustained 150-200% CPU usage in Activity Monitor.

### Information

```
Skaffold version: v0.5.0
Operating system: Mac OS X 10.12.6
Content of skaffold.yaml:

apiVersion: skaffold/v1alpha2
kind: Config
build:
  artifacts:
  - imageName: my-tools-image
    workspace: ./tools
    docker: {}
  local: {}
deploy:
  kubectl:
    manifests:
    - ./kubernetes/*

```

ajbouh:
Also seeing this behavior on 10.11.6. I've been having issues with running out of file descriptors in other apps, though I'm not sure if that's relevant to what's going on here or not...

slaxor:
I have seen it also on linux (Linux turtle 4.16.7-1-ARCH #1 SMP PREEMPT Wed May 2 21:12:36 UTC 2018 x86_64 GNU/Linux)

ajbouh:
I've instrumented my local build of skaffold and found that this line:

https://github.com/GoogleContainerTools/skaffold/blob/762c3ada56d758913e15dfaadfa73b0e68328e8b/pkg/skaffold/kubernetes/log.go#L74

Seems to suddenly return a nil/empty Event repeatedly. I have been able to trigger it by switching from one internet connection to another. I think the watch channel is closed and we're stuck in an infinite loop because of it.

I'm not certain this is the source of the problem, but I'm experimenting locally with:
```golang
	case evt, ok := <-watcher.ResultChan():
		if !ok {
			// Try reopening
			logrus.Infof("Log aggregator watch has stopped, respawning it.")
			a.Start(ctx, client)
			break
		}
```

ajbouh:
I've been running this for a few days now and while the CPU usage issue is gone, I've seen the `"Log aggregator watch has stopped, respawning it."` message spewed to my logs over and over again.

I guess there's something that goes bad elsewhere in the log watching state machine and just respawning it again is not sufficient. This change is better than nothing as it gives a landmark in the logs to launch deeper investigations from. Anyone have ideas about what might be going on?

dlorenc:
[...]

PR #704, opened 2018-06-19
Fix issue with 100% CPU usage in logs.go.

d11wtq:
Resolves: https://github.com/GoogleContainerTools/skaffold/issues/531
See also: https://github.com/kubernetes/client-go/issues/12

There is an issue in which the Pods watcher gets into a infinite tight
loop and begins consuming 100% of the CPU. This happens after skaffold
dev has been running for a while (~30 mins) and once it starts, it
doesn't stop.

The issue was narrowed down by @ajbouh to the event polling loop in
`logs.go`, which was not checking if the `ResultChan()` is closed or not.
Kubernetes actually closes the connection after a timeout (default is in
the range of 30-60 mins according to the related issue linked to above).
In this case, the intended solution is to start the watcher again.

This refactors the polling into two nested loops. One to start (and
restart) the Pods watcher itself and another to receive and process the
events from the watcher. If the `ResultChan()` is closed, the entire
watcher loop is restarted and log tailing continues.

There is a subtle difference in error handling as a result of this
change. Previously any error returned from `client.Pods("").Watch()`
would be immediately returned from the `Start()` func in `logs.go`. This
is no longer possible since the watcher is initialized in the goroutine
started by that func. As such, in the case the watcher cannot be
initialized, we simply log the error and stop tailing logs. Open to
suggestions as to be a better way to handle this error. Retrying in a
tight loop seems potentially problematic in the error scenario.


d11wtq:
I signed it!

d11wtq:
Is that failing integration test correct? It looks like some kind of CI environment config issue https://ci.appveyor.com/project/r2d4/skaffold/build/1.0.6

> The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command.

dgageot:
@d11wtq Can you try to rebase? that should fix the appveyor bug. You should also sign the CLA with the email you've used for the commits.

d11wtq:
Thanks @dgageot. I'm pretty sure I logged into Google with my @w3style.co.uk address used in the commits. Will try again.

d11wtq:
ff945bc9-a49a-4815-b6e2-e8676b89b570
a2164d16-747c-441a-95ac-1e71c2ed48ab

Not sure what else I need to do here.

dgageot:
@d11wtq I had to update the branch. Can you just confirm that you are ok with that?

d11wtq:
I am cool with that 👍

PR #707, opened 2018-06-19
fix: allow environment variables to be used in helm values

### LINKED ISSUE #486: cannot use environment variables in helm templates

jstrachan:
fixes #486

refactored this PR to reuse existing code for env templates


jstrachan:
awesome thanks!!

PR #710, opened 2018-06-19
Update release_notes.sh

balopat:


hack/release_notes.sh
```diff
@@ -17,6 +17,6 @@
 # a simple little utility to list PRs for release notes:
 # Run ./hack/release_notes.sh --help for more info
 
-go build -o out/listpullreqs ./hack/release_note/listpullreqs.go
+go build -o out/listpullreqs ./hack/release_notes/listpullreqs.go
```
r2d4: on second thought could we just `go run` this?

balopat: nice, I'll remove the script!! 
Newbie-me: I did not know about `go run`.

Issue #685, opened 2018-06-15
Dev with dirty submodules


bhack:
### Expected behavior
skaffold dev could work with dirty submodules
### Actual behavior
skaffold dev fail with dirty submodules cause:
`mq FATA[0267] first build: exiting dev mode because the first build failed: generating tag: reading diff: read <mysubmodulepath>: is a directory`
### Information

- Skaffold version: 0.7

Lawouach:
I am also seeing that issue. However I didn't have it with 0.7 but it appeared with 0.8.

bhack:
@Lawouach  Can you test if is it solved with https://github.com/GoogleContainerTools/skaffold/pull/711?

Lawouach:
@bhack I can confirm that with that fix, I don't get the issue I had with 0.8.0 that was released last week.

PR #711, opened 2018-06-20
Fix bug where dirty submodules broke hash generation

mattkelly:
This is a proposal for a fix for https://github.com/GoogleContainerTools/skaffold/issues/685. Please let me know what you think of the approach! I'm new to the Skaffold code, so hopefully I'm not misunderstanding how this should work.

Regarding splitting the status lines: `??` isn't the only special case (two characters). `MM` is also a valid status, for example. If my understanding of the purpose of that code was correct, then my new implementation should be more robust. Ideally I would have split this out as a separate commit, but it's pretty tangled in there.

-------------

Instead of appending the entire contents of modified files to the sha256
input, we should only append the diff as computed by `git diff` in order
to handle all modifications properly, including submodules.

This also fixes a bug where the processing of the git status output was
too fragile. It makes it more generic by splitting the status string on
fields instead.

Signed-off-by: Matt Kelly <matt.kelly@containership.io>

### DIFF
--- a/pkg/skaffold/build/tag/git_commit.go
+++ b/pkg/skaffold/build/tag/git_commit.go
@@ -21,8 +21,6 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
-	"io"
-	"os"
 	"os/exec"
 	"path/filepath"
 	"sort"
@@ -152,7 +150,8 @@ func commitOrTag(currentTag string, tags []string, opts *Options) string {
 	return fmt.Sprintf("%s:%s", opts.ImageName, currentTag)
 }
 
-// The file state is dirty. To generate a unique suffix, let's hash all the modified files.
+// The file state is dirty. To generate a unique suffix, let's hash the diffs
+// of all modified files.
 // We add a -dirty-unique-id suffix to work well with local iterations.
 func dirtyTag(root string, opts *Options, currentTag string, lines []string) (string, error) {
 	h := sha256.New()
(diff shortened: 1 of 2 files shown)


mattkelly:
Whoops, totally missed that there are tests that need fixing now.

Along those lines, this approach needs some tweaking if it's going to have reproducible results. This is because the default `git diff` output includes some index-dependent headers. For example:

```
diff --git a/pkg/skaffold/build/tag/git_commit.go b/pkg/skaffold/build/tag/git_commit.go
index ad5d08c..4bbed3f 100644 // this is a problem
--- a/pkg/skaffold/build/tag/git_commit.go
+++ b/pkg/skaffold/build/tag/git_commit.go
<actual diff contents follow>
```

We should really only hash the diff content itself. That still might be a problem for `git diff` on submodules, though, because the diff content looks like this:

```
-Subproject commit e82a80da031152a55fed6fec2697c6ab2961f27b
+Subproject commit b75d42f81c22009fb9194592c87e6c606fb7a7c6
```

Before I go off and change the diff command and fix the tests (for non-submodules at least), what do people think of this approach in general?

dlorenc:
This makes sense overall to me. Will you be able to sign the CLA?

mattkelly:
@dlorenc yup, sorry - I started getting that sorted out but now I'm traveling for a few days. I'll have the CLA signed and the tests fixed by Monday. Thanks!

mattkelly:
Fixed tests and rebased / fixed conflicts to incorporate https://github.com/GoogleContainerTools/skaffold/commit/b8cfc379fcc4c18f7323bd015991f147f2f87a28#diff-7b47f8007b648de24c6b5c948a869dd6R171 properly. Two notes:

1. I decided to leave the git command `git diff` as it is. Now that I understand how the tests work (and commit hash generation) at a deeper level, I realized that having specific hashes in the diff is fine.

2. I removed the change to use `strings.Fields()` to separate the status code from the file path. The reason is that the [`XY` form of git status codes](https://git-scm.com/docs/git-status#_short_format) can actually start with a blank (for example ` M`). `strings.Fields()` would miss the blank in this case. That said, I do still think that the existing implementation is fragile/broken because there are many possible two-character status codes - none of which are checked for except for `??`. A bigger problem is that if we change the code to properly support `XY` codes, then we break a lot of unit tests because `go-git` (AFAICT) [only has support for single character status codes](https://github.com/src-d/go-git/blob/master/status.go#L48-L70). So in general I think that this should be solved under a new issue because it's a nontrivial change.

dgageot:
Thanks @mattkelly I'm going to merge your fix. However, in https://github.com/GoogleContainerTools/skaffold/pull/714, I stop computing a sha256 of file changes because of the reasons you mention. Instead I use the imageID when the repo is dirty.

PR #713, opened 2018-06-20
Remove warning for an image that’s built and used by fqn

dgageot:


pkg/skaffold/deploy/kubectl.go
```diff
@@ -35,6 +35,8 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+var warner Warner = &logrusWarner{}
```
dlorenc: Maybe add a comment explaining that this is only used so that we can test warnings?

PR #714, opened 2018-06-20
Improve Git tagger

dgageot:
Makes the git tagger more useable in real world situations:
 + Compute tag by artifact working dir
 + Use imageID instead of source sha256 for dirty repositories
 + Fallback on using imageID for non git repositories (with a warning)

### DIFF
--- a/pkg/skaffold/build/tag/git_commit.go
+++ b/pkg/skaffold/build/tag/git_commit.go
@@ -18,12 +18,9 @@ package tag
 
 import (
 	"bufio"
-	"crypto/sha256"
-	"encoding/hex"
 	"fmt"
 	"os/exec"
 	"path/filepath"
-	"sort"
 	"strings"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
@@ -35,8 +32,7 @@ import (
 )
 
 // GitCommit tags an image by the git commit it was built at.
-type GitCommit struct {
-}
+type GitCommit struct{}
 
 func (c *GitCommit) Labels() map[string]string {
 	return map[string]string{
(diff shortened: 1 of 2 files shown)


pkg/skaffold/build/tag/git_commit.go
```diff
@@ -150,74 +178,65 @@ func commitOrTag(currentTag string, tags []string, opts *Options) string {
...
 
-	sort.Strings(changes)
+	return paths
+}
 
-	var lines []string
-	for _, changedPath := range changes {
-		status := status[changedPath].Worktree
-		lines = append(lines, fmt.Sprintf("%c %s", status, changedPath))
+func nornalizePath(path string) (string, error) {
```
mattkelly: Typo: `normalize`

dgageot: Thanks! Fixing it

pkg/skaffold/build/tag/git_commit.go
```diff
@@ -57,36 +53,44 @@ func (c *GitCommit) GenerateFullyQualifiedImageName(workingDir string, opts *Opt
 func generateNameGitShellOut(workingDir string, opts *Options) (string, error) {
```
r2d4: Now that we're adding some more logic to git-shell impl, are we always going to get the same result with git-shell vs. go-git? Do we care? Should we just drop the library?

pkg/skaffold/build/tag/git_commit.go
```diff
@@ -142,6 +151,25 @@ func runGit(workingDir string, arg ...string) (string, error) {
...
+		line := strings.TrimSpace(scanner.Text())
+		if line != "" {
+			lines = append(lines, line)
+		}
+	}
+
+	return lines, nil
+}
+
 func commitOrTag(currentTag string, tags []string, opts *Options) string {
```
r2d4: why does this need a slice of tags?

Can git describe --tags --always return more than one tag?

```shell
$ git tag test
$ git tag test2
$ git describe --tags --always
test
```

pkg/skaffold/build/tag/git_commit.go
dgageot: The same features should be implemented by both paths. Both paths are also tested. If that’s not the case, it’s an error

pkg/skaffold/build/tag/git_commit.go
dgageot: Go-git returns a slice. Don’t know why

pkg/skaffold/build/tag/git_commit.go
r2d4: SGTM

PR #716, opened 2018-06-20
Add labels from options

### LINKED ISSUE #715: Missing labels from options
Labels coming from command line options such as `cleanup` and `namespace` are not set.

dgageot:
Fixes #715


r2d4:
This means that the options weren't being set at all then

dgageot:
@r2d4 They didn't need to be set since before labels they were only used in the constructor

PR #722, opened 2018-06-21
Improve Kaniko code and ns handling

dgageot:


examples/kaniko/skaffold.yaml
```diff
@@ -6,7 +6,7 @@ build:
   kaniko:
     gcsBucket: skaffold-kaniko
     pullSecretName: e2esecret
-    namespace: default
+    # namespace: default
```
r2d4: The way the test was implemented won't make this work. It reuses a secret in the default namespace instead of create one. What we should really do is use the original codepath and provide the secret file that skaffold will use to create/delete the secret everytime.

PR #737, opened 2018-06-24
Support docker build --cache-from

dgageot:


balopat:
I did think a little bit about keeping the `BuildOptions` around but I guess you're right to get rid of it - if there are skaffold specific image building options we can reintroduce a narrower interface + the output buffer really doesn't have to belong there.

dgageot:
@balopat Thanks for your time!

PR #742, opened 2018-06-25
Print the name of the image we are trying to push

dgageot:


balopat:
oops, that was with my test account.

PR #758, opened 2018-06-27
Fix dev loop

dgageot:


cmd/skaffold/app/cmd/dev.go
```diff
@@ -85,7 +85,10 @@ func devLoop(ctx context.Context, out io.Writer, filename string) error {
 				ctxDev, cancelDev := context.WithCancel(ctx)
 				c <- cancelDev
 				if err := runDev(ctxDev, out, filename); err != nil {
-					logrus.Warnln("dev:", err)
+					logrus.Errorln("dev:", err)
+					cancelMainLoop()
```
nkubala: I think this is getting deferred before we enter the dev loop, does it matter that the `cancel` function might get called twice?

dgageot:
It’s ok. It can be cancelled multiple times. Only the first time is useful
Le mer. 27 juin 2018 à 18:54, Nick Kubala <notifications@github.com> a
écrit :

PR #795, opened 2018-07-05
Improve GCR docs

errordeveloper:
The README looked weird on GitHub. It turns out Github doesn't support many list indentation levels for adoc, hence I put it does as one sentence...  Also, `gcloud docker` is deprecated, so I thought we shouldn't mention it.


dgageot:
Thank you @errordeveloper!

PR #805, opened 2018-07-10
GCB now builds artifacts in //

dgageot:
Merging // build for GCB is much less dangerous than for the local builder.


r2d4:
TODO: maybe we can finally add an integration test for the microservices example (and add a GCB profile to test the parallel builds). Seems like a good candidate to test parallel builds.

PR #806, opened 2018-07-10
Add Git Repository Watcher

r2d4:
Adds another implementation of the filewatcher interface that watches a remote git repository and runs the normal `skaffold dev` workflow off that.

Works exactly like the `skaffold dev` workflow except changes are triggered by new commits instead of file change events.

UX:

`$ skaffold dev --git-repository https://github.com/r2d4/saffold`

Behavior:
* Clones repository into a temporary directory
* Watches for new commits (only to master for now)
* Computes the file diff between last build and current commit
* Conditionally rebuilds artifacts/redeploys based on the rules already computed by skaffold

### DIFF
--- a/pkg/skaffold/runner/runner.go
+++ b/pkg/skaffold/runner/runner.go
@@ -45,6 +45,7 @@ type SkaffoldRunner struct {
 
 	watchFactory watch.Factory
 	builds       []build.Artifact
+	opts         *config.SkaffoldOptions
 }
 
 // NewForConfig returns a new SkaffoldRunner for a SkaffoldConfig
@@ -81,6 +82,7 @@ func NewForConfig(opts *config.SkaffoldOptions, cfg *config.SkaffoldConfig) (*Sk
 		Deployer:     deployer,
 		Tagger:       tagger,
 		watchFactory: watch.NewCompositeWatcher,
+		opts:         opts,
 	}, nil
 }
 
@@ -206,7 +208,7 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*v1
 		return nil, errors.Wrap(err, "starting logger")
 	}
 
-	watcher := r.watchFactory(deployDeps, artifacts, PollInterval)
+	watcher := r.watchFactory(deployDeps, artifacts, PollInterval, r.opts)
 	return nil, watcher.Run(ctx, onDeploymentsChange, onArtifactChange)
 }
 
--- a/pkg/skaffold/watch/artifacts.go
+++ b/pkg/skaffold/watch/artifacts.go
@@ -42,10 +42,16 @@ type artifactWatcher struct {
 }
 
 // NewArtifactWatcher creates an ArtifactWatcher for a list of artifacts.
-func NewArtifactWatcher(artifacts []*v1alpha2.Artifact, pollInterval time.Duration) (ArtifactWatcher, error) {
-	fileWatcher, err := NewFileWatcher(workingDirs(artifacts), pollInterval)
-	if err != nil {
-		return nil, errors.Wrap(err, "creating file watcher")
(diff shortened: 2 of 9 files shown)


pkg/skaffold/runner/runner.go
```diff
@@ -206,7 +208,7 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*v1
 		return nil, errors.Wrap(err, "starting logger")
 	}
 
-	watcher := r.watchFactory(deployDeps, artifacts, PollInterval)
+	watcher := r.watchFactory(deployDeps, artifacts, PollInterval, r.opts)
```
dgageot: I'd love to see the `watchFactory` implementation begins different instead of passing opts here.
I tried to make sure opts don't need to be stored in the Runner by moving all the code related to opts in `NewRunner()`

pkg/skaffold/watch/artifacts.go
```diff
@@ -42,10 +42,16 @@ type artifactWatcher struct {
...
 
 // NewArtifactWatcher creates an ArtifactWatcher for a list of artifacts.
-func NewArtifactWatcher(artifacts []*v1alpha2.Artifact, pollInterval time.Duration) (ArtifactWatcher, error) {
-	fileWatcher, err := NewFileWatcher(workingDirs(artifacts), pollInterval)
-	if err != nil {
-		return nil, errors.Wrap(err, "creating file watcher")
+func NewArtifactWatcher(artifacts []*v1alpha2.Artifact, pollInterval time.Duration, gitRepository string) (ArtifactWatcher, error) {
+	var fileWatcher FileWatcher
+	var err error
+	if gitRepository != "" {
```
dgageot: This `if` would be where we create the WatcherFactory instead

pkg/skaffold/watch/git_repo.go
```diff
@@ -0,0 +1,120 @@
...
+
+func NewGitFileWatcher(url string, pollInterval time.Duration) FileWatcher {
+	logrus.Infof("Using git file watcher: %s", url)
+	return &GitFileWatcher{
+		url:          url,
+		pollInterval: pollInterval,
+	}
+}
+
+func (g *GitFileWatcher) Run(ctx context.Context, callback FileChangedFn) error {
```
dgageot: All the code here considers git is installed on the machine. We try to avoid that requirement in the git tagger. I'm ok with that and I'm wondering if we shouldn't maybe get rid of go-git because it's not a high fidelity clone of git anyways.

pkg/skaffold/watch/git_repo.go
```diff
@@ -0,0 +1,120 @@
...
+	ticker := time.NewTicker(g.pollInterval)
+	defer ticker.Stop()
+	lastCommit, err := getHeadRef()
+	if err != nil {
+		return errors.Wrap(err, "getting initial head ref")
+	}
+	for {
+		select {
+		case <-ticker.C:
+			currentCommit, err := getCurrentCommit(g.url)
```
dgageot: If it was named `remoteCommit` or `currentRemoteCommit` it would be easier for me to understand.

pkg/skaffold/watch/git_repo.go
r2d4: Yeah I initially tried to use it but it didnt have `ls-remote` or `diff` support.

pkg/skaffold/watch/git_repo.go
r2d4: Make sense, will change

pkg/skaffold/runner/runner.go
r2d4: I agree --  I just did the lowest friction implementation to implement it without having too many other changes.  I'll change it to be done the right way

PR #809, opened 2018-07-11
Try to guess GCB projectID from the image name

dgageot:
If a GCB projectId is not provided, we try to get the project from the names of the images.

That helps writing skaffold that are a bit more portable.

### DIFF
--- a/pkg/skaffold/build/container_builder.go
+++ b/pkg/skaffold/build/container_builder.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	cstorage "cloud.google.com/go/storage"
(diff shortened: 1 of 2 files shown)


pkg/skaffold/build/container_builder.go
```diff
@@ -222,6 +230,35 @@ watch:
...
+		return "", errors.Wrap(err, "parsing image name for registry")
+	}
+
+	repoInfo, err := registry.ParseRepositoryInfo(ref)
+	if err != nil {
+		return "", err
+	}
+
+	index := repoInfo.Index
+	if !index.Official {
```
r2d4: @jonjohnsonjr is there a cleaner way to do this in go-containerregistry?

jonjohnsonjr: It might be a little bit cleaner to use ggcr, but you're going to end up having to do string surgery regardless, so I wouldn't bother.

There's no notion of "Official" images on GCR, so this isn't super meaningful.

This also won't work for domain-scoped projects, though those are pretty long deprecated by now.

PR #815, opened 2018-07-12
Revert to default grace period

dgageot:
It can now be overridden with additional flags


dgageot:
@r2d4 Can we merge this one?

PR #820, opened 2018-07-13
Cut 0.10.0 release

r2d4:


r2d4:
We don't need to bump the version anymore since its calculated off the git tag.

r2d4:
Fixes #817

Issue #360, opened 2018-04-09
Image handling does not match values.yaml structure from helm create


afrittoli:
The sample chart created by `helm create` uses nested variables for docker images.
The same is a common practice in stable charts as well:

```
image:
  repository: nginx
  tag: stable
  pullPolicy: IfNotPresent
```

The manifest template fetches the details of image from `.Values` accordingly:

```
      containers:
        - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
```

Skaffold overrides the image using the key specified in its pipeline configuration, and it appends the image_name:tag to it, which is not compatible with the value being split into repository and tag.

### Expected behavior

The rendered manifest should be something like:

```
      containers:
        - name: my-chart
           image: image1:<tag-from-skaffold-driven-build>
```
### Actual behavior

Instead what happens is:

```
      containers:
        - name: my-chart
           image: image1:<tag-from-skaffold-driven-build>:<default tag from chart values>
```
### Information

Skaffold version: v0.2.0 (built from source)
```
$ git describe --tags
v0.2.0-91-g88530bd
```

Operating system: macOS Sierra 10.12.6
Content of skaffold.yaml: 

```
build:
  tagPolicy: sha256
  artifacts:
  - imageName: image1
    workspace: image1
  local: {}
deploy:
  helm:
    releases:
    - name: dev
      chartPath: mychart
      namespace: default
      valuesFilePath: skaffold/dev-values.yaml
      values:
        service1_image.repository: image1
        service2_image.repository: image2
```

I appreciate that there is no fixed structure for how the image variable is defined, but the current behaviour does not help using skaffold with many stock charts, and it forces putting image name and tag together.

r2d4:
Thanks for the descriptive issue. This is something I noticed when developing the helm deployer. I'm not sure what the best path forward is.

We should absolutely work with stock charts that separate image name and tag. I'm not sure the best way to expose this. The helm deployer could just inject tag only by default. How does that sound?

afrittoli:
Thanks for the quick reply.

That sounds good. Probably skaffold should support both the current behaviour of appending the tag to the image as well as injecting <image>.tag as well, perhaps with some extra config, so it doesn't break existing skaffold users.

One idea - a bit convoluted but it supports the existing syntax as well.
[...]

PR #826, opened 2018-07-17
add support for helm image convention vs fqn setting

bivas:
This addresses the issue described in #360 
It will enable the user to determine if the image setup schema is using fully qualified named vs the Helm standard of `image.repository` and `image.tag`

### DIFF
--- a/examples/helm-deployment/skaffold-helm/templates/deployment.yaml
+++ b/examples/helm-deployment/skaffold-helm/templates/deployment.yaml
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
         - name: {{ .Chart.Name }}
-          image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
+          image: {{ .Values.image }}
           imagePullPolicy: {{ .Values.pullPolicy }}
           command: ["/bin/bash", "-c", "--" ]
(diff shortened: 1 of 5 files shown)


wstrange:
Could we please see another release (9.1?) when this is merged!

r2d4:
@bivas looks like this is a breaking change for the helm config, can we add this but default to the existing behavior?

@wstrange Once this gets in we'll do another release

bivas:
@r2d4 Wouldn't it be better to default to the helm convention? I can fix the example to match the new default. I'm aware it will be a breaking change but it will support a much wider range of Helm charts.

r2d4:
I don't want to break users current config, even if its really the right behavior. What do you think about leaving the default behavior as is for now, and then we can merge the feature, try it out, and migrate to `skaffold/v1alpha3` config and swap the defaults?

r2d4:
We can even use `skaffold fix` to perform that for users automatically

bivas:
You're right. Better roll it slow and only enable for users who try the new version.

I'll update the behavior accordingly.

bivas:
@r2d4 @dgageot Can we progress with this?

r2d4:
Thanks!

PR #832, opened 2018-07-19
Skip the deployment if no manifests are defined

ccojocar:


ccojocar:
I signed the CLA but it somehow does not update the status.

PR #833, opened 2018-07-19
Watch mode 4th edition

### LINKED ISSUE #729: New files in kubectl.manifests dir are not discovered until an existing file is modified
Skaffold 0.7

I have a setup like:

```
deploy:
  kubectl:
    manifests:
    - ./kubernetes-manifests/**.yaml
```

while `skaffold dev` is watching, if I create a new manifest like `foo.yaml` and keep saving changes to it, these changes are not detected at all.

So I end up changing an existing .yaml in that dir, now it notices `foo.yaml` and adds it to the watch list. Bug or by design?

dgageot:
Here's yet another rewrite of the watch code!

+ Shorter code
+ Can detect new deployment files
+ Won’t stat the files that are not used by the artifacts (.dockerignored for eg.)

Should fix #729 and #620


r2d4:
Some weird behavior I encountered

* logs are no longer muted during build and deploy
* `skaffold dev` fails on microservices example (but `skaffold run` succeeds)

dgageot:
\o/ Thanks for the review @r2d4

PR #837, opened 2018-07-20
Find docker deps 10x faster

dgageot:
+ Reusing the pattern matcher is 5x faster
 + Using godirwalk is 2x faster


pkg/skaffold/docker/parse_test.go
```diff
@@ -308,6 +308,10 @@ func TestGetDependencies(t *testing.T) {
 			}
 
 			deps, err := GetDependencies(test.buildArgs, workspace, "Dockerfile")
+
+			fmt.Println(test.expected)
```
jonjohnsonjr: Forgot to remove these?

dgageot: indeed

Issue #787, opened 2018-07-02
test flake: kaniko_example


r2d4:
```
    --- FAIL: TestRun/kaniko_example (160.68s)
    	run_test.go:213: skaffold: Starting build...
    		Building [gcr.io/k8s-skaffold/skaffold-example]...
    		[kanikohd2mg kaniko] 2018/07/02 17:08:29 No matching credentials found for index.docker.io, falling back on anonymous
    		[kanikohd2mg kaniko] 2018/07/02 17:09:02 mounted blob: sha256:16b5f22d8b231edd526e2606d6636678e40369ceceaafe500696dc44d4c4166a
    		[kanikohd2mg kaniko] 2018/07/02 17:09:02 mounted blob: sha256:ff3a5c916c92643ff77519ffa742d3ec61b7f591b6b7504599d95a4a41134e28
    		[kanikohd2mg kaniko] 2018/07/02 17:09:02 mounted blob: sha256:c6678747892c3d3b99cafaaf96a55f7b0649bb5e0eda79547fb20c593820dd61
    		[kanikohd2mg kaniko] 2018/07/02 17:09:02 mounted blob: sha256:f32d2ea7337847f228582ff9da53aa97c2a76ec1cbdb95f355fdd31ce5d26c9a
    		[kanikohd2mg kaniko] 2018/07/02 17:09:04 pushed blob sha256:e13e4d94e4a1e7ee3713db4e003c57f954b4a73521313da402f5d0fc6c7d84fd
    		[kanikohd2mg kaniko] 2018/07/02 17:09:04 pushed blob sha256:c31d0497cd4056c6106f53cc701d301cf649a34b43ef526ff9147ce01afb5079
    		[kanikohd2mg kaniko] 2018/07/02 17:09:04 pushed blob sha256:cf3691cce5a76b751275248f10c3636ac61fc684f6bc24fc7899b79bf4945eb6
    		[kanikohd2mg kaniko] 2018/07/02 17:09:04 pushed blob sha256:0e52a30bf621387dea8419808e179eaf0cb7405a86f7ae4503d63f8a51668a08
    		[kanikohd2mg kaniko] gcr.io/k8s-skaffold/skaffold-example:317a9b1a32a9f32cd1c1f614a0d60a42: digest: sha256:a346267ffc190febdf50335711ae89fe946e39222e0b8d3c26d046650dd109a4 size: 1400
    		 Running [skaffold run]: stdout Starting build...
[...]

PR #845, opened 2018-07-20
Switch to gcr for the kaniko builder example.

dlorenc:
The flake was a networking issue with Docker Hub. This might help.


dlorenc:
hopefully fixes #787

PR #851, opened 2018-07-24
Initial support for Knative Build

dgageot:
This PR brings initial support for https://github.com/knative/build. It currently focused on supporting Docker build artifacts. Bazel artifacts should be easy to add.


pkg/skaffold/build/knative/build.go
```diff
@@ -0,0 +1,158 @@
...
+
+	if err := waitForCompletion(builds, build); err != nil {
+		return "", errors.Wrap(err, "waiting for completion")
+	}
+
+	return imageDst, nil
+}
+
+func waitForCompletion(builds build_v1alpha1.BuildInterface, build *v1alpha1.Build) error {
+	for {
```
AkihiroSuda: why busy wait?

pkg/skaffold/build/knative/build.go
```diff
@@ -0,0 +1,158 @@
...
+			GenerateName: "skaffold-docker-build",
+			Namespace:    cfg.Namespace,
+		},
+		Spec: v1alpha1.BuildSpec{
+			ServiceAccountName: cfg.ServiceAccountName,
+			Source: &v1alpha1.SourceSpec{
+				Custom: &v1.Container{
+					Image:   "gcr.io/cloud-builders/gsutil", // TODO(dgageot) way too big
+					Command: []string{"sh", "-c"},
+					Args:    []string{fmt.Sprintf("gsutil cp gs://%s/%s - | tar xvz", cfg.GCSBucket, tarName)},
```
AkihiroSuda: Are these escaped?

pkg/skaffold/build/knative/build.go
dgageot: @AkihiroSuda I'm not sure I understand your comment

pkg/skaffold/build/knative/build.go
dgageot: This piece needs fixing.

pkg/skaffold/build/knative/build.go
AkihiroSuda: Does not this loop consume CPU too much?

dgageot: No, it's done just after the logs are done tailing so it usually succeeds i getting the status the first time. But since there's a few ms between the logs being done and the status being updated, it sometimes requires a retry.

pkg/skaffold/build/knative.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+}
+
+// Build builds a list of artifacts with knative.
+func (k *KnativeBuilder) Build(ctx context.Context, out io.Writer, tagger tag.Tagger, artifacts []*v1alpha2.Artifact) ([]Artifact, error) {
+	// TODO(dgageot): parallel builds
+	var builds []Artifact
+
+	for _, artifact := range artifacts {
+		fmt.Fprintf(out, "Building [%s]...\n", artifact.ImageName)
```
bobcatfish: what do you think about [outputting this in color (blue?)](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/color/formatter.go)?

dlorenc:
Is this ready for a look yet?

dlorenc:
Ping, what's left here?

balopat:
ping @dgageot are you okay for me to fork this and continue where you left it?

bkuschel:
This is a pretty tight integration. It would be neat if skaffold had the capability to use pre-existing Build and BuildTemplate resources rather than create fixed implementations with gcs/kaniko.

bkuschel:
I can take a look at this as well now.  I'll pull this code down and see if i can help take it forward.

bkuschel:
@dgageot so i'm thinking about adding support for knative build resources.  Tricky part is generating the input from the local build context.  This current implementation uses an object store as a drop-off. 
Since we have access to the k8s control plane, it could be feasible transfer the files to a customer "skaffold" knative initcontainer responsible for populating the knative build workspace.

So the sequence would be something like. 
1.  Skaffold starts the build, tgz's the build context
2. It loads the build resource specified in the skaffold.yaml.  Ensure the build input source is using a "skaffold" "custom" input container. ( Not sure what to do if an input source specified therein, as it would need a very specific setup here, perhaps it will be overwritten. Perhaps some flexibility of using an object store drop-off etc. ).
3. Skaffold creates the build resource in k8s
4. knative build starts up and "skaffold" input container immediately goes into a waiting state, waiting for a tar process to start
5. Skaffold execs into the waiting "skaffold" container, having it untar the tgz build context that was sent to it
6. knative sees the tar process exit and exits the container.

Thoughts?

dgageot:
@bkuschel Hi! I just heard that Knative Build is going to evolve a LOT in the upcoming weeks. You might want to hold on any effort and see how it turns out

csantanapr:
@dgageot Is there some doc or example on how to use it with knative?

PR #852, opened 2018-07-24
Support setting namespace for every deployer

dgageot:
via flag or env variable.


pkg/skaffold/deploy/kustomize.go
```diff
@@ -107,3 +109,22 @@ func buildManifests(kustomization string) (io.Reader, error) {
 	}
 	return bytes.NewReader(out), nil
 }
+
+// TODO(dgageot): this code is already in KubectlDeployer
```
balopat: why wouldn't we just add the namespace parameter to the kubectl method which we would leave as a package level utility? I'm not sure I follow why both the kustomizeDeployer and the KubeCtlDeployer need kubectl command. What is your plan here?

integration/run_test.go
```diff
@@ -201,70 +145,69 @@ func TestRun(t *testing.T) {
...
 	}
 }
 
 func setupNamespace(t *testing.T) (*v1.Namespace, func()) {
-	namespaceName := util.RandomID()
 	ns, err := client.CoreV1().Namespaces().Create(&v1.Namespace{
 		ObjectMeta: meta_v1.ObjectMeta{
-			Name:      namespaceName,
-			Namespace: namespaceName,
+			GenerateName: "skaffold",
```
balopat: nice!

integration/run_test.go
```diff
@@ -164,31 +123,16 @@ func TestRun(t *testing.T) {
...
-			dir:        "../examples/kaniko",
-			remoteOnly: true,
+			pods:        []string{"getting-started-kaniko"},
+			dir:         "../examples/kaniko",
+			remoteOnly:  true,
 		},
 		{
 			description: "helm example",
 			args:        []string{"run"},
-			deployments: []testObject{
```
balopat: +1 for the yagni cleanup.

pkg/skaffold/deploy/kustomize.go
dgageot: I don't really have a plan. The thing is that both deployer deploy with `kubectl`. I wanted to embed `KubeCtlDeployer` into `kustomizeDeployer` but I'm not sure it's really better than a little duplication.
Another plan would be to extract this function and the state it requires into a shared package

dgageot: I extracted the kubectl CLI code to a separate package to make easier to share between both deployers.

Issue #836, opened 2018-07-19
Unable to retrieve deployment info if helm release name is templated


johnmshields:
### Expected behavior

A helm deployment should retrieve the deployment information after deployment and does not return a warning.

### Actual behavior

The deployment information is not retrieved and a warning is displayed when the release name includes a templated value.

```bash
WARN[0035] error retrieving helm deployment info: Error: invalid release name, must match regex ^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])+$ and the length must not longer than 53
```

The warning above occurs due to the code [here](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/deploy/helm.go#L239) using `r.Name` instead of `releaseName` from [here](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/deploy/helm.go#L115).

So line 239 should be changed from:

```golang
return h.getDeployResults(ns, r.Name), helmErr
```
to
```golang
return h.getDeployResults(ns, releaseName), helmErr
```

dlorenc:
@johnmshields are you interested in sending a PR for this?

johnmshields:
Sure, I'd be happy to... just haven't gone through any of the CLA stuff, yet. I'll put that together ASAP.

dgageot:
Fixed with #855

PR #855, opened 2018-07-26
Issue #836: Use releaseName to get release info.

johnmshields:
The helm release name can be templated so within the `deployRelease`
function the `releaseName` variable should be used any time the release
name is desired.

This commit updates the final call to `getDeployResults` to properly use
the `releaseName` variable.

#836


dgageot:
@johnmshields awesome job. Thanks!

PR #850, opened 2018-07-24
Add a pop of color to terminal output

bobcatfish:
This commit breaks up the Skaffold terminal output with a pop of color!

Skaffold draws a distinction between log output, which goes to stdout,
and output from Skaffold to the user, which goes to stdout and is
interleaved with output from the tools Skaffold users (e.g. docker for
building). This commit changes the output from Skaffold to the user to
be Blue!

The escape codes to make the text blue (or other colors, in the case of
the kube pod-specific logging) will only be output if stdout is a
terminal. If output is being redirected to a file for example, the
escape codes will not be included.

This change makes use of the exciting `colors.go` logic but splits it
into the picker (specifically for mapping images to colors consistently)
from the logic to add escape codes.

Possible alternatives/improvements:
* Pass the formatter object around directly instead of `io.Writer`. One
  big downside to this is it would make it harder for callers to control
  the color of the output (right now it can be arbitrarily changed
  anywhere, e.g. `fix.go` outputs both blue text and plain text)
* Split the formatting from the output.

Fixes #835 

Some example output:
image

Issue #835, opened 2018-07-19
Colorize skaffold output vs. piped output


r2d4:
Imagine the bolded lines being colorized

**Starting build...**
**Building [gcr.io/k8s-skaffold/leeroy-web]...**
Sending build context to Docker daemon  3.072kB
Step 1/6 : FROM golang:1.10.1-alpine3.7 as builder
 ---> 52d894fca6d4
Step 2/6 : COPY web.go .
 ---> Using cache
 ---> a2f24d7c8473
Step 3/6 : RUN go build -o /web .
 ---> Using cache
 ---> bccc7a1c55a9
Step 4/6 : FROM alpine:3.7
 ---> 3fd9065eaf02
Step 5/6 : CMD ["./web"]
 ---> Using cache
 ---> 4594ba11faff
Step 6/6 : COPY --from=builder /web .
 ---> Using cache
 ---> 0c3334a419ab
Successfully built 0c3334a419ab
Successfully tagged 87851671d9c0322b1e2ae520ac185db1:latest
Successfully tagged gcr.io/k8s-skaffold/leeroy-web:v0.10.0-15-g9be665d0
The push refers to repository [gcr.io/k8s-skaffold/leeroy-web]
0939c7023aea: Preparing
cd7100a72410: Preparing
0939c7023aea: Layer already exists
cd7100a72410: Layer already exists
v0.10.0-15-g9be665d0: digest: sha256:b282f3080f95da7e3d07cf8c8b8a04fa0757f443e39e685b1afb0136635432d3 size: 739
**Building [gcr.io/k8s-skaffold/leeroy-app]...**
Sending build context to Docker daemon  3.072kB
Step 1/6 : FROM golang:1.10.1-alpine3.7 as builder
 ---> 52d894fca6d4
Step 2/6 : COPY app.go .
 ---> Using cache
 ---> 567cbfbce599
Step 3/6 : RUN go build -o /app .
 ---> Using cache
 ---> c81ce05f8516
Step 4/6 : FROM alpine:3.7
 ---> 3fd9065eaf02
Step 5/6 : CMD ["./app"]
 ---> Using cache
 ---> b9637259e7db
Step 6/6 : COPY --from=builder /app .
 ---> Using cache
 ---> b59a8f9b007f
Successfully built b59a8f9b007f
Successfully tagged 3a33a2a3607960cf6853fd8fb18e362f:latest
[...]

PR #856, opened 2018-07-27
Implement terminal color using wrapper around io.Writer

bobcatfish:
(I don't want to actually propose this as a solution but I want to make a PR to demonstrate what this proposal would look like)

In #850 I proposed an implementation of terminal color which involves
instantiating a formatter object whenever you want to log to the
terminal in color. This PR instead creates a wrapper around io.Writer
instead, so that theoretically everything output to the terminal will
magically be the right color without callers needing to be aware.

That was how it was in the first iteration HOWEVER when I ran `skaffold
dev` _everything_ was blue, including output from docker, etc. It became
clear that we needed to often use the output writer which didn't apply
formatting, so in the end we need to change pretty much every function
that took io.Writer to take color.Writer instead, so we can control
whether the functions they call output in color or not.

The end result is that when you look at some code, if it still takes
io.Writer as a param, it's very hard to know if that will be in color or
not and I feel this implementation is more confusing and less
maintainable than #850.

(This was a possible fix for #835.)


bobcatfish:
Initially I didn't like this idea (when discussed with @r2d4) but after looking at similar libs (such as [`tabwriter`](https://golang.org/pkg/text/tabwriter/) and [the docker cli does](https://github.com/docker/cli/blob/a6feb55a4821fd2b77fb7412e2955914359ed99d/cli/command/stream.go)) it seemed reasonable. But when I actually tried it, it spread everywhere in the codebase and was very hard to reason about.

PR #866, opened 2018-08-02
Update check

r2d4:
* uploads a VERSION file to latest bucket
* non blocking check for updates on every command, i.e. command will not wait for version check to happen before exiting
* if theres an update, it prints out a message at the very end of output
* won't check on dirty trees - the way that we inject the version in the makefile makes it so that if you don't explicitly pull tags you'll be left with an "old" version.

merge this after this release

### DIFF
--- a/cmd/skaffold/app/cmd/cmd.go
+++ b/cmd/skaffold/app/cmd/cmd.go
@@ -17,11 +17,14 @@ limitations under the License.
 package cmd
 
 import (
+	"context"
+	"fmt"
 	"io"
 
 	cmdutil "github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/cmd/util"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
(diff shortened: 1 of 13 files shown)


r2d4:
Fixes https://github.com/GoogleContainerTools/skaffold/issues/148

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -64,6 +82,25 @@ func NewSkaffoldCommand(out, err io.Writer) *cobra.Command {
...
+	current, err := version.ParseVersion(version.Get().Version)
+	if err != nil {
+		return errors.Wrap(err, "parsing current semver, skipping update check")
+	}
+	latest, err := update.GetLatestVersion(context.Background())
+	if err != nil {
+		return errors.Wrap(err, "getting latest version")
+	}
+	if latest.GT(current) {
+		ch <- fmt.Sprintf("There is a new version (%s) of skaffold available.", latest)
```
balopat: nit: provide a download link

r2d4: good idea, done

r2d4:
We should also probably tell users how to disable this by setting `SKAFFOLD_UPDATE_CHECK=false`

PR #870, opened 2018-08-03
Fix edge cases in build arg processing

jakehow:
With #828 recently merged I tested in some Dockerfiles where we are using build args.

The current implementation misses some functionality and does not handle a couple of edge cases mostly detailed in the documentation here https://docs.docker.com/engine/reference/builder/#arg

* Build Args support default values: `ARG <name>[=<default value>]`
* Docker parses instructions even when there is more than one space between the command and the value.  EX: `ARG  FOO` is valid
* Usage of the build arg can be done with curly braces eg: `COPY ${FOO}`

There are also two additional issues that I didn't have in my files, but should be handled to match the behavior of `docker build`... that are not currently handled by this PR

* Inline defaults: `COPY ${FOO:-server.go}`
* Args are scoped to replace values only below where they are declared in the Dockerfile: 
  For example if we have a build arg FOO set to `start`
   
  ```
  FROM ubuntu:14.04 
  COPY ${FOO:-server.go} .
  ARG  FOO
  CMD $FOO
  ```

  will evaluate to:

  ``` 
  FROM ubuntu:14.04
  COPY server.go .
  ARG  FOO
  CMD start
  ```

Additional tests have been added for the variations handled by this PR and pass for me.

### DIFF
--- a/pkg/skaffold/docker/parse.go
+++ b/pkg/skaffold/docker/parse.go
@@ -60,13 +60,26 @@ func readDockerfile(workspace, absDockerfilePath string, buildArgs map[string]*s
 	for i, value := range res.AST.Children {
 		switch value.Value {
 		case command.Arg:
-			var val string
-			arg := strings.Split(value.Original, " ")[1]
-			valuePtr := buildArgs[arg]
-			if valuePtr == nil {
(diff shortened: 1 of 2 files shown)


jakehow:
@googlebot I've signed the CLA

PR #880, opened 2018-08-07
Add missing filename to error message

tstromberg:


dgageot:
LGTM. Strange that the linter didn't catch this one. Will see if linter's config can be updated

tstromberg:
I only noticed this because I have as separate PR in the works to update our linter to a more recent version, which happened to catch this.

Issue #543, opened 2018-05-12
Production vs Development environment variables.


clanstyles:
What is the right way to store values in a deployment but switch them out for production? If I include secrets, do I mock the secrets with blank data and set them by hand? If the YAML for the secret changes, Skaffold would automatically update them. Some of these values aren't passwords, they might just be flags for "development" to switch out hostnames.

What's the recommended workflow for this?

jstrachan:
if you use Helm for packaging up your application, you can put environment specific configuration values in your `values.yaml` file in your chart - then any installation of your chart can override any of those configuration values via `helm install mychart --set foo.bar=123`

On the Jenkins X project we're big fans of GitOps so prefer to store these environment specific configuration values in a git repository - each environment gets its own git repository to manage those values.

Then to inject the values from `values.yaml` into your actual deployment you can either expose the values as environment variables - or you can put the YAML inside a `ConfigMap` and mount it to the file system as a configuration file. e.g. if you are using `Spring Boot` you could mount the YAML from `values.yaml` as your `application.yml` file on your classpath

dlorenc:
Let's use this issue to track the overall environment variable design/support. There have been a number of issues about this that I'll dupe against this bug.

stevenbarragan:
There is [helm secrets](https://github.com/futuresimple/helm-secrets), I'm trying to figure out how to put it together with skaffold, has anybody tried it?

cpoole:
This would be super helpful for us. 

We have a static container that serves a react app. This app needs to be compiled/minified/etc when we go to dev through prod. But on our local machines we would like to skip the full compile and just run a continuous webpack server. Waiting a full minute for a total react rebuild isnt reasonable. 

If we could pass a build arg to the dockerfile we could skip the react build in the dockerfile and this would be much more reasonable.

Place1:
It'd be nice if skaffold would allow you to defined variables that are rendered into your manifests. These ideally could be global or profile specific and could be sourced from environment variables e.g.

```yaml
apiVersion: skaffold/v1alpha2
kind: Config
profiles:
- name: prod
  # ...  
  deploy:
    kubectl:
      manifests: ./k8s/*
    values:
      domain: "myapp.com"
- name: feature-branch
  # ...  
  deploy:
    kubectl:
      manifests: ./k8s/*
    values:
      domain: "${CI_ENVIRONMENT_SLUG}.myapp.com"
```

then your k8s manifests could have the following:
```yaml
[...]

PR #892, opened 2018-08-11
added support for manifest templating with environment variables and values key in skaffold.yaml

Place1:
Sorry for not communicating at all about this feature idea, I've been thinking this could be a great addition to the tool for awhile but I just never got around to creating a feature request. Then today I read #543 and wrote a response for this idea, thinking it might be a good solution.

After checking every open/closed issue I couldn't find anyone else talking about this feature so I decided to jump in and implement it myself (i'm a golang newbie).

I think the comment I wrote on #543 sums up the feature best so i'll copy it here (with a few minor edits):

---

It'd be nice if skaffold would allow you to defined variables that are rendered into your manifests. These ideally could be global or profile specific and could be sourced from environment variables e.g.

```yaml
apiVersion: skaffold/v1alpha2
kind: Config
profiles:
- name: prod
  # ...  
  deploy:
    kubectl:
      manifests: ./k8s/*
    values:
      namespace: "prod"
      domain: "myapp.com"
- name: feature-branch
  # ...  
  deploy:
    kubectl:
      manifests: ./k8s/*
    values:
      namespace: "{{.CI_ENVIRONMENT_SLUG}}"
      domain: "{{.CI_ENVIRONMENT_SLUG}}.myapp.com"
```

then your k8s manifests could have the following:
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: myservice-ingress
  namespace: "{{.namespace}}"
spec:
  rules:
  - host: "myservice.{{.domain}}"
    http:
      paths:
      - path: /
        backend:
          serviceName: myservice
          servicePort: 80
```

I think this also allows skaffold to fill a nice middle ground between basic deployments with static manifests and helm deployments which adds a cluster component and multiple extra configuration files that feel like boilerplate when all you need is some basic template rendering.

---

### DIFF
--- a/examples/manifest-templating/Dockerfile
+++ b/examples/manifest-templating/Dockerfile
@@ -1,0 +1,7 @@
+FROM golang:1.10.1-alpine3.7 as builder
+COPY main.go .
+RUN go build -o /app main.go
+
+FROM alpine:3.7  
+CMD ["./app"]
+COPY --from=builder /app .
--- a/examples/manifest-templating/README.adoc
+++ b/examples/manifest-templating/README.adoc
@@ -1,0 +1,33 @@
+=== Example: Getting started with a simple go app
+:icons: font
+
+This is a simple example based on
+
+* *building* a single go file app and with a multistage `Dockerfile` using local docker to build
+* *tagging* using the default tagPolicy (`gitCommit`)
+* *deploying* a single container pod using `kubectl`
+
+ifndef::env-github[]
+==== link:{github-repo-tree}/examples/getting-started[Example files icon:github[]]
+
+[source,yaml, indent=3, title=skaffold.yaml]
+----
+include::skaffold.yaml[]
+----
+
+[source,go, indent=3, title=main.go, syntax=go]
+----
+include::main.go[]
+----
+
+[source,docker, indent=3, title=Dockerfile]
+----
+include::Dockerfile[]
+----
+
+[source,yaml, indent=3, title=k8s-pod.yaml]
+----
+include::k8s-pod.yaml[]
+----
+
+endif::[]
--- a/examples/manifest-templating/k8s-pod.yaml
+++ b/examples/manifest-templating/k8s-pod.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{.Name}}
+spec:
+  containers:
+  - name: getting-started
+    image: gcr.io/k8s-skaffold/skaffold-example
--- a/examples/manifest-templating/main.go
+++ b/examples/manifest-templating/main.go
@@ -1,0 +1,14 @@
+package main
+
+import (
+	"fmt"
+	"time"
+)
+
+func main() {
+	for {
+		fmt.Println("Hello world!")
+
+		time.Sleep(time.Second * 1)
+	}
+}
--- a/examples/manifest-templating/skaffold.yaml
+++ b/examples/manifest-templating/skaffold.yaml
@@ -10,7 +10,10 @@ deploy:
     values:
       Name: my-pod
 profiles:
-  - name: gcb
-    build:
-      googleCloudBuild:
-        projectId: k8s-skaffold
+  - name: prod
+    deploy:
+      kubectl:
+        manifests:
+          - k8s-*
+        values:
+          Name: my-production-pod
--- a/examples/manifest-templating/skaffold.yaml
+++ b/examples/manifest-templating/skaffold.yaml
@@ -1,0 +1,16 @@
+apiVersion: skaffold/v1alpha2
+kind: Config
+build:
+  artifacts:
+  - imageName: gcr.io/k8s-skaffold/skaffold-example
+deploy:
+  kubectl:
+    manifests:
+      - k8s-*
+    values:
+      Name: my-pod
+profiles:
+  - name: gcb
+    build:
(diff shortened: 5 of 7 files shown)


Place1:
@googlebot I signed it!

Place1:
@googlebot I signed it!

Place1:
I've added an example in `examples/manifest-templating` that outlines what this feature looks like.

balopat:
Hi @Place1, thank you for the PR and for thinking about this, it is an important aspect of Skaffold that we are trying to figure out and it's great that you started to contribute in golang! However I don't think this aligns with our current thinking: manifest customization is something that we'd like to delegate to solutions downstream, for example kustomize or helm, skaffold itself is not aiming to implement another templating layer for the manifests for now.

Place1:
Hey @balopat cheers for the consideration and I understand your position on this.

But I'll add a few extra thoughts just incase 😛

The motivation to add templating support for the kubectl deployer is to help simplify the tooling requirements for deploying a service.

Here's my company's Gitlab CI use case:

1. build multiple images and tag them with $CI_COMMIT_SHA
2. deploy k8s manifests
    - using $CI_ENVIRONMENT_SLUG as namespace for all resources
    - using $CI_ENVIRONMENT_SLUG as a prefix on all ingress hostnames

We could use a tool like kustomize or helm but it's overkill for basic envsubst functionality.
In the case of helm, you're adding a server side component which developers then need to learn how to secure and operate. Alternatively Kustomize introduces a totally different approach with manifest patching which quite verbose and not as flexible as using template variables in my opinion, patching requires you to understand the structure of the resource thing you're patching, and while Kustomize does support variables, they support is not general [see docs](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/kustomization.yaml#L161).

We were initially drawn to Skaffold because we saw the `envTemplate` feature and saw it solved our image tagging issue. Templating other values in the manifest feels like a natural extension to what skaffold already does - take a manifest and replace values (image names).

So, if adding templating to the kubectl deployer is not on the table, what about adding an `envsubst` or `template` deployer that allows Skaffold to fill this gap? I'd be more than happy to provide the leg work to build and document it 😄

Place1:
I've spent some time modelling this same use-case with helm and here's what i've ended up with:

```yaml
apiVersion: skaffold/v1alpha2
kind: Config
build:
  tagPolicy:
    envTemplate:
      template: "{{.CI_REGISTRY_IMAGE}}/{{.IMAGE_NAME}}:{{.DIGEST_HEX}}"
  artifacts:
    - imageName: my-service
      workspace: ./my-service/
    - imageName: another-service
      workspace: ./another-service/
deploy:
  helm:
    releases:
      - name: my-service
        chartPath: ./my-service/k8s
        setValueTemplates:
          image: "{{.CI_REGISTRY_IMAGE}}/{{.IMAGE_NAME}}:{{.DIGEST_HEX}}"
          domain: "{{.CI_ENVIRONMENT_SLUG}}.mydomain.com"
      - name: another-service
        chartPath: ./another-service/k8s
        setValueTemplates:
          image: "{{.CI_REGISTRY_IMAGE}}/{{.IMAGE_NAME2}}:{{.DIGEST_HEX2}}"
          domain: "{{.CI_ENVIRONMENT_SLUG}}.mydomain.com"
      - name: ingress
        chartPath: ./k8s/ingress
        setValueTemplates:
          domain: "mydomain.com"
      - name: monitoring
        chartPath: ./k8s/monitoring
        setValueTemplates:
          domain: "mydomain.com"
```

This is for using skaffold to automate the build/deploy of:
- 2 backend services
- ingress setup (traefix + consul)
- monitoring setup (prometheus + grafana)

The only reason to use helm here is to be able to render some variables in the k8s manifests, specifically for the ingress resource's host rules, to specify a domain  name that is environment dependent.

As a result of using helm i've had to:
- restructure all my `k8s` folder layout to follow what helm wants/suggests
- add 8 new config files: `Chart.yaml` and `values.yaml` for 4 different _apps_
- duplicate various variable definitions in my `skaffold.yaml` file because each helm release requires the same `domain` variable.

With the feature proposal in this PR it'd look like this:

```yaml
apiVersion: skaffold/v1alpha2
kind: Config
build:
  tagPolicy:
    envTemplate:
      template: "{{.CI_REGISTRY_IMAGE}}/{{.IMAGE_NAME}}:{{.DIGEST_HEX}}"
  artifacts:
    - imageName: registry/user/my-service
      workspace: ./my-service/
    - imageName: registry/user/another-service
      workspace: ./another-service/
deploy:
  kubectl:
    manifests:
      - ./k8s/**
      - ./my-service/k8s/**
      - ./another-service/k8s/**
    values:
      domain: "mydomain.com"
```

With this i wouldn't have to add any new configuration files or install a cluster side component (yay skaffold). Additionally, my manifests look the same as when I use helm because they both use golang templating.

Sorry to be writing so much on this 😬 just trying to explain where i'm coming from.

dlorenc:
Have you tried kustomize or other templating systems? I don't doubt that there is room for improvement here, I just don't believe skaffold is the right place for these improvements.

Place1:
Alright no worries. I'll close this PR.

Place1:
@balopat thanks, i'd be happy to do the leg work building it. thanks for the consideration.

Place1:
@balopat any update on an envsubst deployer?

Issue #635, opened 2018-06-05
Support --target for Docker artifacts


dgageot:
This makes it possible to define multiple artifacts from the same Dockerfile

bryanlarsen:
There's a workaround for --target in #890.

thomas-riccardi:
@dlorenc As it is only a workaround, shouldn't this issue be kept open?

leahneukirchen:
It would be very nice if this was documented.

PR #894, opened 2018-08-11
Add the --target flag as a parameter to the docker builder.

dlorenc:
Fixes https://github.com/GoogleContainerTools/skaffold/issues/635

### DIFF
--- a/pkg/skaffold/docker/image_test.go
+++ b/pkg/skaffold/docker/image_test.go
@@ -171,18 +171,53 @@ func TestDigest(t *testing.T) {
 }
 
 func TestGetBuildArgs(t *testing.T) {
-	artifact := &v1alpha2.DockerArtifact{
-		BuildArgs: map[string]*string{
-			"key1": util.StringPtr("value1"),
-			"key2": nil,
+	tests := []struct {
+		description string
+		artifact    *v1alpha2.DockerArtifact
+		want        []string
+	}{
+		{
+			description: "build args",
+			artifact: &v1alpha2.DockerArtifact{
+				BuildArgs: map[string]*string{
+					"key1": util.StringPtr("value1"),
+					"key2": nil,
+				},
+			},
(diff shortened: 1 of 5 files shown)


dgageot:
@dlorenc there’s much more required to support —target. Local builder has two modes, support needs to be added to gcb and kaniko too, and the computation of the build context should also be updated.

dlorenc:
Yeah sorry, this needs a WIP tag. I got sidetracked checking to see if kaniko supports the target flag yet.

Can you explain the build context part a little more though? Is that an optimization because some of the build context might not be needed based on the target?

dlorenc:
@dgageot can you explain what you mean by the local builder having two modes? Are you referring to the docker API vs. docker binary?

dlorenc:
ping @dgageot I think I actually did this correctly by accident.

There will be a small change in behavior, but I'm not sure what's correct. Previously if someone specified --cache-from with Kaniko, it would be ignored. Now it'll be an error.

pkg/skaffold/docker/image_test.go
```diff
@@ -171,18 +171,64 @@ func TestDigest(t *testing.T) {
...
+		a *v1alpha2.DockerArtifact
+	}
+	tests := []struct {
+		name string
+		args args
+		want []string
+	}{
+		{
+			name: "build args",
+			args: args{
```
dgageot: WDYT of inlining args? Since it contains a single arguments, I think it makes sense.
Also, most tests use `description` instead of `name` to describe the test.

dgageot:
@dlorenc you are right that you had it right the first time. I read your PR too quickly. Sorry

dlorenc:
> @dlorenc you are right that you had it right the first time. I read your PR too quickly. Sorry

It was definitely by accident :) I didn't realize the `GetBuildArgs` function was used everywhere, luckily it works.

dgageot:
@dlorenc linter is failing

PR #898, opened 2018-08-15
Add a simple test for Watcher

dgageot:

### DIFF
--- a/pkg/skaffold/watch/watch_test.go
+++ b/pkg/skaffold/watch/watch_test.go
@@ -1,0 +1,118 @@
+/*
+Copyright 2018 The Skaffold Authors
+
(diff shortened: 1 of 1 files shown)


pkg/skaffold/watch/watch_test.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/GoogleContainerTools/skaffold/testutil"
+)
+
+func TestRun(t *testing.T) {
```
r2d4: TestWatch?

dgageot: Should be good now! thanks!

PR #900, opened 2018-08-16
Remove duplication in kustomize deployer

dgageot:


balopat:
this looks good to me - is this in preparation for swapping out the image replacement logic with kustomize's logic?

dgageot:
@balopat In my mind, it was just to remove some duplication but I'd be happy to do what you said. Can you explain me what you have in mind?

balopat:
I'm not 100% sure that we can use this, but in [kustomize 1.0.5](**https://github.com/kubernetes-sigs/kustomize/releases/tag/v1.0.5**) they introduced an `ImageTagTransformer` - would be interesting to check it out.

PR #901, opened 2018-08-16
Add DEVELOPMENT.md

bobcatfish:
This DEVELOPMENT.md will hopefully help folks ramp up and get started
contributing to skaffold by describing the development process.

After discussions with @r2d4 and @balopat we want to try out keeping the
contributor/developer docs in markdown, while the user facing docs will
continue to be in asciidoc. One of the reasons for this is that github
will automatically link to the CONTRIBUTING.md and code of conduct, and
so if we wrote this in ascii doc, we would _have_ to make the docs (and
any references, links, etc.) that they contain work for both markdown
view in github and ascii doctor view in the generated docs.

Finally made this PR!!!! Sorry for the delay @r2d4 @balopat :sweat_smile:


DEVELOPMENT.md
```diff
@@ -0,0 +1,136 @@
...
+## Building skaffold docs
+
+Before [creating a PR](#creating-a-pr) with doc changes, we recommend that you locally verify the
+generated docs with:
+
+```shell
+make docs
+```
+
+And then open the generated docs/generated folder for `index.html` and `index.pdf`.
```
bobcatfish: One part of this process I'm still not clear on: now that I've made doc changes, do I need to do anything special to publish those?

balopat: no, docs automatically get published after changes are merged to master in the latest builds bucket on https://storage.googleapis.com/skaffold/builds/latest/docs/index.html. When we do a release it gets released into the given versions bucket: https://storage.googleapis.com/skaffold/releases/latest/docs/index.html

Currently we don't host these on anywhere else than GCS.

DEVELOPMENT.md
```diff
@@ -120,6 +120,9 @@ make docs
 
 And then open the generated docs/generated folder for `index.html` and `index.pdf`.
 
+Once PRs with doc changes are merged, the docs will get automatically published to
+[the public docs](https://storage.googleapis.com/skaffold/builds/latest/docs/index.html).
```
balopat: do we want to point out that there are separate builds and releases?

PR #914, opened 2018-08-21
Add --tail flag to stream logs with skaffold run

### LINKED ISSUE #253: Can we see the logs of a deployment when not in dev mode
Adding a `skaffold logs` command would do the trick

priyawadhwa:
Shoulf fix #253. Adds option to stream logs from objects deployed by
skaffold run.

### DIFF
--- a/cmd/skaffold/app/cmd/run.go
+++ b/cmd/skaffold/app/cmd/run.go
@@ -35,6 +35,7 @@ func NewCmdRun(out io.Writer) *cobra.Command {
 		},
 	}
 	AddRunDevFlags(cmd)
+	AddRunFlags(cmd)
 
 	cmd.Flags().StringVarP(&opts.CustomTag, "tag", "t", "", "The optional custom tag to use for images which overrides the current Tagger configuration")
 	return cmd
--- a/cmd/skaffold/app/cmd/run.go
+++ b/cmd/skaffold/app/cmd/run.go
@@ -35,7 +35,7 @@ func NewCmdRun(out io.Writer) *cobra.Command {
 		},
 	}
 	AddRunDevFlags(cmd)
-	AddRunFlags(cmd)
+	AddRunDeployFlags(cmd)
 
 	cmd.Flags().StringVarP(&opts.CustomTag, "tag", "t", "", "The optional custom tag to use for images which overrides the current Tagger configuration")
 	return cmd
--- a/pkg/skaffold/runner/runner.go
+++ b/pkg/skaffold/runner/runner.go
@@ -191,6 +191,8 @@ func (r *SkaffoldRunner) Run(ctx context.Context, out io.Writer, artifacts []*v1
 		select {
 		case <-ctx.Done():
 			return nil
+		default:
+			continue
 		}
 	}
 }
--- a/pkg/skaffold/runner/runner.go
+++ b/pkg/skaffold/runner/runner.go
@@ -177,24 +177,17 @@ func (r *SkaffoldRunner) Run(ctx context.Context, out io.Writer, artifacts []*v1
 	}
 	// If tail is true, stream logs from deployed objects
 	imageList := kubernetes.NewImageList()
-	for _, build := range bRes {
-		imageList.Add(build.Tag)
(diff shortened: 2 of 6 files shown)


pkg/skaffold/runner/runner.go
```diff
@@ -172,7 +172,29 @@ func (r *SkaffoldRunner) Run(ctx context.Context, out io.Writer, artifacts []*v1
...
+		imageList.Add(build.Tag)
+	}
+	colorPicker := kubernetes.NewColorPicker(artifacts)
+	logger := kubernetes.NewLogAggregator(out, imageList, colorPicker)
+
+	if err := logger.Start(ctx); err != nil {
+		return errors.Wrap(err, "starting logger")
+	}
+
+	for {
```
dgageot: Could this be replaced by just `<-ctx.Done()`?

cmd/skaffold/app/cmd/run.go
```diff
@@ -35,6 +35,7 @@ func NewCmdRun(out io.Writer) *cobra.Command {
 		},
 	}
 	AddRunDevFlags(cmd)
+	AddRunFlags(cmd)
```
dgageot: Do you think also adding the flag to `skaffold deploy` would make sense?

pkg/skaffold/runner/runner.go
dgageot: Also, should we setup the signal handling to handle ctrl-c properly?

cmd/skaffold/app/cmd/run.go
priyawadhwa: Yah I think that makes sense, I can add it there as well

pkg/skaffold/runner/runner.go
priyawadhwa: Oh yah good point, I'll replace it with `<-ctx.Done()`

For handling ctrl-c, I wasn't sure what the benefit of setting it up would be since there isn't any cleanup that needs to happen, the log stream just needs to end.

dgageot: Yeah, same for me. I think it's cleaner but I have no idea if it's really better

dgageot: Maybe we can add the ctrl-c handler to every command in another PR

priyawadhwa: Sounds good!

pkg/skaffold/runner/runner.go
```diff
@@ -172,6 +172,21 @@ func (r *SkaffoldRunner) Run(ctx context.Context, out io.Writer, artifacts []*v1
 		return errors.Wrap(err, "deploy step")
 	}
 
+	if !r.opts.Tail {
```
dgageot: @priyawadhwa do you think the duplication could be removed?

r2d4:
Looks like kokoro failed downloading gcloud

```
ERROR: (gcloud.components.install) The component [beta] failed to download.
<urlopen error timed out>
[ID: 2171690] Build finished after 73 secs, exit value: 1
Warning: Permanently added 'localhost' (ECDSA) to the list of known hosts.
[11:52:34] Collecting build artifacts from build VMBuild script failed with exit code: 1
```

rebuilding

r2d4:
https://source.cloud.google.com/results/invocations/9f34f383-9141-4b0d-8679-c9c9feb1a2b2/targets/runtimes-common%2Fskaffold%2Fpresubmit/log

pkg/skaffold/runner/runner.go
priyawadhwa: I thought about moving it into a separate function, but I think four things would have to be passed in, so it would look like:

```golang
func StreamLogsFromDeployedObjects(ctx context.Context, out io.Writer, builds []build.Artifact, artifacts []*v1alpha2.Artifact)
```
and I personally didn't think that was much cleaner, but I don't have a strong opinion so can definitely move it into a function if you prefer that!

dgageot: Let's merge this one any think about how to remove the duplication afterwards!

PR #925, opened 2018-08-23
Allow `skaffold dev —watch image`

dgageot:
The idea is to be able to run a dev loop that is focused on a few artifacts instead of all the artifacts. For eg, with the micro-services example, I could run `skaffold dev --watch-image app` to only watch the code for the `leeroy-app` part.

### DIFF
--- a/cmd/skaffold/app/cmd/cmd.go
+++ b/cmd/skaffold/app/cmd/cmd.go
@@ -129,6 +129,7 @@ func setFlagsFromEnvVariables(commands []*cobra.Command) {
 
 func AddDevFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&opts.Cleanup, "cleanup", true, "Delete deployments after dev mode is interrupted")
+	cmd.Flags().StringArrayVarP(&opts.Watch, "watch-image", "w", nil, "Choose which artifacts to watch. Artifacts with image names that contain the expression will be watched only. Default is to watch sources for all artifacts.")
 }
 
 func AddRunDeployFlags(cmd *cobra.Command) {
--- a/pkg/skaffold/runner/runner.go
+++ b/pkg/skaffold/runner/runner.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/bazel"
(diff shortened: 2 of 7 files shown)


ahmetb:
I’d recommend calling this —watch-image/-w to be verbose. Almost everyone has to read the help description to figure out how to use it otherwise.

dgageot:
Thanks @ahmetb. I think I'll also add support for regexps

ahmetb:
What’s the use case do you have in mind with regexps? I suspect directories might work better than image names if multiple images (that can be selected via a regexp) are involved.

dgageot:
I don't really have one in mind :-) In fact I tend to dislike regexps. You just convinced me to not bother and see if there's a need for it.

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -128,6 +128,7 @@ func setFlagsFromEnvVariables(commands []*cobra.Command) {
 
 func AddDevFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&opts.Cleanup, "cleanup", true, "Delete deployments after dev mode is interrupted")
+	cmd.Flags().StringArrayVarP(&opts.Watch, "watch-image", "w", nil, "Choose which artifacts to watch. Matches against image names. Default is to watch sources for all artifacts.")
```
balopat: "Choose which artifacts to watch.  Artifacts with image names that contain the expression will be watched only. Default is to watch sources for all artifacts."

pkg/skaffold/runner/runner.go
```diff
@@ -271,6 +276,20 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*v1
 	return nil, watcher.Run(ctx, PollInterval, onChange)
 }
 
+func (r *SkaffoldRunner) shouldWatch(artifact *v1alpha2.Artifact) bool {
+	if len(r.opts.Watch) == 0 {
+		return true
+	}
+
+	for _, name := range r.opts.Watch {
```
balopat: nit: `name` -> `watchExpression`

ahmetb:
> I don't really have one in mind :-) In fact I tend to dislike regexps. You just convinced me to not bother and see if there's a need for it.

Yeah I think there are two ways to go here. Either accept --watch-dir[] or --watch-image[]. I think --watch-image is good and I would not mind listing multiple images at once.

Something like "interactive skaffold" came up in another issue (basically ability to stop building/watching certain images on the fly without turning skaffold off) but that's off-topic probably for now.

dgageot:
ping @balopat

PR #928, opened 2018-08-23
Don't unmute logs if an error happened

### LINKED ISSUE #923: Compile errors hard to find in output
skaffold dev outputs logs from Pods in the cluster so when updating a service and getting compile errors the errors flow up in the terminal and can be hard to find.

🚀 Moonshot: It would be really cool if you could enter commands into skaffold while it was running to turn off Pod logging dynamically

dgageot:
I'm not fully happy with the code because it introduces some duplication
But that could be dealt with in another PR

Fix #923 

### DIFF
--- a/pkg/skaffold/runner/runner.go
+++ b/pkg/skaffold/runner/runner.go
@@ -220,6 +220,7 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*v1
 
 		switch {
 		case changed.needsReload:
+			logger.Stop()
 			return ErrorConfigurationChanged
 		case len(changed.dirtyArtifacts) > 0:
 			bRes, err := r.Build(ctx, out, r.Tagger, changed.dirtyArtifacts)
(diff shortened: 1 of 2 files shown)


pkg/skaffold/runner/runner.go
```diff
@@ -207,23 +207,41 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*v1
 	changed := changes{}
 	onChange := func() error {
 		logger.Mute()
+		unmute := true
```
balopat: I would rename this to `hasError`, I think that would be more readable, what do you think?

balopat: ping @dgageot ?

dgageot: Yes, `hasError` is much better! I though I pushed this fix but I must have lost it.

dgageot:
Rebasing and making sure the kokoro tests pass

PR #929, opened 2018-08-24
Make sure that tags created using envTemplates are "more" valid.

agcooke:
Added support to tag an image with branch name (or any other non tag complient environment variable) using envTemplate.

We will use it in our build street as follow:
`BRANCH=feature/GID-001 GIT_COMMIT=addda BUILD_NUMBER=2 skaffold build`

The skaffold file looks like this:

```
apiVersion: skaffold/v1alpha2
kind: Config
build:
  tagPolicy:
    envTemplate:
      template: "{{.IMAGE_NAME}}:{{.BRANCH}}-{{.BUILD_NUMBER}}-{{.GIT_COMMIT}}"
     ...
```
We then get nicely formatted tags which we can easily find the CI build number, git and branch information.


agcooke:
Verified the CLA agreement.

agcooke:
I do not know why googlebot is complaining, the email address matches my address.

agcooke:
I signed it!

balopat:
@agcooke  there are quotes in your email address in your git log, that might be the reason for the mismatch with the CLA!

Can you try to modify your commits with your email without quotes? 

```
commit c55561f331f61b34f0ad3164a2594b87b48fe3fc (HEAD -> 929_master, test)
Author: Adrian Cooke <¨adriangerardcooke@gmail.com¨>
Date:   Wed Sep 5 10:18:54 2018 +0200

    Fixed linter issue.

commit 73eed93ae770742c1b9e3897728b75f46ec496f7
Merge: 3b1d9c77 1354cb4c
Author: Adrian Cooke <¨adriangerardcooke@gmail.com¨>
Date:   Wed Sep 5 09:42:09 2018 +0200

```

maddisondavid:
I agree it may be nice to have skaffold sanitize the tags, but this could also be achieved before calling skaffold, i.e. 

```
#! /bin/bash
branch="feature/test@1234%123"
echo "${branch//[\/@%]/-}"
```
Results in:
```
feature-test-1234-123
```

agcooke:
@maddisondavid Yes that is true. We do that in our pipeline now, but is it something that we want to add to skaffold so that we do not have to do it with scripts?

pkg/skaffold/build/tag/env_template.go
```diff
@@ -50,7 +55,25 @@ func (t *envTemplateTagger) Labels() map[string]string {
 // GenerateFullyQualifiedImageName tags an image with the custom tag
 func (t *envTemplateTagger) GenerateFullyQualifiedImageName(workingDir string, opts Options) (string, error) {
 	customMap := CreateEnvVarMap(opts.ImageName, opts.Digest)
-	return util.ExecuteEnvTemplate(t.Template, customMap)
+	fqin, err := util.ExecuteEnvTemplate(t.Template, customMap)
+	if err != nil {
+		return "", err
+	}
+	return normalizeFQIN(fqin), nil
```
tejal29: I feel skaffold should not normalize the tag name. It should check if a tag is valid and then error out and maybe early on then at the final stage.
Not sure, everyone will agree that `skaffold`  changes all the invalid  image tags pattern under the hood to "-".
We do not want to enforce or suggest any tag naming conventions.

It would be great to fail early so users can change their env values to right one.
WDYT ?

dgageot:
I'm going to close this one because the logic behind `EnvTemplate` has changed a lot since then and because we don't see people complain about current implementation.
Please reopen if you think there's still an issue!

PR #932, opened 2018-08-26
Exclude helm dependency chart packages from watched files

### LINKED ISSUE #931: Skaffold dev creates deployment loop when helm chart has dependencies
### Expected behavior
After running `skaffold dev` and the initial build and deployment is done, it will only trigger a new build / deploy when I make a change to the artifact directory or helm chart directory.

### Actual behavior
After the initial build, the chart is continuously re-deployed in a loop. 
This is because as part of the deployment process, `helm dep build` is run which re-creates the files in the `charts` subdirectory if the chart has dependencies. This change is caught by the change watcher and causes a redeploy, and the process repeats itself.

### Information

- Skaffold version: v0.12
- Operating system: OSX 10.13.6
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1alpha2
kind: Config
build:
  artifacts:
  - imageName: internal-registry/some-app
deploy:
  helm:
    releases:
    - name: some-app
      chartPath: /charts/some-app
      valuesFilePath: /charts/some-app/overrides.yaml
      namespace: default
      imageStrategy:
        helm: {}
      values:
        "some-app.image": internal-registry/some-app
```

1. Create a chart that has a dependency
2. Create a skaffold config that deploys that chart
3. skaffold dev

Uvedale:
Fixes #931 

This excludes files located in the `charts` subdirectory within a chart package from the list of watched files. This does mean that if someone manually does a `helm dep update`, the watcher won't pick up the change. However, a dependency version bump would require a change to `requirements.yaml`, which the watcher will detect.

### DIFF
--- a/pkg/skaffold/deploy/helm.go
+++ b/pkg/skaffold/deploy/helm.go
@@ -79,8 +79,9 @@ func (h *HelmDeployer) Dependencies() ([]string, error) {
 	var deps []string
 	for _, release := range h.Releases {
 		deps = append(deps, release.ValuesFilePath)
+		chart_deps_dir := filepath.Join(release.ChartPath, "charts")
 		filepath.Walk(release.ChartPath, func(path string, info os.FileInfo, err error) error {
-			if !info.IsDir() {
+			if !info.IsDir() && !strings.HasPrefix(path, chart_deps_dir) {
 				deps = append(deps, path)
 			}
 			return nil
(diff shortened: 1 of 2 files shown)


dgageot:
Thanks @Uvedale! could you add one or a few unit tests? That would be awesome!

Uvedale:
Ok, I'll see what I can do.

dgageot:
@Uvedale If you don't have time to write the test, I'm happy to do it.

Uvedale:
@dgageot I'll add the test(s) tonight. The last few days have been a bit hectic.

PR #945, opened 2018-08-30
Port forward pods automatically during `skaffold dev`

r2d4:
Skaffold dev will now automatically port-forward any pods that expose
container ports.

Ports are assigned on a first-come-first-serve basis,
and that container will hold the lock on that port for the remainder of the dev
cycle.

Updates to a container will trigger an update which replaces the
forwarded port with the newest version.

Scaling up a deployment will continue to port-forward the latest pod.
However, scaling down a deployment has the potential to kill the
port-forwarded pod without replacement - since we do not keep track of
candidates that are "waiting" on that port.

### DIFF
--- a/pkg/skaffold/kubernetes/port_forward.go
+++ b/pkg/skaffold/kubernetes/port_forward.go
@@ -78,7 +78,7 @@ func (*kubectlForwader) Forward(pfe *portForwardEntry) error {
 	cmd.Stderr = buf
 
 	if err := cmd.Run(); err != nil && !IsTerminatedError(err) {
-		return errors.Wrapf(err, "port forwarding pod: %s, port: %s, err: %s", pfe.podName, portNumber, string(buf.Bytes()))
+		return errors.Wrapf(err, "port forwarding pod: %s, port: %s, err: %s", pfe.podName, portNumber, buf.String())
 	}
 	return nil
 }
@@ -149,7 +149,7 @@ func (p *PortForwarder) Start(ctx context.Context) error {
 				}
 				if p.podSelector.Select(pod) && pod.Status.Phase == v1.PodRunning && pod.DeletionTimestamp == nil {
 					go func() {
-						if err := p.portForwardPod(ctx, pod); err != nil {
+						if err := p.portForwardPod(pod); err != nil {
 							logrus.Warnf("port forwarding pod failed: %s", err)
 						}
 					}()
@@ -161,7 +161,7 @@ func (p *PortForwarder) Start(ctx context.Context) error {
 	return nil
 }
 
-func (p *PortForwarder) portForwardPod(ctx context.Context, pod *v1.Pod) error {
+func (p *PortForwarder) portForwardPod(pod *v1.Pod) error {
 	resourceVersion, err := strconv.Atoi(pod.ResourceVersion)
 	if err != nil {
 		return errors.Wrap(err, "converting resource version to integer")
--- a/pkg/skaffold/kubernetes/port_forward.go
+++ b/pkg/skaffold/kubernetes/port_forward.go
@@ -17,6 +17,7 @@ limitations under the License.
 package kubernetes
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -26,7 +27,6 @@ import (
 	"syscall"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
-	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


r2d4:
Fixes #94

pkg/skaffold/kubernetes/port_forward.go
```diff
@@ -0,0 +1,228 @@
...
+	if !ok {
+		return false
+	}
+	ws := exitError.Sys().(syscall.WaitStatus)
+	return ws.Signal() == syscall.SIGTERM
+}
+
+// Key is an identifer for the lock on a port during the skaffold dev cycle.
+func (p *portForwardEntry) key() string {
+	return fmt.Sprintf("%s-%d", p.containerName, p.port)
```
dlorenc: Should we also include namespace and podName here to make it fully keyed?

pkg/skaffold/kubernetes/port_forward.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+// PortForwarder is responsible for selecting pods satisfying a certain condition and port-forwarding the exposed
+// container ports within those pods. It also tracks and manages the port-forward connections.
+type PortForwarder struct {
```
balopat: nit: I would rename this to `ExclusivePortForwarder` (or similar) and rename the `Forwarder` interface to `PortForwarder`.

pkg/skaffold/kubernetes/port_forward.go
```diff
@@ -0,0 +1,228 @@
...
+				continue
+			}
+
+			entry := &portForwardEntry{
+				resourceVersion: resourceVersion,
+				podName:         pod.Name,
+				containerName:   c.Name,
+				port:            port.ContainerPort,
+			}
+			v, ok := p.forwardedPods.Load(entry.key())
```
balopat: nit: maybe `ok` -> `isPodForwarded`

pkg/skaffold/kubernetes/port_forward.go
```diff
@@ -0,0 +1,228 @@
...
+				// so only watch "modified" events
+				if evt.Type != watch.Modified {
+					continue
+				}
+
+				pod, ok := evt.Object.(*v1.Pod)
+				if !ok {
+					continue
+				}
+				if p.podSelector.Select(pod) && pod.Status.Phase == v1.PodRunning && pod.DeletionTimestamp == nil {
```
balopat: can we extract the event handling logic and have it covered by unit test? I feel like it's an important piece to protect especially if we want to merge it later with the logging podwatcher.

balopat:
tested it it on @ahmetb's microservices example, and it's running nicely 👍 

Maybe we could add a doc to list the ports as well, it was a bit hard to find what service was mapped where before I found this - but this works only for unix:  
````
$ ps ax | grep port-forw
10369 s006  S+     0:00.19 kubectl port-forward pod/productcatalogservice-7bdb6dc844-cm9qq 3550 3550
10407 s006  S+     0:00.19 kubectl port-forward pod/paymentservice-5977d9d586-4lt4d 50051 50051
10838 s006  S+     0:00.18 kubectl port-forward pod/cartservice-567b597d5-t4dc6 7070 7070
11486 s006  S+     0:00.20 kubectl port-forward pod/checkoutservice-856cf9dd46-bs9hb 5050 5050
11511 s006  S+     0:00.20 kubectl port-forward pod/currencyservice-76cbd59d48-sh5fm 7000 7000
20538 s006  S+     0:00.16 kubectl port-forward pod/frontend-86f9dc4cc-smfzs 8080 8080
```

ahmetb:
What does it do in case of port conflicts? We deliberately kept the port numbers different so we can develop locally easier –but we actually have plans to standardize port numbers.

viglesiasce:
I just tested this out and it is failing when the pods expose a privileged port (ie 80). We need to be able to detect that and map to a higher port maybe by adding a constant (80->10080).

balopat:
In this PR the first one wins. See https://github.com/GoogleContainerTools/skaffold/pull/945/files#diff-f0665c32a151485f611d6517eb543409R175 for conflict "resolution" with known ports. 
The same question comes up with already bound ports on localhost. 
One "mode" would be automatically find an unbound port (+1, random, etc.): pros: no config, cons: user has to "discover" the assigned ports, they are instable. 
Another "mode" could be to enable the user in Skaffold yaml to specify the ports. pros: user knows the ports, they are stable, cons: user has to configure stuff.

ahmetb:
I think if you print something on the screen in the format:

    export A_ADDR=localhost:28937
    export B_ADDR=localhost:12788

then you can randomly choose from available ports, and I'll just paste this to my terminal where I'm launching my app from.

balopat:
Added https://github.com/GoogleContainerTools/skaffold/issues/969 for follow-up on the issues for more complex cases. We should document the caveats and then merge.

Issue #289, opened 2018-03-27
Allow support for helm to recreate pods


PacoVK:
Developing eg NodeJs using helm mechanism, skaffold detects changes and rebuild but helm doesn't recreate pods. With additional flag to recreate pods this would be possible. 
Same for other flags eg dep-up.

Suggestion:

deploy:
  helm:
    releases:
    - name: hello-world
      chartPath: charts/hello-world
      namespace: skaffold
      args:
          - recreate-pods
          - dep-up

PR #946, opened 2018-08-31
Pass --recreate-pods to helm by default in dev mode

stevenjm:
It is common in development to just rebuild a Docker image with the same tag, which will not trigger a rolling update of the deployment because the pod spec does not change. Passing `--recreate-pods` will forcibly terminate the running pods and allow new ones to be spawned with the new image.

This change only applies in dev mode because it would be harmful for production deployments, which should push an image with a new tag and allow the usual rolling update process to replace the old pods.

This change would resolve https://github.com/GoogleContainerTools/skaffold/issues/289, without requiring an additional configuration option. (This behaviour can be disabled at runtime with `--recreate-pods=false` if needed for some reason.)

For now, I have only added a test to verify that a helm deployment works if `recreatePods == true`. Developing a test for the behaviour of actually recreating pods would be a little more involved, and I'm not sure it's worth including in this PR.

### DIFF
--- a/examples/annotated-skaffold.yaml
+++ b/examples/annotated-skaffold.yaml
@@ -151,6 +151,7 @@ deploy:
     #     image: skaffold-helm
     #   namespace: skaffold
     #   version: ""
+    #   recreatePods: false
     #
     #   # setValues get appended to the helm deploy with --set.
     #   setValues:
--- a/examples/helm-deployment/skaffold.yaml
+++ b/examples/helm-deployment/skaffold.yaml
@@ -14,6 +14,8 @@ deploy:
       #valuesFilePath: helm-skaffold-values.yaml
       values:
         image: gcr.io/k8s-skaffold/skaffold-helm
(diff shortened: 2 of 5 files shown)


stevenjm:
@dgageot Sure, that makes sense, and should be a straightforward change. I'll hopefully get some time tomorrow to implement it.

stevenjm:
Whoops, since there's now a new option in skaffold.yaml I should also update the relevant examples/documentation. Will push another version of this change a bit later.

stevenjm:
@dgageot I've made the change you requested. I haven't made the default `true` because this should never be on for production, so it's safer to explicitly turn it on for development.

In an ideal world, I'd still like to have it on by default for `skaffold dev` and not `skaffold run`, but as far as I can tell there is no way to determine whether the invoked command is `dev` or `run` from within the helm deployment implementation.

dgageot:
@stevenjm Thanks! Maybe you could create a ticket to later have a default value based on the command (dev or run)

PR #982, opened 2018-09-13
Add v1alpha3 Config

priyawadhwa:
I added a v1alpha3 version of the skaffold config, since adding
different buildcontexts to the kaniko builder requires a breaking config
change.

Now, new buildcontexts can be easily added to the KanikoBuildContext type.

I also added a transform function so that 'skaffold fix' can convert
v1alpha2 configs to v1alpha3, and fixed unit tests to pass with the new
config.

This is part 1 of 3 PRs to incorporate more kaniko buildcontexts into
the kaniko builder.

Sorry this PR is so huge (it's mostly just changing v1alpha2 to v1alpha3 across the code), so thank you in advance to whoever looks at it!!


integration/run_test.go
```diff
@@ -221,22 +221,39 @@ func setupNamespace(t *testing.T) (*v1.Namespace, func()) {
...
+	tests := []struct {
+		name      string
+		directory string
+	}{
+		{
+			name:      "test v1alpha1 to v1alpha2 fix",
+			directory: "testdata/v1alpha1",
+		},
+		{
+			name:      "test v1alpha2 to v1alpha3 kaniko fix",
```
balopat: nit: we won't need kaniko here I think

PR #989, opened 2018-09-17
Auto configure authentication helper for gcr.io

### LINKED ISSUE #966: Auto configure docker auth to push to gcr.io
When Skaffold is pushing an image to gcr.io, it would be convenient if it could configure automatically the credential helper.

Otherwise, a user has to follow the guide [here](https://cloud.google.com/container-registry/docs/advanced-authentication). All it does is add those lines to `~/.docker/config.json`:

```
  "credHelpers": {
    "gcr.io": "gcloud",
    "us.gcr.io": "gcloud",
    "eu.gcr.io": "gcloud",
    "asia.gcr.io": "gcloud",
    "staging-k8s.gcr.io": "gcloud",
    "marketplace.gcr.io": "gcloud"
  }
```

We could add only the line that is required to push the image we want to push. And Skaffold doesn't even have to edit the docker config file, it can all be done with code.

dgageot:
Fixes #966

### DIFF
--- a/pkg/skaffold/gcp/auth.go
+++ b/pkg/skaffold/gcp/auth.go
@@ -29,14 +29,17 @@ import (
 // This doesn't modify the ~/.docker/config.json. It's only in-memory
 func AutoConfigureGCRCredentialHelper(cf *configfile.ConfigFile, registry string) {
 	if registry != "gcr.io" && !strings.HasSuffix(registry, ".gcr.io") {
+		logrus.Debugln("Skipping credential configuration because registry is not gcr.")
 		return
 	}
 
 	if cf.CredentialHelpers != nil && cf.CredentialHelpers[registry] != "" {
+		logrus.Debugln("Skipping credential configuration because credentials are already configured.")
 		return
 	}
 
 	if path, _ := exec.LookPath("docker-credential-gcloud"); path == "" {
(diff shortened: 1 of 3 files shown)


pkg/skaffold/gcp/auth.go
```diff
@@ -0,0 +1,41 @@
...
+		return
+	}
+
+	if cf.CredentialHelpers == nil {
+		cf.CredentialHelpers = make(map[string]string)
+	}
+
+	_, present := cf.CredentialHelpers[registry]
+	if !present {
+		cf.CredentialHelpers[registry] = "gcloud"
```
dlorenc: Do we make sure gcloud is installed anywhere?

It's also possible to use the "gcr" credential helper which handles it's own authentication without gcloud. I'm not sure how common that is, though.

dgageot: I'm going to add the check indeed.

I'd rather ignore that "gcr" credential helper can be used since It's not the solution documented [here](
https://cloud.google.com/container-registry/docs/advanced-authentication)
Users can still use it if they want though.

dgageot:
@dlorenc it should be good now.

dlorenc:
I added some debug logging on why the helper get's skipped, but now I've made the CLA bot angry :(

I confirm that I'm OK with my commit being used :)

dgageot:
I'm ok with @dlorenc's commit ;-)

Issue #922, opened 2018-08-23
"skaffold deploy" is not useful


ahmetb:
I cannot figure out the use case for "skaffold deploy".

This command is not able to populate the image tags. Without that, all it's doing is corrupting the Deployments by deploying the tag-less images.

This command cannot use the result of "skaffold build" like this (as far as I know):

    skaffold build > images.txt
    skaffold deploy < images.txt

Without these functionality this command just appears like a `kubectl apply` that is not compatible with `skaffold run` to me. 

Please consider either making "deploy" cmd work with "build" cmd or consider eliminating to avoid confusing users.

balopat:
The use case would be to use `skaffold deploy` as part of other orchestration tools. I think we should figure out the tagging situation with `skaffold deploy`.

mxey:
See also https://github.com/GoogleContainerTools/skaffold/issues/653

cliveza:
Yeah, I just picked up on this too. I thought I could `

```
RELEASE=v1.0.0 skaffold build
RELEASE=v1.0.0 skaffold deploy
```
While the build command is able to build an image with the correct tag, using an environment variable and envTemplate, the deploy command just ignore this all together.

r2d4:
I think the right fix for this is that `skaffold deploy` should also generate the tag through the git commit tagger (or env tagger).

r2d4:
Eventually we might want to get rid of the "non-stable" taggers that rely on information from the build artifact itself

rsanders:
Could `skaffold build` write the build artifact names to an output file that is also read by `skaffold deploy`?   I see that it has support for outputting those, but it leaves the plumbing as an exercise for the user.

We are using skaffold in our CD pipeline, and we like having "non-stable" components like build ID in the docker image tag.  We also prefer to have separate build -> test -> deploy stages.

r2d4:
@rsanders yes, skaffold build already outputs the go template of the struct that deploy uses. Deploy should be able to read from that generated output

I imagine something like 

```
skaffold build > build.out
# do some work
skaffold deploy --build-artifacts=build.out
```

or

```
skaffold build | skaffold deploy --build-artifacts -
```

or 
```
skaffold build > build.out
# do some work
skaffold deploy --build-artifacts - < build.out
```

balopat:
cc @tejal29

FX-HAO:
[...]

PR #995, opened 2018-09-19
Support for replacing image names in env vars

ajbouh:
Specifically env vars that end with _IMAGE

### DIFF
--- a/pkg/skaffold/deploy/kubectl/images.go
+++ b/pkg/skaffold/deploy/kubectl/images.go
@@ -17,6 +17,8 @@ limitations under the License.
 package kubectl
 
 import (
+	"strings"
+
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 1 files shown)


mattfysh:
This will be very handy for kicking off jobs needing the skaffold-tagged image names. For now, a workaround is to annotate the pod that needs the image name, using key "image". It will match and get picked up by the replacer

ajbouh:
I don't know how bad the rebase will be so I'm reluctant to invest further in this PR. Also unclear what the appetite for merging this is. Thoughts on these points?

tejal29:
I think we are thinking of Adding Env Based tagger as part of #922 for `skaffold build` images.
See https://github.com/GoogleContainerTools/skaffold/issues/922#issuecomment-474710200 for more details.

I agree, lets close this PR for now and revisit if this is needed once #922 is resolved.

ajbouh:
What is env based tagger? That sounds like a way to use environment variables to decide on the docker image tag?

If so, that's unrelated to what this PR does

tejal29:
Yes it is. Since you mentioned you don't want to invest further at this time, I suggesting revisiting this once build changes are in. 
Feel free to re-open this if you want to get this in before that.

Issue #994, opened 2018-09-19
Helm Image Incorrectly parses tag when docker image contains a port number


maddisondavid:
When an image contains a port number such as `myreg.com:8116/product/ui:test-tag` and an imageStrategy of `helm` the tag is incorrectly parsed

`skaffold run -v debug` provides the following:

```
helm --kube-context vermin install --name ui-helm --version dev charts/ui --namespace testing -f charts/ui/values.yaml --set image.repository=myreg.com,image.tag=8116/product/ui
```

### Expected behavior
It's expected that it would correct parse the tag from the image name by splitting from the colon on the right hand side.

.i.e. It should generate the following Helm command
```
helm --kube-context vermin install --name ui-helm --version dev charts/ui --namespace testing -f charts/ui/values.yaml --set image.repository=myreg.com:8116/product/ui,image.tag=test-tag
```

### Actual behavior
Image is split on the FIRST colon
``` 
helm --kube-context vermin install --name ui-helm --version dev charts/ui --namespace testing -f charts/ui/values.yaml --set image.repository=myreg.com,image.tag=8116/product/ui
```
### Information

- Skaffold version: v0.14.0
- Operating system: macOS 10.12.6
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1alpha2
kind: Config
build:
  tagPolicy:
    envTemplate:
      template: "{{.DOCKER_REGISTRY}}/{{.IMAGE_NAME}}:dev-{{.DIGEST_HEX}}"
  artifacts:
  - imageName: ui
deploy:
  helm:
    releases:
    - name: ui-helm
      chartPath: charts/ui
      valuesFilePath: charts/fui/values.yaml
      namespace: testing
      imageStrategy:
        helm: {}
      values:
        image: ui
      version: "dev"
      recreatePods: false
```

### Steps to reproduce the behavior

1. Use a docker image that points to a private docker registry with a port number
2. skaffold run -v debug

balopat:
Agreed, that is a bug, tag should be the right hand side of the _last_ `:` character.

balopat:
@maddisondavid do you want to take a stab at fixing it? :)

PR #1001, opened 2018-09-20
Correctly parse build tags that contain port numbers

maddisondavid:
Fixes issue #994

When a build tag contains a port number i.e. docker.io:5000/skaffold-helm:my-tag the Helm Strategy was incorrectly performing a basic split and ending up with repository=docker.io,tag=5000/skaffold-helm:my-tag.

This PR fixes the parsing so that the repository is correctly extracted, i.e. repository=docker.io:5000/skaffold-helm,image=my-tag

### DIFF
--- a/pkg/skaffold/deploy/helm.go
+++ b/pkg/skaffold/deploy/helm.go
@@ -135,13 +135,14 @@ func (h *HelmDeployer) deployRelease(ctx context.Context, out io.Writer, r v1alp
 	for k, v := range params {
 		setOpts = append(setOpts, "--set")
 		if r.ImageStrategy.HelmImageConfig.HelmConventionConfig != nil {
-			tagSplit := strings.Split(v.Tag, ":")
-			imageRepositoryTag := fmt.Sprintf("%s.repository=%s,%s.tag=%s", k, tagSplit[0], k, tagSplit[1])
+			tagIndex := strings.LastIndex(v.Tag, ":")
+			imageRepositoryTag := fmt.Sprintf("%s.repository=%s,%s.tag=%s", k, v.Tag[0:tagIndex], k, v.Tag[tagIndex+1:len(v.Tag)-1])
 			setOpts = append(setOpts, imageRepositoryTag)
 		} else {
 			setOpts = append(setOpts, fmt.Sprintf("%s=%s", k, v.Tag))
 		}
 	}
+
 
 	// First build dependencies.
 	logrus.Infof("Building helm dependencies...")
--- a/pkg/skaffold/deploy/helm_test.go
+++ b/pkg/skaffold/deploy/helm_test.go
@@ -37,7 +37,7 @@ import (
 var testBuilds = []build.Artifact{
 	{
 		ImageName: "skaffold-helm",
-		Tag:       "skaffold-helm:3605e7bc17cf46e53f4d81c4cbc24e5b4c495184",
+		Tag:       "docker.io:5000/skaffold-helm:3605e7bc17cf46e53f4d81c4cbc24e5b4c495184",
 	},
 }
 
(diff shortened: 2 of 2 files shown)


pkg/skaffold/deploy/helm.go
```diff
@@ -135,8 +135,8 @@ func (h *HelmDeployer) deployRelease(ctx context.Context, out io.Writer, r v1alp
 	for k, v := range params {
 		setOpts = append(setOpts, "--set")
 		if r.ImageStrategy.HelmImageConfig.HelmConventionConfig != nil {
-			tagSplit := strings.Split(v.Tag, ":")
-			imageRepositoryTag := fmt.Sprintf("%s.repository=%s,%s.tag=%s", k, tagSplit[0], k, tagSplit[1])
+			tagIndex := strings.LastIndex(v.Tag, ":")
```
dgageot: Would it make sense to parse the image tag with something like `docker.ParseReference()`? If it works, it would handle all the edge cases better than any string parsing

maddisondavid: Sounds good, I wasn't aware of `docker.ParseReference()`, it's annoying that it only gives you the baseName and not also the tag.

Code updated to use `docker.ParseReference()` which leads to much nicer code.

I followed the style of the `helm_test.go` to fix up the test, but I'm not super happy about it because the test uses the SAME logic as the code it's testing.

maddisondavid: Doh, I realise that `docker.ParseReference()` is in a `skaffold` package, so probably makes more sense for me to update `docker.ParseReference()` to include the Tag.

maddisondavid:
The Travis build is failing but only on the Linux build with some weird missing packages.  I'm leaning towards it just being an environmental issue blip during the build but I don't have permission to re-run the build.

dgageot:
Let me restart the build

pkg/skaffold/deploy/helm_test.go
```diff
@@ -292,8 +293,8 @@ func TestHelmDeploy(t *testing.T) {
 				t:         t,
 				getResult: fmt.Errorf("not found"),
 				installMatcher: func(cmd *exec.Cmd) bool {
-					builds := strings.Split(testBuilds[0].Tag, ":")
-					expected := map[string]bool{fmt.Sprintf("image.repository=%s,image.tag=%s", builds[0], builds[1]): true}
+					dockerRef, _ := docker.ParseReference(testBuilds[0].Tag)
```
dgageot: Could you handle the error?

pkg/skaffold/docker/reference_test.go
```diff
@@ -72,6 +79,7 @@ func TestParseReference(t *testing.T) {
 			parsed, err := ParseReference(test.image)
 
 			testutil.CheckErrorAndDeepEqual(t, false, err, test.expectedName, parsed.BaseName)
+			testutil.CheckErrorAndDeepEqual(t, false, err, test.expectedTag, parsed.Tag)
```
dgageot: Two instances of `CheckErrorAndDeepEqual` could be replaced with `CheckDeepEqual` here

pkg/skaffold/deploy/helm_test.go
maddisondavid: Done.

maddisondavid:
I've just run into an issue using MASTER whereby the last letter of the tag is missing when Helm is deployed.  Please don't merge this until I can confirm it's not a problem with the code I put in!

maddisondavid:
OK, seems like it must have been something around my environment.  Everything appears to be working fine now and deployment is getting and using the correct tag.  I just wanted to be sure it wasn't something caused by my code!

Issue #942, opened 2018-08-29
support dependencies for kustomize yamls


balopat:
As @mxey pointed out on [Slack](https://kubernetes.slack.com/archives/CABQMSZA6/p1535440771000100): 

"Skaffold does not support dependencies through Kustomize yet. I only saw this TODO comment https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/deploy/kustomize.go#L92, but there is no issue to track this. Would it make sense to add a command to Kustomize or maybe use Kustomize as a library to list the dependencies, rather than Skaffold interpreting the kustomization itself?"

r2d4:
We can do this now

https://github.com/kubernetes-sigs/kustomize/pull/265
https://github.com/kubernetes-sigs/kustomize/issues/144

PR #1015, opened 2018-09-21
Basic support for watching Kustomize dependencies

ajbouh:
This isn't a complete enumeration of Kustomize dependencies, but it's a reasonable initial step.


r2d4:
Fixes https://github.com/GoogleContainerTools/skaffold/issues/942

Thanks!

Issue #367, opened 2018-04-11
Example versioning


r2d4:
Our example are meant to work with skaffold at HEAD, but that won't always work with the latest skaffold release, as we add new fields. Whats the best way to convey this to users?

* Warning on example configs
* Point README to examples at tag of last released version
* Move examples to different repository
* Duplicate examples to "integration" folder, only update on release

ref #363

PR #1020, opened 2018-09-25
Better API version parsing.

dlorenc:
This implements part of the blang/semver surface for k8s-style group/versions.

I couldn't find an easy-to-use version of this in the k8s codebase, but I'm sure it's somewhere.
This gives us a way to tell if the unequal version is higher or lower, so we can give a better error message.

Ref #367

### DIFF
--- a/pkg/skaffold/apiversion/apiversion.go
+++ b/pkg/skaffold/apiversion/apiversion.go
@@ -1,0 +1,119 @@
+/*
+Copyright 2018 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package apiversion
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+)
+
+type ReleaseTrack int
+
+const (
+	alpha ReleaseTrack = 0
+	beta  ReleaseTrack = 1
+	ga    ReleaseTrack = 2
+)
+
+type Version struct {
+	Major   int
+	Minor   int
+	Release ReleaseTrack
+}
+
+var re = regexp.MustCompile(`^skaffold/v(\d)(?:(alpha|beta)(\d))?$`)
+
+// ParseAPIVersion parses a string into a Version.
+func ParseVersion(v string) (*Version, error) {
+	res := re.FindStringSubmatch(v)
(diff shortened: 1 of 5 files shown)


pkg/skaffold/apiversion/apiversion.go
```diff
@@ -0,0 +1,119 @@
...
+type Version struct {
+	Major   int
+	Minor   int
+	Release ReleaseTrack
+}
+
+var re = regexp.MustCompile(`^skaffold/v(\d)(?:(alpha|beta)(\d))?$`)
+
+// ParseAPIVersion parses a string into a Version.
+func ParseVersion(v string) (*Version, error) {
```
r2d4: Here's an easier mapping to blang/semver

v1alpha3 -> 1.0.0-alpha.3
v1 -> 1.0.0

v2beta3 -> 2.0.0-beta.3
v2beta4 -> 2.0.0-beta.4

pkg/skaffold/apiversion/apiversion.go
```diff
@@ -0,0 +1,119 @@
...
+	ga    ReleaseTrack = 2
+)
+
+type Version struct {
+	Major   int
+	Minor   int
+	Release ReleaseTrack
+}
+
+var re = regexp.MustCompile(`^skaffold/v(\d)(?:(alpha|beta)(\d))?$`)
```
r2d4: I think you need to escape the first `/` here to `\/` like `^skaffold\/v(\d)(?:(alpha|beta)(\d))?$`


You already have the capture groups here, so you don't have to manually parse it again.

```golang
		m := r.FindStringSubmatch(s)
		major := m[1]
		prereleaseStr := m[2]
		prereleaseNum := m[3]
```

https://play.golang.org/p/udSiQC7eF_p

pkg/skaffold/apiversion/apiversion.go
r2d4: Once we map it to blang/semver, we can remove the comparison logic and just reuse that library's

dgageot:
@dlorenc should this be closed in favor of #1026?

pkg/skaffold/apiversion/apiversion.go
balopat: My 2 cents: 
The two solution seem pretty comparable, but I like that the blang/semver version because it seems like that the K8s versioning subset _is_ a subset of semver, hence the abstraction is correct. 
The lib will take care of version ordering after the mapping - and for the other blang/semver functionality - like range parsing - will be available, only we'd have to do similar mapping but could be useful and then we don't have to write tests for our own ordering. 

One issue for example with Dan's approach already is that "v1" is not supported. "ga" is not following the K8S versioning either. By mapping `<alpha/beta>X` to Semver's pre-release numbers, this goes away automatically. 

One problem I see with our approach is that there is no minor version numbers - but we made that call when we went with the K8S versioning.

dgageot:
@dlorenc @balopat Maybe I merged  #1026 too quickly?

balopat:
Hey, no problem, I was in favor for #1026 anyway

PR #1029, opened 2018-09-26
Add option to not tail logs in dev mode

MrTrustor:
I need to able to not tail the logs of the containers in dev mode for demos.
It gives a much cleaner and more comprehensible look.

### DIFF
--- a/cmd/skaffold/app/cmd/dev.go
+++ b/cmd/skaffold/app/cmd/dev.go
@@ -38,6 +38,7 @@ func NewCmdDev(out io.Writer) *cobra.Command {
 	}
 	AddRunDevFlags(cmd)
 	AddDevFlags(cmd)
+	AddRunDeployFlags(cmd)
 	return cmd
 }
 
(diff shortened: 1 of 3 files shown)


cmd/skaffold/app/cmd/dev.go
```diff
@@ -38,6 +38,7 @@ func NewCmdDev(out io.Writer) *cobra.Command {
 	}
 	AddRunDevFlags(cmd)
 	AddDevFlags(cmd)
+	AddRunDeployFlags(cmd)
```
dgageot: `tail` flag defaults to `false`. I think it should default to `true` in dev mode.

MrTrustor:
@dgageot Tailing is defaulting to true now, but I'm not sure how to organise the flags. Let me know if that doesn't work for you.

dgageot:
I think you could keep the same ‘tail’ variable for both dev and run modes. But the default value would be different.

MrTrustor:
From my tests, that doesn't seem to be possible: the `Tail` option can only have one default value across the whole project. When I set the default at `true` for the dev command, it doesn't do anything. My guess is that it's because the default is changed later by the definition of the other commands.

dgageot:
@MrTrustor Are you ok if I push a PR that implements that?

MrTrustor:
Sure, go ahead. Thanks!

dgageot:
Closed in favor of #1037

dgageot:
@MrTrustor Ah sorry, it needs to be rebased now... Once rebased, I'll merge it right away

MrTrustor:
Rebased :-)

PR #1039, opened 2018-09-26
File Sync for skaffold dev

r2d4:
This PR introduces file sync for skaffold dev. Each artifact config now has an optional sync map. If the changes for the dev cycle are contained entirely in the sync map, then a rebuild and redeploy will be skipped, and those files will be synced to the running container. If only some of the changes are contained in the sync map, a full rebuild and redeploy will happen as normal.

That can optionally take glob patterns in the "key" part of the map (if the "key" is a pattern, the "value" destination must be a directory).

This PR changes the watcher slightly to compute WatchEvents, which are files that are added, modified, deleted, rather than an aggregate yes/no to whether things have changed. We were already computing this information and throwing it away, so it shouldn't be any slower.

It also adds a "webserver" example

**Caveats**: Files are copied with `kubectl cp` and `kubectl exec rm`. `kubectl cp` requires the `tar` and `cp` binaries to be present in the container. `kubectl exec rm` will require `rm` to be in the container. This feature will not work for scratch containers.

Future work for another PR will directly infer the "destination" from the Dockerfile if it not present.


pkg/skaffold/kubernetes/sync.go
```diff
@@ -0,0 +1,79 @@
...
+	return exec.Command("kubectl", "cp", src, fmt.Sprintf("%s/%s:%s", pod.Namespace, pod.Name, dst), "-c", container.Name)
+}
+
+func perform(image string, files map[string]string, cmdFn func(v1.Pod, v1.Container, string, string) *exec.Cmd) error {
+	logrus.Info("Syncing files:", files)
+	client, err := Client()
+	if err != nil {
+		return errors.Wrap(err, "getting k8s client")
+	}
+	pods, err := client.CoreV1().Pods("").List(meta_v1.ListOptions{})
```
nkubala: I think we want to plumb through the `namespace` to these functions and pass it in here. As per my comment above this should work with kubectl exec

pkg/skaffold/util/util.go
```diff
@@ -106,6 +106,13 @@ func ExpandPathsGlob(workingDir string, paths []string) ([]string, error) {
 	return ret, nil
 }
 
+// HasMeta reports whether path contains any of the magic characters
+// recognized by filepath.Match.
+// This is a copy of filepath/match.go's hasMeta
```
nkubala: the original implementation also includes the filepath separator. why not just use that?

pkg/skaffold/watch/watch_test.go
```diff
@@ -108,8 +108,9 @@ func newCallback() *callback {
 	}
 }
 
-func (c *callback) call() {
+func (c *callback) call(e Events) error {
```
nkubala: what's this change for?

pkg/skaffold/kubernetes/sync.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return errors.Wrap(err, "getting pods")
+	}
+	for _, p := range pods.Items {
+		for _, c := range p.Spec.Containers {
+			if strings.HasPrefix(c.Image, image) {
+				for src, dst := range files {
+					cmd := cmdFn(p, c, src, dst)
```
nkubala: could we call these in parallel?

pkg/skaffold/runner/runner.go
```diff
@@ -269,7 +272,18 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*v1
...
-			func() { changed.Add(artifact) },
+			func(e watch.Events) error {
+				sync, err := r.shouldSync(artifact.ImageName, artifact.Sync, e)
+				if err != nil {
+					return errors.Wrap(err, "checking sync files")
+				}
+				if !sync {
+					changed.Add(artifact)
+				} else {
+					color.Default.Fprintln(out, "Synced:", "copied", append(e.Added, e.Modified...), "deleted", e.Deleted)
```
nkubala: I *think* this is the case, but just to be sure: if we get here, is it always true that we've performed a sync?

pkg/skaffold/kubernetes/sync.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (*KubectlSyncer) CopyFilesForImage(image string, syncMap map[string]string) error {
+	return perform(image, syncMap, copyFileFn)
+}
+
+func (*KubectlSyncer) DeleteFilesForImage(image string, syncMap map[string]string) error {
+	return perform(image, syncMap, deleteFileFn)
+}
+
+// TODO(r2d4): kubectl exec doesn't seem to take a namespace flag?
```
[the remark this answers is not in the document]
r2d4: Good catch, I'll update

pkg/skaffold/util/util.go
r2d4: https://golang.org/pkg/path/filepath/#Match?

pkg/skaffold/watch/watch_test.go
r2d4: callback.call() has to implement ChangeFn which is func(Events) error

pkg/skaffold/kubernetes/sync.go
r2d4: Good idea, I'll make them run in parallel

pkg/skaffold/kubernetes/sync.go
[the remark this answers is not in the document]
r2d4: Do you mind opening up an issue on kubectl to document that? That might be a good contribution to upstream kubectl.

pkg/skaffold/kubernetes/sync.go
r2d4: I don't think we actually know what namespace we're looking for? We don't actually parse the namespace anywhere, and it can be different for kubectl and helm.

dgageot:
@r2d4 when I tried, it would sync files when I don't have a `sync` on the artifacts and it would stop syncing as soon as I had one...

dgageot:
@r2d4 to make the review easier, could you create a PR that introduces watch events?

r2d4:
I’ll split out the watch changes into a separate PR

pkg/skaffold/runner/runner.go
```diff
@@ -317,6 +331,52 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*v1
...
+	if err := r.Syncer.DeleteFilesForImage(image, toDelete); err != nil {
+		return false, errors.Wrap(err, "deleting files for image")
+	}
+	if err := r.Syncer.CopyFilesForImage(image, toCopy); err != nil {
+		return false, errors.Wrap(err, "copying files for image")
+	}
+	return true, nil
+}
+
+func intersect(syncMap map[string]string, files []string) (map[string]string, error) {
```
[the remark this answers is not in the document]
r2d4: Yeah I struggled with the right logic here. Maybe I can rewrite match to return the full match string. Not having the actual filesystem here makes users need to be explicit about dir vs. file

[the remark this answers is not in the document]
r2d4: I rewrote it to fix those issues and added another suite of tests to cover it. I still might rewrite both these functions after I split out watch, they're doing a little too much.

ajbouh:
@r2d4 it's possible that you could use similar logic to keep the build context up to date to make container builds on knative really fast

r2d4:
Yes @ajbouh we can use the same information from the watcher to only update the file deltas on remote contexts 👍

dgageot:
I added a `wip` label to make sure we merge #1064 first and rebase this one on top of it

dgageot:
ping @r2d4, #1064 is merged.

pkg/skaffold/kubernetes/sync.go
```diff
@@ -0,0 +1,93 @@
...
+	return strings.Join(reqs, ",")
+}
+
+func perform(image string, files map[string]string, cmdFn func(v1.Pod, v1.Container, string, string) *exec.Cmd) error {
+	logrus.Info("Syncing files:", files)
+	client, err := Client()
+	if err != nil {
+		return errors.Wrap(err, "getting k8s client")
+	}
+	pods, err := client.CoreV1().Pods("").List(meta_v1.ListOptions{
```
dgageot: Should you provide the ns here?

dgageot:
Sorry, @r2d4 you need to rebase once again...

pkg/skaffold/kubernetes/sync.go
```diff
@@ -0,0 +1,93 @@
...
+				var e errgroup.Group
+				for src, dst := range files {
+					src, dst := src, dst
+					e.Go(func() error {
+						cmd := cmdFn(p, c, src, dst)
+						return util.RunCmd(cmd)
+					})
+				}
+				if err := e.Wait(); err != nil {
+					return errors.Wrap(err, "syncing files:")
```
dgageot: Trailing `:`

pkg/skaffold/runner/runner.go
```diff
@@ -228,13 +231,33 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
...
 		switch {
 		case changed.needsReload:
 			logger.Stop()
 			return ErrorConfigurationChanged
-		case len(changed.dirtyArtifacts) > 0:
-			bRes, err := r.Build(ctx, out, r.Tagger, changed.dirtyArtifacts)
+		case len(changed.needsResync) > 0:
+			for _, s := range changed.needsResync {
+				if err := r.Syncer.Sync(s); err != nil {
+					logrus.Warnln("Skipping build and deploy due to sync error:", err)
```
dgageot: When the sync fails I don't see the warning.

pkg/skaffold/kubernetes/sync.go
```diff
@@ -0,0 +1,93 @@
...
+		return errors.Wrap(err, "copying files")
+	}
+	if err := perform(s.Image, s.Delete, deleteFileFn); err != nil {
+		return errors.Wrap(err, "deleting files")
+	}
+	return nil
+}
+
+func deleteFileFn(pod v1.Pod, container v1.Container, src, dst string) *exec.Cmd {
+	return exec.Command("kubectl", "exec", pod.Name, "--namespace", pod.Namespace, "-c", container.Name, "--", "rm", "-rf", dst)
```
dgageot: A debug log would be useful here.

pkg/skaffold/kubernetes/sync.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func copyFileFn(pod v1.Pod, container v1.Container, src, dst string) *exec.Cmd {
+	return exec.Command("kubectl", "cp", src, fmt.Sprintf("%s/%s:%s", pod.Namespace, pod.Name, dst), "-c", container.Name)
+}
```
dgageot: A debug log would be useful here.

pkg/skaffold/runner/runner.go
```diff
@@ -228,13 +231,33 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
...
+					return nil
+				}
+				color.Default.Fprintf(out, "Synced files for %s...\nCopied: %s\nDeleted: %s\n", s.Image, s.Copy, s.Delete)
```
dgageot: Maybe print the number of files in info and the names in debug

pkg/skaffold/runner/runner.go
dgageot: In fact, sometimes I see it...

pkg/skaffold/kubernetes/sync.go
r2d4: We get a debug log with util.RunCmd. Does that work?

pkg/skaffold/runner/runner.go
r2d4: Should there be any message in stdout?

stdout: ?
info: # of files synced
debug: filenames

pkg/skaffold/kubernetes/sync.go
r2d4: I don't think we have the namespace. In port-forward and log, we have to watch all pods.

r2d4: "" is all namespaces

pkg/skaffold/sync/sync.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	return ""
+}
+
+func intersect(context string, syncMap map[string]string, files []string) (map[string]string, error) {
+	ret := map[string]string{}
+	for _, f := range files {
+		relPath, err := filepath.Rel(context, f)
+		if err != nil {
+			return nil, errors.Wrapf(err, "changed file %s is not relative to context %s", f, context)
```
balopat: nit: changed file %s can't be found relative to context %s

pkg/skaffold/sync/sync.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		for p, dst := range syncMap {
+			match, err := filepath.Match(p, relPath)
+			if err != nil {
+				return nil, errors.Wrap(err, "pattern error")
```
balopat: nit: add `relPath` to error message?

pkg/skaffold/util/util.go
```diff
@@ -107,6 +107,13 @@ func ExpandPathsGlob(workingDir string, paths []string) ([]string, error) {
...
+func HasMeta(path string) bool {
```
balopat: nit: Meta -> WildCard

pkg/skaffold/util/util.go
```diff
@@ -107,6 +107,13 @@ func ExpandPathsGlob(workingDir string, paths []string) ([]string, error) {
...
+	return strings.ContainsAny(path, "*?[")
```
balopat: why does `[` have to be there? and if yes, why not `]` as well?

balopat: I see - this is the copy of hasMeta from filepath/match.go - in that file there is one more meta character `^` for negation...but this is really just nitpicking :)

pkg/skaffold/util/util.go
r2d4: I think its probably best to just use the name from the stdlib

r2d4: Not all of these at wildcards also

r2d4:
rebased again 👍

pkg/skaffold/kubernetes/sync.go
dgageot: yes should be good

pkg/skaffold/kubernetes/sync.go
dgageot: We do have a namespace sometimes but it's ok, your proposal of filtering on label is equally good.

pkg/skaffold/runner/runner.go
dgageot: I like info: # of files synced
debug: filenames

dgageot:
I'm tempted to merge your PR and let you or me fix the nits later.
wdyt?

#nomorerebase

r2d4:
I'll fix them quickly! :)

pkg/skaffold/runner/runner.go
r2d4: done

r2d4:
final rebase :)

integration/examples/hot-reload/skaffold.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: skaffold/v1alpha4
+kind: Config
+build:
+  artifacts:
+  - image: gcr.io/k8s-skaffold/node-example
+    context: node
+    sync:
+      '*.js': .
```
thomas-riccardi: Shouldn't we avoid user data as keys ?

r2d4: Why is that?

thomas-riccardi: I should have justified my comment indeed.
Coming from k8s it is unusual to have user data as keys; it surprised me.

Digging into k8s design doc, we have:
https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#lists-of-named-subobjects-preferred-over-maps
There are whole discussions linked from there, notably about
- readability (mostly in case of subobjects though)
- merging definitions, which may or may not be of concert for skaffold

It lists exceptions to this though: "pure maps [...] as opposed to sets of subobjects".

I suppose if we later want to enrich the format with subobjects it would make things harder with the current format.

I am not (yet) familiar with all the skaffold format, I don't know if there are other cases like that, I couln't find any.

So this is just to open the discussion. Should I move that into an issue?

r2d4: Thanks for the link. That actually makes a lot of sense. Please open an issue!  I think it’s a good suggestion

thomas-riccardi: here we go: #1180

Issue #947, opened 2018-08-31
skaffold trying to create a bucket and failing


ahmetb:
### Actual behavior

`skaffold run` failed with

`build step: building [gcr.io/testproject100000/emailservice]: creating bucket if not exists: googleapi: Error 409: You already own this bucket. Please select another name., conflict`

### Expected behavior

v0.13.0

Just tried `skaffold run -p gcb` with googleCloudBuild on a new project. Project had GCB enabled and at least one image already in GCR.

`skaffold` shouldn't be in the business of creating GCS buckets.

### Information

- Skaffold version: v0.13.0
- Operating system: cloud shell

r2d4:
I agree. However, Cloud Build doesn't create the bucket for you afaik, its actually created by gcloud. This might have changed?

ahmetb:
This wasn't on my project so I had no way of reproing this. But I'm seeing

```
pkg/skaffold/build/container_builder.go
128-
129-	cbBucket := fmt.Sprintf("%s%s", cb.ProjectID, constants.GCSBucketSuffix)
130-	buildObject := fmt.Sprintf("source/%s-%s.tar.gz", cb.ProjectID, util.RandomID())
131-
132-	if err := cb.createBucketIfNotExists(ctx, cbBucket); err != nil {
133:		return nil, errors.Wrap(err, "creating bucket if not exists")
```

Sooo maybe the `if not exists` check isn't working correctly.

MrTrustor:
Just ran into this on a brand new project trying to deploy the [hipster shop](https://github.com/GoogleCloudPlatform/microservices-demo/tree/1bab006af1d7d856b41ea67ec1824277120ae549).

GCR, GCB and GKE APIs manually enabled, a single GKE cluster created (as per instructions in the hipster shop README) and ran a `skaffold run -b gcb`.

ahmetb:
it clearly comes from

```
        if err := c.Bucket(bucket).Create(ctx, projectID, &cstorage.BucketAttrs{
                Name: bucket,
        }); err != nil {
                return err
        }
```

but I'm not sure why the following code fails to detect the bucket.

```
        _, err = c.Bucket(bucket).Attrs(ctx)

        if err == nil {
                // Bucket exists
                return nil
        }

        if err != cstorage.ErrBucketNotExist {
                return errors.Wrapf(err, "getting bucket %s", bucket)
        }
```

Eventual consistency???

r2d4:
[...]

PR #1044, opened 2018-09-26
Catch a 409 when creating a bucket and continue.

dlorenc:
I think this should fix https://github.com/GoogleContainerTools/skaffold/issues/947

### DIFF
--- a/pkg/skaffold/build/gcb/cloud_build.go
+++ b/pkg/skaffold/build/gcb/cloud_build.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net/http"
 	"time"
 
 	cstorage "cloud.google.com/go/storage"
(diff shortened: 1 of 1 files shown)


pkg/skaffold/build/gcb/cloud_build.go
```diff
@@ -253,9 +254,18 @@ func (b *Builder) createBucketIfNotExists(ctx context.Context, projectID, bucket
...
 
-	if err := c.Bucket(bucket).Create(ctx, projectID, &cstorage.BucketAttrs{
+	err = c.Bucket(bucket).Create(ctx, projectID, &cstorage.BucketAttrs{
 		Name: bucket,
-	}); err != nil {
+	})
+	if e, ok := err.(*googleapi.Error); ok {
+		if e.Code == http.StatusConflict {
+			// 409 errors are ok, there could have been a race condition or eventual consistency.
+			logrus.Debugf("Not creating bucket, got a 409 error indicating it already exists.")
```
ahmetb: I'd probably log the err here for diagnostics (i.e. in case some other 409 is returned in the future)

PR #1043, opened 2018-09-26
Add skeleton for jib integration

loosebazooka:
Mostly following the style of the bazel builder

- Not sure yet where to drop tests for Jib for (since there is no bazel equivalent):
  - runner.go
  - local.go
  - config.go

PR #1058, opened 2018-09-27
 Implements jib.GetDependenciesMaven/GetDependenciesGradle.

coollog:
This PR builds off of #1043 and should merge that before merging this.

Implements `jib/jib.go` with `GetDependenciesMaven` and `GetDependenciesGradle`.

For Maven, calls out to `mvn jib:_skaffold-files -q`. Uses the wrapper `mvnw` if present. Uses Windows wrapper `mvnw.cmd` if present and OS is Windows.

For Gradle, calls out to `gradle _jibSkaffoldFiles -q`. Uses the wrapper `gradlew` if present. Uses Windows wrapper `gradlew.bat` if present and OS is Windows.

Note that this only covers the single-module use case and multi-module use case will be added in future PRs.

@GoogleContainerTools/java-tools


pkg/skaffold/jib/jib.go
```diff
@@ -0,0 +1,79 @@
...
+func GetDependenciesGradle(workspace string, _ /*a*/ *v1alpha3.JibGradleArtifact, isWindows bool) ([]string, error) {
+	return getDependencies(workspace, "build.gradle", "gradle", "gradlew.bat", isWindows, []string{"_jibSkaffoldFiles", "-q"}, "jib-gradle")
+}
+
+func exists(workspace string, filename string) bool {
+	_, err := os.Stat(filepath.Join(workspace, filename))
+	return err == nil
+}
+
+func getDependencies(workspace string, buildFile string, defaultExecutable string, windowsExecutable string, isWindows bool, subCommand []string, artifactName string) ([]string, error) {
```
coollog: Would a struct over a long param list be preferred?

pkg/skaffold/jib/jib.go
```diff
@@ -0,0 +1,82 @@
...
+	if !exists(workspace, buildFile) {
+		return nil, errors.Errorf("no %s found", buildFile)
+	}
+
+	executable := defaultExecutable
+	if isWindows {
+		if exists(workspace, windowsExecutable) {
+			executable = "call"
+			subCommand = append([]string{windowsExecutable}, subCommand...)
```
briandealwis: Do we need to use `call` from within Go?

pkg/skaffold/jib/jib.go
```diff
@@ -18,13 +18,65 @@ package jib
...
-func GetDependenciesGradle(_ /*workspace*/ string, _ /*a*/ *v1alpha3.JibGradleArtifact) ([]string, error) {
-	return nil, errors.New("jib gradle support is unimplemented")
+// GetDependenciesGradle finds the source dependencies for the given jib-gradle artifact.
+// All paths are absolute.
+// TODO(coollog): Add support for multi-module projects.
+func GetDependenciesGradle(workspace string, _ /*a*/ *v1alpha3.JibGradleArtifact, isWindows bool) ([]string, error) {
+	return getDependencies(workspace, "build.gradle", "gradle", "gradlew.bat", isWindows, []string{"_jibSkaffoldFiles", "-q"}, "jib-gradle")
+}
+
+func exists(workspace string, filename string) bool {
```
briandealwis: It would be worth making this an `isDir()` that also tests that the `!info.IsDir()`?

pkg/skaffold/jib/jib.go
briandealwis: Could `isWindows` be a package variable? I created some helpers for execution that looks for the maven/gradle wrappers with different extensions. (https://github.com/briandealwis/skaffold/tree/jib-build)

coollog: Linking the commit for reference: https://github.com/briandealwis/skaffold/commit/1ec213dca6f72b22dfadc277b74833eddff4947d

I'll try to change that and see if the testing works well.

pkg/skaffold/jib/jib.go
coollog: Ah, I actually need to use `cmd.exe /C ...` instead. Also, @GoogleContainerTools/container-tools is there a style preference for using back-ticks for string constants vs quotes?

tstromberg: Not that I'm aware of: we generally follow standard Go & Kubernetes Coding Conventions: https://github.com/kubernetes/community/blob/master/contributors/guide/coding-conventions.md

In practice, raw string literals (back-ticks) in Go are only used for strings containing a back-slash or double quotes. Other string declarations should probably use the interpreted string literal form (double quotes) unless there is a special requirement to do so.

coollog: Thanks for clarifying!

pkg/skaffold/jib/jib.go
```diff
@@ -0,0 +1,82 @@
...
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+// GetDependenciesMaven finds the source dependencies for the given jib-maven artifact.
+// All paths are absolute.
+// TODO(coollog): Add support for multi-module projects.
+func GetDependenciesMaven(workspace string, _ /*a*/ *v1alpha3.JibMavenArtifact, isWindows bool) ([]string, error) {
+	return getDependencies(workspace, "pom.xml", "mvn", "mvnw.cmd", isWindows, []string{"jib:_skaffold-files", "-q"}, "jib-maven")
```
loosebazooka: This might need to specify the profile in use.

pkg/skaffold/schema/v1alpha3/config.go
```diff
@@ -224,6 +226,17 @@ type BazelArtifact struct {
 	BuildTarget string `yaml:"target"`
 }
 
+type JibMavenArtifact struct {
+	// Only multi-module
+	Module  string `yaml:"module"`
+	Profile string `yaml:"profile"`
```
loosebazooka: Turns out profile is not multimodule only.

coollog:
@balopat Will change into that format.

pkg/skaffold/jib/jib.go
```diff
@@ -18,13 +18,101 @@ package jib
...
+	mavenExecutable     = "mvn"
+	mavenWrapper        = "mvnw"
+	mavenWindowsWrapper = "mvnw.cmd"
+
+	gradleExecutable     = "gradle"
+	gradleWrapper        = "gradlew"
+	gradleWindowsWrapper = "gradlew.bat"
+)
+
+func getCommandMaven(workspace string, a *v1alpha3.JibMavenArtifact, isWindows bool) (executable string, subCommand []string) {
```
briandealwis: Can we generalize this and getCommandGradle to take the subcommand as an argument?

coollog: `getCommand` does take `subCommand` (`defaultSubCommand`) as an argument

coollog:
Changed to use *_GOOS filename format - LMK if this looks okay.

pkg/skaffold/jib/jib_darwin.go
```diff
@@ -0,0 +1,33 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package jib
+
+func getWrapper() string {
+	return "gradlew"
```
briandealwis: This doesn't seem right?

pkg/skaffold/jib/jib_darwin.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func getCommand(workspace string, defaultExecutable string, defaultSubCommand []string) (executable string, subCommand []string) {
+	executable = defaultExecutable
+	subCommand = defaultSubCommand
+
+	wrapperExecutable := getWrapper()
+	if exists(workspace, wrapperExecutable) {
+		executable = "./" + wrapperExecutable
```
briandealwis: I think it's worth making the path absolute just in case the command-execution sets a different working directory

pkg/skaffold/jib/jib_linux.go
```diff
@@ -0,0 +1,33 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package jib
+
+func getWrapper() string {
+	return "gradlew"
```
briandealwis: Doesn't seem right?

pkg/skaffold/jib/jib_linux.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func getCommand(workspace string, defaultExecutable string, defaultSubCommand []string) (executable string, subCommand []string) {
+	executable = defaultExecutable
+	subCommand = defaultSubCommand
+
+	wrapperExecutable := getWrapper()
+	if exists(workspace, wrapperExecutable) {
+		executable = "./" + wrapperExecutable
```
briandealwis: Ditto returning an absolute path

pkg/skaffold/jib/jib_darwin.go
```diff
@@ -0,0 +1,33 @@
...
+limitations under the License.
+*/
+
+package jib
+
+func getWrapper() string {
+	return "gradlew"
+}
+
+func getCommand(workspace string, defaultExecutable string, defaultSubCommand []string) (executable string, subCommand []string) {
```
briandealwis: Can you capitalize this so that it can be used elsewhere?

Or provide `InvokeMaven` and `InvokeGradle` variants that hide the wrapper bit.

pkg/skaffold/jib/jib.go
```diff
@@ -18,13 +18,76 @@ package jib
...
+	return getDepsFromStdout(string(stdout)), nil
+}
+
+const (
+	mavenExecutable  = "mvn"
+	gradleExecutable = "gradle"
+)
+
+func getCommandMaven(workspace string, a *v1alpha3.JibMavenArtifact) (executable string, subCommand []string) {
+	subCommand = []string{"jib:_skaffold-files", "-q"}
```
briandealwis: `com.google.cloud.tools:jib-maven-plugin::_skaffold-files` — I discovered by accident that we need to use partially-qualified plugin references in case the plugin isn't configured in the pom.

coollog: I believe currently we are only planning to support projects that have the plugin configured due to difficulties with multi-module projects if the plugin isn't configured @loosebazooka

pkg/skaffold/jib/jib_darwin.go
coollog: I think I'll leave that to another PR - one that would use this method from another package - it's probably good to keep visibility as low as possible until it needs to be exposed.

briandealwis: I want to replace what I have :-)

pkg/skaffold/jib/jib_darwin.go
coollog: Oh is this missing something?

coollog: Oh oops. Hmm, how did this even pass the tests.

coollog:
Hmm, it seems if I name a test like `jib_non_windows_test.go`, it doesn't execute when I do `make test` on my linux machine. @GoogleContainerTools/container-tools

pkg/skaffold/jib/jib_darwin.go
coollog: Would making `getCommandMaven` and `getCommandGradle` exposed be good? I'll change it to include a custom subcommand to use.

pkg/skaffold/jib/jib.go
loosebazooka: This is more of a concern in multimodule projects, where the processing needs to be multi module aware. But I would much prefer it was explicit.

loosebazooka: sorry, I meant explicitly defined in the pom.xml.

coollog:
Alright, finally got Windows to pass.

pkg/skaffold/jib/jib_linux.go
```diff
@@ -0,0 +1,38 @@
+/*
```
loosebazooka: do we really need both jib_linux and jib_darwin? @balopat ?

r2d4: no just call it `jib.go` and put `// +build linux darwin` at the top

r2d4: you can call it `jib_unix.go` if you really want

coollog: Okay, will go with `jib_unix.go` since `jib.go` has common logic shared with `jib_windows.go`

pkg/skaffold/jib/jib.go
```diff
@@ -17,14 +17,86 @@ limitations under the License.
...
+	}
+	if info.IsDir() {
+		return "", errors.Errorf("%s is a directory", file)
+	}
+	return filepath.Abs(file)
+}
+
+func getDependencies(workspace string, buildFile string, executable string, subCommand []string, artifactName string) ([]string, error) {
+	if _, err := resolveFile(workspace, buildFile); err != nil {
+		return nil, errors.Errorf("no %s found", buildFile)
```
balopat: nit: use `errors.Wrap` instead with `err`

pkg/skaffold/jib/jib.go
```diff
@@ -17,14 +17,86 @@ limitations under the License.
...
+
+// GetDependenciesGradle finds the source dependencies for the given jib-gradle artifact.
+// All paths are absolute.
+func GetDependenciesGradle(workspace string, a *v1alpha3.JibGradleArtifact) ([]string, error) {
+	executable, subCommand := getCommandGradle(workspace, a)
+	return getDependencies(workspace, "build.gradle", executable, subCommand, "jib-gradle")
+}
+
+// resolveFile resolves the absolute path of the file named filename in directory workspace, erroring if it is not a file
+func resolveFile(workspace string, filename string) (string, error) {
```
balopat: - this could live in `pkg/skaffold/util/util.go` 
- I would rename it to just `AbsFile`. 
- `workspace` -> `dir`

pkg/skaffold/jib/jib_unix.go
```diff
@@ -0,0 +1,40 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package jib
+
+func getWrapper(defaultExecutable string) string {
+	switch defaultExecutable {
```
balopat: this smells a little bit like we are trying to decide between maven/gradle logic using arbitrary consts (lack of enums I guess?), while in other cases we are using "Maven/Gradle" postfix. It's starting to get confusing what logic belongs to maven/gradle and what's common. I'd like to see if introducing the subpackages under jib/maven, jib/gradle would help or not.

coollog:
Rearranged the structure into separate `jib_maven.go`/`jib_gradle.go` and their respective OS-specific code. Let me know if this structure looks okay and I'll migrate the tests over as well.

balopat:
It looks good to me, go ahead with the tests!

pkg/skaffold/jib/jib_unix.go
```diff
@@ -0,0 +1,37 @@
...
+*/
+
+package jib
+
+import (
+	"os/exec"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+)
+
+func getCommand(workspace string, defaultExecutable string, wrapperExecutable string, defaultSubCommand []string) *exec.Cmd {
```
loosebazooka: if there's no alt to the subCommand, do we need to specify default?

coollog: Nope, I'll rename to just `subCommand`

loosebazooka:
We don't need to cover expanding directories in this pr, right?

coollog:
Tests are updated as well and added TODO for expanding directories in follow-up PR.

PR #1064, opened 2018-09-28
Watch returns filemap

r2d4:
splits out the watch changes from #1039 

**Why are watch events computed in the component onChange instead of the watch's onChange?** 
Because watch events are computed per component, and they will need to be used per component. This will also make the resync/rebuild logic a lot easier

The current file sync PR actually does the sync in the component on change, but I think we can do something similar to the rebuild and redeploy where we just do it in the main watch loop.


balopat:
it would be interesting to see the performance impact of storing an actual file map instead of just the count of filechanges in `fileMap`. Both in memory and operations is now O(n) instead of O(1) in number of artifact dependencies.

pkg/skaffold/watch/changes.go
```diff
@@ -18,59 +18,86 @@ package watch
...
+	Deleted  []string
+}
 
-	for _, path := range paths {
-		stat, err := os.Stat(path)
-		if err != nil {
-			if os.IsNotExist(err) {
-				continue // Ignore files that don't exist
-			}
+func (e Events) HasChanged() bool {
```
balopat: I'd risk putting a boolean field on `Events` that I'd flip in the if statements in `events()` below...?

balopat:
Oh I guess the number of operations doesn't change significantly as we were O(n) already anyway, it's O(2*n) now and the memory impact is probably insignificant.

pkg/skaffold/watch/changes.go
```diff
@@ -18,59 +18,86 @@ package watch
...
+	added, deleted, modified := len(e.Added), len(e.Deleted), len(e.Modified)
+	if added > 0 {
+		logrus.Debugf("[watch event] added: %s", e.Added)
```
balopat: having the logs in a getter is a bit surprising (nothing prevents a client of this API would call HasChanged multiple times, but it can generate unnecessary clutter in the logs) I'd rather have them in the `events()` at calculation time.

pkg/skaffold/watch/changes.go
balopat: On second thought this is elegant enough and O(1) and there is no extra state to be maintained. Please ignore.

pkg/skaffold/watch/changes.go
r2d4: Good call!  I’ll change

dgageot:
@r2d4 I'll review the PR on Monday and will run some performance tests. People already complain that the watch consumes too much CPU. I'll run a comparison

r2d4:
Thanks! I think the CPU usage is from the continuous filesystem Stat, which hasn't changed

pkg/skaffold/watch/watch.go
```diff
@@ -73,16 +74,18 @@ func (w *watchList) Run(ctx context.Context, pollInterval time.Duration, onChang
...
 		case <-ticker.C:
 			changed := 0
-
+			var e Events
 			for i, component := range *w {
 				state, err := stat(component.deps)
 				if err != nil {
 					return errors.Wrap(err, "listing files")
 				}
+				e = events(component.state, state)
```
dgageot: Could we write `e := events(component.state, state)` here?

pkg/skaffold/watch/changes.go
```diff
@@ -17,60 +17,96 @@ limitations under the License.
...
+type Events struct {
+	Added    []string
+	Modified []string
+	Deleted  []string
+}
 
-	for _, path := range paths {
-		stat, err := os.Stat(path)
-		if err != nil {
-			if os.IsNotExist(err) {
```
dgageot: This test should be kept. And we might want to had a unit test to protect against breaking it.
Watch shouldn't fail on a broken link or if a file disappear just after the artifact says it relies on this file.

r2d4: Good catch. Added it back and added some explicit test coverage of those cases.

pkg/skaffold/watch/watch.go
r2d4: Yes, leftover artifact of some refactoring, changed

PR #1070, opened 2018-10-01
Transform values files in profiles to v1alpha3

GeertJohan:
I made a mistake in #985. With the v0.15.0 `skaffold fix` the `valuesFilePath` field in profiles is completely dropped. These changes fix that.

These changes are based on the `v0.15.0` tag. I think this PR shouldn't actually point to master but to a v0.15.x branch from which we may release v0.15.1

Apologies for the inconvenience. I hope it's possible to push out a hotfix release fast.

### DIFF
--- a/pkg/skaffold/config/transform/transforms.go
+++ b/pkg/skaffold/config/transform/transforms.go
@@ -148,6 +148,16 @@ func ToV1Alpha3(vc util.VersionedConfig) (util.VersionedConfig, error) {
 			return nil, errors.Wrap(err, "converting new profile")
 		}
 	}
+	// if the helm deploy config was set for a profile, then convert ValueFilePath to ValuesFiles
+	for p, oldProfile := range oldConfig.Profiles {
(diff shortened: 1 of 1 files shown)


balopat:
Yeah, I think with transformation logic we could go as thorough as covering quick-check style a lot of combinations, or the lazy way is to provide a couple of full golden examples and provide some coverage. 

We do have a "the conversion doesn't fail and the result is deployable" test at https://github.com/GoogleContainerTools/skaffold/blob/master/integration/run_test.go#L224 but that should already cover your case. 

And with that, I'm going to unblock this PR. I think we should improve the test coverage on these though on the long run or do a better validation integration test.

GeertJohan:
@balopat Can we release this hotfix as v0.15.1? Otherwise users will break their skaffold.yaml when switching to v0.15.0 and v0.16.0 is still a few weeks away I guess?
cc @r2d4

balopat:
Yes, I will work on the v0.15.1 release today!

PR #1085, opened 2018-10-03
Implement a manual trigger for watch mode

dgageot:
One can now run `skaffold dev --trigger=manual` to:
 + don't poll for files changes
 + let the user press any key to then compute the diff and rebuild and redeploy what has changed.

The default trigger type is `polling`. Another trigger type could expose a named pipe or socket to send refresh commands.

### DIFF
--- a/pkg/skaffold/watch/triggers.go
+++ b/pkg/skaffold/watch/triggers.go
@@ -1,0 +1,111 @@
+/*
+Copyright 2018 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


pkg/skaffold/watch/triggers.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
+	"github.com/sirupsen/logrus"
+)
+
+// Trigger describes a mechanism that triggers the watch.
+type Trigger interface {
+	Start() (<-chan bool, func())
+	WatchingForChanges(io.Writer)
```
balopat: nit: Watching -> Watch

r2d4:
What if we had implemented manual triggering as a different watcher altogether? Since really its a `keybinding watcher`. 

I still think you always want both at the same time

dgageot:
@r2d4 with current code, we could do both but I have no idea how that would work.

Issue #1096, opened 2018-10-04
Resolve issues to finish Jib builder.


coollog:
- [x] Rename error message to `jibMaven` and `jibGradle`. #1097
- [x] `GetDependenciesGradle` process `JibGradleArtifact.Project` #1097 
- [x] `runner.dependenciesForArtifact` seems to try to resolve absolute deps against workspace #1097
- [x] `GetDependencies*` to list files recursively #1097
- [x] Error if trying to use Maven multi-module in `jib.GetDependenciesMaven` #1097 
- [x] Combine logic for resolving Maven/Gradle executable and wrapper #1105
- [x] Build directly to registry when `push` (#1114) is true #1124 #1132
- [x] Handle Maven multi-module (lower priority - will not be in `0.16.0`) #1152
- [x] Use `godirwalk.walk` (https://github.com/GoogleContainerTools/skaffold/pull/1097#discussion_r223378336) #1125
- [x] Don't walk dep if is workspace root #1125
- [x] Don't use fully-qualified `jib-maven-plugin` goal in builder #1129

/cc @GoogleContainerTools/java-tools

briandealwis:
> - [ ] Build directly to registry when `skipPush` is true

The `local` builder [handles this already](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/local/local.go#L79). I don't think there's anything for us to do here.

briandealwis:
I'm reworking the wrapper logic.

saturnism:
Quick question.. if `skipPush: true`, then it will build directly to registry (which is likely remote), what does `skipPush: false` mean?

coollog:
That was actually supposed to be when `skipPush` is false. Fixed it. When `skipPush` is true, Jib will build to the Docker daemon.

balopat:
FYI. I added #1114  which will invert SkipPush to Push.

coollog:
The logic for handling `push=true` seems to be not as trivial as originally thought, mainly due to the fact that 3 out of the 5 taggers require the image digest to generate the intended tag (`ChecksumTagger`, `GitCommit`, and `envTemplateTagger`).

`local.go#buildArtifact` performs a series of steps for a build:

1. `b.runBuildForArtifact` builds to the Docker daemon and gets the initial tag
1. `docker.Digest` gets the digest for the initial tag from the Docker daemon
1. `tagger.GenerateFullyQualifiedImageName` generates the intended tag
[...]

PR #1097, opened 2018-10-04
 Lists files recursively in jib.getDependencies and other fixes.

coollog:
Resolves issues in #1096:

- Rename error message to `jibMaven` and `jibGradle`.
- `GetDependenciesGradle` process `JibGradleArtifact.Project`
- `runner.dependenciesForArtifact` seems to try to resolve absolute deps against workspace
- `GetDependencies*` to list files recursively
- Error if trying to use Maven multi-module in `jib.GetDependenciesMaven`


loosebazooka:
should we be rebased against that feature branch?

pkg/skaffold/jib/jib.go
```diff
@@ -30,14 +35,34 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
...
-		if l == "" {
+	for _, dep := range lines {
+		if dep == "" {
+			continue
+		}
+
+		// Resolves directories recursively.
+		info, err := os.Stat(dep)
+		if err != nil {
+			if os.IsNotExist(err) {
```
coollog: This is modeled after the logic in `changes.go#stat`. I think ignoring non-existent files should be fine here since the files would be ignored by `stat` anyways. The alternative is to just add the file to the `deps` list upon this `os.Stat` error. Thoughts?

pkg/skaffold/runner/runner.go
```diff
@@ -428,7 +428,10 @@ func dependenciesForArtifact(a *latest.Artifact) ([]string, error) {
 
 	var p []string
 	for _, path := range paths {
-		p = append(p, filepath.Join(a.Workspace, path))
+		if !filepath.IsAbs(path) {
```
coollog: This should be fine in regards to the other artifacts right? There should be no case where a dep would be absolute yet need to resolve against workspace?

dgageot:
Sorry, this needs to be rebased

coollog:
This is rebased and good for review.

pkg/skaffold/jib/jib.go
```diff
@@ -30,14 +35,34 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
...
+				logrus.Debugf("could not stat dependency: %s", err)
+				continue // Ignore files that don't exist
+			}
+			return nil, errors.Wrapf(err, "unable to stat file %s", dep)
+		}
+		if info.IsDir() {
+			err := filepath.Walk(dep, func(path string, info os.FileInfo, err error) error {
```
dgageot: `godirwalk.Walk` is much faster at walking directories (like 10x faster), which is important since this is going to run every second in watch mode.
It's ok to leave that this way but we should take note to improve it in another PR.

PR #1102, opened 2018-10-05
Improve sync

dgageot:
+ Add more logs and fix some of them
+ Improve errors
+ Pass context to commands

### DIFF
--- a/pkg/skaffold/kubernetes/sync.go
+++ b/pkg/skaffold/kubernetes/sync.go
@@ -20,9 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
-	"strings"
 
-	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/sync"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 	"github.com/sirupsen/logrus"
@@ -63,23 +61,13 @@ func copyFileFn(ctx context.Context, pod v1.Pod, container v1.Container, src, ds
 	return exec.CommandContext(ctx, "kubectl", "cp", src, fmt.Sprintf("%s/%s:%s", pod.Namespace, pod.Name, dst), "-c", container.Name)
 }
 
-func labelSelector() string {
-	var reqs []string
-	for k, v := range constants.Labels.DefaultLabels {
-		reqs = append(reqs, fmt.Sprintf("%s=%s", k, v))
-	}
-	return strings.Join(reqs, ",")
-}
-
 func perform(ctx context.Context, image string, files map[string]string, cmdFn func(context.Context, v1.Pod, v1.Container, string, string) *exec.Cmd) error {
 	client, err := Client()
 	if err != nil {
 		return errors.Wrap(err, "getting k8s client")
 	}
 
-	pods, err := client.CoreV1().Pods("").List(meta_v1.ListOptions{
-		LabelSelector: labelSelector(),
-	})
+	pods, err := client.CoreV1().Pods("").List(meta_v1.ListOptions{})
 	if err != nil {
 		return errors.Wrap(err, "getting pods")
 	}
--- a/pkg/skaffold/kubernetes/sync.go
+++ b/pkg/skaffold/kubernetes/sync.go
@@ -20,14 +20,12 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
-	"sync/atomic"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/sync"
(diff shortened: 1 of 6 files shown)


pkg/skaffold/runner/runner.go
```diff
@@ -271,12 +271,12 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
 			return ErrorConfigurationChanged
 		case len(changed.needsResync) > 0:
 			for _, s := range changed.needsResync {
-				if err := r.Syncer.Sync(s); err != nil {
+				color.Default.Fprintf(out, "Syncing %d files for %s\n", len(s.Copy)+len(s.Delete), s.Image)
+
+				if err := r.Syncer.Sync(ctx, s); err != nil {
 					logrus.Warnln("Skipping build and deploy due to sync error:", err)
 					return nil
```
r2d4: I forgot to add this but if we get an error here, we should add that dirty artifact back to the rebuild list

pkg/skaffold/runner/runner.go
```diff
@@ -271,12 +271,12 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
 			return ErrorConfigurationChanged
 		case len(changed.needsResync) > 0:
 			for _, s := range changed.needsResync {
-				if err := r.Syncer.Sync(s); err != nil {
+				color.Default.Fprintf(out, "Syncing %d files for %s\n", len(s.Copy)+len(s.Delete), s.Image)
```
r2d4: update to latest code

pkg/skaffold/kubernetes/sync.go
```diff
@@ -34,60 +34,76 @@ import (
...
+func perform(ctx context.Context, image string, files map[string]string, cmdFn func(context.Context, v1.Pod, v1.Container, string, string) *exec.Cmd) error {
 	client, err := Client()
 	if err != nil {
 		return errors.Wrap(err, "getting k8s client")
 	}
-	pods, err := client.CoreV1().Pods("").List(meta_v1.ListOptions{
-		LabelSelector: labelSelector(),
-	})
+
+	pods, err := client.CoreV1().Pods("").List(meta_v1.ListOptions{})
```
r2d4: use the skaffold labelselector here

pkg/skaffold/kubernetes/sync.go
```diff
@@ -34,60 +34,76 @@ import (
...
+				})
 			}
 		}
 	}
+
+	if err := e.Wait(); err != nil {
+		return errors.Wrap(err, "syncing files")
+	}
+
+	if int(performed) != len(files) {
```
r2d4: I don't think you need to count here. errgroup handles this

pkg/skaffold/kubernetes/sync.go
```diff
@@ -34,60 +34,76 @@ import (
 
 type KubectlSyncer struct{}
 
-func (k *KubectlSyncer) Sync(s *sync.Item) error {
-	if err := perform(s.Image, s.Copy, copyFileFn); err != nil {
-		return errors.Wrap(err, "copying files")
+func (k *KubectlSyncer) Sync(ctx context.Context, s *sync.Item) error {
+	if len(s.Copy) > 0 {
```
r2d4: maybe move the len check into `perform`? a few lines of code saved

pkg/skaffold/kubernetes/sync.go
r2d4: ah good point 👍

pkg/skaffold/runner/runner.go
r2d4: Yeah just maybe the last commit. I changed this message to the debug and info statement.

pkg/skaffold/kubernetes/sync.go
r2d4: Hm, I guess this is a bit tricky. What if we have 2 replicas?

pkg/skaffold/runner/runner.go
dgageot: Oh, the one I asked you to add? It didn't work well in reality :-) Didn't give enough feedback

r2d4: :) SGTM

pkg/skaffold/kubernetes/sync.go
dgageot: Good point, I'll see how to fix it. Each file should be synced at least once, I guess.

r2d4: Yeah I think thats the best we can do. But making sure we sync at least once is a good check.

pkg/skaffold/runner/runner.go
dgageot: How would you do that?

PR #1103, opened 2018-10-05
Add a test to check version upgrades

dgageot:
Each version should upgrade to the next version.
This test is an additional safety net when adding
a new version.


dgageot:
ping @balopat this is a test to protect us from breaking the upgrade mechanism in the future

Issue #1104, opened 2018-10-05
SkaffoldRunner Dev should pass along its context


briandealwis:
`SkaffoldRunner#Dev` creates a watcher and registers artifact watchers via `dependenciesForArtifact()`, and should pass along its context.

PR #1105, opened 2018-10-05
Unify Jib command wrapper usage

briandealwis:
Create a new `CommandWrapper` structure to specify the executable and wrapper names.  Command lines are generated by a `CommandWrapper#CreateCommand()` method. I found examples of Maven having a .bat and .cmd wrappers, so we leave the extension determining to the `CreateCommand()` method.

The builder codepath passes along a context, but the dependency-tracking codepath does not (#1104) so I pass along a dummy context for now.


pkg/skaffold/util/wrapper_unix.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// CreateCommand creates an `exec.Cmd` that is configured to call the
+// executable (possibly using a wrapper in `workingDir`, when found) with the given arguments,
+// with working directory set to `workingDir`.
+func (cw CommandWrapper) CreateCommand(ctx context.Context, workingDir string, args []string) *exec.Cmd {
+	executable := cw.Executable
+
+	if cw.Wrapper != "" && !SkipWrapperCheck {
```
coollog: I noticed `SkipWrapperCheck` is always false - was this intentional?

PR #1107, opened 2018-10-07
azure container registry runner

cedrickring:
Added an acr runner similar to `az acr build` to build and push images on acr.


cedrickring:
I signed it!

pkg/skaffold/schema/latest/config.go
```diff
@@ -22,7 +22,7 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
-const Version string = "skaffold/v1alpha4"
+const Version string = "skaffold/v1alpha5"
```
dgageot: Hi @cedrickring, since you're only adding new fields, it's ok to not create a new version.

pkg/skaffold/build/acr/acr_build.go
```diff
@@ -0,0 +1,178 @@
...
+			}
+		}
+
+		time.Sleep(2 * time.Second)
+	}
+}
+
+// ACR needs the image tag in the following format
+// <repository>:<tag>
+func getImageTagWithoutFQDN(imageTag string) (string, error) {
```
dgageot: Could you add a few test for this piece of code? Could you replace the manual parsing a an image name with something like `docker.ParseReference()` in `github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker` package?

pkg/skaffold/build/acr/azure_storage.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return err
+	}
+	req.Header.Add("x-ms-blob-type", "BlockBlob")
+	req.Header.Add("x-ms-version", VERSION)
+	req.Header.Add("x-ms-date", time.Now().String())
+	req.Header.Add("Content-Length", fmt.Sprint(s.Buffer.Len()))
+
+	client := http.Client{}
+	response, err := client.Do(req)
```
dgageot: I think the response needs to be closed at some point.

pkg/skaffold/schema/latest/config.go
```diff
@@ -118,6 +119,23 @@ type KanikoBuild struct {
 	Image          string             `yaml:"image,omitempty"`
 }
 
+// AzureContainerBuild contains the fields needed to do a build
+// on Azure Container Registry
+type AzureContainerBuild struct {
+	Credentials       AzureContainerBuildCredentials `yaml:"credentials,omitempty"`
+	ContainerRegistry string                         `yaml:"containerRegistry,omitempty"`
```
dgageot: Could a few of those field have sensible default values or be, by default, derived from the image name. Like if I push an image called `myregistry.azurecr.io/skaffold-example` would it make sense to set `containerRegistry` to `myregistry` by default?

That's something we do on GCB to make sure the Skaffold config is as short as possible and contains as few duplication as required.

Gopkg.lock
```diff
@@ -16,6 +16,17 @@
   revision = "050b16d2314d5fc3d4c9a51e4cd5c7468e77f162"
   version = "v0.17.0"
 
+[[projects]]
+  digest = "1:fd5f5c00789da5a6ae78b6c6050eccea46bcd4a519a4913ac82d22bafd7fd368"
+  name = "github.com/Azure/azure-sdk-for-go"
```
dgageot: Should `github.com/Azure/azure-sdk-for-go` be added as a top level `constraint` in Gopkg.toml`?

examples/acr/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM gcr.io/google-appengine/golang
```
dgageot: nit, you might want to use a smaller image to accelerate the build. (on GCR, we use `gcr.io/*` images because some of them are pre-pulled on the build node.)

examples/acr/skaffold.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: skaffold/v1alpha3
+kind: Config
+build:
+  artifacts:
+  - imageName: myregistry.azurecr.io/skaffold-example
+  acr:
+    containerRegistry: myregistry
+    resourceGroup: myresourcegroup
+    credentials:
```
dgageot: Can the credentials be found automatically by Skaffold like it's done for GCB? With GCB, you don't have to configure anything in Skaffold if `gcloud` is properly authenticated to the right project. 

I thing both scenario make sense : 
 + One were you assume the user's machine is properly configured. In that case Skaffold can have the minimal configuration.
 + One were we prefer to configure Skaffold explicitly. We should add that mode to the GCP builder.

examples/acr/README.adoc
```diff
@@ -0,0 +1,34 @@
+=== Example: Azure Container Registry
```
dgageot: We are in a nasty chicken and egg situation were a sample config can't be put in `examples` folder until a release of Skaffold actually supports that config.

You should move the whole sample to `integration/examples` folder and it will be copied to top level `examples` next time a release is cut.

examples/acr/skaffold.yaml
cedrickring: I could create an Authorizer which uses the bearer token from the azure cli config.

So the priority would be: try to use provided credentials, if they're not present try to authenticate with bearer from the config, if the bearer has expired throw an error. I could try to call the azure cli to refresh the bearer though...

pkg/skaffold/build/acr/acr_build.go
```diff
@@ -0,0 +1,178 @@
...
+
+func (b *Builder) buildArtifact(ctx context.Context, out io.Writer, tagger tag.Tagger, artifact *latest.Artifact) (string, error) {
+	client := cr.NewRegistriesClient(b.Credentials.SubscriptionID)
+	authorizer, err := auth.NewClientCredentialsConfig(b.Credentials.ClientID, b.Credentials.ClientSecret, b.Credentials.TenantID).Authorizer()
+	if err != nil {
+		return "", errors.Wrap(err, "authorizing client")
+	}
+	client.Authorizer = authorizer
+
+	result, err := client.GetBuildSourceUploadURL(ctx, b.ResourceGroup, b.ContainerRegistry)
```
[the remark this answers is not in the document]
cedrickring: Unfortunately not. When uploading to this URL, it saves the tar to a specific location `result.RelativePath` which has to be passed to the `DockerBuildRequest` as `sourceLocation`.

pkg/skaffold/schema/latest/config.go
cedrickring: Yes, I can get the container registry name from the image name, but the resource group needs to be provided separately since it may differ from the image name.

Gopkg.lock
cedrickring: I don't think that's necessary, because I'm using a versioned import path for azure sdk. See [here](https://github.com/GoogleContainerTools/skaffold/blob/88f8e32642a289beb04ebfdf61ac6b423f258c94/pkg/skaffold/build/acr/acr_build.go#L27)

pkg/skaffold/build/acr/acr_build.go
cedrickring: `docker.ParseReference()` only provides the `BaseName` consisting of `fqdn/repository`, but Azure only needs the repository name. I could extend the function to extract the repository name from the BaseName.

dgageot:
@nkubala @priyawadhwa @r2d4 Could one of you TAL?
@cedrickring One thing that is missing is the ability to interrupt the build. It's also missing on the kaniko builder but was implemented in the GCB builder. I'm ok to add that in another PR though

cedrickring:
@dgageot I'd like to do that in another PR. I'm not sure if it's possible to stop the build actually. But I'll dig into it later.

examples/annotated-skaffold.yaml
```diff
@@ -107,6 +107,19 @@ build:
   #   namespace: default
   #   timeout: 20m
 
+  # Docker artifacts can be built on an Azure Container Registry.
```
priyawadhwa: I think [integration/examples/annotated-skaffold.yaml](https://github.com/GoogleContainerTools/skaffold/blob/master/integration/examples/annotated-skaffold.yaml) should be modified instead (or in addition to this one), so that when we copy over config changes we don't lose this.

pkg/skaffold/build/acr/acr_build.go
```diff
@@ -0,0 +1,190 @@
...
+	cr "github.com/Azure/azure-sdk-for-go/services/containerregistry/mgmt/2018-09-01/containerregistry"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/tag"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+	"github.com/pkg/errors"
+)
+
+const BuildStatusHeader = "x-ms-meta-Complete"
```
r2d4: nit: I think you can leave this unexported

pkg/skaffold/build/acr/acr_build.go
```diff
@@ -0,0 +1,190 @@
...
+			//so the resourceGroup is the fourth element of the split
+			return strings.Split(*registry.ID, "/")[4], nil
+		}
+	}
+
+	return "", errors.New("Couldn't find resource group of registry")
+}
+
+func getImageTagWithoutFQDN(imageTag string) string {
+	return imageTag[strings.Index(imageTag, "/")+1:]
```
r2d4: I think theres a better way to do this using the go-containerregistry. Can you look at an example of the other builders

r2d4: Maybe add a test or two?

cedrickring: Thanks for the tip with go-containerregistry 👍

cedrickring:
Should be ready to merge imo

priyawadhwa:
Hey @cedrickring  -- we just released support for custom builders and skaffold. If you're still interested in using skaffold with Azure, this could be an option for you! Here's an [example](https://github.com/GoogleContainerTools/skaffold/tree/master/integration/testdata/custom) using a custom build script and associated [docs](https://skaffold.dev/docs/how-tos/builders/#custom-build-script-run-locally). If you end up trying it out, please open an issue if you run into any trouble, and let us know if it works for you!

cedrickring:
@priyawadhwa sounds really cool! I'd love to create a custom builder with Azure. Is there a way to specify a e.g. GitHub repo for the builder? So if I want to share my custom builder with others, they don't need to clone it everywhere they need it.

priyawadhwa:
hey @cedrickring , not as of now, but I've updated it as a TODO in #2009

PR #1114, opened 2018-10-08
skipPush -> push

### LINKED ISSUE #1113: Rename skipPush -> push
`skipPush` is a negated field, can we change it to `push`?  

`skipPush: false` -> `push: true`
`skipPush: true` -> `push: false`

balopat:
Fixes #1113.


balopat:
Thanks for catching that @dgageot! Done.

skaffold.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: skaffold/v1alpha3
 kind: Config
 build:
   artifacts:
-  - image: gcr.io/k8s-skaffold/skaffold
+  - imageName: gcr.io/k8s-skaffold/skaffold
     docker:
       dockerfile: deploy/skaffold/Dockerfile
```
dgageot: I think `dockerfile` should be `dockerfilePath`

balopat: agreed - I changed it

integration/examples/annotated-skaffold.yaml
```diff
@@ -77,7 +77,8 @@ build:
   # `useBuildkit` can also be set to activate the experimental BuildKit feature.
   #
   # local:
-  #   skipPush: true
+  #   false by default for local clusters, true for remote clusters
+  #   push: false
```
thomas-riccardi: The doc in the previous paragraph still talks about skip push, with inverted boolean value: this is confusing.

PR #1128, opened 2018-10-09
Trim the dockerfile a bit:

dlorenc:
- remove gcloud tarball (24M)
- switch to python instead of python-dev (28M vs 90M)
- rm var/lib/apt/lists (100M)

Overall savings of around 200M (1.9GB vs. 1.68 GB)


dlorenc:
```shell
    		, stderr: time="2018-10-09T14:17:56Z" level=fatal msg="build step: building [gcr.io/k8s-skaffold/skaffold-example]: could not create build: googleapi: Error 503: The service is currently unavailable., backendError"
```

PR #1132, opened 2018-10-09
 Handles Jib build directly to registry when push=true.

### LINKED ISSUE #1124: Proposal: How to handle Jib build directly to registry.
Currently, Jib can build to the Docker daemon and share the same logic as the rest of the `local` builders (`docker` and `bazel`):

`local.go#buildArtifact` performs a series of steps for a build:

1. `b.runBuildForArtifact` builds to the Docker daemon and gets the initial tag
1. `docker.Digest` gets the digest for the initial tag from the Docker daemon
1. `tagger.GenerateFullyQualifiedImageName` generates the intended tag
1. `b.api.ImageTag` retags the initial tag to the intended tag
1. if `push=true`, `docker.RunPush` pushes the intended tag

However, to leverage the full benefits of Jib, we would need to add the ability to call `jib:build` to build directly to the registry. The implementation logic would be very similar to how the `kaniko` builder works:

1. Generate initial tag as random ID.
1. Run `jib:build` to initial tag.
1. `docker.RemoteDigest` to get the digest
1. `tagger.GenerateFullyQualifiedImageName` to get the intended tag
1. `docker.AddTag` to retag with intended tag

The original plan was the perform this logic when `local.push=true`, but we're proposing a new builder type `jib` (only supports `JibMavenArtifact` and `JibGradleArtifact`).

Therefore, the following would be what users could do with the Jib artifacts:
- `local` and `push=false` - `jib:dockerBuild` builds to Docker daemon
- `local` and `push=true` - `jib:dockerBuild` builds to Docker daemon and then the image is pushed
- `jib` - `jib:build` builds directly to the registry

The caveat, though, is that the `jib` builder would not support other artifact types. The alternative is to just keep the original approach, where users could do:
- `local` and `push=false` - `jib:dockerBuild` builds to Docker daemon
- `local` and `push=true` - `jib:build` builds directly to the registry

@GoogleContainerTools/java-tools @GoogleContainerTools/container-tools

coollog:
Builds off of #1130, so will merge master once that is merged.

Fixes #1124 and implements the logic proposed in https://github.com/GoogleContainerTools/skaffold/issues/1124#issuecomment-428201600. This required adding some more per-artifact-type logic that should probably be refactored into separate implementations in `docker.go`/`bazel.go`/`jib-*.go`. Let me know how this looks and whether or not I should do that refactoring. I will add tests once the structure is more finalized.

### DIFF
--- a/pkg/skaffold/build/local/jib_gradle.go
+++ b/pkg/skaffold/build/local/jib_gradle.go
@@ -35,11 +35,12 @@ func (b *Builder) buildJibGradle(ctx context.Context, out io.Writer, workspace s
 	cmd := jib.GradleCommand.CreateCommand(ctx, workspace, commandLine)
 	cmd.Stdout = out
 	cmd.Stderr = out
+
 	logrus.Infof("Building %s: %s, %v", workspace, cmd.Path, cmd.Args)
-	err := util.RunCmd(cmd)
-	if err != nil {
+	if err := util.RunCmd(cmd); err != nil {
 		return "", errors.Wrap(err, "gradle build failed")
 	}
+
 	return skaffoldImage, nil
 }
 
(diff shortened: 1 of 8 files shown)


dgageot:
I'm ok for this PR to use my commits

PR #1138, opened 2018-10-10
Move sync code to pkg/skaffold/sync/kubectl

### LINKED ISSUE #1136: move pkg/skaffold/kubernetes/sync* to pkg/skaffold/sync/kubectl
That is the correct place for the code

priyawadhwa:
This is the correct place for this code. Might require some
reorganization if we add other syncers, like rsync, but that shouldn't
be difficult.

Should fix #1136


r2d4:
and the linter issue

pkg/skaffold/sync/kubectl/kubectl.go:34:6: type name will be used as kubectl.KubectlSyncer by other packages, and that stutters; consider calling this Syncer (golint)

pkg/skaffold/sync/sync.go
```diff
@@ -108,3 +112,44 @@ func intersect(context string, syncMap map[string]string, files []string) (map[s
 	}
 	return ret, nil
 }
+
+func Perform(ctx context.Context, image string, files map[string]string, cmdFn func(context.Context, v1.Pod, v1.Container, string, string) *exec.Cmd) error {
```
r2d4: You can still leave this as unexported, I dont think we want it used outside the package.

priyawadhwa: so the `Sync` function in the kubectl package calls it, so I think it has to be exported?

r2d4: doh you are correct, LGTM then

PR #1139, opened 2018-10-10
Fix kaniko default behaviour

priyawadhwa:
If buildcontext isn't specified or localdir is specified, return localdir as the source
context. Otherwise, return a GCS bucket. If no value is passed in for
the bucket, one will be inferred from the project ID.

### DIFF
--- a/pkg/skaffold/build/kaniko/sources/sources.go
+++ b/pkg/skaffold/build/kaniko/sources/sources.go
@@ -35,7 +35,7 @@ type BuildContextSource interface {
 
 // Retrieve returns the correct build context based on the config
 func Retrieve(cfg *latest.KanikoBuild) (BuildContextSource, error) {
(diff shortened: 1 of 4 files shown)


pkg/skaffold/build/kaniko/sources/sources.go
```diff
@@ -35,7 +35,7 @@ type BuildContextSource interface {
 
 // Retrieve returns the correct build context based on the config
 func Retrieve(cfg *latest.KanikoBuild) (BuildContextSource, error) {
-	if cfg.BuildContext.LocalDir != nil {
+	if cfg.BuildContext == nil || cfg.BuildContext.LocalDir != nil {
```
dgageot: Setting default config should be done in https://github.com/GoogleContainerTools/skaffold/blob/e73775c0304872600bf6e7c643fdbdae715ed5f8/pkg/skaffold/schema/latest/defaults.go#L39 so that `BuildContext` is never nil

PR #1155, opened 2018-10-13
Canonicalizes paths when attempting to compensate for odd `mvnw` working-directory echo bug

briandealwis:
There is an odd bug in some `mvnw` wrappers (particularly the one generated by the Spring Initializr) where it prints out the current directory name.  This bug affects the Scaffold-Jib integration as Skaffold requests `mvnw jib:_skaffold-files`.  For example:
```
$ pwd
/Users/bsd/Projects/skaffold-jib/spring-boot-logging

$ ./mvnw -q jib:_skaffold-files
/Users/bsd/Projects/skaffold-jib/spring-boot-logging
/Users/bsd/Manumitting/Projects/skaffold-jib/spring-boot-logging/pom.xml
/Users/bsd/Manumitting/Projects/skaffold-jib/spring-boot-logging/src/main/java
/Users/bsd/Manumitting/Projects/skaffold-jib/spring-boot-logging/src/main/resources
/Users/bsd/Manumitting/Projects/skaffold-jib/spring-boot-logging/src/main/resources
/Users/bsd/Manumitting/Projects/skaffold-jib/spring-boot-logging/src/main/jib
```
A consequence is that Skaffold then monitors the entire source directory and picks up files from the build, and continuously triggers new builds.

To compensate, the Scaffold-Jib dependencies parser [compares each path to the workspace directory](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/jib/jib.go#L49).  But there are two problems:
   1. The workspace directory is the `context` field value and is usually relative to the current working directory, whereas the files reported by `jib:_skaffold-files` are absolute.
   2. I have a symlink from `/Users/bsd/Projects/skaffold-jib` to `/Users/bsd/Manumitting/Projects/skaffold-jib`. So just using `filepath.Abs` is insufficient as it does not expand symlinks.

This patch adds a `util.Canonical(path)` method that attempts to expand a provided path, and uses this new method to compare canonicalized paths.

PR #1167, opened 2018-10-15
Use os.SameFile() to check for mvnw working-dir echo bug

briandealwis:
Alternative implementation to workaround the `mvnw` bug described in #1155.  Instead of canonicalizing each file to compare with the working directory, use `os.SameFile()` to compare the `os.Stat()` structures.  Since we were obtaining the stat structures already, this should have minimal overhead.

### DIFF
--- a/pkg/skaffold/jib/jib.go
+++ b/pkg/skaffold/jib/jib.go
@@ -35,18 +35,17 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


briandealwis:
Turns out the mvnw.cmd for Windows outputs `Found "<maven-wrapper.jar path>"`

https://github.com/takari/maven-wrapper/blob/4096f68f48873e98af8376a30167afec4946d37d/mvnw.cmd#L129-L138

briandealwis:
takari/maven-wrapper#115

Issue #1135, opened 2018-10-10
skaffold.yaml schema generation


etanshaul:
This is to start the discussion around generating a schema for `skaffold.yaml` consumable by other tools like IDEs and editors.

In [google-container-tools-intellij](https://github.com/GoogleContainerTools/google-container-tools-intellij) we are building out editing support for Skaffold. This includes things autocomplete, file inspections etc. To do this in a robust way (keeping up with schema changes and so on), we'd like to have a versioned schema that ideally is auto-generated, possibly using a Go struct to JSON schema generator.

As an example, the IntelliJ Kubernetes plugin currently provides support for editing Kubernetes manifests. It accomplishes this by consuming the published [k8s openapi-spec](https://github.com/kubernetes/kubernetes/tree/master/api/openapi-spec). Within this spec, are definitions (in the form of JSON schema) such as the following Deployment schema:

```
   "io.k8s.api.apps.v1.Deployment": {
    "description": "Deployment enables declarative updates for Pods and ReplicaSets.",
    "properties": {
     "apiVersion": {
      "description": "APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources",
      "type": "string"
     },
     "kind": {
      "description": "Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds",
      "type": "string"
     },
     "metadata": {
      "description": "Standard object metadata.",
      "$ref": "#/definitions/io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta"
     },
     "spec": {
      "description": "Specification of the desired behavior of the Deployment.",
      "$ref": "#/definitions/io.k8s.api.apps.v1.DeploymentSpec"
     },
     "status": {
      "description": "Most recently observed status of the Deployment.",
      "$ref": "#/definitions/io.k8s.api.apps.v1.DeploymentStatus"
     }
    },
    "x-kubernetes-group-version-kind": [
     {
      "group": "apps",
      "kind": "Deployment",
      "version": "v1"
     }
    ]
   }
```

[...]

PR #1188, opened 2018-10-22
[wip] openapi experiment using kube-openapi

balopat:
DO NOT MERGE.

An experiment generating openapi spec in context of https://github.com/GoogleContainerTools/skaffold/issues/1135. 

This introduces a `make out/skaffold-schema.json` target that generates an OpenAPI schema for the latest Skaffold config. It uses https://github.com/kubernetes/kube-openapi to generate the schema.

Key design issues to discuss: 

* kube-openapi 
  * kube-openapi generates a go file with the OpenApi specs first and then it is used to generate the schema - this is a bit painful as the go file (pkg/skaffold/schema/openapi/build.go) needs a placeholder to compile, but I don't think we should check in the generated go file, so after generation it needs to be reverted (see Makefile target)
  * kube-openapi forces json fieldName to be the same as the go struct field name (https://github.com/kubernetes/kube-openapi/blob/0cf8f7e6ed1d2e3d47d02e3b6e559369af24d803/pkg/generators/rules/names_match.go#L54-L76) - this caused a lot of renames - that also lead to some odd failures with the yaml marshalling/unmarshalling
  * I vendored in kube-openapi but that should be fine 
  * I'm not sure how to generate one-of tags yet with kube-openapi
  * kube-openapi only uses a subset of OpenAPI that kubernetes uses - however it does not depend on k8s code as such - this could be a good thing if we want to stay aligned to the k8s toolings
* distributing the schema - we can generate a schema and publish it to a well known location, e.g. our bucket gs://skaffold/schema/v1alpha1/schema.json,  gs://skaffold/schema/v1alpha2/schema.json, etc.


etanshaul:
thanks for putting this together so quickly!

Overall the schema is looking good. A couple preliminary observations when testing the output with a schama-based editing prototype in IntelliJ:

### The naming of the "top-level" element

By top level element, I mean the definition element `latest.[]` that references the top-level properties in `skaffold.yaml`, i.e. these properties:

`apiVersion`, `kind`, `build`, `deploy`, `profiles` ...

Currently, in the schema output you've produced, this top level element looks as follows:

`"latest.SkaffoldPipeline": { ... }`

As is, this doesn't play nicely with the IntelliJ Kubernetes plugin which expects this top level element to be named `Config`, matching the `kind: Config` definition in skaffold.yaml:

```
"latest.Config": { 
  properties {
   "apiVersion": ...
   "kind": ...
   "build": ...
   ...
  }
 }
```

I'm not sure if this is a convention (or part of the spec), but I'm looking into it.

### Quick documentation isn't working

Quick documentation refers to the doc snippets you can view inline in the code:
screen shot 2018-10-23 at 2 56 11 pm

These are pulled from the `description` element in the schema.

As is, the produced schema doesn't have these elements in the correct places (for the IntelliJ Kubernetes integration). For the top-level element, the config is expected as follows:

```
"latest.Config": { 
  "description": "This description is what appears on the quick-docs when opening them on the 'Config' element in `kind: Config` 
 }
```

for every other element, the description field is expected within the `properties element`:

```
...
"properties": {
          "apiVersion": {
            "type": "string"
          },
          "build": {
            "description": "**This is the description for the 'build' key**",
            "$ref": "#/definitions/latest.BuildConfig"
          },
          ...
}
...      
```

I suspect these things are simply due to the Kubernetes integration in IntelliJ "hard-coding" certain things to adhere to the k8s OpenApi spec.

etanshaul:
And so, I guess my follow up question to the observations above is: can the schema be updated to conform to what I describe (and I suppose by extension, the go structs as well)? (so far I've only been looking at the JSON schema output of this PR)

balopat:
just to capture our conversation: 
* description fields are fine and kube-openapi actually generates them from comments on the struct - which is great 
* if we switch to  `kind: SkaffoldPipeline` we are fine - I'm okay with that. @r2d4  @dgageot are you okay to switch to `kind: SkaffoldPipeline` instead of `kind: Config`?

balopat:
Today's update: 
- I was playing with adding a dummy path to make the Kubernetes plugin happy for IntelliJ - that should be okay but is starting to stretch the approach 
- I tried to use the plugin with the generated yaml, mostly fine, the biggest issue I have is with our  "inlined oneOf" type fields.

Inline oneOf fields: 
for example: `artifact.ArtifactType` on `Artifact`: 
image
- oneOf: there is no good way to represent oneOf in swagger 2.0. OpenAPI 3.0.0 does support oneOf fields but that is not aligned with kubernetes. Bottom line: probably we'll have to live with this "choose one" as a documented thing.
- inline: I am going to submit a proposal to move these type of fields to a specific field - maybe we save a couple of bytes for the user by having them inline, but it is very confusing to see a list of choices mixed with other properties. Candidates: Artifact.ArtifactType -> `artifact.type`, BuildConfig.BuildType -> `build.location`, `DeployConfig.DeployType` and `HelmImageStrategy.HelmImageConfig` can stay because there are no other fields mixed in!

etanshaul:
A couple things:

- The quick documentation (the feature that pops up a description of the key/value in the IDE) won't work for properties that don't have a `description` field. `artifacts` for example under `BuildConfig`:
image
```
"latest.BuildConfig": {
    "description": "BuildConfig contains all the configuration for the build steps",
    "properties": {
     "artifacts": {
      "type": "array",
      "items": {
       "$ref": "#/definitions/latest.Artifact"
      }
     ...
```
The `artfiacts` object would need a `description` field for this to work. 

- More of a nit, but regarding the `latest.*` pattern (e.g. `latest.TagPolicy`). The k8s style of doing this is `io.k8s.[api | apimachinery ..]`. I wonder if we would want to consider following this style for Skaffold. Something like: `com.google.skaffold.TagPolicy`.

balopat:
https://github.com/GoogleContainerTools/skaffold/issues/1203

balopat:
This is going to be a larger effort to include in a cleaner way but it's a good prototype for the design. Closing for now until we pick up this task again.

PR #1201, opened 2018-10-24
Add some code to webhook

priyawadhwa:
This commit adds validation for the pull request event, making sure that
the pull request is open and labeled with the docs-modifications label.
It also adds code to cleanup all deployments associated with a pull
request when a pull request is closed or merged.


pkg/webhook/labels/labels.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// retrieveLabel returns the key and value for a label associated with this PR number
+func retrieveLabel(prNumber int) (string, string) {
+	value := fmt.Sprintf("docs-controller-deployment-%d", prNumber)
+	return "deployment", value
+}
+
+// DocsLabelExists returns true if the label exists
+func DocsLabelExists(labels []*github.Label) bool {
```
balopat: nothing else but this function is used in this PR, or am I missing something?

balopat: oh my bad, I see Selector used in Cleanup. 
How about `GenerateLabelsFromPR`?

webhook/webhook.go
```diff
@@ -35,5 +42,43 @@ func main() {
 }
 
 func handleGithubEvent(w http.ResponseWriter, r *http.Request) {
-	// TODO (priyawadhwa@): Add logic to handle a Github event here
+	eventType := r.Header.Get("X-GitHub-Event")
```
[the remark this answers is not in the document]
balopat: it works for me :)

pkg/webhook/labels/labels.go
priyawadhwa: oh whoops, that isn't used in this PR but will be used to label deployments and services (which will be the next PR!)

webhook/webhook.go
balopat: `"X-GitHub-Event"` can this go to a constant?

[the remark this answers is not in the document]
priyawadhwa: Ahhh it's just me then :-1:

webhook/webhook.go
```diff
@@ -35,5 +42,43 @@ func main() {
...
+
+	if event.GetPullRequest() == nil {
+		return nil
+	}
+
+	// Make sure pull request is open
+	if event.GetPullRequest().GetMerged() {
+		return nil
+	}
+
```
balopat: we can combine the above into a single boolean expression ORed together

balopat: or maybe we can put debug statements - that could make these useful

webhook/webhook.go
```diff
@@ -35,5 +42,43 @@ func main() {
...
+	if event.GetAction() == constants.ClosedAction {
+		return kubernetes.CleanupDeployment(event)
+	}
+
+	// Only continue if the docs-modifications label was added
+	if event.GetAction() != constants.LabeledAction {
+		return nil
+	}
+
+	if event.GetPullRequest() == nil {
```
balopat: is this possible on a `PullRequestEvent`?

webhook/webhook.go
```diff
@@ -35,5 +42,43 @@ func main() {
...
+		return nil
+	}
+
+	// Make sure pull request is open
+	if event.GetPullRequest().GetMerged() {
```
balopat: would this for not merged but closed PRs evaluate to `false`?

priyawadhwa: I think so, but I actually don't think we need this check since we first check if a PR's `action="closed"`, which is true for both merged and closed PRs. I'll go ahead and remove it.

webhook/webhook.go
priyawadhwa: Just read the docs and it doesn't seem like it, I'll remove this check as well

webhook/webhook.go
balopat: Isn't `action = "closed"` when it's an actual "Close" event? 
I think we need this check, as we do want to make sure we only deploy for open PRs. 
Here's the (contrived) scenario I'm thinking that this would have gone through:
- a PR is closed  (but not merged!)
- someone puts a "documentation-preview" label on it 
we do not want to build it

balopat: My point is that I think we need to check for `IsOpen()` or something similar - like `! Merged && ! Closed ` :)

pkg/webhook/labels/labels.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, l := range labels {
+		if l != nil {
```
balopat: you can combine the two ifs into 1

priyawadhwa: done

webhook/webhook.go
priyawadhwa: for sure, done!

PR #1202, opened 2018-10-24
Add cloudbuild.yaml for releasing docs-controller-image

priyawadhwa:
This cloudbuild.yaml will release the docs-controller image. It first
gets the credentials for the docs cluster, builds the image tagged with
:commit_sha and :latest, and pushes both images. It then deletes the
docs-controller pod running in cluster, so that when the deployment
recreates it, the new changes are picked up.

### DIFF
--- a/deploy/webhook/cloudbuild.yaml
+++ b/deploy/webhook/cloudbuild.yaml
@@ -1,0 +1,41 @@
+steps:
+  # Get cluster credentials
+- name: gcr.io/cloud-builders/gcloud
+  args: 
+  - 'container'
+  - 'clusters'
+  - 'get-credentials'
+  - 'docs'
+  - '--zone'
+  - 'us-west2-a'
+  - '--project'
+  - 'k8s-skaffold'
+  # Build and push the new docs-controller image
+- name: gcr.io/cloud-builders/docker
+  args:
+    - 'build'
+    - '-t'
+    - 'gcr.io/k8s-skaffold/docs-controller:${COMMIT_SHA}'
+    - '-f'
+    - deploy/webhook/Dockerfile
+    - .
+- name: gcr.io/cloud-builders/docker
+  args:
+    - 'tag'
+    - 'gcr.io/k8s-skaffold/docs-controller:${COMMIT_SHA}'
+    - 'gcr.io/k8s-skaffold/docs-controller:latest'
+- name: gcr.io/cloud-builders/docker
+  args:
+    - 'push'
+    - 'gcr.io/k8s-skaffold/docs-controller:${COMMIT_SHA}'
+- name: gcr.io/cloud-builders/docker
+  args:
+    - 'push'
+    - 'gcr.io/k8s-skaffold/docs-controller:latest'
+  # Delete the current controller pod in cluster so that the changes are picked up
+- name: gcr.io/cloud-builders/kubectl
+  args:
+    - 'delete'
+    - 'pod'
+    - '--selector'
+    - 'run=docs'


deploy/webhook/cloudbuild.yaml
```diff
@@ -0,0 +1,41 @@
+steps:
+  # Get cluster credentials
+- name: gcr.io/cloud-builders/gcloud
```
r2d4: you could simplify this by using skaffold :)

r2d4:
```
steps:
- name: gcr.io/k8s-skaffold/skaffold:v0.16.0
  args: ['skaffold', 'run', '-f=skaffold.yaml']
```

priyawadhwa: thanks Matt! good call :)

deploy/webhook/cloudbuild.yaml
```diff
@@ -0,0 +1,31 @@
+steps:
+  # Get cluster credentials
+- name: gcr.io/k8s-skaffold/skaffold:v0.16.0
+  args:
```
r2d4: I don think you need this step, cloud build already has the permissions to deploy to clusters in the project

deploy/webhook/cloudbuild.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - '${COMMIT_SHA}'
+  # Build and deploy the image with the :latest tag
+- name: gcr.io/k8s-skaffold/skaffold:v0.16.0
+  args:
+  - 'skaffold'
+  - '-f'
+  - 'deploy/webhook/skaffold.yaml'
+  - 'run'
+  - '--tag'
+  - 'latest'
```
r2d4: Sorry I didn't follow the whole design, but what is :latest used for?

deploy/webhook/cloudbuild.yaml
priyawadhwa: when I tried with my own cluster this step was needed for some reason, but am happy to remove it for now and add it back in later if it doesn't work!

deploy/webhook/cloudbuild.yaml
priyawadhwa: I just realised that's left over from the previous design and not needed anymore! I'll remove it.

deploy/webhook/cloudbuild.yaml
r2d4: thats because we've given the cloud build service account cluster admin access explicitly. if you want it to work on you're own cluster in a different project it won't work

priyawadhwa: gotcha, thanks!

deploy/webhook/cloudbuild.yaml
```diff
@@ -0,0 +1,10 @@
+steps:
+  # Build and push the image with the :commit_sha tag
+- name: gcr.io/k8s-skaffold/skaffold:v0.16.0
+  args:
+  - 'skaffold'
+  - '-f'
+  - 'deploy/webhook/skaffold.yaml'
+  - 'run'
+  - '--tag'
+  - '${COMMIT_SHA}'
```
r2d4: one last nit: why don't you let skaffold compute the commit sha! thats what the gitTagger does and its the default

r2d4: this PR is becoming an advertisement for the power of skaffold :)

priyawadhwa: LOL solid i had no idea! hypeee

PR #1204, opened 2018-10-25
Improve runner code

dgageot:
Let's remove some duplication and simplify the runner's code.


dgageot:
ping @GoogleContainerTools/container-tools can you please take a look. I tried to remove a lot of duplication in the runner code. It was becoming a bit too complicated to maintain. 
It's now super simple yet but there should be less room for mistakes.

dgageot:
@r2d4 done in https://github.com/GoogleContainerTools/skaffold/pull/1255

pkg/skaffold/runner/runner.go
```diff
@@ -205,67 +204,68 @@ func getTagger(t latest.TagPolicy, customTag string) (tag.Tagger, error) {
...
+func (r *SkaffoldRunner) buildTestDeploy(ctx context.Context, out io.Writer, artifacts []*latest.Artifact) error {
+	bRes, err := r.Build(ctx, out, r.Tagger, artifacts)
+	if err != nil {
+		return errors.Wrap(err, "build failed")
 	}
 
-	imageList := kubernetes.NewImageList()
-	for _, b := range bRes {
-		imageList.Add(b.Tag)
+	if err = r.Test(ctx, out, bRes); err != nil {
```
r2d4: How about skip if test is nil and get rid of noop tester?

pkg/skaffold/build/prebuilt.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Labels are labels applied to deployed resources.
+func (b *prebuiltImagesBuilder) Labels() map[string]string {
+	return map[string]string{
+		constants.Labels.Builder: "pre-built",
+	}
+}
+
+func (b *prebuiltImagesBuilder) Build(ctx context.Context, out io.Writer, tagger tag.Tagger, artifacts []*latest.Artifact) ([]Artifact, error) {
+	tags := make(map[string]string)
```
r2d4: Since we’re already here, how about we generate the git tags if no images were supplied?

r2d4: We can also resolve the taggers from the yams if you want to be really correct (to make sure we can generate reproducibly)

pkg/skaffold/runner/runner.go
```diff
@@ -119,66 +116,68 @@ func NewForConfig(opts *config.SkaffoldOptions, cfg *latest.SkaffoldPipeline) (*
 		Syncer:       &kubectl.Syncer{},
 		opts:         opts,
 		watchFactory: watch.NewWatcher,
+		imageList:    kubernetes.NewImageList(),
 	}, nil
 }
 
-func getBuilder(cfg *latest.BuildConfig, kubeContext string) (build.Builder, error) {
+func getBuilder(cfg *latest.BuildConfig, kubeContext string, opts *config.SkaffoldOptions) (build.Builder, error) {
```
r2d4: Slightly weird but I get it

pkg/skaffold/runner/runner.go
r2d4: And the helper function

dgageot: `if tests is nil`? Do you mean we'd have to store the `SkaffoldPipeline` in the SkaffoldRunner instance?

pkg/skaffold/build/prebuilt.go
dgageot: Sorry @r2d4, I don't understand

pkg/skaffold/runner/runner.go
dgageot: yes, I know

pkg/skaffold/build/prebuilt.go
r2d4: If images isn't provided, try to seed it by using the passed in tagger to generate a tag for each artifact
```golang
if images == nil {
  for _, a := range artifacts { 
    tag, err := tag.GenerateFullyQualifiedTag(a.Image, nil)
    if err != nil // this is the case where we ended up with a tagger that needs some options passed in, fail
    images == append(images, tag)
}

r2d4: which would fix https://github.com/GoogleContainerTools/skaffold/issues/653

Issue #1012, opened 2018-09-21
Completion support for zsh


GeertJohan:
Currently, the `skaffold completion` subcommand ignores any arguments and always outputs bash completion. Since `skaffold completion zsh` didn't error, I expected it to return a valid zsh completion script only to find out it was running the bash completion script.

I thinks two seperate PR's can be created to fix this issue.

The first is very simple and just manages expectation. The shell argument to `skaffold completion SHELL` should be validated; if it's not "bash" then write an error and exit with non-zero exitcode. If maintainers agree with this I would like to create a PR for it. It's simple enough to make it in the next release.

The second PR would actually add zsh completion.
I've seen cobra supports very rudimentary Zsh now, just command structure. I guess for that reason both kubectl and minikube use a converter/wrapper from bash to zsh.

kubectl: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/completion.go
minikube: https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/completion.go

They both seem very similar. Probably we could just diff the files and pick one based on which is best for skaffold.

r2d4:
Agreed. Both the kubectl and minikube zsh completion should be same (except s/minikube/kubectl/) and we can just copy that over. We would absolutely be open to a PR for either, or both.

When I wrote the minikube completion, I used kubectl completion as the reference, and when I wrote the skaffold completion, I used minikube as the reference :)

PR #1206, opened 2018-10-26
Improve error handling for `completion` command

varunkashyap:
We currently only support bash completion scripts, however as raised as part (1) of #1012 this means we output bash completion script irrespective of what the user types as the shell.

This PR makes typing bash explicit and mandatory, introduces valid args and checks for length and presence of valid args


varunkashyap:
I signed it!

PR #1207, opened 2018-10-26
 feat: IBM Cloud Container Registry Build Support

bkuschel:
A new build type called "icr" allows build to be sent to the IBM Cloud Container Registry Build engine by either using an API key or a pre-authenticated CLI session.


balopat:
Hi, thank you for opening this! We can't take this at the moment, we have to think through our strategy around integration tests with cloud providers outside of Google and the collaboration and maintenance models around them. As soon as that is clarified we'll get back to you on this. Thank you for your understanding!

bkuschel:
I signed it!

PR #1208, opened 2018-10-26
cut v0.17.0

balopat:


balopat:
I think I messed up the versions, as I merged David's PR first and then ran the copy command. I'll open a new version of this PR.

PR #1209, opened 2018-10-26
Get credentials before releasing docs controller image

priyawadhwa:
The build trigger for releasing the docs-controller-image failed with this error:

`error retrieving kubernetes dynamic client: getting client config for
dynamic client: Error creating kubeConfig: invalid configuration: no
configuration has been provided`

Explicitly getting the credentials should generate the correct
kubeconfig for skaffold. I tested locally with cloud-build-local and
this should work now.


r2d4:
Is this running in the same cluster it's deploying to? You can use the in cluster config then.

PR #1226, opened 2018-10-30
Add hugo and git to webhook image

priyawadhwa:
Adding hugo and git to the skaffold image will allow the webhook to
deploy it and:

1. Use git to clone a user's branch with docs changes
2. Use hugo to set up a static website which will show those changes


r2d4:
The skaffold image shouldn't have hugo. It's made to be minimal and only have the dependencies for skaffold

priyawadhwa:
For sure, moved to the webhook image

PR #1233, opened 2018-11-03
Allow flags to be passed to kustomize

vreon:
:wave:

`kustomize build` accepts a `--transformer-config` argument, but skaffold didn't expose a mechanism for adding flags there.

`KustomizeDeploy` already had a `flags` field that added flags to kubectl invocations, so I renamed it to `kubectlFlags` and co-opted the `flags` field for kustomize flags; not sure if that's kosher.

### DIFF
--- a/integration/examples/annotated-skaffold.yaml
+++ b/integration/examples/annotated-skaffold.yaml
@@ -161,9 +161,14 @@ deploy:
  # kustomize:
     # path: .
     # kustomize deploys manifests with kubectl.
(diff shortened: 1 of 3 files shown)


mxey:
FYI: [Kustomize is moving from a flag to a setting in kustomization.yaml](https://github.com/kubernetes-sigs/kustomize/pull/544)

vreon:
@mxey That's great news, thank you!

PR #1248, opened 2018-11-07
Fix jib errors on ctrl-c

### LINKED ISSUE #1247: Jib errors when dev mode is interrupted with ctlr-c
There are still errors when `skaffold dev` is interrupted with ctrl-c.
 + Sometimes because the gradle or maven command is killed with a signal.
 + Sometimes because it fails because the context is already cancelled.

dgageot:
Fixes #1247


pkg/skaffold/util/process.go
```diff
@@ -26,13 +26,14 @@ import (
 // IsTerminatedError returns true if the error is type exec.ExitError and the corresponding process was terminated by SIGTERM
 // This error is given when a exec.Command is ran and terminated with a SIGTERM.
```
briandealwis: I forgot to fix up the comments.

pkg/skaffold/util/process.go
```diff
@@ -26,13 +26,14 @@ import (
...
-	exitError, ok := err.(*exec.ExitError)
-	if !ok {
-		return false
+	cause := errors.Cause(err)
+
+	switch exitErr := cause.(type) {
+	case *exec.ExitError:
+		status := exitErr.Sys().(syscall.WaitStatus)
+
+		return status.Signaled()
```
briandealwis: I don't think SIGSEGV, SIGBUS, etc. should be masked.

pkg/skaffold/runner/runner.go
```diff
@@ -465,6 +465,12 @@ func DependenciesForArtifact(ctx context.Context, a *latest.Artifact) ([]string,
 	}
 
 	if err != nil {
+		// if the context was cancelled act as if all is well
+		// TODO(dgageot): this should be even higher in the call chain.
+		if ctx.Err() == context.Canceled {
```
briandealwis: This will mask *all* errors.  Perhaps you could log the error at a DEBUG level?

dgageot: Well, the idea is to not show annoying errors when the user presses ctrl-c. No matter what the error is

briandealwis: That's why I suggested DEBUG :-)  Just in case those errors are due to something other than SIGTERM/KILL.  Anyways, it's not a blocked from me.

dgageot: I'm going to add debug logs

PR #1250, opened 2018-11-07
Port forward multiple ports

### LINKED ISSUE #1249: Port-forwarding works with only one port
It seems that the code that loops on ports to be forwarded, blocks on the first port.
@r2d4 can you confirm?

dgageot:
Fixes #1249

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -748,6 +748,14 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:b521f10a2d8fa85c04a8ef4e62f2d1e14d303599a55d64dabf9f5a02f84d35eb"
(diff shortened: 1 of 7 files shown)


r2d4:
I think we still want to log the error? Maybe wrap this in a sync Group and return an error at the end if any of them errored (but still port forward?)

dgageot:
@r2d4 Good point! Should be ok now

r2d4:
So if we port forward some but then fail and return an error, I’m not sure we’ll clean up the rest properly. 

I guess this was also an issue before when the first port forward in a pod succeeded and the second one failed. 

Let’s merge for now and keep an eye out for any forwards that don’t get cleaned up properly

PR #1258, opened 2018-11-08
cut v0.18.0

balopat:


hack/release_notes/listpullreqs.go
```diff
@@ -61,7 +61,7 @@ func printPullRequests() {
 	fmt.Println(fmt.Sprintf("Collecting pull request that were merged since the last release: %s (%s)", *releases[0].TagName, lastReleaseTime))
 
 	listSize := 1
-	for page := 0; listSize > 0; page++ {
+	for page := 1; listSize > 0; page++ {
```
r2d4: ?

balopat: it is fixing a bug that was duplicating the list of the PRs, technically this could go in a separate PR, but I figured we can let this sneak in :)

PR #1270, opened 2018-11-14
Support system's LDFLAGS, make compilation reproducible

maximbaz:
* Supporting `$LDFLAGS` provided by the build system for example allows [enabling RELRO](https://medium.com/@HockeyInJune/relro-relocation-read-only-c8d0933faef3).
* Using `-trimpath` in `-gcflags` and `-asmflags` makes your application [reproducible](https://reproducible-builds.org/).


maximbaz:
Signed CLA

PR #1287, opened 2018-11-16
Add caching to kaniko builder

### LINKED ISSUE #1275: Kaniko caching
Skaffold should auto-configure Kaniko to leverage caching.

priyawadhwa:
I updated the version of the kaniko image so that I could add caching.
To use caching, the user has to specify 'cache' in their skaffold.yaml,
with an optional field for 'repo', which is where cached layers will be
stored.

Fixes #1275

### DIFF
--- a/pkg/skaffold/constants/constants.go
+++ b/pkg/skaffold/constants/constants.go
@@ -46,7 +46,7 @@ const (
 
 	DefaultKustomizationPath = "."
 
-	DefaultKanikoImage             = "gcr.io/kaniko-project/executor:v0.4.0@sha256:0bbaa4859eec9796d32ab45e6c1627562dbc7796e40450295b9604cd3f4197af"
+	DefaultKanikoImage             = "gcr.io/kaniko-project/executor@sha256:434bbb1d998ba1bd8ebc04c90d93afa859fd5c7ff93326bca9f6e7da0d6277ff"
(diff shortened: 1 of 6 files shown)


priyawadhwa:
For sure, should be an easy change.

pkg/skaffold/constants/constants.go
```diff
@@ -46,7 +46,7 @@ const (
 
 	DefaultKustomizationPath = "."
 
-	DefaultKanikoImage             = "gcr.io/kaniko-project/executor:v0.4.0@sha256:0bbaa4859eec9796d32ab45e6c1627562dbc7796e40450295b9604cd3f4197af"
+	DefaultKanikoImage             = "gcr.io/kaniko-project/executor@sha256:434bbb1d998ba1bd8ebc04c90d93afa859fd5c7ff93326bca9f6e7da0d6277ff"
```
dgageot: Can you leave the version here, in addition to the sha256? It makes it easier to find which version we are using.

priyawadhwa: kaniko actually hasn't released a version with the new caching stuff yet -- I was going to come back to this and add the new version in once that happened.

priyawadhwa: I'll open an issue for it!

PR #1289, opened 2018-11-17
Allow passing arguments to bazel build

jamielennox:
To make a container capable of being used by skaffold I need to be able
to provide additional parameters on the command line. Add an args:
element to the bazel build configuration that is a list of strings that
will be passed to the build command.


jamielennox:
My specific requirement here is that I'm running bazel on OSX with a golang codebase. There are a couple of problems to do with the toolchain, gazelle, that don't resolve dependencies correctly when simply specify cross compile parameters. The advice from bazel is to pass "--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64" to the build process. 

This will provide a way to set these parameters.

PR #1332, opened 2018-11-28
Apply default values upgraded configurations

dgageot:
Default values need to be set after upgrade, not before. In fact, default values are only needed to simplify the code of the runner, so that it doesn't have to compensate missing fields.

I used this opportunity to:
 + improve the code and coverage for `skaffold fix`
 + Remove duplication in code and tests

### DIFF
--- a/cmd/skaffold/app/cmd/init.go
+++ b/cmd/skaffold/app/cmd/init.go
@@ -106,7 +106,7 @@ func doInit(out io.Writer) error {
 
 	for _, file := range potentialConfigs {
 		if !force {
-			config, err := schema.ParseConfig(file, true)
+			config, err := schema.ParseConfig(file, true, false)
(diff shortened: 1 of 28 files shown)


cmd/skaffold/app/cmd/init.go
```diff
@@ -106,7 +106,7 @@ func doInit(out io.Writer) error {
 
 	for _, file := range potentialConfigs {
 		if !force {
-			config, err := schema.ParseConfig(file, true)
+			config, err := schema.ParseConfig(file, false)
```
balopat: I'm not sure I follow this one. Why wouldn't we want to upgrade?

dgageot: In fact, here, we don't really care. We just want to check that those are valid config files.

PR #1357, opened 2018-12-03
Add gcb profile to example

halvards:
This updates the Skaffold manifest to allow new users to follow
the Quick start for GKE document.


priyawadhwa:
Hey @halvards, thanks for your contribution! This change will need to be made in [skaffold/integration/examples/getting-started/skaffold.yaml](https://github.com/GoogleContainerTools/skaffold/blob/master/integration/examples/getting-started/skaffold.yaml) as well because of how we handle config changes.

PR #1364, opened 2018-12-04
Cancellable builds

### LINKED ISSUE #916: skaffold dev: not cancelling the in-flight build on change
(skaffold v0.12.0)

**Observed behavior:** 

- During `skaffold dev`, when I change+save a file, docker build gets kicked off.
- while it's going on, when I change+save a file again, the previous build still goes on, skaffold doesn't cancel/stop the build.

This is significantly noticeable in the user experience.

**Expected:** 
- skaffold cancels the ongoing build when it detects a change to the relevant workspace
- building the old context is no longer useful to me, so I probably should not be waiting for that build to finish as I won't use it.

balopat:
Fixes #916. Well, after the first build. The first build is still non-cancellable, that can come in another PR - needs more rework to get that working as we'll need to start the watcher before the first build starts.

- [X] quick and dirty prototype
- [X] write integration test 
- [X] need to figure out why the unit test is failing when building is done in a go func
- [x] need to scope cancelling to the actual artifacts that need rebuild 
- [x] need to garbage collect finished child contexts


balopat:
@dgageot ready for review

balopat:
That was just a warning - but I removed it for "cancel", as it's unnecessary noise, agreed. Maybe we can completely remove it as well - the user can't really do anything if this happens - it's more of an assertion. PTAL.

dgageot:
@balopat I've tried again the feature. I have a few problems:\
 + Logs are not correctly muted. Logs used to ba muted during a build because it's annoying to see the logs mixed with build commands.
 + `Watching for changes every 500ms...` message used to be displayed just after a successful deploy. Now, it seems to be displayed just before a build.
 + I still see warnings when a build is interrupted and when the subsequent build finished. That's annoying because, as a user, I can do anything about it.

balopat:
Thank you for the detailed feedback, I'll have a look at them! This might sound crazy but - as logging is an important part of our UX - we might want to write tests for them to protect the functionality? As I don't use skaffold (yet) that often, I completely missed these points - same would go for a new contributor. What do you think?

balopat:
I'll rework this when I have the time :) Closing for now...

PR #1374, opened 2018-12-07
Local builders return a digest

dgageot:
Each builder is in charge of returning the digest of what it built. 
`digest` is still `digest|imageID` though.


dgageot:
@balopat Thank you, I put a `wip` label because I need to triple check if the Bazel builder still works

PR #1390, opened 2018-12-12
Create directory before kubectl cp

### LINKED ISSUE #1365: Sync does not sync new directories
### Expected behavior
- Creating a new directory when sync is enabled should create the directory in the target container
- (`Syncing 1 files for example:xxxxx` should appear in the console when a new directory is created)
- This is a problem if you subsequently create a file inside the directory. It will not sync the file because the target directory does not exist, resulting in an error

### Actual behavior
A new directory is not created in the target container.

### Information

- Skaffold version: v0.19.0
- Operating system: Ubuntu 18.04.1

`kubectl cp` supports copying a directory but it appears as though Skaffold skips new directories. I believe the problem originates in `pkg/skaffold/docker/parse.go` in `GetDependencies`. 

It looks like execution for a directory executes `case mode.IsDir():` at: https://github.com/GoogleContainerTools/skaffold/blob/v0.19.0/pkg/skaffold/docker/parse.go#L305

and `if info.IsDir() {` at: https://github.com/GoogleContainerTools/skaffold/blob/v0.19.0/pkg/skaffold/docker/parse.go#L323

Note that `files` is never modified in this case. 

EDIT: Even though `kubectl cp` will copy a directory, after further testing this won't be a good solution. It is fine when a directory is initially created, however as new files are added to it, the directories modification time is updated. This would trigger a resync of the directory, and the resulting `kubectl cp` of the directory would (unnecessarily) copy all contents as well...  A possible solution could be to use `kubectl exec` and `mkdir`. 

#### skaffold.yaml
```yaml
apiVersion: skaffold/v1beta1
kind: Config
build:
  artifacts:
    - image: example
      context: .
      docker:
        dockerfile: ./Dockerfile
      sync:
        '**': /root

deploy:
  kubectl:
    manifests:
      - "example.yaml"
```

#### Dockerfile
```
FROM ubuntu:18.04

COPY . /root

CMD exec /bin/bash -c "trap : TERM INT; sleep infinity & wait"
```

#### example.yaml
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: example
  labels:
    component: example
spec:
  containers:
    - name: example
      image: example
```

1. `skaffold dev`
2. `mkdir this-doesnt-sync`

priyawadhwa:
This PR should fix a bug where syncing files into new directories wasn't working because those directories didn't exist. This change will create the directory in the container via kubectl exec. This change also requires that the container has 'mkdir' in it, but I couldn't think of a better way to get around this.

Also adds an integration test for file sync.

fix #1365


pkg/skaffold/sync/kubectl/kubectl.go
```diff
@@ -46,10 +47,14 @@ func (k *Syncer) Sync(ctx context.Context, s *sync.Item) error {
...
+func deleteFileFn(ctx context.Context, pod v1.Pod, container v1.Container, src, dst string) []*exec.Cmd {
+	delete := exec.CommandContext(ctx, "kubectl", "exec", pod.Name, "--namespace", pod.Namespace, "-c", container.Name, "--", "rm", "-rf", dst)
+	return []*exec.Cmd{delete}
 }
 
-func copyFileFn(ctx context.Context, pod v1.Pod, container v1.Container, src, dst string) *exec.Cmd {
-	return exec.CommandContext(ctx, "kubectl", "cp", src, fmt.Sprintf("%s/%s:%s", pod.Namespace, pod.Name, dst), "-c", container.Name)
+func copyFileFn(ctx context.Context, pod v1.Pod, container v1.Container, src, dst string) []*exec.Cmd {
+	dir := filepath.Dir(dst)
+	makeDir := exec.CommandContext(ctx, "kubectl", "exec", pod.Name, "-c", container.Name, "-n", pod.Namespace, "--", "mkdir", "-p", dir)
```
r2d4: I'd hate to nearly double the sync time by running the mkdir command every file synced. What if we minimized the time we run mkdir at the cost of a bit more complexity in the code?

priyawadhwa: That's a good point, I changed the code to keep track of directories and only run the command when it comes across a new one!

PR #1398, opened 2018-12-18
Improve runner unit tests

dgageot:
Test were broken and wouldn't test what we thought they'd test.

### DIFF
--- a/pkg/skaffold/kubernetes/log.go
+++ b/pkg/skaffold/kubernetes/log.go
@@ -29,7 +29,7 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/watch"
 )
 
@@ -116,7 +116,9 @@ func (a *LogAggregator) Start(ctx context.Context) error {
 
 // Stop stops the logger.
 func (a *LogAggregator) Stop() {
-	a.cancel()
+	if a.cancel != nil {
+		a.cancel()
+	}
 }
 
 func sinceSeconds(d time.Duration) int64 {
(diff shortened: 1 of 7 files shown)


dgageot:
ping @balopat This is the first iteration on fixing/adding tests on the Runner.

I have more to come to test the sync. There's a lot of issues with this piece of code

pkg/skaffold/kubernetes/log.go
```diff
@@ -116,7 +116,9 @@ func (a *LogAggregator) Start(ctx context.Context) error {
 
 // Stop stops the logger.
 func (a *LogAggregator) Stop() {
-	a.cancel()
+	if a.cancel != nil {
```
balopat: now that we're touching this: `a.cancel` can be `nil` when it's not yet started, right? 
Also, what happens when we call Stop() twice? I.e. the cancelFunc called twice? 
What if instead of `nil` the "default" value of the cancel is a "noop" func, i.e `cancel: func() {}` in NewLogAggregator - and then still, `Stop`will have to remember to set it back to `func() {}` - this way we don't have to check for `nil`.
Alternatively we can keep it as it is and have `a.cancel=nil` represent the Stopped state -  but then we have to set `a.cancel = nil` here. WDYT?

pkg/skaffold/runner/dev.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/sync"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/watch"
+)
+
+// ErrorConfigurationChanged is a special error that's returned when the skaffold configuration was changed.
+var ErrorConfigurationChanged = errors.New("configuration changed")
+
+// Dev watches for changes and runs the skaffold build and deploy
+// pipeline until interrrupted by the user.
```
balopat:
```suggestion
// pipeline until interrupted by the user.
```

pkg/skaffold/kubernetes/log.go
dgageot: Let me handle that in another PR. Current state is that `cancel` can be called multiple time, which is a noop passed the first time.

PR #1401, opened 2018-12-18
Change getting started to emojinetes

bobcatfish:
Instead of displaying "hello world", with this change the example will cycle through emojis and a rainbow of terminal colors.

Not sure if we should make this the default example, since the terminal escape colors will look terrible if looking at logs directly (e.g. piping into a file, offloading them somewhere), so feel free to reject this :D It looks pretty fun though:

image

_This is based on the Google skaffold OSS ecosystem booth demo from Kubecon Seattle 2018_

### DIFF
--- a/examples/getting-started/main.go
+++ b/examples/getting-started/main.go
@@ -5,9 +5,30 @@ import (
 	"time"
 )
 
+const (
+	startingEmoji = '🐰'
+	maxEmoji      = 200
+
+	colorReset = "\u001b[0m"
+)
+
+var colors = []string{
+	"\u001b[31;1m", // red bold
+	"\u001b[33;1m", // yellow bold
+	"\u001b[32;1m", // green bold
+	"\u001b[34;1m", // blue bold
+	"\u001b[36;1m", // cyan bold
+	"\u001b[35;1m", // magenta bold
+}
+
 func main() {
 	for {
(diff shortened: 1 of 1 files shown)


r2d4:
One thing that might be cool is to add more features that don't require configuration options to the getting-started config like port-forward

bobcatfish:
> @bobcatfish I'm afraid @vdemeester influenced you in using emoticons everywhere

@dgageot YOU CAUGHT ME :rofl::rofl::rofl: DEFINITELY this would not have happened without @vdemeester 's influence

bobcatfish:
> I dont know how these are going to render in powershell man_shrugging. Maybe not as the main example but still checked in is my take +1

That's a great question @r2d4 !! I'm not sure either - the colors I'm _assuming_ should work, since I think skaffold itself is using the same escape codes, but maybe not the emojis :thinking:

balopat:
windows screenshot for terminal/powershell:

image

bobcatfish:
Just ack-ing your requests @balopat - I haven't gotten a chance to get to them yet, hoping I will! But feel free to close this if I take too long, I can always open a new PR :)

balopat:
Sounds good, please open a new one whenever you're ready, I'll close this for now to reduce our queue a little bit :)

PR #1406, opened 2018-12-20
Add some tests for Sync

dgageot:


dgageot:
The idea here is I think the opposite of what you’re saying. The TestBench does not produce a MockRunner, it creates an actual runner with mocked dependencies. To test the code of the Runner, we test need a TestBuilder, a TestTester, a TestSyncer and a TestDeployer. The TestBench implements all those deps, with a shared state that facilitates the assertions. It’s teally a bench on which you plug a runner and collect information. Is it clearer?

balopat:
I see, true, calling it a `MockRunner` would miss the point as well. Let's go with it, I just haven't seen this before - maybe some comments can help clarifying the responsibility of `TestBench` that it's a facade for all of the mocked collaborators of Runner.

balopat:
Not even a facade - it's a "union mock" of all collaborators :)

PR #1421, opened 2018-12-23
added poc way of dealing with docker auth issues

robbert229:
This is meant as an example of what I as a user need to use kaniko. I need a way of using regular docker auth helpers with kaniko.

### DIFF
--- a/pkg/skaffold/build/kaniko/sources/sources.go
+++ b/pkg/skaffold/build/kaniko/sources/sources.go
@@ -70,6 +70,10 @@ func podTemplate(cfg *latest.KanikoBuild, args []string) *v1.Pod {
 							Name:      constants.DefaultKanikoSecretName,
 							MountPath: "/secret",
 						},
+						{
+							Name:      "regcred",
+							MountPath: "/root/",
+						},
 					},
 				},
 			},
@@ -79,6 +83,19 @@ func podTemplate(cfg *latest.KanikoBuild, args []string) *v1.Pod {
 				VolumeSource: v1.VolumeSource{
 					Secret: &v1.SecretVolumeSource{
 						SecretName: cfg.PullSecretName,
+					},
+				},
+			}, {
+				Name: "regcred",
+				VolumeSource: v1.VolumeSource{
+					Secret: &v1.SecretVolumeSource{
+						SecretName: cfg.PullSecretName,
+						Items: []v1.KeyToPath{
+							{
+								Key:  constants.DefaultKanikoSecretName,
+								Path: ".docker/config.json",
+							},
+						},
 					},
 				},
 			}},


pkg/skaffold/build/kaniko/sources/sources.go
```diff
@@ -70,6 +70,10 @@ func podTemplate(cfg *latest.KanikoBuild, args []string) *v1.Pod {
 							Name:      constants.DefaultKanikoSecretName,
 							MountPath: "/secret",
 						},
```
robbert229: So why not overload the credentials and mount them as the normal creds in `/secret` as well as in `/root/.docker/config.json`?

In my case this allowed me to set the `kanikoSecret: ~/.docker/config.json`  inside of the kaniko builder configuration.

balopat: Interesting idea. My thoughts: 
1.) However there is no way _both_ can work. What if you want to build two artifacts, one with a GCR repo, the other one using Docker? The first would want to use the `GOOGLE_APPLICATION_CREDENTIALS` env var, the latter would use the config.json. 
2.) based on https://github.com/GoogleContainerTools/kaniko#pushing-to-amazon-ecr I see that it should be /kaniko/.docker/config.json and could be a configmap... @priyawadhwa can you advise? 
3.) What if we introduce a `mountDockerConfig` boolean flag, by default `false` and a `dockerConfigPath:` string  flag pointing to `~/.docker/config.json` by default instead?

priyawadhwa: Yup option 3 SGTM -- basically skaffold.yaml should have separate options for registry specific credentials and the dockerconfig in case both are needed, as @balopat said. We could create a configmap or another secret (probably better) for the dockerconfig and mount it into `/kaniko/.docker/config.json` (that's where kaniko [expects](https://github.com/GoogleContainerTools/kaniko/blob/master/deploy/Dockerfile#L38) to find it)

garethjevans:
Hi, just wondering if anyone is looking at an implementation for option 3 - am happy to help if I can.

pkg/skaffold/build/kaniko/sources/sources.go
robbert229: Sorry for the delay. I wasn't intending for the configuration to be reused, I only did it to simplify the work required to test the concept. Realistically option 3 is probably be the best.

robbert229:
@garethjevans I would appreciate it if you found the time to take a swing at it! I am just a little too busy right now. I really just opened this pull request to start the conversation

garethjevans:
@robbert229 no problem, I'll take a look at it tomorrow, see what I can come up with.

garethjevans:
@robbert229 just to give you a heads up on where i'm up to, I'm working on a branch (https://github.com/garethjevans/skaffold/tree/docker-creds), pushing the image to dockerhub works nicely, but using skaffold to tag the image is currently failing due to a PUT request with a zero length body when adding the manifest.  I'm still trying to track this down.

garethjevans:
@robbert229 implementation is in #1466

PR #1431, opened 2018-12-27
Fix localDir context for Kaniko on Windows

cedrickring:
When using the `localDir` context on Windows, the `kubectl cp` command fails when trying to copy files from `C:\Users\<user>\AppData\Local\Temp` (`C` is interpreted as the source file, `\Users\...` is interpreted as the destination file).

This solution is basically a `kubectl cp` command without the `src:dest` limitation.


pkg/skaffold/build/kaniko/sources/localdir.go
```diff
@@ -24,7 +24,7 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
cedrickring: Can someone explain me why `golangci-lint` wants this? Goland removed this since this is an unnecessary import alias. Running `goimports` didn't fix this and I had to add it back manually.

balopat: So, did a bit of research on this: 
1.) goimports _should_ fix it - that's why `golangcli-lint` detects it
2.) goimports generates an alias to packages where the "real package name" defined in the go file on disk does not equal to the "basic package name" (https://github.com/golang/tools/blob/master/imports/fix.go#L402). Now the "basic package name" is  usually the last folder (base) in the import path, except that if the path ends in foo/vN, it assumes the package name is foo (see [importPathToBasic](https://github.com/golang/tools/blob/d00ac6d27372a4273825635281f2dc360d4be563/imports/fix.go#L672-L674)) 

Now - to keep things simple we rely on tools like goimports. 
Here it seems that the k8s code is deviating from the assumed format of goimports - i.e. the package name is not `core` under the `v1` folder...which creates the above issue.

dgageot: This is fixed by very recent versions of goimports. You might want to update with `go get -u golang.org/x/tools/cmd/goimports`

dgageot:
@cedrickring I wonder if there's a way of escaping the path for `kubectl cp` to understand. Another solution would be to change the command's working directory to the folder containing the tarball.

pkg/skaffold/build/kaniko/sources/localdir.go
cedrickring: For some reason running `go get golang.org/x/tools/cmd/goimports` without the `-u` flag doesn't install the latest version if its not installed yet. But I think it would be nice to have a `Makefile` target which goimports/gofmt's all files (including checking for the latest version of `goimports` 😉). 

@balopat Thanks for the explanation 👍

pkg/skaffold/kubernetes/exec.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+//Exec contains all required information to execute a command in a container
+type Exec struct {
```
priyawadhwa: Is it possible to reuse [Run](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/exec/exec.go#L268) to execute commands in the container and use the [ExecOptions](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/exec/exec.go#L137) type, instead of creating a new `Exec` type?

cedrickring: I tried that before but didn't get the RESTConfig right and tried to do it that way. Though I did some more debugging in the `kubectl` code and maybe found a way to get the rest config to work. I'll push another version tomorrow where I'll use the `ExecOptions`.

cedrickring: So I did manage to get everything to work now. Seems like the k8s api has changed much since 1.11. I'd consider upgrading all k8s dependencies to 1.12 as there have been many significant and good changes to the api...

cedrickring:
Yes I had the same thought when implementing it. But I'd rather go with running a `kubectl exec` with the file as stdin instead of changing workdirs and using `kubectl cp`. I'll close this pr and open a new one.

PR #1439, opened 2019-01-07
Implement a notification based watcher

ltouati:
Hi Team,

This is to support file system notifications instead of reading the files one by ones. It is specially important on MacOS where the current notification scheme is killing the laptop battery.

Let me know your feedback


pkg/skaffold/watch/triggers.go
```diff
@@ -21,11 +21,14 @@ import (
...
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
+	"github.com/rjeczalik/notify"
+
```
dgageot: This empty line should be removed

pkg/skaffold/watch/triggers.go
```diff
@@ -109,3 +114,51 @@ func (t *manualTrigger) Start() (<-chan bool, func()) {
...
 	return trigger, func() {}
 }
+
+// notifyTrigger watches for changes when fsnotify
+type fsNotifyTrigger struct {
+}
+
+// Debounce tells the watcher to not debounce rapid sequence of changes.
+func (t *fsNotifyTrigger) Debounce() bool {
+	return false
```
dgageot: I think we should activate debounce, here.
If a lot of changes are made to the sources, very quickly, we don't want to trigger a rebuild too many times.

pkg/skaffold/watch/triggers.go
```diff
@@ -109,3 +114,51 @@ func (t *manualTrigger) Start() (<-chan bool, func()) {
...
+}
+
+func (t *fsNotifyTrigger) WatchForChanges(out io.Writer) {
+	color.Yellow.Fprintln(out, "FS Notify starting")
```
dgageot: This message is for the user to understand what's going on.
Should be something like `Watching for changes...`

pkg/skaffold/watch/triggers.go
```diff
@@ -109,3 +114,51 @@ func (t *manualTrigger) Start() (<-chan bool, func()) {
...
+}
+
+// Start starts listening to pressed keys.
+func (t *fsNotifyTrigger) Start() (<-chan bool, func()) {
+	trigger := make(chan bool)
+	c := make(chan notify.EventInfo, 1)
+
+	if err := notify.Watch("./...", c, notify.All); err != nil {
+		logrus.Fatalf("Unable to start notification: %v", err)
```
dgageot: Maybe `Start()` should return an error, now.

pkg/skaffold/watch/triggers.go
```diff
@@ -109,3 +114,51 @@ func (t *manualTrigger) Start() (<-chan bool, func()) {
...
+				trigger <- true
+			}
+		}
+	}()
+	return trigger, func() {
+		notify.Stop(c)
+	}
+}
+
+func isMatchingFile(event string) bool {
```
dgageot: This should not be trigger's job to ignore files.

Makefile
```diff
@@ -55,7 +55,7 @@ $(BUILD_DIR)/$(PROJECT): $(BUILD_DIR)/$(PROJECT)-$(GOOS)-$(GOARCH)
 	cp $(BUILD_DIR)/$(PROJECT)-$(GOOS)-$(GOARCH) $@
 
 $(BUILD_DIR)/$(PROJECT)-%-$(GOARCH): $(GO_FILES) $(BUILD_DIR)
-	GOOS=$* GOARCH=$(GOARCH) CGO_ENABLED=0 go build -ldflags $(GO_LDFLAGS) -gcflags $(GO_GCFLAGS) -asmflags $(GO_ASMFLAGS) -o $@ $(BUILD_PACKAGE)
+	GOOS=$* GOARCH=$(GOARCH) CGO_ENABLED=1 go build -ldflags $(GO_LDFLAGS) -gcflags $(GO_GCFLAGS) -asmflags $(GO_ASMFLAGS) -o $@ $(BUILD_PACKAGE)
```
dgageot: Do we have to enable CGO because of the new dependency? Is it documented somewhere?

hack/check-docs.sh
```diff
@@ -17,14 +17,15 @@
 cp docs/content/en/docs/references/cli/index_header docs/content/en/docs/references/cli/_index.md
 go run cmd/skaffold/man/man.go >> docs/content/en/docs/references/cli/_index.md
 
-readonly CLI_CHANGES=`git status -s | grep "docs/" | wc -l`
+readonly CLI_CHANGES=`git status -s | grep "docs/" | grep -x vendor | wc -l`
 
 if [ $CLI_CHANGES -gt 0 ]; then
+echo $CLI_CHANGES
```
dgageot: Could be removed, right?

pkg/skaffold/watch/triggers.go
```diff
@@ -43,6 +46,8 @@ func NewTrigger(opts *config.SkaffoldOptions) (Trigger, error) {
 		return &pollTrigger{
 			Interval: time.Duration(opts.WatchPollInterval) * time.Millisecond,
 		}, nil
+	case "notify":
```
dgageot: This should be added to the description of the `trigger` command line flag

pkg/skaffold/watch/triggers.go
dgageot: Oh, I see why you returned `false`, it doesn't work with `true`...

pkg/skaffold/watch/triggers.go
```diff
@@ -109,3 +114,51 @@ func (t *manualTrigger) Start() (<-chan bool, func()) {
...
+func (t *fsNotifyTrigger) WatchForChanges(out io.Writer) {
+	color.Yellow.Fprintln(out, "FS Notify starting")
+}
+
+// Start starts listening to pressed keys.
+func (t *fsNotifyTrigger) Start() (<-chan bool, func()) {
+	trigger := make(chan bool)
+	c := make(chan notify.EventInfo, 1)
+
+	if err := notify.Watch("./...", c, notify.All); err != nil {
```
dgageot: This should watch all the artifact folders instead.

pkg/skaffold/watch/triggers.go
ltouati: Indeed

pkg/skaffold/watch/triggers.go
ltouati: Removed is as per our discussion

hack/check-docs.sh
ltouati: indeed

pkg/skaffold/watch/triggers.go
ltouati: right

Makefile
ltouati: it is required by the fsnotify requirement that build differently depending on the platform

ltouati:
Hi @dgageot just implemented the requested changes, let me know if ok for you

Gopkg.toml
```diff
@@ -66,3 +66,7 @@
 [[constraint]]
   name = "github.com/bmatcuk/doublestar"
   revision = "v1.1.1"
+
+[[constraint]]
+  name = "github.com/rjeczalik/notify"
```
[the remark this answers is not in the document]
ltouati: Hi @r2d4
 this one support directory notification on all platforms. Specially useful on macOS where on relatively large projects you can exhaust number of open files as you need to watch files one by ones, which on top of this creates an issue when files are added into a directory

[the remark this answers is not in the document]
r2d4: 👍

r2d4:
FWIW I tried this out and it seemed to work quite well.

dgageot:
@ltouati Are you ok with my changes? The CLA requires that we both agree. I do

PR #1454, opened 2019-01-11
Show duplication in jib code

dgageot:
There's some duplication in jib code with regards to maven/gradle args.
This PR moves similar code fragments next to one another to make them easier to maintain.

Ideally, this duplication can be removed but I don't know enough jib to ensure I'm not breaking something.


dgageot:
ping @GoogleContainerTools/java-tools

PR #1459, opened 2019-01-12
Configuration layers

ltouati:
This PR add support for configuration layers based on activated profiles to skaffold.

When skaffold is triggered with
```
skaffold dev -p basis -p development
```
Skaffold will merge the skaffold_basis.y[a]ml and skaffold_development.y[a]ml with the main skaffold.yaml file

Obviously is any of those file do not exist this PR will not change the way skaffold load configuration files


pkg/skaffold/util/util.go
```diff
@@ -67,6 +67,16 @@ func StrSliceContains(sl []string, s string) bool {
 	return false
 }
 
+// Tests if a file exists
+func FileExists(filapath string) bool {
```
balopat:
```suggestion
func FileExists(filepath string) bool {
```

balopat: we should propagate the error if there is error os.Stat-ing the file, instead of swallowing it to `false`

ltouati:
Hi @balopat 

The use case I'm trying to solve is when you want to specialize a skaffold file. One example is when creating an application that uses angular. If using file-sync to update a container you probably don't want to do the same for your production environment. Here's an example of configuration
```
build:
  artifacts:
  - image: ${image.registry.url}/${image.prefix}-frontend
    context: ./frontend
```
and a skaffold_dev.yaml
````
build:
  artifacts:
  - image: ${image.registry.url}/${image.prefix}-frontend
    sync:
      'src/**': './src/'
    docker:
      dockerfile: Dockerfile.dev
````

This will allow to store in a single skaffold files all options. 

Another use case is if you want to have slightly different K8S config between your dev and env. You'll be able to add extra configuration files

dgageot:
@ltouati what would be the issue of using the existing profiles feature?

balopat:
I don't see this clearly separated from the current profiles feature - this might have been due to visibility: we  recently documented the full featureset here: https://skaffold.dev/docs/how-tos/profiles/
I'm going to close this now - please reopen it if you feel strongly about it.

PR #1466, opened 2019-01-15
feat(docker creds) can mount docker config into kaniko pod

garethjevans:


pkg/skaffold/build/kaniko/sources/sources.go
```diff
@@ -81,7 +81,39 @@ func podTemplate(cfg *latest.KanikoBuild, args []string) *v1.Pod {
 						SecretName: cfg.PullSecretName,
 					},
 				},
-			}},
+			},
+			},
 		},
 	}
+
+	if cfg.MountDockerConfig {
```
priyawadhwa: It might be cleaner here to check the negative, so that a bunch of code isn't nested in the if statement:

```golang
if !cfg.MountDockerConfig {
  return pod
}
```

pkg/skaffold/build/kaniko/sources/sources.go
```diff
@@ -81,7 +81,39 @@ func podTemplate(cfg *latest.KanikoBuild, args []string) *v1.Pod {
...
+		mountPath := constants.DefaultKanikoDockerConfigPath
+		if cfg.DockerConfigPath != "" {
+			mountPath = cfg.DockerConfigPath
```
priyawadhwa: I believe the mountPath should still be `constants.DefaultKanikoDockerConfigPath` here. If a user passes in `cfg.DockerConfigPath`, that should be the path to config.json on their local machine. If you pass in this path to `v1.VolumeMount` later on, it will mount in whatever's at that path on the cluster node, not on the machine.

To get around this, it might make sense to first create a secret from `cfg.DockerConfigPath`, and then mount that into the pod.  For reference, this is similar to how kaniko handles the pull [secret](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/kaniko/secret.go#L32) as well.

Let me know if that didn't make sense! :)

pkg/skaffold/build/kaniko/sources/sources.go
garethjevans: Done

pkg/skaffold/build/kaniko/sources/sources.go
garethjevans: Ok, that makes sense, I've taken the opposite approach here by assuming the `config.json` was already available as a secret in the cluster.  I'll get the code updated to handle this in the same way as the pull secret.

pkg/skaffold/build/kaniko/secret.go
```diff
@@ -65,12 +65,58 @@ func (b *Builder) setupSecret(out io.Writer) (func(), error) {
...
+		return nil, errors.Wrap(err, "reading docker config")
+	}
+
+	secret := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   b.DockerConfigSecretName,
+			Labels: map[string]string{"skaffold-kaniko": "skaffold-kaniko"},
+		},
+		Data: map[string][]byte{
+			constants.DefaultKanikoDockerConfigSecretName: secretData,
```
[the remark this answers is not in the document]
garethjevans: Done, good spot!

pkg/skaffold/schema/versions_test.go
```diff
@@ -88,6 +88,9 @@ build:
     pullSecretName: secret-name
     namespace: nskaniko
     timeout: 120m
+    mountDockerConfig: true
```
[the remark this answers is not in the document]
garethjevans: I've updated the code to match the suggestion, and added the example into annotated-skaffold.yaml

PR #1468, opened 2019-01-15
fix(docs) updated references for imageName to be image

garethjevans:


dgageot:
Thanks a lot @garethjevans 
Looks like there's more broken bits. I'll try to fix them

PR #1471, opened 2019-01-15
Use Maven wrapper for Jib example easier start.

ivanporty:
When working on coming IntelliJ IDE plugin demos, running Jib example is not as easy as it could be with a added Maven wrapper. Maven is required in `PATH` for example to work, but it may not be used at all (i.e. developer uses Gradle instead) or not used in CLI (internal IDE support for Maven). Maven wrapper will download and provide Maven automatically and Jib knows how to pick it up.

Also, this PR exposes container port for port forwarding in the container section, without it `skaffold` does not port-forward 8080 automatically.


ivanporty:
Sure, good idea, I'll bump up the version.

dgageot:
@ivanporty this is awesome. One issue, you have to also fix the same sample under `integration/examples`. I know this is painful...

ivanporty:
Ah, good to know, let me make the changes there as well.

ivanporty:
@dgageot the integration example is updated as well, let me know if something else is missing, thanks!

Issue #1225, opened 2018-10-30
Port Forward with client-go


r2d4:
https://github.com/kubernetes/client-go/blob/dcf16a0f3b52098c3d4c1467b6c80c3e88ff65fb/tools/portforward/portforward.go

Instead of shelling out to kubectl

dgageot:
I'm going to close this one. We don't have a problem right now to fix

PR #1474, opened 2019-01-15
Refactor port forwarding

### LINKED ISSUE #969: Stabilize port forwarding
We introduced port-forwarding for simple cases in #945.
This is an epic for all port-forwarding related issues that we want to solve in 1.0.0.

priyawadhwa:
Refactored port-forwarding to account for cases where:

1. Two containers map to the same port
2. Ports are already allocated on the machine

Now, when the port forwarder comes upon a new container, it follows
these steps:

1. It checks if the port is available on the machine. If so, then it
marks that port as taken.
2. If the port is unavailable, skaffold will find a new open port to map
to. It checks ports 4503-4533, which are unassigned user ports, first. If
none of those are open (which seems unlikely), then skaffold will find a
random open port.

Starts to fix #969

### DIFF
--- a/pkg/skaffold/kubernetes/port_forward.go
+++ b/pkg/skaffold/kubernetes/port_forward.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"os/exec"
 	"strconv"
 
@@ -42,7 +43,7 @@ type PortForwarder struct {
 	// forwardedPods is a map of portForwardEntry.key() (string) -> portForwardEntry
 	forwardedPods map[string]*portForwardEntry
 
-	// forwardedPorts is a map of port (int32) -> container name (string)
+	// forwardedPorts is a map of local port (int32) -> container name (string)
 	forwardedPorts map[int32]string
 }
 
@@ -52,6 +53,7 @@ type portForwardEntry struct {
 	namespace       string
 	containerName   string
 	port            int32
+	localPort       int32
 
 	cancel context.CancelFunc
 }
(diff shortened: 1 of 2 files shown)


pkg/skaffold/kubernetes/port_forward.go
```diff
@@ -175,42 +181,101 @@ func (p *PortForwarder) portForwardPod(ctx context.Context, pod *v1.Pod) error {
...
+	}
+	// If we have, return the current entry
+	oldEntry, ok := p.forwardedPods[entry.key()]
+	if ok {
+		entry.localPort = oldEntry.localPort
+		return entry, nil
+	}
+	// If another container isn't using this port...
+	if _, exists := p.forwardedPorts[port.ContainerPort]; !exists {
+		// ...Then make sure the port is availble
```
dgageot: linter is complaining s/availble/available/

priyawadhwa: thanks!

pkg/skaffold/kubernetes/port_forward.go
```diff
@@ -175,42 +181,101 @@ func (p *PortForwarder) portForwardPod(ctx context.Context, pod *v1.Pod) error {
...
+	if err := p.Forward(ctx, entry); err != nil {
+		return errors.Wrap(err, "port forwarding failed")
+	}
 	return nil
 }
 
+// From https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt,
+// ports 4503-4533 are unassigned user ports; first check if any of these are available
+// If not, return a random port, which hopefully won't collide with any future containers
+func getAvailablePort() (int32, error) {
```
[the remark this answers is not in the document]
priyawadhwa: hmm took a quick look at the code and couldn't find a function we could use for this  -- but when we do #1225 i think it'll take care of that for us.

PR #1482, opened 2019-01-17
New tagging mechanism

dgageot:
This completely rewrites the image tagging mechanism. Here's how it'll work:
 + Skaffold computes a tag for an image BEFORE it's built. This removes a lot of complexity.
 + Images pushed to registries will always be referenced by name, tag and digest in kubernetes manifests. That is to make sure the wrong image is not used and to remove this complexity from tagger.
 + Images used locally can't be referenced by digest (because we don't have it) or imageID (because k8s doesn't support that). So images used locally will be tagged with their imageID to make sure we produce a unique tag that can be used in kubernetes manifests
+ taggers (git, sha256, envTemplate, datetime) are all modified so that they can't use the image digest/imageID. It makes their code simpler but breaks the envTemplate tagger as it is.


pkg/skaffold/build/tag/env_template.go
```diff
@@ -32,10 +32,21 @@ type envTemplateTagger struct {
 
 // NewEnvTemplateTagger creates a new envTemplateTagger
 func NewEnvTemplateTagger(t string) (Tagger, error) {
+	if strings.Contains(t, "{{.DIGEST}}") {
```
balopat: This wouldn't work if it has a space or do we normalize it somewhere? e.g `{{ .DIGEST }}`

balopat:
Otherwise everything looks good. 
My recommendation would be to have 

`DIGEST` = `DIGEST`
`DIGEST_ALGO` = `sha256` 
`DIGEST_HEX`= `DIGEST`

and whenever these template params are used, we would print out a deprecation warning for 3 months. 
WDYT?

dgageot:
@balopat how do you see the deprecation warning working?

dgageot:
@priyawadhwa @balopat {{.DIGEST}} and friends should be handled better now.
+ In cases where it was used as an image name suffix (like for JenkinX), a deprecation warning is printed and this suffix is ignored because it'll be done automatically now.
+ In other cases, an error is printed and can't be recovered.

pkg/skaffold/build/tag/env_template.go
```diff
@@ -48,25 +64,31 @@ func (t *envTemplateTagger) Labels() map[string]string {
...
+		tag = strings.TrimSuffix(tag, "@[DEPRECATED_DIGEST]")
+
+	case strings.HasSuffix(tag, "@[DEPRECATED_DIGEST_ALGO]:[DEPRECATED_DIGEST_HEX]"):
+		warner.Warnf("{{.DIGEST}}, {{.DIGEST_ALGO}} and {{.DIGEST_HEX}} are deprecated, image digest will now automatically be apppended to image tags")
+		tag = strings.TrimSuffix(tag, "@[DEPRECATED_DIGEST_ALGO]:[DEPRECATED_DIGEST_HEX]")
+
+	case strings.Contains(tag, "[DEPRECATED_DIGEST]") ||
+		strings.Contains(tag, "[DEPRECATED_DIGEST_ALGO]") ||
+		strings.Contains(tag, "[DEPRECATED_DIGEST_HEX]"):
+		return "", errors.New("{{.DIGEST}}, {{.DIGEST_ALGO}} and {{.DIGEST_HEX}} are deprecated, image digest will now automatically be apppended to image tags")
```
nkubala: s/`apppended`/`appended`

dgageot: Let me fix that

pkg/skaffold/build/tag/env_template.go
dgageot: This has been fixed

pkg/skaffold/build/tag/env_template.go
```diff
@@ -32,10 +32,21 @@ type envTemplateTagger struct {
 
 // NewEnvTemplateTagger creates a new envTemplateTagger
 func NewEnvTemplateTagger(t string) (Tagger, error) {
+	if strings.Contains(t, "{{.DIGEST}}") {
+		return nil, errors.New("{{.DIGEST}} is deprecated, image digest will automatically be apppended to image tags")
```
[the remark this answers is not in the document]
dgageot: This has been fixed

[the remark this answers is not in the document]
dgageot: This has been fixed

pkg/skaffold/build/tag/env_template.go
```diff
@@ -48,25 +64,31 @@ func (t *envTemplateTagger) Labels() map[string]string {
...
+		tag = strings.TrimSuffix(tag, "@[DEPRECATED_DIGEST]")
+
+	case strings.HasSuffix(tag, "@[DEPRECATED_DIGEST_ALGO]:[DEPRECATED_DIGEST_HEX]"):
+		warner.Warnf("{{.DIGEST}}, {{.DIGEST_ALGO}} and {{.DIGEST_HEX}} are deprecated, image digest will now automatically be appended to image tags")
+		tag = strings.TrimSuffix(tag, "@[DEPRECATED_DIGEST_ALGO]:[DEPRECATED_DIGEST_HEX]")
+
+	case strings.Contains(tag, "[DEPRECATED_DIGEST]") ||
+		strings.Contains(tag, "[DEPRECATED_DIGEST_ALGO]") ||
+		strings.Contains(tag, "[DEPRECATED_DIGEST_HEX]"):
+		return "", errors.New("{{.DIGEST}}, {{.DIGEST_ALGO}} and {{.DIGEST_HEX}} are deprecated, image digest will now automatically be appended to image tags")
```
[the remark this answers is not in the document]
dgageot: This is fixed now

balopat:
As dicussed let's change the "[" "]" chars to "_" so that the image names are compatible with docker, then we can merge it!

dgageot:
Woot! All green 💚 🍏 📗 
Let's merge!

PR #1495, opened 2019-01-20
Pre-pull cache-from images

### LINKED ISSUE #936: cacheFrom option doesn't work
Thanks a lot for building skaffold! 
I feel like I will not need any CI/CD tools soon :) ...almost.

I want to speed up the container building process for my project on Google Cloud.
The problem is that **cacheFrom** option is ignored completely.

### Expected behavior
**cacheFrom** should force build process to use cache from the previously built image
### Actual behavior
build process repeats every time without using the cache 
### Information
I have checked the previously build image on gcr.io - it exists

- Skaffold version: v0.12.0
- Operating system: mac os 10.13.3
- Contents of skaffold.yaml:
```
  - name: stage
    build:
      googleCloudBuild:
        projectId: neural-pub
      artifacts:
        - imageName: gcr.io/neural-pub/api
          workspace: ./api
          docker:
            cacheFrom:
            - gcr.io/neural-pub/api
        - imageName: gcr.io/neural-pub/web
          workspace: ./web
          docker:
            cacheFrom:
            - gcr.io/neural-pub/web
    deploy:
      kubectl:
        manifests:
          - ./ingress/k8s-stage-*
          - ./api/k8s-stage-*
          - ./web/k8s-stage-*
```

1. skaffold run --profile stage
2. run this command again

I am wondering - does this feature work on Google Cloud?

dgageot:
Fixes #936

### DIFF
--- a/integration/examples/annotated-skaffold.yaml
+++ b/integration/examples/annotated-skaffold.yaml
@@ -56,8 +56,8 @@ build:
         key2: "value2"
       # Images to consider as cache sources
       cacheFrom:
-      - image1
-      - image2
+      - golang:1.10.1-alpine3.7
+      - alpine:3.7
       # Dockerfile target name to build.
       # target: stageName
 
--- a/pkg/skaffold/build/local/docker.go
+++ b/pkg/skaffold/build/local/docker.go
@@ -29,6 +29,10 @@ import (
 )
 
 func (b *Builder) buildDocker(ctx context.Context, out io.Writer, workspace string, a *latest.DockerArtifact) (string, error) {
+	if err := b.pullCacheFromImages(ctx, out, a); err != nil {
+		return "", errors.Wrap(err, "pulling cache-from images")
+	}
+
 	initialTag := util.RandomID()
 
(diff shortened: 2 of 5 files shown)


dgageot:
> Will it also pull new versions even if it has a local image of the specified tag?

Do you think it should? It currently doesn't but both behaviour would make sense to me. I get always pulling is more predictable.

tjerkw:
> > Will it also pull new versions even if it has a local image of the specified tag?
> 
> Do you think it should? It currently doesn't but both behaviour would make sense to me. I guess always pulling is more predictable.

It depends on the context, for us, in our CI environment, we always need to pull, because that local images might be stale. 

However for local development, you don't want to pull everytime.

So maybe it could be a flag? 
Or, if we need to choose: Dont pull if its locally available, then we can still pre-pull before calling skaffold (thats what we currently doing).

If you always pull, you have to make sure you dont pull the same cacheFrom image twice.
What we do in our scripts, is to gather all cacheFrom images in a Set, so we dont have dups, and then pull them.

dgageot:
@tjerkw let's merge it as it is and I'll create a ticket to configure 'pull always'

Issue #1328, opened 2018-11-28
Purge helm release option on skaffold delete


bholemt:
In case of change in helm template, one needs to helm delete --purge. Otherwise helm upgrade tries to reuse the template from the old delete release. Wish this option is available somehow with skaffold delete.

bholemt:
I need to catch up on Go. Let me try but will need time.

tjerkw:
I think the --force flag for "helm upgrade" also solves this, right?

https://github.com/helm/helm/pull/2280

tejal29:
This is addressed in #2280

PR #1499, opened 2019-01-20
Add support for helm --force flag

tjerkw:
We're using skaffold and helm to deploy. In our ci-cluster deploys sometimes fail,
but then skaffold cannot install it anymore if it finds a failed helm installation.
The --force flag solves this. 

See the --force functionailtiy added in helm 2.5.0:
https://github.com/helm/helm/pull/2280

Potential solution for #1328 
Makes it more easy to use skaffold in a ci-pipeline. (deploy from a clean sheet)

### DIFF
--- a/examples/annotated-skaffold.yaml
+++ b/examples/annotated-skaffold.yaml
@@ -183,6 +183,7 @@ deploy:
 
  # helm:
     # helm releases to deploy.
+    # Uses "helm install" if not installed yet, otherwise "helm upgrade" is executed
     # releases:
     # - name: skaffold-helm
     #   chartPath: skaffold-helm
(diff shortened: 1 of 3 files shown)


tjerkw:
I signed it!

examples/annotated-skaffold.yaml
```diff
@@ -183,6 +183,7 @@ deploy:
 
  # helm:
     # helm releases to deploy.
+    # Uses "helm install" if not installed yet, otherwise "helm upgrade" is executed
```
priyawadhwa: So, unfortunately because of the way we have to release config changes, this change also needs to be made in [integration/examples/annotated-skaffold.yaml](https://github.com/GoogleContainerTools/skaffold/blob/master/integration/examples/annotated-skaffold.yaml). Would you mind adding it there as well? Thanks!

tjerkw:
As agreed here, lets use the flag based approach:
https://github.com/GoogleContainerTools/skaffold/pull/1506

PR #1522, opened 2019-01-23
[wip] Add --port flag to skaffold dev

### LINKED ISSUE #1311: User defined port-forwarding
It would be useful to let the developer remap ports to be forwarded locally, say when they want to work on deeper level components that are using the same port.

priyawadhwa:
I added a --port flag to skaffold dev so users could specify additional ports they want to port forward, to be used with Java debugging (#1480) . This will also allow users to specify local ports as well (right now, skaffold randomly forwards to an open port in the event of a collision), which should fix #1311.

The flag takes in strings in the form `pod/container:localPort:containerPort`. When a pod event happens, skaffold first goes through the user defined ports and forwards any ports assigned to that pod. It then goes through the ports specified in the manifest file and forwards any additional ports there.

We could also add a field to the skaffold.yaml, which should be pretty easy to add:

```yaml
deploy:
  portForward:
  - pod: mypod
    container: mycontainer
    containerPort: 8080 
    localPort: 9000
```

cc @etanshaul


etanshaul:
fyi as @briandealwis mentioned here (https://github.com/GoogleContainerTools/skaffold/issues/1480#issuecomment-457040326) this approach has some limitations

priyawadhwa:
for sure, I'm going to close this PR until we've decided what the best approach is.

PR #1531, opened 2019-01-24
Add zsh completion

### LINKED ISSUE #1012: Completion support for zsh

corneliusweig:
This PR adds command completion for skaffold in zsh. It uses the cobra zsh completion generator. Cobra has only limited support for completion in zsh, for example options are not included in the completion unlike for bash completion. However, a revised completion generator is in the works (https://github.com/spf13/cobra/pull/646), so that it didn't seem reasonable to build a custom temporary solution. Therefore, this zsh completion for skaffold is still a bit basic, but hopefully will improve soon.

Fixes #1012

### DIFF
--- a/cmd/skaffold/app/cmd/completion.go
+++ b/cmd/skaffold/app/cmd/completion.go
@@ -24,30 +24,46 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// completionCmd represents the completion command
+const longDescription = `
+	Outputs skaffold shell completion for the given shell (bash or zsh)
+
+	This depends on the bash-completion binary.  Example installation instructions:
+	OS X:
+		$ brew install bash-completion
+		$ source $(brew --prefix)/etc/bash_completion
+		$ skaffold completion bash > ~/.skaffold-completion  # for bash users
+		$ skaffold completion zsh > ~/.skaffold-completion   # for zsh users
+		$ source ~/.skaffold-completion
+	Ubuntu:
+		$ apt-get install bash-completion
+		$ source /etc/bash-completion
+		$ source <(skaffold completion bash) # for bash users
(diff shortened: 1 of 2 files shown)


cmd/skaffold/app/cmd/completion.go
```diff
@@ -24,30 +24,46 @@ import (
...
+	Additionally, you may want to output the completion to a file and source in your .bashrc
+`
+
 var completionCmd = &cobra.Command{
-	// Only bash is supported for now. However, having args after
-	// "completion" will help when supporting multiple shells
-	Use: "completion bash",
+	Use: "completion SHELL",
 	Args: func(cmd *cobra.Command, args []string) error {
 		if len(args) != 1 {
```
dgageot: Could it default to `bash` to not break existing usage?

PR #1541, opened 2019-01-28
Better profiles

dgageot:
This PR implements two features related to Profiles:

 + It makes it possible to patch individual configuration values using json patch syntax.
 + It makes it possible to auto-activate profiles by skaffold command, env variable or kubeContext

### DIFF
--- a/pkg/skaffold/schema/profiles.go
+++ b/pkg/skaffold/schema/profiles.go
@@ -21,33 +21,32 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-
-	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/defaults"
-	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	yaml "gopkg.in/yaml.v2"
 )
 
 // ApplyProfiles returns configuration modified by the application
 // of a list of profiles.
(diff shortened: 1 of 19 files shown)


pkg/skaffold/schema/profiles.go
```diff
@@ -18,36 +18,114 @@ package schema
...
+
+	currentKubeContext, err := kubectx.CurrentContext()
+	if err != nil {
+		return false, errors.Wrap(err, "getting current cluster context")
+	}
+
+	return equalValue(kubeContext, currentKubeContext), nil
+}
+
+func equalValue(expected, actual string) bool {
```
balopat:
```suggestion
func satisfies(expected, actual string) bool {
```

pkg/skaffold/schema/profile_test.go
```diff
@@ -142,13 +156,164 @@ func TestApplyProfiles(t *testing.T) {
...
+
+func TestActivatedProfiles(t *testing.T) {
+	tests := []struct {
+		description string
+		profiles    []latest.Profile
+		opts        *cfg.SkaffoldOptions
+		expected    []string
+		shouldErr   bool
+	}{
+		{
```
balopat: There are no tests covering the case where the `Activation` array has multiple activations. It is unclear if it should be OR or AND  - it looks like from the implementation that it's a logical OR.

dgageot:
@balopat Thanks for the review!
 + I'm fixing the nits
 + I'll add an integration test
 + The doc is currently unclear about json patches. Those are not external files. I'll work on a sample and an integration test to make that clearer
 + I'll document the behaviour with regards to AND and OR

PR #1546, opened 2019-01-29
Add an http trigger

dgageot:
This is an additional trigger that makes it possible to drive `skaffold dev` from an outside such as a file watcher or an IDE.

**How-to**:

 + Run `skaffold dev --trigger=http`
 + Send a POST query to `http://localhost:45678/skaffold/v1/build`, for eg with: `curl -X POST http://localhost:45678/skaffold/v1/build`
 + If any file was actually changed, then Skaffold will rebuild what's necessary.
 + Skaffold stops watching for file changes but maintains a snapshot of file change dates, at startup and each time the trigger is called.
 + When the trigger is called rapidly, then Skaffold debounces events to not be flooded.

The address on which Skaffold listens can be changed with `--trigger-address tcp://0.0.0.0:45678` command line arg. It can also listen to a unix socket with `--trigger-address unix:///tmp/skaffold`.Named pipes are not yet implemented.

ping @briandealwis

### DIFF
--- a/cmd/skaffold/app/cmd/dev.go
+++ b/cmd/skaffold/app/cmd/dev.go
@@ -38,7 +38,8 @@ func NewCmdDev(out io.Writer) *cobra.Command {
 	}
 	AddRunDevFlags(cmd)
 	cmd.Flags().BoolVar(&opts.TailDev, "tail", true, "Stream logs from deployed objects")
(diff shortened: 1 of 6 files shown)


dgageot:
Let's clos that PR since it conflicts with what @nkubala is working on

PR #1557, opened 2019-01-30
Add missing tests for watch/triggers

shuheiktgw:
Added missing tests for watch/triggers! Also, I found profile_test.go is supposed to be named profiles_test.go, so I renamed it so. Thanks 👍


pkg/skaffold/watch/triggers_test.go
```diff
@@ -0,0 +1,95 @@
...
+		expected    Trigger
+		shouldErr   bool
+	}{
+		{
+			description: "polling trigger",
+			opts:        &config.SkaffoldOptions{Trigger: "polling", WatchPollInterval: 1},
+			expected: &pollTrigger{
+				Interval: time.Duration(1) * time.Millisecond,
+			},
+			shouldErr: false,
```
dgageot: We usually skip `shouldErr: false` in out tests.

pkg/skaffold/watch/triggers_test.go
```diff
@@ -0,0 +1,95 @@
...
+		{
+			description: "manual trigger",
+			opts:        &config.SkaffoldOptions{Trigger: "manual"},
+			expected:    &manualTrigger{},
+			shouldErr:   false,
+		},
+		{
+			description: "unknown trigger",
+			opts:        &config.SkaffoldOptions{Trigger: "unknown"},
+			expected:    nil,
```
dgageot: This could be skipped too

PR #1576, opened 2019-02-01
Fix DEVELOPMENT.md fragment

shuheiktgw:
Hi, I think the fragment below does not exist on the page, so I fixed it! 👍


priyawadhwa:
Thanks for fixing this!

PR #1600, opened 2019-02-05
[WIP] Use a pseudo json schema to generate the documentation

dgageot:
This is work in progress to improve our documentation.

The idea is to manually maintain a json file that serves as a schema for `skaffold.yaml`. That shema would help:
 1. Generate large portions of the documentation
 2. Generate the `annotated-skaffold.yaml`
 3. Generate schemas for IDEs to validate `skaffold.yaml` files

This PR currently achieves 1.

Next step is to make sure that this schema is in sync with the Go code with a unit test.

Why not generating this schema from Go code? I'm not sure every information can be put in Go comments (default values for example).

### DIFF
--- a/docs/content/en/docs/how-tos/builders/_index.md
+++ b/docs/content/en/docs/how-tos/builders/_index.md
@@ -42,11 +42,7 @@ of `skaffold.yaml`.
 
 The `local` type offers the following options:
 
-| Option        | Description | Default |
-|---------------|-------------|---------|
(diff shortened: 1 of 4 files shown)


balopat:
This is an interesting direction - how do you plan to keep this json schema in sync with the config structs?

dgageot:
So, my idea is to maintain that schema file manually but have a set of tests that make sure it is in sync with the Go code.

PR #1609, opened 2019-02-05
Add custom build arguments in jib artifacts

### LINKED ISSUE #1565: Provide additional flags for Maven/Gradle
It would be useful to set additional flags at the artifact level that would be passed to gradle or maven.
For eg, if one wants to skip tests.

Something like: 
```
apiVersion: skaffold/v1beta3
kind: Config
build:
  artifacts:
  - image: gcr.io/k8s-skaffold/skaffold-jib
    jibMaven:
      additionalFlags:
      - -x
      - -Dmaven.test.skip
```

corneliusweig:
Before, there was no way to pass custom arguments to the jib artifacts.
Now, there is the `mavenArgs` and `gradleArgs` field which expect a list of strings which is forwarded to the maven and gradle builder, respectively.

Fixes #1565

### DIFF
--- a/examples/annotated-skaffold.yaml
+++ b/examples/annotated-skaffold.yaml
@@ -76,14 +76,14 @@ build:
     # jibMaven:
     #  module: modulename     # selects which maven module to build, for a multimodule project
     #  profile: profilename   # selects which maven profile to activate
-    #  mavenArgs:             # additional arguments to pass to maven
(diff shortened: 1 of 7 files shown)


dgageot:
@corneliusweig Can you rename `mavenArgs` and `gradleArgs` to `args`?

PR #1614, opened 2019-02-06
Fix unsupported transitive target case

dgageot:
When a Dockerfile is built and a target is provided, Skaffold needs to find the dependencies for that target.

The code was missing an important case of transitive dependency when a target `A` depends on a target `B` because it has a `COPY --from=B` instruction. (Could also be a `COPY --from=1` assuming target `B` is the second target)

### DIFF
--- a/pkg/skaffold/docker/parse.go
+++ b/pkg/skaffold/docker/parse.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strconv"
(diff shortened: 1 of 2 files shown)


dgageot:
I've done more testing and it seems that `--target` in docker build is much more dumb. It doesn't look for transitive dependencies between targets. It builds every target until the provided target. I'm going to rollback the code that filters deps on the target. It's much safer.

Issue #1623, opened 2019-02-07
Allow specifying an explicit image tag


michaelfig:
### Expected behavior

Specifying an explicit image tag will be used by Skaffold.  See the `skaffold.yaml` below.

I have a PR for this change: #1622

The rationale is that I want to make a staged `./Dockerfile` that depends on the `./docker/**/Dockerfile` to create build and install base images that are pushed to the build registry so that builder processes can also find them there as a cache:

```
ARG BUILD_REGISTRY=registry.example.com
ARG BUILD_BASE=${BUILD_REGISTRY}/citest:build-base-1.0.0
ARG INSTALL_BASE=${BUILD_REGISTRY}/citest:install-base-1.0.0
FROM ${BUILD_BASE} AS build
[...]
FROM ${INSTALL_BASE}
COPY --from=build /some/path /some/new/path
[...]
```

### Actual behavior

```
$ skaffold build
[...]
Successfully built 12d5565cc0f3
Successfully tagged 975c05a3af0fe406bb5244d510eef4b4:latest
FATA[0000] build failed: building [michaelfig/citest:build-base-1.0.0]: tagging: Error parsing reference: "michaelfig/citest:build-base-1.0.0:0dd8dfd" is not a valid repository/tag: invalid reference format
$
```

### Information

- Skaffold version: 0.22.0
- Operating system: darwin
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta4
kind: Config
build:
  artifacts:
  - image: michaelfig/citest:build-base-1.0.0
    context: ./docker/build-base
  - image: michaelfig/citest:install-base-1.0.0
    context: ./docker/install-base
  - image: michaelfig/citest
    docker:
      buildArgs:
        BUILD_REGISTRY: michaelfig
deploy:
  kubectl:
    manifests:
    - k8s/*.yaml
```

### Steps to reproduce the behavior

1. Run `skaffold build` or `skaffold dev` with artifacts that have explicit tags.

michaelfig:
As @balopat explained, this works as expected with:

```yaml
build:
  tagPolicy:
    sha256: {}
```

PR #1634, opened 2019-02-09
Don't push all tags when sha256 builds just one

michaelfig:
I don't yet know how to describe the problem, except for that sha256 sends thousands of pushes if you use it with an image that doesn't have its own tag.

Forcing the tag to `:skaffold` (`:latest` would work, too) makes it behave properly.


pkg/skaffold/build/tag/sha256.go
```diff
@@ -31,5 +36,17 @@ func (c *ChecksumTagger) Labels() map[string]string {
...
+	}
+	noRepo := repoPortRe.ReplaceAllLiteralString(imageName, "")
+
+	if strings.Contains(noRepo, ":") {
+		// They have a tag, so use it.
+		return imageName, nil
+	}
+
+	// No supplied tag, so use skaffold.
+	return imageName + ":skaffold", nil
```
dgageot: I'd rather we use `:latest`, here

pkg/skaffold/build/tag/sha256.go
```diff
@@ -31,5 +36,17 @@ func (c *ChecksumTagger) Labels() map[string]string {
 }
 
 func (c *ChecksumTagger) GenerateFullyQualifiedImageName(workingDir, imageName string) (string, error) {
-	return imageName, nil
+	// Strip repositories that have a port number.
+	if repoPortRe == nil {
```
dgageot: Please, use something like `parsed, err := docker.ParseReference(tag)` to parse image names.
Also, image names on artifacts should never have a tag. We should maybe enforce that somewhere else.

dgageot:
@michaelfig Could you create a ticket with a procedure to reproduce? I couldn't reproduce the issue here.

pkg/skaffold/build/tag/sha256.go
michaelfig: #1623 is where I explained a use case for having tags on built/pushed artifacts.

Thanks so much for your feedback!

michaelfig:
@dgageot working on finding a simple example for a new issue.

michaelfig:
@dgageot #1638 describes this problem in detail, with minimal steps to reproduce.

On my system, I have a lot of tags on one image, so it's especially slow and onerous to push them all.

PR #1636, opened 2019-02-09
Add missing tests for update package

shuheiktgw:
- Refactored GetLatestAndCurrentVersion method for testability
- Fix a small bug with GetLatestAndCurrentVersion, which does not return error when http status is not 200
- Add tests for update package

Thanks! 👍

### DIFF
--- a/pkg/skaffold/update/update.go
+++ b/pkg/skaffold/update/update.go
@@ -17,6 +17,7 @@ limitations under the License.
 package update
 
 import (
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -28,7 +29,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-const latestVersionURL = "https://storage.googleapis.com/skaffold/releases/latest/VERSION"
+const LatestVersionURL = "https://storage.googleapis.com/skaffold/releases/latest/VERSION"
 
 // IsUpdateCheckEnabled returns whether or not the update check is enabled
 // It is true by default, but setting it to any other value than true will disable the check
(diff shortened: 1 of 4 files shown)


pkg/skaffold/update/update.go
```diff
@@ -44,15 +45,15 @@ func IsUpdateCheckEnabled() bool {
...
+func GetLatestAndCurrentVersion(url string) (semver.Version, semver.Version, error) {
 	none := semver.Version{}
-	resp, err := http.Get(latestVersionURL)
+	resp, err := http.Get(url)
 	if err != nil {
 		return none, none, errors.Wrap(err, "getting latest version info from GCS")
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return none, none, errors.Wrapf(err, "http %d, error: %s", resp.StatusCode, resp.Status)
```
shuheiktgw: `err` will be always nil since it is sure `err` is nil as [line 50](https://github.com/GoogleContainerTools/skaffold/pull/1636/files#diff-3bf14b4d33cbb50151fe5326e8cb450bL50) suggests,  and `errors.Wrapf` returns nil when err is nil.

pkg/skaffold/update/update_test.go
```diff
@@ -0,0 +1,105 @@
...
+				"SKAFFOLD_UPDATE_CHECK": test.updateCheckEnvVar,
+			})
+			defer reset(t)
+
+			testutil.CheckDeepEqual(t, test.expected, IsUpdateCheckEnabled())
+		})
+	}
+}
+
+func TestGetLatestAndCurrentVersion(t *testing.T) {
```
shuheiktgw: I could not add a case when `GetLatestAndCurrentVersion` succeeds since `version` variable in version package is enexported and `version.ParseVersion` for current version always fails.

shuheiktgw:
@dgageot 

Thank you for your comment!

> See comment about errors.Wrapf

I guess you meant the comment below?

> If err is nil, Wrapf returns nil.

I think my comment was not clear enough, but this behavior of `Wrapf` is the problem here because we know at (line 55)[https://github.com/GoogleContainerTools/skaffold/pull/1636/files#diff-3bf14b4d33cbb50151fe5326e8cb450bL55] err is always nil (This is cleate from [line 50](https://github.com/GoogleContainerTools/skaffold/pull/1636/files#diff-3bf14b4d33cbb50151fe5326e8cb450bL50)). As a result, even if the http status is not OK, `GetLatestAndCurrentVersion` won't return error, so I think we have to create an error and return it not using Wrapf!

Thanks! 👍

Issue #1639, opened 2019-02-12
Single exec command for sync


r2d4:
### Expected behavior
Sync should only make one exec call per sync round

### Actual behavior
Sync makes a call per copy and a call per delete

### Information

Step 1: Bunch the delete operations into a single exec command
Step 2: Replace `kubectl cp` command with 

- Tar (in Go) all files-to-be-copied with the correct destination path
- Exec a single `cp sync.tar / && tar -xvf /sync.tar && rm /sync.tar` for all copies

Step 3 (maybe?): Combine delete and copy operations into a single large command

michaelfig:
#1641 implements steps 1 and 2.

corneliusweig:
@r2d4 Do you really think combining copy & delete (step3) would achieve much? It would not be trivial to combine both, because copy needs to read the tar from stdin, whereas delete does not. Probably not worth the complexity..

r2d4:
Agreed @corneliusweig

PR #1641, opened 2019-02-12
Optimise sync

michaelfig:
This PR implements steps 1 and 2 of #1639.

### DIFF
--- a/pkg/skaffold/sync/kubectl/kubectl.go
+++ b/pkg/skaffold/sync/kubectl/kubectl.go
@@ -32,8 +32,6 @@ type Syncer struct {
 	namespaces []string
 }
 
-var syncedDirs = map[string]struct{}{}
-
 func NewSyncer(namespaces []string) *Syncer {
 	return &Syncer{
 		namespaces: namespaces,
--- a/pkg/skaffold/sync/kubectl/kubectl.go
+++ b/pkg/skaffold/sync/kubectl/kubectl.go
@@ -18,11 +18,11 @@ package kubectl
 
 import (
 	"context"
-	"fmt"
+	"io"
 	"os/exec"
-	"path/filepath"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/sync"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


michaelfig:
Notably, the touch-after-sync is avoided by using the `m` flag to `tar`, which does the touch itself.

Maybe controversially, I added the `--no-same-owner` flag, which is the only thing that really makes sense to me when syncing a source file directly to a container.

pkg/skaffold/sync/kubectl/kubectl.go
```diff
@@ -60,18 +58,30 @@ func (k *Syncer) Sync(ctx context.Context, s *sync.Item) error {
...
-	if _, ok := syncedDirs[dir]; !ok {
-		cmds = []*exec.Cmd{exec.CommandContext(ctx, "kubectl", "exec", pod.Name, "-c", container.Name, "-n", pod.Namespace, "--", "mkdir", "-p", dir)}
-		syncedDirs[dir] = struct{}{}
-	}
-	copy := exec.CommandContext(ctx, "kubectl", "cp", src, fmt.Sprintf("%s/%s:%s", pod.Namespace, pod.Name, dst), "-c", container.Name)
-	return append(cmds, copy)
+func copyFileFn(ctx context.Context, pod v1.Pod, container v1.Container, files map[string]string) []*exec.Cmd {
+	// Use "m" flag to touch the files as they are copied.
+	reader, writer := io.Pipe()
+	copy := exec.CommandContext(ctx, "kubectl", "exec", pod.Name, "--namespace", pod.Namespace, "-c", container.Name, "-i",
```
r2d4: Remove the tar after extraction?

Also TIL about `-m` or `--touch` flag. Cool

pkg/skaffold/sync/kubectl/kubectl.go
```diff
@@ -60,18 +58,30 @@ func (k *Syncer) Sync(ctx context.Context, s *sync.Item) error {
...
+func deleteFileFn(ctx context.Context, pod v1.Pod, container v1.Container, files map[string]string) []*exec.Cmd {
+	// "kubectl" is below...
+	deleteCmd := []string{"exec", pod.Name, "--namespace", pod.Namespace, "-c", container.Name, "--", "rm", "-rf"}
+	args := make([]string, 0, len(deleteCmd)+len(files))
+	args = append(args, deleteCmd...)
+	for _, dst := range files {
+		args = append(args, dst)
+	}
+	delete := exec.CommandContext(ctx, "kubectl", args...)
 	return []*exec.Cmd{delete}
```
r2d4: Both copyFn and deleteFn can just return a single exec.Cmd? WDYT?

pkg/skaffold/sync/sync.go
```diff
@@ -140,35 +140,32 @@ func Perform(ctx context.Context, image string, files map[string]string, cmdFn f
...
+					numSynced++
 				}
 			}
 		}
+	}
 
-		if len(synced) != len(files) {
-			return errors.New("couldn't sync all the files in " + ns)
-		}
+	if numSynced == 0 {
```
r2d4: Not sure if this check makes sense anymore?

pkg/skaffold/sync/kubectl/kubectl.go
michaelfig: The tar file is piped from Go's memory directly to `kubectl exec`'s stdin.  So there's nothing to remove.

pkg/skaffold/sync/kubectl/kubectl.go
michaelfig: No harm in leaving them as slices, but I'm indifferent.

pkg/skaffold/sync/kubectl/kubectl.go
r2d4: Ah missed that 👍

pkg/skaffold/sync/sync.go
michaelfig: I put this in to pass the "no copy" test at `sync_test.go`:410.  If you want, both this and the test can be removed.

PR #1642, opened 2019-02-12
Add `required` tags

dgageot:


dgageot:
@balopat @priyawadhwa @nkubala The PR is failing on an integration test that has no artifact to build. Should we really support that usecase?

priyawadhwa:
Is the idea that people would use skaffold just to deploy manifests? Personally I'm not sure how useful that would be.

PR #1644, opened 2019-02-12
Generate json schema

dgageot:
Make sure a json schema is generated for each version of the `skaffold.yaml`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,3 +193,9 @@ build-docs-preview:	docs-preview-image
 .PHONY: clean-docs-preview
 clean-docs-preview: docs-preview-image
 	docker run --rm -ti -v $(PWD):/app --workdir /app/ -p 1313:1313 skaffold-docs-previewer bash -xc deploy/docs/clean.sh
+
+# schema generation
+
+.PHONY: generate-schemas
+generate-schemas:
+	go run schemas/main.go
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -193,3 +193,9 @@ build-docs-preview:	docs-preview-image
 .PHONY: clean-docs-preview
 clean-docs-preview: docs-preview-image
 	docker run --rm -ti -v $(PWD):/app --workdir /app/ -p 1313:1313 skaffold-docs-previewer bash -xc deploy/docs/clean.sh
+
+# schema generation
+
+.PHONY: generate-schemas
+generate-schemas:
```
r2d4: How are you going to keep this file up to date?

balopat: +1 we should write a script similar to hack/check-docs.sh that runs the generate-schemas target on schema changes and if there are changes then fails the build.

balopat: ignore me - there is a unit test for that - https://github.com/GoogleContainerTools/skaffold/pull/1644/files#diff-075a0df1686535f0bc3a3408646079c6

dgageot: yes, there's a unit test. That'll do it for now. Another option would be to generate those files only on a release.

dgageot:
@balopat This should be better now

PR #1646, opened 2019-02-12
Config upgrade: handle helm overrides

### LINKED ISSUE #1585: upgrading old config fails with unsupported type error
### Expected behavior
my file should correctly parse and update

### Actual behavior
parsing fails but file is valid and works if we just manually change the apiVersion.

### Information

- Skaffold version: 0.21.0 to 0.22.0
- Operating system: linux - ubuntu 18.04
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta3
kind: Config
build:
  artifacts:
  - image: local/app1
    context: .
  - image: local/redis-cl
    context: ./redis-local/
deploy:
  helm:
    releases:
    - name: app1
      chartPath: app1
      values:
        image: local/app1
        redisImage: local/redis-cl
      overrides:
        global:
          localstack:
            enabled: true
          environment: "local"
      namespace: "default"
      wait: false
      recreatePods: false
```

run skaffold fix and the following is printed:
```
WARN[0000] config version (skaffold/v1beta3) out of date: upgrading to latest (skaffold/v1beta4) 
FATA[0000] creating runner: parsing skaffold config: transforming skaffold config: converting deploy config: marshalling old: json: unsupported type: map[interface {}]interface {} 
```

but if I simply update the `apiVersion` to beta4 it works fine and skaffold is totally happy actually parsing the config

corneliusweig:
Json serialization does not cope with helm overrides, because its type `map[string]interface{}` is not serialized by the json encoder. To deal with that, implement a fallback which includes the helm overrides as a yaml string.

That way, the config upgrade is forwards compatible, so that future api-versions do not need to do any manual adaptions for a working json serialization. I also tried a much more manual approach which would have required manual tasks for future api versions :-1: 

One drawback of my current approach is that it requires a new struct `HelmOverrides` which is shared among all api-version schemas. However, since the content of `HelmOverrides` is opaque to Skaffold anyway, this should be ok.

Fixes #1585

### DIFF
--- a/docs/content/en/schemas/v1alpha2.json
+++ b/docs/content/en/schemas/v1alpha2.json
@@ -472,11 +472,7 @@
         "namespace": {
           "type": "string"
         },
-        "overrides": {
-          "additionalProperties": {},
-          "type": "object",
-          "default": "{}"
-        },
+        "overrides": {},
         "packaged": {
           "$ref": "#/definitions/HelmPackaged"
         },
--- a/docs/content/en/schemas/v1alpha3.json
+++ b/docs/content/en/schemas/v1alpha3.json
@@ -472,11 +472,7 @@
         "namespace": {
           "type": "string"
         },
-        "overrides": {
-          "additionalProperties": {},
-          "type": "object",
-          "default": "{}"
-        },
+        "overrides": {},
         "packaged": {
           "$ref": "#/definitions/HelmPackaged"
         },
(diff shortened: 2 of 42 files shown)


corneliusweig:
@nkubala Do you think this is a viable approach? Since you commented on the bug report you seem to be the right person to ask that. Besides, this PR has been open for almost a month now, and I would like to resolve this at some point.

balopat:
Thank you @corneliusweig for contributing this! This looks good to me - as in theory this shouldn't break anybody. Let's see integration tests pass and then we can merge it.

PR #1648, opened 2019-02-13
Add support for Kaniko AWS secret

fpighi:
Kaniko supports an AWS secret used to push/pull from ECR repositories.
This adds support for it to Skaffold.
I'm happy to review the design if necessary.
I tested this end to end in a custom build of mine against our internal ECR repository and it works fine.
CLA is in progress (we want to sign it as a company, so it's taking a bit more time).

### DIFF
--- a/pkg/skaffold/build/kaniko/secret.go
+++ b/pkg/skaffold/build/kaniko/secret.go
@@ -93,7 +93,7 @@ func (b *Builder) setupDockerConfigSecret(out io.Writer) (func(), error) {
 		logrus.Debug("No docker config specified. Checking for one in the cluster.")
 
 		if _, err := secrets.Get(b.DockerConfig.SecretName, metav1.GetOptions{}); err != nil {
-			return nil, errors.Wrap(err, "checking for existing kaniko secret")
+			return nil, errors.Wrap(err, "checking for existing kaniko docker secret")
 		}
 
 		return func() {}, nil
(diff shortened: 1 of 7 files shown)


pkg/skaffold/build/kaniko/secret.go
```diff
@@ -124,3 +124,53 @@ func (b *Builder) setupDockerConfigSecret(out io.Writer) (func(), error) {
 		}
 	}, nil
 }
+
+func (b *Builder) setupAWSSecretSecret(out io.Writer) (func(), error) {
```
r2d4: What is specific to AWS about this? Looks like the only thing thats different is the filename in the secret, which should already be configurable through the YAML (if not, I think thats the feature that we'd want instead of a separate code path).

fpighi: Thanks for the feedback. That makes sense.

Is something like this what you had in mind?

```
kaniko:
  # ...
  secrets:
    - path: <secret1-path>
      name: <secret1-name>
    - path: <secret2-path>
      name: <secret2-name>
```

If so I can give this a try in the next few days.

fpighi:
I'll close this as I'm working on an alternative PR to set arbitrary secrets as suggested above and another one to set arbitrary annotations (useful when authenticating Kaniko on EC2 via Kube2IAM).

Issue #1299, opened 2018-11-20
Document JIB maven and gradle on the new site


balopat:
ref #1060

patflynn:
@balopat 
@loosebazooka 

Should we assign this to someone on the jib team?

patflynn:
@balopat 
@loosebazooka 
@tadcordle 

friendly ping :)

loosebazooka:
Oh right, yeah, @TadCordle can probably own this for now, as he'll be doing the jib optimizations.

briandealwis:
I'm taking a stab at this.

briandealwis:
- [ ] create example multi-module projects too.
- [ ] document GCB + Jib

tejal29:
@briandealwis would you be able to do this by Nov 7th?

PR #1653, opened 2019-02-14
Explain what is required for multi-module builds

ahaeber:
This line about multi-module builds confused me because it seems like lot's of details was required. However, in the Gitter chat I was advised that it is only one field (module) that should be set. This update attempts to clarify this to help other folks with multi-module projects.

### DIFF
--- a/examples/jib/README.adoc
+++ b/examples/jib/README.adoc
@@ -19,7 +19,7 @@ build:
 Please note that this example is for a standalone Maven project, where
 all dependencies are resolved from outside. The Jib builder requires
 that the projects are configured to use the Jib plugins for Maven or Gradle.
-Multi-module builds require a bit additional configuration.
+For multi-module builds the `module` field should be set to the module which should be built.
 
 ifndef::env-github[]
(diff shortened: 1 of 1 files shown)


ahaeber:
I signed the CLA!

briandealwis:
Thank you @ahaeber for making this contribution.  Unfortunately I don't think this change is enough to help a newcomer: it's been a couple of months since I last fiddled with a multi-module project and I couldn't remember what the module field was supposed to be :-(   So I'm -1 to this change.

There's a long outstanding issue to add documentation for Jib (#1299) and I'll make a start on that today.

briandealwis:
@ahaeber please give a look at #1661 and suggest any improvements — it's already been merged, but happy to make any other changes.

ahaeber:
First of all, in this change I wanted to simply help clarify what is required for a multi-module build instead of leaving readers with more questions. #1661 looks good to me with details I have been missing to know about.

I think that a small example for a multi-module project would be very helpful. So instead of complicating the docs for this example we could refer readers to as separate multi-module example instead.

PR #1662, opened 2019-02-15
Added bazel in local execution environment

### LINKED ISSUE #1553: Implement Bazel as a builder plugin in local ExecutionEnvironment

priyawadhwa:
I added bazel as a builder plugin in the local execution environment.
This required moving DependenciesForArtifact into the build interface;
if people want to write their own plugins, they will need some way of
telling the skaffold watcher what their dependencies are.

fixes #1553


integration/examples/test-plugin/gcb/skaffold.yaml
```diff
@@ -10,10 +10,10 @@ build:
     context: ./leeroy-app/
     plugin: 
       name: docker
+      properties:
+        target: targetStage
   executionEnvironment:
     name: googleCloudBuild
-    properties:
```
balopat: why are we removing this?

priyawadhwa: I figured it would be easier for people to test this example if skaffold guessed the project from the image name instead of having to change this field in the skaffold.yaml every time

priyawadhwa:
I think the switch statement may still be the cleanest option, since build.DependenciesForArtifact applies to the overall "local" builder -- is that what you meant?

balopat:
> I think the switch statement may still be the cleanest option, since build.DependenciesForArtifact applies to the overall "local" builder instead of the different types of artifacts -- is that what you meant?

No, what I meant is that the builder itself now should know what the dependencies are for a given artifact. It is weird to see bazel.DependenciesForArtifact calling out to a global build.DependenciesForArtifact where it checks whether the artifact is docker, jib or bazel... Instead the logic should be something like this in bazel.DependenciesForArtifact: 

```
if a.BazelArtifact == nil {
  return nil, fmt.Errorf("%s is not a bazel artifact", a)  
}
return bazel.GetDependencies(ctx, a.Workspace, a.BazelArtifact)
```

integration/examples/test-plugin/gcb/skaffold.yaml
balopat: 👍

priyawadhwa:
@balopat thanks for clarifying, that should be done now!

pkg/skaffold/build/prebuilt.go
```diff
@@ -82,3 +82,7 @@ func (b *prebuiltImagesBuilder) Build(ctx context.Context, out io.Writer, _ tag.
 
 	return builds, nil
 }
+
+func (b *prebuiltImagesBuilder) DependenciesForArtifact(ctx context.Context, artifact *latest.Artifact) ([]string, error) {
+	return DependenciesForArtifact(ctx, artifact)
```
balopat: this can't be right - a prebuilt image should have no source dependencies.

balopat:
Great, almost there - I think we should go all the way, we can just nuke [deps.go](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/deps.go).

priyawadhwa:
So removing deps.go would cause code duplication of the switch statement since we need it [here](https://github.com/GoogleContainerTools/skaffold/blob/c289a6bfbba7b8643d90d08057579e8e3dc52d7a/pkg/skaffold/build/local/local.go#L92) for the local builder and [here](https://github.com/GoogleContainerTools/skaffold/blob/c289a6bfbba7b8643d90d08057579e8e3dc52d7a/cmd/skaffold/app/cmd/diagnose.go#L113) for the diagnose command.

I'm leaning towards keeping deps.go, WDYT?

balopat:
This looks great! 
1.) local.go will go away with all the builders turning into plugins. We can keep the switch logic there for now as it's an "obsolete" builder
2.) diagnose should actually delegate the call now to the builder that the runner created!

priyawadhwa:
Ahh got it that makes sense, thanks! I updated diagnose to do that :)

PR #1674, opened 2019-02-19
v1beta6

balopat:


pkg/skaffold/schema/v1beta5/upgrade.go
```diff
@@ -0,0 +1,77 @@
...
+		APIVersion: next.Version,
+		Kind:       config.Kind,
+		Build:      newBuild,
+		Test:       newTest,
+		Deploy:     newDeploy,
+		Profiles:   newProfiles,
+	}, nil
+}
+
+func convert(old interface{}, new interface{}) error {
```
priyawadhwa: This is the same as the function [CloneThroughJSON](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/util/util.go#L260)  I added to the util package for build plugins, maybe we should use that instead?

dgageot:
LGTM but needs to be rebased

docs/content/en/schemas/v1beta5.json
```diff
@@ -326,7 +326,7 @@
         },
         "image": {
           "type": "string",
-          "description": "Docker image used by the Kaniko pod. Defaults to the latest released version of <code>gcr.io/kaniko-project/executor</code>."
+          "description": "Docker image used by the Kaniko pod. Defaults to the v1beta5 released version of <code>gcr.io/kaniko-project/executor</code>."
```
dgageot: @balopat this is not supposed to be replaced :-)

balopat: LOL nice catch!!!!

pkg/skaffold/schema/v1beta5/config.go
```diff
@@ -0,0 +1,591 @@
...
+	// Namespace is the Kubernetes namespace.
+	// Defaults to current namespace in Kubernetes configuration.
+	Namespace string `yaml:"namespace,omitempty"`
+
+	// Timeout is the amount of time (in seconds) that this build is allowed to run.
+	// Defaults to 20 minutes (`20m`).
+	Timeout string `yaml:"timeout,omitempty"`
+
+	// Image is the Docker image used by the Kaniko pod.
+	// Defaults to the v1beta5 released version of `gcr.io/kaniko-project/executor`.
```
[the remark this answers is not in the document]
balopat: sigh. thanks for catching these!

pkg/skaffold/schema/v1beta5/upgrade.go
balopat: good catch, I think that's a good idea. we'll separate them if necessary.

pkg/skaffold/schema/v1beta5/config.go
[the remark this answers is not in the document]
balopat: actually, this was the source of the previous one through schema generation. Fixed!

PR #1675, opened 2019-02-20
Generate skaffold references

dgageot:


balopat:
Follow-up thoughts for next iterations (we don't have to do it in this PR, but we have to do it before the release)

 - we need to lift over all the comments from annotated-skaffold.yaml and then get rid of it, but I'm guessing that that's your plan. For example the 'test' section just says "describes how images are tested." - which is obviously not enough. 
- Also, we should turn off the "Edit page" button on the right hand side for this page and the CLI page as they are autogenerated

Issue #1668, opened 2019-02-17
Skaffold v0.23.0 + Windows 10 regression: didn't sync any files


WoLfulus:
### Expected behavior

Sync files when it changes

### Actual behavior

Did not sync files. Always shows "didn't sync any files"

### Information

- Skaffold version: v0.23.0
- Operating system: Windows 10
- Contents of skaffold.yaml:

```yaml
Any yamls I tried
```

### Steps to reproduce the behavior

1. Update to v0.23.0 from v0.22.0
2. Sync stops working

priyawadhwa:
Hey @WoLfulus, I tried to repro this with our [hot reload](https://github.com/GoogleContainerTools/skaffold/tree/master/examples/hot-reload) example and it worked as expected. Could you provide the skaffold.yaml you tried?

jonaseck2:
Can confirm, compiled tag v0.22.0 works, master and release 0.23.0 doesn't

v0.23.0:
Syncing 1 files for some_node_app:d7db341f059c1ea8eaf66d3c0c41593e724860471c7fa2bf84486ad16c28775a
Watching for changes every 1s...

Files aren't updated in the image at all

v.0.22.0
Syncing 1 files for some_node_app:03050b781def4228b6b64f6865293434c633d24393f395448df4bee9d1e309d7
Watching for changes every 1s...

updated, and hot reload works

skaffold yaml:
```
apiVersion: skaffold/v1beta4
kind: Config
build:
  tagPolicy:
    # gitCommit tags the image with the git commit of your current repository.
    #gitCommit: {}
    # sha256 tags the image with the checksum of the built image (image id).
    sha256: {}
  artifacts:
  - image: some_java_api
    context: ./api
#    jibGradle: {}
    # Skaffold can sync local files with remote pods (alpha) instead
    # of rebuilding the whole artifact's image. This is a mapping
    # of local files to sync to remote folders.
    # sync:
    #   '*.py': .
  - image: some_node_app
    context: ./app
    sync:
      '**/*.js': .
    docker:
      buildArgs:
        NODE_ENV: development
      target: build-deps
# The test section has all the information needed to test images.
test:
    # For each image listed here, Skaffold will run a series of structure tests using
    # the [Container Structure Tests](https://github.com/GoogleContainerTools/container-structure-test)
    # project.
    # - image: gcr.io/k8s-skaffold/skaffold-example
    #   structureTests:
    #     - ./test/*

[...]

PR #1680, opened 2019-02-20
Make the hot-reload example more exemplary

jonaseck2:
While investigating #1668 I found myself modifying the hot-reload example to actually test how sync works in 0.22, and doesn't in 0.23/master. The example is a little too succinct to understand how hot-reload works:

First off the example syncs exactly one source file. The Dockerfile does copy a glob, but the destination isn't a folder, so adding a second source file sibling crashes the build.

Secondly I was suspecting that nested folders might be the issue, so i added a src folder, which is probably closer to a real world example anyways. It also clarifies that the pattern isn't recursive like .gitignore (unless the folder name matches, then it's actually traversed), even though the k8s path does hint at that.

Now when the Dockerfile and sync pattern no longer exactly overlap, you can easily test that only files intersected by the Dockerfile and sync pattern are synced, files included by the Dockerfile but not the sync pattern fall back to docker build, and files matched by the pattern but outside the Dockerfile are ignored.

In hindsight it feels kinda obvious, but we'll probably end up with fewer issues on hot-reload with an example that is a little more verbose.

My comment on #1668 is probably a separate issue

/jonas

Signed-off-by: Jonas Eckerström <jonaseck@gmail.com>


jonaseck2:
I signed it!

dgageot:
CI is green despite the status reported on GitHub.

PR #1685, opened 2019-02-22
completion: add wrapper code to transform bash to zsh completion

### LINKED ISSUE #1671: ZSH completion load: can only be called from completion function
### Expected behavior

ZSH completion loading.

### Actual behavior

```
~ at ☸️  microk8s 
➜ source <(skaffold completion zsh)
_arguments:comparguments:325: can only be called from completion function
_arguments:comparguments:325: can only be called from completion function
```

### Information

- Skaffold version: v0.23.0
- Operating system: Linux sweethome 4.18.0-15-generic #16-Ubuntu SMP Thu Feb 7 10:56:39 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
- Contents of skaffold.yaml: Not related.

1. `source <(skaffold completion zsh)`

corneliusweig:
Cobra zsh completion is pretty broken. The idea is therefore, to use the working bash completion from cobra and wrap that with zsh adapter code. This code is based on the kubectl wrapper script.

I'm a bit unsure about the proper copyright notice. I left it as-is, but the beef of the logic comes from https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/completion/completion.go . Must I therefore include a proper attribution for the wrapper script?

Fixes #1671

### DIFF
--- a/cmd/skaffold/app/cmd/completion.go
+++ b/cmd/skaffold/app/cmd/completion.go
@@ -14,6 +14,26 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+/*
+NOTICE: The zsh wrapper code below is derived from the completion code
+in kubectl (k8s.io/kubernetes/pkg/kubectl/cmd/completion/completion.go),
+with the following license:
+
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


balopat:
Thank you for fixing this!!! 

> Must I therefore include a proper attribution for the wrapper script?
I think it's fine just as a reference to put a comment "inspired by kubectl completion" and a link

balopat:
On second thought - reading the Apache license maybe actually adding their copyright notice as attribution is not a bad idea. So in the end we would have 
* skaffold copyright notice at the top as it is now
* then a comment describing that this is a derivative work of the kubectl wrapper with link 
* then the copyright notice from the kubectl file

corneliusweig:
@balopat Ok, I added the attribution to kubernetes. Is the go package path ok as a link?

PR #1701, opened 2019-02-27
Fix schemas

dgageot:
I've tested the current schemas in multiple places and this change makes them work everywhere.

Also improves the rendering of the properties map on skaffold.dev. They are now displayed in the preferred order for easier reading.


dgageot:
Thank you @nkubala. I've addressed the nits.

PR #1702, opened 2019-02-27
Add debugging support for Skaffold: `skaffold debug`

briandealwis:
This PR adds a new command to Skaffold called `debug`.  `debug` provides a zero configuration solution for setting up containers for debugging.  It is a mix of `run` and `dev` in that it builds and deploys artifacts similar to `run`, but monitors and cleans-up like `dev`.  `debug` transforms the Kubernetes manifests to setup a container for debugging as required for the container's runtime technology.

Comments and suggestions are welcome.  I'm know my Go is terrible and can be improved.

### About `skaffold debug`

`skaffold debug` is like `skaffold dev` with the following differences:
  - Kubernetes pod-bearing objects that reference built images are transformed to setup runtime-specific debugging technologies:
      - JVM applications have a `JAVA_TOOL_OPTIONS` environment variable added/adjusted to configure the JDWP agent to listen for a specific port
      - NodeJS applications have a `--inspect=port` added to configure a DevTools connection
  - The associated debug ports are exposed and labelled
  - An annotation is added to the `PodSpec` metadata to describe the JDWP/DevTools port named `debug.cloud.google.com/config` which points to a string containing a JSON object mapping _container-name_ &rarr; a debug runtime configuration object.  Each debug configuration object has at least a `runtime` field that points to a simple ID describing the underlying language runtime technology (e.g., `jvm`, `nodejs`); other fields specify any configuration specific to that language runtime technology.  For example , a pod with two containers named `microservice` and `web` (line breaks added for legibility)
```
debug.cloud.google.com/config: '{
   "microservice":{"devtools":9229,"runtime":"nodejs"},
   "adapter":{"jdwp":5005,"runtime":"jvm"}
}'
```

#### Example transforms

For NodeJS, involving a PodSpec from `examples/hot-reload`.  The changes in this example:
  1. An annotation `debug.cloud.google.com/config` has a JSON object that describes the connection configuration details.  As Kubernetes _container_ objects don't carry metadata, we have to put this metadata on the container's parent; as a pod/podspec can have multiple containers, we use a key-value map, keyed by the container name.
  2. A new argument `--inspect=9229` is added to the node command line
  3. The devtools port 9229 is exposed

Note that we identify this artifact as Node based on the container attempting to run `nodemon`.  (`npm run xxx` is not handled as it doesn't support passing through `--inspect`.)

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    debug.cloud.google.com/config: '{"node":{"devtools":9229,"runtime":"nodejs"}}'
  creationTimestamp: null
  labels:
    cleanup: "true"
    docker-api-version: "1.38"
    skaffold-builder: local
    skaffold-deployer: kubectl
    skaffold-tag-policy: git-commit
    tail: "true"
  name: node
  namespace: default
spec:
  containers:
  - args:
    - nodemon
    - --inspect=9229
    - --legacy-watch
    - server.js
    image: gcr.io/k8s-skaffold/node-example:c42c314176d27cd0cfc1d7fbe983ab25f97c85dfcd7faf3000d86ba08bc5c3cd
    name: node
    ports:
    - containerPort: 3000
    - containerPort: 9229
      name: devtools
```

For Java (a deployment from `examples/jib`).  The changes in this example:
  1. An annotation `debug.cloud.google.com/config` has a JSON object that describes the connection configuration details.  As Kubernetes _container_ objects don't carry metadata, we have to put this metadata on the container's parent; as a pod/podspec can have multiple containers, we use a key-value map, keyed by the container name.
  2. A new environment variable `JAVA_TOOL_OPTIONS` is added; this is picked up by the JVM at launch.
  3. The JDWP port 5005 is exposed.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    cleanup: "true"
    docker-api-version: "1.38"
    skaffold-builder: local
    skaffold-deployer: kubectl
    skaffold-tag-policy: git-commit
    tail: "true"
  name: web
  namespace: default
spec:
  selector:
    matchLabels:
      app: web
  strategy: {}
  template:
    metadata:
      annotations:
        debug.cloud.google.com/config: '{"web":{"jdwp":5005,"runtime":"jvm"}}'
      creationTimestamp: null
      labels:
        app: web
        cleanup: "true"
        docker-api-version: "1.38"
        skaffold-builder: local
        skaffold-deployer: kubectl
        skaffold-tag-policy: git-commit
        tail: "true"
    spec:
      containers:
      - env:
        - name: JAVA_TOOL_OPTIONS
          value: -agentlib:jdwp=transport=dt_socket,server=y,address=5005,suspend=n,quiet=y
        image: gcr.io/k8s-skaffold/skaffold-jib:d6b74559a3f72c7ed03add0aaeed5bbddfd35be583b68db57896c24d61c29ce4
        name: web
        ports:
        - containerPort: 8080
        - containerPort: 5005
          name: jdwp
        resources: {}
```

### Change Overview

This implementation happens entirely at deploy time.  When I initially starting writing this code, my plan was to implement it as a mutating admission controller, where the code only had access to the images.  So this code walks the manifests to be deployed and examines and transforms any containers referenced in a pod/podspec that use a supported language runtime.

We guess the runtime technology for an image container by examining the referenced image container configuration, specifically looking at the environment variables and command-line.  It turns out that most of the language runtime base images  [define a `XXXX_VERSION` environment variable](https://github.com/GoogleContainerTools/distroless/issues/276#issuecomment-460375014) or use a well-known command-name (`java` or `node` or `nodemon`).

- most debugging transform code is in `pkg/skaffold/debugging`
- adds a hook to `deploy/kubectl.go` to configure a manifest transformer (`ManifestList`)
- debugging's manifest transformer (`ApplyDebuggingTransforms` in `pkg/skaffold/debugging/debug.go`) walks the Kubernetes objects and applies changes to podspecs based on the guessed runtime technology
- _Update:_ **{{Skaffold Change}}**  A `skaffold/build/Artifact` alone isn't enough to introspect on the corresponding container image as it may have been loaded into the local docker daemon or pushed to  a registry. I changed `skaffold/build/Artifact` to include a `Location` enum, and changed the Builder interface to return an Artifact with the location, rather than just return the built tag.
- guessing the runtime technology is a bit involved as it requires retrieving the container config and looking at the environment and command-line
   -~~ the container image's config (`go-containerregistry/pkg/v1/config/Config`) is contained by re-defining the `skaffold/build/Artifact` to also provide a _configuration retriever_  function.~~ _Update: now entirely hidden inside the `debugging` package._  There are two retriever implementations, one that fetches from a registry and the other from the Docker daemon.  This must be determined at build-time as it depends on the local-push.
  - passing this configuration retriever back is a bit … ugly
- the transforms examine the pod's ports to allocate a non-conflicting port that is close to the standard port (e.g., 5005 for JDWP, 9229 for NodeJS DevTools)

### To Do

I'd prefer to leave these follow-ups to separate PRs:
- [ ] Add other languages: these will require installing files at startup.  The plan is to use an `initContainer` that copies files into an `emptyDir` volume.
  - [ ] python requires [ptvsd](https://github.com/Microsoft/ptvsd/)
  - [ ] go requires providing delve and/or gdb
  - [ ] .NET
  - [ ] will need to make this initContainer location configurable (e.g., for air-gapped clusters)
- [ ] Re-enable watching: we don't want to kill containers, but file changes are ok.  So for artifacts with debugging, we should continue to sync, for all other artifacts we should continue with watching.  But it's unclear how to twiddle the file-watching setting for an artifact from the manifest transforms. Hmm, perhaps I should front-load some of this into the `cmd/debug.go`.
- [ ] Add explicit artifact runtime configuration: requires changing the `skaffold.yaml` and associated objects.


cmd/skaffold/app/cmd/debug.go
```diff
@@ -0,0 +1,105 @@
...
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return debug(out)
+		},
+	}
+	AddRunDebugFlags(cmd)
+	cmd.Flags().BoolVar(&opts.TailDev, "tail", true, "Stream logs from deployed objects")
+	cmd.Flags().StringVar(&opts.Trigger, "trigger", "polling", "How are changes detected? (polling or manual)")
+	cmd.Flags().BoolVar(&opts.Cleanup, "cleanup", true, "Delete deployments after dev mode is interrupted")
+	cmd.Flags().StringArrayVarP(&opts.TargetImages, "watch-image", "w", nil, "Choose which artifacts to watch. Artifacts with image names that contain the expression will be watched only. Default is to watch sources for all artifacts")
```
tejal29: The help mentions, Artifacts that are debugged are not watched. Here it mentions, "by default all sources are watched". 
These contradict each other

cmd/skaffold/app/cmd/debug.go
```diff
@@ -0,0 +1,105 @@
...
+		Use:   "debug",
+		Short: "Runs a pipeline file in debug mode",
+		Long:  "Similar to `dev`, but runs the pipeline for debugging. Artifacts that are debugged are not watched.",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return debug(out)
+		},
+	}
+	AddRunDebugFlags(cmd)
+	cmd.Flags().BoolVar(&opts.TailDev, "tail", true, "Stream logs from deployed objects")
```
dgageot: Could this duplication be removed?

pkg/skaffold/deploy/kubectl/debug.go
```diff
@@ -0,0 +1,316 @@
...
+)
+
+// portAllocator is a function that takes a desired port and returns an available port
+// Ports are normally uint16 but Kubernetes ContainerPort.containerPort is an integer
+type portAllocator func(int32) int32
+
+// ApplyDebuggingTransforms applies language-platform-specific transforms to a list of manifests.
+func ApplyDebuggingTransforms(l ManifestList, builds []build.Artifact) (ManifestList, error) {
+	var updated ManifestList
+	//decode := api.Codecs.UniversalDeserializer().Decode
```
tejal29: remove commented lines

cmd/skaffold/app/cmd/debug.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	output := &config.Output{
+		Main: out,
+		Logs: out,
+	}
+
+	// HACK: prevent redeploying changed containers during debugging
+	// TODO: build to ignore debuggable artifacts
+	if len(opts.TargetImages) == 0 {
```
dgageot: This is to deactivate the watcher altogether, right?

pkg/skaffold/deploy/kubectl/debug.go
```diff
@@ -0,0 +1,316 @@
...
+	switch guessRuntime(config) {
+	case JVM:
+		logrus.Debugf("Configuring %v for JVM", container.Name)
+		return configureJvmDebugging(container, config, portAlloc)
+	case NODEJS:
+		logrus.Debugf("Configuring %v for NODEJS", container.Name)
+		return configureNodeJSDebugging(container, config, portAlloc)
+	default:
+		logrus.Debugf("Unable to determine runtime for %v\n", container.Name)
+		return nil
```
tejal29: maybe return an error here so that skaffold debug fails for unsupported runtimes instead of silently passing.

pkg/skaffold/deploy/kubectl/debugjvm.go
```diff
@@ -0,0 +1,175 @@
...
+
+	var port int32
+	if spec != nil {
+		port = int32(spec.port)
+	} else {
+		// no standard port for JDWP; most examples use 5005 or 8000
+		port = portAlloc(5005)
+
+		javaToolOptions := v1.EnvVar{
+			Name:  "JAVA_TOOL_OPTIONS",
```
tejal29: shd we make these constants?

tejal29:
Great PR!!

cmd/skaffold/app/cmd/debug.go
briandealwis: I'll remove the help description: no-watch is an to simplify the implementation.  And we may watch in the future, and instead not redeploy on change.

cmd/skaffold/app/cmd/debug.go
briandealwis: Yes

briandealwis:
Marking this is ready for review: PTAL.  I've updated the description at the top; the other items will come as separate PRs.  

I'll add some docs in the meantime to describe what is and isn't supported.

briandealwis:
(Failure its because docs out of date)

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -154,6 +155,17 @@ func AddRunDevFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&opts.CacheFile, "cache-file", "", "", "Specify the location of the cache file (default $HOME/.skaffold/cache)")
 }
 
+func AddDevDebugFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&opts.TailDev, "tail", true, "Stream logs from deployed objects")
+	cmd.Flags().StringVar(&opts.Trigger, "trigger", "polling", "How are changes detected? (polling, manual or notify)")
```
tejal29: Can we change this to be a descriptive help instead of a question.
" Mode for detecting changes. One of [polling, manual or notification]."

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -154,6 +155,17 @@ func AddRunDevFlags(cmd *cobra.Command) {
...
+	cmd.Flags().BoolVar(&opts.Cleanup, "cleanup", true, "Delete deployments after dev mode is interrupted")
```
tejal29: its not clear whether setting it to true will delete. Can we change this to "if true, deletes deployments after dev mode is interrupted"

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -154,6 +155,17 @@ func AddRunDevFlags(cmd *cobra.Command) {
...
+	cmd.Flags().StringArrayVarP(&opts.TargetImages, "watch-image", "w", nil, "Choose which artifacts to watch. Artifacts with image names that contain the expression will be watched only. Default is to watch sources for all artifacts")
+	cmd.Flags().IntVarP(&opts.WatchPollInterval, "watch-poll-interval", "i", 1000, "Interval (in ms) between two checks for file changes")
```
tejal29: Can we also mention, this flag is respected when "trigger is set to polling"

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -154,6 +155,17 @@ func AddRunDevFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&opts.CacheFile, "cache-file", "", "", "Specify the location of the cache file (default $HOME/.skaffold/cache)")
 }
 
+func AddDevDebugFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&opts.TailDev, "tail", true, "Stream logs from deployed objects")
+	cmd.Flags().StringVar(&opts.Trigger, "trigger", "polling", "How are changes detected? (polling, manual or notify)")
+	cmd.Flags().BoolVar(&opts.Cleanup, "cleanup", true, "Delete deployments after dev mode is interrupted")
+	cmd.Flags().StringArrayVarP(&opts.TargetImages, "watch-image", "w", nil, "Choose which artifacts to watch. Artifacts with image names that contain the expression will be watched only. Default is to watch sources for all artifacts")
```
tejal29: Its confusing, the option name is "watch-image" but the description says "Choose **artifacts** .."
Consider using the same term i.e. either artifact or image in option name and help

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -154,6 +155,17 @@ func AddRunDevFlags(cmd *cobra.Command) {
...
+	cmd.Flags().IntVarP(&opts.WatchPollInterval, "watch-poll-interval", "i", 1000, "Interval (in ms) between two checks for file changes")
+	cmd.Flags().BoolVar(&opts.PortForward, "port-forward", true, "Port-forward exposed container ports within pods")
```
tejal29: mention if ports are exposed is done when it is set true.
"if True, port forward exposed container ports within pods."

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -154,6 +155,17 @@ func AddRunDevFlags(cmd *cobra.Command) {
...
+	cmd.Flags().StringArrayVarP(&opts.CustomLabels, "label", "l", nil, "Add custom labels to deployed objects. Set multiple times for multiple labels")
```
tejal29: Can you also mention how we can specify multiple labels.
"Adds custom labels to deployed objects. To specify multiple labels, repeat the option per label"

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -154,6 +155,17 @@ func AddRunDevFlags(cmd *cobra.Command) {
...
+	cmd.Flags().BoolVar(&opts.ExperimentalGUI, "experimental-gui", false, "Experimental Graphical User Interface")
```
tejal29: If this an advanced user option, can we hide it using the flags.MarkHidden("experimental-gui") if not, can we provide more helpful description.

testutil/util.go
```diff
@@ -28,6 +28,13 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
+func CheckEqual(t *testing.T, options cmp.Options, expected, actual interface{}) {
```
tejal29: Rename this to more descriptive name like CheckDeepEqualWithOptions

cmd/skaffold/app/cmd/cmd.go
balopat: This is a great suggestion - however it was not introduced in this PR - we can raise that as a separate contribution.

balopat:
Not sure if this is related - but I started testing on jib-multimodule and the port allocation logic seems to allocate 5005 for both web apps:

````
deployment.apps/web1 created
deployment.apps/web2 created
Deploy complete in 4.101437229s
Watching for changes every 1s...
Forwarding container web2 to local port 0.
Port Forwarding web2-7d8574bd5-vwzt4/web2 8080 -> 0
Port Forwarding web2-7d8574bd5-vwzt4/web2 5005 -> 5005
[web2-7d8574bd5-vwzt4 web2] Picked up JAVA_TOOL_OPTIONS: -agentlib:jdwp=transport=dt_socket,server=y,address=5005,suspend=n,quiet=y
Forwarding container web1 to local port 0.
Port Forwarding web1-d65cfbf67-dttrd/web1 8080 -> 0
Forwarding container web1 to local port 0.
Port Forwarding web1-d65cfbf67-dttrd/web1 5005 -> 0
````
Also web1 8080 and 5005 are forwarded to 0...

**Comment by briandealwis** [comment]
The `jib-multimodule` example sets up two separate deployments, so the two Java containers are deployed to separate pods.  So allocating port 5005 is fine.

The mapping to port 0 is odd though, but I haven't changed anything there.  When I merge with master, I don't see this problem but I do see port 5005 being mapped to port 5005 on the localhost:
```
Deploy complete in 365.125704ms
Watching for changes every 1s...
Port Forwarding web2-b8c44c668-bbt5t/web2 8080 -> 8080
INFO[0015] Stream logs from pod: web2-b8c44c668-bbt5t container: web2 
DEBU[0015] Port forwarding web2-b8c44c668-bbt5t/web2:8080 
DEBU[0015] Running command: [kubectl logs --since=2s -f web2-b8c44c668-bbt5t -c web2 --namespace default] 
Port Forwarding web2-b8c44c668-bbt5t/web2 5005 -> 5005
DEBU[0015] Port forwarding web2-b8c44c668-bbt5t/web2:5005 
INFO[0015] Stream logs from pod: web1-6f448c9679-rwqfg container: web1 
DEBU[0015] Running command: [kubectl logs --since=3s -f web1-6f448c9679-rwqfg -c web1 --namespace default] 
Port Forwarding web1-6f448c9679-rwqfg/web1 8080 -> 8080
DEBU[0015] Port forwarding web1-6f448c9679-rwqfg/web1:8080 
Port Forwarding web1-6f448c9679-rwqfg/web1 5005 -> 5005
```

**Comment by briandealwis** [comment]
> @tejal29: _I added a bunch of nits, but then realized, you just copied the flags over from skaffold dev._

I'll leave these for now, especially since they're repeated in other places.  But I will move the watching-related options back to `dev` since watching isn't supported in `debug` (yet).

PTAL @balopat: the port issues were unrelated and fixed in #1780

**Comment by briandealwis** [comment]
What are _controller logs_ and how do I go about setting them? Or are they the Kokoro logs?

(The Kokoro failure is an unrelated kaniko test time out.)

**Comment by balopat** [comment]
Nah, it's our docs previewer acting up...don't worry about this - I'll fix this on our end.

File: pkg/skaffold/debugging/transformjvm.go [hunk 15]
```diff
@@ -0,0 +1,202 @@
...
+
+func (t jdwpTransformer) IsApplicable(config imageConfiguration) bool {
+	if _, found := config.env["JAVA_TOOL_OPTIONS"]; found {
+		return true
+	}
+	if _, found := config.env["JAVA_VERSION"]; found {
+		return true
+	}
+	if len(config.entrypoint) > 0 {
+		if config.entrypoint[0] == "java" || strings.HasSuffix(config.entrypoint[0], "/java") {
```
dgageot: `return config.entrypoint[0] == "java" || strings.HasSuffix(config.entrypoint[0], "/java")`

pkg/skaffold/debugging/transformjvm.go
```diff
@@ -0,0 +1,202 @@
...
+			return true
+		}
+	} else if len(config.arguments) > 0 {
```
dgageot: remote the `else`

pkg/skaffold/debugging/transformjvm.go
```diff
@@ -0,0 +1,202 @@
...
+		if config.arguments[0] == "java" || strings.HasSuffix(config.arguments[0], "/java") {
```
dgageot: `return config.arguments[0] == "java" || strings.HasSuffix(config.arguments[0], "/java")`

pkg/skaffold/debugging/transformjvm.go
```diff
@@ -0,0 +1,202 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+func extractJdwpArg(spec string) *jdwpSpec {
+	if strings.Index(spec, "-agentlib:jdwp=") == 0 {
+		return parseJdwpSpec(spec[15:])
+	} else if strings.Index(spec, "-Xrunjdwp:") == 0 {
```
dgageot: remove the `else`

pkg/skaffold/debugging/transformjvm.go
```diff
@@ -0,0 +1,202 @@
...
+		return parseJdwpSpec(spec[10:])
+	}
+	return nil
+}
+
+func (spec jdwpSpec) String() string {
+	result := "transport=" + spec.transport
```
dgageot: maybe build a []string, joined with a comma

pkg/skaffold/debugging/debug.go
```diff
@@ -0,0 +1,119 @@
...
+		if err := s.Encode(o, w); err != nil {
+			return nil, err
+		}
+		w.Flush()
+		return b.Bytes(), nil
+	}
+)
+
+// ApplyDebuggingTransforms applies language-platform-specific transforms to a list of manifests.
+func ApplyDebuggingTransforms(l kubectl.ManifestList, builds []build.Artifact) (kubectl.ManifestList, error) {
```
dgageot: Are there unit tests for `ApplyDebuggingTransforms`?

pkg/skaffold/build/build.go
```diff
@@ -21,13 +21,21 @@ import (
...
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 // Artifact is the result corresponding to each successful build.
 type Artifact struct {
 	ImageName string
 	Tag       string
+	Config    ConfigurationRetriever
```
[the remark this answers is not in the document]
dgageot: Anyhow, builders should maybe return some else that a `string`

cmd/skaffold/app/cmd/debug.go
```diff
@@ -0,0 +1,99 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return debug(out)
+		},
+	}
+	AddRunDevFlags(cmd)
+	AddDevDebugFlags(cmd)
+	return cmd
+}
+
+func debug(out io.Writer) error {
```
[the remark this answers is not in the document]
briandealwis: looks like I can delegate everything to dev().

pkg/skaffold/build/build.go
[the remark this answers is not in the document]
briandealwis: Urp! This is a really good point.

briandealwis:
Ok, I've updated the patch with feedback from reviews.  I made some changes to how Skaffold managed information about the build artifacts:
  - I modified `skaffold/build/Artifact` to track whether the artifact was built to the local docker daemon or to a registry (I need to this to be able to examine the resulting container image)
  - I modified the `Builder` interface to return a `skaffold/build/Artifact` instead of the resulting build tag, rather than have the instances built inside the parallel and sequential builders (which seemed an odd location!).  (This should be fine on the builder-plugins since they were reusing the same parallel/sequential builders?)
  - moved the container image configuration retrieval into the `debugging` package

PTAL!

pkg/skaffold/build/build.go
```diff
@@ -41,3 +49,54 @@ type Builder interface {
...
+//
+// - RegistryConfigurationRetriever for querying a registry based on the provided image reference
+//
+// - DockerConfigurationRetriever for querying the local docker daemon's cache
+type ConfigurationRetriever func(ctx context.Context) (config.Config, error)
+
+// RegistryConfigurationRetriever returns a function to retrieve an image configuration from a registry
+func RegistryConfigurationRetriever(image string) ConfigurationRetriever {
+	return func(ctx context.Context) (config.Config, error) {
+		logrus.Debugf("Retrieving image configuration for %v", image)
```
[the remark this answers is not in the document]
briandealwis: I've moved and simplified this code into the `debugging` package.

cmd/skaffold/app/cmd/debug.go
```diff
@@ -0,0 +1,53 @@
...
+	// HACK: disable watcher to prevent redeploying changed containers during debugging
+	// TODO: avoid redeploys of debuggable artifacts, but still enable file-sync
```
balopat: file sync does not work as such - the comment is a bit misleading

pkg/skaffold/debugging/transformjvm.go
```diff
@@ -0,0 +1,200 @@
+/*
```
balopat: nit: `transform_jvm.go`

pkg/skaffold/debugging/transform.go
```diff
@@ -0,0 +1,201 @@
...
+		if desiredPort < 1024 || desiredPort > maxPort {
+			desiredPort = 1024 // skip reserved ports
+		}
+		windowSize := maxPort - desiredPort + 1
+		// check pod containers for the next 20 ports
+		if windowSize > 20 {
+			windowSize = 20
+		}
+		var allocated = make([]bool, windowSize)
+		for _, container := range podSpec.Containers {
```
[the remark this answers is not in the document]
balopat: I'm not sure if there is a need for the window size then either, you can just walk the range of `max(desiredPort + i mod 65535, 1024)`

pkg/skaffold/debugging/transformnodejs.go
```diff
@@ -0,0 +1,166 @@
+/*
```
balopat: nit: `transform_nodejs.go`

pkg/skaffold/debugging/transformnodejs.go
```diff
@@ -0,0 +1,166 @@
...
+	v1 "k8s.io/api/core/v1"
+)
+
+type nodeTransformer struct{}
+
+func init() {
+	containerTransforms = append(containerTransforms, nodeTransformer{})
+}
+
+// captures the useful nodejs devtools options
```
balopat:
```suggestion
// inspectSpec captures the useful nodejs devtools options
```

pkg/skaffold/debugging/transformnodejs.go
```diff
@@ -0,0 +1,166 @@
...
+func (t nodeTransformer) Apply(container *v1.Container, config imageConfiguration, portAlloc portAllocator) map[string]interface{} {
+	logrus.Infof("Configuring [%s] for node.js debugging", container.Name)
+
+	// try to find existing `--inspect` command
+	spec := retrieveNodeInspectSpec(config)
+	// todo: find existing containerPort "devtools" and use port. But what if it conflicts with command-line spec?
+
+	if spec == nil {
+		// most examples use 9229
+		spec = &inspectSpec{port: portAlloc(9229)}
```
balopat: nit: extract to constant `defaultDebugPort` (with comment)

docs/content/en/docs/how-tos/debug/_index.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Limitations
+
+`skaffold debug` has some limitations:
+
+  - Only JVM and NodeJS applications are supported:
+      - JVM applications are configured using the `JAVA_TOOL_OPTIONS` environment variable
+        which causes extra debugging output on launch.
+      - NodeJS applications must be launched using `node` or `nodemon`
+  - File watching is disabled for all artifacts, irregardless of whether
```
balopat:
```suggestion
  - File watching is disabled for all artifacts, regardless of whether
```

docs/content/en/docs/how-tos/debug/_index.md
```diff
@@ -0,0 +1,46 @@
...
+  - a JDWP agent is configured for Java applications,
+  - the Chrome DevTools inspector is configured for NodeJS applications.
+      
+`skaffold debug` uses a set of heuristics to identify the runtime technology.
+The Kubernetes manifests are transformed on-the-fly such that the on-disk
+representations are untouched. 
+
+## Limitations
+
+`skaffold debug` has some limitations:
```
balopat: only kubectl is supported - helm and kustomize are not yet available

pkg/skaffold/deploy/kubectl.go
```diff
@@ -93,6 +101,13 @@ func (k *KubectlDeployer) Deploy(ctx context.Context, out io.Writer, builds []bu
 		return errors.Wrap(err, "setting labels in manifests")
 	}
 
+	for _, transform := range manifestTransforms {
```
balopat: we could theoretically add the same thing to kustomize deployer as well - but it is okay to follow-up with a new PR. 
helm will be a bit trickier.

pkg/skaffold/debugging/debug_test.go
```diff
@@ -0,0 +1,489 @@
...
+func TestFindArtifact(t *testing.T) {
+	buildArtifacts := []build.Artifact{
+		{ImageName: "image1", Tag: "tag1"},
+	}
+	tests := []struct {
+		description string
+		source      string
+		returnNil   bool
+	}{
+		{description: "found",
```
balopat:
```suggestion
		{
		    description: "found",
```

pkg/skaffold/debugging/debug_test.go
```diff
@@ -0,0 +1,489 @@
...
+			source:    "image1",
+			returnNil: false,
+		},
+		{description: "not found",
```
balopat:
```suggestion
		{
		    description: "not found",
```

cmd/skaffold/app/cmd/debug.go
briandealwis: Reworded as `TODO: enable file-sync but avoid redeploys of artifacts being debugged`

etanshaul:
Thanks Brian. The logic lgtm. I’ll defer to the Skaffold folks for the golang specifics.

briandealwis:
Thanks @balopat for the thorough review.

> It is a little bit annoying how the multiple containers can race for ports so sometimes web1 gets the 8081 sometimes web2.

IIUC, this is related to the port-forwarding of declared container ports to the local host, right?  
It would be even worse if we don't rewrite the replicas to 1!  I see this as a tooling issue — the GWT tooling for Eclipse, for example, provides a view with an annotated list of URLs that a developer might want to examine.  We should have enough information from the events to do something similar.

> sync and rebuild would be actually nice - now the iteration requires restarting skaffold debug - which I usually do by changing the yaml file :)

+1

pkg/skaffold/build/build.go
```diff
@@ -24,10 +24,19 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 )
 
+// Location describes the location to which a container image was built
+type ImageLocation int
```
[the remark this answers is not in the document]
briandealwis: Oh!  If the image may be in both the local daemon and the remote registry, and presumably querying the local daemon should be quick, I'll just query the local daemon first and then try the remote.

Then the question is: should I back out the location and builder API changes?  I think having the parallel and sequential builders creating the actual build.Artifact is a bit odd.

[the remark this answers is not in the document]
balopat: Good question - let's try the local first +remote fallback direction - it should be okay to query the local daemon for image existence and then, I would revert the location part - and although having the builders return a build.artifact might not be a bad idea - we are not using it yet, so "YAGNI"...

pkg/skaffold/debugging/transform.go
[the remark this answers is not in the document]
briandealwis: Very true: container port space seem to be sparse.  Will radically simplify.

pkg/skaffold/debugging/debug.go
```diff
@@ -0,0 +1,177 @@
...
+			logrus.Debugf("Found artifact for image [%s]", image)
+			return &artifact
+		}
+	}
+	return nil
+}
+
+// retrieveImageConfiguration retrieves the image container configuration for
+// the given build artifact
+func retrieveImageConfiguration(image string, artifact *build.Artifact) (imageConfiguration, error) {
```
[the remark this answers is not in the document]
briandealwis: Oh!  Indeed it is: https://github.com/GoogleContainerTools/skaffold/blob/9e23d320369b1586d13d6dddcf27023fb5929862/pkg/skaffold/docker/image.go#L111

pkg/skaffold/build/build.go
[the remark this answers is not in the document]
briandealwis: It turns out that requesting the config from the local docker API Client tries fetching from the remote registry does the trick.  I'll back out these changes.

briandealwis:
PTAL @balopat

PR #1707, opened 2019-02-28
Move bazel code into plugins directory

priyawadhwa:
Refactoring to move as much bazel code into the plugins directory as possible, so it should be easier to remove the old local builder later on.

continues work on #1765


priyawadhwa:
Let's hold off on merging this until the package structure is confirmed --

priyawadhwa:
CI actually passed (red because of the netlify checks) so I'm gonna go ahead and merge this.

PR #1717, opened 2019-03-02
Refactor local builder docker code into plugins directory

priyawadhwa:
Move all docker builder code in plugins directory; have current local
builder code call out to the plugin builder.

first step in #1765


priyawadhwa:
@nkubala I think I was able to fix it -- this should be RFAL. 

cc @balopat

PR #1723, opened 2019-03-04
Bail out on docker build error

### LINKED ISSUE #1712: Skaffold ignores errors in Dockerfiles
### Expected behavior
Give an error information when there are errors in the Dockerfile:
image

### Actual behavior
No explicit information showed
 
image

- Skaffold version: v0.23.0
-  Operating system: macos
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta5
kind: Config
build:
  tagPolicy:
    sha256: {}
  artifacts:
    - image: my-image-name
```

1. An incorrect Dockerfile

michaelfig:
Fixes #1712

### DIFF
--- a/pkg/skaffold/build/local/docker.go
+++ b/pkg/skaffold/build/local/docker.go
@@ -45,7 +45,7 @@ func (b *Builder) buildDocker(ctx context.Context, out io.Writer, workspace stri
 		imageID, err = b.localDocker.Build(ctx, out, workspace, a, tag)
 	}
 
-	if b.pushImages {
+	if err == nil && b.pushImages {
 		return b.localDocker.Push(ctx, out, tag)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/skaffold/build/local/docker.go
```diff
@@ -45,7 +45,7 @@ func (b *Builder) buildDocker(ctx context.Context, out io.Writer, workspace stri
 		imageID, err = b.localDocker.Build(ctx, out, workspace, a, tag)
 	}
 
-	if b.pushImages {
+	if err == nil && b.pushImages {
```
priyawadhwa: Instead of checking `if err==nil` , let's instead check

```golang
if err != nil {
   return "", err
}
```
before pushing, and then at the end of the function we can

```golang
return imageID, nil
```

michaelfig: Done.

dgageot:
I'll merge this and add an integration test

PR #1736, opened 2019-03-06
Upgrade Kaniko to 0.9.0

dgageot:

### DIFF
--- a/pkg/skaffold/constants/constants.go
+++ b/pkg/skaffold/constants/constants.go
@@ -45,7 +45,7 @@ const (
 
 	DefaultKustomizationPath = "."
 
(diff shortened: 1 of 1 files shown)


dgageot:
@priyawadhwa any idea why it's failing on kokoro?

priyawadhwa:
@dgageot I can't think of any reason why it would fail; I tried running it locally and it passed -- I feel like it may have been a weird one-off error? (since `TestRun/kaniko_local_-_sub_folder` passed)

Only thing I can think of is to just rerun the tests but that's maybe not the greatest solution...

balopat:
I retriggered the build...let's see what it's going to do. it feels like a flake.

balopat:
added the flake as an issue: https://github.com/GoogleContainerTools/skaffold/issues/1746

PR #1741, opened 2019-03-06
Improve artifact caching

### LINKED ISSUE #1740: Improve artifact caching
Based on the following feedback from @dgageot, there are a few things we can do to make artifact caching better:

TODO:
- [x] Even if there is no local daemon, check if images exist remotely
- [x] Include filenames in hash
- [x] Sort dependencies
- [x] Loop on all images only once (#1743)
- [x] Don't create a function/file for the hasher
- [ ] skaffold cache command to add images built by others to the cache (not doable without tagging with hash)
- [x] images built locally but pushed (push:true) are pushed each time
- [x] sometimes images cannot be retagged:
      - something like `rm ~/.skaffold/cache && skaffold build --cache-artifacts=true` in `examples/kaniko` should reproduce the issue
- [x] Changes to the builder config or the artifact config should invalidate the cache
- [x] only get image list from local docker daemon once per build loop
- [x] check for remote images in parallel
- [x] allow caching to work with remote builds (gcb, in-cluster via kaniko)
- [ ] store artifacts in cache as they are built
- [ ] tags provided with `--images` to `skaffold deploy` are kinda ignored.
- [x] Should not show a red error if an image is not already built. (`Unable to retrieve ... from cache; this image will be rebuilt.`)
- [x] TestNewCache unit test shouldn't depend on cluster the user is running
- [x] In a fully local environment, using minikube or Docker Desktop, and with `push` either undefined or set to `false`, we should NOT check for remote digests.

priyawadhwa:
some improvements to artifact caching:

1. Include filename in the cache key generation
2. Sort dependencies to ensure the key is reproducible
3. Add unit tests for both of these additions

Also, simplify the cache hasher function

begins to fix #1740

### DIFF
--- a/pkg/skaffold/build/cache.go
+++ b/pkg/skaffold/build/cache.go
@@ -27,6 +27,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"time"
 
 	skafconfig "github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
@@ -42,6 +43,11 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	yaml "gopkg.in/yaml.v2"
+)
+
+var (
+	// For testing
+	hashFunction = cacheHasher
 )
 
 // ImageDetails holds the Digest and ID of an image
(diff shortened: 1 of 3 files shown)


dgageot:
Sorry @priyawadhwa, this needs to be rebased

PR #1750, opened 2019-03-07
Simplify integration tests

dgageot:
+ Move `Gcloud` command to the Makefile
 + Remove TestMain. It makes it complicated to run a single test
 + Replace build tags with a check on `-short` so that tests are always compiled and linted
 + Fix lint errors on rpc_test.go
 + Some test projects should not be in ./integration/examples because they shouldn’t be in ./examples
 + Fix flakiness on TestDev


integration/dev_test.go
```diff
@@ -29,38 +27,44 @@ import (
...
+	deployName := "test-dev"
+	if err := kubernetesutil.WaitForDeploymentToStabilize(context.Background(), client, ns.Name, deployName, 10*time.Minute); err != nil {
+		t.Fatalf("Timed out waiting for deployment to stabilize")
 	}
 
-	job, err := Client.BatchV1().Jobs(ns.Name).Get(jobName, meta_v1.GetOptions{})
+	dep, err := client.AppsV1().Deployments(ns.Name).Get(deployName, meta_v1.GetOptions{})
 	if err != nil {
-		t.Fatalf("Could not find job: %s %s", ns.Name, jobName)
+		t.Fatalf("Could not find dep: %s %s", ns.Name, deployName)
```
balopat:
```suggestion
		t.Fatalf("Could not find deployment: %s %s", ns.Name, deployName)
```

PR #1754, opened 2019-03-07
Fix bug in jib in GCB

priyawadhwa:
Dependencies for any non-Docker artifact type in GCB aren't being
calculated, resulting in a panic:

```
jib-multimodule$ skaffold dev --default-repo gcr.io/priya-wadhwa -p gcb
ERRO[0000] failed to start grpc server: grpc: the server has been stopped
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xaa5cfc]

goroutine 1 [running]:
github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker.GetDependencies(0x17ed0c0, 0xc000214640, 0x1621ffd, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	pkg/skaffold/docker/parse.go:282 +0x8c
```

this should fix the panic, although it seems like jibMaven in GCB may
not be working, potentially related to this
[issue](https://github.com/GoogleContainerTools/jib/issues/1500).


dgageot:
@priyawadhwa something fail on AppVeyor. Could you also add an integration test for jib on GCB?

PR #1773, opened 2019-03-11
Faster doc preview

dgageot:
Clone the docs theme inside a Docker image, not on the developer's machine.
Instead of mounting the whole workspace, we mount only the relevant doc sources.

This makes working on the documentation much faster, gets rid of the submodules, doesn't require a cleanup phase...

### DIFF
--- a/deploy/docs/build.sh
+++ b/deploy/docs/build.sh
@@ -19,7 +19,7 @@ set -exu
 readonly BASE_URL=${1}
 
 cd docs
-mkdir themes
+mkdir -p themes
 ln -s /app/docs/themes/docsy ./themes/docsy
 ln -s /app/docs/node_modules ./node_modules
 hugo --baseURL=${BASE_URL}
--- a/deploy/docs/build.sh
+++ b/deploy/docs/build.sh
@@ -19,7 +19,7 @@ set -exu
 readonly BASE_URL=${1}
 
 cd docs
-mkdir -p themes
+mkdir themes
 ln -s /app/docs/themes/docsy ./themes/docsy
 ln -s /app/docs/node_modules ./node_modules
 hugo --baseURL=${BASE_URL}
--- a/deploy/docs/build.sh
+++ b/deploy/docs/build.sh
@@ -14,21 +14,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -exu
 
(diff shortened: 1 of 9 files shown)


deploy/docs/build.sh
```diff
@@ -14,21 +14,12 @@
...
-
-git submodule init && \
-git submodule update --init && \
-cd  docs/themes/docsy && \
-git checkout ${DOCSY_COMMIT} && \
-git submodule update --init --recursive && \
-cd  ${REPO_DIR}/docs && \
-npm i -D autoprefixer && \
+cd docs
+mkdir themes
```
balopat:
```suggestion
mkdir -p themes
```

balopat: currently the controller is failing on this step. Can we try with `-p`?

balopat: image

dgageot: Thank you @balopat for investigating.

I don't think there's an issue with `themes` directory.
I think the problem is that it uses an *old* version of the `docs-controller` image. This PR changes the image. I don't know how to handle that case properly

balopat: You can temporarily deploy your version of the controller using `gcloud submit deploy/webhook/cloudbuild.yaml --project k8s-skaffold` or similar

deploy/docs/preview.sh
```diff
@@ -14,16 +14,12 @@
...
-readonly REPO_DIR=$(pwd)
-readonly BASE_URL=${1:-"http://localhost:1313"}
-
-bash ${REPO_DIR}/deploy/docs/build.sh ${BASE_URL}
-
-cd ${REPO_DIR}/docs
+readonly BASE_URL=${1}
 
+cd docs
+mkdir themes
```
balopat:
```suggestion
mkdir -p themes
```

deploy/docs/build.sh
balopat: so...the trick was that the docs-controller image was not pushed/published only with the `deploy/docs/cloudbuild.yaml` - after I did that, we can run `deploy/webhook/cloudbuild.yaml` which then deploys the new controller  but more importantly it can spin up the previews with the right base-image.

balopat: I'm not a big fan of this setup - I should sit down and design the proper dependency graph of these steps soon...

PR #1779, opened 2019-03-11
Refactor caching

priyawadhwa:
I refactored caching and moved it into `pkg/skaffold/build/cache`, as described in #1765 

I split up the giant `cache.go` file into four smaller ones:
1. `cache.go` which initializes the `Cache` type
2. `hash.go` which contains all functions related to getting workspace hashes for artifacts
3. `retrieve.go` which is responsible for retrieving an artifact from the cache and deciding if it needs to be rebuilt/retagged/repushed and executing those actions
4. `save.go` which saves artifacts to the cache


dgageot:
@priyawadhwa Looks like there's still a few Gofmt errors

priyawadhwa:
@dgageot the gofmt errors are happening for Go 1.10 but not Go 1.11, I think waiting for #1775 to merge could fix this?

dgageot:
That's weird...

dgageot:
@priyawadhwa Could you rebase?

PR #1789, opened 2019-03-13
Faster integration tests

dgageot:
+ Use `--cache-from` to build the integration tests Docker image.
+ Run `make test` and `make coverage` in a single step

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,9 @@ clean:
 
 .PHONY: integration-in-docker
 integration-in-docker:
+	-docker pull gcr.io/$(GCP_PROJECT)/skaffold-builder
 	docker build \
+		--cache-from gcr.io/$(GCP_PROJECT)/skaffold-builder \
 		-f deploy/skaffold/Dockerfile \
 		--target integration \
 		-t gcr.io/$(GCP_PROJECT)/skaffold-integration .
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ $(BUILD_DIR):
 cross: $(foreach platform, $(SUPPORTED_PLATFORMS), $(BUILD_DIR)/$(PROJECT)-$(platform).sha256)
 
 .PHONY: test
-test:
+test: $(BUILD_DIR)
 	@ ./test.sh
 
 .PHONY: install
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -94,10 +94,6 @@ ifeq ($(REMOTE_INTEGRATION),true)
 endif
 	REMOTE_INTEGRATION=$(REMOTE_INTEGRATION) go test -v $(REPOPATH)/integration -timeout 10m
 
-.PHONY: coverage
```
tejal29: Can we still keep this in? 
Its used for code coverage i guess. 
https://github.com/GoogleContainerTools/skaffold/blob/master/.travis.yml#L14

tejal29: Looks like you removed it .travis.yaml but we need it for https://github.com/codecov/example-go plugin.

Makefile
```diff
@@ -164,6 +162,7 @@ integration-in-docker:
 		-e DOCKER_CONFIG=/root/.docker \
 		-e GOOGLE_APPLICATION_CREDENTIALS=$(GOOGLE_APPLICATION_CREDENTIALS) \
 		gcr.io/$(GCP_PROJECT)/skaffold-integration
+	docker push gcr.io/$(GCP_PROJECT)/skaffold-integration
```
balopat: why do we need this? GCB pushes the images automatically

dgageot: I tried to pull this image and it didn't exist. So I'm not sure.

Makefile
dgageot: `make test` is supposed to produce the same output, now.

dgageot:
@balopat Is there a GCB job that pushes this image?

Makefile
dgageot: I mean `make test` is supposed to produce the coverage file that codecov needs.

Makefile
balopat: Hmm...yeah you're right - deploy/cloudbuild.yaml only pushes up to the `builder` target in that dockerfile... why do we need to push this?

dgageot: When you say `this`, do you mean the `builder` image or the `skaffold-integration` image?
In case it's the latter, we need to push it to make sure it can be used by `--cache-from` two lines up in the script

balopat: but integration is always different for every different version of skaffold as it is only `builder` + skaffold binary + a specific CMD? can't we just use --cache-from from the builder image?

PR #1790, opened 2019-03-13
Make integration tests shorter and more stable

dgageot:
This PR improves the integration tests to make them easier to debug and more reliable.

+ Use k8s's watch mechanism instead of polling to check for what was deployed
+ Add more logs
+ Mark a few tests as remoteOnly
+ Ensure all go samples use the same code to maximize caching
+ Use smaller base images when possible

Also make TravisCI run faster by not fetching the submodules


dgageot:
kokoro seems to be running forever. Let's wait for the logs to see if the changes help the debugging

dgageot:
Let's try to run this PR a few times:
So far, 1 green run

PR #1791, opened 2019-03-13
Initialize LocalCluster in docker local builder plugin

chanseokoh:
I see `LocalCluster` is used in line 80 but seems not initialized.


chanseokoh:
I wonder if this is related to [this](https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/2454#discussion_r265266938).

PR #1792, opened 2019-03-13
Implement Jib builder (Maven and Gradle) as plugins

chanseokoh:
Just started learning go and about the repo. I've seen a lot of restructuring in the repo around the plugin system recently, and I'd like to see if I am on the right track.

There is a lot of code duplication between Bazel, Docker, Jib Gradle, and Jib Maven plugins, but I guess we'd probably handle that separately with further refactoring in the future?

~~Probably this is in conflict with @TadCordle's new Jib integration (#1744)~~, but I think I should be able to resolve conflicts after it's merged.


chanseokoh:
Still working on adding integration tests.

integration/testdata/plugin/local/jib/skaffold.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: skaffold/v1beta7
+kind: Config
+build:
+  artifacts:
+  - image: gcr.io/sk8-skaffold/skaffold-jib
+    context: .
```
chanseokoh: @priyawadhwa I noticed that if I don't give `context: .`, Jib fails [here](https://github.com/GoogleContainerTools/skaffold/blob/v0.24.0/pkg/skaffold/jib/jib.go#L36) because `cmd.Dir` (which is `artifact.Workspace`) is an empty string. I noticed `context: .` is also given for the [Bazel plugin example project](https://github.com/GoogleContainerTools/skaffold/blob/v0.24.0/integration/examples/test-plugin/local/bazel/skaffold.yaml#L6). Is this required? Shouldn't someone set `artifact.Workspace` if this is missing?

priyawadhwa: Hey @chanseokoh, so there's a [setDefaultWorkspace](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/schema/defaults/defaults.go#L186) function but it looks like it [doesn't get set](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/schema/defaults/defaults.go#L65) when using plugins.

This looks like a bug I introduced, if you move it out to before the plugins [check](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/schema/defaults/defaults.go#L62) this should work as expected (and thanks for catching the bug!)

chanseokoh:
I'll keep working on adding unit and integration tests.

chanseokoh:
Turns out #1744 is independent of this, so this is ready for review.

Summary of changes:
- Two local build integration tests, one for Maven and the other for Gradle.
   - One new sample project (`integration/testdata/plugin/jib`) shared by both tests. Mostly a copy of the `integration/testdata/jib` sample.
- Most of the content of `pkg/skaffold/build/local/jib_{gradle|maven}.go` moved to `pkg/skaffold/plugin/builders/jib/local_{gradle|maven}.go`, following the Bazel/Docker structure.
   - The original `pkg/skaffold/build/local/jib_{gradle|maven}.go` became a minimal adaptor, like the Bazel/Docker siblings.
 - `pkg/skaffold/plugin/builders/jib/builder_{gradle|maven}.go` is basically a copy of the Bazel/Docker siblings.

So, although there seem a lot of file changes, the gist of the changes is not big. Mostly moving and copying contents and files, with some new glue code to run things as a plugin.

pkg/skaffold/plugin/builders/jib/builder_maven.go
```diff
@@ -0,0 +1,133 @@
...
+func setMavenArtifact(artifact *latest.Artifact) error {
+	if artifact.ArtifactType.JibMavenArtifact != nil {
+		return nil
+	}
+	var a *latest.JibMavenArtifact
+	if err := yaml.UnmarshalStrict(artifact.BuilderPlugin.Contents, &a); err != nil {
+		return errors.Wrap(err, "unmarshalling Jib Maven artifact")
+	}
+	if a == nil {
+		a = &latest.JibMavenArtifact{}
```
priyawadhwa: Are there any defaults that need to be set here?

pkg/skaffold/plugin/builders/jib/builder_gradle.go
```diff
@@ -0,0 +1,133 @@
...
+func setGradleArtifact(artifact *latest.Artifact) error {
+	if artifact.ArtifactType.JibGradleArtifact != nil {
+		return nil
+	}
+	var a *latest.JibGradleArtifact
+	if err := yaml.UnmarshalStrict(artifact.BuilderPlugin.Contents, &a); err != nil {
+		return errors.Wrap(err, "unmarshalling Jib Gradle artifact")
+	}
+	if a == nil {
+		a = &latest.JibGradleArtifact{}
```
priyawadhwa: Same question as below, are there any defaults that need to be set on the jibGradle artifact?

pkg/skaffold/plugin/builders/jib/local_gradle_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+	"github.com/GoogleContainerTools/skaffold/testutil"
+	"github.com/docker/docker/api/types"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/pkg/errors"
+)
+
+type mockLocalDaemon struct {
```
priyawadhwa: Instead of creating a new `mockLocalDaemon`, I think you can use our [FakeAPIClient](https://github.com/GoogleContainerTools/skaffold/blob/master/testutil/fake_image_api.go#L32) to mock out the docker daemon, which should make this test a bit simpler. 

An example unit test that uses it and calls `ImageID` (which seems to be the function you need for this test) can be found [here](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/docker/image_test.go#L126).

pkg/skaffold/plugin/builders/jib/builder_maven.go
chanseokoh: In the simplest case, nothing needs to be set, as in this example: https://github.com/GoogleContainerTools/skaffold/blob/master/examples/jib/README.adoc

pkg/skaffold/plugin/builders/jib/builder_gradle.go
chanseokoh: Likewise, nothing needs to be set in the simplest case in Gradle.

pkg/skaffold/plugin/builders/jib/local_gradle_test.go
chanseokoh: Will do so, thanks!

chanseokoh:
Please take a look. This should be good unless there's something off from the recent repo changes pulled in by merging master.

chanseokoh:
Maintaining the PR up-to-date. I see the recent change in master affecting the last merge is the introduction of the "prune" option. Hopefully I got this right with Jib. I believe the option isn't meant to attempt to remove images pushed to a remote registry.

chanseokoh:
@GoogleContainerTools/java-tools-build talked with the Skaffold team, and we agree it's better to close this than prolonging the burden of maintaining the PR to keep up with a lot of potential design and structural changes in the repo.

PR #1808, opened 2019-03-15
cut v0.25.0

balopat:


CHANGELOG.md
```diff
@@ -1,5 +1,98 @@
+# v0.25.0 Release - 3/15/2019
+
+*Note*: This release comes with a config change, use `skaffold fix` to permanently upgrade your config to `v1beta7`, however old versions are now auto-upgraded. 
+See [deprecation-policy.md](/deprecation-policy.md) for details on what beta means.
+
+
+*Deprecation notice*: With this release we mark for deprecation the `additionalFlags` field in kaniko, instead the kaniko builder plugin .
```
priyawadhwa: whoops, looks like you may have cut off the sentence by accident, maybe: 

With this release we mark for deprecation the `additionalFlags` field in kaniko; instead any additional flags will now be represented as unique fields in the KanikoArtifact type.

deprecation-policy.md
```diff
@@ -84,3 +87,17 @@ Commands and their flags are subject to the deprecation policy based on the foll
 - init:  alpha 
 - run:  beta 
 - version:  beta 
+
+
+## Current deprecation notices 
+
+
+03/15/2019: With release v0.25.0 we mark for deprecation the `additionalFlags` field in kaniko, instead the `buildArgs` field should be used.
```
priyawadhwa: This doesn't account for additional flags that aren't `--build-arg`. Maybe we could add in a sentence like:

The buildArgs field should now be used for build arguments, any additional flags kaniko accepts should be added as fields in the new KanikoArtifact type.

CHANGELOG.md
```diff
@@ -1,5 +1,98 @@
+# v0.25.0 Release - 3/15/2019
+
+*Note*: This release comes with a config change, use `skaffold fix` to permanently upgrade your config to `v1beta7`, however old versions are now auto-upgraded.
```
tejal29: i didn't understand what you meant by "permanently upgrade" does that mean, once i upgrade to v1beta7, i can't downgrade?

CHANGELOG.md
```diff
@@ -1,5 +1,98 @@
+# v0.25.0 Release - 3/15/2019
+
+*Note*: This release comes with a config change, use `skaffold fix` to permanently upgrade your config to `v1beta7`, however old versions are now auto-upgraded. 
+See [deprecation-policy.md](/deprecation-policy.md) for details on what beta means.
+
+
+*Deprecation notice*: With this release we mark for deprecation the `additionalFlags` field in kaniko, instead the kaniko builder plugin .
+This flag will will be removed earliest 06/15/2019.
```
tejal29: typo: will will

CHANGELOG.md
balopat: yeah, this is our standard notice - what it means that the versions are auto-upgraded - so you don't have to upgrade yourself but you'll have the warning message every time. If you "permanently upgrade" with `skaffold fix` the upgrade is persisted.

I guess we could rephrase to something like this: to upgrade your skaffold.yaml, use `skaffold fix`. If you don't upgrade, skaffold will auto-upgrade in memory as best it can, and print a warning message.

CHANGELOG.md
```diff
@@ -1,5 +1,98 @@
...
+See [deprecation-policy.md](/deprecation-policy.md) for details on what beta means.
+
+
+*Deprecation notice*: With this release we mark for deprecation the `additionalFlags` field in kaniko, instead the kaniko builder plugin .
+This flag will will be removed earliest 06/15/2019. 
+
+New features: 
+
+* Config upgrade: handle helm overrides [#1646](https://github.com/GoogleContainerTools/skaffold/pull/1646)
+* make initContainer in localDir type of kaniko build param [#1727](https://github.com/GoogleContainerTools/skaffold/pull/1727)
```
tejal29: umm, maybe not an expert, so i cant understand what this means. However, digging into the PR, the description makes more sense.
shd we change this to
"Add init container image used to run init container which mounts kaniko context."

deprecation-policy.md
```diff
@@ -17,11 +17,14 @@ A breaking change is when the primary functionality of a feature changes in a wa
...
-1. we document the deprecation in 
+1. we document the deprecation in the following places if applicable 
 a.) docs
 b.) release notes 
-c.) command help (if applicable)
-d.) https://skaffold.dev/docs/references/yaml/ (if applicable)
+c.) command help 
+d.) logs
+e.) https://skaffold.dev/docs/references/yaml/ 
+f.) this doc
```
tejal29: name the doc instead of this doc in case this moves around.
f.) [Deprecation Policy Document](./deprecation-policy.md)

balopat:
`make test` passes locally, Travis is having a bad day today - I'm Admin-merging this.

PR #1809, opened 2019-03-15
Add tests for skaffold init walk flow.

tejal29:
As we support other helm and kustomize deployers in `skaffold init` flow, walk will become more complex.
Refactoring it out and adding some tests.

### DIFF
--- a/pkg/skaffold/initializer/init_test.go
+++ b/pkg/skaffold/initializer/init_test.go
@@ -77,7 +77,7 @@ func TestWalk(t *testing.T) {
 		err                 bool
 	}{
 		{
-			name: "shd return correct k8 configs and dockerfiles",
+			name: "should return correct k8 configs and dockerfiles",
 			filesWithContents: map[string][]byte{
 				"config/test.yaml":  emptyFile,
 				"k8pod.yml":         emptyFile,
@@ -97,7 +97,7 @@ func TestWalk(t *testing.T) {
 			err: false,
 		},
 		{
-			name: "shd skip hidden dir",
+			name: "should skip hidden dir",
 			filesWithContents: map[string][]byte{
 				".hidden/test.yaml":  emptyFile,
 				"k8pod.yml":          emptyFile,
@@ -115,7 +115,7 @@ func TestWalk(t *testing.T) {
 			err: false,
 		},
 		{
-			name: "shd not error when skaffold.config present and force = true",
+			name: "should not error when skaffold.config present and force = true",
 			filesWithContents: map[string][]byte{
 				"skaffold.yaml": []byte(`apiVersion: skaffold/v1beta6
 kind: Config
@@ -139,7 +139,7 @@ deploy:
 			err: false,
 		},
 		{
-			name: "shd  error when skaffold.config present and force = false",
+			name: "should  error when skaffold.config present and force = false",
 			filesWithContents: map[string][]byte{
 				"config/test.yaml":  emptyFile,
 				"k8pod.yml":         emptyFile,
--- a/pkg/skaffold/initializer/init_test.go
+++ b/pkg/skaffold/initializer/init_test.go
@@ -18,6 +18,10 @@ package initializer
 
 import (
 	"bytes"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/GoogleContainerTools/skaffold/testutil"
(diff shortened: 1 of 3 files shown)


pkg/skaffold/initializer/init_test.go
```diff
@@ -59,3 +63,146 @@ func TestPrintAnalyzeJSON(t *testing.T) {
...
+	tests := []struct {
+		name                string
+		filesWithContents   map[string][]byte
+		expectedConfigs     []string
+		expectedDockerfiles []string
+		force               bool
+		err                 bool
+	}{
+		{
+			name: "shd return correct k8 configs and dockerfiles",
```
balopat:
```suggestion
			name: "should return correct k8 configs and dockerfiles",
```

pkg/skaffold/initializer/init_test.go
```diff
@@ -59,3 +63,146 @@ func TestPrintAnalyzeJSON(t *testing.T) {
...
+				"k8pod.yml",
+			},
+			expectedDockerfiles: []string{
+				"Dockerfile",
+				"deploy/Dockerfile",
+			},
+			err: false,
+		},
+		{
+			name: "shd skip hidden dir",
```
balopat:
```suggestion
			name: "should skip hidden dir",
```

pkg/skaffold/initializer/init_test.go
```diff
@@ -59,3 +63,146 @@ func TestPrintAnalyzeJSON(t *testing.T) {
...
+			expectedConfigs: []string{
+				"k8pod.yml",
+			},
+			expectedDockerfiles: []string{
+				"Dockerfile",
+			},
+			err: false,
+		},
+		{
+			name: "shd not error when skaffold.config present and force = true",
```
balopat:
```suggestion
			name: "should not error when skaffold.yaml present and force = true",
```

pkg/skaffold/initializer/init_test.go
```diff
@@ -59,3 +63,146 @@ func TestPrintAnalyzeJSON(t *testing.T) {
...
+				"k8pod.yml",
+			},
+			expectedDockerfiles: []string{
+				"Dockerfile",
+				"deploy/Dockerfile",
+			},
+			err: false,
+		},
+		{
+			name: "shd  error when skaffold.config present and force = false",
```
balopat:
```suggestion
			name: "should error when skaffold.config present and force = false",
```

pkg/skaffold/initializer/init_test.go
```diff
@@ -59,3 +63,146 @@ func TestPrintAnalyzeJSON(t *testing.T) {
...
+			},
+			force:               false,
+			expectedConfigs:     nil,
+			expectedDockerfiles: nil,
+			err:                 true,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			rootDir, err := ioutil.TempDir("", "test")
```
dgageot: There's an util to do that: `testutil.NewTempDir(t)`. It should make `createDirStructure` easier to write.

pkg/skaffold/initializer/init_test.go
```diff
@@ -59,3 +63,146 @@ func TestPrintAnalyzeJSON(t *testing.T) {
...
+
+func TestWalk(t *testing.T) {
+	emptyFile := []byte("")
+	tests := []struct {
+		name                string
+		filesWithContents   map[string][]byte
+		expectedConfigs     []string
+		expectedDockerfiles []string
+		force               bool
+		err                 bool
```
dgageot: This is called `shouldErr` in every test.

pkg/skaffold/initializer/init_test.go
```diff
@@ -59,3 +63,146 @@ func TestPrintAnalyzeJSON(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		f.Write(content)
+		f.Close()
+	}
+}
+
+func convertToAbsPath(dir string, files []string) []string {
+	if files == nil {
```
dgageot: This can be removed if you change line 203 to `var absPaths []string`

pkg/skaffold/initializer/init_test.go
```diff
@@ -59,3 +61,120 @@ func TestPrintAnalyzeJSON(t *testing.T) {
 		})
 	}
 }
+
+func TestWalk(t *testing.T) {
+	emptyFile := ""
+	tests := []struct {
+		name                string
```
dgageot: This is named `description` is almost every test.

pkg/skaffold/initializer/init_test.go
```diff
@@ -59,3 +61,120 @@ func TestPrintAnalyzeJSON(t *testing.T) {
...
+			force: false,
+			expectedConfigs: []string{
+				"config/test.yaml",
+				"k8pod.yml",
+			},
+			expectedDockerfiles: []string{
+				"Dockerfile",
+				"deploy/Dockerfile",
+			},
+			shouldErr: false,
```
dgageot: We usually avoid those `shouldErr: false`

pkg/skaffold/initializer/init_test.go
```diff
@@ -59,3 +61,120 @@ func TestPrintAnalyzeJSON(t *testing.T) {
...
+			expectedConfigs:     nil,
+			expectedDockerfiles: nil,
+			shouldErr:           true,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			testDir, deleteFunc := testutil.NewTempDir(t)
+			rootDir := testDir.Root()
+			deleteFunc()
```
dgageot: This should be `defer deleteFunc()` and be moved one line higher

dgageot: We also usually don't use a `Func` suffix

pkg/skaffold/initializer/init_test.go
```diff
@@ -59,3 +61,120 @@ func TestPrintAnalyzeJSON(t *testing.T) {
...
+			writeAllFiles(testDir, test.filesWithContents)
+			potentialConfigs, dockerfiles, err := walk(rootDir, test.force, testValidDocker)
+			testutil.CheckErrorAndDeepEqual(t, test.shouldErr, err,
+				testDir.Paths(test.expectedConfigs), potentialConfigs)
+			testutil.CheckErrorAndDeepEqual(t, test.shouldErr, err,
+				testDir.Paths(test.expectedDockerfiles), dockerfiles)
+			os.Remove(rootDir)
```
dgageot: This is not needed.

dgageot: One point that I never mentioned to other members of the team but that I should have:
I find it much easier to read when there's some kind of structure in a test:
```
setup1()
setup2()

action()

assertions()
```

When we add a space before and after the action, it creates a nicer structure.

dgageot: If we create a style guide, we should put all those preferences in.

tejal29: umm actually i have this point of view. I don't remember which testing doc it was, but there were few points which stood out for me
1.  Repetition of code in tests is ok instead of re-direction in another file or method. Having long test function so that all code is contained in the test case instead of having them someother method like `action()` and `assertions()` especially where `action()` and `assertions` is the crux of the code. (Code Proximity Vs Modular approach)
```
setup1()
setup()

// do the action here

// test the assertions here

```
2. Make test case input clear.
To avoid guesswork,  write all the input values.
```
should = false  // though by default bool are false.
```
3. Only refactor code and convert it into a helper function, if its used in more than place.

WDYT?

pkg/skaffold/initializer/init_test.go
tejal29: will fix this.

pkg/skaffold/initializer/init_test.go
tejal29: Actually, i follow the guidelines to make tests easy to read and hence spell it out there.

pkg/skaffold/initializer/init_test.go
tejal29: Does `description` also appear in the test output when it fails?
I know, when the test table does not have name, each testcase is assinged a name `TestWalkXXX` and then its hard to know which of the test case failed.

pkg/skaffold/initializer/init_test.go
dgageot: I understand

pkg/skaffold/initializer/init_test.go
tejal29: Actually, it could be i read this on TOT :) maybe i can find those and we can discuss them.

pkg/skaffold/initializer/init_test.go
tejal29: Changed to description

pkg/skaffold/initializer/init_test.go
```diff
@@ -18,46 +18,163 @@ package initializer
...
+			expectedConfigs:     nil,
+			expectedDockerfiles: nil,
+			shouldErr:           true,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			testDir, cleanUp := testutil.NewTempDir(t)
+			rootDir := testDir.Root()
+			defer cleanUp()
```
dgageot: nit: `defer cleanUp()` should always be called as soon as we get the `cleanup`

dgageot:
@tejal29 approved with a nit. I would also have loved seeing more spaces in tests:

```
```

Issue #1710, opened 2019-03-01
Extract skaffold init functionality into a map response


balopat:
Skaffold init detects Dockerfiles and image names from k8s manifests and then asks the user to map them together. 
For IDE integration purposes it would be beneficial to expose the inital lists of Dockerfiles and imagenames (potentially with pre-matching based on heuristics).

PR #1810, opened 2019-03-15
Error if no Dockerfiles are found for skaffold init --analyze 

### LINKED ISSUE #1795: Skaffold init --analyze should do the same validations as skaffold init.
### Expected behavior
When there are no valid DockerFile or kubernetes manifests in the directory, skaffold init --analyze should provide errors as skaffold init does.
Ex: 
FATA[0000] one or more valid Dockerfiles must be present to run skaffold; please provide at least one Dockerfile and try again
FATA[0000] one or more valid kubernetes manifests is required to run skaffold

### Actual behavior
Skaffold init --analyze doesn't provide any error info.

### Information

- Skaffold version: version...7212422
- Operating system: ... Windows
- Contents of skaffold.yaml: NA

```yaml
<paste your skaffold.yaml here>
```

1. Create a directory that doesn't have valid DockerFiles or kubernetes manifests
2. Run skaffold init --analyze and skaffold init.

priyawadhwa:
This should fix #1795; now, if no Dockerfiles are found in the workspace
then skaffold will error out. If no k8s manifests are found it will
error out (already implemented).


pkg/skaffold/initializer/init.go
```diff
@@ -290,6 +290,9 @@ func generateSkaffoldPipeline(k Initializer, dockerfilePairs []dockerfilePair) (
 }
 
 func printAnalyzeJSON(out io.Writer, dockerfiles, images []string) error {
+	if len(dockerfiles) == 0 {
```
tejal29: This condition is used here as well. 
https://github.com/GoogleContainerTools/skaffold/blob/0211c2c7a5deaf2c5f14ae6b4e3581bb72a054d6/pkg/skaffold/initializer/init.go#L124

@dgageot  and @priyawadhwa should we consider failing `skaffold init` if no docker files are present?
AFAIK, `analyze` is used by IDE? What is it used for?

tejal29: Can we get some more context around why `skaffold init` should fail when there are no docker files?

priyawadhwa: `skaffold init` currently fails if there are no Dockerfiles or Kubernetes manifests found, this PR adds the same functionality to `skaffold init --analyze`.

I'm guessing `skaffold init` fails if either is not present because then there is nothing to match either an image or a k8s manifest with when generating the skaffold.yaml. The `--analyze` flag was recently added so that IDEs could support `skaffold init` (ref #1710)

tejal29: It does not right now. Can we move this check to [L112](https://github.com/GoogleContainerTools/skaffold/blob/0211c2c7a5deaf2c5f14ae6b4e3581bb72a054d6/pkg/skaffold/initializer/init.go#L112)
And then remove it from `if c.SkipBuild` flow.

priyawadhwa: Sorry, I thought we only want to see this error `if !c.skipBuild` as written? Because if we're skipping the build section of the skaffold.yaml, it shouldn't matter if there are any Dockerfiles?

In this case, should I check `if !c.SkipBuild`  for `--analyze` as well?

tejal29: yes makes sense. I think we should also change the error message. Technically you don't need a Dockerfile for running skaffold. You can't "create" skaffold config file with sane config without docker files.

Issue #1807, opened 2019-03-15
How to sync subtrees in 0.25.0?


michaelfig:
### Expected behavior

I want to sync subtrees like:

```
src/foo/bar/baz.js -> /usr/src/app/src/foo/bar/baz.js
src/foo2/bot.js -> /usr/src/app/src/foo2/bot.js
src/other.js -> /usr/src/app/src/other.js
something.txt -> usr/src/app/something.txt
```

In 0.23.0, I could accomplish this simply with:

```yaml
    sync:
      '**': /usr/src/app/
```

### Actual behavior

0.25.0 (and 0.24.0) strips all path information from the destination files, so they all get collapsed and synced directly to `/usr/src/app/` (and not subdirectories of it).  This is due to PR #1709.

I would argue that PR #1709 broke working use-cases, because now the only way I know how to achieve this simple subtree mapping is to add explicit sync rules for every subdirectory in my project (!).  The use-case that #1709 was trying to fix (collapsing certain subdirectories into a root) can be achieved by adding explicit rules for specific directories.

Docker does not support any kind of `**` glob match, so I would argue that skaffold's behaviour here should be more predictable (substituting the glob's match into the resulting destination path), rather than trying to follow some notion of how Docker's `COPY` command works.

Sync rules are *not* COPYs, they are tarball generators.   I think recursive tar is the model that should be used for syncing, with its defaults of preserving paths unless explicitly pruned by a mapping.

### Information

- Skaffold version: 0.25.0
- Operating system: darwin-amd64
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta4
kind: Config
build:
  artifacts:
  - image: michaelfig/something
    sync:
      '**': /usr/src/app/
deploy:
  kubectl:
    manifests:
    - 'k8s-*.yaml'
```

### Steps to reproduce the behavior

1. Run `skaffold dev`
2. Create/modify a file somewhere deep in the tree.

davidarich:
Loving Skaffold so far, but ran into this same issue. For my current project, I had to generate 64 lines of sync rules. Unfortunately the need to manage a large list of sync rules doesn't seem like a maintainable solution for my team. We're using PHP and in most cases sync is preferred over an image rebuild, simply due to speed. I suspect other users working with interpreted languages may have a similar preference.

TheSabby:
This issue seems like it was fixed: #1722 

Update to skaffold v0.25.0 and try again.

[...]

Issue #1581, opened 2019-02-01
Use Dockerfile file paths when using Sync


alewis001:
### Desired Behavior
Files matching the sync pattern are copied to their locations as defined in the Dockerfile.

E.g.

`Dockerfile`:
```
# Start with OL runtime.
FROM open-liberty:kernel

# Configure the features and JVM settings for the Server
COPY server.xml /config/
COPY jvm.options /config/

# Copy shared/sttic libraries into a global location.
COPY lib /config/lib/global

# Add the Azure Certificate for Redis to enable HTTPS connection. 
COPY azure-redis.cer /opt/ol/wlp/output/defaultServer/resources/security/
WORKDIR /opt/ol/wlp/output/defaultServer/resources/security/
RUN keytool -noprompt -importcert -file azure-redis.cer -alias azure-redis -keystore key.jks -storepass Liberty -storetype jks

# Copy in the application
COPY myapp.war /config/dropins
```

In my particular case, as OpenLiberty supports dynamic updates of the `server.xml` and any wars deployed to it, I believe it would be useful to have the sync feature be able to copy those files to the active container based on their location as specified in the Dockerfile rather than having to duplicate the "config" in two places.

I'm able to get the functionality I'm looking for with Skaffold 0.22.0 but I currently need to:
* Update the Dockerfile to include `WORKDIR /` at the end of the file.
* Set the sync targets in `skaffold.yaml` to be the same paths as defined in the Dockerfile.

r2d4:
+1 ref https://github.com/GoogleContainerTools/skaffold/issues/1166

ketaanhshah:
Hey guys, I am building a Docker container with Open Liberty and I used the following in my DockerFile to copy the Azure Cert for HTTPS.

# Add the Azure Certificate for Redis to enable HTTPS connection. 
COPY azurewebsites.cer /opt/ol/wlp/output/defaultServer/resources/security/
WORKDIR /opt/ol/wlp/output/defaultServer/resources/security/
RUN keytool -noprompt -importcert -file azurewebsites.cer -alias azureazurewebsites -keystore azurewebsites.jks -storepass Liberty -storetype jks

but I am getting this error:- 
Step 6/6 : RUN keytool -noprompt -importcert -file azurewebsites.cer -alias azurewebsites -keystore azurewebsites.jks -storepass Liberty -storetype jks
 ---> Running in 8acb3a438510
Certificate was added to keystore
[...]

Issue #1180, opened 2018-10-19
File Sync format specification: maybe don't have user data in the key part


thomas-riccardi:
#1039 added a `key: value` map under `sync` in the skaffold format.
The key is a local glob pattern, the value is the destination directory.

Example:
```yaml
    sync:
      '*.py': .
```

Coming from k8s it is unusual to have user data as keys; it surprised me.

Digging into k8s design doc, we have:
https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#lists-of-named-subobjects-preferred-over-maps
There are whole discussions linked from there, notably about
- readability (mostly in case of subobjects though)
- merging definitions, which may or may not be of concert for skaffold

It lists exceptions to this though: "pure maps [...] as opposed to sets of subobjects".

I suppose if we later want to enrich the format with subobjects it would make things harder with the current format.

I am not (yet) familiar with all the skaffold format, I don't know if there are other cases like that, I couln't find any.

# Question

Shouldn't we forbit storing user data as keys?

# Suggestions

Either go full sub-object like in k8s:
```yaml
    sync:
    - localPattern: '*.py'
      destination: .
```
(better names are welcome)

Or maybe support lighter format with string:
```yaml
    sync:
    - '*.py:.'
```
But I'm not sure it's a good idea: it doesn't seem that much readable...

Issue #1704, opened 2019-02-27
Builders should be able to provide a sync-map


briandealwis:
It would be useful for builders to be able to provide a sync-map.

_(Updated: I had my example projects mixed up:)_
For example, Maven web projects (`.war` files) normally place web resources, such as html/css/images, in `src/main/webapp`.  As part of the Maven build process, these files are processed  into `target/<artifactId>-<version>/` (`target` is location where Maven normally places build-time files).  This directory forms the root of the resulting `.war` file (a glorified zip file).  Jib actually copies the contents of `target/<artifactId>-<version>` directly into in the container as `/jetty/webapps/ROOT`.  The motivation for this issue is that all of these locations are *by convention* and can be reconfigured to point elsewhere.  And other [file locations can be configured too](https://maven.apache.org/plugins/maven-war-plugin/examples/adding-filtering-webresources.html).

So the `jibMaven` builder would return a map like:
- `.../target/<artifactId>-<version>/**` &rarr; `/jetty/webapps/ROOT/`

tejal29:
This sounds reasonable. Just to get more details, the workflow would be, 
- a user changes a class file or a static resource,
- The builder plug complies and puts the binaries in some location.
- skaffold dev should then detect these new binaries  and deploy the new changes.

Is that right?

briandealwis:
Hmm, thinking some more, this is more complex than I thought!  (And my example above was wrong: I've edited it.)

Here's the workflow as a Java developer:

1. the user changes some file or static resource such as `src/main/webapp/index.html`
2. _skaffold dev_ detects the change and starts the build:
    - the normal part of the maven build copies and interpolates the file into `target/artifact-version/index.html`
    - jib copies `target/artifact-version/index.html` into the container at `/jetty/webapps/ROOT/index.html`
3. _skaffold dev_ somehow needs to know that `src/main/webapp/index.html` can be sync'd as  `target/artifact-version/index.html` &rarr; `/jetty/webapps/ROOT/index./html` and so the container doesn't need to be redeployed

I suppose this workflow may be less common as most developers now separate their backend from their frontend.  But I think this would correspond to a frontend webpack-style development flow?

tejal29:
Ah. got it. 
[...]

PR #1812, opened 2019-03-15
WIP Sync: infer destination from Dockerfile

### LINKED ISSUE #1166: Sync map can be inferred from artifact Dockerfile
We already parse the source in a dockerfile's COPY and ADD, but we should also capture the destination.

```Dockerfile
FROM python:3.7-slim
CMD ["python", "-m", "flask", "run"]
ENV FLASK_DEBUG=1
ENV FLASK_APP=app.py

COPY requirements.txt .
RUN pip install -r requirements.txt
COPY *.py . # Should parse dst here
```
So that instead of writing this:
```yaml
  - image: gcr.io/k8s-skaffold/python-reload
    context: python
    sync:
      '*.py': .
```

Users could write this
```yaml
  - image: gcr.io/k8s-skaffold/python-reload
    context: python
    sync:
      '*.py':
```

However, there are many corner cases to watch out for. COPY can take multiple arguments to copy to the same destination folder, and there are rules of when the destination must be inferred as a directory or not.

Overlapping copies may be another corner case. Ultimately I think it will need to be modeled as a map of source files to possibly multiple locations.

Different artifact types can also provide different hints on how to fill in the sync map.

corneliusweig:
**On hold until #1847 is resolved**
## Context
So far, artifact sync required a specification of sync patterns like
```
sync:
  '*.js': app/
```
This is error prone and unnecessarily hard to use, because the destination is already contained in the Dockerfile (see #1166, #1581). In addition, the syncing needs to handle special cases for globbing and often requires a long list of sync patterns (#1807)

## Changes
This PR extracts the destination location from the Dockerfile and uses that information to infer the sync location. For example
```
FROM nginx
WORKDIR /var
COPY static *.html www/
ADD nginx.conf .
```
will infer sync destinations according to docker rules such as
- `nginx.conf -> /var/nginx.conf`
- `static/page1.html -> /var/www/page1.html` (! no `static` folder at destination !)
- `static/assets/style.css -> /var/www/assets/style.css`
- `index.html -> /var/www/index.html`

#### Possible issues
- I did not consider how to resolve overlapping copies to the same destination. For example
   ```
   COPY foo /foo
   COPY bar/ /   # bar contains a file called "foo"
   ```
   In that case it is indeterminate what the content of `/foo` will be. However, simpler cases such as
   ```
   COPY foo baz
   COPY bar baz
   ```
   guarantee that `baz` has content from `bar`. 
- Transitive copies across stages of multi-stage dockerfiles are not resolved:
   ```
   FROM scratch
   ADD foo .
   FROM scratch
   COPY --from=0 foo bar
   ```
   The current implementation will resolve `foo -> foo` and not `foo -> bar`. In practice, this should not be a problem, because this does not really make sense and can easily be circumvented.

#### Simplified pipeline config
Since the pipeline config sync map competes with the sync inference, and it is not clear to me how we can handle preference, I also changed the config field to a boolean:
```
sync: true # false
```
This means however, that sync can no longer be used with Dockerfile-less builders (i.e. only kaniko and docker, right?). I think
- for jib syncing does not make sense IMHO, so this should be fine.
- for bazel, I also don't see the use-case. Also, I'm not a bazel expert and maybe there is a way to get the file destination.

For me, this is reason enough to avoid the complexity of maintaining the original sync map and bother about preference. Let me know your thoughts..

Fixes #1166, #1581 
Related #1180, #1807 
See #1844 for a design discussion

### DIFF
--- a/cmd/skaffold/app/cmd/diagnose.go
+++ b/cmd/skaffold/app/cmd/diagnose.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"sort"
 	"time"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
@@ -118,8 +119,16 @@ func diagnoseArtifacts(out io.Writer, builder build.Builder, artifacts []*latest
 
 func timeToListDependencies(ctx context.Context, builder build.Builder, a *latest.Artifact) (time.Duration, []string, error) {
 	start := time.Now()
-	paths, err := builder.DependenciesForArtifact(ctx, a)
-	return time.Since(start), paths, err
+	dependencies, err := builder.DependenciesForArtifact(ctx, a)
+	duration := time.Since(start)
+
+	paths := make([]string, 0, len(dependencies))
+	for path := range dependencies {
+		paths = append(paths, path)
+	}
+	sort.Strings(paths)
+
+	return duration, paths, err
 }
 
 func timeToComputeMTimes(deps []string) (time.Duration, error) {
--- a/docs/content/en/schemas/v1beta7.json
+++ b/docs/content/en/schemas/v1beta7.json
@@ -70,16 +70,10 @@
               "x-intellij-html-description": "plugin used to build this artifact."
             },
             "sync": {
-              "additionalProperties": {
-                "type": "string"
-              },
-              "type": "object",
-              "description": "*alpha* local files synced to pods instead of triggering an image build when modified. This is a mapping of local files to sync to remote folders.",
-              "x-intellij-html-description": "<em>alpha</em> local files synced to pods instead of triggering an image build when modified. This is a mapping of local files to sync to remote folders.",
-              "default": "{}",
-              "examples": [
-                "{\"*.py\": \".\", \"css/**/*.css\": \"app/css\"}"
-              ]
+              "type": "boolean",
+              "description": "*alpha* local files synced to pods instead of triggering an image build when modified.",
+              "x-intellij-html-description": "<em>alpha</em> local files synced to pods instead of triggering an image build when modified.",
+              "default": "false"
             }
           },
           "preferredOrder": [
@@ -117,16 +111,10 @@
               "x-intellij-html-description": "plugin used to build this artifact."
             },
             "sync": {
-              "additionalProperties": {
-                "type": "string"
-              },
-              "type": "object",
-              "description": "*alpha* local files synced to pods instead of triggering an image build when modified. This is a mapping of local files to sync to remote folders.",
-              "x-intellij-html-description": "<em>alpha</em> local files synced to pods instead of triggering an image build when modified. This is a mapping of local files to sync to remote folders.",
-              "default": "{}",
-              "examples": [
-                "{\"*.py\": \".\", \"css/**/*.css\": \"app/css\"}"
-              ]
+              "type": "boolean",
+              "description": "*alpha* local files synced to pods instead of triggering an image build when modified.",
+              "x-intellij-html-description": "<em>alpha</em> local files synced to pods instead of triggering an image build when modified.",
+              "default": "false"
             }
           },
           "preferredOrder": [
@@ -165,16 +153,10 @@
               "x-intellij-html-description": "plugin used to build this artifact."
             },
             "sync": {
-              "additionalProperties": {
-                "type": "string"
-              },
-              "type": "object",
-              "description": "*alpha* local files synced to pods instead of triggering an image build when modified. This is a mapping of local files to sync to remote folders.",
-              "x-intellij-html-description": "<em>alpha</em> local files synced to pods instead of triggering an image build when modified. This is a mapping of local files to sync to remote folders.",
-              "default": "{}",
-              "examples": [
-                "{\"*.py\": \".\", \"css/**/*.css\": \"app/css\"}"
-              ]
+              "type": "boolean",
+              "description": "*alpha* local files synced to pods instead of triggering an image build when modified.",
+              "x-intellij-html-description": "<em>alpha</em> local files synced to pods instead of triggering an image build when modified.",
+              "default": "false"
             }
           },
           "preferredOrder": [
@@ -213,16 +195,10 @@
               "x-intellij-html-description": "plugin used to build this artifact."
             },
             "sync": {
-              "additionalProperties": {
-                "type": "string"
-              },
-              "type": "object",
-              "description": "*alpha* local files synced to pods instead of triggering an image build when modified. This is a mapping of local files to sync to remote folders.",
-              "x-intellij-html-description": "<em>alpha</em> local files synced to pods instead of triggering an image build when modified. This is a mapping of local files to sync to remote folders.",
-              "default": "{}",
-              "examples": [
-                "{\"*.py\": \".\", \"css/**/*.css\": \"app/css\"}"
-              ]
+              "type": "boolean",
+              "description": "*alpha* local files synced to pods instead of triggering an image build when modified.",
+              "x-intellij-html-description": "<em>alpha</em> local files synced to pods instead of triggering an image build when modified.",
+              "default": "false"
             }
           },
           "preferredOrder": [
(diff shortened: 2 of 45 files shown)


michaelfig:
Instead of making `sync` a boolean, could you instead still support legacy sync maps, and add a magic value to say that inference is also used, in case inference needs some help?

```yaml
  sync:
    "bar.txt": foo
    "$infer": docker
```

I'm not attached to the syntax, just not a fan of breaking changes when we're unsure if all the existing use cases are covered.

corneliusweig:
@michaelfig There also was some discussion to change the sync spec to subobjects (#1180). If the old sync spec should still be supported, it needs to come this way I think. For example
```
sync:
    infer: false   # default to true
    override:      # default to empty list
    - local: '*.py'
      destination: .
```

But before implementing this, I would wait for some more feedback. The overrides invite to do "clever" things and we don't really want to encourage this. If the overrides are not necessary, I would prefer to avoid it, because it considerably reduces complexity.
@r2d4 as the original sync author, what do you think?

corneliusweig:
I thought about this again. Maybe the important use-case for the sync patterns is to restrict the set of files for which syncing makes sense. For example, if you build a service from `main.go` that serves static assets. Then you'd want the static stuff to be synced, whereas changes in `main.go` should trigger a rebuild. This case is not covered by a plain `sync: true/false` config.
However, what still remains to be clarified is whether overriding the destination remains a valid use-case with destination inference.

tejal29:
@corneliusweig  Thank you for starting the discussion.
While finding the best way to represent sync, we shd also think of supporting existing usecases
1. Magic path "***"
2. Preserve subtree sync https://github.com/GoogleContainerTools/skaffold/pull/1813/files 
And probably potential others listed here https://github.com/GoogleContainerTools/skaffold/labels/area%2Fsync

tejal29:
One more requirement https://github.com/GoogleContainerTools/skaffold/issues/1704

tejal29:
in summary, i think we should still provide a way to override which can be used by builder plugins.

Regarding your comment, 
> for jib syncing does not make sense IMHO, so this should be fine.

There is an issue open #1704 

> for bazel, I also don't see the use-case. Also, I'm not a bazel expert and maybe there is a way to get the file destination.

In bazel, you can also bundle static resources as dependencies and we should be able to infer these paths.

>For example, if you build a service from main.go that serves static assets. Then you'd want the static stuff to be synced, whereas changes in main.go should trigger a rebuild. This case is not covered by a plain sync: true/false config.

Maybe if sync is false, changes in any file trigger a full build probably what we had before adding sync map?

corneliusweig:
@tejal29 Thanks for your insight here. So based on your input, plan to implement the following:
```
sync:
- from: 'src/***'
  to: '/app/src'
- from: *
  # to omitted
```
- Only host paths matching any of the patterns will be sync'ed.
- The `from` field basically corresponds to the key field in the current sync object and should understand magic sequences
- If specified, the `to` location will always be taken from the sync rule
- If *not* specified, the `to` location will be inferred from the Dockerfile. Thus, a rule such as
  ```
  sync:
  - from: '**'
  ```
  will basically sync everything.
- Builder plugins will then have the possibility to provide additional sync patterns
- Every from/to rule will be considered for syncing and not just the first match
Did I miss anything?


> In bazel, you can also bundle static resources as dependencies and we should be able to infer these paths.

If you know how to get this from bazel, can you send me a link to the docs?

michaelfig:
@corneliusweig your plan looks really good!

Just to restate (maybe the obvious?):

```yaml
sync:
- from: '***'
```

should also sync everything with inference.

tejal29:
@corneliusweig this looks good.

>* Only host paths matching any of the patterns will be sync'ed.

+1
> * The from field basically corresponds to the key field in the current sync object and should understand > magic sequences
The magic sequences we support right now are "*", "**" and "***". Just enlisting them here.
> * If specified, the to location will always be taken from the sync rule

+1
> * If not specified, the to location will be inferred from the Dockerfile. Thus, a rule such as
> ```
> sync:
> - from: '**'
> ```
> will basically sync everything.

Right now we support "\*\*" and "\*\*\*".  To  add to @michaelfig point, we need to support both. However not sure if we use a completely different notation here. In normal case where "from" and "to" are specified
1. "\*\*": denotes match all the files and collapse it to path mentioned in "to"  field.
2. "\*\*\*": denotes match all files and sync them to respective sub dirs.
However, in the case when files will be inferred users might wonder if they are being collapsed or not.
Does that make sense?
Should we just use  another filed for inference like?

```
sync: 
   inferred : true
   - from:'**.js'
     to: /home/js
   - from: '***/*.html"
     to: : "/home/webapp"
```
> * If not specified, and not a docker artifact, bail out with an error
Makes sense right now since we don't support any other inference. 
In future, when bazel builder plugin 
> * Builder plugins will then have the possibility to provide additional sync patterns

Correct. will this part of Builder Interface?
```
type Builder interface {
  Labels() map[string]string

  // NEW
  SyncMap()map[string]string 
}
```
> * Every from/to rule will be considered for syncing and not just the first match
Umm probably, 2 source files being synced to same destination should be an error?

> Did I miss anything?

Currently, i  can't think of any. But this is flexible enough to add more use cases.

> If you know how to get this from bazel, can you send me a link to the docs?

So currently we do a (bazel query)[https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/plugin/builders/bazel/dependencies.go#L51] to find dependencies.
We can extend that to find dependencies of certain types. 
e.g.: https://docs.bazel.build/versions/master/query-how-to.html#what-rule-dependencies-exist
An example of using .js files for a webapp is here
https://github.com/bazelbuild/rules_nodejs/blob/master/examples/webapp/BUILD.bazel#L6

Another way to pull in files in bazel at runtime is by specifying them in ("data")[https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes] field.
So, we should be able to directly "sync" files which are listed as "data" dependencies or of types like "rollup_bundle".
Some other links to look at 
* https://docs.bazel.build/versions/master/build-ref.html#types_of_dependencies

tejal29:
So, after discussing this with the team, we are thinking of having a new process for Design proposals.

We would love to have the community work with us on new features. How about we capture all this discussion in a Design Proposal doc that could live in docs/design_proposals?
I will go ahead and add a template there.

What we are looking for are:
1. What is this new feature?
2. Why?
3. Design
4. Is this going to break existing usage?
5. If its a backward compatible config change, is there a migration path?
6. Implementation plan or break it down in smaller issues making small incremental changes.
   for e.g. in this case,
   a. We will first add `infer` key to sync map and if it is set to true, docker inference will be used. If both `infer` and user defined map is present, the sync will error out.
   b. Support both with precedence rules implemented. 
   d. Finally, support builder plugins provide sync maps.
7. Integration test

Does that make sense?

corneliusweig:
Hi @tejal29, sounds as if not everyone on the team was happy about how things panned out. Sorry for any inconvencience.

I'll create a separate PR for the design proposal, once the template is there. Let's continue this discussion on this design proposal, because inline comments on github comments are not possible :)
However, one thing really got me confused, because this is not valid yaml
```
sync: 
   inferred : true
   - from:'**.js'
     to: /home/js
   - from: '***/*.html"
     to: : "/home/webapp"
```
and I don't quite get why `inferred` is true, but there are `to` fields.

tejal29:
> Hi @tejal29, sounds as if not everyone on the team was happy about how things panned out. Sorry for any inconvencience.

No there wasn't any inconvenience, as i was discussing with balint, we realized this is a bigger change and shd have a design discussion.

> I'll create a separate PR for the design proposal, once the template is there. Let's continue this discussion on this design proposal, because inline comments on github comments are not possible :)

The PR is already up here #1838 
> However, one thing really got me confused, because this is not valid yaml
> 
> ```
> sync: 
>    inferred : true
>    - from:'**.js'
>      to: /home/js
>    - from: '***/*.html"
>      to: : "/home/webapp"
> ```
> and I don't quite get why `inferred` is true, but there are `to` fields.
My bad. One thing me and balint were discussing was,
Can users have sync maps which hints + inference ? and if yes, what should take precedence?
We can definitely decide it to be either or. Again something to note down in design proposal.

bhack:
Will the destination from Dockerfile mandatory? What about NoDockerfile use cases  https://github.com/GoogleContainerTools/skaffold/issues/1396?

corneliusweig:
> Will the destination from Dockerfile mandatory? What about NoDockerfile use cases #1396?

The destination inference is a convenience feature. All existing use-cases will still be possible. Cf. #1844

corneliusweig:
I'll close this, because the discussion on this PR is completely outdated. See design proposal #1844 for details.

Issue #1538, opened 2019-01-27
Helm integration test fails with `namespaces "skaffold6plvk" not found`


dgageot:
All the helm integration tests fail with `Error: UPGRADE FAILED: failed to create resource: namespaces "skaffold6plvk" not found`.
What is strange is that it's always complaining about the same namespace.

dgageot:
I ran `helm reset --force` and then `helm init` and now tests fail with `Error: configmaps is forbidden: User "system:serviceaccount:kube-system:default" cannot list configmaps in the namespace "kube-system"`

dgageot:
Had to run `helm delete skaffold-helm --purge`

PR #1813, opened 2019-03-16
Preserve sync subtree for '***'.

### LINKED ISSUE #1807: How to sync subtrees in 0.25.0?

michaelfig:
Fixes #1807 

This PR implements subtree syncing for '***' globs in the sync specification.

Given:
```yaml
    sync:
      'foo/***/*.js': /usr/src/app/
```

and changing `foo/baz/bot/test.js`, `test.js` will be synchronized to `/usr/src/app/baz/bot/test.js`.

Given:
```yaml
    sync:
      '***': subdir
```

and changing `my/src.txt`, `src.txt` will be synchronized to `${WORKDIR}/subdir/my/src.txt`.

All the existing sync specs in 0.25.0 mean the same as they did before this PR (i.e. they collapse the subPath, so all files in the subtree are synced to the same destination directory, not its children).  To allow for backward compatibility with 0.23.0's subtree syncing, you can use:

```yaml
    sync:
      '**': /usr/src/app/ # v0.23.0
      '***': /usr/src/app/ # this PR
```

The line marked `this PR` will take precedence over the `v0.23.0` line (triple-stars are matched first before other patterns), except when running a version of Skaffold before this PR.  In that case, the `v0.23.0` line will be used, and only `v0.23.0` will produce the same behaviour as this PR.

Hope this helps,
Michael.

### DIFF
--- a/pkg/skaffold/sync/sync.go
+++ b/pkg/skaffold/sync/sync.go
@@ -123,58 +123,93 @@ func latestTag(image string, builds []build.Artifact) string {
 func intersect(context string, syncMap map[string]string, files []string, workingDir string) (map[string]string, error) {
 	ret := map[string]string{}
 
+	tripleStarSyncMap, otherSyncMap := segregateSyncMaps(syncMap)
 	for _, f := range files {
 		relPath, err := filepath.Rel(context, f)
 		if err != nil {
 			return nil, errors.Wrapf(err, "changed file %s can't be found relative to context %s", f, context)
 		}
-		var matches bool
-		// Prioritize '***' sync map rules, for backward compatibility.
-		for _, doTripleStar := range []bool{true, false} {
-			for p, dst := range syncMap {
-				// Replace triple-star with double-star to make a pattern.
-				pat := strings.Replace(p, "***", "**", -1)
-				isTripleStar := len(pat) != len(p)
-				if isTripleStar != doTripleStar {
-					// Wait until the right phase.
-					continue
-				}
-				match, err := doublestar.PathMatch(filepath.FromSlash(pat), relPath)
-				if err != nil {
-					return nil, errors.Wrapf(err, "pattern error for %s", relPath)
-				}
-				if match {
-					var subPath string
-					if isTripleStar {
-						// Map the paths as a tree from the prefix.
-						subtreePrefix := strings.Split(p, "***")[0]
-						subPath = strings.TrimPrefix(relPath, filepath.FromSlash(subtreePrefix))
-					} else {
-						// Collapse the paths.
-						subPath = filepath.Base(relPath)
-					}
-					if filepath.IsAbs(dst) {
-						dst = filepath.ToSlash(filepath.Join(dst, subPath))
-					} else {
-						dst = filepath.ToSlash(filepath.Join(workingDir, dst, subPath))
-					}
-					// Every file must match at least one sync pattern, if not we'll have to
-					// skip the entire sync
-					matches = true
-					if _, ok := ret[f]; !ok {
-						// Don't overwrite existing returns.
-						ret[f] = dst
-					}
-				}
+
+		var match bool
+
+		// First try all tripleStarSyncMaps.
+		match, dst, err := matchTripleStarSyncMap(tripleStarSyncMap, relPath)
+		if err != nil {
+			return nil, err
+		}
+
+		if !match {
+			// Try matching the other rules.
+			match, dst, err = matchOtherSyncMap(otherSyncMap, relPath)
+			if err != nil {
+				return nil, err
 			}
 		}
-		if !matches {
+
+		if !match {
 			logrus.Infof("Changed file %s does not match any sync pattern. Skipping sync", relPath)
 			return nil, nil
 		}
 
+		// Convert relative destinations to absolute via the workingDir.
+		if filepath.IsAbs(dst) {
+			dst = filepath.ToSlash(dst)
+		} else {
+			dst = filepath.ToSlash(filepath.Join(workingDir, dst))
+		}
+
+		// Record the final destination.
+		ret[f] = dst
 	}
+
 	return ret, nil
+}
+
+func segregateSyncMaps(syncMap map[string]string) (tripleStarPattern, doubleStarPattern map[string]string) {
+	tripleStarPattern = make(map[string]string)
+	doubleStarPattern = make(map[string]string)
+	for p, dst := range syncMap {
+		if strings.Contains(p, "***") {
+			tripleStarPattern[p] = dst
+		} else {
+			doubleStarPattern[p] = dst
+		}
+	}
+	return
+}
+
+func matchTripleStarSyncMap(syncMap map[string]string, relPath string) (bool, string, error) {
+	for p, dst := range syncMap {
+		pat := strings.Replace(p, "***", "**", -1)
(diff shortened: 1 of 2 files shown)


pkg/skaffold/sync/sync.go
```diff
@@ -128,21 +129,43 @@ func intersect(context string, syncMap map[string]string, files []string, workin
...
+				match, err := doublestar.PathMatch(filepath.FromSlash(pat), relPath)
+				if err != nil {
+					return nil, errors.Wrapf(err, "pattern error for %s", relPath)
+				}
+				if match {
+					var subPath string
+					if isTripleStar {
+						// Map the paths as a tree from the prefix.
+						subtreePrefix := strings.Split(p, "***")[0]
+						subPath = strings.TrimPrefix(relPath, subtreePrefix)
```
michaelfig: Need:
```go
filepath.FromSlash(subtreePrefix)
```
Otherwise the prefix-pruning fails on Windows.

pkg/skaffold/sync/sync_test.go
```diff
@@ -275,6 +275,61 @@ func TestNewSyncItem(t *testing.T) {
...
+			},
+			workingDir: "/some/dir",
+			builds: []build.Artifact{
+				{
+					ImageName: "test",
+					Tag:       "test:123",
+				},
+			},
+			evt: watch.Events{
+				Added: []string{filepath.Join("dir1", "dir2/node.js")},
```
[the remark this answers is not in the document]
tejal29: i dont like the fact that this for loop is so long and it gets difficult to read.
How about?
```

tripleStarSyncMaps, otherSyncMaps = segregateSyncMaps(syncMaps)
for _, f := range files {
 ...
 var match bool
 // first try all tripleStarSyncMaps
 match, dst, err := matchesTripleStarSyncMap(f)
 if err != nil {
    return nil, err
 }
 if match {
   ret[f] = dst
   continue
 }
 // Try matching others
 match, dst, err := match(otherSyncMaps, f)
 if !match {
  logrus.Infof("Changed file %s does not match any sync pattern. Skipping sync", relPath)
  return nil, nil
 }
 ret[f] = dst
}
...
}


func segregateSyncMaps(syncMap map[string]string) (map[string]string, map[string] string) {
  for p, dst := range syncMap {
     pat := strings.Replace(p, "***", "**", -1)
     if len(pat) != len(p) {
       tripleStarPatter[pat] = dst
     } else {
      doubelStarPattern[p] = dst
    } 
 } 
 return tripleStarPattern, doubleStarPatter
}
```

That ways, we individually test smaller functions.

pkg/skaffold/sync/sync.go
```diff
@@ -128,21 +129,43 @@ func intersect(context string, syncMap map[string]string, files []string, workin
 			return nil, errors.Wrapf(err, "changed file %s can't be found relative to context %s", f, context)
 		}
 		var matches bool
-		for p, dst := range syncMap {
```
tejal29: i dont like the fact that this for loop is so long and it gets difficult to read.
How about we first find all the TripleStar Patterns, match them and then try all the others?

```

tripleStarSyncMaps, otherSyncMaps = segregateSyncMaps(syncMaps)
for _, f := range files {
 ...
 var match bool
 // first try all tripleStarSyncMaps
 match, dst, err := matchesTripleStarSyncMap(f)
 if err != nil {
    return nil, err
 }
 if match {
   ret[f] = dst
   continue
 }
 // Try matching others
 match, dst, err := matchOthers(otherSyncMaps, f)
 if !match {
  logrus.Infof("Changed file %s does not match any sync pattern. Skipping sync", relPath)
  return nil, nil
 }
 ret[f] = dst
}
...
}


func segregateSyncMaps(syncMap map[string]string) (map[string]string, map[string] string) {
  for p, dst := range syncMap {
     pat := strings.Replace(p, "***", "**", -1)
     if len(pat) != len(p) {
       tripleStarPatter[pat] = dst
     } else {
      doubelStarPattern[p] = dst
    } 
 } 
 return tripleStarPattern, doubleStarPatter
}
```

That ways, we individually test smaller functions.

tejal29:
Thank you for the bug fix! The kokoro job failed due to intergration test flakiness #1538. 
This should be fixed. I have the kokoro-run label. Hopefully a new build will be kicked off soon.

pkg/skaffold/sync/sync_test.go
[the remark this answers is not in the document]
michaelfig: AFAICT, the file watcher only returns relative events.  What, specifically, are you suggesting should be the absolute path?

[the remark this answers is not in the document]
michaelfig: No, I see you're right.  I'll have a look.

pkg/skaffold/sync/sync.go
michaelfig: Good suggestion.  Thanks!

michaelfig:
@tejal29 I implemented your proposed changes.  I found and fixed some test failures with absolute paths on Windows.  Would you mind please relabeling `kokoro-run`?

pkg/skaffold/sync/sync.go
```diff
@@ -119,41 +120,108 @@ func latestTag(image string, builds []build.Artifact) string {
...
+	tripleStarPattern = make(map[string]string)
+	doubleStarPattern = make(map[string]string)
+	for p, dst := range syncMap {
+		if strings.Contains(p, "***") {
+			tripleStarPattern[p] = dst
+		} else {
+			doubleStarPattern[p] = dst
+		}
+	}
+	return
```
tejal29: Thanks, i never knew you can return like this in go.

PR #1826, opened 2019-03-19
Add React example app featuring hot module reload

### LINKED ISSUE #1173: Add a React example with file-sync

corneliusweig:
Close #1173

I checked that hot module reloading is working with version v0.25.0. In other words, I cannot reproduce #1640 with this example.

Note that the sync spec quickly becomes unwieldy with many subfolders. #1812 and #1813 should fix this when merged.

### DIFF
--- a/examples/react-reload/README.adoc
+++ b/examples/react-reload/README.adoc
@@ -1,0 +1,16 @@
+=== Example: React app with hot-reload
+:icons: font
+
+Simple React app demonstrating the file synchronization mode in conjunction with webpack hot module reload.
+
+==== Init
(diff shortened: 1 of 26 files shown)


corneliusweig:
@dgageot I copied the example to integration/examples as requested. Anything else that needs to be done so that it is picked up by the integration tests?

PR #1833, opened 2019-03-19
Fixing nit

dgageot:


dgageot:
The usual kaniko integration test is failing

PR #1834, opened 2019-03-19
Increase integration tests timeout to 15minutes

dgageot:


dgageot:
I'm going to ignore travis that seems to have problems with OSX hosts

PR #1840, opened 2019-03-20
Improve test coverage

dgageot:
This improves the coverage of a few pieces of code.

It also configures `go test` so that test coverage is counted on every package, not the package that the test belongs to. This should show our real test coverage.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -21,7 +21,7 @@ GREEN='\033[0;32m'
 RESET='\033[0m'
 
 echo "Running go tests..."
-go test -count=1 -race -cover -short -timeout=60s -coverprofile=out/coverage.txt -covermode=atomic ./... | awk -v FAIL="${RED}FAIL${RESET}" '! /no test files/ { gsub("FAIL", FAIL, $0); print $0 }'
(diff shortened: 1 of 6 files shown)


dgageot:
I still don't understand how Codecov is computing coverage...

test.sh
```diff
@@ -21,7 +21,8 @@ GREEN='\033[0;32m'
 RESET='\033[0m'
 
 echo "Running go tests..."
-go test -count=1 -race -cover -short -timeout 60s -coverprofile=out/coverage.txt -covermode=atomic ./... | grep -v 'no test files' | sed ''/PASS/s//$(printf "${GREEN}PASS${RESET}")/'' | sed ''/FAIL/s//$(printf "${RED}FAIL${RESET}")/''
+go test -cover -short -timeout=60s -coverprofile=out/coverage.txt -coverpkg="./pkg/...,./cmd/..." ./... | awk -v FAIL="${RED}FAIL${RESET}" '! /no test files/ { gsub("FAIL", FAIL, $0); print $0 }'
```
balopat: this removes the race detector - is that intentional?

dgageot: Good catch!

dgageot:
@balopat should be all good now

PR #1841, opened 2019-03-20
Fix skaffold build templating output and add tests

### LINKED ISSUE #1839: Fix skaffold build output templating
While reproducing #922 i discovered "--output" flag does not work as documented.

### Expected behavior
`skaffold build -o` should work as documented.
### Actual behavior
Running `skaffold build -o` does not work without `--quiet` flag present.
The help does not mention this.

The help also mentions,
```
tejaldesai@ (deploy_build) $ ../../out/skaffold build help
Usage:
  skaffold build [flags]

Flags:
  -b, --build-image stringArray      Choose which artifacts to build. Artifacts with image names that contain the expression will be built only. Default is to build sources for all artifacts
      --cache-artifacts              Set to true to enable caching of artifacts.
      --cache-file string            Specify the location of the cache file (default $HOME/.skaffold/cache)
  -d, --default-repo string          Default repository value (overrides global config)
      --enable-rpc skaffold dev      Enable gRPC for exposing Skaffold events (true by default for skaffold dev)
  -f, --filename string              Filename or URL to the pipeline file (default "skaffold.yaml")
  -h, --help                         help for build
  -n, --namespace string             Run deployments in the specified namespace
  -o, --output *flags.TemplateFlag   Format output with go-template. For full struct documentation, see https://godoc.org/github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/cmd#BuildOutput (default {{range .Builds}}{{.ImageName}} -> {{.Tag}}
```
The default value is incorrect. it should be `{{range .Builds}}{{.ImageName}} -> {{.Tag}}{{end}}`
Also, this default is not respected.
```
tejaldesai@ (deploy_build) $ ../../out/skaffold build -q -o 
Usage:
  skaffold build [flags]

Flags:
  -b, --build-image stringArray      Choose which artifacts to build. Artifacts with image names that contain the expression will be built only. Default is to build sources for all artifacts
      --cache-artifacts              Set to true to enable caching of artifacts.
      --cache-file string            Specify the location of the cache file (default $HOME/.skaffold/cache)
  -d, --default-repo string          Default repository value (overrides global config)
      --enable-rpc skaffold dev      Enable gRPC for exposing Skaffold events (true by default for skaffold dev)
  -f, --filename string              Filename or URL to the pipeline file (default "skaffold.yaml")
  -h, --help                         help for build
  -n, --namespace string             Run deployments in the specified namespace
  -o, --output *flags.TemplateFlag   Format output with go-template. For full struct documentation, see https://godoc.org/github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/cmd#BuildOutput (default {{range .Builds}}{{.ImageName}} -> {{.Tag}}
                                     {{end}})
  -p, --profile stringArray          Activate profiles by name
  -q, --quiet                        Suppress the build output and print image built on success
      --rpc-port int                 tcp port to expose event API (default 50051)
      --skip-tests                   Whether to skip the tests after building
      --toot                         Emit a terminal beep after the deploy is complete

Global Flags:
      --color int          Specify the default output color in ANSI escape codes (default 34)
  -v, --verbosity string   Log level (debug, info, warn, error, fatal, panic) (default "warning")

FATA[0000] flag needs an argument: 'o' in -o  
```
### Information

- Skaffold version: version...
- Operating system: ...
- Contents of skaffold.yaml:

```yaml
<paste your skaffold.yaml here>
```

1. ...
2. ...

tejal29:
fixes #1839 

In this commit

- Add better description on flags so its clear, `skaffold build -o` is only useful with --quiet.
- suppress "Complete in " line with --quiet flag so that  `skaffold build --quiet > build.out` output can consumed to by `skaffold deploy` to correctly parse and construct a build.Artifacts object.
- Refactor runBuild to move creating a runner and actually building artifacts to another function, so it can be mocked in tests.
- add tests for --quiet flag and make sure output is as expected when a template is specified vs when its not.
- change default template to "{{.}}"

### DIFF
--- a/cmd/skaffold/app/cmd/build.go
+++ b/cmd/skaffold/app/cmd/build.go
@@ -98,15 +98,15 @@ func runBuild(out io.Writer) error {
 
 func createRunnerAndBuild(ctx context.Context, buildOut io.Writer) ([]build.Artifact, error) {
 	runner, config, err := newRunner(opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "creating runner")
+	}
+	defer runner.RPCServerShutdown()
 	var targetArtifacts []*latest.Artifact
 	for _, artifact := range config.Build.Artifacts {
 		if runner.IsTargetImage(artifact) {
 			targetArtifacts = append(targetArtifacts, artifact)
 		}
 	}
-	if err != nil {
-		return nil, errors.Wrap(err, "creating runner")
-	}
-	defer runner.RPCServerShutdown()
 	return runner.BuildAndTest(ctx, buildOut, targetArtifacts)
 }
--- a/cmd/skaffold/app/cmd/build.go
+++ b/cmd/skaffold/app/cmd/build.go
@@ -71,13 +71,17 @@ func runBuild(out io.Writer) error {
 		}
 	}()
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	catchCtrlC(cancel)
+
 	buildOut := out
 	if quietFlag {
 		buildOut = ioutil.Discard
 	}
 
-	bRes, cleanUp, err := createRunnerAndBuildFunc(buildOut)
-	defer cleanUp()
+	bRes, err := createRunnerAndBuildFunc(ctx, buildOut)
+
 	if err != nil {
 		return err
 	}
@@ -92,10 +96,7 @@ func runBuild(out io.Writer) error {
 	return nil
 }
 
-func createRunnerAndBuild(buildOut io.Writer) ([]build.Artifact, func(), error) {
-	ctx, cancel := context.WithCancel(context.Background())
-	catchCtrlC(cancel)
-
+func createRunnerAndBuild(ctx context.Context, buildOut io.Writer) ([]build.Artifact, error) {
 	runner, config, err := newRunner(opts)
 	var targetArtifacts []*latest.Artifact
 	for _, artifact := range config.Build.Artifacts {
@@ -104,13 +105,8 @@ func createRunnerAndBuild(buildOut io.Writer) ([]build.Artifact, func(), error) 
 		}
 	}
 	if err != nil {
-		return nil, func() { cancel() }, errors.Wrap(err, "creating runner")
+		return nil, errors.Wrap(err, "creating runner")
 	}
-
-	cleanUp := func() {
-		cancel()
-		runner.RPCServerShutdown()
-	}
-	a, err := runner.BuildAndTest(ctx, buildOut, targetArtifacts)
-	return a, cleanUp, err
+	defer runner.RPCServerShutdown()
+	return runner.BuildAndTest(ctx, buildOut, targetArtifacts)
 }
--- a/cmd/skaffold/app/cmd/build.go
+++ b/cmd/skaffold/app/cmd/build.go
@@ -35,6 +35,11 @@ var (
 	buildFormatFlag = flags.NewTemplateFlag("{{range .Builds}}{{.ImageName}} -> {{.Tag}}\n{{end}}", BuildOutput{})
 )
 
+// For testing
+var (
+	createRunnerAndBuildFunc = createRunnerAndBuild
+)
+
 // NewCmdBuild describes the CLI command to build artifacts.
 func NewCmdBuild(out io.Writer) *cobra.Command {
 	cmd := &cobra.Command{
@@ -48,8 +53,8 @@ func NewCmdBuild(out io.Writer) *cobra.Command {
 	}
 	AddRunDevFlags(cmd)
 	cmd.Flags().StringArrayVarP(&opts.TargetImages, "build-image", "b", nil, "Choose which artifacts to build. Artifacts with image names that contain the expression will be built only. Default is to build sources for all artifacts")
-	cmd.Flags().BoolVarP(&quietFlag, "quiet", "q", false, "Suppress the build output and print image built on success")
-	cmd.Flags().VarP(buildFormatFlag, "output", "o", buildFormatFlag.Usage())
+	cmd.Flags().BoolVarP(&quietFlag, "quiet", "q", false, "Suppress the build output and print image built on success. See --output to format output. ")
+	cmd.Flags().VarP(buildFormatFlag, "output", "o", "Used in conjuction with --quiet flag. "+buildFormatFlag.Usage())
 	return cmd
 }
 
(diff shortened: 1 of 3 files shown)


cmd/skaffold/app/cmd/build.go
```diff
@@ -100,3 +91,26 @@ func runBuild(out io.Writer) error {
...
+
+	runner, config, err := newRunner(opts)
+	var targetArtifacts []*latest.Artifact
+	for _, artifact := range config.Build.Artifacts {
+		if runner.IsTargetImage(artifact) {
+			targetArtifacts = append(targetArtifacts, artifact)
+		}
+	}
+	if err != nil {
+		return nil, func() { cancel() }, errors.Wrap(err, "creating runner")
```
dgageot: Could you return `cancel` instead of `func() { cancel() }`?

cmd/skaffold/app/cmd/build.go
```diff
@@ -100,3 +91,26 @@ func runBuild(out io.Writer) error {
 
 	return nil
 }
+
+func createRunnerAndBuild(buildOut io.Writer) ([]build.Artifact, func(), error) {
+	ctx, cancel := context.WithCancel(context.Background())
```
dgageot: Can you move those two lines outside the function to avoid returning a func()?

cmd/skaffold/app/cmd/build_test.go
```diff
@@ -0,0 +1,124 @@
...
+			Tag:       "test",
+		}}, func() {}, nil
+	}
+
+	orginalCreateRunner := createRunnerAndBuildFunc
+	defer func(c func(buildOut io.Writer) ([]build.Artifact, func(), error)) {
+		createRunnerAndBuildFunc = c
+	}(orginalCreateRunner)
+	var tests = []struct {
+		name           string
```
dgageot: This is called `description` is almost every test

cmd/skaffold/app/cmd/build_test.go
```diff
@@ -0,0 +1,124 @@
...
+		template       string
+		expectedOutput []byte
+		mock           func(io.Writer) ([]build.Artifact, func(), error)
+		shdErr         bool
```
dgageot: This is called `shouldErr` is almost every test

cmd/skaffold/app/cmd/build_test.go
```diff
@@ -0,0 +1,124 @@
...
+	}{
+		{
+			name:           "quiet flag print build images with no template",
+			expectedOutput: []byte("{[{gcr.io/skaffold/example test}]}"),
+			shdErr:         false,
```
dgageot: We usually skip those `shdErr: false`

cmd/skaffold/app/cmd/build.go
```diff
@@ -100,3 +95,18 @@ func runBuild(out io.Writer) error {
...
+	var targetArtifacts []*latest.Artifact
+	for _, artifact := range config.Build.Artifacts {
+		if runner.IsTargetImage(artifact) {
+			targetArtifacts = append(targetArtifacts, artifact)
+		}
+	}
+	if err != nil {
+		return nil, errors.Wrap(err, "creating runner")
+	}
+	defer runner.RPCServerShutdown()
```
dgageot: Shouldn't it be called just after `runner, config, err := newRunner(opts)`

cmd/skaffold/app/cmd/build.go
```diff
@@ -100,3 +95,18 @@ func runBuild(out io.Writer) error {
...
+
+func createRunnerAndBuild(ctx context.Context, buildOut io.Writer) ([]build.Artifact, error) {
+	runner, config, err := newRunner(opts)
+	var targetArtifacts []*latest.Artifact
+	for _, artifact := range config.Build.Artifacts {
+		if runner.IsTargetImage(artifact) {
+			targetArtifacts = append(targetArtifacts, artifact)
+		}
+	}
+	if err != nil {
```
dgageot: Should be just after we get the `err`

cmd/skaffold/app/cmd/build.go
tejal29: yeah i think, i moved code over by mistake.

cmd/skaffold/app/cmd/build_test.go
```diff
@@ -0,0 +1,127 @@
...
+	errRunner := func(c context.Context, buildOut io.Writer) ([]build.Artifact, error) {
+		return nil, errors.New("some error")
+	}
+	mockCreateRunner := func(c context.Context, buildOut io.Writer) ([]build.Artifact, error) {
+		return []build.Artifact{{
+			ImageName: "gcr.io/skaffold/example",
+			Tag:       "test",
+		}}, nil
+	}
+	orginalCreateRunner := createRunnerAndBuildFunc
```
dgageot: This can be simplified:

```
defer func(f func(context.Context, io.Writer) ([]build.Artifact, error)) {createRunnerAndBuildFunc = f}(createRunnerAndBuildFunc)
```

The original value of `createRunnerAndBuildFunc` is captured so you don't need to create an `orginalCreateRunner` variable

cmd/skaffold/app/cmd/build_test.go
```diff
@@ -0,0 +1,127 @@
...
+			createRunnerAndBuildFunc = test.mock
+			var output bytes.Buffer
+			err := runBuild(&output)
+			testutil.CheckErrorAndDeepEqual(t, test.shouldErr, err, string(test.expectedOutput), output.String())
+		})
+	}
+}
+
+func TestRunBuild(t *testing.T) {
+	errRunner := func(c context.Context, buildOut io.Writer) ([]build.Artifact, error) {
```
dgageot: I would omit the argument names since they are not used.

cmd/skaffold/app/cmd/build_test.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, errors.New("some error")
+	}
+	mockCreateRunner := func(c context.Context, buildOut io.Writer) ([]build.Artifact, error) {
```
dgageot: I would omit the argument names since they are not used.

cmd/skaffold/app/cmd/build_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+func TestQuietFlag(t *testing.T) {
+	mockCreateRunner := func(c context.Context, buildOut io.Writer) ([]build.Artifact, error) {
+		return []build.Artifact{{
+			ImageName: "gcr.io/skaffold/example",
+			Tag:       "test",
+		}}, nil
+	}
+
+	orginalCreateRunner := createRunnerAndBuildFunc
```
dgageot: same as below (Sorry I read from bottom to top...)

cmd/skaffold/app/cmd/build_test.go
tejal29: umm the value of createRunnerAndBuildFunc is set to a mock function here, 
```
	for _, test := range tests {
			quietFlag = true
			defer func() { quietFlag = false }()
			if test.template != "" {
				buildFormatFlag = flags.NewTemplateFlag(test.template, BuildOutput{})
			}
			defer func() { buildFormatFlag = nil }()
			createRunnerAndBuildFunc = test.mock
```

cmd/skaffold/app/cmd/build_test.go
tejal29: same. they are set to mock functions in the test runs.

cmd/skaffold/app/cmd/build_test.go
tejal29: fixed.

cmd/skaffold/app/cmd/build_test.go
tejal29: fixed

PR #1844, opened 2019-03-20
Design proposal for sync improvements

corneliusweig:
As discussed in #1812

### DIFF
--- a/docs/design_proposals/sync-improvements.md
+++ b/docs/design_proposals/sync-improvements.md
@@ -221,8 +221,10 @@ Resolution: descoped, open separate issue
 
 
 #### Migration plan
-The new configuration supports all existing use-cases and is fully backwards compatible.
 An automatic schema upgrade shall be implemented.
+The former default behavior to flatten directories at the destination will be dropped.
+This is possible, because sync is still an alpha feature.
+Users who are using incompatible sync patterns will receive a warning during upgrade.
 
 ### Open Issues/Question
 
--- a/docs/design_proposals/sync-improvements.md
+++ b/docs/design_proposals/sync-improvements.md
@@ -67,61 +67,42 @@ Whether the bazel builder supports this, is unclear at the moment.
 
 ### Interface changes
 
-There are at least two relevant changes
+The builder implementations need to fulfill a new requirement:
+They need to provide a default sync map for changed or deleted files.
+For that, the `Builder` interface (pkg/skaffold/build/build.go) will receive the new method `SyncMap`.
 
-1. When builders are queried for their dependencies, they must provide a list of destinations together with local source paths. The signature of `Builder.GetDependencies` (pkg/skaffold/build/build.go) therefore needs to change to:
-   ```go
-   type Builder interface {
-       Labels() map[string]string
+```go
+type Builder interface {
+   Labels() map[string]string
 
-       Build(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact) ([]Artifact, error)
+   Build(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact) ([]Artifact, error)
 
-       // OLD
-       // DependenciesForArtifact(ctx context.Context, artifact *latest.Artifact) ([]string, error)
-       // NEW
-       DependenciesForArtifact(ctx context.Context, artifact *latest.Artifact) (map[string][]string, error)
-   }
-   ```
+   DependenciesForArtifact(ctx context.Context, artifact *latest.Artifact) ([]string, error)
 
-   So far, `DependenciesForArtifact` returned a list of local paths (relative to the working directory of the artifact) that are input files for the container.
-   Now, `DependenciesForArtifact` serves two purposes. As before, the map keys are relative local paths of input files for the container.
-   In addition, the map values list all absolute remote destination paths where the input files are placed in the container.
-   This has to be a list, because a single input file may be copied to several locations in the container.
+   // ADDED
+   SyncMap() (map[string][]string, error)
+}
+```
 
-   Thus, `DependenciesForArtifact` serves two purposes: firstly, it lists all input files to watch for changes.
-   Secondly, it provides default sync destinations for changed or deleted files.
-   Note that this does not mean, that every file will be sync'ed. If a file is entitled for syncing depends on the user-provided sync rules.
+`SyncMap` returns a map whose keys paths of source files for the container.
+These host paths will be relative to the workingdir of the respective artifact.
+The map values list all absolute remote destination paths where the input files are placed in the container.
+This has to be a list, because a single input file may be copied to several locations in the container.
+Note that this does not mean, that every file will be sync'ed. If a file is entitled for syncing depends on the user-provided sync rules.
 
-   For example, given the above Dockerfile, it will return something like
-   ```go
-   map[string][]string{
-    "static/page1.html": []string{"/var/www/page1.html"},
-    "static/assets/style.css": []string{"/var/www/assets/style.css"},
-    "index.html": []string{"/var/www/index.html"},
-    "nginx.conf": []string{"/var/nginx.conf", "/etc/nginx.conf"},
-   }
-   ```
+For example, given the above Dockerfile, it will return something like
+```go
+map[string][]string{
+  "static/page1.html": []string{"/var/www/page1.html"},
+  "static/assets/style.css": []string{"/var/www/assets/style.css"},
+  "index.html": []string{"/var/www/index.html"},
+  "nginx.conf": []string{"/var/nginx.conf", "/etc/nginx.conf"},
+}
+```
 
-2. To support additional sync maps from builders, the builder interface will need an additional method `SyncMap`:
-   ```go
-   type Builder interface {
-       Labels() map[string]string
+Unsupported builders will return a not-supported error.
 
-       Build(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact) ([]Artifact, error)
-
-       DependenciesForArtifact(ctx context.Context, artifact *latest.Artifact) (map[string][]string, error)
-    
-       // ADDED
-       SyncMap() map[string]string
-   }
-   ```
-
-   The new `SyncMap` function shall return pairs of
-
-   - **key**: a glob pattern of local files to sync into the container
-   - **value**: a destination path in the container
-
-   By convention, the sync rule from a builder never flattens directories but does subdir syncing instead (also see open question below).
+**Note**: In the original version of this proposal, the signature of `GetDependencies` was changed. This is no longer the case
 
 ### Config changes
 
--- a/docs/design_proposals/sync-improvements.md
+++ b/docs/design_proposals/sync-improvements.md
@@ -24,6 +24,28 @@ requires a long list of sync patterns (#1807).
 Furthermore, builders should be able to inform Skaffold about additional sync paths (#1704).
 This will result in much faster deploy cycles, without users having to bother about these optimizations.
 
+#### Problems with existing syntax
+The current `sync` config syntax has the following problems:
+
+- The flattening of the file structure at the destination is surprising. For example:
+  ```yaml
+  sync:
+    'src/**/*.js': dest/
+  ```
+  will put all `.js` files in the `dest` folder.
+- The triple-star syntax is is quite implicit and a surprising extension of the standard globbing syntax. For example:
+  ```yaml
+  sync:
+    'src/b/c/***/*.py': dest/
+  ```
+  will recreate subfolders _below_ `src/b/c` at `dest/` and sync `.py` files.
+  The triple-star therefore triggers two actions which should not depend on each other:
+  - do not flatten the directory structure
+  - strip everything on the source path up to `***`
+- The current syntax does not allow extension to further sync modes unless further magic strings are introduced.
+  
+The syntax should therefore be revised.
+
 ## Design
 
 Skaffold sync shall have three different modes:
@@ -62,7 +84,7 @@ Builders know where to put static files in the container.
 This information needs to be pulled from the builders and made available in Skaffold.
 Currently, we know how to do this for Docker artifacts (@r2d4 suggested this in #1166).
 Jib transforms some input files and includes others as static resources.
-The specific sync maps need to be detailed out but are should be straightforward.
+The specific sync maps need to be detailed out but are straightforward.
 Whether the bazel builder supports this, is unclear at the moment.
 
 ### Interface changes
@@ -84,15 +106,15 @@ type Builder interface {
 }
 ```
 
-`SyncMap` returns a map whose keys paths of source files for the container.
+`SyncMap` returns a map whose keys are paths of source files for the container.
 These host paths will be relative to the workingdir of the respective artifact.
 The map values list all absolute remote destination paths where the input files are placed in the container.
-This has to be a list, because a single input file may be copied to several locations in the container.
-Note that this does not mean, that every file will be sync'ed.
+This must be a list, because a single input file may be copied to several locations in the container.
+Note that this does not mean that every file will be sync'ed.
 The intent to sync a file needs to be explicitly specified by user-provided sync rules (see config changes).
 In the future, some builders may specify default sync rules, but this is out of scope for this document.
 
-For example, given the above Dockerfile, it will return something like
+For example, given the above Dockerfile, `SyncMap` will return something like
 ```go
 map[string][]string{
   "static/page1.html": []string{"/var/www/page1.html"},
@@ -104,8 +126,6 @@ map[string][]string{
 
 Unsupported builders will return a not-supported error.
 
-**Note**: In the original version of this proposal, the signature of `GetDependencies` was changed. This is no longer the case
-
 ### Config changes
 
 The `artifact.sync` config needs to support additional requirements:
@@ -114,28 +134,11 @@ The `artifact.sync` config needs to support additional requirements:
   - Glob patterns on the host path
   - Transplant a folder structure, ~~or flatten the folder structure~~[this behavior was a bug and will be removed]
 - Offer a way to use destination inference.
-- Avoid ambiguity
-
-#### Problems with existing syntax
-- The flattening of the file structure at the destination is surprising. For example:
-  ```yaml
-  sync:
-    'src/**/*.js': dest/
-  ```
-  will put all `.js` files in the `dest` folder.
-- The triple-star syntax is is quite implicit and a surprising extension of the standard globbing syntax. For example:
-  ```yaml
-  sync:
-    'src/b/c/***/*.py': dest/
-  ```
-  will recreate subfolders _below_ `src/b/c` at `dest/` and sync `.py` files.
-  The triple-star therefore triggers two actions which should not depend on each other:
-  - do not flatten the directory structure
-  - strip everything on the source path up to `***`
+- Provide a concise and unambiguous configuration syntax to avoid the problems outlined [above](#Background).
+- Be open for extension, such as the `smart` sync mode.
 
 #### Suggestion
-These above problems can be made less surprising by a more explicit configuration syntax.
-The `artifact.sync` will be upgraded as follows:
+The `artifact.sync` config will be upgraded as follows:
 
 ```yaml
 sync:
@@ -154,7 +157,7 @@ sync:
   smart: {}       # oneOf=sync
 ```
 
-In the `sync` config, _one of_ the modes `manual` or `infer` may be specified, but not both.
+In the `sync` config, exactly one of the modes `manual` or `infer` may be specified, but not both.
 Both `manual` and `infer` contain a list of sync rules.
 When determining the destination for a given file, all sync rules will be considered and not just the first match.
 Thus, a single file may be copied to multiple destinations.
@@ -162,11 +165,11 @@ Thus, a single file may be copied to multiple destinations.
 Error cases:
 - It should be an error to specify `strip` with a prefix which which does not match the static prefix of the `from` field.
   ```yaml
-  - from: 'src/**/*.js'
-    to: dest/
+  - src: 'src/**/*.js'
+    dest: dest/
     strip: src/sub        # ERR
-  - from: 'src/**/*.js'
-    to: dest/
+  - src: 'src/**/*.js'
+    dest: dest/
     strip: app/           # ERR
   ```
 
@@ -196,10 +199,10 @@ The current config only suggests additive rules. For some applications it might 
 ```yaml
 sync:
 # include all files below src...
-- from: 'src/**/*'
-  to: app/
+- src: 'src/**/*'
+  dest: app/
 # ...but do not consider js files for sync
-- from: 'src/**/*.js'
+- src: 'src/**/*.js'
   exclude: true
 ```
 Also see #1766.
@@ -226,10 +229,10 @@ Overlapping copies can be caused by at least two ways:
 2. Two sync rules may copy files to the same destination. For example:
    ```yaml
    sync:
-   - from: foo
-     to: /baz
-   - from: bar
-     to: /baz
+   - src: foo
+     dest: /baz
+   - src: bar
+     dest: /baz
    ```
 
 Should such cases be treated as errors, warnings, or not be detected at all?
@@ -285,7 +288,9 @@ Resolution: This is an implementation detail.
    Should already support multiple destination paths. (#1847)
 2. Add inference logic for docker and examples. (#1812)
 3. Support support sync rules with inference. (former #1812, will become separate PR)
-4. (out of scope) Support smart sync mode by allowing builders to specify default sync rules.
+4. (out of scope) Smart sync mode by allowing builders to specify default sync rules.
+   - Support smart sync for jib.
+   - Possibly support smart sync for Dockerfile when `ADD` or `COPY` are not followed by a `RUN` command.
 5. (future) Add sync map logic for jib and examples.
 6. (future) Add sync map logic for bazel and examples.
 
--- a/docs/design_proposals/sync-improvements.md
+++ b/docs/design_proposals/sync-improvements.md
@@ -3,7 +3,7 @@
 * Author(s): Cornelius Weig (@corneliusweig)
 * Design Shepherd: Tejal Desai (@tejal29)
 * Date: 03/20/2019
-* Status: Draft
+* Status: Reviewed
 
 ## Background
 
@@ -28,9 +28,9 @@ This will result in much faster deploy cycles, without users having to bother ab
 
 Skaffold sync shall have three different modes:
 
-1. _direct_: the user specifies both the source and destination.
+1. _manual_: the user specifies both the source and destination.
 3. _auto-inferred destination_: the user specifies the syncable files, and Skaffold infers their destination in the container, e.g based on Dockerfile
-2. _smart_: builder recommended sync mapping for files, e.g. jib specifies what to sync and provides both the source and dest of files to sync. The user does not have to specify anything.
+2. _smart_: builder recommended sync rules for files, e.g. jib specifies what to sync and provides both the source and dest of files to sync. The user does not have to specify anything.
 
 The scope of this design proposal is the _direct_ and _inferred_ mode.
 The smart sync mode is out of scope for this document but may be mentioned at times.
@@ -88,7 +88,9 @@ type Builder interface {
 These host paths will be relative to the workingdir of the respective artifact.
 The map values list all absolute remote destination paths where the input files are placed in the container.
 This has to be a list, because a single input file may be copied to several locations in the container.
-Note that this does not mean, that every file will be sync'ed. If a file is entitled for syncing depends on the user-provided sync rules.
+Note that this does not mean, that every file will be sync'ed.
+The intent to sync a file needs to be explicitly specified by user-provided sync rules (see config changes).
+In the future, some builders may specify default sync rules, but this is out of scope for this document.
 
 For example, given the above Dockerfile, it will return something like
 ```go
@@ -133,46 +135,31 @@ The `artifact.sync` config needs to support additional requirements:
 
 #### Suggestion
 These above problems can be made less surprising by a more explicit configuration syntax.
-The `artifact.sync` will be upgraded to a list of *sync rules*.
+The `artifact.sync` will be upgraded as follows:
 
 ```yaml
 sync:
-# Existing use-case: Copy all js files under 'src' flat into 'app'
-# This corresponds to the current `'src/**/*.js': app/`
-- from: 'src/**/*.js'
-  to: app/
-  flatten: true  # default to false
+  # Existing use-case: Copy all js under 'src' into 'dest', re-creating the directory structure below 'src'
+  # This corresponds to the current `'src/***/*.js': app/`
+  manual:         # oneOf=sync
+  - src: 'src/**/*.js' # required
+    dest: app/         # required
+    strip: 'src/'      # optional, default to ""
 
-# Existing use-case: Copy all js under 'src' into 'dest', re-creating the directory structure below 'src'
-# This corresponds to the current `'src/***/*.js': dest/`
-- from: 'src/**/*.js'
-  to: dest/
-  strip: 'src'   # default to ""
-
-# New use-case: Copy all python files and infer their destination.
-- from: '**/*.py'
-  inferTo: true   # <- tbd, default to false
+  # New use-case: Copy all python files and infer their destination.
+  infer:          # oneOf=sync
+  - '**/*.py'
+    
+  # New use-case: smart sync mode for jib (out of scope)
+  smart: {}       # oneOf=sync
 ```
 
-When determining the destination for a given file, all sync rules will be considered.
+In the `sync` config, _one of_ the modes `manual` or `infer` may be specified, but not both.
+Both `manual` and `infer` contain a list of sync rules.
+When determining the destination for a given file, all sync rules will be considered and not just the first match.
 Thus, a single file may be copied to multiple destinations.
 
 Error cases:
-- It should be an error to specify `inferTo=true` together with `to` in the same sync rule.
-  ```yaml
-  - from: '*.js'
-    to: dest/
-    inferTo: true
-  ```
-- It should be an error to specify either of `flatten` or `strip` together with `inferTo`.
-  ```yaml
-  - from: '*.js'
-    inferTo: true
-    flatten: true        # ERR
-  - from: 'src/**/*.js'
-    inferTo: true
-    strip: src/          # ERR
-  ```
 - It should be an error to specify `strip` with a prefix which which does not match the static prefix of the `from` field.
   ```yaml
   - from: 'src/**/*.js'
@@ -195,14 +182,14 @@ Resolution:  YES
 A different approach could be to fall back to inference, if the `to` field is left blank.
 This would have the advantage that there needs to be no validation about having `to` and `inferTo` in the same rule, which should cause an error.
 
-Resolution: __Not Yet Resolved__
+Resolution: Obsolete, the improved version excludes this error case.
 
 **\<Error for `strip` and `flatten`\>**
 Given the new sync config, users may try to combine various flags in unintended ways.
 For example, should `strip` and `flatten` in the same rule always be an error?
 My intuition would say, that `strip+flatten` is the same as `flatten`, but this should to be discussed.
 
-Resolution: __Not Yet Resolved__
+Resolution: Obsolete.
 
 **\<Do we need an `excludes` rule?\>**
 The current config only suggests additive rules. For some applications it might be easier to also remove matched host paths from the syncable set. For example:
@@ -217,12 +204,12 @@ sync:
 ```
 Also see #1766.
 
-Resolution: descoped, open separate issue
+Resolution: Descoped, open separate issue
 
 
 #### Migration plan
 An automatic schema upgrade shall be implemented.
-The former default behavior to flatten directories at the destination will be dropped.
+The former default behavior to flatten directories at the destination will no longer be supported.
 This is possible, because sync is still an alpha feature.
 Users who are using incompatible sync patterns will receive a warning during upgrade.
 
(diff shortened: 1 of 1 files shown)


corneliusweig:
@tejal29 I can't finish this right now, but I'll try to find time later today.

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,146 @@
+# Title
+
+* Author(s): @corneliusweig
+* Design Shepherd: As mentioned in the document here.
```
tejal29: Sorry my bad. I forgot to add a link to here which points to Readme

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,265 @@
...
+
+1. When builders are queried for their dependencies, they must provide a list of destinations together with local source paths. The signature of `Builder.GetDependencies` (pkg/skaffold/build/build.go) therefore needs to change to:
+   ```go
+   type Builder interface {
+       Labels() map[string]string
+
+       Build(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact) ([]Artifact, error)
+
+       // OLD
+       // DependenciesForArtifact(ctx context.Context, artifact *latest.Artifact) ([]string, error)
```
tejal29: Was `[]string` before, just a list of local destination paths?

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,265 @@
+# Sync improvements
+
+* Author(s): Cornelius Weig (@corneliusweig)
+* Design Shepherd: Techal Desai (@tejal29)
```
tejal29:
```suggestion
* Design Shepherd: Tejal Desai (@tejal29)
```

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,265 @@
...
+   ```go
+   type Builder interface {
+       Labels() map[string]string
+
+       Build(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact) ([]Artifact, error)
+
+       // OLD
+       // DependenciesForArtifact(ctx context.Context, artifact *latest.Artifact) ([]string, error)
+       // NEW
+       DependenciesForArtifact(ctx context.Context, artifact *latest.Artifact) (map[string][]string, error)
```
tejal29: Two questions here
1. Should `map[string][]string` be `map[string]string` i.e. one source should be mapped to one destination?
For Clarification, it will really nice if you provide an example of this func will return for the above dockerfile.
Also, is this list going to be source for sync?

2.  DependenciesForArtifacts provides a list of source file dependencies. `skaffold dev` will then watch all these files and do a rebuild if any file changes.  Is that accurate?
If yes, some of these files can be source files and do need a rebuild and not all sources from here should be synced.

There should be a way to distinguish between source files and resources like data files, html, css etc?

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,265 @@
...
+   }
+   ```
+
+### Config changes
+
+The `artifact.sync` config needs to support additional requirements:
+
+- Support existing use-cases for backwards compatibility, i.e.
+  - Glob patterns on the host path
+  - Transplant a folder structure, or flatten the folder structure
```
tejal29: Thank you for putting this up here. 
We are actually thinking we should revisit if we need `***` vs `**` notation.

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,265 @@
...
+  will recreate subfolders _below_ `src/b/c` at `dest/` and sync `.py` files.
+  The triple-star therefore triggers two actions which should not depend on each other:
+  - do not flatten the directory structure
+  - strip everything on the source path up to `***`
+
+#### Suggestion
+These above problems can be made less surprising by a more explicit configuration syntax.
+The `artifact.sync` will be upgraded to a list of *sync rules*.
+
+```yaml
```
tejal29: This is amazing! I think we should definitely get other's input in this and discuss in the biweekly.

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,265 @@
...
+2. Two sync rules may copy files to the same destination. For example:
+   ```yaml
+   sync:
+   - from: foo
+     to: /baz
+   - from: bar
+     to: /baz
+   ```
+   
+Should such cases be treated as errors, warnings, or not be detected at all?
```
tejal29: We can `warn` for now and then `error` later.
Right now, these do not get detected.
Changing it to `warn` will give users time to fix their config and still be able to proceed with their normal workflow.
After that, we can change this to `error`

WDYT?

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,265 @@
...
+FROM scratch
+COPY --from=0 baz bar
+```
+Ideally, the sync destination should be inferred as `foo -> bar`.
+
+This is quite an edge case and can easily be circumvented.
+However, it is quite complex to implement.
+Should we bother about this special case?
+   
+Resolution: __Not Yet Resolved__
```
tejal29: I would say, not right now. If people ask for it we can do a 2 step roll out where 
- first detect and warn
- then fail.

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,265 @@
...
+
+**\<Other builders\>**
+We don't know how to obtain a mapping of local paths to remote paths for other builders (jib, bazel).
+This may even have to be implemented upstream.
+Until we can support those builders, we to handle the case when a user tries to use inference with those builders.
+
+- Option 1: bail out with an error
```
tejal29: Bail out with "not yet supported error"
It would be great if we can bail out at the start but checking if "SyncMap" is implemented and the Builder who don't implement it throw an error.
Something like that.

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,265 @@
...
+For example: given the following Dockerfile
+
+```dockerfile
+FROM nginx
+WORKDIR /var
+COPY static *.html www/
+ADD nginx.conf .
+```
+
+The sync locations should be inferred as
```
briandealwis: The sync works for this example as it's a straightforward copy.  But as soon as we involve some transformation tool (e.g., webpack or babel or tsc), then this no longer works.

docs/design_proposals/sync-improvements.md
corneliusweig: Yes, only the local paths were extracted, thus `[]string`. What we want is now a mapping of local paths to several remote destinations.

docs/design_proposals/sync-improvements.md
corneliusweig: @me :man_facepalming:

docs/design_proposals/sync-improvements.md
corneliusweig: So for case 1 where the overlapping copies come from the builder, we should follow what the builders do. Docker does not treat this as warnings, so neither should we. However, our syncing must be consistent with the builders. In this case, we need to follow the "last-one-wins" policy.

For case 2, where the overlapping copies may come from (mis-)configuration, I am not a big fan of warnings, too. This is for two reasons:
1. The unix copy command also does not warn about existing destination paths. Instead it simply overwrites. The sync rules are simply a list of automated copy instructions, so the same safety guard should apply (namely none).
2. Overlapping destination paths are costly to detect. Yet, this check needs to run for every sync just to cover an edge case.

docs/design_proposals/sync-improvements.md
corneliusweig: If we want to throw an error, we must do it upfront during the Skaffold pipeline config validation. Otherwise we cannot differentiate between dependencies that are not copied into the container (e.g. Dockerfile) and builders that can not provide destinations paths.

But I also prefer Option 1, because it does not surprise users.

docs/design_proposals/sync-improvements.md
corneliusweig: I think for webpack/tsc the sync would still do the right thing, as long as webpack/tsc runs inside the container (see #1826).

However, you are right that this
```dockerfile
FROM golang as builder
COPY main.go .
RUN go build

FROM scratch
COPY --from=builder main .
```
will not work as intended.

However, I think that there can be no general solution here. In the end, only the developers know what files are static and what needs to be processed. This is what the sync rules are for after all.

Nevertheless, we could think about a warn/info if typical static files cause a rebuild. For example, if somebody changes a `index.html` file and no sync rule is set up, we could advise to do so. But I have to say, I'm not a big fan of this.

docs/design_proposals/sync-improvements.md
corneliusweig: 1. As suggested, I added further explanation about the function signature and the internal use-cases for `DependenciesForArtifact`.

2. Yes, `skaffold dev` watches the local files which are returned in the map keys.

You are right that some files are source input and should trigger a rebuild, whereas other files are just static assets which can be sync'ed. The way to distinguish between these groups is something that only the user can answer. For that, she needs to set up sync rules.

Could you have a look at the revised versionn if this point becomes clear enough?

corneliusweig:
@tejal29 I'm done for now. At your convenience, please have a second look.

tejal29:
@corneliusweig This looks good.

The next skaffold bi-weekly meeting is on 3rd April 9:30 am to 10:am
I put this DD on the agenda, Will you be able to join?

Please join https://groups.google.com/forum/#!forum/skaffold-users to get meeting invite.

corneliusweig:
@tejal29 Is that LA time (UTC-07:00)? If so, I can try to attend, but can't promise.

tejal29:
yes it is going to be 9:30 AM PST.

tejal29:
do-not merge until reviewed on 3rd April

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+       // DependenciesForArtifact(ctx context.Context, artifact *latest.Artifact) ([]string, error)
+       // NEW
+       DependenciesForArtifact(ctx context.Context, artifact *latest.Artifact) (map[string][]string, error)
+   }
+   ```
+   
+   So far, `DependenciesForArtifact` returned a list of local paths (relative to the working directory of the artifact) that are input files for the container.
+   Now, `DependenciesForArtifact` serves two purposes. As before, the map keys are relative local paths of input files for the container.
+   In addition, the map values list all absolute remote destination paths where the input files are placed in the container.
+   This has to be a list, because a single input file may be copied to several locations in the container.
```
briandealwis: I think it would be better to separate the inferred sync-map from `GetDependencies()`.  `GetDependencies` is called on any file change so it should be as lightweight as possible.  For some builders there are input files that are compiled/transformed/merged that won't map to a destination so we'd likely have a lot of entries that just map to an empty array.

tejal29:
Summary after Today's discussion:
1. Everyone agrees with the proposed Sync Config. Looks much better and we shd start working on it!
   1. `flatten` true is is fishy, and maybe people are unknowingly using it to get wrong results. For backwards compatibility we should keep it there and  later could be marked for deprecation( We shd open an issue for it to address later once the new config changes land). We should not see much of "flatten = true"
      
1. Do we need to change the signature for `DependenciesForArtifact` to return `map[string][]string` and instead use `SyncMap()` ?
   PROs: we don't want to parse the Dockerfile twice and get that information when doing this.
1.  If we want to use SyncMaps instead of `DependenciesForArtifacts` its signature shd be changed.
     - Shd we change it regardless?

Looking forward to get this in!

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+  ```
+
+Open questions about the configuration:
+
+**\<Is the syntax clear enough?\>**
+Is the syntax clear?
+Is the configuration intuitive to use?
+Is there any ambiguity which needs attention?
+
+Resolution: __Not Yet Resolved__
```
tejal29:
```suggestion
Resolution:  YES
```

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+
+**\<Does it improve clarity to have the `inferTo` field?\>**
+A different approach could be to fall back to inference, if the `to` field is left blank.
+This would have the advantage that there needs to be no validation about having `to` and `inferTo` in the same rule, which should cause an error.
+
+Resolution: __Not Yet Resolved__
```
tejal29:
```suggestion
Resolution:  Yes, `inferTo` makes it more verbose.
```

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+Currently, skaffold config supports `artifact.sync` as a way to sync files
+directly to pods. So far, artifact sync requires a specification of sync
+patterns like
+
+```yaml
+sync:
+  '*.js': app/
+```
+
+This is error prone and unnecessarily hard to use, because the destination is
```
briandealwis:
```suggestion
This is error prone and unnecessarily hard to use because the destination can often be determined by the builder. For example a destination path is
```

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+# include all files below src...
+- from: 'src/**/*'
+  to: app/
+# ...but do not consider js files for sync
+- from: 'src/**/*.js'
+  exclude: true
+```
+Also see #1766.
+
+Resolution: __Not Yet Resolved__
```
tejal29: For now, lets go ahead and de-scope it.
With the this Proposal, it will be easy to support this in future.
@balopat WDYT?

docs/design_proposals/sync-improvements.md
tejal29: Makes sense. I think when you run skaffold dev with `--verbose` we do see all the computed sync files and users can use that to debug.

docs/design_proposals/sync-improvements.md
briandealwis: When I read through the earlier version, it wasn't clear that your approach required the user to opt-in to the change by defining a sync-rule in the skaffold.yaml.  Could you please call this out more explicitly above?

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+   sync:
+   - from: foo
+     to: /baz
+   - from: bar
+     to: /baz
+   ```
+   
+Should such cases be treated as errors, warnings, or not be detected at all?
+
+Resolution: __Not Yet Resolved__
```
tejal29:
```suggestion
Resolution: No detection which is the current behavior.
```

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+
+**\<Other builders\>**
+We don't know how to obtain a mapping of local paths to remote paths for other builders (jib, bazel).
+This may even have to be implemented upstream.
+Until we can support those builders, we to handle the case when a user tries to use inference with those builders.
+
+- Option 1: bail out with an error
+- Option 2: copy the file into the container with the destination set to the relative path wrt the context directory.
+
+Resolution: __Not Yet Resolved__
```
tejal29:
```suggestion
Resolution: Option 1?
```

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+       SyncMap() map[string]string
+   }
+   ```
+   
+   The new `SyncMap` function shall return pairs of
+   
+   - **key**: a glob pattern of local files to sync into the container
+   - **value**: a destination path in the container
+   
+   By convention, the sync rule from a builder does never flatten directories but does subdir syncing instead (also see open question below).
```
briandealwis:
```suggestion
   By convention, the sync rule from a builder never flattens directories but does subdir syncing instead (also see open question below).
```

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+- Option 2: copy the file into the container with the destination set to the relative path wrt the context directory.
+
+Resolution: __Not Yet Resolved__
+
+**\<Upgrade hint\>**
+The path inference allows to simplify the sync specification considerably.
+Besides, subdirectory syncing makes the sync functionality even feasible for large projects that need to sync many folders into the container.
+Therefore, we could consider to advertise this new functionality when running Skaffold or when an old-style sync map as found.
+
+Resolution: __Not Yet Resolved__
```
tejal29:
```suggestion
Resolution: Agree. We should support auto-fixing.
```

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+**\<Is the syntax clear enough?\>**
+Is the syntax clear?
+Is the configuration intuitive to use?
+Is there any ambiguity which needs attention?
+
+Resolution: __Not Yet Resolved__
+
+**\<Does it improve clarity to have the `inferTo` field?\>**
+A different approach could be to fall back to inference, if the `to` field is left blank.
+This would have the advantage that there needs to be no validation about having `to` and `inferTo` in the same rule, which should cause an error.
```
briandealwis: I prefer this approach: it seems pretty intuitive that a missing `to` means _fill me in_.

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+
+Resolution: __Not Yet Resolved__
+
+**\<SyncMap signature\>** 
+The new `SyncMap` interface provides some sync rules from the builder.
+However, the suggested signature is `func() map[string]string` with the convention to not flatten directories.
+To be more in line with the other sync functionality, it could also return `[]SyncRule`.
+Although being clearer, this has the downside of introducing a dependency from the builder package to the pipeline config package.
+If going that route, we should probably duplicate this struct in a proper place.
+However, is it necessary at all, or is the current signature good enough?
```
tejal29: I like the idea of having `SyncRule` however not sure how the implementation will look , like you mentioned.
Shd we consider this as an implementation detail and proceed?

Re: `SyncMap` maybe we should change the signature to`map[string][]string` to be in Sync with the `DependenciesForAtifacts` and also more flexible if same file is synced in multiple places.

docs/design_proposals/sync-improvements.md
briandealwis: This situation seem similar to a situation when the destination cannot be inferred by the builder.
```
FROM scratch
  ADD foo baz
  RUN mv baz bar
FROM scratch
  COPY --from=0 bar bar
```

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+- `nginx.conf` -> `/var/nginx.conf`, `/etc/nginx.conf`
+
+A fundamental change compared to the current sync mechanism is that one source path may by sync'ed to multiple destinations in the future.
+
+### Assumptions
+
+Builders know where to put static files in the container.
+This information needs to be pulled from the builders and made available in Skaffold.
+Currently, we know how to do this for Docker artifacts (@r2d4 suggested this in #1166).
+Whether jib and bazel builders can provide that information too, is unclear at the moment.
```
briandealwis: Jib can provide this information.  For Jib, the static files are usually in `src/main/resources/` and then copied into the container image under `/app/resources`.  So Jib could return a sync map of `src/main/resources/** -> /app/resources`.

docs/design_proposals/sync-improvements.md
balopat: agreed, we can scope this out for now, create a separate issue to discuss.

docs/design_proposals/sync-improvements.md
balopat: +1 on error - preferably at schema validation time to fail fast but could be just inference time too

docs/design_proposals/sync-improvements.md
balopat: +1

docs/design_proposals/sync-improvements.md
balopat: +1 let's punt this to implementation time - I think you can remove this from here - but I do like the idea of `SyncRule`.

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+   ```go
+   type Builder interface {
+       Labels() map[string]string
+
+       Build(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact) ([]Artifact, error)
+
+       DependenciesForArtifact(ctx context.Context, artifact *latest.Artifact) (map[string][]string, error)
+    
+       // ADDED
+       SyncMap() map[string]string
```
balopat: technically this should be map[string][]string as well (`SyncRule`) to match copy semantics

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+This is error prone and unnecessarily hard to use, because the destination is
+already contained in the Dockerfile for docker build. (see #1166, #1581).
+In addition, the syncing needs to handle special cases for globbing and often
+requires a long list of sync patterns (#1807).
+
+Furthermore, builders should be able to inform Skaffold about additional sync paths (#1704).
+This will result in much faster deploy cycles, without users having to bother about these optimizations.
+
+## Design
+
```
balopat: I'd add a section here about 3 sync modes: 

   - "smart": builder recommended sync mapping for files, e.g. jib gives you both the source and dest of files to sync, you don't have to spec anything else 
   - "auto-inferred destination": you specify the files you're interested in, we infer the destination in the container, e.g based on Dockerfile 
   - "direct": you specify both the source and destination

I don't want to combine these three modes into hybrid scenarios _at first_. We could just have them implemented separately and figure out the combination rules later, if people are asking for it! 

"smart" is only available for jib (maybe?) and it is itself a separate design, so maybe we should just mention it here that the details are out of scope for this design proposal, but it should fit in the model as one of the options users can choose.

docs/design_proposals/sync-improvements.md
corneliusweig: I take it this does not refer to
> Is there any ambiguity which needs attention?

:wink:

docs/design_proposals/sync-improvements.md
corneliusweig: Just to clarify: is `src/main/resources` below the workingdir of the artifact? If this is not the case, we need to think how we can handle absolute paths on the host as well, because the current assumption is that `SyncMap` returns _relative_ paths wrt the workingdir.

docs/design_proposals/sync-improvements.md
corneliusweig: Yes, after thinking about this, I'm also much in favor of @briandealwis suggestion. Besides, avoiding the double responsibility for `GetDependencies` will probably save some headache in the future.
cc @balopat

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+Resolution: __Not Yet Resolved__
+
+**\<Does it improve clarity to have the `inferTo` field?\>**
+A different approach could be to fall back to inference, if the `to` field is left blank.
+This would have the advantage that there needs to be no validation about having `to` and `inferTo` in the same rule, which should cause an error.
+
+Resolution: __Not Yet Resolved__
+   
+**\<Error for `strip` and `flatten`\>**
```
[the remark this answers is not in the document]
corneliusweig: :bulb:

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+   Should already support multiple destination paths. (#1847)
+2. Add inference logic for docker and examples. (#1812)
+3. Support support sync rules with inference. (former #1812, will become separate PR)
+4. Finally, support builder plugins to add sync patterns.
+5. (future) Add inference logic for jib and examples.
+6. (future) Add inference logic for bazel and examples.
+
+## Integration test plan
+
+- **step 3** Change one of the existing sync examples so that it uses inference (e.g. #1826).
```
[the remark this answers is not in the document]
corneliusweig: So my idea was to reference the item in the implementation plan when this test case needs to be added. Therefore the numbers I put were correct, right?

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+# include all files below src...
+- from: 'src/**/*'
+  to: app/
+# ...but do not consider js files for sync
+- from: 'src/**/*.js'
+  exclude: true
+```
+Also see #1766.
+
+Resolution: __Not Yet Resolved__
```
corneliusweig:
```suggestion
Resolution: descoped, open separate issue
```

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+   
+   By convention, the sync rule from a builder does never flatten directories but does subdir syncing instead (also see open question below).
+
+### Config changes
+
+The `artifact.sync` config needs to support additional requirements:
+
+- Support existing use-cases for backwards compatibility, i.e.
+  - Glob patterns on the host path
+  - Transplant a folder structure, or flatten the folder structure
```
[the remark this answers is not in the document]
corneliusweig: So do I understand correctly that you want to drop support for `flatten` with the overhaul of the sync config? And when upgrading the config we should print a warning that the semantics has changed.
+1 for that :)

docs/design_proposals/sync-improvements.md
corneliusweig: @balopat @briandealwis So it just occurred to me that in my original proposal, `GetDependencies` and `SyncMap` were answering different questions. `GetDependencies` would answer _how_ to sync files, whereas `SyncMap` would answer _what_ files to sync.

For the reasons outlined above, we want to keep `GetDependencies` unchanged and `SyncMap` will get the semantics of _how_ to sync. However, this leaves open the question of _what_ to sync. As @balopat outlined above, this magic jib sync will just be mentioned and not be further detailed by this design proposal. However, I wanted to highlight that this is now out of scope for this design proposal. Correct?

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,319 @@
...
+  ```yaml
+  sync:
+    'src/b/c/***/*.py': dest/
+  ```
+  will recreate subfolders _below_ `src/b/c` at `dest/` and sync `.py` files.
+  The triple-star therefore triggers two actions which should not depend on each other:
+  - do not flatten the directory structure
+  - strip everything on the source path up to `***`
+
+#### Suggestion
```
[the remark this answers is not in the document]
corneliusweig: My thoughts about this alternative:
- Multiple destinations should be a rare exception. An exception we must to handle internally, but none that we need to expose to our users. If somebody wants to sync to multiple destinations, she can as well add another rule for the same src pattern.
- Should the user have to enable magic sync manually? I think this would be easily forgotten, because most users would not be aware of this. IMHO magic sync should be implicitly on, if available (and not overridden by a manual sync spec). So I would not consider this for the pipeline config.
- Finally, I'm not a big fan of the explicit `inferDest`. But maybe it makes the feature more discoverable.
@briandealwis @tejal29 WDYT?

docs/design_proposals/sync-improvements.md
corneliusweig: About the hybrid scenario: "direct" and "auto" mode can easily be combined. Is there any reason why you want to avoid this in particular?

corneliusweig:
@balopat @briandealwis @tejal29 Thanks a lot for your invaluable input here! At your convenience, please take another look. I still left the new sync config format open for discussion.

docs/design_proposals/sync-improvements.md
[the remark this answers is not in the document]
tejal29: sounds good to me given sync is in alpha.

docs/design_proposals/sync-improvements.md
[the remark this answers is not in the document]
tejal29: @corneliusweig and @balopat i am little bit skeptical about the key `direct`. Its confusing to me, maybe `manual` makes it better.

> Multiple destinations should be a rare exception.

Agree with this. If a file syncing to multiple directories is a side-effect of multiple rules, it ok.

@balopat  with your suggestion, would this  skaffold config be valid? 
```
sync: 
  manual: 
     - src: 'special-directory/**/*.js'
       strip: "special-directory/"  # default "" 
       dest: "dir1/"
  infer:
      - 'src/**/*.js'
      - 'src/**/*.html'
```
I really don't have a preference to be honest. This is something we can find out with Critical User Journeys.
But yes, you do end up writing less config with suggested change. Maybe that a win?

docs/design_proposals/sync-improvements.md
tejal29: +1 for out of scope now

tejal29:
LGTM for me! I don't have a preference on the config syntax like i mentioned.
nit: Do mention "Builder provided syncmaps is out of scope for now"

docs/design_proposals/sync-improvements.md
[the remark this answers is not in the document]
balopat: AHhhhhhh, I see. Then may be change it to "Implementation step 3"

docs/design_proposals/sync-improvements.md
balopat: yes!

docs/design_proposals/sync-improvements.md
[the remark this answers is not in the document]
balopat: > * Should the user have to enable magic sync manually? I think this would be easily forgotten, because most users would not be aware of this. IMHO magic sync should be implicitly on, if available (and not overridden by a manual sync spec). So I would not consider this for the pipeline config.

Filesync is a serious business - we are mutating the user's container. I think sync should be opt-in, and even if we get comfortable with it and turn it on by default after a while, I believe there will be users who want to be able to turn it off. 

I'm okay with `manual` maybe instead of `explicit`, I do like `infer`. 

Multiple dest: I'm convinced to not exposed that to the user, sounds good to keep dest single valued.

docs/design_proposals/sync-improvements.md
balopat: My guess is that: 

- For non-docker users it's only going to be "manual". 
- For Dockerfile users 99% of the cases should be covered by "infer". 
- For jib users hopefully most of it will be covered by "smart"...

Thus, I did not want to deal with the complexity of combining the "manual" and "infer" scenarios, unless it's really required. Why maintain precedence rules and merging logic code if no one will use it?

balopat: this is not saying that _when_ the above assumptions fail and users are asking for hybrid features we can't implement them - we can, but then we'll have some more data to work with.

docs/design_proposals/sync-improvements.md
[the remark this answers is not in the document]
balopat: discoverability: totally true - in IntelliJ, VSCode this structure will help a lot for users to find the options.

docs/design_proposals/sync-improvements.md
corneliusweig: Alright, you convinced me :)

corneliusweig:
@tejal29 Please take a (final?) look :)

docs/design_proposals/sync-improvements.md
briandealwis: Hmm… it should be.  Jib assumes the artifact context points to a Maven/Gradle project.  If it's a standalone project, it should have `src/main/...`.  If it's a multi-module project, where there are several subproject — like libraries — that are joined together to produce an image, there will be multiple `xxx/src/main/...` directories, but they should all be under the context directory too.

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,305 @@
...
+
+A fundamental change compared to the current sync mechanism is that one source path may by sync'ed to multiple destinations in the future.
+
+### Assumptions
+
+Builders know where to put static files in the container.
+This information needs to be pulled from the builders and made available in Skaffold.
+Currently, we know how to do this for Docker artifacts (@r2d4 suggested this in #1166).
+Jib transforms some input files and includes others as static resources.
+The specific sync maps need to be detailed out but are should be straightforward.
```
briandealwis:
```suggestion
The specific sync maps need to be detailed out but are straightforward.
```

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,305 @@
...
+In the future, some builders may specify default sync rules, but this is out of scope for this document.
+
+For example, given the above Dockerfile, it will return something like
+```go
+map[string][]string{
+  "static/page1.html": []string{"/var/www/page1.html"},
+  "static/assets/style.css": []string{"/var/www/assets/style.css"},
+  "index.html": []string{"/var/www/index.html"},
+  "nginx.conf": []string{"/var/nginx.conf", "/etc/nginx.conf"},
+}
```
briandealwis: Interesting: this also gives us the ability to perform renames: `static/page1.html` &rarr; `/var/www/_page1_.html`.

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+Unsupported builders will return a not-supported error.
+
+**Note**: In the original version of this proposal, the signature of `GetDependencies` was changed. This is no longer the case
```
briandealwis: I don't think this line is necessary.

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,305 @@
...
+
+The `artifact.sync` config needs to support additional requirements:
+
+- Support existing use-cases for backwards compatibility, i.e.
+  - Glob patterns on the host path
+  - Transplant a folder structure, ~~or flatten the folder structure~~[this behavior was a bug and will be removed]
+- Offer a way to use destination inference.
+- Avoid ambiguity
+
+#### Problems with existing syntax
```
briandealwis: This part provides a motivation for changing the syntax and might be worth moving to the top.

docs/design_proposals/sync-improvements.md
briandealwis: It seems to me that the Docker builder can be "smart" too: just apply the inferred rules *providing* there are no `RUN` statements after an `ADD` or `COPY`.  I don't know how common this case would be in practice though.

corneliusweig: That's a funny idea. I'll add that as possible future work.

docs/design_proposals/sync-improvements.md
corneliusweig: agreed

corneliusweig:
Thanks @briandealwis for another thorough review. If you like, take another look.

docs/design_proposals/sync-improvements.md
corneliusweig: moved it to the "Background" section

docs/design_proposals/sync-improvements.md
```diff
@@ -0,0 +1,310 @@
...
+- Be open for extension, such as the `smart` sync mode.
+
+#### Suggestion
+The `artifact.sync` config will be upgraded as follows:
+
+```yaml
+sync:
+  # Existing use-case: Copy all js under 'src' into 'dest', re-creating the directory structure below 'src'
+  # This corresponds to the current `'src/***/*.js': app/`
+  manual:         # oneOf=sync
```
tejal29: great!

tejal29:
great! this is ready to merge once kokoro build goes green

corneliusweig:
Thanks again to everyone who shaped this design proposal! In particular @tejal29 who initiated the whole design proposal process. Doing these revisions on the PR has helped a lot!

corneliusweig:
@tejal29 Do you think this is ready to be merged?

balopat:
@dgageot can you please have a look - it would be great to have your input before we jump in implementing this.

balopat:
Let's merge this in. @dgageot can comment on the actual PR later, and we'll adjust accordingly.

briandealwis:
Let me throw in an additional wrinkle that may be worth thinking about.  I glossed over some details previously when [I described how](https://github.com/GoogleContainerTools/skaffold/pull/1844#discussion_r271857743) Jib copies from `src/main/resources` to `/app/resources`.  Maven and Gradle actually copy and (optionally) transform static file into somewhere under their build results (`target/` for Maven, `build/` for Gradle), and then Jib copies those file copies into the destination container.  For example, Maven allows treating files as a template and interpolates them using variables from the build context.  

For Skaffold to handle this properly, Skaffold would need to know if a builder requires a rebuild prior to performing a sync.  It would require separating `SkaffoldRunner.buildTestDeploy()` to separate rebuilding from redeploying.  The rebuild could continue to push the resulting image to a registry — that is, Skaffold can push, but it shouldn't shove :-)

corneliusweig:
@briandealwis Just to clarify, I think "build" is used in two different contexts:
1. The container build
2. The java build, aka compilation

So I guess, 
>  if a builder requires a __rebuild__ prior to performing a sync

could be rephrased to 
>  if a builder requires a __recompile__ prior to performing a sync

Correct?

That kind of change probably needs to be picked up by the JIB team. However, the whole thing will only work, if the container allows some hot-swapping or hot-reloading of updated files. Otherwise, the main process in the container needs to be restarted, which kind of defeats the whole point of sync. Does JIB support a hot-swapping mode?

briandealwis:
You're right @corneliusweig; I just used "rebuild" as a shorthand as for Jib-based projects, as a _container build_ performs a recompile.  It might be advantageous for the builder interface to support _compile_ and _build_ (to build the container image), but it seems easy enough to just use _build_ since generating a container image is cheap.

It's not a real stretch to imagine Skaffold being able to sync files from a built container image: it's just a matter of trawling through the layers.

The hot-swapping/reloading is something that needs to be supported by the language runtime.  The JVM does support it via helper libraries like [_Spring Boot Dev Tools_](https://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-devtools.html).

corneliusweig:
@loosebazooka (I hope this is the correct github handle :) We talked yesterday about the magic/smart sync mode.
What I had in mind is https://github.com/GoogleContainerTools/skaffold/pull/1844/files#diff-9f76156c150848a57c51506d8c4354d1R55  and https://github.com/GoogleContainerTools/skaffold/pull/1844\#issuecomment-487944328 . But I think your ideas are already much further evolved.

PR #1848, opened 2019-03-21
removing unnecessary exit from plugin processes

balopat:
We thought that the default warning message means something is wrong when we saw it by default from hashicorps go-plugin after a Ctrl+C signal is received: 

```
{"@level":"debug","@message":"plugin received interrupt signal, ignoring","@timestamp":"2019-03-20T16:40:56.592099-07:00","count":1}
2019-03-20T16:40:56.593-0700 [DEBUG] plugin.skaffold: plugin received interrupt signal, ignoring: count=1 timestamp=2019-03-20T16:40:56.592-0700

```

It turns out it behaves exactly what it needs to: it swallows the signal and ignores it - leaving the chance for the main process to cleanup. 

The `plugins.CleanupPlugins()` call should be called from the main process which is now implemented there. 

The logging level for plugins can be configured via the `DefaultPluginLogLevel` that is now set to `Info` instead of `Debug`. A follow-up PR could be to get this from an environment variable that the main process sets for the plugin to set the same level of debugging as the main process.

### DIFF
--- a/cmd/skaffold/app/cmd/dev.go
+++ b/cmd/skaffold/app/cmd/dev.go
@@ -21,6 +21,8 @@ import (
 	"io"
 	"strings"
 
+	"github.com/hashicorp/go-plugin"
+
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner"
@@ -97,7 +99,6 @@ func dev(out io.Writer, ui bool) error {
 			if err != nil {
 				return errors.Wrap(err, "creating runner")
 			}
-			defer r.RPCServerShutdown()
 
 			err = r.Dev(ctx, output, config.Build.Artifacts)
 			if r.HasDeployed() {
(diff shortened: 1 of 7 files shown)


pkg/skaffold/build/plugin/core.go
```diff
@@ -17,62 +17,39 @@ limitations under the License.
...
-
-	go func() {
-		<-sigs
-		errCh <- cancelError
-	}()
-
-	err := <-errCh
-
-	if err == cancelError {
-		hashiplugin.CleanupClients()
```
balopat: this is the key removal - on Ctrl+C the plugin was self-destructing.

cmd/skaffold/app/cmd/dev.go
```diff
@@ -109,9 +110,13 @@ func dev(out io.Writer, ui bool) error {
 			}
 			if err != nil {
 				if errors.Cause(err) != runner.ErrorConfigurationChanged {
+					plugin.CleanupClients()
```
priyawadhwa: just a question -- why did you opt to have these twice and remove the defer function?

balopat: because you're not suppose to call defer in a for loop this way - it might actually end up causing problems across dev cycles. one nice explanation is here: https://blog.learngoprogramming.com/gotchas-of-defer-in-go-1-8d070894cb01 - I could have switched over to an anonymous function as well...

PR #1850, opened 2019-03-21
Don't expose ports to the outside and fix a race condition

dgageot:
Ports used for eventing shouldn't be exposed to the outside world. Also this removed an alert on OSX each time a port needs to be opened on the firewall.

I also took the opportunity to fix a race condition where a port could be said available when it is not.

### DIFF
--- a/pkg/skaffold/event/server.go
+++ b/pkg/skaffold/event/server.go
@@ -25,14 +25,13 @@ import (
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/event/proto"
+	gw "github.com/GoogleContainerTools/skaffold/pkg/skaffold/event/proto"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 	empty "github.com/golang/protobuf/ptypes/empty"
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
-
-	gw "github.com/GoogleContainerTools/skaffold/pkg/skaffold/event/proto"
 )
 
 type server struct{}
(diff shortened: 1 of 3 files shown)


pkg/skaffold/util/port.go
```diff
@@ -32,7 +32,6 @@ import (
 // See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt,
 func GetAvailablePort(port int, forwardedPorts *sync.Map) int {
 	if isPortAvailable(port, forwardedPorts) {
-		forwardedPorts.Store(port, true)
```
briandealwis: These stores are necessary as there are race conditions: port-forwarding gets an available port but doesn't use it immediately

pkg/skaffold/util/port.go
```diff
@@ -41,37 +40,39 @@ func GetAvailablePort(port int, forwardedPorts *sync.Map) int {
...
 	p := l.Addr().(*net.TCPAddr).Port
+
 	forwardedPorts.Store(p, true)
+	l.Close()
 	return p
 }
 
 func isPortAvailable(p int, forwardedPorts *sync.Map) bool {
-	if _, ok := forwardedPorts.Load(p); ok {
+	alreadyUsed, loaded := forwardedPorts.LoadOrStore(p, true)
```
briandealwis: It seems odd for an `is*` method to mutate state

briandealwis: Maybe rename the method to something like `getPortIfAvailable()`?

pkg/skaffold/event/server.go
```diff
@@ -116,12 +116,12 @@ func newGRPCServer(port int) (func() error, error) {
...
 	mux := runtime.NewServeMux()
 	opts := []grpc.DialOption{grpc.WithInsecure()}
-	err := gw.RegisterSkaffoldServiceHandlerFromEndpoint(context.Background(), mux, fmt.Sprintf(":%d", proxyPort), opts)
+	err := gw.RegisterSkaffoldServiceHandlerFromEndpoint(context.Background(), mux, fmt.Sprintf("127.0.0.1:%d", proxyPort), opts)
 	if err != nil {
 		return func() error { return nil }, err
 	}
 
-	l, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+	l, err := net.Listen("tcp", fmt.Sprintf("127.0.0.1:%d", port))
```
tejal29: maybe move "127.0.0.1" to a constant.

dgageot:
@tejal29 @briandealwis should be all good now

PR #1865, opened 2019-03-22
Tail logs for init containers

ottonello:
This PR makes output for init containers to be tailed together with the rest of the containers in the pods. So far these logs were being ignored.

### DIFF
--- a/pkg/skaffold/kubernetes/log.go
+++ b/pkg/skaffold/kubernetes/log.go
@@ -104,7 +104,7 @@ func (a *LogAggregator) Start(ctx context.Context) error {
 					continue
 				}
 
-				for _, container := range pod.Status.ContainerStatuses {
+				for _, container := range append(pod.Status.ContainerStatuses, pod.Status.InitContainerStatuses...) {
 					if container.ContainerID == "" {
 						if container.State.Waiting != nil && container.State.Waiting.Message != "" {
 							color.Red.Fprintln(a.output, container.State.Waiting.Message)


ottonello:
I signed it!

balopat:
Thank you for your first contribution! 
Now we just have to fight our way through tests :) 
1. The Travis issue is a flake - we are working on it. 
2. Kokoro is failing with a kaniko_local test - that might be a real issue, given that kaniko is using init containers. Can you please check? 
3. An integration test would be fantastic for this.

ottonello:
Hi @balopat indeed I've merged master. Haven't found anything in the kaniko_local test, seems to be ok. I'm working on an integration test, thanks!

pkg/skaffold/kubernetes/log.go
```diff
@@ -104,7 +104,7 @@ func (a *LogAggregator) Start(ctx context.Context) error {
 					continue
 				}
 
-				for _, container := range pod.Status.ContainerStatuses {
+				for _, container := range append(pod.Status.ContainerStatuses, pod.Status.InitContainerStatuses...) {
```
tejal29: InitContainers run to completion. Running `kubeclt logs --since` for completed initContainer is going to be expensive.
```
tejaldesai@tejaldesai~/initcontainers:$ time kubectl logs --since=20s -f init-demo -c install 

real	0m5.120s
user	0m0.105s
sys	0m0.020s

tejaldesai@tejaldesai~/initcontainers:$ time kubectl logs init-demo -c install 
init is running!
init is up again.

real	0m0.108s
user	0m0.106s
sys	0m0.015s
tejaldesai@tejaldesai~/initcontainers:$
```
this is going to slow down a lot if a user has multiple init containers. 
Please add a check to filter completed init containers if its run to completion.

It would be great to see if you add an integration test like @balopat mentioned.

PR #1869, opened 2019-03-22
[ DRAFT] Add two new flag types and Inputfilepath, Outputfilepath

tejal29:
Add two new flags, Inputfilepath, Outputfilepath
    1. Inputfilepath:
            Inputfilepath flag, makes sure the file present exists.
    2. Outputfilepath:
            Outputfilepath flag does not do any validation right now.
    
Motivation:
For, #922 I am going to add `--build-output` as new flag to `skaffold build` to
 dump all artifacts built with with their tags.
    
Similarly, the deploy will have additional flag `--build-artifacts` which will point to the output written by `skaffold build --build-output <file>`
   
 The Inputfilepath, can also be used to replace all the places where we
 expect an input file without repeating validation.
    
    ```
    (add_flag_to_deploy)$ git grep 'Filename or URL to the pipeline'
    cmd/skaffold/app/cmd/cmd.go:    cmd.Flags().StringVarP(&opts.ConfigurationFile, "filename", "f", "skaffold.yaml", "Filename or URL to the pipeline file")
    cmd/skaffold/app/cmd/diagnose.go:       cmd.Flags().StringVarP(&opts.ConfigurationFile, "filename", "f", "skaffold.yaml", "Filename or URL to the pipeline file")
    cmd/skaffold/app/cmd/fix.go:    cmd.Flags().StringVarP(&opts.ConfigurationFile, "filename", "f", "skaffold.yaml", "Filename or URL to the pipeline file")
    cmd/skaffold/app/cmd/init.go:   cmd.Flags().StringVarP(&opts.ConfigurationFile, "filename", "f", "skaffold.yaml", "Filename or URL to the pipeline file")
    ```

TODO
1. What validation for outputfile path?
1.  Find all flag which expect to read to write file and replace them with these new files ( will done in a follow up Pr)

### DIFF
--- a/cmd/skaffold/app/flags/filepath.go
+++ b/cmd/skaffold/app/flags/filepath.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package flags
+
+import (
+	"os"
+)
+
+type filepathFlag struct {
+	path        string
+	shouldExist bool
+}
+
+func (f *filepathFlag) SetIfValid(value string) error {
+	copy := f
+	copy.path = value
+	if err := copy.isValid(); err != nil {
+		return err
+	}
+	f = copy
+	return nil
+}
+
+func newFilepathFlag(value string, exists bool) filepathFlag {
+	return filepathFlag{
+		path:        value,
+		shouldExist: exists,
+	}
+}
+
+func (f filepathFlag) isValid() error {
+	if !f.shouldExist {
+		// Currently no validation implemented for output files.
+		return nil
+	}
+	if _, err := os.Stat(f.path); os.IsNotExist(err) {
+		return err
+	}
+	return nil
+}
--- a/cmd/skaffold/app/flags/filepath.go
+++ b/cmd/skaffold/app/flags/filepath.go
@@ -26,20 +26,13 @@ type filepathFlag struct {
 }
 
 func (f *filepathFlag) SetIfValid(value string) error {
-	copy := f
+	copy := *f
 	copy.path = value
 	if err := copy.isValid(); err != nil {
 		return err
 	}
-	f = copy
+	*f = copy
 	return nil
-}
-
-func newFilepathFlag(value string, exists bool) filepathFlag {
-	return filepathFlag{
-		path:        value,
-		shouldExist: exists,
-	}
 }
 
 func (f filepathFlag) isValid() error {
--- a/cmd/skaffold/app/flags/filepath.go
+++ b/cmd/skaffold/app/flags/filepath.go
@@ -26,12 +26,12 @@ type filepathFlag struct {
 }
 
 func (f *filepathFlag) SetIfValid(value string) error {
-	copy := *f
-	copy.path = value
-	if err := copy.isValid(); err != nil {
+	clone := *f
+	clone.path = value
+	if err := clone.isValid(); err != nil {
 		return err
 	}
-	*f = copy
+	*f = clone
 	return nil
 }
 
--- a/cmd/skaffold/app/flags/inputfilepath.go
+++ b/cmd/skaffold/app/flags/inputfilepath.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package flags
+
+import (
+	"fmt"
+)
+
+// InputFilepath represents a input file command line argument
+// InputFilepath type makes sure the file exists.
+type InputFilepath struct {
+	filepathFlag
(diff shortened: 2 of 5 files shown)


cmd/skaffold/app/flags/filepath.go
```diff
@@ -0,0 +1,54 @@
...
+	"os"
+)
+
+type filepathFlag struct {
+	path        string
+	shouldExist bool
+}
+
+func (f *filepathFlag) SetIfValid(value string) error {
+	copy := f
```
dgageot: This copies the reference, not the value. You should use `copy := *f`

cmd/skaffold/app/flags/filepath.go
```diff
@@ -0,0 +1,54 @@
...
+	copy.path = value
+	if err := copy.isValid(); err != nil {
+		return err
+	}
+	f = copy
+	return nil
+}
+
+func newFilepathFlag(value string, exists bool) filepathFlag {
```
dgageot: Is this constructor useful?

cmd/skaffold/app/flags/filepath.go
tejal29: umm let me see if the tests caught it or write tests.

tejal29: so, so i found the bug and these tests did not catch it. will fix.

cmd/skaffold/app/flags/filepath.go
tejal29: yeah its not used right now. let me delete.

cmd/skaffold/app/flags/filepath.go
```diff
@@ -0,0 +1,47 @@
...
+	"os"
+)
+
+type filepathFlag struct {
+	path        string
+	shouldExist bool
+}
+
+func (f *filepathFlag) SetIfValid(value string) error {
+	copy := *f
```
corneliusweig: `copy` collides with a builtin. Maybe rename to `clone`?

cmd/skaffold/app/flags/inputfilepath.go
```diff
@@ -0,0 +1,51 @@
...
+func NewInputFilepath(value string) *InputFilepath {
+	return &InputFilepath{
+		filepathFlag{
+			path:        value,
+			shouldExist: true,
+		}}
+}
+
+func (f *InputFilepath) Usage() string {
+	return "Path to an input file."
```
corneliusweig: I think a non-configurable usage string for a flag is useless. Can you make it configurable?

cmd/skaffold/app/flags/outputfilepath.go
```diff
@@ -0,0 +1,49 @@
...
+		path: value,
+	}}
+}
+
+func (f *OutputFilepath) Type() string {
+	return fmt.Sprintf("%T", f)
+}
+
+func (f *OutputFilepath) Usage() string {
+	return "Path to an output filepath."
```
corneliusweig: Should be configurable, too.

cmd/skaffold/app/flags/inputfilepath.go
tejal29: makes sense. I will change this to default usage and then add another method `Usage`

cmd/skaffold/app/flags/filepath.go
tejal29: thanks! Will do

cmd/skaffold/app/flags/inputfilepath.go
```diff
@@ -0,0 +1,54 @@
+/*
```
balopat: nit: `input_filepath.go`

cmd/skaffold/app/flags/outputfilepath.go
```diff
@@ -0,0 +1,53 @@
+/*
```
balopat: nit: `output_filepath.go`

cmd/skaffold/app/flags/filepath.go
```diff
@@ -0,0 +1,47 @@
...
+	if err := clone.isValid(); err != nil {
+		return err
+	}
+	*f = clone
+	return nil
+}
+
+func (f filepathFlag) isValid() error {
+	if !f.shouldExist {
+		// Currently no validation implemented for output files.
```
balopat: why are we talking here about output files? Isn't this a shared/abstracted struct?

cmd/skaffold/app/flags/filepath.go
```diff
@@ -0,0 +1,47 @@
...
+limitations under the License.
+*/
+
+package flags
+
+import (
+	"os"
+)
+
+type filepathFlag struct {
```
balopat: comments would be nice to describe the struct

tejal29:
> I'm a little bit worried introducing these abstractions without the context they will be used in - without actually tying it to a feature.

I added a TODO to follow up with a PR for replacing all flags, but if it helps  i can add them in this same PR.

> the usage is a bit clunky: the constructors are not validating - so I have to create an Inputfilepath or

 Outfilepath first and then ask if it is "IsValid()" - or even worse, I have to create it with empty values and then try to "Set()" and check for errors.
e.g usage:
build_command.go
```
var (
readfile = flags.NewInputFilepath("", "Input file for the build command")
)
cmd := &cobra.Command{
		Use:   "build",
		Short: "Builds the artifacts",
		Args:  cobra.NoArgs,
			opts.Command = "build"
			return runBuild(out)
		},
	}

    cmd.Flags().VarP(readfile, "readfile", "r", readfile.Usage())
```

That is correct. The way way flags works, you can create a flag with default value. Only `Set` does the validation. See https://github.com/spf13/pflag/blob/master/flag.go#L187

> usage appears in both input and output: I think there is no real Inputfilepath and Outputfilepath flag just yet
the shouldExist seems a bit limiting as well: what if you want to append? What if you want to express shouldNotExist?

The shouldExist is to distinguish between InputfilePath and OutputfilePath.
An input file path should always exist. But if its not clear for now, i can remove OutputfilePath and add it later if needed. So just have filepath for now.

> I would recommend folding these changes into the context of the development you thought about them, but maybe just start small, identify the repetitions and triangulate the design from there. It looks like to me that maybe just a single filepathFlag struct with a func based validation for existence / non-existence, etc. could be enough? What do you think?

I can add some examples in the follow up code changes that will make the usage clear.

tejal29:
@balopat  i am going to hold on this change. I just realized the skaffold config file has a lot more validation than just file exists.
1. A File could be a local file and needs to exists.
2. A file could be a remote url and should be valid url
3. A file could be   "-" in which case we  read from std in.

balopat:
Sounds good!

PR #1874, opened 2019-03-24
Fix go routine leak

dgageot:


balopat:
haha, I missed this one - should have checked I guess :)

Issue #1380, opened 2018-12-10
Add parallel local image building.


tmc:
#248 added parallel building to gcb+kaniko, but not local builds.

### Expected behavior
That local builds use InParallel

### Actual behavior
Local builds use InSequence

balopat:
I would love to have this somehow. 

Thoughts: 

This is only for `local` builds. 
On `cluster` and `gcb` builds we have parallel builds. 

1. We experimented with this before for local builds but things were breaking back then . @dgageot maybe it's worth having another look now
2. There _are_ people who depend on sequenced builds - they use it as a poor man's dependency management solution.

PR #1895, opened 2019-03-29
Configure Jib builds to use plain progress updates

briandealwis:
Fixes GoogleContainerTools/jib#1430

This PR changes Skaffold's Jib binding to invoke Maven and Gradle with `-Djib.console=plain` to cause Jib to use its impoverished plain progress updater.

Jib provides a _rich_ progress updater that uses ANSI cursor sequences to provide a more GUI-like  build monitor.  Although we attempt to log these sequence strings as a single output, there can be interference if there is simultaneous output.  This rich progress updater is automatically disabled on Jib-Maven when invoked from Skaffold, and this patch explicitly disables Jib's rich progress updater on Jib-Gradle too.

The use of `jib.console` is a compromise.  Although Maven and Gradle provide options to run in a batch mode, doing so disables all colour output which is actually very useful. These batch modes can be enabled from the `args` field in the `skaffold.yaml`, and are automatically set when console output is redirected to a log file.

### DIFF
--- a/pkg/skaffold/jib/jib_gradle.go
+++ b/pkg/skaffold/jib/jib_gradle.go
@@ -42,13 +42,13 @@ func GetDependenciesGradle(ctx context.Context, workspace string, a *latest.JibG
 }
 
 func getCommandGradle(ctx context.Context, workspace string, a *latest.JibGradleArtifact) *exec.Cmd {
-	args := []string{gradleCommand(a, "_jibSkaffoldFiles"), "-q"}
+	args := append(gradleCommand(a, "_jibSkaffoldFiles"), "-q")
 	return GradleCommand.CreateCommand(ctx, workspace, args)
 }
 
 // GenerateGradleArgs generates the arguments to Gradle for building the project as an image.
 func GenerateGradleArgs(task string, imageName string, a *latest.JibGradleArtifact, skipTests bool) []string {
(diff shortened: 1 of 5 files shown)


dgageot:
For the record, Skaffold will never interleave build outputs. 
 + Local builds are always in sequence
 + Remote builds are run in parallel but their output is displayed in sequence.

briandealwis:
I was going on this TODO and #1380.
https://github.com/GoogleContainerTools/skaffold/blob/6ea8c06a03339aa92dab4272fe7ce94eb3c6b0f5/pkg/skaffold/build/local/local.go#L45-L46

Aside: If Skaffold ever does allow parallel local builds, it would be good to let the builders provide an indication of their support: I'm not sure if Maven and Gradle can really do parallel builds within the same workspace such as for multi-module type builds — I think there may be interference if there are shared project dependencies that needs to be rebuilt.

At the very least, this patch brings consistency between the two jib builders.  I'm not sure how Skaffold drains the builder logs, but there is a possibility that the Maven/Gradle side may buffer output.

briandealwis:
I didn't quite finish that last thought:  _ I'm not sure how Skaffold drains the builder logs, but there is a possibility that the Maven/Gradle side may buffer output_ **and so the log string may only be partially output in multiple parts rather than as a single output.**

chanseokoh:
@dgageot this PR is self-contained and good to go anyway.

PR #1897, opened 2019-03-30
Fail fast when dockerfile is outside of artifact context

corneliusweig:
GetDependencies must return relative paths wrt the artifact context. This was not always the case for the dockerfile. Before this PR, dockerfiles outside the context were accepted, but Skaffold would fail lateron with a hard to understand error message.

Note that returning an absolute dockerfile path was introduced in b572d4b2602305b1611dbaeebdcef563cdba4aad. However even back then this could not work, because pkg/skaffold/docker/context.go prefixes every path again.

Relates to #1893

### DIFF
--- a/pkg/skaffold/docker/parse.go
+++ b/pkg/skaffold/docker/parse.go
@@ -243,6 +243,7 @@ func expandPaths(workspace string, copied [][]string) ([]string, error) {
 }
 
 // NormalizeDockerfilePath returns the absolute path to the dockerfile.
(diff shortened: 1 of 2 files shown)


corneliusweig:
After second thoughts, I don't think that failing fast is such a good idea after all. `GetDependencies` only should fail, when the builder is the go-library builder or kaniko. If it is docker-cli, everything is fine, wherever the Dockerfile may be located (also see #1893). This PR would rule out that possibility.

PR #1906, opened 2019-04-02
[Proposal] Generalize configuration of kaniko pod

venkatk-25:


balopat:
hey @venkatk-25 - can you add me as design shepherd? I'll start to have a look tomorrow and give feedback!

venkatk-25:
@balopat updated

docs/design_proposals/generalizing_kaniko_config.md
```diff
@@ -0,0 +1,137 @@
...
+```
+new scheme:
+``` yaml
+build:
+  cluster:
+    secrets: 
+    - secretName: e2esecret
+      mountPath:  "/secret"
+    env:
+       "GOOGLE_APPLICATION_CREDENTIALS" : "/secret/kaniko-secret"
```
balopat: this still feels like we could generate it - just because the user can make an error easily here - for example this example itself has inconsistency, the env should be: ` "GOOGLE_APPLICATION_CREDENTIALS" : "/secret/e2esecret"`

docs/design_proposals/generalizing_kaniko_config.md
```diff
@@ -0,0 +1,137 @@
...
+
+	DockerConfig   *DockerConfig `yaml:"dockerConfig,omitempty"`
+}
+
+// DockerConfig contains information about the docker `config.json` to mount.
+type DockerConfig struct {
+	Path       string `yaml:"path,omitempty"`
+	SecretName string `yaml:"secretName,omitempty"`
+}
+```
```
tejal29: Can you add a section of related bugs in Background ?
- https://github.com/GoogleContainerTools/skaffold/issues/731

venkatk-25: Added the section for this

docs/design_proposals/generalizing_kaniko_config.md
venkatk-25: I wasn't sure of this when I wrote it. From code this looked like a constant regardless of secret name. https://github.com/GoogleContainerTools/skaffold/blob/533378cad3a92c391dca9d1656c2c30967c2c5cb/pkg/skaffold/build/kaniko/sources/sources.go#L69-L70

venkatk-25: > * I would still think about generating/helping the user with the env var somehow for GCR.

Regardless of above comment, I agree this is an additional and error prone step for GCR or maybe other secrets. I can think of following way to solve this:

. Have one more optional field called type for secret which can be an enum of special secrets, so we can do some extra work on them when specified.

docs/design_proposals/generalizing_kaniko_config.md
```diff
@@ -0,0 +1,140 @@
...
+Please list any open questions here in the format.
+
+**\<Question\>**
+
+Resolution: Please list the resolution if resolved during the design process or
+specify __Not Yet Resolved__
+
+## Implementation plan
+1. We will implement the new configurations first, so ECR can be supported
+2. Write skaffold fix for converting GCR/Docker secrets and give dprecation warnings for same
```
balopat:
```suggestion
2. Write skaffold fix for converting GCR/Docker secrets and give deprecation warnings for same
```

docs/design_proposals/generalizing_kaniko_config.md
```diff
@@ -0,0 +1,140 @@
...
+    dockerConfig:
+      secretName: docker-secret
+```
+new scheme:
+``` yaml
+build:
+  cluster:
+    secrets: 
+    - secretName: docker-secret
+      mountPath:  "/kaniko/.docker"
```
balopat:
```suggestion
      mountPath:  "/kaniko/.docker/"
```

balopat:
Friendly ping @venkatk-25

corneliusweig:
@venkatk-25 can you also add a reminder to update the docs at `docs/content/en/docs/how-tos/builders/_index.md`?

venkatk-25:
@balopat sorry was travelling and couldn't respond earlier. 

I suggest the following changes instead to be more consistent by introducing type field to the secret.
User would need to provide one of mountPath or type
```go
type SecretConfig struct {
	LocalPath  string `yaml:"localPath"`
	SecretName string `yaml:"secretName,omitempty"`
	MountPath  string `yaml:"mountPath,omitempty"`
        Type: string `yaml:"type,omitempty"`
}
```

and yaml will look like.
```yaml
build:
  cluster:
    secrets: 
    - secretName: e2esecret
      type: "GCR"
    - secretName: e2esecret
      mountPath:  "/secret"
    env:
       "GOOGLE_APPLICATION_CREDENTIALS" : "/secret/e2esecret"
```

venkatk-25:
> @venkatk-25 can you also add a reminder to update the docs at `docs/content/en/docs/how-tos/builders/_index.md`?

@corneliusweig I did not understand this. Is this related to this design or you are asking me to work on docs for existing features

corneliusweig:
@venkatk-25 Sorry for being so unclear about this. What I mean is to add a reminder in the implementation plan that the docs at this location need to be updated. Doc updates are so easily forgotten...

balopat:
> @balopat sorry was travelling and couldn't respond earlier.

No worries, I've been super busy too :)  

> I suggest the following changes instead to be more consistent by introducing type field to the secret.
> User would need to provide one of mountPath or type( enum of GCR, ECR, Docker, ... )

1. I like the idea to standardize on `secretName` - maybe we can just call it `name` as we are under `secrets` anyway? 

2. You still mentioned in your example `GOOGLE_APPLICATION_CREDENTIALS` env var - that shouldn't need to be there if there is a GCR type secret! We should mention this in the doc. Also - as I'm thinking about this - we should mention that with the `cluster` details are extracted we are working with the key assumption that a single GCR/ECR type secret covers most use cases (i.e. users typically don't want to push to two different GCR registries from a single multi-artifact skaffold project). 

3. `type` field could work, but I'd prefer to do something similar to how we handle artifact types in the `build` section. That enables to give more structure to these options if we need to - and while I was thinking about this I realized one more issue: 

4.  The `path` in dockerConfig is not a mountPath! It is the path for the local json that skaffold can create the secret from on the cluster for the duration of the build! We have to cater for this too. Or handle it separately.

tejal29:
@venkatk-25 friendly ping

balopat:
Hi @venkatk-25, I'm going to close this now as it is a bit stalled. Please resubmit if you want to keep working on it! Thank you very much!

PR #1910, opened 2019-04-02
Refactor pod selection logic for log-tailing

corneliusweig:
**Note**: Part of design proposal #1911

So far, pods were selected based on known image names in their container spec. This is restricting, because it does not allow to add additional pods for log-tailing, nor is it feasible to exclude uninteresting pods. Besides, the book-keeping code for the deployed images was scattered in the code.

Now, pods are selected for log-tailing by looking for a `tail=true` label on the pods. This is mainly a preparatory refactoring to implement more interesting label selectors in the future.

References #666, #588
Depends on #2074

### DIFF
--- a/pkg/skaffold/kubernetes/colorpicker_test.go
+++ b/pkg/skaffold/kubernetes/colorpicker_test.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestColorPicker(t *testing.T) {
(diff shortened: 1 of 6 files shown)


pkg/skaffold/kubernetes/colorpicker_test.go
```diff
@@ -30,46 +31,43 @@ func TestColorPicker(t *testing.T) {
...
 			},
 			expectedColor: colorCodes[1],
 		},
 	}
 
-	picker := NewColorPicker([]string{"image", "second"})
+	picker := NewColorPicker()
+	// register "second" twice and still expect colorCodes[1]
+	for _, v := range []string{"pod", "second", "second"} {
+		picker.Register(&v1.Pod{
```
corneliusweig: `Register` and `Pick` now involve locking. However, the current tests do not assert on concurrent use. Do we need that?

corneliusweig:
Helm does currently not label pods created via Deployment resources (or any pod from a podTemplateSpec for that matter). Therefore, this PR can only be merged once #2074 is resolved.

dgageot:
@corneliusweig Is this PR still active? I'd like to close stalled PRs so that we have a better idea of which PR needs a review

corneliusweig:
This one is pending on #1911 and #2074 . I'll close for now and reopen if these other ones are solved.

PR #1923, opened 2019-04-04
Minor doc updates

chanseokoh:


chanseokoh:
Test is broken due to #1924.

balopat:
Thanks for asking the right question @nkubala, I was too fast to approve.
Actually it is not true either - our release and integration tests are also based on go 1.11 - so we should change that first `deploy/skaffold/Dockerfile`

chanseokoh:
Perhaps my go installation had some problem, which made me to think upgrading to 1.12 resolved the issue. I reverted the 1.1.2 requirement.

chanseokoh:
I merged with master, and tests should pass.

Issue #1814, opened 2019-03-16
Sync does not work in containers without the tar command


corneliusweig:
### Expected behavior
Changed files are synced to the container. The example below does not make much sense, but it does also affect distroless containers.

### Actual behavior
Sync fails with
```
DEBU[0007] Running command: [kubectl exec getting-started --namespace default -c getting-started -i -- tar xmf - -C / --no-same-owner] 
WARN[0007] Skipping deploy due to sync error: copying files: exit status 126
```

### Information

- Skaffold version: >0.24

### Steps to reproduce the behavior
Based on the examples/getting-started example
1. Modify Dockerfile to
```
FROM golang:1.10.1-alpine3.7 as builder
COPY main.go .
RUN go build -o /app main.go

FROM scratch
ENTRYPOINT ["./app"]
COPY . /unused
COPY --from=builder /app .
```
2. Modify skaffold.yaml
```
apiVersion: skaffold/v1beta7
kind: Config
build:
  artifacts:
  - image: gcr.io/k8s-skaffold/skaffold-example
    sync:
    - '*.go': /
deploy:
  kubectl:
    manifests:
      - k8s-*
```
3. `skaffold dev -v debug`
4. Change `main.go` and wait for sync.

corneliusweig:
A way out is obviously to offer a fallback from the tar-optimized copy to a single-file copy. I would like to fix this, and I can think of two options:
1. try an opportunistic tar-sync. If that fails, change the syncer from to single-file for the rest of this skaffold session. This has the advantage of not adding yet another configuration option which users don't really want to bother about. On the other hand, if the sync fails for a different reason, it will be stuck on the slow syncer.
2. add a new configuration to switch between tar-based and single-file syncer. In that case, what is a good name? Has the drawback that users need to be aware of this option and makes the configuration even bigger.

What do you think?

corneliusweig:
So I just tried the fallback version but it will not work, because `kubectl cp` also builds a tar :roll_eyes: 
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/cp/cp.go#L245

I also had a look at the distroless images from https://github.com/GoogleContainerTools/distroless/ and it looks like they do not package any executable like `cat` or `tar` (that's what they are made for, after all). So that rules out `kubectl cp` for distroless and scratch.

[...]

PR #1926, opened 2019-04-05
Fix syncing for Jib

### LINKED ISSUE #1921: Sync does not work for Skaffold-Jib
For Maven and Gradle, `src/main/resources` is a standard location for placing static files.  But attempting to enable sync files from `src/main/resources` fails with Skaffold+Jib. 
```diff
+    sync:
+      'src/main/resources/**': '/app/resources'
```

### Expected behavior

The `README.md` file should be sync'd.

### Actual behavior

I have `~/go/src/github.com/GoogleContainerTools/skaffold` linked to `~/Projects/GPE/repo-skaffold`, and normally do my builds from `~/Projects/GPE/repo-skaffold`.

With `skaffold dev`, a change to a file with a sync rule will fail with:
```
FATA[0094] calling final callback: sync: intersecting sync map and added, modified files: changed file /Users/bsd/go/src/github.com/GoogleContainerTools/skaffold/examples/jib/src/main/resources can't be found relative to context .: Rel: can't make /Users/bsd/go/src/github.com/GoogleContainerTools/skaffold/examples/jib/src/main/resources relative to . 
```

### Information

The Jib builder's `GetDependencies` works by invoking an internal Jib build goal (`_skaffold-files` on Maven, `_jibSkaffoldFiles` on Gradle) that returns the build files and directories to source files.  The Jib Builder then expands the source directories to return the contents.

In digging into the problem, there are two issues:
  1. When expanding the paths returned by Jib's `skaffold-files`, `getDependencies()` returns both files and directories.  In the example below, the binding returns the directory `src/main/resources` and its contents `src/main/resources/README`.  But the directory doesn't match a sync-rule and so the attempt to sync fails.  
  1. The Jib's `skaffold-files` goal returns absolute canonical paths (where symlinks are expanded), which can be a problem when the context is actually in a symlinked directory (either itself or the parent) as neither `filepath.Abs()` nor `filepath.Rel()` expand symlinks.  So the attempt [in `sync.go`'s `intersect()` to relativize the paths fails](https://github.com/GoogleContainerTools/skaffold/blob/b1dce640bd28bbf80f4097c5eb04ae3531f11628/pkg/skaffold/sync/sync.go#L142).

1. `mkdir examples/jib/src/main/resources`
2. `cp README.md examples/jib/src/main/resources/`
3. Apply the following diff to add a sync rule.
4. `cd examples/jib; skaffold dev` and then make a change to `src/main/resources/README.md`

```diff
--- a/examples/jib/skaffold.yaml
+++ b/examples/jib/skaffold.yaml
@@ -4,6 +4,8 @@ build:
   artifacts:
   - image: gcr.io/k8s-skaffold/skaffold-jib
     jibMaven: {}
+    sync:
+      'src/main/resources/**': '/app/resources'
 
 # optional profile to run the jib build on Google Cloud Build
 profiles:
```

### Proposed Fix

Fixing Problem 1 is straightforward: fix Jib's [`getDependencies()`](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/jib/jib.go#L70) to only return files.

I can fix Problem 2 by processing changing Jib's `getDependencies()` to apply `filepath.EvalSymlinks()` to the context directory prior to relativizing the files.  But the code in `sync.go`'s `intersect()` looks like it's trying to handle absolute files anyways.  Perhaps I should push this expansion here?

https://github.com/GoogleContainerTools/skaffold/blob/b1dce640bd28bbf80f4097c5eb04ae3531f11628/pkg/skaffold/sync/sync.go#L135-L142

briandealwis:
Syncing with a Jib project doesn't work (#1921):
  - jib.go's `getDependencies()` returns both files and directories.  A directory `a/b/c` doesn't match a sync-rule with a glob `a/b/c/**` and so changing file `a/b/c/README` causes sync to fail.
  - The Skaffold-Jib binding calls out to Maven-Jib and Gradle-Jib to return paths files and directories.  These are normally _canonical paths_ where symlinks are expanded, and so the paths may not match the specified workspace which is usually relative to the `skaffold.yaml` like ".".

This PR fixes Jib's `getDependencies()` to relativize paths to the given workspace or the canonicalized workspace.  It also filters out directories.

Fixes #1921

### DIFF
--- a/pkg/skaffold/jib/jib.go
+++ b/pkg/skaffold/jib/jib.go
@@ -17,9 +17,12 @@ limitations under the License.
 package jib
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"sort"
+	"strings"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 	"github.com/karrick/godirwalk"
@@ -27,10 +30,19 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func getDependencies(cmd *exec.Cmd) ([]string, error) {
+func getDependencies(workspace string, cmd *exec.Cmd) ([]string, error) {
 	stdout, err := util.RunCmdOut(cmd)
 	if err != nil {
 		return nil, err
+	}
+
+	// Jib's dependencies are absolute, and usually canonicalized, so must canonicalize the workspace
+	if workspace, err = filepath.Abs(workspace); err != nil {
+		return nil, errors.Wrapf(err, "unable to resolve workspace %s", workspace)
+	}
+	canonicalWorkspace, err := filepath.EvalSymlinks(workspace)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to canonicalize workspace %s", workspace)
 	}
 
 	cmdDirInfo, err := os.Stat(cmd.Dir)
@@ -60,6 +72,10 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
 		}
 
 		if !info.IsDir() {
+			// try to relativize the path
+			if relative, err := relativize(dep, workspace, canonicalWorkspace); err == nil {
+				dep = relative
+			}
 			deps = append(deps, dep)
 			continue
 		}
@@ -67,7 +83,13 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
 		if err = godirwalk.Walk(dep, &godirwalk.Options{
 			Unsorted: true,
 			Callback: func(path string, _ *godirwalk.Dirent) error {
-				deps = append(deps, path)
+				if info, err := os.Stat(path); err == nil && !info.IsDir() {
+					// try to relativize the path
+					if relative, err := relativize(path, workspace, canonicalWorkspace); err == nil {
+						path = relative
+					}
+					deps = append(deps, path)
+				}
 				return nil
 			},
 		}); err != nil {
@@ -77,4 +99,19 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
 
 	sort.Strings(deps)
 	return deps, nil
+}
+
+// relativize tries to make path relative to the root location(s)
+func relativize(path string, roots ...string) (string, error) {
+	if !filepath.IsAbs(path) {
+		return path, nil
+	}
+	dotDotSlash := fmt.Sprintf("..%c", filepath.Separator)
+	for _, root := range roots {
+		// check that the path can be made relative and is contained (since `filepath.Rel("/a", "/b") => "../b"`)
+		if rel, err := filepath.Rel(root, path); err == nil && !strings.HasPrefix(rel, dotDotSlash) {
+			return rel, nil
+		}
+	}
+	return "", errors.New("could not relativize path")
 }
--- a/pkg/skaffold/jib/jib.go
+++ b/pkg/skaffold/jib/jib.go
@@ -39,7 +39,10 @@ func getDependencies(workspace string, cmd *exec.Cmd) ([]string, error) {
 		return nil, err
 	}
 
-	// Jib's dependencies are absolute, and usually canonicalized, so must canonicalize the workspace
+	// Skaffold prefers to deal with relative paths.  In *practice*, Jib's dependencies
+	// are *usually* absolute (relative to the root) and canonical (with all symlinks expanded).
+	// But that's not guaranteed, so we try to relativize paths against the workspace as
+	// both an absolute path and as a canonicalized workspace.
 	workspaceRoots, err := calculateRoots(workspace)
 	if err != nil {
 		return nil, errors.Wrapf(err, "unable to resolve workspace %s", workspace)
@@ -72,7 +75,8 @@ func getDependencies(workspace string, cmd *exec.Cmd) ([]string, error) {
 		}
 
 		if !info.IsDir() {
-			// try to relativize the path
+			// try to relativize the path: an error indicates that the file cannot
+			// be made relative to the roots, and so we just use the full path
 			if relative, err := relativize(dep, workspaceRoots...); err == nil {
 				dep = relative
 			}
@@ -84,7 +88,8 @@ func getDependencies(workspace string, cmd *exec.Cmd) ([]string, error) {
 			Unsorted: true,
 			Callback: func(path string, _ *godirwalk.Dirent) error {
 				if info, err := os.Stat(path); err == nil && !info.IsDir() {
-					// try to relativize the path
+					// try to relativize the path: an error indicates that the file cannot
+					// be made relative to the roots, and so we just use the full path
 					if relative, err := relativize(path, workspaceRoots...); err == nil {
 						path = relative
 					}
--- a/pkg/skaffold/jib/jib.go
+++ b/pkg/skaffold/jib/jib.go
@@ -17,7 +17,6 @@ limitations under the License.
 package jib
 
 import (
-	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -30,6 +29,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	dotDotSlash = ".." + string(filepath.Separator)
+)
+
 func getDependencies(workspace string, cmd *exec.Cmd) ([]string, error) {
 	stdout, err := util.RunCmdOut(cmd)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


briandealwis:
I should note that sync with Jib requires building against something like `openjdk` as sync relies on the presence of `tar`, which doesn't exist in the `gcr.io/distroless/java` (Jib's default image).

briandealwis:
🤦‍♂️ WIndows paths

pkg/skaffold/jib/jib.go
```diff
@@ -78,3 +100,18 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
 	sort.Strings(deps)
 	return deps, nil
 }
+
+// relativize tries to make path relative to the root location(s)
```
corneliusweig:
```suggestion
// relativize tries to make path relative below one of the root locations
```

pkg/skaffold/jib/jib.go
```diff
@@ -17,22 +17,34 @@ limitations under the License.
...
 )
 
-func getDependencies(cmd *exec.Cmd) ([]string, error) {
+func getDependencies(workspace string, cmd *exec.Cmd) ([]string, error) {
 	stdout, err := util.RunCmdOut(cmd)
 	if err != nil {
 		return nil, err
 	}
 
+	// Jib's dependencies are absolute, and usually canonicalized, so must canonicalize the workspace
```
corneliusweig: I take it that "usually" means that they are not always canonicalized? Or do you simply want to cover that possibility `relativize`?

pkg/skaffold/jib/jib.go
```diff
@@ -60,14 +72,24 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
 		}
 
 		if !info.IsDir() {
+			// try to relativize the path
+			if relative, err := relativize(dep, workspace, canonicalWorkspace); err == nil {
```
corneliusweig: I wonder if `dep` is most of the time canonical here. If so, you could swap `workspace` and `canonicalWorkspace` so that the good case comes first.

pkg/skaffold/jib/jib.go
```diff
@@ -78,3 +100,18 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
 	sort.Strings(deps)
 	return deps, nil
 }
+
+// relativize tries to make path relative to the root location(s)
+func relativize(path string, roots ...string) (string, error) {
+	if !filepath.IsAbs(path) {
+		return path, nil
+	}
+	dotDotSlash := fmt.Sprintf("..%c", filepath.Separator)
```
corneliusweig: I wonder if `dotDotSlash` can be put in a const variable?

corneliusweig:
> I should note that sync with Jib requires building against something like `openjdk` as sync relies on the presence of `tar`, which doesn't exist in the `gcr.io/distroless/java` (Jib's default image).

@briandealwis Do you think there's a chance to get a distroless variant with tar? Currently, this affects _all_ copy operations into the k8s container (also see https://github.com/GoogleContainerTools/skaffold/issues/1814#issuecomment-473654285)

briandealwis:
Distroless doesn't include a shell, so even if distroless included tar, any use would have to explicitly specify a full path to the executable.

pkg/skaffold/jib/jib.go
briandealwis: I had hoped so too but the compiler complained

pkg/skaffold/jib/jib.go
```diff
@@ -60,14 +72,24 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
...
 			continue
 		}
 
 		if err = godirwalk.Walk(dep, &godirwalk.Options{
 			Unsorted: true,
 			Callback: func(path string, _ *godirwalk.Dirent) error {
-				deps = append(deps, path)
+				if info, err := os.Stat(path); err == nil && !info.IsDir() {
+					// try to relativize the path
+					if relative, err := relativize(path, workspace, canonicalWorkspace); err == nil {
```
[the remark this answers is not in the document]
briandealwis: It's still valid to return an absolute path: Skaffold-Jib also returns references to SNAPSHOT jars found in the user's local repository (e.g., under ~/.m2/repository).

pkg/skaffold/jib/jib.go
briandealwis: To cover that possibility: the Jib `skaffold-files` goal simply reports builder-provided information.  So it's best not to assume anything, really.

pkg/skaffold/jib/jib.go
briandealwis: Good idea.

pkg/skaffold/jib/jib.go
briandealwis: Aha: `const dotDotSlash = ".." + string(filepath.Separator)`

briandealwis:
Huh, `filepath.IsAbs("\a")` is false on Windows.

corneliusweig:
> Huh, `filepath.IsAbs("\a")` is false on Windows.

:scream:

briandealwis:
PTAL @balopat

pkg/skaffold/jib/jib.go
```diff
@@ -60,14 +72,24 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
 		}
 
 		if !info.IsDir() {
+			// try to relativize the path
+			if relative, err := relativize(dep, workspaceRoots...); err == nil {
```
balopat: we should handle the relativization errors here

pkg/skaffold/jib/jib.go
balopat: Frankly, I don't understand this comment yet either :) What do we exactly mean by "canonicalization"? Everything relative? Everything absolute? Symlinks resolved to absolute? Not sure I'm following.

pkg/skaffold/jib/jib.go
```diff
@@ -78,3 +100,33 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
...
 	return deps, nil
 }
+
+// calculateRoots returns a list of possible symlink-expanded paths
+func calculateRoots(path string) ([]string, error) {
+	path, err := filepath.Abs(path)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to resolve %s", path)
+	}
+	canonical, err := filepath.EvalSymlinks(path)
```
balopat: based on my understanding reading `EvalSymlinks`'s godoc, it will leave relative paths as relative. This should be fine right?

pkg/skaffold/jib/jib.go
```diff
@@ -78,3 +100,33 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to canonicalize workspace %s", path)
+	}
+	if path == canonical {
+		return []string{path}, nil
+	}
+	return []string{canonical, path}, nil
```
balopat: Why are we returning both the symlink expanded and the non expanded version?

pkg/skaffold/jib/jib.go
```diff
@@ -78,3 +100,33 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
...
+}
+
+// relativize tries to make path relative to one of the given roots
+func relativize(path string, roots ...string) (string, error) {
+	if !filepath.IsAbs(path) {
+		return path, nil
+	}
+	for _, root := range roots {
+		// check that the path can be made relative and is contained (since `filepath.Rel("/a", "/b") => "../b"`)
+		if rel, err := filepath.Rel(root, path); err == nil && !strings.HasPrefix(rel, dotDotSlash) {
```
balopat: I think we should handle the errors here explicitly as well - it should be easier to debug errors.

balopat: Maybe concatenate them, if there are more. There are max two roots as I'm reading this...

pkg/skaffold/jib/jib.go
```diff
@@ -78,3 +100,33 @@ func getDependencies(cmd *exec.Cmd) ([]string, error) {
...
+	}
+	if path == canonical {
+		return []string{path}, nil
+	}
+	return []string{canonical, path}, nil
+}
+
+// relativize tries to make path relative to one of the given roots
+func relativize(path string, roots ...string) (string, error) {
+	if !filepath.IsAbs(path) {
```
balopat: Why are we assuming here that if the path is relative, it is relative to the right root? Can we assume that?

balopat: Why are we assuming here that if the path is relative, it is relative to the right root? Can we assume that?

balopat: Why are we assuming here that if the path is relative, it is relative to the right root? Can we assume that?

balopat: Why are we assuming here that if the path is relative, it is relative to the right root? Can we assume that?

pkg/skaffold/jib/jib.go
briandealwis: This is @corneliusweig's [question](https://github.com/GoogleContainerTools/skaffold/pull/1926/files/4d799283170dde3fa008587382ae7a1a43a291fd#r272745633) too: basically Jib's `skaffold-files` goal just returns information provided from the Maven/Gradle.  These files *seem* to be absolute and canonical, but the docs don't actually guarantee that.  Expanding symlinks for *every* file seems like it could be expensive, so I instead expand the workspace so we can try comparing the expanded and unexpanded workspace paths as prefixes.

pkg/skaffold/jib/jib.go
briandealwis: These aren't real errors: it just means that the file could not be made relative, which usually means that `path` is already relative.  (Can also happen on Windows if they're on different volumes.)

briandealwis:
> I'll take that on

I meant separately.

corneliusweig:
> > I'll take that on
> 
> I meant separately.

Keep in mind that the sync config will be changed soon.

PR #1939, opened 2019-04-09
Full validation for `required` and `oneOf` config fields

### LINKED ISSUE #1920: `yamltags.ProcessStruct` is insufficient
### Information
`yamltags.ProcessStruct` was intended to set defaults in the schema and validate tags such as `required` and `oneOf`. However, in its current shape, it only processes structs and child fields which are structs. This means that config fields of kind `slice` or `Ptr` are not processed. This means that validation of `yamltags=required` is currently broken.

This is also the reason why it appeared that Skaffold did not validate `oneOf` tags at all (#1836 & #1881). I tried to fix this and recurse into arrays&pointers as well, but it seems there is a fundamental problem here. The reason why it cannot work is that `ProcessStruct` tries to set default values, however if we include slices here, we would try to set a value in a non-addressable value (you can play around with this [yamltags-patch.txt](https://github.com/GoogleContainerTools/skaffold/files/3043099/yamltags.txt)).

### Proposal
- I checked if `yamltags=default` is being used at all and I didn't find anything. My suggestion is therefore to remove `yamltags=default`, unless somebody comes up with an idea how to fix this. Thus, `yamltags.ProcessStruct` should only do validation.
- `yamltags.ProcessStruct` is a much cleaner code than `validation.ValidateSchema`. However, validation needs to be done after applying Skaffold config profiles (cmd/skaffold/app/cmd/runner.go:35). I therefore suggest to keep `validation.ValidateSchema` but use the validation implementation from `yamltags.ProcessStruct`. This will ensure that we also have a place to add new validation requirements which has nothing to do with `yamltags`. 

I'm happy to submit a PR for this, if you agree with that proposal.

corneliusweig:
As outlined in #1920, the current validation of the Skaffold config was quite inconsistent:
- validation for `yamltags=oneOf` was duplicated in packages `yamltags` and `validation`
- validation for `yamltags=required` was broken and not doing much
- `yamltags=default` had the same problem as `yamltags=required` but it cannot be fixed
- the original yamltags validation was done before applying profiles, iow too early

This PR fixes the abovementioned problems. Validation still remains in the `validation` package, so that there will be a good location to add further validation in the future. The implementation of `yamltags` validation is now taken from the `yamltags` package.

Close #1920

### DIFF
--- a/pkg/skaffold/schema/versions.go
+++ b/pkg/skaffold/schema/versions.go
@@ -19,11 +19,7 @@ package schema
 import (
 	"fmt"
 
-	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
-	yaml "gopkg.in/yaml.v2"
-
-	apiversion "github.com/GoogleContainerTools/skaffold/pkg/skaffold/apiversion"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/apiversion"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/v1alpha1"
@@ -39,7 +35,9 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/v1beta6"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/v1beta7"
 	misc "github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
-	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/yamltags"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
 )
 
 type APIVersion struct {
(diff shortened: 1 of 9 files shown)


pkg/skaffold/yamltags/tags_test.go
```diff
@@ -64,7 +64,7 @@ func TestProcessStructRequired(t *testing.T) {
 			wantErr: false,
 		},
 		{
-			name: "missng some",
+			name: "missing some",
```
tejal29: Can we add a test for `processTags` or `ValidateStructs` with an invalid tag i.e. not "oneOf", "required"

pkg/skaffold/schema/versions.go
```diff
@@ -102,10 +100,6 @@ func ParseConfig(filename string, upgrade bool) (util.VersionedConfig, error) {
 		return nil, errors.Wrap(err, "unable to parse config")
 	}
 
-	if err := yamltags.ProcessStruct(cfg); err != nil {
```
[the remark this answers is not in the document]
corneliusweig: I put some explanation in the commit messages, but here I can be more verbose :). The fundamental problem with validation at this point is that after profile application, the configuration may become invalid. For example, if somebody configures a docker build in the base and applies a profile which _adds_ another kaniko build config. Profile application merges the two, so that the `oneOf=build` is violated, because both kaniko and docker are set in the result.
Therefore, I think it is better to generally not do validation here.

pkg/skaffold/yamltags/tags_test.go
corneliusweig: Good idea! Done.

corneliusweig:
@tejal29 PTAL

PR #1986, opened 2019-04-18
Remove `PreBuiltImageBuilder` and make it a function instead.

tejal29:
For #922, i am making changes to `skaffold deploy` to  not run `skaffold build`.
`--images` flag is only used `deploy` flow.
The way it works is when `--images` flag is present, runner creates a
`PreBuiltImageBuilder`. All it does is, parses the images and converts
them to build.Artifact. If during deploy, an image in the list was also
built, then it uses the built image tag vs the tag present on the
commmand line flag.

In this change, we remove creating a PreBuiltImageBuilder in the runner.
(Remember this flag is only available for deploy and hence other command
won't get affected)
The `--images` will now be converted to build.Artifact in the
runner.Deploy flow.


tejal29:
@dgageot, Keeping `PreBuiltImageBuilder` adds unnecessary `build` flow into deploy which we need to get rid off.
Other way to fix this, introducing if by adding one more flow in the runner where if command is deploy and  `opts.PreBuiltImages`  is present, then run `build` or else only `deploy` which is add more complexity to the code.

cmd/skaffold/app/cmd/deploy.go
```diff
@@ -30,11 +33,33 @@ func NewCmdDeploy(out io.Writer) *cobra.Command {
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// Same actions as `skaffold run`, but with pre-built images.
-			return run(out)
+			opts.Command = "deploy"
```
dgageot: This line is not needed anymore. It's done in `rootCmd.PersistentPreRunE()`

cmd/skaffold/app/cmd/deploy.go
```diff
@@ -30,11 +33,33 @@ func NewCmdDeploy(out io.Writer) *cobra.Command {
...
 	}
 	AddRunDevFlags(cmd)
 	AddRunDeployFlags(cmd)
 	cmd.Flags().StringSliceVar(&opts.PreBuiltImages, "images", nil, "A list of pre-built images to deploy")
 	return cmd
 }
+
+func runDeploy(out io.Writer) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
dgageot: in run.go, we also have a `defer plugin.CleanupClients()` here. Should it be added?

cmd/skaffold/app/cmd/deploy.go
```diff
@@ -30,11 +33,33 @@ func NewCmdDeploy(out io.Writer) *cobra.Command {
...
+	defer runner.RPCServerShutdown()
+
+	// TODO: This should be fixed with https://github.com/GoogleContainerTools/skaffold/issues/922#issuecomment-474710200
+	// where buildArtifacts will consume `skaffold build` output
+	buildArtifacts := []build.Artifact{}
+	if err := runner.Deploy(ctx, out, buildArtifacts); err != nil {
+		return err
+	}
+
+	runner.TailLogs(ctx, out, nil, buildArtifacts)
```
dgageot: The error is not handled here.

Also, the `TailLogs` function was unused before this PR. `runner.Run()` uses its own code.
I think we should make sure both `Run` and `Deploy` use the same code.

pkg/skaffold/runner/runner.go
```diff
@@ -331,6 +327,15 @@ func (r *SkaffoldRunner) BuildAndTest(ctx context.Context, out io.Writer, artifa
 
 // Deploy deploys the given artifacts
 func (r *SkaffoldRunner) Deploy(ctx context.Context, out io.Writer, artifacts []build.Artifact) error {
+	if len(r.runCtx.Opts.PreBuiltImages) > 0 {
+		logrus.Debugln("Deploy using pre-built images")
+		pBuilts, err := convertImagesToArtifact(r.runCtx.Opts.PreBuiltImages)
+		if err != nil {
+			return errors.Wrap(err, "deploy failed")
+		}
+		// Merge current build artifacts with previously built
```
dgageot: I don't understand this comment. This portion of code is usually called in a dev loop to merge the artifacts built in the last loop and the artifacts built in the previous loops.

cmd/skaffold/app/cmd/deploy.go
tejal29: no it should not be added here since the plugins should not be called in deploy flow, But i can check again.

pkg/skaffold/runner/runner.go
tejal29: this is to merge the build artifacts, which is empty now. Something #2001, will add in.

cmd/skaffold/app/cmd/deploy.go
tejal29: i can look in to this.

cmd/skaffold/app/cmd/deploy.go
priyawadhwa: I’m in the process of removing the plugin code, so you don’t need to add it!

pkg/skaffold/runner/runner.go
balopat: What do you think about removing this until #2001 introduces it then?

tejal29: yes. Like i mentioned in the channel, i am going to remove the last commit from this branch. 
I pushed it for David to see the whole change. 
Please re-view this.

balopat:
> The idea behind curent implementation is to remove as many branches as possible in the Runner's functions. Too many branches make it too hard to maintain. I'd like to see the bug fixed without removing the `PreBuiltImageBuilder`

I agree that introducing an if statement and overriding the `artifacts` parameter in the `Deploy` method is not the nicest, it would be great to have the Build step be conditionally skipped in case we are using prebuilt images and pass in the artifacts to be deployed to the `Deploy` method.  

This conditional branching is implemented currently by the switch statement used in the Build step. Which is - I think one of those "clever" solutions that are a bit of a surprising  as 
- for `skaffold deploy` we are doing a build - and led to the bug of introducing a build step during a deploy-only operation. 
- we are "building prebuilt images"... which is just weird 

So I think we can do better than the current design in terms of non-surpising, more maintainable code. 

My suggestion would be to introduce the explicit conditional in the runner above the build step.

PR #2018, opened 2019-04-24
Make Deploy flags more relevant

tejal29:
When rewriting `skaffold deploy` i realized `skaffold deploy` has `--tail` option which is not required.
`skaffold run` uses `--tail`

Some of the `skaffold deploy` flags are not relevant like `--insecure-registries`, `--no-prune`.
Please let me know if this is incorrect.

Please see https://github.com/GoogleContainerTools/skaffold/compare/master...tejal29:make_sane_deploy_flags?expand=1#diff-2e110939e227b787655acb75b7eea64f to see the flag difference.

Flags removed:

|  Flag | Reason   |  
|---|---|
|   --cache-artifacts    |  Relevant to Build |   
| --cache-file  |  Same as above  |  
|--insecure-registry   | Relevant to Build, deploy does not pull or push.  |   
|    --no-prune  | deploy does not invoke delete flow. Relevant to dev, run |
|      --skip-tests  | deploy does not run tests |
|     --tail            |  I think this is more relevant to run and dev. Deploy should only deploy and return |

### DIFF
--- a/cmd/skaffold/app/cmd/cmd.go
+++ b/cmd/skaffold/app/cmd/cmd.go
@@ -156,6 +156,7 @@ func AddRunCommonFlags(cmd *cobra.Command) {
 }
 
 func AddRunDeployCommonFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&opts.Tail, "tail", false, "Stream logs from deployed objects")
 	cmd.Flags().BoolVar(&opts.Force, "force", false, "Recreate kubernetes resources if necessary for deployment (default: false, warning: might cause downtime!)")
 	cmd.Flags().StringArrayVarP(&opts.CustomLabels, "label", "l", nil, "Add custom labels to deployed objects. Set multiple times for multiple labels.")
 }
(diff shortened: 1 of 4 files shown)


tejal29:
Let me know if  `--tail` is something we want to keep for deploy.

balopat:
I think --tail still makes sense for `skaffold deploy` the rest LGTM that you removed.

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -155,12 +155,22 @@ func AddRunCommonFlags(cmd *cobra.Command) {
...
 }
 
-func AddRunDeployFlags(cmd *cobra.Command) {
-	cmd.Flags().BoolVar(&opts.Tail, "tail", false, "Stream logs from deployed objects")
+func AddRunDeployCommonFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&opts.Force, "force", false, "Recreate kubernetes resources if necessary for deployment (default: false, warning: might cause downtime!)")
 	cmd.Flags().StringArrayVarP(&opts.CustomLabels, "label", "l", nil, "Add custom labels to deployed objects. Set multiple times for multiple labels.")
 }
 
+func AddRunDeployOnlyFlags(cmd *cobra.Command) {
```
balopat: I'm starting to get really lost in this naming scheme. What does it mean that `RunDeployOnly` and how does it differ from `RunDeployCommon`? I'm starting to think that we should just add `commands=[]string{"run", "dev", "deploy"}` annotation to the flags  somehow...:)

tejal29: yes. its a good idea to add scope annotation to flags.  Its kind of confusing right now. Especially when they call each other.

PR #2022, opened 2019-04-25
Add Annotations to command and flags per phase annotation.

tejal29:
Inspired by @balopat feedback on #2018 , in this change Command have annotations with phases they run like "build", "test", "deploy".

Annotations Description:

| Annotation | Description |
|----------|-----|
| build | Runs a build |
| test | Runs tests |
| deploy | Performs Deploy |
| cleanup| flags related to clean up  |
| events| uses events api to send events|

With this new change, hopefully only relevant flags are added for commands.

Flags Removed or Added per Command

| Command | Flags removed| Flags Added| Note|
|------|-------|---|--|
| build | --no-prune, --toot| | |
| delete | --enable-rpc,  --rpc-port, --rpc-int-port, --toot| |Events api not used in this flow|
|deploy| --cache-artifacts , --cache-file,--insecure-registry, --skip-tests| | |
|dev | |  --force (for deploy phase,false by default) | |
 
Some considerations:
Maybe `toot` should be a common flag for build, delete, deploy and run without --tail.

### DIFF
--- a/cmd/skaffold/app/cmd/dev.go
+++ b/cmd/skaffold/app/cmd/dev.go
@@ -30,15 +30,15 @@ import (
 
 // NewCmdDev describes the CLI command to run a pipeline in development mode.
 func NewCmdDev(out io.Writer) *cobra.Command {
+	cmdUse := "dev"
 	return commands.
 		New(out).
-		WithDescription("dev", "Runs a pipeline file in development mode").
+		WithDescription(cmdUse, "Runs a pipeline file in development mode").
 		WithFlags(func(f *pflag.FlagSet) {
 			f.StringVar(&opts.Trigger, "trigger", "polling", "How are changes detected? (polling, manual or notify)")
 			f.StringSliceVarP(&opts.TargetImages, "watch-image", "w", nil, "Choose which artifacts to watch. Artifacts with image names that contain the expression will be watched only. Default is to watch sources for all artifacts")
 			f.IntVarP(&opts.WatchPollInterval, "watch-poll-interval", "i", 1000, "Interval (in ms) between two checks for file changes")
-			AddRunDevFlags(f)
-			AddDevDebugFlags(f)
+			AddFlags(f, cmdUse)
 		}).
 		NoArgs(cancelWithCtrlC(context.Background(), doDev))
 }
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -17,114 +17,240 @@ limitations under the License.
 package cmd
 
 import (
+	"reflect"
+
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
-	"github.com/spf13/cobra"
-	flag "github.com/spf13/pflag"
+	"github.com/spf13/pflag"
 )
 
-var (
-	AllFlags = []*flag.FlagSet{
-		commonFlagSet("common"),
-		buildFlagSet("build"),
-		eventsAPIFlagSet("events"),
-		deployPhaseFlagSet("deploy-phase"),
-		deployCommandFlagSet("deploy-cmd"),
-		testFlagSet("test"),
-		cleanupFlagSet("cleanup"),
-		devFlagSet("dev"),
+// Flag defines a Skaffold CLI flag which contains a list of
+// subcommands the flag belongs to in `DefinedOn` field.
+type Flag struct {
+	Name          string
+	Shorthand     string
+	Usage         string
+	Value         interface{}
+	DefValue      interface{}
+	FlagAddMethod string
+	DefinedOn     []string
+}
+
+// FlagResitry is a list of all Skaffold CLI flags.
+// When adding a new flag to the registry, please specify the
+// command/commands to which the flag belongs in `DefinedOn` field.
+// If the flag is a global flag, or belongs to all the subcommands,
+/// specify "all"
+// FlagAddMethod is method which defines a flag value with specified
+// name, default value, and usage string. e.g. `StringVar`, `BoolVar`
+var FlagResitry = []Flag{
+	{
+		Name:          "filename",
+		Shorthand:     "f",
+		Usage:         "Filename or URL to the pipeline file",
+		Value:         &opts.ConfigurationFile,
+		DefValue:      "skaffold.yaml",
+		FlagAddMethod: "StringVar",
+		DefinedOn:     []string{"all"},
+	},
+	{
+		Name:          "profile",
+		Shorthand:     "p",
+		Usage:         "Activate profiles by name",
+		Value:         &opts.Profiles,
+		DefValue:      []string{},
+		FlagAddMethod: "StringSliceVar",
+		DefinedOn:     []string{"all"},
+	},
+	{
+		Name:          "namespace",
+		Shorthand:     "n",
+		Usage:         "Run deployments in the specified namespace",
+		Value:         &opts.Namespace,
+		DefValue:      "",
+		FlagAddMethod: "StringVar",
+		DefinedOn:     []string{"all"},
+	},
+	{
+		Name:          "default-repo",
+		Shorthand:     "d",
+		Usage:         "Default repository value (overrides global config)",
+		Value:         &opts.DefaultRepo,
+		DefValue:      "",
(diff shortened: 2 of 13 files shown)


cmd/skaffold/app/cmd/dev.go
```diff
@@ -36,9 +36,16 @@ func NewCmdDev(out io.Writer) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return dev(out)
 		},
+		Annotations: map[string]string{
+			BuildAnnotation:  "ture",
```
priyawadhwa: nit: true

cmd/skaffold/app/cmd/flags.go
```diff
@@ -0,0 +1,117 @@
...
+		DevAnnotation:    devFlagSet(DevAnnotation),
+		BuildAnnotation:  buildFlagSet(BuildAnnotation),
+		EventsAnnotation: eventsAPIFlagSet(EventsAnnotation),
+		DeployAnnotation: deployFlagSet(DeployAnnotation),
+		TestAnnotation:   testFlagSet(TestAnnotation),
+		DebugAnnotation:  debugFlagSet(DebugAnnotation),
+	}
+)
+
+func commandFlagSet(name string) *flag.FlagSet {
```
priyawadhwa: nit: commandFlagSet/commonFlagSet?

cmd/skaffold/app/cmd/dev.go
priyawadhwa: maybe a 

```golang
map[string]struct{}
```
would make sense here if the string value isn't being used?

tejal29: Annotations is Builtin Field for cobra.Command Struct.

tejal29:
TODO: `--tail` has different defaults in depoy, dev and run.
Fix that

cmd/skaffold/app/cmd/flags.go
```diff
@@ -0,0 +1,111 @@
...
+	for a := range annotations {
+		flags, ok := AnnotationToFlag[a]
+		if ok {
+			allFlags.AddFlagSet(flags)
+		}
+	}
+	return allFlags
+}
+
+func addTailDevFlag(cmd *cobra.Command) {
```
priyawadhwa: Why is this a separate function, instead of an annotation, like `devFlagSet`?

cmd/skaffold/app/cmd/flags_test.go
```diff
@@ -0,0 +1,56 @@
...
+			annotations:   map[string]string{"some": "true", "test": "true"},
+			expectedFlags: []string{"skip-tests"},
+		},
+		{
+			description:   "not register annotations should return no flags",
+			annotations:   map[string]string{"some": "true"},
+			expectedFlags: nil,
+		},
+		{
+			description:   "union of anotations",
```
priyawadhwa: nit: spelling annotations

cmd/skaffold/app/cmd/flags.go
tejal29: This is  because of existing flag "--tail" is used for setting "opts.TailDev" and "opts.Tail".
In Deploy flow, "--tail" is set to opts.Tail with default set to false
However, in run and dev flow, "--tail" is set to opts.TailDev with default set to true

Because the code adds annotations as map, the order of keys 
```
		Annotations: map[string]string{
			BuildAnnotation:   "true",
			DeployAnnotation:  "true",
			TestAnnotation:    "true",
			CleanupAnnotation: "true",
			EventsAnnotation:  "true",
		},
	}
```
Dev and Build, also run deploy, so it gets messy. 
So inorder to solve this, in Dev and Run flow, we first register the "--tail" flag and bind it to opts.TaildDev.
Then in the `AddFlags` function when we register the same flag, it ignores it.

cmd/skaffold/app/cmd/flags_test.go
tejal29: thanks! will do

cmd/skaffold/app/cmd/flags.go
tejal29: Ideally, this should have been another flag like "--tail-dev" or a way to set defaults for "--tail" flag based on the command. But since flagSets do not have command reference, there isn't a way to do it.

tejal29:
@priyawadhwa  Added comment to describe the same.

tejal29:
Great! i feel with another example ("--force", "--tail") i should think a way to specify default behavior based on command. 
I will make changes to this PR.

tejal29:
@priyawadhwa  and @balopat  please take another look.
I actually ended up moving the annotations to flags now. 
This is was something @balopat  and I had discussed previously but i veto-ed it :)
Looks like i have to do it that way to add flags with different defaults.

tejal29:
Freeze for now until @balopat come back. As per last discussion, he would want to see flagset per flag with annotation for which command they belong to versus right now its a group of flags.

cmd/skaffold/app/cmd/flags.go
```diff
@@ -0,0 +1,264 @@
...
+	Name          string
+	Shorthand     string
+	Usage         string
+	Value         interface{}
+	DefValue      interface{}
+	FlagAddMethod string
+	DefinedOn     []string
+}
+
+// FlagResitry is a list of all Skaffold CLI flags.
```
balopat:
```suggestion
// FlagRegistry is a list of all Skaffold CLI flags.
```

cmd/skaffold/app/cmd/flags.go
```diff
@@ -0,0 +1,264 @@
...
+// When adding a new flag to the registry, please specify the
+// command/commands to which the flag belongs in `DefinedOn` field.
+// If the flag is a global flag, or belongs to all the subcommands,
+/// specify "all"
+// FlagAddMethod is method which defines a flag value with specified
+// name, default value, and usage string. e.g. `StringVar`, `BoolVar`
+var FlagResitry = []Flag{
```
balopat:
```suggestion
var FlagRegistry = []Flag{
```

pkg/skaffold/config/options.go
```diff
@@ -35,6 +35,7 @@ type SkaffoldOptions struct {
 	CacheArtifacts     bool
 	EnableRPC          bool
 	Force              bool
+	ForceDev           bool
```
balopat: do we still need this after the refactor? I think we should just be able to use `Force`.

cmd/skaffold/app/cmd/flags.go
```diff
@@ -0,0 +1,264 @@
...
+		Name:          "tail",
+		Usage:         "Stream logs from deployed objects",
+		Value:         &opts.Tail,
+		DefValue:      false,
+		FlagAddMethod: "BoolVar",
+		DefinedOn:     []string{"deploy"},
+	},
+	{
+		Name:          "force",
+		Usage:         "Recreate kubernetes resources if necessary for deployment, warning: might cause downtime!)",
```
balopat:
```suggestion
		Usage:         "Recreate kubernetes resources if necessary for deployment (warning: might cause downtime!)",
```

cmd/skaffold/app/cmd/flags.go
```diff
@@ -0,0 +1,264 @@
...
+		Name:          "tail",
+		Usage:         "Stream logs from deployed objects",
+		Value:         &opts.TailDev,
+		DefValue:      true,
+		FlagAddMethod: "BoolVar",
+		DefinedOn:     []string{"dev", "run", "debug"},
+	},
+	{
+		Name:          "force",
+		Usage:         "Recreate kubernetes resources if necessary for deployment, warning: might cause downtime!)",
```
balopat:
```suggestion
		Usage:         "Recreate kubernetes resources if necessary for deployment (warning: might cause downtime!)",
```

balopat: can we move the two "force"-s next to each other with a comment that yes, this is intentional?

tejal29: done! added a commit

cmd/skaffold/app/cmd/flags.go
tejal29: done

PR #2029, opened 2019-04-25
cut v0.28.0

balopat:
Fixes #2028

PR #2030, opened 2019-04-25
Remove upgrade to v1beta10

priyawadhwa:
I accidentally upgraded to v1beta10 in #2016 when v1beta9 had yet to be
released. To fix this, I've removed v1beta10 and merged the original
v1beta9->v1beta10 upgrade  into the current v1beta8->v1beta9 upgrade.

Should allow #2029 to go through


pkg/skaffold/schema/v1beta9/upgrade_test.go
```diff
@@ -1,127 +0,0 @@
...
-    plugin:
-      name: docker
-      properties:
-        dockerfile: path/to/Dockerfile
-  - image: gcr.io/k8s-skaffold/bazel
-    plugin:
-      name: bazel
-      properties:
-        target: //mytarget
-  executionEnvironment:
```
corneliusweig: This is gone in the v1beta8.upgrade_test. Is this intentional?

balopat: Good catch - I think it should be in v1beta8.upgrade_test as we had plugins/executionEnv in v1beta8 and removing them in v1beta9
@priyawadhwa ?

priyawadhwa: There should be tests for this now, thanks!

PR #2032, opened 2019-04-26
Correctly parse env-var for multi-valued flags

corneliusweig:
Before, multi-valued env-vars were not parsed correctly. For example
```
SKAFFOLD_INSECURE_REGISTRY="insecure1.io,insecure2.io" skaffold run
```
resulted in _one_ entry in the string array `["insecure1.io,insecure2.io"]`. This is because the cobra flag type `StringArrayVar` does not split its argument on commas.

This PR switches all string array flags to the cobra flag type `StringSliceVar`, which _does_ split its argument on commas. For example, the following is now possible:
```
SKAFFOLD_INSECURE_REGISTRY="insecure1.io,insecure2.io" skaffold run \
  --insecure-registry xyz.io \
  --insecure-registry abc.io,def.io  
# results in ["insecure1.io", "insecure2.io", "xyz.io", "abc.io", "def.io"]
```

See #2031

### DIFF
--- a/cmd/skaffold/app/cmd/build.go
+++ b/cmd/skaffold/app/cmd/build.go
@@ -51,7 +51,7 @@ func NewCmdBuild(out io.Writer) *cobra.Command {
 		},
 	}
 	AddRunDevFlags(cmd)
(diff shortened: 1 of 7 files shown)


tejal29:
yes looks good to me! But we still need to address an design issue if we shd have whitelist of flags being set from env variable.

PR #2035, opened 2019-04-26
Freeze v1beta9 config

corneliusweig:


corneliusweig:
I'll be away for a few days. So if anything needs to be amended, either ditch this PR or feel free to fix it without me.

PR #2037, opened 2019-04-29
Refactor kaniko builder to cluster builder

priyawadhwa:
Custom artifacts will need to work with both the local builder and the
cluster builder. However, currently the "cluster" builder is really just
the kaniko builder.

This refactor renames `pkg/build/kaniko` to `pkg/build/cluster`, and
reorganizes `pkg/build/cluster` so that it looks like `pkg/build/local`. This way, we
should easily be able to add a new CustomArtifact to the cluster
builder just as easily as we can to the local builder.


pkg/skaffold/build/cluster/cluster.go
```diff
@@ -42,11 +43,32 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, tags tag.ImageTags,
...
+	onlyKanikoBuilds := true
+	for _, a := range artifacts {
+		onlyKanikoBuilds = onlyKanikoBuilds && a.ArtifactType.KanikoArtifact != nil
+	}
+	if onlyKanikoBuilds {
+		return build.InParallel(ctx, out, tags, artifacts, b.buildArtifactWithKaniko)
+	}
+
+	// Otherwise, run the builds in sequence.
+	return build.InSequence(ctx, out, tags, artifacts, b.runBuildForArtifact)
```
balopat: hmm...it's a bit more complicated but I think we should collect the kaniko builds and run them in parallel with building all the custom artifacts in sequence. Building all kaniko artifacts in sequence sounds like a very slow solution and a waste of resources on a cluster.

tejal29: Something we should consider documenting when we add CustomBuilder.

tejal29: With the way this is organized right now, i dont think this is going to used until we support another builder type for in cluster.
Maybe we shd remove it for now? It will help us get our test coverage up too since this branch isn't tested

priyawadhwa: >Is it because we don't know what the scripts do and they might use same resources and could not run in parallel?

Yah, exactly. There's a section in my DD about letting users define if they want their builds to run in parallel or not, which could be an option.

Am happy to remove it for now and add it back later.

priyawadhwa:
@balopat I ended up removing the logic for running in sequence, and I'll add it back when I add the CustomArtifact to the cluster builder.

Issue #1815, opened 2019-03-18
with skaffold dev, the forwarded ports change on the code changes


jiraguha:
Hi,

Adopting skaffold dev for java dev, I am experiencing an issue similar to  #1594

Each time I make a change in my java code, all ports that are forwarded are incremented by 1.

### Information

- Skaffold version: v0.25.0
- Operating system: Mac OS X 10.14.3 x86_64
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta7
kind: Config
build:
  artifacts:
    - image: ijp/color-service
      jibGradle: {}
```
### Steps to reproduce the behavior

- When I run `skaffold dev`, I have: 8081 -> 8081
- On the 1st change, I have: 8081 -> 8082
- On the 2nd change, I have: 8081 -> 8083
- ...

Br,
JP

priyawadhwa:
Hey @jiraguha could you provide more information regarding your project (the contents of your k8s yamls would be helpful), or more specifically are pods recreated upon every code change?

We recently updated the port forward [key](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/kubernetes/port_forward.go#L244) to include pod name, so if a pod is being recreated upon every code change that would explain why this is happening.

Rio:
I just started using skaffold yesterday and I'm running into the same issue.
Here is my deployment and service as spit out by kustomize, nothing fancy:
```
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/component: ...
    app.kubernetes.io/name: ...
  name: ...
spec:
  ports:
  - name: http
    port: 3000
    protocol: TCP
    targetPort: http
  selector:
    app.kubernetes.io/component: ...
    app.kubernetes.io/name: ...
  type: ClusterIP
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/component: ...
    app.kubernetes.io/name: ...
  name: ...
spec:
  selector:
    matchLabels:
      app.kubernetes.io/component: ...
      app.kubernetes.io/name: ...
  template:
    metadata:
      labels:
        app.kubernetes.io/component: ...
        app.kubernetes.io/name: ...
    spec:
      containers:
      - envFrom:
        - secretRef:
            name: secret-248g9bh45m
            optional: false
        image: ...
        name: ...
        ports:
        - containerPort: 3000
          name: http
```

Rio:
[...]

PR #2047, opened 2019-05-01
Remove podname from port forward key

priyawadhwa:
This way, when pods are regenerated, they will be mapped to the same
port.

This should fix some of the issues users have been facing in #1815 and #1594.


tejal29:
is there a way to add a test for it? The test will help not breaking it in future.

PR #2048, opened 2019-05-01
Add custom artifact type to cluster builder

priyawadhwa:


pkg/skaffold/constants/constants.go
```diff
@@ -73,6 +73,9 @@ var (
 
 	// BuildContext is the absolute path to a directory this artifact is meant to be built from for custom artifacts
 	BuildContext = "BUILD_CONTEXT"
+
+	// Kubecontext is the expected kubecontext to build an artifact with a custom build script on cluster
+	Kubecontext = "KUBECONTEXT"
```
tejal29: nit: KubeContext shd match the "BUILD_CONTEXT" convention
```suggestion
	KubeContext = "KUBE_CONTEXT"
```

pkg/skaffold/build/cluster/cluster.go
```diff
@@ -42,7 +46,46 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, tags tag.ImageTags,
 		defer teardownDockerConfigSecret()
 	}
 
-	return build.InParallel(ctx, out, tags, artifacts, b.buildArtifactWithKaniko)
+	var kanikoArtifacts []*latest.Artifact
+	var otherArtifacts []*latest.Artifact
+
+	for _, a := range artifacts {
```
tejal29: Can we refactor this out to another function say
```
kaninoArtifacts, otherArtifacts := segregateArtifacts(artifacts)
```

That ways you add tests for this `segregateArtifacts`

pkg/skaffold/build/cluster/cluster.go
```diff
@@ -42,7 +46,46 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, tags tag.ImageTags,
...
+		}
+		otherArtifacts = append(otherArtifacts, a)
+	}
+
+	pArts, err := build.InParallel(ctx, out, tags, kanikoArtifacts, b.buildArtifactWithKaniko)
+	if err != nil {
+		return nil, errors.Wrap(err, "building artifacts in parallel with kaniko")
+	}
+
+	sArts, err := build.InSequence(ctx, out, tags, otherArtifacts, b.runBuildForArtifact)
```
tejal29: I think we should run build.InSequence even if build.InParallel failed.  In general we want to try building all artifacts and hopefully with @nkubala eventing change, users will get notifications of build passed or failed as they occur.

In another PR, we are thinking of adding a `--fail-fast` flag in cases where users do not want builds to continue building if there is a failure. The way you have it right now is `--fail-fast` mode.

pkg/skaffold/build/cluster/cluster.go
priyawadhwa: Thanks, done.

pkg/skaffold/build/cluster/cluster.go
priyawadhwa: sounds good, I changed it to run everything and return the error at the end

pkg/skaffold/constants/constants.go
```diff
@@ -73,6 +73,9 @@ var (
 
 	// BuildContext is the absolute path to a directory this artifact is meant to be built from for custom artifacts
 	BuildContext = "BUILD_CONTEXT"
+
+	// Kubecontext is the expected kubecontext to build an artifact with a custom build script on cluster
+	Kubecontext = "KUBE_CONTEXT"
```
tejal29:
```suggestion
	KubeContext = "KUBE_CONTEXT"
```

pkg/skaffold/build/cluster/cluster.go
```diff
@@ -42,7 +47,41 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, tags tag.ImageTags,
...
+}
+
+func (b *Builder) runBuildForArtifact(ctx context.Context, out io.Writer, artifact *latest.Artifact, tag string) (string, error) {
+	switch {
+	case artifact.KanikoArtifact != nil:
+		return b.buildArtifactWithKaniko(ctx, out, artifact, tag)
+
+	case artifact.CustomArtifact != nil:
+		return b.buildArtifactWithCustomBuilder(ctx, out, artifact, tag)
+
```
tejal29: I see you added bazel artifact in the test cases. 
shd there be a case of bazelArtifact ? or something you are going to add later?

priyawadhwa: I replaced bazelArtifact with customArtifact (guessing that's what you meant?), since bazel actually can't be run on cluster it's not really relevant here. 

Thanks for pointing that out!

docs/content/en/docs/how-tos/builders/_index.md
```diff
@@ -243,11 +243,15 @@ Skaffold will pass in the following additional environment variables for the fol
 | ------------- |-------------| -----|
 | Docker daemon environment variables     | Inform the custom builder of which docker daemon endpoint we are using. Allows custom build scripts to work with tools like Minikube. For Minikube, this is the output of `minikube docker-env`.| None. | 
 
+##### Cluster Builder
+| Environment Variable         | Description           | Expectation  |
+| ------------- |-------------| -----|
+| $KUBECONTEXT    | The expected kubecontext in which the image will be built.| None. |
```
balopat: This just occured to me: should we pass namespace as well?

balopat: How about the secret names? 
Like...theoretically in cluster one should be able to generate a kaniko pod definition and run the build through that, right?

priyawadhwa: good point, added the secrets, namespace and timeout.

I'm not sure what the best way to pass these [resource requirements](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/schema/latest/config.go#L246) is yet, but I'll tackle that in another PR.

pkg/skaffold/build/cluster/cluster.go
```diff
@@ -42,7 +47,47 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, tags tag.ImageTags,
...
+	default:
+		return "", fmt.Errorf("undefined artifact type: %+v", artifact.ArtifactType)
+	}
+}
+
+func (b *Builder) buildArtifactWithCustomBuilder(ctx context.Context, out io.Writer, artifact *latest.Artifact, tag string) (string, error) {
```
balopat: one last thing: can we cover that the right env vars are passed in to the builder in tests?

pkg/skaffold/build/cluster/cluster.go
```diff
@@ -42,7 +47,47 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, tags tag.ImageTags,
...
 
-	return build.InParallel(ctx, out, tags, artifacts, b.buildArtifactWithKaniko)
+	kanikoArtifacts, otherArtifacts := splitArtifacts(artifacts)
+
+	pArts, parallelErr := build.InParallel(ctx, out, tags, kanikoArtifacts, b.buildArtifactWithKaniko)
+	if parallelErr != nil {
+		logrus.Warnf("error building kaniko artifacts in parallel: %v", parallelErr)
+	}
+
+	sArts, err := build.InSequence(ctx, out, tags, otherArtifacts, b.runBuildForArtifact)
```
balopat: why did we decide at `InSequence`? Maybe we should mention this in the docs if we have a good reason - but it might actually just work with `InParallel` as well - in the end the majority of the work is just spinning up pods on cluster, right?

priyawadhwa: I think we did that in case one artifact depended on another -- but since we don't explicitly promise that artifacts will be built in order I think it's reasonable to just do `InParallel` for cluster builds.

pkg/skaffold/build/cluster/cluster.go
priyawadhwa: done

pkg/skaffold/build/cluster/cluster.go
balopat: Yeah, I think we should stick to `InParallel`

PR #2062, opened 2019-05-03
Prefix Skaffold labels with 'skaffold-'

charlyx:
Hi 👋

First of all, thanks for your work!

I hope this PR meets #2002 expectations 😄 

Thanks to @pyaillet for his help 🤗

### DIFF
--- a/pkg/skaffold/config/options.go
+++ b/pkg/skaffold/config/options.go
@@ -55,23 +55,15 @@ func (opts *SkaffoldOptions) Labels() map[string]string {
 	labels := map[string]string{}
 
 	if opts.Cleanup {
-		// deprecated
-		labels["cleanup"] = "true"
 		labels["skaffold-cleanup"] = "true"
 	}
 	if opts.Tail || opts.TailDev {
-		// deprecated
-		labels["tail"] = "true"
 		labels["skaffold-tail"] = "true"
 	}
 	if opts.Namespace != "" {
-		// deprecated
-		labels["namespace"] = opts.Namespace
 		labels["skaffold-namespace"] = opts.Namespace
 	}
 	if len(opts.Profiles) > 0 {
-		// deprecated
-		labels["profiles"] = strings.Join(opts.Profiles, "__")
 		labels["skaffold-profiles"] = strings.Join(opts.Profiles, "__")
 	}
 	for _, cl := range opts.CustomLabels {
--- a/pkg/skaffold/config/options.go
+++ b/pkg/skaffold/config/options.go
@@ -55,16 +55,16 @@ func (opts *SkaffoldOptions) Labels() map[string]string {
 	labels := map[string]string{}
 
 	if opts.Cleanup {
-		labels["skaffold-cleanup"] = "true"
+		labels["skaffold.dev/cleanup"] = "true"
 	}
 	if opts.Tail || opts.TailDev {
-		labels["skaffold-tail"] = "true"
+		labels["skaffold.dev/tail"] = "true"
 	}
 	if opts.Namespace != "" {
-		labels["skaffold-namespace"] = opts.Namespace
+		labels["skaffold.dev/namespace"] = opts.Namespace
 	}
 	if len(opts.Profiles) > 0 {
-		labels["skaffold-profiles"] = strings.Join(opts.Profiles, "__")
+		labels["skaffold.dev/profiles"] = strings.Join(opts.Profiles, "__")
 	}
 	for _, cl := range opts.CustomLabels {
 		l := strings.SplitN(cl, "=", 2)
--- a/pkg/skaffold/config/options.go
+++ b/pkg/skaffold/config/options.go
@@ -55,16 +55,24 @@ func (opts *SkaffoldOptions) Labels() map[string]string {
 	labels := map[string]string{}
 
 	if opts.Cleanup {
+		// deprecated
 		labels["cleanup"] = "true"
+		labels["skaffold-cleanup"] = "true"
 	}
 	if opts.Tail || opts.TailDev {
+		// deprecated
 		labels["tail"] = "true"
+		labels["skaffold-tail"] = "true"
 	}
 	if opts.Namespace != "" {
+		// deprecated
 		labels["namespace"] = opts.Namespace
+		labels["skaffold-namespace"] = opts.Namespace
 	}
 	if len(opts.Profiles) > 0 {
+		// deprecated
 		labels["profiles"] = strings.Join(opts.Profiles, "__")
+		labels["skaffold-profiles"] = strings.Join(opts.Profiles, "__")
 	}
 	for _, cl := range opts.CustomLabels {
 		l := strings.SplitN(cl, "=", 2)
--- a/pkg/skaffold/config/options_test.go
+++ b/pkg/skaffold/config/options_test.go
@@ -36,7 +36,6 @@ func TestLabels(t *testing.T) {
 			description: "cleanup",
 			options:     SkaffoldOptions{Cleanup: true},
 			expectedLabels: map[string]string{
-				"cleanup":          "true",
 				"skaffold-cleanup": "true",
 			},
 		},
@@ -44,7 +43,6 @@ func TestLabels(t *testing.T) {
 			description: "namespace",
 			options:     SkaffoldOptions{Namespace: "NS"},
 			expectedLabels: map[string]string{
-				"namespace":          "NS",
 				"skaffold-namespace": "NS",
 			},
 		},
@@ -52,7 +50,6 @@ func TestLabels(t *testing.T) {
 			description: "profile",
 			options:     SkaffoldOptions{Profiles: []string{"profile"}},
 			expectedLabels: map[string]string{
-				"profiles":          "profile",
 				"skaffold-profiles": "profile",
 			},
 		},
@@ -60,7 +57,6 @@ func TestLabels(t *testing.T) {
 			description: "profiles",
 			options:     SkaffoldOptions{Profiles: []string{"profile1", "profile2"}},
 			expectedLabels: map[string]string{
-				"profiles":          "profile1__profile2",
 				"skaffold-profiles": "profile1__profile2",
 			},
 		},
@@ -72,9 +68,6 @@ func TestLabels(t *testing.T) {
 				Profiles:  []string{"p1", "p2"},
 			},
 			expectedLabels: map[string]string{
-				"cleanup":            "true",
-				"namespace":          "namespace",
-				"profiles":           "p1__p2",
 				"skaffold-cleanup":   "true",
 				"skaffold-namespace": "namespace",
 				"skaffold-profiles":  "p1__p2",
@@ -92,7 +85,6 @@ func TestLabels(t *testing.T) {
 				},
 			},
 			expectedLabels: map[string]string{
-				"cleanup":          "true",
 				"skaffold-cleanup": "true",
 				"one":              "first",
 				"two":              "second",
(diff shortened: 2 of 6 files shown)


charlyx:
I signed it!

corneliusweig:
I'm a bit worried that somebody may forget to add the `skaffold-` prefix when this issue has been forgotten. I would prefer, if the labeler would check each label whether it needs to be prefixed. WDYT?

pkg/skaffold/config/options_test.go
```diff
@@ -77,11 +92,12 @@ func TestLabels(t *testing.T) {
 				},
 			},
 			expectedLabels: map[string]string{
-				"cleanup": "true",
-				"one":     "first",
-				"two":     "second",
-				"three":   "",
-				"four":    "",
+				"cleanup":          "true",
```
tejal29: why  cleanup and skaffold-cleanup both exist here?

tejal29:
> I'm a bit worried that somebody may forget to add the `skaffold-` prefix when this issue has been forgotten. I would prefer, if the labeler would check each label whether it needs to be prefixed. WDYT?


Thats is a great thought.  The labeller could have something is "AddSkaffoldPrefixIfRequired". However there is an exception to this rule when you want to add [k8 recommended style labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/). 
Maybe we could do some more strict validations like if its one of the recommended style label or not?

corneliusweig:
> However there is an exception to this rule when you want to add k8 recommended style labels .


@tejal29 good point. So at the very least, we should check if the label already has a `app.kubernetes.io` prefix, and if it does skip adding our prefix.

Further, it seems to be common convention to delimit label namespaces with a `/`. Therefore, if a label already contains a `/`, we should assume that the label is already namespaced and skip the skaffold prefix. WDYT? 

Now that I think about it, should we not use a more expressive prefix than `skaffold-`? My favorite would be `skaffold.dev/`, e.g. `skaffold.dev/tail`. That way, a completely oblivious user would be directed to the skaffold web page.

pkg/skaffold/config/options_test.go
pyaillet: If I understand correctly, in #2002, it's question of keeping both for sometimes to make sure not to break potential use from skaffold users.
If this is excessive precaution, we might just replace the existing ones.

pyaillet:
@corneliusweig I like the idea of a `skaffold.dev/` prefix.
However, what should be done of user label passed with `skaffold ... -l userlabel=uservalue` ?

corneliusweig:
@pyaillet good catch! I as a user would be surprised, if my label was prefixed. So I would say, no prefix for any explicitly user-specified labels.

corneliusweig:
@charlyx there is one contrived piece of code where the prefixing everything is not correct. The helm deployer needs to apply labels after creating the objects in k8s. For that, the labeler needs to merge existing and new skaffold labels. Those existing labels should of course _not_ be prefixed with  `skaffold...` .

pkg/skaffold/config/options_test.go
tejal29: ah understood. this is for backward compatibility.  However these are labels that skaffold uses internally to track pods that needed to cleaned-up. Users should not have to care about these labels.
So. removing them should be fine.

pyaillet:
I see three options there:
1. Add a `AddSkaffoldPrefixIfRequired` method and call it where [merge](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/deploy/labels.go#L51) is called. However user labels would be prefixed, which is not really satisfying
2. Add the `AddSkaffoldPrefixIfRequired` method, call it [there](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/config/options.go#L68), just before user labels handling. Keeping prefix hardcoded [here](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/constants/constants.go#L82)
3. Create a way to distinguish between ToBePrefixedLabels and ToBeKeptLabels by extending the `Labeller` interface, but this one seems a bit overengineered to me
WDYT ?

tejal29:
Thanks @pyaillet i feel we should go with your current implementation and enforce "skaffold-" prefix via code reviews and not use `AddSkaffoldPrefixIfRequired` at all. We don't want to add too many branches to the flow.

pyaillet:
I signed it!

charlyx:
I'm ok with my commits being contributed to this project.

pkg/skaffold/config/options.go
```diff
@@ -55,16 +55,16 @@ func (opts *SkaffoldOptions) Labels() map[string]string {
 	labels := map[string]string{}
 
 	if opts.Cleanup {
-		labels["cleanup"] = "true"
+		labels["skaffold-cleanup"] = "true"
```
corneliusweig: I still think that `skaffold.dev/` would be a better prefix than plain `skaffold-`. Either way will be fine though.

charlyx: How about labels that are already prefixed with `skaffold-` like the ones in `constants.go`?

For instance:

```golang
	TagPolicy:        "skaffold-tag-policy",
	Deployer:         "skaffold-deployer",
	Builder:          "skaffold-builder",
```

corneliusweig: They, too, should follow the same prefix convention. Whatever it will be :)

charlyx: OK, this is done.

tejal29:
@pyaillet Looks like you dont have signed the CLA.

Screenshot from 2019-05-06 11-08-25

pyaillet:
I signed it!

pyaillet:
Really strange, when I got to the CLA page I got:
"It looks like you've already signed this CLA. If you'd like to edit your contact information, you may do so below."
Sorry for the troubles :\

pyaillet:
Is it related to [this comment](https://github.com/GoogleContainerTools/skaffold/pull/2062#issuecomment-489262582) from GoogleBot ?

charlyx:
> Please have them confirm that here in the pull request

We have @tejal29 :)

priyawadhwa:
I think the CLA has been signed, looks like it just hasn't updated on the PR @tejal29 

image

tejal29:
Thanks @pyaillet and @priyawadhwa. I forced a Re-Scan and the  error is "CLAs are signed, but unable to verify author consent" which is terminal 
So i have to manually remove the label.
Thanks

Jonathan34:
Is it possible that this broke CI/CD:
```
Resource: "networking.istio.io/v1alpha3, Resource=virtualservices", GroupVersionKind: "networking.istio.io/v1alpha3, Kind=VirtualService"
Name: "istio-routes", Namespace: "development"
Object: &{map["apiVersion":"networking.istio.io/v1alpha3" "kind":"VirtualService" "metadata":map["annotations":map["kubectl.kubernetes.io/last-applied-configuration":"{\"apiVersion\":\"networking.istio.io/v1alpha3\",\"kind\":\"VirtualService\",\"metadata\":{\"annotations\":{},\"labels\":{\"app.kubernetes.io/managed-by\":\"skaffold-f25b09f\",\"environment\":\"development\",\"skaffold.dev/builder\":\"google-cloud-build\",\"skaffold.dev/cleanup\":\"true\",\"skaffold.dev/deployer\":\"kustomize\",\"skaffold.dev/profiles\":\"gcb\",\"skaffold.dev/tag-policy\":\"git-commit\",\"skaffold.dev/tail\":\"true\"},\"name\":\"istio-routes\",\"namespace\":\"development\"},\"spec\":{<redacted>}}\n"] "creationTimestamp":"2019-04-29T15:32:45Z" "generation":'\x01' "labels":map["app.kubernetes.io/managed-by":"skaffold-f25b09f" "environment":"development" "skaffold.dev/builder":"google-cloud-build" "skaffold.dev/cleanup":"true" "skaffold.dev/deployer":"kustomize" "skaffold.dev/profiles":"gcb" "skaffold.dev/tag-policy":"git-commit" "skaffold.dev/tail":"true"] "name":"istio-routes" "namespace":"development" "resourceVersion":"20715208" "selfLink":"/apis/networking.istio.io/v1alpha3/namespaces/development/virtualservices/istio-routes" "uid":"090cfcd6-6a94-11e9-af1e-42010a800034"] "spec":map[<redacted>]]}
for: "STDIN": VirtualService.networking.istio.io "istio-routes" is invalid: metadata.labels: Invalid value: "skaffold-": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
time="2019-05-08T15:10:51Z" level=fatal msg="deploy failed: kubectl error: kubectl apply: exit status 1"
```

pyaillet:
As the prefix set here is `skaffol.dev/`, I would say no

tejal29:
Looks like something else might be adding the label. 
Can you provide logs of skaffold run with --verbosity=debug ?

pyaillet:
It's strange, the only label prefixed by `skaffold-` in the logs is: `"app.kubernetes.io/managed-by":"skaffold-f25b09f"`

tejal29:
yeah looks like the version is not get populated. :( maybe we shd fix this to check if version is not empty.

corneliusweig:
But the version string (albeit weird) is not the problem, as it fits the regexp.

@Jonathan34 I think you are using kustomize to deploy, correct? Can you check if you do some label magic there? And are you using the `AdmissionWebhookController` or do you call istioctl manually? Maybe the clue as to what is going on is hidden there.

Jonathan34:
> Looks like something else might be adding the label.
> Can you provide logs of skaffold run with --verbosity=debug ?

is there a good place i can confidentially share the logs? 
I am running skaffold from the latest skaffold docker image. the deploy step is a kustomize overlay that worked yesterday morning (probably should not use skaffold:latest :P)

The guilty label seems to be the following:
`app.kubernetes.io/managed-by: skaffold-`

the tag policy i set is:
```yaml
 tagPolicy:
    gitCommit: {}
```
here is an abstract of the log:
```
Generating tags...
time="2019-05-08T17:03:56Z" level=debug msg="Running command: [git describe --tags --always]"
 - gcr.io/<redacted>/myservice -> time="2019-05-08T17:03:56Z" level=info msg="update check failed: get latest and current Skaffold version: parsing current semver, skipping update check: parsing semver: Version string empty"
time="2019-05-08T17:03:57Z" level=debug msg="Command output: [62577b7\n]"
time="2019-05-08T17:03:57Z" level=debug msg="Running command: [git status . --porcelain]"
gcr.io/<redacted>/myservice:62577b7
Tags generated in 1.4113427s

...
Build complete in 2m20.0544816s
Starting test...
Test complete in 20.9µs
Starting deploy...
kubectl client version: 1.14

```

corneliusweig:
> I am running skaffold from the latest skaffold docker image

I take it you are talking about `gcr.io/k8s-skaffold/skaffold:latest`?

Jonathan34:
> > I am running skaffold from the latest skaffold docker image
> 
> I take it you are talking about `gcr.io/k8s-skaffold/skaffold:latest`?

yes

corneliusweig:
So the skaffold image _does_ have a version set (albeit a strange one):
```sh
docker run --rm -ti gcr.io/k8s-skaffold/skaffold:latest skaffold version \
  -o "{{.Version}}  {{.GitCommit}} {{.GitTreeState}}"
```
yields
```
a117236  a1172365f8c099635277a46c20acfa1e1d6ed2a7 clean%
```

I think the reason why the version is as it is, is that `skaffold-v0.29.0` would not be a valid label due to the dots.

Overall, to me it does not look like `DefaultLabeller` would cause the bad label.

tejal29:
umm we shd fix that. Sorry about it.

PR #2072, opened 2019-05-06
Fix typo

posquit0:
Just typo issue


priyawadhwa:
Hey @posquit0 thanks for fixing this! I think you'll actually need to make this change in our schema [here](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/schema/latest/config.go#L208) and then run `make generate-schemas` to update the schema config (you can't directly change `v1beta9.json`, so you'll have to revert that change)

posquit0:
Thanks for the advice! I did it! :)

PR #2075, opened 2019-05-06
Add labels to podTemplates in helm deployments

### LINKED ISSUE #2074: Helm deployments do not add labels to PodTemplates
### Expected behavior
Skaffold should also label pods created via Deployment resources.

### Actual behavior
Skaffold adds labels to the `deployment.metadata.labels` but not to `deployment.spec.template.metadata.labels`.

### Information

- Skaffold version: HEAD

This clearly comes from this function where the live api-resource is patched: https://github.com/GoogleContainerTools/skaffold/blob/25f1b66402507440a12a0c1173806b190ff76eb8/pkg/skaffold/deploy/labels.go#L95
It's clear, that it only adds labels to the owning resource, but not to the pods it manages.

1. `cd examples/helm-deployment`
2. `skaffold dev`
3. `kubectl get deploy skaffold-helm -o jsonpath='{..labels}'` -> yields the expected set of skaffold labels:
    ```go
    map[app:skaffold-helm app.kubernetes.io/managed-by:skaffold-v0.28.0-80-g25f1b664 chart:skaffold-helm-0.1.0 cleanup:true docker-api-version:1.38 heritage:Tiller release:skaffold-helm skaffold-builder:local skaffold-deployer:helm skaffold-tag-policy:sha256 tail:true] map[app:skaffold-helm release:skaffold-helm]
    ```
4. `kubectl get pod skaffold-helm-XYZ -o jsonpath='{..labels}' yields _only the helm labels_
    ```go
    map[app:skaffold-helm pod-template-hash:6b75889977 release:skaffold-helm]
    ```

I think it would be preferrable, if also the dependent resources have the Skaffold labels.

corneliusweig:
Helm deployments need to be labeled after they are created and live. For that, Skaffold fetches the live resources and adds its labels. However, Skaffold only adds its labels to the outer `metatdata.labels` field, but not to any pod templates, such as the `deployment.spec.template`.

This PR ensures that for all common public api-resources with pod templates, the labels are also live-patched to the pod-template.

Fix #2074

### DIFF
--- a/pkg/skaffold/deploy/labels.go
+++ b/pkg/skaffold/deploy/labels.go
@@ -99,6 +99,15 @@ func addLabels(labels map[string]string, accessor metav1.Object) {
 	copyMap(kv, labels)
 
 	accessor.SetLabels(kv)
+
+	if podTpl := kubernetes.GetPodTemplateSpec(accessor); podTpl != nil {
+		dkv := make(map[string]string)
+
+		copyMap(dkv, podTpl.GetLabels())
+		copyMap(dkv, labels)
+
+		podTpl.SetLabels(dkv)
+	}
 }
 
 func updateRuntimeObject(client dynamic.Interface, disco discovery.DiscoveryInterface, labels map[string]string, res Artifact) error {
--- a/pkg/skaffold/kubernetes/util.go
+++ b/pkg/skaffold/kubernetes/util.go
@@ -1,0 +1,81 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kubernetes
+
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/api/apps/v1beta1"
+	"k8s.io/api/apps/v1beta2"
+	batchv1 "k8s.io/api/batch/v1"
+	batchv1beta1 "k8s.io/api/batch/v1beta1"
+	v1 "k8s.io/api/core/v1"
+	extv1beta1 "k8s.io/api/extensions/v1beta1"
+)
+
+// GetPodTemplateSpec extracts the PodTemplateSpec from public k8s api-resources.
+// This list will need to be extended manually for new api-versions.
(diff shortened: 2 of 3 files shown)


tejal29:
@corneliusweig , i am re-designing the labeller interface. 
Would you mind if i tackle this issue as a part of the re-design?

corneliusweig:
@tejal29 sounds as if I would cause more work, if I said no :wink:
So please do go ahead and tackle #2074 as part of your re-design. I'd be happy to review once it's done. Do you already have a target when you want to be done? Just curious, because #2074 is now a pre-requisite for the re-design of log-tailing (#1910).

Btw, I already looked into your idea to reuse the exiting recursive update strategy. I think it would work at the expense of additional marshalling/unmarshalling. I'm not a big fan of this, because it's much slower than the podTemplate getter from this PR. On the other hand, it will be guaranteed to work with all resource types.

tejal29:
@corneliusweig Thanks!
Regarding:
>Btw, I already looked into your idea to reuse the exiting recursive update strategy. I think it would work at the expense of additional marshalling/unmarshalling. I'm not a big fan of this, because it's much slower than the podTemplate getter from this PR. On the other hand, it will be guaranteed to work with all resource types.

Thanks for looking in to it. Another advantage of this approach is consistency. In of the reviews, i found helm was adding ""deployed-with": "skaffold"," label #2055.
So, having two flows is confusing.

corneliusweig:
I'll close this for now, because @tejal29 will take this up in a different PR.

tejal29:
Re-open!

tejal29:
@corneliusweig please wait for https://github.com/GoogleContainerTools/skaffold/pull/2140 and add tests for pod labels!

corneliusweig:
@tejal29 Can you briefly point out why you think this approach is better than what you tried?

If it is about the issue that pods are immediately destroyed and recreated -- I think it's the same with this approach (even though I haven't tested).

corneliusweig:
@tejal29 After thinking about this a bit more, I think the whole approach of re-labeling live objects is a flawed. For example, when `skaffold run` creates a new release and the user does a `helm` rollback to a previous revision. Do we expect the skaffold labels to be there or not? The least surprise would be to have labels, but this will not happen, because they are not part of the helm manifests. WDYT?

dgageot:
@corneliusweig @tejal29 What's the status on this PR?

corneliusweig:
So the current state is that it does solve #2074. But there are two drawbacks:

- Due to adding new labels to the podTemplate, the deployment recreates every pod after the first deployment. This may be unacceptable.
- When rolling back a helm release, the labels are gone, because they are applied manually after releasing a pod. This is not ideal, but probably acceptable (and also the current behavior).

---

Both problems come from patching the live resources instead of the original helm template. However neither @tejal29 nor me could find a proper way to do that.

corneliusweig:
Closing in favor of #2105 .
@tejal29 If the manifest labeller approach does not work out, I can re-open this one.

PR #2079, opened 2019-05-07
docs: correct header name for jump.

persevereVon:
Origin problem:
image

when I click `skaffold completion`, it does not work as expected.

I think i should change the header name to `skaffold completion`, but the usage is changed too. And the command's description is not clarified.

I wonder the `SHELL` is arguments or sub command, if sub command is better?

Thank you~

### DIFF
--- a/cmd/skaffold/app/cmd/completion.go
+++ b/cmd/skaffold/app/cmd/completion.go
@@ -230,7 +230,7 @@ func completion(cmd *cobra.Command, args []string) {
 // NewCmdCompletion returns the cobra command that outputs shell completion code
 func NewCmdCompletion(out io.Writer) *cobra.Command {
 	return &cobra.Command{
-		Use: "completion",
+		Use: "completion SHELL",
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) != 1 {
 				return fmt.Errorf("requires 1 arg, found %d", len(args))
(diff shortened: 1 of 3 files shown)


balopat:
@persevereVon, thanks for catching this - do you mind signing the CLA?

persevereVon:
OK, I signed it!

balopat:
1. CLA: @persevereVon - I think the issue with the CLA is that the email you gave is not tied to your github account - I see that the email in the commit is signed `wu5430...@163.com` but the CLA robot needs the _sender_ of the PR (your github account) to sign it too. 

2. as I'm looking at this, I find it a bit confusing that we are missing the SHELL argument. This command would be used like this: `skaffold completion bash` or `skaffold completion zsh`

persevereVon:
@balopat sorry, I filled in the wrong name in the GLA, I have updated it now. My fault :rofl: 

> 2\. as I'm looking at this, I find it a bit confusing that we are missing the SHELL argument. This command would be used like this: `skaffold completion bash` or `skaffold completion zsh`

yes, so I asked if SHELL is a argument or a subcommand, if it is a subcommand, then it can be clearly described in the usage.

balopat:
> @balopat sorry, I filled in the wrong name in the GLA, I have updated it now. My fault 
> 
> > 2. as I'm looking at this, I find it a bit confusing that we are missing the SHELL argument. This command would be used like this: `skaffold completion bash` or `skaffold completion zsh`
> 
> yes, so I asked if SHELL is a argument or a subcommand, if it is a subcommand, then it can be clearly described in the usage.

yes, SHELL is an argument

dgageot:
@persevereVon Thank you for this submission. `SHELL` is indeed an argument, not a subcommand.

I think the issue here is that the docs generator should be improved. 
The definition of the command is ok.

I think using `command.CommandPath()` instead of `command.UseLine()` in `skaffold/cmd/skaffold/man/man.go` does the job of fixing the doc.

WDYT?

persevereVon:
@dgageot Thank you!! you are right. It works. I pushed the commit to fix it just now.

PR #2083, opened 2019-05-07
[Design Proposal] Preventing kaniko logs leakage

prary:
Kaniko logs not being fetch in case either too many logs are produced or kaniko exit before logger is attached to it.

### DIFF
--- a/docs/design_proposals/kaniko-logs-leakage.md
+++ b/docs/design_proposals/kaniko-logs-leakage.md
@@ -64,8 +64,6 @@ Extra podGracePeriodSeconds field would be added, so that user can configure it.
 #1978
 
 
-Resolution: This is an implementation detail.
-
 ## Implementation plan
 
 1. Modify kaniko pod definition
--- a/docs/design_proposals/kaniko-logs-leakage.md
+++ b/docs/design_proposals/kaniko-logs-leakage.md
@@ -1,0 +1,76 @@
+# Skaffold logs improvements
+
+* Author(s): Prashant Arya (@prary)
+* Design Shepherd: 
+* Date: 7/05/2019
+* Status: 
+
+## Background
+
+Kaniko pod exiting before errors could be read by Skaffold. Do a normal skaffold run and put a invalid 
+base image or any other condition that would immediately(idea is to kill the kaniko pod before logger
+is attached ) fail to build a image and exit kaniko pod. In this condition user would not get acutual 
+reason of failure, he will just see following error
+
+build step: building [someBaseImageUrl]: kaniko build for [someBaseImageUrl]: waiting for pod to complete: pod already in terminal phase: Failed   
+
+
+Where as user should get actual error
+
+     Error while retrieving image from cache: could not parse reference
+     INFO[0002] Downloading base image gcr.io/invalidImageNameORUnReachableImage
+     error building image: could not parse reference
+
+
+This problem can also be exacted to a situation where user has put a such command which produce tons of output and before skaffold logger can pull all the logs kaniko pod  dies.
+
+## Design
+
+There should be a way where user can configure kaniko pod to stay alive for certain amount of time, so 
+that no logs are flushed before being fetched. There should be podGracePeriodSecond which ensures that
+kaniko pod stays alive for certain seconds after it finishes off building and pushing to registry. 
+
+Possible solution to counter above situation would be create a extra container in same kaniko pod
+which should continously monitor kaniko process. As soon as kaniko process terminate it should put 
+other container to sleep for specified mount of time. Extra container could be called a side car 
+which would not be very resource agnostic.
+
+### Kaniko Pod Definition changes
+
+```yaml
+Containers: []v1.Container{
+  {
+    Name:            constants.DefaultKanikoContainerName,
+    Image:           cfg.Image,
+    Args:            args,
+    ImagePullPolicy: v1.PullIfNotPresent,
+    Env:             []v1.EnvVar{},
+    VolumeMounts:    []v1.VolumeMount{},
+  },
+  {
+    Name:            "side-car",
+    Image:           constants.DefaultBusyboxImage,
+    ImagePullPolicy: v1.PullIfNotPresent,
+    Command: []string{"sh", "-c", "while [[ $(ps -ef | grep kaniko | wc -l) -gt 1 ]] ; do sleep 1; done; sleep " + cfg.PodGracePeriodSeconds},
+  },
+}
+```
+### Config changes
+
+Extra podGracePeriodSeconds field would be added, so that user can configure it.
+
+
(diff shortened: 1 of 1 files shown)


docs/design_proposals/kaniko-logs-leakage.md
```diff
@@ -0,0 +1,74 @@
...
+    Args:            args,
+    ImagePullPolicy: v1.PullIfNotPresent,
+    Env:             []v1.EnvVar{},
+    VolumeMounts:    []v1.VolumeMount{},
+  },
+  {
+    Name:            "side-car",
+    Image:           constants.DefaultBusyboxImage,
+    ImagePullPolicy: v1.PullIfNotPresent,
+    Command: []string{"sh", "-c", "while [[ $(ps -ef | grep kaniko | wc -l) -gt 1 ]] ; do sleep 1; done; sleep " + cfg.PodGracePeriodSeconds},
```
tejal29: I see you specify this Container as a side-car container which adds a sleep until cgf.PodGracePeriodSeconds which is 2 seconds. 
Shd we instead use the preStopContainer hook to block container deletion ?
Shd we block it for number of seconds or something else is a question to discuss here.
```
lifecycle:
          preStop:
            exec:
              # SIGTERM triggers a quick exit; gracefully terminate instead
              command: ["some blocking command"]
```

docs/design_proposals/kaniko-logs-leakage.md
```diff
@@ -0,0 +1,74 @@
+# Skaffold logs improvements
+
+* Author(s): Prashant Arya (@prary)
+* Design Shepherd:
```
tejal29: Tejal Desai

tejal29: Feel free to add me here! i am interested in this discussion!

docs/design_proposals/kaniko-logs-leakage.md
prary: hi @tejal29 ,

IMHO I think preStopContainer are applicable only when api server sends kill pod command to a running pod, not when pod itself kills. Below is the link I have previously commented, correct me if i am missing something. We can have other solutions as well, I more of the option kaniko is the one where logs should be stored and streamed.

https://github.com/GoogleContainerTools/skaffold/issues/1978#issuecomment-489061535

docs/design_proposals/kaniko-logs-leakage.md
prary: sure

docs/design_proposals/kaniko-logs-leakage.md
tejal29: yes, i think we discussed about `kubectl describe` when we internally triaged #1978 . I just forgot about it. We can definitely rely on kubectl describe.  Should we do that when logs are not present?

prary: Correct me if I am wrong, you are pointing when kaniko is not producing any log in that case we should rely on kubectl describe, is that what you mean?

tejal29: yes. if we see an error or no lines when retrieving logs, we should run kubectl describe

tejal29: btw, Thanks @prary  for pointing out preStopContainers are only applicable when api server kills the pod.

prary: I think there is no need for kubectl describe we are already monitoring the status of pod in WaitForPodComplete function in https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/kubernetes/wait.go . Or we may just add a log in PodRunning switch case i.e 

```
                case v1.PodRunning:
                        logrus.Infof("Pod is running")
			return false, nil
		case v1.PodFailed:
			return false, fmt.Errorf("pod already in terminal phase: %s", pod.Status.Phase)
```

cedrickring:
> too many logs are produced or kaniko exit before logger is attached to it

Kubernetes is pretty good at handling big log streams. The problem relies in the current log stream implementation as it waits until the pod is ready. If the pod exits during that period, no logs are streamed.

Maybe we could reduce the interval time here:
https://github.com/GoogleContainerTools/skaffold/blob/b13924053dc1345ca31b4cd751c3a36caef45e22/pkg/skaffold/build/cluster/logs.go#L55

or just look for the lines streamed. If no lines have been streamed before exiting, one more log call should be done there.

e.g.

```go
var read int64
...
n, _ = io.Copy(out, r)
read += n
```
and here
https://github.com/GoogleContainerTools/skaffold/blob/b13924053dc1345ca31b4cd751c3a36caef45e22/pkg/skaffold/build/cluster/logs.go#L64

we can check if `read == 0` and request the logs.

WDYT

docs/design_proposals/kaniko-logs-leakage.md
priyawadhwa: Personally I like providing the output of kubectl describe upon failure because it could provide valuable information, and because the user won't be able to access it themselves since the pod will be cleaned up by skaffold.

prary: You mean 
```
case v1.PodFailed:
        **kubectl describe over here**
	return false, fmt.Errorf("pod already in terminal phase: %s", pod.Status.Phase)
```
Is that where you think kubectl describe should  be?

prary:
> Maybe we could reduce the interval time here:
I personally think it won't be a good idea.

> ```go
> var read int64
> ...
> n, _ = io.Copy(out, r)
> read += n
> ```
Will this work even when logger is not attached?

docs/design_proposals/kaniko-logs-leakage.md
prary: @priyawadhwa @dgageot I think fetching events from kubernetes would be better approach than kubectl describe i.e
```
case v1.PodFailed:
   kubectl get events --namespace namespace-name --field-selector involvedObject.name=kaniko-pod-name
```
WDYT? This would fetch all event logs even when pod is failed due to some kubernetes constraints failure like image-builder(which we use for kaniko pod) service account not present or maybe attaching secret volume timeout due to some random reason.

cedrickring:
I don't know why the logger shouldn't attach to the process. Even if the executable only runs for a ms, the logger will still catch it's output.

When I implemented the fix for me, the logs were pulled properly from the pod.

@prary maybe I'm getting this question wrong, but are you missing logs when Kaniko exits immediately?

tejal29:
is blocked by #2311

tejal29:
running kokoro build  here: https://sponge.corp.google.com/invocation?id=46060480-67a6-43f5-848f-237e248c6243&searchFor=

tejal29:
Discarding this

Issue #1980, opened 2019-04-18
skaffold not working with kind and docker for mac


sbhardwaj-mt:
### Expected behavior.
skaffold deployment with helm should be able to pull the images that are build during development phase.

### Actual behavior
skaffold deployment with helm is not able to pull the images that are built during development mode at the run time in pod

### Information

- Skaffold version: version... 0.27.0
- Operating system: ... Mac OSx 10.14.2
- Contents of skaffold.yaml: 

```yaml
apiVersion: skaffold/v1beta8
kind: Config
build:
  artifacts:
    - image: munimji
      context: ./../munimG
      sync:
        'src/*.go': .
        'src/**/*.go': .

  tagPolicy:
    envTemplate:
      template: "{{.IMAGE_NAME}}:latest"

deploy:
  helm:
    releases:
      - name: initservice
        version: 0.1.0
        chartPath: ./../baton-helm-charts/initservice

      - name: munimji
        version: 0.1.0
        chartPath: ./../baton-helm-charts/munimji
        recreatePods: true
        setValues:
          env.GIN:
```

### Steps to reproduce the behavior
1. install kind with docker. 
2. try to build any image and deploy using helm.

corneliusweig:
Skaffold does not recognize the kubernetes context created by kind as local by default. Kind uses rather generic context names, like `kubernetes-admin@1`, so that it would be questionable for Skaffold to treat contexts matching `kubernetes-admin@\d+` as local by default.

Try to configure your kubernetes context as local, see https://skaffold-latest.firebaseapp.com/docs/concepts/#local-development for details. You probably need to do
```bash
skaffold config set -k 'kubernetes-admin@1' local-cluster true
```

sbhardwaj-mt:
kind was by default running on context name with kubernetes-admin@kind.
we tried  skaffold config set -k 'kubernetes-admin@kind' local-cluster true. 
and redeployed but we are getting the same error. 

`Warning  Failed     13s (x2 over 30s)  kubelet, test-control-plane  Failed to pull image "wait-for-it": rpc error: code = Unknown desc = Error response from daemon: pull access denied for wait-for-it, repository does not exist or may require 'docker login'`

corneliusweig:
@sbhardwaj-mt Have you also set up the docker environment as described in the docs?

corneliusweig:
@sbhardwaj-mt do you by any chance set an
`imagePullPolicy`? And could you also show the pod manifest?

[...]

PR #2090, opened 2019-05-08
Mention kind in docs for local development

corneliusweig:
See #1980


corneliusweig:
I'm asking myself, if it is acceptable to link to answers on github issues from the skaffold.dev docs.

I think this could help people getting more detailed information while at the same time keeping the docs somewhat manageable. On the other hand, the "doc experience" with github issues is not very comfortable. WDYT?

balopat:
> I'm asking myself, if it is acceptable to link to answers on github issues from the skaffold.dev docs.
> 
> I think this could help people getting more detailed information while at the same time keeping the docs somewhat manageable. On the other hand, the "doc experience" with github issues is not very comfortable. WDYT?

I think we should in general copy and sanitize reference type of information from Github issues as they are "mutable" things can change, deleted, moved and originally meant for discussion not for documentation. For historical reference we can definitely link to them - but not more than that.

PR #2091, opened 2019-05-08
[WIP] Single channel

balopat:
experiment


tejal29:
i am going to take another stab and update what i have in mind.

tejal29:
@balopat i added this commit https://github.com/GoogleContainerTools/skaffold/pull/2091/commits/99a4da97d3aac6173fa7ba32de7b69a262102e4e, 
Where the signature of Builder is different than what you have, 
1. The runner provides a channel to right too
2. The builder just sends artifacts to these channels as they are built
VS 
the current design where the Builder returns a single channel.

Thanks

balopat:
I'm not sure I agree with this change @tejal29 - this creates 2 ways to return the same information: through the channel and as the return slice. I find that confusing.

balopat:
After multiple days invested in this direction I still feel that it is not worthwhile to go down this direction: the complexity of adding go channels adds a lot of complexity and the design more error prone for subtle bugs. The current arguments for going towards this direction is the Event API redesign to centralize event management in the Runner. This in itself is a valuable idea but at this point the current design is simpler to reason about this new design with the go channels, plus if we want to extend the events sent by builders, e.g. progress bar events, we will have to send data through the go channels anyway that represents those events and convert them to events, or directly call the `event` package - at which point the complexity of "handling events from the builders" is going to be the same. 

PR #2099, opened 2019-05-09
Fix image release process: master -> edge, tag -> latest

balopat:
Also fixes a bug that caused sometimes the version to disappear - now if the Makefile receives an empty string, it will still override the `VERSION` variable with the abbreviated git commit.


tejal29:
Awesome find @balopat

PR #2108, opened 2019-05-10
Profiles syntax version dependency

venkyvb:
Trying out profiles with an earlier apiVersion: v1beta7 and realized that this needs v1beta10. Wanted to make sure that this is captured in the documentation.

### DIFF
--- a/docs/content/en/docs/how-tos/profiles/_index.md
+++ b/docs/content/en/docs/how-tos/profiles/_index.md
@@ -16,6 +16,8 @@ For a detailed discussion on Skaffold configuration, see
 
 ## Profiles (`profiles`)
 
+**NOTE:** To use profiles you should use the skaffold apiVersion: `skaffold/v1beta10`
+
 Each profile has six parts:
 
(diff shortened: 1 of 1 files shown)


venkyvb:
I have signed the CLA.

venkyvb:
> Hi @venkyvb thanks for opening this, it is an interesting point.
> 
> 1. I don't think we should hard code the version into a single page. Maybe adding the skaffold config version in the footer/header could help?
> 2. We do have a `{{< skaffold-version >}}` macro that resolves to the latest skaffold config version.

Thanks. However I am wondering if it would make sense to peg the version here as this is an API dependency. I assume that the macro will always reflect the latest version right isnt it?

venkyvb:
> Hi! I'm not sure I understand what's not supported before `v1beta10`

This was about the support for profiles in skaffold.yaml

dgageot:
@venkyvb Profiles were added in `v1alpha3`. Maybe I misunderstand your issue about `v1beta10` being the first version to support them.

PR #2109, opened 2019-05-10
Freeze v1beta10 config

corneliusweig:


tejal29:
@corneliusweig do you have a config change PR ready to merge?
We don't usually bump it unless there is a config change in pipeline. 
If there is an PR, please add this to the ticket

corneliusweig:
@tejal29 So there are #2088 and #1905.
- #1905 only makes additional enum values valid, so id may not necessarily need a new apiVersion.
- #2088 is the inferred sync mode, but it may take a while to get merged.

I could as well cherry-pick this PR on top of #2088 and #1905 goes under the radar. WDYT?

PR #2131, opened 2019-05-16
Fix and improve sync samples

dgageot:
+ Fixed the `sync` configuration on `hot-reload` sample
+ Fix all samples so that they actually work
+ Convert pods into deployments. Now that port forwarding is not activated, deployments need to be exposed with a service
+ Simplify the `skaffold.yaml` files with regards to deploy config.
+ Improve the Dockerfiles (caching, smaller and more recent base images...)


dgageot:
@corneliusweig Could you take a look at those fixes? Thanks!

integration/examples/react-reload/skaffold.yaml
```diff
@@ -5,9 +5,8 @@ build:
   - image: gcr.io/k8s-skaffold/react-reload
     context: app
     sync:
-      'src/components/*': src/components/
```
corneliusweig: oops.. this should not have happened (@.myself). The reason why it could is that `react-reload` has no test-case in `run_test.go`.

integration/examples/hot-reload/node/Dockerfile
```diff
@@ -1,5 +1,9 @@
...
 
+WORKDIR /app
 EXPOSE 3000
-CMD ["nodemon","--legacy-watch", "server.js"]
-COPY src ./
+CMD ["npm", "run", "dev"]
+
+COPY package* ./
+RUN npm install
+COPY src .
```
corneliusweig:
```suggestion
COPY src src/
```
Otherwise docker flattens the directory at the destination. And in the `package.json` the main file is `src/index.js`. I think it only works by accident, because node looks for `index.js` in the root folder if the entrypoint is not found.

dgageot: I think the idea was to flatten the directory. in the skaffold.yaml, sync is configured to strip the `src/` prefix.
I'll fix the `package.json` though.

dgageot:
Good point. I'll push a PR to test more samples (not sure it's easy to test them all) after this PR is merged.

corneliusweig:
@dgageot LGTM

PR #2138, opened 2019-05-16
skaffold debug: log unsupported objects or versions

briandealwis:
We don't support workload objects in `apps/v1beta1` or `apps/v1beta2`, which have been deprecated since Kubernetes 1.9.  Log these in a more visible manner:
```
ERRO[0003] deprecated versions not supported by debug: deployment.apps/web2 (v1beta1) 
DEBU[0004] no debug transformation for: service/web          
WARN[0005] no debug transformation for: foobar.apps/name
```
The warning should only be issued workload API objects (`apps` and `batch`) that aren't handled.

### DIFF
--- a/pkg/skaffold/debug/transform.go
+++ b/pkg/skaffold/debug/transform.go
@@ -18,6 +18,8 @@ package debug
 
 import (
 	"encoding/json"
+	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -25,6 +27,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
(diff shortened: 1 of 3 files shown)


pkg/skaffold/debug/transform.go
```diff
@@ -200,3 +213,24 @@ func encodeConfigurations(configurations map[string]map[string]interface{}) stri
...
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	group = gvk.Group
+	version = gvk.Version
+	kind = gvk.Kind
+	if group == "" {
+		description = fmt.Sprintf("%s/%s", strings.ToLower(kind), name)
+	} else {
+		description = fmt.Sprintf("%s.%s/%s", strings.ToLower(kind), group, name)
+	}
+	return
```
tejal29: we dont return anything here or am i missing some hidden go mojo that happens here?

pkg/skaffold/debug/transform.go
```diff
@@ -200,3 +213,24 @@ func encodeConfigurations(configurations map[string]map[string]interface{}) stri
 	}
 	return string(bytes)
 }
+
+func describe(obj runtime.Object) (group, version, kind, description string) {
+	// get metadata/name; shamelessly stolen from from k8s.io/cli-runtime/pkg/printers/name.go
+	name := "<unknown>"
+	if acc, err := meta.Accessor(obj); err == nil {
```
tejal29: We can add test here by creating mock objects like:
```
type FakeObject struct {
     kind schema.ObjectKind
}

func (f *FakeObject) GetObjectKind() schema.ObjectKind {
   return f.kind
}

func (f *FakeObject) DeepCopyObject() {
  return f
}

type FakeObjectKind struct {
     group string
     kind string
     version string
}

func (o *FakeObjectKind) SetGroupVersionKind(kind GroupVersionKind) {
    fmt.Errorf("nyi")
}
func (o *FakeOjectKind) GroupVersionKind() schema.GroupVersionKind {
  return schema.GroupVersionKind {
     Group: o.group, 
     Version: o.version, 
     Kind: o.kind,
}
}

```

We can pass this object to `describe` and see if we get expected results.

Thanks
Tejal

pkg/skaffold/debug/transform.go
briandealwis: [_named return values_](https://tour.golang.org/basics/7)

pkg/skaffold/debug/transform.go
briandealwis: I did write some tests, but the setup was getting hairy.

/me hangs head in shame 😭

briandealwis: I pushed an alternative version that uses real resources.

PR #2141, opened 2019-05-16
Add support for npm run-script-based launches for `skaffold debug`

### LINKED ISSUE #2044: Can `skaffold debug` support npm run-scripts
`skaffold debug` supports configuring NodeJS applications for debugging providing their the container entrypoint or command is `node` or `nodemon`.  But some apps, like our own `examples/nodejs` use `npm run xxx` to run a script: (https://github.com/GoogleContainerTools/skaffold/blob/11f8a1bc593906c6cb00b696b4317e85b8a48d05/examples/nodejs/backend/Dockerfile#L5
which itself executes `nodemon`:
https://github.com/GoogleContainerTools/skaffold/blob/11f8a1bc593906c6cb00b696b4317e85b8a48d05/examples/nodejs/backend/package.json#L7

Perhaps `skaffold debug` for NodeJS should parse and retrieve the script line.

briandealwis:
Fixes #2044

Adds support for launches by passing the node `--inspect=xxxx` option via npm's `--node-options` command line argument.  This must appear after the run-script but before a `--` (should it be present).  npm uses this `--node-options` argument to configure the `NODE_OPTIONS` environment which is picked up by NodeJS 8 and higher.

Note: npm run-scripts that use `nodemon` (a node-based script) instead of `node` will not work as `nodemon`'s `node` will interpret the `NODE_OPTIONS` and configure itself for debugging.

  - add limitations to docs
  - add `debug` integration tests for `jib`, `nodejs`, and `npm` artifacts


chanseokoh:
The Maven test project looks good.

PR #2157, opened 2019-05-20
Add smoke test for `skaffold diagnose`

corneliusweig:
The `skaffold diagnose` command has bad test coverage (both integration and unit tests). There are reasons why it may not make sense to have full unit test coverage there. However, basic test coverage should be there to check whether the subcommand runs at all.

This PR adds basic smoke tests which check if `skaffold diagnose` runs successfully. It covers all artifact types that have an example.


balopat:
Nice!!! Thank you for this!

PR #2163, opened 2019-05-23
Improve cmd creation

dgageot:
+ Make sure a command name is always passed to the Builder
 + Remove a few duplicate flags
 + Make it easier to add common flags

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -59,7 +59,7 @@ var FlagRegistry = []Flag{
 		Value:         &opts.Profiles,
 		DefValue:      []string{},
 		FlagAddMethod: "StringSliceVar",
-		DefinedOn:     []string{"all"},
+		DefinedOn:     []string{"dev", "build", "run", "debug", "deploy", "delete", "diagnose"},
 	},
 	{
 		Name:          "namespace",
@@ -68,7 +68,7 @@ var FlagRegistry = []Flag{
 		Value:         &opts.Namespace,
 		DefValue:      "",
 		FlagAddMethod: "StringVar",
-		DefinedOn:     []string{"all"},
+		DefinedOn:     []string{"dev", "build", "run", "debug", "deploy", "delete"},
 	},
 	{
 		Name:          "default-repo",
@@ -77,11 +77,11 @@ var FlagRegistry = []Flag{
 		Value:         &opts.DefaultRepo,
 		DefValue:      "",
 		FlagAddMethod: "StringVar",
-		DefinedOn:     []string{"all"},
+		DefinedOn:     []string{"dev", "build", "run", "debug", "deploy", "delete"},
 	},
 	{
 		Name:          "cache-artifacts",
-		Usage:         "Set to true to enable caching of artifacts.",
+		Usage:         "Set to true to enable caching of artifacts",
 		Value:         &opts.CacheArtifacts,
 		DefValue:      false,
 		FlagAddMethod: "BoolVar",
--- a/cmd/skaffold/app/cmd/build.go
+++ b/cmd/skaffold/app/cmd/build.go
@@ -22,7 +22,6 @@ import (
 	"io/ioutil"
 	"time"
 
-	"github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/cmd/commands"
 	"github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/flags"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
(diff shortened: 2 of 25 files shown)


hack/check-docs.sh
```diff
@@ -14,16 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cp docs/content/en/docs/references/cli/index_header docs/content/en/docs/references/cli/_index.md
-go run cmd/skaffold/man/man.go >> docs/content/en/docs/references/cli/_index.md
-
-readonly CLI_CHANGES=`git status -s | grep "docs/" | wc -l`
-
-if [ $CLI_CHANGES -gt 0 ]; then
```
balopat: This consistency check is important to have in travis.

cmd/skaffold/app/cmd/flags.go
```diff
@@ -17,243 +17,164 @@ limitations under the License.
...
-		Value:         &opts.EnableRPC,
-		DefValue:      false,
-		FlagAddMethod: "BoolVar",
-		DefinedOn:     []string{"dev", "build", "run", "debug", "deploy"},
+		DefinedOn: []string{"dev", "build", "run", "debug", "deploy"},
+		Setup: func(f *pflag.FlagSet) {
+			f.BoolVar(&opts.EnableRPC, "enable-rpc", false, "Enable gRPC for exposing Skaffold events (true by default for `skaffold dev`)")
+		},
 	},
+	//
```
balopat: nit: remove unnecessary comment

cmd/skaffold/app/cmd/flags.go
```diff
@@ -17,243 +17,164 @@ limitations under the License.
...
 	{
-		Name:          "tail",
-		Usage:         "Stream logs from deployed objects. (default false)",
-		Value:         &opts.Tail,
-		DefValue:      false,
-		FlagAddMethod: "BoolVar",
-		DefinedOn:     []string{"deploy"},
+		DefinedOn: []string{"deploy"},
+		Setup: func(f *pflag.FlagSet) {
+			// TODO: default?
```
balopat: what's the TODO?

cmd/skaffold/app/cmd/flags.go
```diff
@@ -17,243 +17,164 @@ limitations under the License.
...
+		Setup: func(f *pflag.FlagSet) {
+			f.BoolVar(&opts.PortForward, "port-forward", false, "Port-forward exposed container ports within pods")
+		},
 	},
 }
 
-var commandFlags []*pflag.Flag
-
-// SetUpFlags creates pflag.Flag for all registered flags
-func SetUpFlags() {
```
balopat: I was personally okay with having the reflection based logic to have the nice struct.
It is admittedly a bit of a hack in the sense that it uses reflect and that it bets on the shared structure of the  flagset API but results in a named structure that I find more readable/usable than the single line version of the flags and that is the key for me. I guess this is one of those "opinion" based decisions. Do you feel strongly about this direction?

hack/check-docs.sh
dgageot: It’s done as a unit test now

cmd/skaffold/app/cmd/flags.go
dgageot: I’ve got an updated version without this comment, that I couldn’t push before I took the plane :-)

cmd/skaffold/app/cmd/flags.go
dgageot: Yeah, I felt bad of kinda rolling back @tejal29´s effort. I personally like the fact that most flags are now defined globally. I didn’t really like the super verbose structure that uses reflection when every flag is a one liner. But as you say, it’s a matter taste. I’m happy to remove that part of the PR if you prefer

cmd/skaffold/app/cmd/flags.go
dgageot: I’ve remove that too.

cmd/skaffold/app/cmd/flags.go
balopat: @tejal29 wdyt? I'm okay either way.

tejal29: I think the motivation for making it more verbose was so that the flags are easy to read.
Right now,  one needs to dig deeper in the method signature to identify what is &opts.PortForward, "false" or "Port-forward exposed container ports within pods"

```
f.BoolVar(&opts.PortForward, "port-forward", false, "Port-forward exposed container ports within pods")
```

I had other iterations on my branch where i used BoolVar and related methods.

balopat: I agree - for me too, it looks like that the space-saving reduces readability. Let's revert to structs, if you don't mind, @dgageot?

dgageot: I'll do that!

dgageot:
@balopat @tejal29 I reverted part of the changes. This PR should be ready for a review.

dgageot:
Let me find why the coverage is decreased and fix that.

dgageot:
I'm going to close this PR and rework on it

PR #2180, opened 2019-05-27
Check man page with a unit test instead of a script

dgageot:


dgageot:
It should be the same for contributors. Yes, the idea is to move the testing steps from `make test` into the normal testing flow.

dgageot:
@balopat Could you TAL at this one. It's very minor

PR #2182, opened 2019-05-27
Print logs from init containers

corneliusweig:
This is based on #1865, but now skips terminated containers.

Close #1865.
CC @tejal29  who raised this concern on #1865 and @ottonello who opened #1865 .

### DIFF
--- a/pkg/skaffold/kubernetes/log.go
+++ b/pkg/skaffold/kubernetes/log.go
@@ -104,7 +104,7 @@ func (a *LogAggregator) Start(ctx context.Context) error {
 					continue
 				}
 
-				for _, container := range pod.Status.ContainerStatuses {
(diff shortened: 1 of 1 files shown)


corneliusweig:
@ottonello Looks like you need to explicitly express your consent that my additional changes are ok (https://opensource.google.com/docs/cla/#github-consent).

corneliusweig:
@ottonello I totally missed that there were other issues with the original PR.
I don't think that the kaniko failure is related to this change though, at least I could not reproduce an error.

I'll try to set up an integration test for logs from init containers.

tejal29:
Thanks @corneliusweig for taking this up again. Running kokoro now. 
Unfortunately there are no tests for log.go :(. Maybe we can cover it later

PR #2196, opened 2019-05-29
Use kind to run integration tests on TravisCI

dgageot:
This is a first try at using [kind](https://github.com/kubernetes-sigs/kind) for our integration tests.

+ It creates a kind cluster in Travis.
+ It runs the test suite in the same container that's currently used on kokoro.
+ I had to add a few lines of code in skaffold to handle kind clusters. The main change is that docker images have to be explicitly loaded before they are used by a k8s workload.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,3 +34,10 @@ jobs:
       script:
         - go build -o out/skaffold.exe cmd/skaffold/skaffold.go
         - go test -short -timeout 60s ./...
+    - stage: integration
+      os: linux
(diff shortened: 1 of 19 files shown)


dgageot:
@tejal29 @balopat @nkubala This PR is ready to be reviewed. Be aware that it doesn't run all the tests we used to run on kokoro. Because it has no gcp credentials, it can't use cloud build or have kaniko push to gcr. Tell me what you think

PR #2222, opened 2019-06-05
Use test wrapper in more tests

dgageot:

### DIFF
--- a/cmd/skaffold/app/cmd/config/config_test.go
+++ b/cmd/skaffold/app/cmd/config/config_test.go
@@ -39,36 +39,43 @@ var baseConfig = &Config{
 var emptyConfig = &Config{}
 
 func TestReadConfig(t *testing.T) {
-	c, _ := yaml.Marshal(*baseConfig)
-	cfg, teardown := testutil.TempFile(t, "config", c)
-	defer teardown()
-
 	var tests = []struct {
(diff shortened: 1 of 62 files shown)


cmd/skaffold/app/cmd/config/config_test.go
```diff
@@ -217,23 +226,21 @@ func TestSetAndUnsetConfig(t *testing.T) {
...
 
-			if test.shouldErrSet {
+			if test.shouldErr {
 				// if we expect an error when setting, don't try and unset
 				return
 			}
 
 			// unset the value
 			err = unsetConfigValue(test.key)
 			newConfig, cfgErr := readConfig()
```
tejal29: not an issue with this PR. But we shd de-couple testing set and unset behavior in the same test into 2 separate tests.

testutil/checks.go
```diff
@@ -0,0 +1,45 @@
...
+limitations under the License.
+*/
+
+package testutil
+
+import (
+	"reflect"
+)
+
+func (t *T) CheckErrorAndTypeEquality(shouldErr bool, err error, expected, actual interface{}) {
```
tejal29: Very nice! There were times when i felt i shd check error type!
Thank you for adding this.

PR #2246, opened 2019-06-10
Mark `debug` as alpha

briandealwis:


balopat:
Note: please use forks instead of branches to avoid duplicated Travis builds (not sure why they are duped)

briandealwis:
Sorry, just used Github's edit feature on the files which created a branch since I now have write permission.  The Travis `push` build is for this branch, and the `pr` branch is for the merge to HEAD.  Forks just do the `pr` branch.

PR #2250, opened 2019-06-11
ability to `make integration` only on a chosen set of integration tests

balopat:


DEVELOPMENT.md
```diff
@@ -170,6 +170,14 @@ _These tests require push access to a project in GCP, and so can only be run
 by maintainers who have access. These tests will be kicked off by [reviewers](#reviews)
 for submitted PRs._
 
+### Running a chosen integration tests
+
+You can select a chosen integration tests via the `INTEGRATION_TEST_ARGS` env var:
```
priyawadhwa: nit: "select a chosen" is a bit unclear, might be rephrased as "You can select specific integration tests to run via the ..."

PR #2258, opened 2019-06-12
Fix remoteManifests

tanner-bruce:
This commit address issue #2160. Remote manifests in the config are
properly pulled again.

This also addresses another unmentioned issue: when using a remote
manifest with `skaffold dev`, the remote manifest will have its image
updated to include the new tag. When shutting down, the remote manifest
will be left the same, including the updated tag. This causes further
runs of `skaffold dev` to no longer match the image, and not update it
properly.

This is addressed by saving all of the images in the manifests found the
first time apply is ran, and then on cleanup replacing the images with
those initially found.

### DIFF
--- a/pkg/skaffold/deploy/kubectl.go
+++ b/pkg/skaffold/deploy/kubectl.go
@@ -144,7 +144,7 @@ func (k *KubectlDeployer) Cleanup(ctx context.Context, out io.Writer) error {
 	}
 
 	// pull remote manifests
-	var rm kubectl.ManifestList
+	var rm deploy.ManifestList
 	for _, m := range k.RemoteManifests {
 		manifest, err := k.readRemoteManifest(ctx, m)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


gbird3:
Would love to get this merged in as soon as possible. We want to add skaffold to our development workflow, but due to our current setup this is a requirement before it would make sense to move over.

dgageot:
@tanner-bruce any chance you could update this PR?

tanner-bruce:
@dgageot @nkubala I've rebased and added a test for cleaning up remote manifests, i.e reverting to original. I feel like there may be something wrong with it but was having some issues figuring out the testing framework. Let me know and I will make changes as necessary. Thanks!

tejal29:
@tanner-bruce i added another test case which you missed.
If the CI is green i will merge this in today.
Thanks for your contribution and we apologize for the delay

PR #2260, opened 2019-06-12
Add resources to Kaniko init container

### LINKED ISSUE #2159: Error creating kaniko pod because of resource quotas
I am facing this issue where my build using Kaniko w/ localDir fails. The issue is with quotas, however, I think while #1683 addresses part of the issue, the one I have is because my K8s cluster has mandatory resource quotas and in the Kaniko pod an init container is not defining any limits/requests.

### Expected behavior

Kaniko build starts

### Actual behavior

Build fails

```
FATA[0004] build failed: building [transpondertf/transponder]: kaniko build for [transpondertf/transponder]: creating kaniko pod: pods "kaniko-w6ntf" is forbidden: failed quota: compute-resources: must specify limits.cpu,limits.memory,requests.cpu,requests.memory
```

### Information

- Skaffold version: `v0.29.0`
- Operating system: Ubuntu 16.04 in WSL
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta10
kind: Config
build:
  artifacts:
  - image: transpondertf/transponder
    kaniko:
      buildContext:
        localDir: {}
      cache: {}
  cluster:
    pullSecretName: default-dockercfg-ps74r
    resources:
      requests:
        cpu: 100m
        memory: 64Mi
      limits:
        cpu: 1
        memory: 1Gi
deploy:
  kubectl:
    manifests:
      - deployments/*
```

1. Have a cluster with [Resource Quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/) enabled.
2. Use *localDir* Kaniko builder
3. `skaffold dev`

I think this is the offending init container breaking the pod creation:

https://github.com/GoogleContainerTools/skaffold/blob/0f2cb02b866be0311ee4d9539b7b2fb3da7ef5d7/pkg/skaffold/build/cluster/sources/localdir.go#L80-L92

cedrickring:
Fixes #2159 

I used fixed resource quotas for the init container as I don't think a while-loop needs configurable resources.


pkg/skaffold/build/cluster/sources/localdir.go
```diff
@@ -83,6 +88,16 @@ func (g *LocalDir) Pod(args []string) *v1.Pod {
 		Image:        g.artifact.BuildContext.LocalDir.InitImage,
 		Command:      []string{"sh", "-c", "while [ ! -f /tmp/complete ]; do sleep 1; done"},
 		VolumeMounts: []v1.VolumeMount{vm},
+		Resources: v1.ResourceRequirements{
```
priyawadhwa: Instead of setting new defaults and parsing them here,  WDYT of reusing the [resourceRequirements](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/cluster/sources/sources.go#L78) function? It seems reasonable to apply the same requirements specified in the skaffold config to each container.

cedrickring: I'm using the same resource requirements now 👍

pkg/skaffold/build/cluster/sources/localdir_test.go
```diff
@@ -101,5 +116,5 @@ func TestPod(t *testing.T) {
 		},
 	}
 
-	testutil.CheckDeepEqual(t, expectedPod, pod)
+	testutil.CheckDeepEqual(t, expectedPod, pod, cmpopts.IgnoreUnexported(resource.Quantity{}))
```
priyawadhwa: We typically use `reflect.DeepEqual` directly when there are unexported fields. Here's an [example](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/custom/custom_test.go#L122). This way we won't have to vendor in any new dependencies!

cedrickring: Alright. I was wondering why it wasn't vendored in the first place 😄

PR #2267, opened 2019-06-13
Docker is case sensitive about networks

tomaratyn:
host/bridge/none must be lower case or will be treated as network names.


dgageot:
Fixed with https://github.com/GoogleContainerTools/skaffold/pull/2288. Thanks @tomaratyn for getting this started.

PR #2270, opened 2019-06-14
Add K8sManagedBy function to labeller

priyawadhwa:
This function will be used by the port forwarder to parse out deployed services to automatically port forward. I've added the defaultLabeller to the runner so that the label can be passed in to the port forwarder in an upcoming PR.

### DIFF
--- a/pkg/skaffold/runner/labeller.go
+++ b/pkg/skaffold/runner/labeller.go
@@ -51,3 +51,7 @@ func (d *DefaultLabeller) Labels() map[string]string {
 		K8ManagedByLabel: fmt.Sprintf("skaffold-%s", version),
 	}
 }
+
+func (d *DefaultLabeller) K8sMangedByLabel() string {
+	return fmt.Sprintf("%s=skaffold-%s", K8ManagedByLabel, d.version)
+}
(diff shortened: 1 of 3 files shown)


pkg/skaffold/runner/labeller.go
```diff
@@ -51,3 +51,7 @@ func (d *DefaultLabeller) Labels() map[string]string {
 		K8ManagedByLabel: fmt.Sprintf("skaffold-%s", version),
 	}
 }
+
+func (d *DefaultLabeller) K8sMangedByLabel() string {
+	return fmt.Sprintf("%s=skaffold-%s", K8ManagedByLabel, d.version)
```
balopat: This took me a while to understand - can we rename/refactor some things to make it clearer? 
*  `K8ManagedByLabel`  -> `K8sManagedByLabelKey` 
* `K8sMangedByLabel()` -> `K8sManagedByLabelKeyValueString()`
* `fmt.Sprintf("skaffold-%s", version)` -> `skaffoldVersion()` and use that in `K8sManagedByLabelKeyValueString()`

priyawadhwa: for sure, done!

PR #2272, opened 2019-06-14
Add resourceType and resourceName to PortForward event

priyawadhwa:
Adds resourceType and resourceName to the PortForward event in skaffold.proto and regenerates the proto files.
Updates current port forwarding code to pass in this information for automatically forwarded pods.


priyawadhwa:
@balopat I added a unit test to make sure the correct number of events were sent during port forwarding -- I think making sure the port forward event itself is correct would require some sort of unit testing framework around eventing which would probably make more sense to do separately. WDYT?

balopat:
> @balopat I added a unit test to make sure the correct number of events were sent during port forwarding -- I think making sure the port forward event itself is correct would require some sort of unit testing framework around eventing which would probably make more sense to do separately. WDYT?

Hmmm...I think that we could start that mini framework small - just by setting the `event.PortForwarded` to a global var and in test replace it with a `mock` that captures the argument that we can check against then. On the long run we can implement a couple of these and then extract the functionality to a fully functional mock for the event package. 

I see more benefit in checking the actual events compared to just the number of events.

PR #2274, opened 2019-06-17
 Refactor skaffold init for more flexible builder detection

TadCordle:
This PR introduces an `InitBuilder` interface so that skaffold init can more easily detect non-Docker builders in the future. The interface allows us to:
- `Name()`: Get the name of the builder that was detected
- `Describe()`: Use a specialized prompt string in the interactive CLI while choosing builder/image pairs
- `CreateArtifact()`: Generate the corresponding build config artifact for the builder
- `GetConfiguredImage()`: Detect a builder's configured target image to skip choosing it from the CLI
- `GetPath()`: Get the path to the builder's config file

Follow-up for updating `--analyze`: #2327

Follow-up for Jib support: #2276


pkg/skaffold/docker/docker_init.go
```diff
@@ -0,0 +1,89 @@
...
+	return ""
+}
+
+// GetPath returns the path to the dockerfile
+func (d Dockerfile) GetPath() string {
+	return string(d)
+}
+
+// ValidateDockerfile makes sure the given Dockerfile is existing and valid.
+var ValidateDockerfile = func(path string) bool {
```
TadCordle: Note: this function was moved from `docker/validate.go`, to keep all the init-related docker stuff in one file. If it makes more sense to leave the `InitBuilder` implementation by itself, I can bring back the original file and move this back.

pkg/skaffold/docker/docker_init_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	"github.com/GoogleContainerTools/skaffold/testutil"
+)
+
+func TestValidateDockerfile(t *testing.T) {
```
TadCordle: Note: moved from `docker/validate_test.go`.

pkg/skaffold/initializer/init.go
```diff
@@ -150,94 +172,136 @@ func DoInit(out io.Writer, c Config) error {
 	return nil
 }
 
-func processCliArtifacts(artifacts []string) ([]dockerfilePair, error) {
-	var pairs []dockerfilePair
+// autoSelectBuilders takes a list of builders and images, checks if any of the builders' configured target
+// images match an image in the image list, and returns a list of the matching builder/image pairs. Also
+// returns the images from the original image list that didn't match any build configurations.
+func autoSelectBuilders(buildConfigs []InitBuilder, images []string) ([]BuilderImagePair, []string) {
```
TadCordle: Note: This function is only really useful for builders that can configure a target image themselves, so there will be a test for this function in the Jib follow-up PR, since Jib makes better use of this than Dockerfiles.

pkg/skaffold/initializer/init.go
```diff
@@ -261,16 +325,16 @@ func generateSkaffoldConfig(k Initializer, dockerfilePairs []dockerfilePair) ([]
 	return pipelineStr, nil
 }
 
-func printAnalyzeJSON(out io.Writer, skipBuild bool, dockerfiles, images []string) error {
-	if !skipBuild && len(dockerfiles) == 0 {
-		return errors.New("one or more valid Dockerfiles must be present to build images with skaffold; please provide at least one Dockerfile and try again or run `skaffold init --skip-build`")
+func printAnalyzeJSON(out io.Writer, skipBuild bool, buildConfigs []InitBuilder, images []string) error {
```
chanseokoh: Is this only for printing debugging output? Does changing the JSON output format not matter?

pkg/skaffold/initializer/init.go
```diff
@@ -88,22 +105,27 @@ func DoInit(out io.Writer, c Config) error {
...
 	// conditionally generate build artifacts
+	var pairs []BuilderImagePair
 	if !c.SkipBuild {
-		if len(dockerfiles) == 0 {
-			return errors.New("one or more valid Dockerfiles must be present to build images with skaffold; please provide at least one Dockerfile and try again or run `skaffold init --skip-build`")
+		if len(buildConfigs) == 0 {
+			return errors.New("one or more valid Dockerfiles must be present to build images with skaffold; please provide at least Dockerfile and try again or run `skaffold init --skip-build`")
 		}
 
+		var filteredImages []string
```
chanseokoh: `unresolvedImages`?

pkg/skaffold/initializer/init.go
```diff
@@ -150,94 +172,136 @@ func DoInit(out io.Writer, c Config) error {
...
 
-func processCliArtifacts(artifacts []string) ([]dockerfilePair, error) {
-	var pairs []dockerfilePair
+// autoSelectBuilders takes a list of builders and images, checks if any of the builders' configured target
+// images match an image in the image list, and returns a list of the matching builder/image pairs. Also
+// returns the images from the original image list that didn't match any build configurations.
+func autoSelectBuilders(buildConfigs []InitBuilder, images []string) ([]BuilderImagePair, []string) {
+	// Auto-select builders that have a definite target image
+	pairs := []BuilderImagePair{}
+	filteredImages := []string{}
```
chanseokoh: `unresolvedImages`?

pkg/skaffold/docker/docker_init.go
TadCordle: Ah, this is another thing that happens in the followup, I end up mocking this function in a test.

pkg/skaffold/initializer/init.go
```diff
@@ -88,22 +105,27 @@ func DoInit(out io.Writer, c Config) error {
...
 	}
-	var pairs []dockerfilePair
+
 	// conditionally generate build artifacts
+	var pairs []BuilderImagePair
 	if !c.SkipBuild {
-		if len(dockerfiles) == 0 {
-			return errors.New("one or more valid Dockerfiles must be present to build images with skaffold; please provide at least one Dockerfile and try again or run `skaffold init --skip-build`")
+		if len(buildConfigs) == 0 {
+			return errors.New("one or more valid Dockerfiles must be present to build images with skaffold; please provide at least Dockerfile and try again or run `skaffold init --skip-build`")
```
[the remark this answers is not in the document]
TadCordle: Yeah, already done in the followup.

pkg/skaffold/initializer/init.go
```diff
@@ -150,94 +172,136 @@ func DoInit(out io.Writer, c Config) error {
...
+			buildConfigs = append(buildConfigs[:matchingConfigIndex], buildConfigs[matchingConfigIndex+1:]...)
+		} else {
+			// No definite pair found, add to images list
+			filteredImages = append(filteredImages, image)
+		}
+	}
+	return pairs, filteredImages
+}
+
+func detectBuildFile(path string) ([]InitBuilder, error) {
```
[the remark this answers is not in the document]
TadCordle: Sorry, another followup thing. Since Jib can return multiple build configs from a single file (i.e. parent pom in a multimodule project) we need to return a list here.

pkg/skaffold/initializer/init.go
TadCordle: Hmm that's a good point, I'm not sure.

pkg/skaffold/docker/docker_init.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	"github.com/moby/buildkit/frontend/dockerfile/command"
+	"github.com/moby/buildkit/frontend/dockerfile/parser"
+	"github.com/sirupsen/logrus"
+)
+
+// Dockerfile is the path to a dockerfile. Implements the InitBuilder interface.
+type Dockerfile string
```
chanseokoh: I wonder if there is a better name, since `Dockerfile` is very general. I see in the follow-up PR the name is `Config` for the Jib InitBuilder, so that's a little bit inconsistent too.

TadCordle: We could call this a `Config` as well (so there's a `jib.Config` and a `docker.Config`), the reason I called this one `Dockerfile` is because it's an alias for `string` that just holds a path to a Dockerfile.

chanseokoh:
The logic looks good to me (except for the breaking schema change we talked about.)

Others with Go expertise will have a change to give feedback on style, naming, effective code, etc.

pkg/skaffold/initializer/init.go
```diff
@@ -150,94 +172,136 @@ func DoInit(out io.Writer, c Config) error {
...
+
+			// Found more than one match; can't auto-select.
+			if matchingConfigIndex != -1 {
+				matchingConfigIndex = -1
+				break
+			}
+			matchingConfigIndex = i
+		}
+
+		if matchingConfigIndex != -1 {
```
briandealwis: Doesn't this mean that initBuilders that don't have a matching image are never returned?

pkg/skaffold/initializer/init.go
```diff
@@ -150,94 +172,136 @@ func DoInit(out io.Writer, c Config) error {
...
+			ImageName: images[0],
 		}}
 	}
-	pairs := []dockerfilePair{}
+
+	// Build map from choice string to builder config struct
+	choices := make([]string, len(buildConfigs))
+	choiceMap := make(map[string]InitBuilder, len(buildConfigs))
+	for i, buildConfig := range buildConfigs {
+		choice := buildConfig.Describe()
```
briandealwis: Also worth mentioning on `Describe`'s godoc that the return result must be unique between artifacts

pkg/skaffold/initializer/init.go
```diff
@@ -150,94 +172,136 @@ func DoInit(out io.Writer, c Config) error {
...
+		choices[i] = choice
+		choiceMap[choice] = buildConfig
```
briandealwis: Should we try to sort the list?

pkg/skaffold/initializer/init.go
```diff
@@ -150,94 +172,136 @@ func DoInit(out io.Writer, c Config) error {
 	return nil
 }
 
-func processCliArtifacts(artifacts []string) ([]dockerfilePair, error) {
-	var pairs []dockerfilePair
+// autoSelectBuilders takes a list of builders and images, checks if any of the builders' configured target
+// images match an image in the image list, and returns a list of the matching builder/image pairs. Also
+// returns the images from the original image list that didn't match any build configurations.
```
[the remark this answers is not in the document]
TadCordle: I think at some point I was passing a pointer and modifying the original list, then forgot to return the filtered list when I changed it to pass a copy instead of a pointer, whoops.

pkg/skaffold/initializer/init.go
TadCordle: I changed this a bit so the new output is more consistent across builders/more easily digested by IDEs, but I'm going to double check with the vs code team since I think they use this.

pkg/skaffold/initializer/init.go
```diff
@@ -150,94 +177,136 @@ func DoInit(out io.Writer, c Config) error {
...
+	if len(builderConfigs) > 0 {
+		logrus.Warnf("unused builder configs found in repository: %v", builderConfigs)
 	}
 	return pairs
 }
 
-func promptUserForDockerfile(image string, dockerfiles []string) dockerfilePair {
-	var selectedDockerfile string
-	options := append(dockerfiles, NoDockerfile)
+var promptUserForBuildConfig = func(image string, choices []string) string {
```
priyawadhwa: Same comment as above!

TadCordle:
I think I'm going to revert the `--analyze` changes and make another PR for those, since I don't want this to get too much bigger.

chanseokoh:
The last commit to revert --analyze looks OK. That can be done later. Other code remains same, so LGTM for the overall logic.

pkg/skaffold/initializer/init.go
```diff
@@ -54,6 +59,22 @@ type Initializer interface {
...
 }
 
+// InitBuilder represents a builder that can be chosen by skaffold init.
+type InitBuilder interface {
+	// Name returns the name of the builder
+	Name() string
+	// Describe returns the initBuilder's string representation, used when prompting the user to choose a builder.
+	// Must be unique between artifacts.
+	Describe() string
+	// GetArtifact creates an Artifact to be included in the generated Build Config
```
dgageot: Should be `CreateArtifact`

pkg/skaffold/initializer/init.go
```diff
@@ -150,94 +182,136 @@ func DoInit(out io.Writer, c Config) error {
 	return nil
 }
 
-func processCliArtifacts(artifacts []string) ([]dockerfilePair, error) {
-	var pairs []dockerfilePair
+// autoSelectBuilders takes a list of builders and images, checks if any of the builders' configured target
+// images match an image in the image list, and returns a list of the matching builder/image pairs. Also
+// separately returns the builder configs and images that didn't have any matches.
+func autoSelectBuilders(builderConfigs []InitBuilder, images []string) ([]builderImagePair, []InitBuilder, []string) {
+	pairs := []builderImagePair{}
```
dgageot: var pairs []builderImagePair

pkg/skaffold/initializer/init.go
```diff
@@ -150,94 +182,136 @@ func DoInit(out io.Writer, c Config) error {
...
+	unresolvedImages := []string{}
```
dgageot: var unresolvedImages []string

PR #2275, opened 2019-06-17
adding tests for cluster builder

balopat:


pkg/skaffold/build/cluster/types_test.go
```diff
@@ -0,0 +1,122 @@
...
+	return &runcontext.RunContext{
+		InsecureRegistries: insecureRegistries,
+		Cfg: &latest.Pipeline{
+			Build: latest.BuildConfig{
+				BuildType: latest.BuildType{
+					Cluster: clusterDetails,
+				},
+			},
+		},
+		Opts: &config.SkaffoldOptions{
```
priyawadhwa: nit: we don't need to explicitly state any of these options

Issue #1264, opened 2018-11-09
skaffold init for Java applications


saturnism:
### Expected behavior
`skaffold init` on a Java project (with Maven or Gradle) will produce the default `skaffold.yaml` with the  corresponding Jib builder configuration.

### Actual behavior
`skaffold init` requires Dockerfile to be present and fails with:
`FATA[0001] one or more valid Dockerfiles must be present to run skaffold; please provide at least one Dockerfile and try again `

However, Dockerfile is not needed for Jib.

### Information

- Skaffold version: 0.17
- Operating system: OSX

### Steps to reproduce the behavior

1. Create a new Java project that uses either Maven or Gradle.
1. Run `skaffold init`

dgageot:
That's a duplicate of https://github.com/GoogleContainerTools/skaffold/issues/1093 but I'm going to keep your ticket because it's more detailed :-)

coollog:
Some options for when there might be a mix of Dockerfiles and Maven/Gradle.

- default to Dockerfile if present, choose between Dockerfiles if multiple
- default to Maven/Gradle if `pom.xml/mvnw`/`build.gradle/gradlew` detected, choose between Maven/Gradle if both present
- survey with choices for use Dockerfile, jibGradle, or jibMaven

I think we should go with the survey with all the choices since otherwise, it would be hard to differentiate which artifact should use which.

loosebazooka:
I like the survey idea.

Is skaffold init designed to only work for single artifacts at the moment? I imagine it would get pretty complex pretty fast if it started searching for one or more artifacts to configure.

dgageot:
This can be close, we are waiting for https://github.com/GoogleContainerTools/skaffold/issues/2842 to make this feature active.

PR #2276, opened 2019-06-17
Add Jib detection to skaffold init

TadCordle:
Review after #2327.

This PR adds Jib detection to `skaffold init`. Skaffold walks the directory searching for dockerfiles, `build.gradle`s, and `pom.xml`s. If a `build.gradle` or a `pom.xml` is found, skaffold will attempt to run an internal Jib gradle task/maven goal to determine if Jib is configured on any of the projects (Jib v1.3.0 or higher is required for this to work).

If Jib is detected and `jib.to.image` is configured with an image found in the k8s manifest, the Jib builder is added to the skaffold config automatically.

If Jib is detected but `jib.to.image` is not configured to build a k8s image, the Jib project is added to the CLI prompt so the user can decide which image Jib should build, and the user's choice is added using the `-Djib.to.image=...` flag.

Issues left to solve:
- Configuring to build with Jib via CLI artifacts
- Gradle/Maven edge cases (config files that don't use the default names, unorthodox project structure, etc.)
- Generating Jib config automatically if it's missing?

Towards https://github.com/GoogleContainerTools/skaffold/issues/1264 (probably not completely fixed until there's some idea how to automatically apply Jib to maven/gradle projects).


pkg/skaffold/initializer/init.go
```diff
@@ -221,7 +222,19 @@ func autoSelectBuilders(builderConfigs []InitBuilder, images []string) ([]builde
 	return pairs, builderConfigs, unresolvedImages
 }
 
-func detectBuilders(path string) ([]InitBuilder, error) {
+func detectBuilders(enableJibInit bool, path string) ([]InitBuilder, error) {
+	// TODO: Remove backwards compatibility if statement (not entire block)
+	if enableJibInit {
+		// Check for jib
+		if builders := jib.ValidateJibConfigFunc(path); builders != nil {
```
chanseokoh: Isn't `ValidateJibConfig` more natural, given that `...Func` is like an ad-hoc variable only used by testing?

pkg/skaffold/jib/jib_init.go
```diff
@@ -0,0 +1,165 @@
...
+	return fmt.Sprintf("%s (%s)", j.BuilderName, j.FilePath)
+}
+
+// CreateArtifact creates an Artifact to be included in the generated Build Config
+func (j Jib) CreateArtifact(manifestImage string) *latest.Artifact {
+	workspace := filepath.Dir(j.FilePath)
+
+	a := &latest.Artifact{ImageName: j.Image}
+	if j.Image == "" {
+		a.ImageName = manifestImage
```
chanseokoh: I forgot. So if a target image is configured in Jib (in `pom.xml` for example), then we can (or should) skip defining `image` in `skaffold.yaml`? This [repo example](https://github.com/GoogleContainerTools/skaffold/tree/master/examples) defines `image`s even if the images are configured in Jib.

pkg/skaffold/jib/jib_init.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Parse Jib output. Multiple JSON strings may be printed for multi-project/multi-module setups.
+	matches := regexp.MustCompile(`BEGIN JIB JSON\r?\n({.*})`).FindAllSubmatch(stdout, -1)
+	if len(matches) == 0 {
+		return nil
+	}
+
+	results := make([]Jib, len(matches))
+	for i, match := range matches {
+		line := bytes.Replace(match[1], []byte(`\`), []byte(`\\`), -1)
```
chanseokoh: What is this replacement for? Handling Windows path separators?

pkg/skaffold/jib/jib_init.go
```diff
@@ -0,0 +1,165 @@
...
+		parsedJSON := jibJSON{}
+		if err := json.Unmarshal(line, &parsedJSON); err != nil {
+			return nil
```
chanseokoh: This should usually not happen, but may happen if we change the JSON structure? Should we log this or print a warning?

pkg/skaffold/jib/jib_init.go
chanseokoh: Never mind. I see `ImageName` is initialized at first.

pkg/skaffold/jib/jib_init.go
TadCordle: Yup, I can add a comment.

pkg/skaffold/initializer/init.go
TadCordle: I'm having trouble finding an example now, but I think I noticed this was a convention used elsewhere in the codebase.

pkg/skaffold/jib/jib_init.go
chanseokoh: It's unlikely path names on non-Windows will have `\`, so I think this is acceptable. BTW, how do you do a test for Windows?

Issue #2294, opened 2019-06-19
write tests for protobuf generation 


balopat:
We should have tests to ensure that the Event API proto is in sync with the generated grpc code.
Also we should remove the protobufs from codecov.

dgageot:
Fixed with https://github.com/GoogleContainerTools/skaffold/pull/2419

PR #2316, opened 2019-06-21
Add scripts to test and generate files from proto

priyawadhwa:
This PR adds a Dockerfile which generates files from skaffold.proto and compares them to the existing files.

The dockerfile is used in `generate-proto.sh`, where generated files are created in a container and then copied into a local filesystem. It's also used in `test-generated-proto.sh`, which compares the local files to the generated ones.


balopat:
Fixes #2294.

hack/Dockerfile_proto
```diff
@@ -0,0 +1,43 @@
+# Install protoc
+FROM ubuntu AS protoc
```
balopat: can we pin the ubuntu version?

hack/Dockerfile_proto
```diff
@@ -0,0 +1,43 @@
+# Install protoc
```
balopat: let's move this under a dir in hack/proto/Dockerfile so that syntax highlighting can work

hack/Dockerfile_proto
priyawadhwa: yup!

hack/Dockerfile_proto
priyawadhwa: done

PR #2318, opened 2019-06-22
Logs: Show final container's status

dgageot:


dgageot:
This is buggy (I hate those logs)

Issue #2190, opened 2019-05-28
kaniko pull secret should be optional


prary:
Right now PullSecretName is created while building image using kaniko with buildcontext of localdir.

### Expected behavior
When build context is set to localdir, then there is no need to create a pull secret.
### Actual behavior
Secret is getting created
### Information

- Skaffold version: 30
- Operating system: linux
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta11
kind: Config
build:
  artifacts:
# denotes the user’s image push location in artifactory, ensure user write permission in artifactory, this should be the same location referenced in your respective deployment yml
  - image: www.someurl.com/docker-sandbox/kube/pythonserver
    context: .
    kaniko:
      buildContext:
        localDir: {}
  cluster:
    namespace: 103000-pras-dev
#    pullSecretName: regcred
    dockerConfig:
      path: ~/.docker/config.json
deploy:
  kubectl:
    manifests:
#must match the kubernetes deploy file name of application
      - deploy.yml
```

### Steps to reproduce the behavior

1. Run the skaffold wiht kaniko builder and build context equal to localdir

jimanvlad:
+1 

` - name: oncluster
    build:
      artifacts:
        - image: artifactory.local:5000/data/image
          context: src
          kaniko:
            buildContext:
              localDir: {}
            cache: {}`

`time="2019-08-16T14:22:12+01:00" level=fatal msg="failed to build: build failed: build failed: setting up pull secret: checking for existing kaniko secret: secrets \"kaniko-secret\" not found"
`

priyawadhwa:
Hey @prary -- so this field is definitely a bit misnamed. `pullSecretName` is used to pull base images and buildcontexts from GCS, but also to push the final image to a registry. Looks like you're using the docker config for that instead?

Perhaps we should change the logic to require either the `pullSecret` or the `dockerConfigSecret` be specified. If insecure registries are specified, then we don't require either. WDYT?

jimanvlad:
Agree that pullSecret is only needed for GCS (but naming is not very clear). AFAIK, dockerConfigSecret (or dockerConfigPath) have a completely different purpose, which is to authenticate to the artifact registry. I don't think "insecure registries" and this should be linked either because the former should just bypass the tls verification on the registry, but the latter may still be required for the registry credentials.

priyawadhwa:
[...]

PR #2321, opened 2019-06-22
Kaniko secret

prary:
kaniko pull secret should be optional
Resolving issue #2190


prary:
This is how it should be right? 
```
func setupPullSecret() {
	if b.PullSecret == "" {
		return func() {}, nil
        }
        rest of the secret creation code
}
```

dgageot:
@prary could you add a unit test for the new code?

prary:
> @prary could you add a unit test for the new code?

Sure I will add

tejal29:
@prary, Here is an example of creating a mock client and adding secrets

https://github.com/GoogleContainerTools/skaffold/blob/6ed64d6d7ee85aa5538c85eb9ba61df20bac5795/pkg/skaffold/deploy/status_check_test.go#L180
```
// Create Secrets	
		objs := make([]runtime.Object, len(test.secrets))
			for i, s := range test.secrets {
				objs[i] = dep
			}
// Create a fake client with Secrets added
			client := fakekubeclientset.NewSimpleClientset(objs...)
// Pass the fake client to the method using it

actual, err := getDeployments(client, "test", labeller, time.Duration(200)*time.Second)

```

tejal29:
In your case, you need to mock the function call to `kubernetes.GetClientset()` to actually return the fakeclient.

tejal29:
I think @dgageot covered this https://github.com/GoogleContainerTools/skaffold/pull/2321

PR #2323, opened 2019-06-23
Improve the logs

dgageot:
+ Don’t leak go routines 
 + Never reprint the log for a given container
 + Simplify code
 + Handle init containers first 
 + Notify that container logs ended


dgageot:
@balopat I simplified my PR because it was buggy. I decided to stop printing the container's final state because it's too unreliable. Printing when container logs end is, if think, good enough.

PR #2347, opened 2019-06-26
Restart the dev loop when the skaffold config changes

dgageot:
I thought I fixed that recently but it turns out I
haven’t!

Move default labeller to deploy since it used in deployer.


MarkBennett:
Great changeset! I've bumped into this when getting started with skaffold and seeing it fixed would definitely help out with the ergonomics when you're setting up. Thanks @dgageot! :)

PR #2361, opened 2019-06-27
Schemas: Make sure preferredOrder is in sync with field order in Go structs

### LINKED ISSUE #2003: profiles should be shown last on https://skaffold.dev/docs/references/yaml/
I tried to make sure the `skaffold.yaml` reference (https://skaffold.dev/docs/references/yaml/) shows the most used configuration sections first. However, I should have written a test for that because the order has changed.

It used to show sections in that order `build`, `test `, `deploy`, `profiles`.
It's now `profiles`, `build`, `test `, `deploy`.

Maybe the order of other sub sections has changed too

dgageot:
Fields from inlined definitions were always at the end.

Fixes #2003

### DIFF
--- a/docs/content/en/schemas/v1beta10.json
+++ b/docs/content/en/schemas/v1beta10.json
@@ -1509,11 +1509,11 @@
       },
       "preferredOrder": [
         "name",
-        "patches",
-        "activation",
         "build",
         "test",
-        "deploy"
+        "deploy",
+        "patches",
+        "activation"
       ],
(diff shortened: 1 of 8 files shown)


balopat:
LGTM! it would be cool to cover this ordering in tests though...I added tests on the schema generator a while ago, it shouldn't be too hard to extend them in theory :)

dgageot:
@balopat I've updated a test

PR #2368, opened 2019-06-28
Try Buildkit on TravisCI

dgageot:
Let’s see if it’s supported and if images are built
faster.


dgageot:
Looks like it's not supported

PR #2384, opened 2019-07-01
Design proposal for configurable kubecontext

corneliusweig:
This is a design proposal for making the kubecontext configurable. The goal here is to lay out different options with their tradeoffs and come to an agreement what is the best way to proceed. I may have missed better alternatives, so if you find better ones, please leave a comment.
I deliberately only present different options, so that the discussion starts unbiased.

### DIFF
--- a/docs/design_proposals/configurable-kubecontext.md
+++ b/docs/design_proposals/configurable-kubecontext.md
@@ -1,96 +1,181 @@
-# Title
+# Configurable kubecontext
 
-* Author(s): \<your name\>
-* Design Shepherd: \<skaffold-core-team-member\>
-
-    If you are already working with someone mention their name.
-    If not, please leave this empty, it will be assigned to a core team member.
-* Date: \<date\>
+* Author(s): Cornelius Weig (@corneliusweig)
+* Design Shepherd: tbd
+* Date: 29 June 2019
 * Status: [Reviewed/Cancelled/Under implementation/Complete]
-
-Here is a brief explanation of the Statuses
-
-1. Reviewed: The proposal PR has been accepted, merged and ready for
-   implementation.
-2. Under implementation: An accepted proposal is being implemented by actual work.
-   Note: The design might change in this phase based on issues during
-   implementation.
-3. Cancelled: During or before implementation the proposal was cancelled.
-   It could be due to:
-   * other features added which made the current design proposal obsolete.
-   * No longer a priority.
-4. Complete: This feature/change is implemented.
 
 ## Background
 
-In this section, please mention and describe the new feature, redesign
-or refactor.
+So far, Skaffold always uses the currently active kubecontext when interacting with a kubernetes cluster.
+This is problematic when users want to deploy multiple projects with different kubecontexts, because the user needs to manually switch the context before starting Skaffold.
+In particular when working on multiple such projects in parallel, the current behavior is limiting.
 
-Please provide a brief explanation for the following questions:
+Open issues concerning this problem are
 
-1. Why is this required?
-2. If this is a redesign, what are the drawbacks of the current implementation?
-3. Is there any another workaround, and if so, what are its drawbacks?
-4. Mention related issues, if there are any.
+- Allow option to specify the kubectl context (#511)
+- Support kube.config and kube.context for specifying alternative Kubernetes config file or context (#2325)
+- Feature: Support regex in profile activation via kubeContext (#1677)
+- Skaffold.yaml files are not portable (#480)
 
-Here is an example snippet for a new feature:
+There also was an attempt to add a configuration option to `skaffold.yaml` (Support for overriding kubectl context during deployment #1540).
 
-___
-Currently, skaffold config supports `artifact.sync` as a way to sync files
-directly to pods. So far, artifact sync requires a specification of sync
-patterns like
-
-```yaml
-sync:
-  '*.js': app/
-```
-
-This is error prone and unnecessarily hard to use, because the destination is
-already contained in the Dockerfile for docker build (see #1166, #1581).
-In addition, the syncing needs to handle special cases for globbing and often
-requires a long list of sync patterns (#1807).
-
-___
+The goal of this document is to create an agreement on what options should be supported and identify edge cases.
 
 ## Design
 
-Please describe your solution. Please list any:
+There are four places where kubecontext activation can be added:
+<table>
+    <thead>
+        <th>Precedence</th> <th>Kind</th> <th>Details</th>
+    </thead>
+    <tbody>
+        <tr>
+            <td>1. (highest)</td>
+            <td>CLI option</td>
+            <td>
+              The kubernetes standard to set the kubecontext is <code>--context</code>.
+              However, in Skaffold this term is so overloaded that it should more precisely be named <code>--kube-context</code>.
+              This flag is necessary for IDE integration.
+            </td>
+        </tr>
+        <tr>
+            <td>2.</td>
+            <td>Env variable</td>
+            <td>
+              <code>SKAFFOLD_KUBE_CONTEXT</code>, similar to other Skaffold flags.
+            </td>
+        </tr>
+        <tr>
+            <td>3.</td>
+            <td><code>skaffold.yaml</code></td>
+            <td>
+              Json-path <code>deploy.kubeContext</code>.
+              This option is shareable, but also creates a lot of complexity due to profile activation (see detailed discussion below).
+            </td>
+        </tr>
+        <tr>
+            <td>4. (lowest)</td>
+            <td>Global Skaffold config</td>
+            <td>
+              This should give users the possibility to define a default context globally or per project.
+              This variant is not shareable.
+            </td>
+        </tr>
+    </tbody>
+</table>
 
-* new config changes
-* interface changes
-* design assumptions
+---
 
-For a new config change, please mention:
+Beside the kubecontext, also the namespace needs to be specified.
+It should also be possible to specify in various places:
 
-* Is it backwards compatible? If not, what is the deprecation policy?
-  Refer to the [deprecation policy requirements.](./../../deprecation-policy.md#how-do-we-deprecate-things)
-  for details.
+1. `--namespace` CLI flag
+2. `SKAFFOLD_NAMESPACE` env var
+3. In `skaffold.yaml` at json-path `deploy.namespace` 
+4. globally or per project in the Skaffold config
+
+### Detailed discussion
+#### Option in `skaffold.yaml`
+A configuration option in `skaffold.yaml` has the advantage of being most discoverable:
+it is in the place where users configure all aspects of their pipeline.
+
+However, it also has questionable implications:
+
+- `skaffold.yaml` is meant to be shared, but kubecontext names vary across users.
+  Sharing therefore makes only sense in a corporate setting where context names are the same across many users.
+  There is however a risk of abuse in settings where sharing the context name does not make sense, for example in open source projects.
+- Due to profile activation by Skaffold profiles, there can be circular problems.
+  For example, the current context is `A` and activates some profile.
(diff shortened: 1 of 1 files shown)


docs/design_proposals/configurable-kubecontext.md
```diff
@@ -0,0 +1,181 @@
+# Configurable kubecontext
+
+* Author(s): Cornelius Weig (@corneliusweig)
+* Design Shepherd: tbd
```
balopat:
```suggestion
* Design Shepherd: Balint Pato (@balopat)
```

docs/design_proposals/configurable-kubecontext.md
```diff
@@ -0,0 +1,183 @@
...
+            <td>Env variable</td>
+            <td>
+              <code>SKAFFOLD_KUBE_CONTEXT</code>, similar to other Skaffold flags.
+            </td>
+        </tr>
+        <tr>
+            <td>3.</td>
+            <td><code>skaffold.yaml</code></td>
+            <td>
+              Json-path <code>deploy.kubeContext</code>.
```
balopat:
```suggestion
              Json-path <code>[profile.]deploy.kubeContext</code>.
```

docs/design_proposals/configurable-kubecontext.md
```diff
@@ -0,0 +1,183 @@
...
+A configuration option in `skaffold.yaml` has the advantage of being most discoverable:
+it is in the place where users configure all aspects of their pipeline.
+In addition, it allows to define the kubecontext per Skaffold profile.
+
+However, it also has questionable implications:
+
+- `skaffold.yaml` is meant to be shared, but kubecontext names vary across users.
+  Sharing therefore makes only sense in a corporate setting where context names are the same across many users.
+  There is however a risk of abuse in settings where sharing the context name does not make sense, for example in open source projects.
+- Due to profile activation by Skaffold profiles, there can be circular problems.
```
balopat:
```suggestion
- Due to profile activation by Skaffold profiles, there can be confusing/surprising situations.
```

This is not really a circular problem as we don't change the "current context".

docs/design_proposals/configurable-kubecontext.md
```diff
@@ -0,0 +1,183 @@
...
+  For example, the current context is `A` and activates some profile.
+  This profile activates a different kubecontext `B`.
```
balopat:
```suggestion
  This profile is deploying to a different kubecontext `B`. This can be surprising to the user.
```

docs/design_proposals/configurable-kubecontext.md
```diff
@@ -0,0 +1,183 @@
...
+it is in the place where users configure all aspects of their pipeline.
+In addition, it allows to define the kubecontext per Skaffold profile.
+
+However, it also has questionable implications:
+
+- `skaffold.yaml` is meant to be shared, but kubecontext names vary across users.
+  Sharing therefore makes only sense in a corporate setting where context names are the same across many users.
+  There is however a risk of abuse in settings where sharing the context name does not make sense, for example in open source projects.
+- Due to profile activation by Skaffold profiles, there can be circular problems.
+  For example, the current context is `A` and activates some profile.
```
balopat:
```suggestion
  For example, the current context is `minikube` and activates some profile.
```

docs/design_proposals/configurable-kubecontext.md
```diff
@@ -0,0 +1,183 @@
...
+  This profile activates a different kubecontext `B`.
+  A solution could be to forbid specifying a kubecontext in a profile if it is activated by a kubecontext (and validate that).
```
balopat: This should be the recommended solution. 
We should mention that this check should on the "effective profile"  that is the result of merging all the activated profiles.

docs/design_proposals/configurable-kubecontext.md
```diff
@@ -0,0 +1,183 @@
...
+1. How to identify projects
+2. How to save the project/kubecontext relation
+
+##### How to identify projects
+
+There are at least two possibilities:
+
+- Identify projects by their absolute host path.
+  This is guaranteed to be unique, but may break if a user moves his project to another location.
+- Identify projects by a new `metadata.name` entry in `skaffold.yaml` (see also #2200).
```
balopat:
```suggestion
- Identify projects by a new `metadata.name` entry in `skaffold.yaml` (see also [#2200](https://github.com/GoogleContainerTools/skaffold/issues/2200)).
```

docs/design_proposals/configurable-kubecontext.md
corneliusweig: My current feeling is that we should have the option also in the main pipeline. Although this may not be the recommended way to use this feature, I have a hunch that this will be requested nevertheless.

docs/design_proposals/configurable-kubecontext.md
corneliusweig: I'm a bit puzzled about the effective profile. If we forbid to override the kubecontext twice by two profiles, then we are deviating from the normal behavior of how profiles are usually applied. Usually profiles are applied one after the other, and last one wins. So why should that be different for the kubecontext?

I think a static check where a profile with activation by kubecontext must not override the kubecontext is easy to understand and fine. However, as soon as the logic becomes dynamic this quickly becomes rather intransparent.

docs/design_proposals/configurable-kubecontext.md
balopat: agreed, `[profile.]` meant to signify main pipeline or optionally in a profile...maybe there is a better way to express that :)

docs/design_proposals/configurable-kubecontext.md
balopat: Hm...but then 

given: current context = minikube 
activated profiles: 

P1 <- envvar1
P2 <- cur_ctx=minikube  

and P1.deploy.kubecontext = gke_foobar

I'll end up with the minikube specific profile that deploys to gke...

corneliusweig: I would say

given: current context = minikube

P1 <- envvar1,   changes the kubeContext to `gke_foobar`
P2 _no longer applies_, because kubeContext is not `minikube` anymore

---

I just checked the `skaffold.yaml` and the profile activation by kube-context is simply called `kubeContext`. It does not necessarily mean that it only considers the current context as defined in the kubeconfig. That would even be very misleading: when somebody changes the kube-context on the CLI with `skaffold --kube-context gke_foobar`, but the kubeconfig says `minikube`, I would certainly not expect that Skaffold activates a minikube profile.

corneliusweig: I think profile activation is confusing, and one can construct pretty contrived examples:

given: current context = `minikube`
P1 <- envvar, changes kubeContext to `gke_foobar`
P2 <- activated by kubeContext `gke_foobar`
P3 <- activated by envvar, changes kubeContext to `minikube`

In that sequence, the gke-specific profile would be activated on minikube. :zap: 

---

This example suggests that a dynamic check is necessary in any case. I think the validation should look out for profiles that were activated by a kube-context which is no longer active in the effective `skaffold.yaml` (Pardon me, if that is what you were trying to tell me all along).

corneliusweig: Never mind my previous comments. I should have read the code first.

So profile activation and profile application are separate, meaning that all profile activations use the same current kubecontext. Iow there can be no transitive kubecontext activation like
```
P1 activates minikube
P2 is activated by minikube
...and so on
```
A reasonable validation is therefore:
- check if any profile is activated by the current kubecontext before profile activation
- after profile application, check if the effective kubecontext `skaffold.yaml` has changed.
- if yes, and there was a context-activated profile, return error
- otherwise, everything is fine.
(Pardon me, if this is what you wanted to tell me all along..)
This check also makes any static check of the `skaffold.yaml` obsolete.

The case where subsequent profiles repeatedly switch the kubecontext is probably fine, as profile activation follows a last-one-wins policy which should also apply to kubecontext.

docs/design_proposals/configurable-kubecontext.md
```diff
@@ -0,0 +1,183 @@
...
+
+Currently, the Skaffold config uses the kubecontext as identifier.
+
+There are two possibilities to add the relation:
+
+- Reverse the mapping project/kubecontext and save as list under `kube-context` entries:
+  ```yaml
+  kubecontexts:
+  - kube-context: my-context
+    projects:
```
corneliusweig: @balopat Please correct me if I'm wrong, but my impression was that this variant was the preferred one.

Also, we should carefully think about how to name the new field, as the global config is not versioned and field renaming will be much harder. @tejal29 confused it with GCB projects, and this may happen to other users too. We could make it clearer that this ties the kubecontext to a specific named `skaffold.yaml`, and call the field `skaffoldConfigs` instead.

docs/design_proposals/configurable-kubecontext.md
```diff
@@ -0,0 +1,183 @@
...
+
+**\<What Skaffold config structure has the best tradeoffs?\>**
+
+Resolution: __Not Yet Resolved__
+
+### Open Issues/Questions
+
+**\<Should there be a config option in `skaffold.yaml`?\>**
+
+Resolution: Yes.
```
corneliusweig: I think the overall agreement was that we do want to have the option in `skaffold.yaml`

docs/design_proposals/configurable-kubecontext.md
balopat: This sounds good, yes, this is exactly what I meant :)

docs/design_proposals/configurable-kubecontext.md
balopat: +1 on `skaffoldConfigs`

docs/design_proposals/configurable-kubecontext.md
balopat: Yes, confirmed!

docs/design_proposals/configurable-kubecontext.md
```diff
@@ -0,0 +1,201 @@
...
+    - my-project
+  - kube-context: context2
+    skaffoldConfigs:
+    - my-project
+  ```
+- Add a new top-level entry in Skaffold config:
+  ```yaml
+  global: {}
+  kubecontexts: []
+  projects:
```
corneliusweig: Should be 
```yaml
skaffoldConfigs:
  my-project: my-context
  my-other-project: my-other-context
```

docs/design_proposals/configurable-kubecontext.md
```diff
@@ -0,0 +1,201 @@
...
+    my-project: # project name as the key
+      kube-context: context1
+  ```
+  This option will be more complex to implement wrt `skaffold config`.
+
+### Open Issues/Questions
+
+**\<What Skaffold config structure has the best tradeoffs?\>**
+
+Resolution: __Not Yet Resolved__
```
corneliusweig: @balopat Let's resolve this last open question, so that this design proposal can be merged.

My feeling is that option 2 (map-like structure) is overall better suited to store project-to-context mappings. Besides, option 1 (`skaffoldConfigs` list under kubecontext entry) requires additional special cases for the `skaffold config set/unset` commands. So I'm more inclined towards the second option:
```yaml
skaffoldConfigs:
  my-project: my-context
  my-other-project: my-other-context
```
WDYT?

corneliusweig: I just found another advantage of option 2: It allows a much more seamless implementation of the wildcard global default context like so:
```yaml
skaffoldConfigs:
  '*': default-context
  my-project: my-context
```

balopat: I'm okay with this direction!

corneliusweig:
@balopat I have included the latest design decisions into this document. From my side, it's ready to be merged now.

tejal29:
@balopat This looks good to me now.

PR #2390, opened 2019-07-02
Jib Builder: add more tests

dgageot:

### DIFF
--- a/pkg/skaffold/build/local/local.go
+++ b/pkg/skaffold/build/local/local.go
@@ -35,6 +35,11 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// For testing
+var (
+	getRemoteDigest = docker.RemoteDigest
+)
+
 // Build runs a docker build on the host and tags the resulting image with
 // its checksum. It streams build progress to the writer argument.
 func (b *Builder) Build(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact) ([]build.Artifact, error) {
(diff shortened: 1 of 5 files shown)


pkg/skaffold/build/local/local.go
```diff
@@ -35,6 +35,11 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// For testing
+var (
```
loosebazooka: I don't really understand this specific part? Does this just become a variable in the scope of the local pkg? Was there some problem with using the direct reference (`docker.RemoteDigest`) in the jib files?

dgageot: This is the simplest way to mock the call to `docker.RemoteDigest`. We use that pattern in. a lot of places.

dgageot:
@loosebazooka Looks like the changes to the owners file are still not enough

TadCordle:
Might be because you made a change to `local.go`?

loosebazooka:
@dgageot as a followup perhaps change all of the `docker.RemoteDigest` -> `getRemoteDigest` ?
```
build/local/bazel.go:   return docker.RemoteDigest(tag, insecureRegistries)
build/local/custom.go:          return docker.RemoteDigest(tag, b.insecureRegistries)
build/cache/save.go:            return docker.RemoteDigest(img, c.insecureRegistries)
build/cache/save.go:            return docker.RemoteDigest(img, c.insecureRegistries)
build/cache/cache.go:   remoteDigest    = docker.RemoteDigest
build/cluster/kaniko.go:        return docker.RemoteDigest(tag, b.insecureRegistries)
build/cluster/cluster.go:       return docker.RemoteDigest(tag, b.insecureRegistries)
```

dgageot:
Well, let’s do it only when a test requires it.

Issue #2398, opened 2019-07-03
TestDebug is too long and flaky


dgageot:
Here are the integration tests durations on a random PR:
```
--- PASS: TestBuild (5.55s)
--- PASS: TestBuild/dateTime_tagger (0.09s)
--- PASS: TestBuild/docker_build (5.14s)
--- PASS: TestBuild/envTemplate_tagger (0.10s)
--- PASS: TestBuild/git_tagger (0.12s)
--- PASS: TestBuild/sha256_tagger (0.10s)
--- PASS: TestBuildDeploy (15.70s)
--- PASS: TestConfigListForAll (0.02s)
--- PASS: TestConfigListForContext (0.02s)
--- PASS: TestDebug (267.99s)
--- PASS: TestDebug/kubectl (212.36s)
--- PASS: TestDebug/kustomize (55.63s)
--- PASS: TestDeploy (0.81s)
--- PASS: TestDev (5.63s)
--- PASS: TestDevSync (5.94s)
--- PASS: TestDiagnose (30.20s)
--- PASS: TestDiagnose/bazel_builder (16.00s)
--- PASS: TestDiagnose/docker_builder (0.81s)
--- PASS: TestDiagnose/jib_maven_builder (13.36s)
--- PASS: TestDiagnose/kaniko_builder (0.03s)
--- PASS: TestEventLogHTTP (1.23s)
--- PASS: TestEventLogRPC (1.19s)
--- PASS: TestFailToSetUnrecognizedValue (0.02s)
--- PASS: TestFix (16.57s)
--- PASS: TestGetStateHTTP (3.15s)
--- PASS: TestGetStateRPC (2.19s)
--- PASS: TestInit (10.60s)
--- PASS: TestInit/compose (2.16s)
--- PASS: TestInit/getting-started (1.24s)
--- PASS: TestInit/microservices (7.20s)
--- PASS: TestRun (122.71s)
--- PASS: TestRun/bazel (24.69s)
--- PASS: TestRun/custom_builder (35.23s)
--- PASS: TestRun/envTagger (16.53s)
--- PASS: TestRun/getting-started (12.72s)
--- PASS: TestRun/microservices (3.52s)
--- PASS: TestRun/nodejs (19.14s)
--- PASS: TestRun/structure-tests (10.86s)
--- PASS: TestSetDefaultRepoForContext (0.05s)
--- PASS: TestSetGlobalDefaultRepo (0.04s)
```

[...]

PR #2399, opened 2019-07-03
Simplify test debug

dgageot:
+ Remove code that's not required
+ Use base images that are already used elsewhere. It might help reduce the number of images we pull

See #2398

### DIFF
--- a/integration/debug_test.go
+++ b/integration/debug_test.go
@@ -33,11 +33,9 @@ func TestDebug(t *testing.T) {
 	tests := []struct {
 		description string
 		dir         string
-		filename    string
 		args        []string
 		deployments []string
 		pods        []string
-		env         []string
 	}{
 		{
 			description: "kubectl",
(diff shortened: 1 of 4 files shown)


integration/debug_test.go
```diff
@@ -55,23 +53,24 @@ func TestDebug(t *testing.T) {
...
+
 			ns, client, deleteNs := SetupNamespace(t)
 			defer deleteNs()
 
-			stop := skaffold.Debug(test.args...).WithConfig(test.filename).InDir(test.dir).InNs(ns.Name).WithEnv(test.env).RunBackground(t)
+			stop := skaffold.Debug(test.args...).InDir(test.dir).InNs(ns.Name).RunBackground(t)
 			defer stop()
 
 			client.WaitForPodsReady(test.pods...)
-			client.WaitForDeploymentsToStabilize(test.deployments...)
```
briandealwis: The `jib` example is a deployment.

integration/debug_test.go
```diff
@@ -55,23 +53,24 @@ func TestDebug(t *testing.T) {
...
 			for _, depName := range test.deployments {
```
briandealwis: Hmm this should be testing the pods too.  I'll fix that separately.

integration/debug_test.go
dgageot: Not sure I understand your point

briandealwis: By deleting this line, the deployments may not be stabilized before the test continues. Right?

dgageot: `GetDeployment` does it

dgageot:
BEFORE:

```
--- PASS: TestDebug/kubectl (212.36s)
--- PASS: TestDebug/kustomize (55.63s)
```

AFTER:

```
--- PASS: TestDebug/kubectl (83.07s)
--- PASS: TestDebug/kustomize (23.91s)
```

Issue #2360, opened 2019-06-27
sync to compeleted job fails, terminating the rest of the sync


MarkBennett:
### Expected behavior

Given I've got a few pods running in my cluster and using the same build (such as Rails webserver, worker, and rspec job), then when I make a change to synced file skaffold to restart the job rather than try to sync files to a completed/terminated pod. Alternatively, as a less ideal outcome skaffold could ignore the completed job and not do anything.

### Actual behavior

Currently, if I make a change to a synced file then skaffold will try to sync it to all pods using the build, including jobs which have completed. When trying to exec into a completed pod this error is generated:

```
WARN[2020] Skipping deploy due to sync error: copying files: Running [kubectl exec dev-run-specs-kfqfs --namespace default -c run-specs -i -- tar xmf - -C / --no-same-owner]: stdout , stderr: error: canno
t exec into a container in a completed pod; current phase is Succeeded
, err: exit status 1: exit status 1
```

### Information

- Skaffold version: v0.31.0
- Operating system: Mac OS 10.14.5
- Cluster: Minikube v1.1.1
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta11
kind: Config
build:
  artifacts:
    - image:  gcr.io/litdistco-pro/litdistco-rails
      docker:
        cacheFrom:
          - gcr.io/litdistco-pro/litdistco-rails 
  googleCloudBuild:
    projectId: litdistco-pro
deploy:
  kustomize:
    path: k8s/overlays/production
profiles:
  - name: dev
    build:
      artifacts:
      - image:  gcr.io/litdistco-pro/litdistco-rails
        docker:
          target: developer
        sync:
          manual:
            - src: 'app/**'
              dest: /project
      local: {}
    deploy:
      kustomize:
        path: k8s/overlays/development
    activation:
      - command: dev
  - name: staging
    build:
      artifacts:
      - image:  gcr.io/litdistco-pro/litdistco-rails
        docker:
          target: developer
      local: {}
    deploy:
      kustomize:
        path: k8s/overlays/staging
```

### Steps to reproduce the behavior

1. Create `skaffold.yaml` which uses `build.artificats.sync` and deploys to both a pod and a job.
2. Start development with `skaffold dev`
3. Change a synced file.
4. `skaffold dev` fails to sync changed files to pods as running `exec` in the terminated job seems to abort the entire sync.

[...]

PR #2403, opened 2019-07-04
Feat(sync): skip sync on non-running pods

seblegall:
Fix issue #2360

The goal of this PR is to avoid trying to sync files in a pod which is
not in a running state. It should avoid errors when working on a single
artifact which is deployed in different pods. For example : one running
a webserver and another running a work or a job. This is typically
something that could happend when a microservice is able to run both web
server and a job which make some sort of batch treatments on the database.

### DIFF
--- a/pkg/skaffold/sync/sync.go
+++ b/pkg/skaffold/sync/sync.go
@@ -169,6 +169,12 @@ func Perform(ctx context.Context, image string, files syncMap, cmdFn func(contex
 		}
 
 		for _, p := range pods.Items {
+
+			if p.Status.Phase != v1.PodRunning {
+				logrus.Infof("Skipping sync with pod %s because it's not running", p.Name)
+				continue
+			}
+
 			for _, c := range p.Spec.Containers {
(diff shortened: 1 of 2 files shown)


pkg/skaffold/sync/sync.go
```diff
@@ -169,6 +169,12 @@ func Perform(ctx context.Context, image string, files syncMap, cmdFn func(contex
 		}
 
 		for _, p := range pods.Items {
+
+			if p.Status.Phase != v1.PodRunning {
+				logrus.Infof("Skipping sync with pods %s because it's not running", p.Name)
```
dgageot: s/pods/pod

dgageot:
@seblegall When https://github.com/GoogleContainerTools/skaffold/pull/2413 is merged, you can rebase and the test suite should be all green

seblegall:
ok, great.

dgageot:
@seblegall you should be able to rebase on top of the now better test suite!

PR #2406, opened 2019-07-04
Better integration tests

dgageot:
+ Switch to latest TravisCI env (xenial) because it seems faster and has more disk space for integration tests
+ Print more information when an integration test goes wrong
+ Re-enable TestDev that wasn't flaky
+ Increase the duration of the test suite to increase the chance of running all the tests when one goes wrong.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,6 +109,7 @@ ifeq ($(GCP_ONLY),true)
 		--zone $(GKE_ZONE) \
 		--project $(GCP_PROJECT)
 endif
+	kubectl get nodes -oyaml
 	GCP_ONLY=$(GCP_ONLY) go test -v $(REPOPATH)/integration -timeout 15m $(INTEGRATION_TEST_ARGS)
 
 .PHONY: release
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ ifeq ($(GCP_ONLY),true)
 		--project $(GCP_PROJECT)
 endif
 	kubectl get nodes -oyaml
-	GCP_ONLY=$(GCP_ONLY) go test -v $(REPOPATH)/integration -timeout 15m $(INTEGRATION_TEST_ARGS)
+	GCP_ONLY=$(GCP_ONLY) go test -v $(REPOPATH)/integration -timeout 20m $(INTEGRATION_TEST_ARGS)
 
 .PHONY: release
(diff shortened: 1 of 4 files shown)


dgageot:
Worked 1 time

dgageot:
@corneliusweig would you mind taking a look at this PR? I'm trying to fix all known flaky tests

corneliusweig:
Sure, I can do that.

Makefile
```diff
@@ -109,27 +109,28 @@ ifeq ($(GCP_ONLY),true)
 		--zone $(GKE_ZONE) \
 		--project $(GCP_PROJECT)
 endif
-	GCP_ONLY=$(GCP_ONLY) go test -v $(REPOPATH)/integration -timeout 15m $(INTEGRATION_TEST_ARGS)
+	kubectl get nodes -oyaml
+	GCP_ONLY=$(GCP_ONLY) go test -v $(REPOPATH)/integration -timeout 20m $(INTEGRATION_TEST_ARGS)
```
corneliusweig: I don't fully understand how this relates to your comment:

> Increase the duration of the test suite to increase the chance of running all the tests when one goes wrong.

Do you mean that this gives enough time for the remaining tests to finish, so that they are not marked as failed?

integration/util.go
```diff
@@ -103,6 +103,8 @@ func (k *NSKubernetesClient) WaitForPodsReady(podNames ...string) {
 	waitLoop:
 		select {
 		case <-ctx.Done():
+			k.printDiskFreeSpace()
+			k.debug("nodes")
```
corneliusweig: Does `k.debug("nodes")` add anything? After all, the nodes are already unconditionally printed in `make integration` (https://github.com/GoogleContainerTools/skaffold/pull/2406/files#diff-b67911656ef5d18c4ae36cb6741b7965R112).

Makefile
dgageot: Exactly! It's so better said!

integration/util.go
dgageot: When the `TestDebug` used to fail, I could see an issue with the k8s nodes.
I added those logs before the test suite and in case of an error, to detect cases where the kind cluster might get broken after running a few tests.

corneliusweig: Ok, that makes sense.

Issue #2416, opened 2019-07-05
Add support for Kustomize Kustomization resources


jmcclell:
As of version 2.1, Kustomize has deprecated the concept of "bases" in favor of combining the concept into the "resources" array by enforcing merge order. This means that as of 2.1, a resource can refer to a Kustomization (a directory containing a kustomization config at its root) rather than simply a file. This means that Skaffold's Kustomization support needs to be expanded to treat resources as if they are potentially bases in order to properly construct the dependency array for file monitoring.

This also tangentially relates to Issue #2082 (fixed in PR #2269), as the handling of resources needs to to take into account the possibility that they may now be remote just like bases were.

jmcclell:
@dgageot This is probably actually a bug rather than a feature request, because today if you use version 2.1 and opt to use the new resource functionality rather than bases, Skaffold won't properly watch all the files as it won't follow the dependency tree. I'm putting a PR together now.

dgageot:
great

jmcclell:
Closed by the merge of #2420

PR #2269, opened 2019-06-14
Allow for remote kustomize bases

jmcclell:
Fixes #2082 

Kustomize will error if it can't find a base dependency, so it's safe for
Skaffold to assume that any base file that does not exist locally is intended
to be a remote path that can be ignored since Skaffold can't monitor remote
paths for changes.

PR #2420, opened 2019-07-05
Add support for kustomization resources (#2416)

jmcclell:
As of version 2.1, Kustomize has deprecated bases in favor of combining
the functionality with resources by enforcing merge order based on
resource array order. This means resources now may act as bases do today
instead of being simply files - they may point to kustomiziations
(directories with a kustomization config in the root) or even to remote
locations. The code to generate the list of dependencies for file
watching has been updated accordingly.

----

This will close  Issue #2416. Please review it carefully as it's a non-trivial change to the behavior of dependency gathering. I've added the basic test cases, but if you can think of any edge cases to add let me know.

Note: tests now must create all files referenced within both the "resources" and "bases" section of any kustomization configs encountered during the walk, otherwise those files will not be added to the dependency array and the test will fail. This is to allow support for remote files, as first introduced in PR #2269.

### DIFF
--- a/pkg/skaffold/deploy/kustomize.go
+++ b/pkg/skaffold/deploy/kustomize.go
@@ -167,23 +167,30 @@ func dependenciesForKustomization(dir string) ([]string, error) {
 		return nil, err
 	}
 
-	for _, base := range content.Bases {
+	deps = append(deps, path)
+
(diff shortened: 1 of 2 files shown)


jmcclell:
@dgageot As promised.

PR #2424, opened 2019-07-06
Wait for parallel builds to be cancelled on error

### LINKED ISSUE #2268: One Kaniko artifact build failure can stop other pod(s) from being cleaned up
### Expected behavior
When building two artifacts with Kaniko, if building the first artifact fails, the pod for building the second artifact is eventually cleaned up

### Actual behavior
Skaffold exits with a fatal error after the first build failed, and (sometimes?) the second pod stays around (whether in a "Completed" or "Failed" status)
```
Starting build...
Creating kaniko secret [kaniko-secret]...
Creating docker config secret [docker-cfg]...
Building [dummy]...
Storing build context at /tmp/context-cd5dec741b7cd259ed9591258fda8298.tar.gz
INFO[0000] No base image, nothing to extract
INFO[0000] Unpacking rootfs as cmd RUN false requires it.
INFO[0000] Taking snapshot of full filesystem...
INFO[0000] Skipping paths under /kaniko, as it is a whitelisted directory
INFO[0000] Skipping paths under /secret, as it is a whitelisted directory
INFO[0000] Skipping paths under /var/run, as it is a whitelisted directory
INFO[0000] Skipping paths under /dev, as it is a whitelisted directory
INFO[0000] Skipping paths under /sys, as it is a whitelisted directory
INFO[0000] Skipping paths under /proc, as it is a whitelisted directory
INFO[0000] RUN false
INFO[0000] cmd: /bin/sh
INFO[0000] args: [-c false]
error building image: error building stage: starting command: fork/exec /bin/sh: no such file or directory
FATA[0011] build failed: building [dummy]: kaniko build for [dummy]: waiting for pod to complete: condition error: pod already in terminal phase: Failed
```
It seems like pods for both builds are launched at first. The first pod is cleaned up, but the second one stays:
```
$ kubectl get pod
NAME           READY   STATUS            RESTARTS   AGE
kaniko-4qpdz   0/1     Error             0          25s
```

### Information

- Skaffold version: v0.31.0
- Operating system: Ubuntu 16.04
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta11
kind: Config
build:
  artifacts:
    - image: dummy
      kaniko:
        dockerfile: Dockerfile.fail
        buildContext:
          localDir: {}
    - image: dummy2
      kaniko:
        buildContext:
          localDir: {}
  cluster:
    dockerConfig: {}
    resources:
      limits:
        cpu: 200m
        memory: 200M
```

1. Create Dockerfiles and `skaffold.yaml`
`Dockerfile`:
```
FROM scratch
RUN sleep 10  # this fails anyway because there's no shell, but it's ok
```
`Dockerfile.fail`:
```
FROM scratch
RUN false
```
2. `skaffold run`

tbarrella:
to fix #2268 

Previously, if building the first cluster artifact failed, skaffold
would exit immediately after. Build goroutines for other artifacts may
have still been running and not get a chance to clean up

The Kaniko build has logic to handle cancellation, but allow custom
commands to be cancelled as well

I'm still new to Go so be liberal with requesting changes

### DIFF
--- a/pkg/skaffold/build/parallel.go
+++ b/pkg/skaffold/build/parallel.go
@@ -50,6 +50,9 @@ func InParallel(ctx context.Context, out io.Writer, tags tag.ImageTags, artifact
 		return runInSequence(ctx, out, tags, artifacts, buildArtifact)
 	}
 
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
@@ -57,6 +60,7 @@ func InParallel(ctx context.Context, out io.Writer, tags tag.ImageTags, artifact
 	outputs := make([]chan []byte, len(artifacts))
 
 	// Run builds in //
+	wg.Add(len(artifacts))
 	for i := range artifacts {
 		outputs[i] = make(chan []byte, buffSize)
 		r, w := io.Pipe()
@@ -64,7 +68,7 @@ func InParallel(ctx context.Context, out io.Writer, tags tag.ImageTags, artifact
 
 		// Run build and write output/logs to piped writer and store build result in
 		// sync.Map
-		go runBuild(ctx, cw, tags, artifacts[i], results, buildArtifact)
+		go runBuild(ctx, &wg, cw, tags, artifacts[i], results, buildArtifact)
 		// Read build output/logs and write to buffered channel
 		go readOutputAndWriteToChannel(r, outputs[i])
 	}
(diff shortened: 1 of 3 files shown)


pkg/skaffold/build/parallel.go
```diff
@@ -50,21 +50,25 @@ func InParallel(ctx context.Context, out io.Writer, tags tag.ImageTags, artifact
...
+	wg.Add(len(artifacts))
 	for i := range artifacts {
 		outputs[i] = make(chan []byte, buffSize)
 		r, w := io.Pipe()
 		cw := setUpColorWriter(w, out)
 
 		// Run build and write output/logs to piped writer and store build result in
 		// sync.Map
-		go runBuild(ctx, cw, tags, artifacts[i], results, buildArtifact)
+		go runBuild(ctx, &wg, cw, tags, artifacts[i], results, buildArtifact)
```
tbarrella: If I put a closure here to avoid passing in `wg`, it would cause tests to fail for me by changing the build order

dgageot: This is because `i` needs to be captured with something like `i := i` before running the go routine. I'd rather we don't pass the waitGroup as an argument to runBuild. It makes the code harder to reason about.

tbarrella: Ah, thank you!

PR #2283, opened 2019-06-18
Kaniko proxy

prary:
Implementation of PR #2064 Setting proxy for Kaniko Pod

PR #2430, opened 2019-07-08
Freeze v1beta12 and prepare v1beta13

dgageot:
A new version is required by #2188


tejal29:
I have another approved PR #2283 which is waiting on config freeze.

PR #2437, opened 2019-07-09
Add a `metadata.name` field to skaffold.yaml

### LINKED ISSUE #2200: Add 'metadata' field to skaffold schema
My team is using kustomize with skaffold, we've moved common config to centralized repos and are using kustomize's 'remote base' feature. We have taken this further and centralized a generic `skaffold.yaml` which we modify by repo to specify the artifact details. So during CI/CD the `skaffold.yaml` file is dynamically created from remote bases and local kustomizations.

I've added a
```
metadata:
  name: base-config
```
block to the central `skaffold.yaml` file as kustomize will error out without one, but this is not a legal field in the current skaffold schema so I have to pipe the generated skaffold config through sed, to remove those two lines; not ideal.

I propose a metadata field be added to the `skaffold.yaml` schema. This can hold, at the minimum, name inside it so that skaffold configs can be modified with kustomize, as well as other values that may be used for other things.

corneliusweig:
This allows to name pipelines such that multiple skaffold.yaml can be merged with kustomize (#2200).

Close #2200

### DIFF
--- a/pkg/skaffold/schema/latest/config.go
+++ b/pkg/skaffold/schema/latest/config.go
@@ -35,11 +35,20 @@ type SkaffoldConfig struct {
 	// Kind is always `Config`. Defaults to `Config`.
 	Kind string `yaml:"kind" yamltags:"required"`
 
+	// Metadata holds additional information about the config.
+	Metadata Metadata `yaml:"metadata,omitempty"`
+
 	// Pipeline defines the Build/Test/Deploy phases.
 	Pipeline `yaml:",inline"`
 
 	// Profiles *beta* can override be used to `build`, `test` or `deploy` configuration.
 	Profiles []Profile `yaml:"profiles,omitempty"`
+}
+
+// Metadata holds an optional name of the project.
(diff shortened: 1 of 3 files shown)


pkg/skaffold/schema/latest/config.go
```diff
@@ -35,13 +35,22 @@ type SkaffoldConfig struct {
...
 	// Pipeline defines the Build/Test/Deploy phases.
 	Pipeline `yaml:",inline"`
 
 	// Profiles *beta* can override be used to `build`, `test` or `deploy` configuration.
 	Profiles []Profile `yaml:"profiles,omitempty"`
 }
 
+// Metadata holds an optional name of the project.
+type Metadata struct {
+	// Name is an identifier for the project.
```
corneliusweig: @balopat Would this be a good place to warn about non-unique project names?

balopat: Maybe, as we implement the kubecontext switching. It wouldn't make sense before it.

PR #2439, opened 2019-07-10
Support alternative Kustomization config filenames (#2422)

### LINKED ISSUE #2422: Add support for alternative Kustomization config names
### Expected behavior
When using `kustomization.yml` or `Kustomization` config files (as  opposed to the more common `kustomization.yaml`) Skaffold should function normally and watch all files in dev mode.

### Actual behavior
Skaffold fails to recognize these files as Kustomization configurations and it produces an error since the code expects `kustomization.yaml` to exist at the root of any Kustomizatiton base.

I have code to fix this, but it relies on code in #2416. I'll open a PR once that PR is ultimately merged.

jmcclell:
Closes #2422

### DIFF
--- a/pkg/skaffold/deploy/kustomize.go
+++ b/pkg/skaffold/deploy/kustomize.go
@@ -18,6 +18,7 @@ package deploy
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


jmcclell:
Not sure why tests failed for linux. Looks unrelated to the patch, however?

priyawadhwa:
@jamielennox we just merged a change to our proto files, could you try rebasing on master?

PR #2447, opened 2019-07-10
Add CLI option `--kube-context` to override the kubecontext in Skaffold

corneliusweig:
This PR adds a new CLI option `--kube-context` to Skaffold `run`, `dev`, `debug`, `delete`, `deploy`, and `build`. It's the frist step in providing several options to override the kubecontext to use in Skaffold.

See #2384
~~Depends on #2509~~

### DIFF
--- a/pkg/skaffold/kubernetes/context/context.go
+++ b/pkg/skaffold/kubernetes/context/context.go
@@ -36,8 +36,8 @@ var (
 	kubeContext    string
 )
 
-// ResetConfig is used by tests
-func ResetConfig() {
+// resetConfig is used by tests
+func resetConfig() {
 	kubeConfigOnce = sync.Once{}
 }
 
@@ -47,12 +47,16 @@ func UseKubeContext(overrideKubeContext string) {
 	kubeContext = overrideKubeContext
 }
 
+// GetRestClientConfig returns a REST client config for API calls against the Kubernetes API.
+// If UseKubeContext was called before, the CurrentContext will be overridden.
+// The result will be cached after the first call.
 func GetRestClientConfig() (*restclient.Config, error) {
 	clientConfig, err := getKubeConfig().ClientConfig()
 	return clientConfig, errors.Wrap(err, "error creating kubeConfig")
 }
 
 // getCurrentConfig retrieves the kubeconfig file. If UseKubeContext was called before, the CurrentContext will be overridden.
+// The result will be cached after the first call.
 func getCurrentConfig() (clientcmdapi.Config, error) {
 	cfg, err := getKubeConfig().RawConfig()
 	if kubeContext != "" {
@@ -62,7 +66,7 @@ func getCurrentConfig() (clientcmdapi.Config, error) {
 	return cfg, errors.Wrap(err, "loading kubeconfig")
 }
 
-// getKubeConfig creates a cached kubeConfig. If UseKubeContext was called before, the CurrentContext will be overridden.
+// getKubeConfig retrieves and caches the kubeConfig. If UseKubeContext was called before, the CurrentContext will be overridden.
 func getKubeConfig() clientcmd.ClientConfig {
 	kubeConfigOnce.Do(func() {
 		loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
--- a/pkg/skaffold/kubernetes/context/context.go
+++ b/pkg/skaffold/kubernetes/context/context.go
@@ -42,7 +42,6 @@ func resetConfig() {
 }
 
 // UseKubeContext sets an override for the current context in the k8s config.
-// This override must be set before calling CurrentConfig.
 func UseKubeContext(overrideKubeContext string) {
 	kubeContext = overrideKubeContext
 }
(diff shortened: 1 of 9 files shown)


pkg/skaffold/kubernetes/context/context.go
```diff
@@ -30,26 +31,44 @@ var (
 )
 
 var (
-	currentConfigOnce sync.Once
-	currentConfig     clientcmdapi.Config
-	currentConfigErr  error
+	kubeConfigOnce sync.Once
+	kubeConfig     clientcmd.ClientConfig
+	kubeContext    string
```
corneliusweig: I'm a bit sad about introducing some global state here. However, I have considered several alternatives and this appeared to be the one with the least head-ache overall. Let me know your thoughts...

pkg/skaffold/kubernetes/context/context.go
```diff
@@ -30,26 +31,44 @@ var (
...
+func ResetConfig() {
+	kubeConfigOnce = sync.Once{}
 }
 
+// UseKubeContext sets an override for the current context in the k8s config.
+// This override must be set before calling CurrentConfig.
+func UseKubeContext(overrideKubeContext string) {
+	kubeContext = overrideKubeContext
+}
+
```
balopat: provide docstring please

pkg/skaffold/kubernetes/context/context.go
```diff
@@ -30,26 +31,44 @@ var (
...
+	kubeConfigOnce sync.Once
+	kubeConfig     clientcmd.ClientConfig
+	kubeContext    string
 )
 
-// ResetCurrentConfig is used by tests
-func ResetCurrentConfig() {
-	currentConfigOnce = sync.Once{}
+// ResetConfig is used by tests
+func ResetConfig() {
```
balopat: does this need to be public/exported? can we make it private?

pkg/skaffold/kubernetes/context/context.go
```diff
@@ -30,26 +31,44 @@ var (
...
+}
+
+func GetRestClientConfig() (*restclient.Config, error) {
+	clientConfig, err := getKubeConfig().ClientConfig()
+	return clientConfig, errors.Wrap(err, "error creating kubeConfig")
+}
+
+// getCurrentConfig retrieves the kubeconfig file. If UseKubeContext was called before, the CurrentContext will be overridden.
 func getCurrentConfig() (clientcmdapi.Config, error) {
-	currentConfigOnce.Do(func() {
```
balopat: why did we change the semantics of the `currentConfigOnce`? 
Doesn't this open up the possibilities of inconsitencies in a skaffold process that started off version1 of a kubeconfig file but then someone mutated that file and a second read would be of version2?

corneliusweig: The once-semantic is now encapsulated in `getKubeConfig`. So for the moment we are safe here. However, in the future when the kube-context is also overridden by `skaffold.yaml`, we will have to re-read the kubeconfig. This will occur during the `skaffold.yaml` parsing/profile phase and should not take very long. Do you think Skaffold should take extra precautions to work with the identical kubeconfig as in the first kubeconfig read?

pkg/skaffold/kubernetes/context/context.go
corneliusweig: Yes, it can be private. Don't know why it was public so far.

pkg/skaffold/kubernetes/context/context.go
corneliusweig: I was already investigating in that direction. That requires to pass down the RunContext in quite a few places, so I decided against it. But it's a good idea nevertheless.

pkg/skaffold/kubernetes/context/context.go
corneliusweig: I found a nice way to ensure that Skaffold always uses the kubeconfig from the first read. Note that this has not been the case with respect to the kubernetes REST client so far, and is much more consistent now.

Note that there is still the possibility, that the kubecontext changes due to a change in `skaffold.yaml` (via `UseKubeContext`). However, that is the desired behavior.

corneliusweig:
@dgageot I'm desparate. I don't understand why my test isn't working. If you have time and energy to have a look, I'd be really grateful! (I'm asking you because you set up kind).

What I'm trying to do: Set up an integration test where skaffold uses a modified kubeconfig. The test runs locally (inside and outside of kind), and it runs when I run all tests or just a single one (locally). But on travis, it consistently shows exactly this error (https://travis-ci.org/GoogleContainerTools/skaffold/jobs/559384994#L1935):

> deployment test-dev failed due to Running [kubectl --context modified-context@kind rollout status deployment test-dev --watch=false]: stdout , stderr: Error from server (NotFound): deployments.extensions "test-dev" not found
            , err: exit status 1: exit status 1

However, I don't understand why it's looking for deployment `test-dev` and I can't even find the location which does this check. The previous test-case does have this check, but it should be over when this test starts. I'm somewhat out of ideas...

balopat:
I will have a look tomorrow - but as a tip: can you rebase on top of master? Sometimes changes might impact Travis vs local - as Travis does merge master in your branch.

corneliusweig:
@balopat :man_facepalming: Thanks that gives me a lead as to what is going wrong. You don't have to waste your time, I can debug further now.

corneliusweig:
Alright, I found what is going on:
When my new test-case runs, the deployment `test-dev` from the previous test is still around. This deployment is erroneously picked up by `pkg/skaffold/deploy/status_check.go:80`. From there on, the test is doomed.

I think this is mainly a problem how the `status_check` code determines the deployments to consider for watching. It would be better, if the status-check was looking for a UUID label which is generated for each Skaffold run. @tejal29 WDYT?

balopat:
yup, we're missing kubecontext here: https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go#L54

It could be useful to just do a mechanical search for anything `kubectl` and double check if kubecontext is missing.

corneliusweig:
Oh no! I wasn't aware that so many places called `kubectl` directly without going through `pkg/skaffold/deploy/kubectl`. I think the cleaner approach would be to consistently use our wrapper to call `kubectl`, but I will have to see how much refactoring this will require.

Thanks for catching this early!

corneliusweig:
@nkubala Do you know why I can't get a clean build here? The tests on mac are always failing. I have tried rebasing several times, but it never succeeded.

corneliusweig:
Rebase is done. Thanks everyone for reviewing and helping out!

corneliusweig:
Looks like another glitch in the CI. I've rebased one more time :)

tejal29:
Sorry @corneliusweig, we changed from travis-ci.org to premium. We cancelled all jobs in the old free travis and hence you must have seen a failure. 
You job is green on Travis. I just started kokoro job

PR #2456, opened 2019-07-12
[caching] Simpler code and fixing nits

dgageot:


dgageot:
ping @priyawadhwa

priyawadhwa:
One question, why are we removing the `WorkspaceHash` from some older versions?

dgageot:
@priyawadhwa because it's not used

PR #2476, opened 2019-07-15
File output flag for writing built images to a specified file

marlon-gamez:
This is essentially the same as running skaffold build --quiet > file.out, but doesn't silence the standard output from skaffold.

### DIFF
--- a/cmd/skaffold/app/cmd/build.go
+++ b/cmd/skaffold/app/cmd/build.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"io"
 	"io/ioutil"
+	"os"
 	"time"
 
 	"github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/flags"
@@ -35,6 +36,7 @@ import (
 var (
 	quietFlag       bool
 	buildFormatFlag = flags.NewTemplateFlag("{{json .}}", flags.BuildOutput{})
+	buildOutputFlag string
 )
 
 // NewCmdBuild describes the CLI command to build artifacts.
@@ -46,6 +48,7 @@ func NewCmdBuild() *cobra.Command {
 			f.StringSliceVarP(&opts.TargetImages, "build-image", "b", nil, "Choose which artifacts to build. Artifacts with image names that contain the expression will be built only. Default is to build sources for all artifacts")
 			f.BoolVarP(&quietFlag, "quiet", "q", false, "Suppress the build output and print image built on success. See --output to format output.")
 			f.VarP(buildFormatFlag, "output", "o", "Used in conjunction with --quiet flag. "+buildFormatFlag.Usage())
+			f.StringVar(&buildOutputFlag, "file-output", "", "Filename to write build images to")
 		}).
 		NoArgs(cancelWithCtrlC(context.Background(), doBuild))
 }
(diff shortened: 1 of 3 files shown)


tejal29:
This looks good! Few comments
- if you have a `--file-output` flag on, the build output log lines should go to stdout. Is that happenning?
- Write tests where
   -- both --quiet and --file-output is present. In this case, either you shd see build output on std out and also in the file or error out saying both are not supported. I feel 1st is better.
   -- both --file-output and -o are present the file has the correct formatted output.
   - also make sure, if the builder prints some output it is written to std out when --file-output is present and no --quiet.

cmd/skaffold/app/cmd/build.go
```diff
@@ -64,8 +67,19 @@ func doBuild(ctx context.Context, out io.Writer) error {
 	return withRunner(ctx, func(r runner.Runner, config *latest.SkaffoldConfig) error {
 		bRes, err := r.BuildAndTest(ctx, buildOut, targetArtifacts(opts, config))
 
-		if quietFlag {
+		if quietFlag || buildOutputFlag != "" {
 			cmdOut := flags.BuildOutput{Builds: bRes}
+
+			if buildOutputFlag != "" {
+				f, err := os.OpenFile(buildOutputFlag, os.O_RDWR|os.O_CREATE, 0755)
```
tejal29: Can you use ioutil.WriteFile like skaffold fix uses 
 here https://github.com/GoogleContainerTools/skaffold/blob/master/cmd/skaffold/app/cmd/fix.go#L73
`if err := ioutil.WriteFile(configFile, newCfg, 0644); err != nil {
}
`

tejal29: ah ok. 
Can you create a bytes.buffer and execute the template there
```
buildOut bytes.Buffer
t.Execute(buildOut, r)
```

And the if --quiet is present, 
```
out.Write(buildOut)
```
if "--file-output" is present, 
```
err := ioutil.WriteFile(buildOutputFlag, buildOut, 0644)
if err != nil {
  log.Fatal(err)
}
```
That ways its cleaner and we do not overwrite  `out`

docs/content/en/docs/references/cli/_index.md
```diff
@@ -104,6 +104,7 @@ Options:
       --cache-file='': Specify the location of the cache file (default $HOME/.skaffold/cache)
   -d, --default-repo='': Default repository value (overrides global config)
       --enable-rpc=false: Enable gRPC for exposing Skaffold events (true by default for `skaffold dev`)
+      --file-output='': Filename to write build images to
```
AndiDog: Do we still need "Used in conjunction with --quiet flag" for `-o, --output`? I guess from now on that flag even applies without quiet mode?!

twuttk:
Is there something off with the documentation? It works fine with `skaffold build --file-output=blub` but the documentation (and skaffold build --help) claims it should be `skaffold build --output=blub` which doesn't work.

docs/content/en/docs/references/cli/_index.md
marlon-gamez: I think that should still be there, but be updated to say "Used in conjunction with --quiet or --file-output", since those are the only two flags whose output it effects.

marlon-gamez:
> Is there something off with the documentation? It works fine with `skaffold build --file-output=blub` but the documentation (and skaffold build --help) claims it should be `skaffold build --output=blub` which doesn't work.

--output is actually the flag used to format the output of --quiet (and now --file-output as well). The naming is a bit confusing, sorry about that. Running skaffold build --help should show documentation for both --output and --file-output.

PR #2477, opened 2019-07-15
Fix port forwarding in dev loop

priyawadhwa:
Port forwarding wasn't being restarted after each loop of the dev cycle,
so I've moved it to happen after `runner.Deploy` is called and before `ErrorConfigurationChanged` error is returned. This PR also
fixes a bug where used local ports were being stored as type int but not
deleted because the code was trying to look up the keys as type int32.
I amended a unit test to make sure local ports are cleaned up correctly
as well.

Another bug is fixed here: The check for `portForwardSuccessful` never succeeded the first time as it is a race condition with the command execution - checks binding to the port too fast. So the poller tried a second time to port forward - now, the command `kubectl forward` gets executed but fails in a go func where the error is swallowed - but the `portForwardSuccessful` returns no error as the first port forward already bound to the port - however, this "seemingly successful" second port forward overrides the `cancel` function on the portforward entry ... so when we call `Terminate` - the first portforward keeps running happily, while the second one's cancel cancels an already exited context. Fix: 
- introduce a 5 second polling wait with 200 millisecond poll interval to determine if the port forward is successful 
- added a safety call to `cancel` in case the same portforward entry is retried the previous context is cleared up 
- added error handling and logging in the go func for the `cmd.Wait()` instead of swallowing it


pkg/skaffold/runner/dev.go
```diff
@@ -63,6 +63,7 @@ func (r *SkaffoldRunner) doDev(ctx context.Context, out io.Writer) error {
 		}
 		fallthrough
 	case r.changeSet.needsRedeploy:
+		r.forwarderManager.Stop()
```
tejal29: would this gaurantee portforward is stopped when dev context is cancelled when user presses "cntrl c"?

pkg/skaffold/runner/dev.go
```diff
@@ -151,10 +151,5 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
 		}
 	}
 
-	// Forward ports
-	if err := forwarderManager.Start(ctx); err != nil {
```
tejal29: would this be effective change as moving this at L147?  and add at L70 again. its easier to read since, 
- you stopped portforwading 
- did a deploy
- start port forwarding again.

The deploy does not get cluttered with portforward management.
WDYT?

pkg/skaffold/runner/dev.go
priyawadhwa: yah, so each go routine created by portforwarding depends on the `ctx context.Context` passed in to this function, so when the user presses "cntrl-c" they should all be stopped.

pkg/skaffold/kubernetes/portforward/entry_manager_test.go
```diff
@@ -62,6 +64,10 @@ func TestStop(t *testing.T) {
 	em.forwardedResources.Store("pod-resource-default-0", pfe1)
 	em.forwardedResources.Store("pod-resource2-default-0", pfe2)
 
+	em.forwardedPorts = &sync.Map{}
+	em.forwardedPorts.Store(9000, struct{}{})
```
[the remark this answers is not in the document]
priyawadhwa: Once a goroutine has selected an unused port, it will store it in this map. Before other goroutines try and use that port, they will check and make sure it doesn't exist in the map, so multiple goroutines shouldn't find the same port unused. 

This is also convenient because it allows us to use the [GetAvailablePort](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/util/port.go#L37) function which takes in a sync.Map and is also used by the event API, so I'd prefer not to change it, unless you feel strongly. WDYT?

pkg/skaffold/runner/dev.go
priyawadhwa: that makes sense, thanks! just changed it.

pkg/skaffold/kubernetes/portforward/entry_manager_test.go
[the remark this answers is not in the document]
tejal29: like you mentioned. "Once a goroutine has selected an unused port, it will store it in this map" The bug is when multiple routines happen to detect it as unused at the same time.  Does "GetAvailablePort" get called in go subroutines?
happy to chat offline.

pkg/skaffold/kubernetes/portforward/entry_manager.go
```diff
@@ -89,6 +89,6 @@ func (b *EntryManager) Stop() {
 // Terminate terminates a single port forward entry
 func (b *EntryManager) Terminate(p *portForwardEntry) {
 	b.forwardedResources.Delete(p.key())
-	b.forwardedPorts.Delete(p.localPort)
+	b.forwardedPorts.Delete(int(p.localPort))
```
dgageot: Shouldn't we use `uint16` everywhere?

pkg/skaffold/kubernetes/portforward/entry_manager_test.go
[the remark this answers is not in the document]
dgageot: We should probably introduce our own types that wrap a `map` and a lock. That would make the rest of the code easier to read and maintain.

pkg/skaffold/runner/dev.go
```diff
@@ -144,17 +147,16 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
 		return errors.Wrap(err, "exiting dev mode because first deploy failed")
 	}
 
+	if err := r.forwarderManager.Start(ctx); err != nil {
+		logrus.Warnln("Error starting port forwarding:", err)
```
dgageot: It used to be an error. Now it's a warning. Why would we not fail?

pkg/skaffold/runner/dev.go
```diff
@@ -79,8 +83,7 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
 	r.createLogger(out, artifacts)
 	defer r.logger.Stop()
 
-	forwarderManager := portforward.NewForwarderManager(out, r.imageList, r.runCtx.Namespaces, r.defaultLabeller.K8sManagedByLabelKeyValueString(), r.runCtx.Opts.PortForward, r.portForwardResources)
-	defer forwarderManager.Stop()
```
dgageot: Why not defer the stop anymore?

pkg/skaffold/runner/dev.go
```diff
@@ -79,8 +83,7 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
...
+	r.forwarderManager = portforward.NewForwarderManager(out, r.imageList, r.runCtx.Namespaces, r.defaultLabeller.K8sManagedByLabelKeyValueString(), r.runCtx.Opts.PortForward, r.portForwardResources)
```
dgageot: Could this be extracted to a `createForwarder()` function, similar to `createLogger`?

balopat:
> Note: Ports are forwarded after each loop of the dev cycle, but they
> aren't being forwarded to the same port.
> 
> So, on loop 1, svc/mysvc will be forwarded to port 9000.
> Then, on loop 2, svc/mysvc will be forwarded to port 9001.
> 
> Currently, port 9000 is coming back as "in use" even though it's been
> removed from the map of local ports and the go function calling it has
> been cancelled. I'll look into this in another PR.

I would love to tackle this in one go if possible. This non-stable / self-colliding port forward is not great UX. Deployments can be very fast, if the user has to open a new browser on every dev cycle is not nice. Also, in a multiservice setup we generate a new port for all services regardless whether it changed or not...

pkg/skaffold/kubernetes/portforward/entry_manager_test.go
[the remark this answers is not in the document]
priyawadhwa: Makes sense, thanks @tejal29  and @dgageot for the explanations! I've updated this.

pkg/skaffold/kubernetes/portforward/entry_manager.go
priyawadhwa: So I updated to use int because a lot of the helper functions expect an int -- we've been using int32 because that's the type used for `ContainerPort` in Kubernetes, but it's only read in one place so I don't think we need to stick with it.

I can change it to uint16 if you feel strongly, but this seems to require the least amount of casting.

pkg/skaffold/runner/dev.go
priyawadhwa: So I think having the stop here is actually a bug. The stop should happen at the end of every dev loop, not at the end of the dev command, since we want to clean up ports and port-forward again on every cycle.

pkg/skaffold/runner/dev.go
priyawadhwa: sure, done

pkg/skaffold/runner/dev.go
priyawadhwa: I figured if port forwarding doesn't work, we don't want the dev loop to error out, since users might still be making changes or checking logs. To me, port forwarding isn't a feature that should cause skaffold to break, since it's kind of an extra feature (nice to have, but not necessary to the dev loop). WDYT?

balopat:
I added a fix for the port repetition with an ugly whitebox integration test. I'll try to clean up on it tomorrow.

pkg/skaffold/kubernetes/portforward/port_forward_integration.go
```diff
@@ -0,0 +1,71 @@
...
+	nextPort := retrieveAvailablePort(localPort, em.forwardedPorts)
+
+	// theoretically we should be able to bind to the very same port
+	// this might get flaky when multiple tests are ran. However
+	// we shouldn't collide with our own process because of poor cleanup
+	if nextPort != localPort {
+		t.Fatalf("the same port should be still open, instead first port: %d != second port: %d", localPort, nextPort)
+	}
+
+	defer em.Stop()
```
priyawadhwa: This was already called on line 48

balopat: we can remove the second defer!

integration/white_box_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+func TestPortForward(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+	ns, _, deleteNs := SetupNamespace(t)
+	defer deleteNs()
+	dir := "examples/microservices"
+	skaffold.Run().InDir(dir).InNs(ns.Name).RunOrFailOutput(t)
+	defer skaffold.Delete().InDir(dir).InNs(ns.Name).RunOrFailOutput(t)
```
dgageot: You can remove the delete. I’ll be removed with the ns

integration/white_box_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/portforward"
+)
+
+func TestPortForward(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+	ns, _, deleteNs := SetupNamespace(t)
```
dgageot: I think it also needs a check on GCP only tests

integration/white_box_test.go
balopat: right, that makes sense!

integration/white_box_test.go
balopat: why? as long as I have a cluster this can work, and it did run on kind on travis

dgageot: To be skipped on kokoro

balopat: ah right. 🤦‍♂

pkg/skaffold/kubernetes/portforward/entry_manager.go
```diff
@@ -31,28 +31,124 @@ import (
...
+}
+
+func (f forwardedPorts) Store(key, value interface{}) {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+	val, ok := key.(int)
+	if !ok {
+		panic("only store keys of type int in forwardedPorts")
+	}
+	f.ports[val] = struct{}{}
```
balopat: is this a bug? shouldn't we store value instead of empty struct?

priyawadhwa: whoops, val should be renamed to `k`, thanks for catching this!

pkg/skaffold/runner/dev.go
balopat: It's a slightly related problem? with Nick's PR #2501 the config change scenario is back in action. And I just tried it and because of removing this `Stop` here, we leave kubectl processes hanging around. We could add the stop in the devLoop before returning the error though.

balopat: I added the `Stop` before throwing the `ErrorConfigurationChanged` as well.

pkg/skaffold/runner/dev.go
tejal29: Warning sounds good.

Issue #2496, opened 2019-07-18
Interference between status check of multiple Skaffold runs


corneliusweig:
Skaffold now checks the deployment status. However, the checks for multiple runs can interfere.

For example, with this patch on master:
```patch
diff --git a/examples/react-reload/k8s/deployment.yaml b/examples/react-reload/k8s/deployment.yaml
index e4d64794..edc57ba2 100644
--- a/examples/react-reload/k8s/deployment.yaml
+++ b/examples/react-reload/k8s/deployment.yaml
@@ -12,7 +12,7 @@ spec:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: node
+  name: node-react
 spec:
   selector:
     matchLabels:
diff --git a/pkg/skaffold/deploy/status_check.go b/pkg/skaffold/deploy/status_check.go
index 96b101e8..64317e89 100644
--- a/pkg/skaffold/deploy/status_check.go
+++ b/pkg/skaffold/deploy/status_check.go
@@ -96,6 +96,7 @@ func getDeployments(client kubernetes.Interface, ns string, l *DefaultLabeller)
 		} else {
 			deadline = *d.Spec.ProgressDeadlineSeconds
 		}
+		logrus.Debugf("Register deployment %s for status check", d.Name)
 		depMap[d.Name] = deadline
 	}
 

```
Skaffold prints the watched deployments.

When doing
```
cd examples/nodejs
skaffold run
cd ../../examples/react-reload
skaffold dev -v debug 2>&1 | grep Register
```
It shows that the second run also watches the first deployment:
```
time="2019-07-18T13:26:16+02:00" level=debug msg="Register deployment node for status check"
time="2019-07-18T13:26:16+02:00" level=debug msg="Register deployment node-react for status check"
```

Deleting the first deployment with
```
cd examples/nodejs
skaffold delete
```
will eventually (after the timeout) lead to a failure of the `skaffold dev` run.

---
A possible way out would be to add a new label with a UUID that is generated in every skaffold run. In other words deployments should be selected via 
```go
deps, err := client.AppsV1().Deployments(ns).List(metav1.ListOptions{
		LabelSelector: "skaffold-run-uuid=xyz-xyz",
	})
```
in `pkg/skaffold/deploy/status_check.go:80`,

cc @tejal29

PR #2499, opened 2019-07-18
Set statuscheck to false.

### LINKED ISSUE #2495: Skaffold dev fails when user cannot list deployments at cluster scope
The deployment status check from Skaffold tries to list all deployments at cluster scope, when running Skaffold without the `--namespace` flag. For example:

```
cd examples/nodejs
kubectl auth can-i list deploy --all-namespaces  # -> no
skaffold dev
```

Fails with 
```
FATA[0001] exiting dev mode because first deploy failed: could not fetch deployments: could not fetch deployments: deployments.apps is forbidden: User "default-user" cannot list resource "deployments" in API group "apps" at the cluster scope
```
On the other hand:
```
kubectl auth can-i list deploy -n default  # -> yes
skaffold dev -n default
```
works as expected.

---

The reason for this is that `pkg/skaffold/deploy/status_check.go:80` does 
```go
	})
```
with `ns==""` (due to not specifying a namespace), and that means _all namespaces_.

tejal29:
Turn status-check false by default as a workaround for #2496 and #2495.


cmd/skaffold/app/cmd/flags.go
```diff
@@ -66,7 +66,7 @@ var FlagRegistry = []Flag{
 		Shorthand:     "n",
 		Usage:         "Run deployments in the specified namespace",
 		Value:         &opts.Namespace,
-		DefValue:      "",
+		DefValue:      "default",
```
balopat: let's separate this out to a different PR with tests and everything, I'm not sure about that this is what we want to do (we might but I need a bit more time to think it through).

balopat: more thought around this: there is a difference between leaving it empty and letting the resources determine their namespace vs enforcing `default`.

PR #2539, opened 2019-07-24
Let the user specify a path and a secret name

### LINKED ISSUE #2522: Automated kaniko secret generation for docker configuration used to work in v0.30.0 and now fails in v0.34.0
### Expected behavior
Building images using Kaniko as builder with automated secret generation should work.

### Actual behavior
```
Creating kaniko secret [kaniko-secret]...
Creating docker config secret []...
time="2019-07-22T13:03:17Z" level=fatal msg="build failed: setting up docker config secret: creating docker config secret: Secret \"\" is invalid: metadata.name: Required value: name or generateName is required: Secret \"\" is invalid: metadata.name: Required value: name or generateName is required"
```
### Information

We are using kaniko to build our images. Kaniko needs two secrets which contain the credentials for image pushing and the docker configuration. When used in conjunction with skaffold, one has the option to:
1. Create these secret manually and reference them in the skaffold.yaml
2. Reference the path to the docker/config.json in the skaffold yaml. In this case skaffold will create the required secrets automatically and delete them after image building has finished.

We are using method 2, which is working fine in skaffold v0.30.0. In skaffold v0.34.0, this fails with the above error. The error message itself is rather straight forward: Skaffold can't create a secret with an empty name. 

It's not clear how one would set the name for this secret though, in the past the secret was automatically named "docker-cfg". When used in conjunction with the additional field "secretName: foo", skaffold fails with the error that one should use either "secretName" OR "path". Which is correct, as these fields decide whether one uses option 1 or 2 from above. You can't do both.

I haven't found any documentation which explains how one would set the name for the docker-cfg secret in v0.34.0. 

We haven't changed the version of kaniko we use.

This is the relevant part from our skaffold.yaml:
```
build:
  artifacts:
    - image: <image>
      context:  <path-to-context>
      kaniko:
        dockerfile: <path-to-dockerfile>
        image: gcr.io/kaniko-project/executor:2fc8a7f4bcd4f91052c7defd80f90631b9e2b4cb
        cache:
          repo: <url-to-cache>
        buildContext:
          localDir: {}
  cluster:
    pullSecret: /root/.docker/config.json
    dockerConfig:
      path: /root/.docker/config.json
  tagPolicy:
    gitCommit: {}
```

dgageot:
Fixes #2522

### DIFF
--- a/pkg/skaffold/schema/defaults/defaults.go
+++ b/pkg/skaffold/schema/defaults/defaults.go
@@ -206,18 +206,18 @@ func setDefaultClusterDockerConfigSecret(cluster *latest.ClusterDetails) error {
 		return nil
 	}
 
-	if cluster.DockerConfig.Path != "" {
-		absPath, err := homedir.Expand(cluster.DockerConfig.Path)
-		if err != nil {
-			return fmt.Errorf("unable to expand dockerConfig.path %s", cluster.DockerConfig.Path)
-		}
+	cluster.DockerConfig.SecretName = valueOrDefault(cluster.DockerConfig.SecretName, constants.DefaultKanikoDockerConfigSecretName)
 
-		cluster.DockerConfig.Path = absPath
+	if cluster.DockerConfig.Path == "" {
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


dgageot:
ping @priyawadhwa do you think it makes sense?

priyawadhwa:
makes sense to me!

PR #2541, opened 2019-07-24
Go Modules

dgageot:


nkubala:
just tested the release build locally and it works for me, so looks like setting `GOMODULE11=on` in the makefile was enough to solve the xgo issue. @dgageot what more do you think needs to be done here?

dgageot:
@nkubala I think we're good. I was wondering if we should vendor dependencies or re-download them on each build. I think vendoring is faster. Let me remove the `[WIP]` prefix!

nkubala:
I'm not 100% sure but I think if we're keeping the vendored deps then we want to pass `-mod=vendor` to `go build`? otherwise won't it fetch the deps from `go.mod` and use those to build instead?

dgageot:
@nkubala TIL. It's fixed now

nkubala:
cool. and because of this we don't need to worry about downloading from a proxy right?

dgageot:
Yes

Le lun. 29 juil. 2019 à 21:50, Nick Kubala <notifications@github.com> a
écrit :

nkubala:
```
0.09s$ go build -mod=vendor -o out/skaffold.exe cmd/skaffold/skaffold.go
The command "go build -mod=vendor -o out/skaffold.exe cmd/skaffold/skaffold.go" exited with 1.

390.10s$ go test -short -timeout 60s ./...
build flag -mod=vendor only valid when using modules
```
not sure why this is happening, or why it only happened with the windows build...

PR #2544, opened 2019-07-25
Make Jib test projects more lightweight

briandealwis:
- Change Jib-based integration test projects (`integration/testdata/jib` and  `integration/testdata/debug/jib`) to use the built-in HTTP server from the JDK rather than pull in Spring Boot.  Build times for these projects goes from about 17s to 9s when all artifacts are downloaded from scratch.
- Change `integration/diagnose_test.go` to use the now-simpler `integration/testdata/jib`.  `skaffold diagnose` in this project goes from approx 6.9s to 5.5s.


dgageot:
@briandealwis This seems to have no impact on test duration on kokoro. but it looks good to me

PR #2548, opened 2019-07-25
Improve Jib support on gcb

### LINKED ISSUE #2546: Skaffold can't use the latest gcr.io/cloud-builders/mvn to with with Jib on GCB
The version of the `gcr.io/cloud-builders/mvn` builder that Skaffold uses on GCB to build jib artifacts was pinned a [while ago](https://github.com/GoogleContainerTools/skaffold/commit/e25305966453caf7d41bc6894315a534f0e5e275) to a version that contained `docker`.

The reason is that Skaffold runs `jib:dockerBuild` and lets Cloud Build push the result image. Instead, it could use `jib:build` and let Jib push the image itself. Which was complicated because of https://github.com/GoogleContainerTools/jib/issues/1500

dgageot:
Fix #2546


pkg/skaffold/build/gcb/jib.go
```diff
@@ -22,17 +22,22 @@ import (
...
+	base.Steps = []*cloudbuild.BuildStep{{
 		Name: b.GradleImage,
-		Args: jib.GenerateGradleArgs("jibDockerBuild", tag, artifact, b.skipTests),
+		Args: fixHome(jib.GenerateGradleArgs("jib", tag, artifact, b.skipTests)),
 	}}
+	return base
+}
+
+func fixHome(args []string) []string {
+	return append([]string{"-Duser.home=/builder/home"}, args...)
```
chanseokoh: I am concerned with this workaround, which changes the user home JVM-wise globally affecting all code including Maven and Gradle. For example, the user may be caching `/root/.m2` manually using GCS, and this will be a breaking change. I still believe GCB should fix this at the core. Either set `$HOME` to `/builder/home` or have `.docker/config.json` under `/root`. If GCB can't fix this soon, alternatively I think it is better to introduce a new flag in Jib to allow configuring the location of the Docker config, which can be a useful feature. @GoogleContainerTools/java-tools-build WDYT?

dgageot: I don't see how it's dangerous. The user has no control over how Skaffold uses GCB. So they can't "be caching /root/.m2 manually". Skaffold creates the whole `cloudbuild.yaml` for the user.

Also, GCB does set `HOME` to `/builder/home`.

chanseokoh: Oh, yeah, my mistake about `$HOME`, sorry. To be fair, I did know `HOME` points to `/builder/home`, which is very confusing actually, because the OS-defined home directory for the user `root` is still `/root`. (The Maven runs as `root` and stores artifacts under `/root/.m2`.)

You may be right that it is not dangerous in the context of Skaffold usage. It is just that I don't really like the idea of switching out the home directory of one user to a different one only because of the issue in GCB, and I think the right solution is that GCB fixes the root cause.

chanseokoh: Okay, if anyone think this workaround is fine, I am good.

briandealwis: Referencing `$HOME` seems safe.  The corresponding bug [JDK-7069190](https://bugs.openjdk.java.net/browse/JDK-7069190) doesn't look likely to be resolved anytime soon.

chanseokoh: Agreed, "setting home to home" seems reasonable in the Skaffold GCB context.

pkg/skaffold/build/gcb/desc.go
```diff
@@ -24,48 +24,40 @@ import (
...
 }
 
-func (b *Builder) buildSteps(artifact *latest.Artifact, tag string) ([]*cloudbuild.BuildStep, error) {
+func (b *Builder) buildDescription(artifact *latest.Artifact, tag, bucket, object string) (cloudbuild.Build, error) {
+	base := b.baseDescription(bucket, object)
+
 	switch {
 	case artifact.DockerArtifact != nil:
-		return b.dockerBuildSteps(artifact.DockerArtifact, tag)
+		return b.dockerBuildDescription(artifact.DockerArtifact, tag, base)
```
briandealwis: I found this passing `base` a bit puzzling, especially since it wasn't being passed as a pointer.  You're now turning these `xxxBuildSteps()` methods into something that re-casts or morphs the `base` instance for that builder.  Perhaps moving the `cloudbuild.Builder` argument to be first, so that it's a bit more clear that the methods are transforming or rewriting the instance?

(Or could the `xxxBuildDescription()` methods just call `baseDescription` themselves?)

pkg/skaffold/build/gcb/jib_test.go
```diff
@@ -25,70 +25,92 @@ import (
...
-			GoogleCloudBuild: &latest.GoogleCloudBuild{
+		testutil.Run(t, test.description, func(t *testutil.T) {
+			artifact := &latest.Artifact{
+				ArtifactType: latest.ArtifactType{
+					JibMavenArtifact: &latest.JibMavenArtifact{},
+				},
+			}
+
+			builder := newBuilder(latest.GoogleCloudBuild{
 				MavenImage: "maven:3.6.0",
```
briandealwis: Do you think this should be 3.6.1?

dgageot: This is just a test. Could be `foobar`

PR #2567, opened 2019-07-29
skaffold generate-pipeline command

marlon-gamez:
This pr adds a new skaffold command that allows users to generate a Tekton pipeline (https://github.com/tektoncd/pipeline) by reading their skaffold config file and generating corresponding build and deploy tasks. The command also adds a profile to the users skaffold.yaml, which is used by the pipeline when running the tasks on cluster.

In order to use the generated pipeline, the user must run the command, then commit and push the changes to their skaffold config file.

Currently, building on this branch won't work because the tekton api has outdated dependencies, which has been addressed in this pr (https://github.com/tektoncd/pipeline/pull/1117). Building should be fixed once the next release of Tekton pipelines is out.

TODO: 
- add integration tests
- update Tekton dependency when new release occurs


pkg/skaffold/runner/generate_pipeline.go
```diff
@@ -0,0 +1,388 @@
...
+	ReadString(delim byte) (string, error)
+}
+
+func (r *SkaffoldRunner) GeneratePipeline(ctx context.Context, out io.Writer, config *latest.SkaffoldConfig, fileOut string) error {
+	err := createSkaffoldProfile(config)
+	if err != nil {
+		return errors.Wrap(err, "setting up profile")
+	}
+
+	fmt.Println("Generating Pipeline...")
```
priyawadhwa: nit: use color.Default here

pkg/skaffold/runner/generate_pipeline.go
```diff
@@ -0,0 +1,388 @@
...
+	for _, item := range jsons {
+		itemYaml, err := yaml.JSONToYAML(item)
+		if err != nil {
+			return errors.Wrap(err, "converting jsons to yamls")
+		}
+		output.Write(append(itemYaml, []byte("---\n")...))
+	}
+
+	// write all yaml pieces to output
+	if err := ioutil.WriteFile(fileOut, output.Bytes(), 0755); err != nil {
```
priyawadhwa: nit: can simplify to 

```golang
return ioutil.WriteFile(fileOut, output.Bytes(), 0755)
```

pkg/skaffold/runner/generate_pipeline.go
```diff
@@ -0,0 +1,388 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func generateGitResource() (*tekton.PipelineResource, error) {
```
priyawadhwa: I think it would be useful to move all of these helper function into a pkg so the code is easier to find, like `pkg/skaffold/pipeline`

then you could have a file for each, like

`git_resource.go` and `git_resource_test.go` for the unit tests.

pkg/skaffold/runner/generate_pipeline.go
```diff
@@ -0,0 +1,388 @@
...
+	getGitRepo := exec.Command("git", "config", "--get", "remote.origin.url")
+	gitRepo, err := getGitRepo.Output()
+	if err != nil {
+		return nil, errors.Wrap(err, "getting git repo from git config")
+	}
+
+	// Create git resource for pipeline from users current git repo
+	return &tekton.PipelineResource{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "PipelineResource",
```
priyawadhwa: If there's a `constants.go` in tekton you could use that here too!

pkg/skaffold/runner/generate_pipeline.go
```diff
@@ -0,0 +1,388 @@
...
+					{
+						Name: "source",
+						Type: tekton.PipelineResourceTypeGit,
+					},
+				},
+			},
+			Steps: []corev1.Container{
+				{
+					Name:       "run-build",
+					Image:      "gcr.io/k8s-skaffold/skaffold:v0.34.0",
```
priyawadhwa: I'd move this into `constants.go`so that it can be easily updated in the future

pkg/skaffold/runner/generate_pipeline.go
```diff
@@ -0,0 +1,388 @@
...
+					WorkingDir: "/workspace/source",
+					Command:    []string{"skaffold"},
+					Args: []string{"build",
+						"--filename", "skaffold.yaml",
+						"--profile", "oncluster",
```
priyawadhwa: are we assuming that the user will have a profile named `oncluster` in their skaffold.yaml?

pkg/skaffold/runner/generate_pipeline.go
```diff
@@ -0,0 +1,388 @@
...
+					},
+				},
+			},
+		},
+	}, nil
+}
+
+func generateDeployTask(deployConfig latest.DeployConfig) (*tekton.Task, error) {
+	if deployConfig.HelmDeploy == nil && deployConfig.KubectlDeploy == nil && deployConfig.KustomizeDeploy == nil {
+		return nil, errors.New("No Helm/Kubectl/Kustomize deploy config")
```
priyawadhwa: nit: go errors typically start lower case :)

pkg/skaffold/runner/generate_pipeline.go
```diff
@@ -0,0 +1,388 @@
...
+						"--build-artifacts", "build.out",
+					},
+				},
+			},
+		},
+	}, nil
+}
+
+func generatePipeline(tasks []*tekton.Task) (*tekton.Pipeline, error) {
+	if len(tasks) == 0 {
```
priyawadhwa: could this ever happen? or will errors from build/deploy always return first?

pkg/skaffold/runner/generate_pipeline.go
```diff
@@ -0,0 +1,388 @@
...
+			pipelineTask.RunAfter = []string{pipeline.Spec.Tasks[i-1].Name}
+		}
+		pipeline.Spec.Tasks = append(pipeline.Spec.Tasks, pipelineTask)
+	}
+
+	return pipeline, nil
+}
+
+func createSkaffoldProfile(config *latest.SkaffoldConfig) error {
+	fmt.Println("Checking for oncluster skaffold profile...")
```
priyawadhwa: nit: color.Default

pkg/skaffold/runner/generate_pipeline.go
```diff
@@ -0,0 +1,388 @@
...
+	for _, profile := range config.Profiles {
+		if profile.Name == "oncluster" {
+			profileExists = true
+			break
+		}
+	}
+
+	// Check for existing oncluster profile, if none exists then prompt to create one
+	if profileExists {
+		fmt.Println("profile \"oncluster\" found!")
```
priyawadhwa: we could probably log this at the debug level 
`logrus.Debug`

pkg/skaffold/runner/generate_pipeline.go
```diff
@@ -0,0 +1,388 @@
...
+	if err != nil {
+		return errors.Wrap(err, "generating profile \"oncluster\"")
+	}
+
+	bProfile, err := yamlv2.Marshal([]*latest.Profile{profile})
+	if err != nil {
+		return errors.Wrap(err, "marshaling new profile")
+	}
+
+	fileContents, err := ioutil.ReadFile("skaffold.yaml")
```
priyawadhwa: the user may have passed in a different path to their skaffold.yaml, which you can find in the runcontext!

pkg/skaffold/runner/generate_pipeline.go
priyawadhwa: answered my own question by reading the code :sweat_smile:

integration/testdata/generate_pipeline/no_profiles/expectedSkaffold.yaml
```diff
@@ -0,0 +1,22 @@
...
+  artifacts:
+  - image: gcr.io/first-week-244218/getting-started
+deploy:
+  kubectl:
+    manifests:
+    - k8s-*
+profiles:
+- name: oncluster
+  build:
+    artifacts:
```
tejal29: @priyawadhwa is kaniko defined at per image level?
Is there a way to apply "onclluster" profile for all artifacts? Not for this phase, but for phase 2 when we support multiple artifacts, the config will become longer..

pkg/skaffold/runner/generate_pipeline.go
tejal29: yes, we can make this further more better in a follow up PR and ask the user if there is already one exists from the given list. ? 
esp when its a custom builder, it might run on cluser. we just don't know
or just go through profiles and see if "kaniko" section is defined.

pkg/skaffold/runner/generate_pipeline.go
tejal29: i think its `runCtx.Opts.Config`

pkg/skaffold/runner/generate_pipeline.go
tejal29: We shd use the current skaffold version which is stored in `version.Get().Version`
```
fmt.Sprintf(""gcr.io/k8s-skaffold/skaffold:%s", version.Get().Version)
```

pkg/skaffold/runner/generate_pipeline_test.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func TestGenerateBuildTask(t *testing.T) {
+	var tests = []struct {
+		description string
+		buildConfig latest.BuildConfig
+		shouldErr   bool
+	}{
+		{
+			description: "successfully generate build task",
```
tejal29: Add a build task with non docker artifact like `jib`

pkg/skaffold/runner/generate_pipeline_test.go
```diff
@@ -0,0 +1,278 @@
...
+					HelmDeploy: &latest.HelmDeploy{},
+				},
+			},
+			shouldErr: false,
+		},
+		{
+			description: "fail generating deploy task",
+			deployConfig: latest.DeployConfig{
+				DeployType: latest.DeployType{
+					HelmDeploy:      nil,
```
tejal29: you can get rid of nil assignments.

pkg/skaffold/runner/generate_pipeline_test.go
```diff
@@ -0,0 +1,278 @@
...
+func TestGenerateProfile(t *testing.T) {
+	var tests = []struct {
+		description     string
+		skaffoldConfig  *latest.SkaffoldConfig
+		expectedProfile *latest.Profile
+		responses       []string
+		shouldErr       bool
+	}{
+		{
+			description: "successful profile generation",
```
tejal29: add a test when artifact type is `jib-maven` and no profile is added.

tejal29:
@marlon-gamez Can you also run `go mod tidy` to prune unused dependencies?
I dont' think we are using all of `Tekton`. Let me know if you don't see the number of files reduced.

integration/generate_pipeline_test.go
```diff
@@ -0,0 +1,81 @@
...
+		},
+		{
+			description: "existing other profile",
+			dir:         "testdata/generate_pipeline/existing_other",
+			responses:   []byte("y"),
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
```
tejal29: Please add 			if testing.Short() {
				t.Skip("skipping integration test")
			}
			if ShouldRunGCPOnlyTests() {
				t.Skip("skipping test that is not gcp only")
			}

integration/testdata/generate_pipeline/no_profiles/expectedSkaffold.yaml
priyawadhwa: yah, kaniko is defined per artifact. I'm not sure if there's any way to get around a longer config :/

tejal29:
one small nit remaing
```
INFO [loader] SSA repr building timing: packages building 287.10442ms, total 6.187174947s 
204integration/generate_pipeline_test.go:38:3: U1000: field `args` is unused (unused)
205		args        []string
```

PR #965, opened 2018-09-11
Remove service labeling temporarily

dgageot:
Fixes #959

PR #2568, opened 2019-07-30
Bring back applying labels to services deployed with helm

### LINKED ISSUE #2379: Port forwarding broken for helm releases
### Expected behavior

Port forwarding works properly for helm releases.

### Actual behavior

Port forwarding does not occur for helm releases

### Information

- Skaffold version: [latest version](https://github.com/GoogleContainerTools/skaffold/commit/734e5e3aee4611445237fd34f1abd03e36fdfb64) built from master
- Operating system: ubuntu
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta11
kind: Config
build:
  tagPolicy:
    sha256: {}
  artifacts:
  - image: gcr.io/k8s-skaffold/skaffold-helm
deploy:
  helm:
    releases:
    - name: skaffold-helm
      chartPath: skaffold-helm
      #wait: true
      #valuesFiles:
      #- helm-skaffold-values.yaml
      values:
        image: gcr.io/k8s-skaffold/skaffold-helm
      #recreatePods will pass --recreate-pods to helm upgrade
      #recreatePods: true
      #overrides builds an override values.yaml file to run with the helm deploy
      #overrides:
      # some:
      #   key: someValue
      #setValues get appended to the helm deploy with --set.
      #setValues:
        #some.key: someValue
```

1. Run `skaffold dev` with [helm-deployment example](https://github.com/GoogleContainerTools/skaffold/tree/master/examples/helm-deployment) using skaffold built from HEAD
2. No port forward occurs

cc/ @priyawadhwa since this might be related to #2215

tejal29:
Fixes #2379 

In #965, we turned off labelling services due to kubernetes/kubernetes#68087
This bug is in the kube-controller-manager which runs on the cluster master side.
 I tested this PR on my cluster with 1.11.7-gke.4 and port-forwarding for LoadBalancer type services works.

Impacted versions for this bug

|GKE cluster Version | State |  Notes |
------------------|-------|--------|
| v1.12.0 and onwards| Never Occurred | As per the Changelog for v1.12.0 [here](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md#v1120), this bug never appeared in 1.12.0 .  The very first version lists kubernetes/kubernetes#68087 being fixed in notes. See [L1243](https://github.com/kubernetes/kubernetes/blame/master/CHANGELOG-1.12.md#L1243) |
| v1.11.x| Occurred in v1.11.3 and before| This fix was patched in v1.11.4 as per the Changelog for for v1.11.0 [here](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#changelog-since-v1113). As per release notes on [Jan 2019](https://cloud.google.com/kubernetes-engine/docs/release-notes#january-21-2019), all cluster masters running 1.11.2 through 1.11.4 will be upgraded to 1.11.5-gke.5.  which should contain the fix.| 

As per GKE version skew policy,  [June 2019 release notes mention](https://cloud.google.com/kubernetes-engine/docs/release-notes#jun_27_2019) GKE will only support the three newest minor versions. 
So in june end, for all node pools older than v1.11.X  Google will enable node auto-upgrade and these nodes will be updated to 1.11.10-gke.5. ( we are v1.14 now)

In conclusion, i feel this is safe way to enable port-forwarding for services deployed by helm (helm2) to get in.

### DIFF
--- a/integration/dev_test.go
+++ b/integration/dev_test.go
@@ -17,23 +17,19 @@ limitations under the License.
 package integration
 
 import (
-	"bytes"
 	"context"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
-	"regexp"
 	"strings"
 	"testing"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/wait"
-
 	"github.com/GoogleContainerTools/skaffold/integration/skaffold"
 	"github.com/GoogleContainerTools/skaffold/proto"
 	"github.com/GoogleContainerTools/skaffold/testutil"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 func TestDev(t *testing.T) {
@@ -111,17 +107,7 @@ func TestDevAPITriggers(t *testing.T) {
 	client, shutdown := setupRPCClient(t, rpcAddr)
 	defer shutdown()
 
-	// read the event log stream from the skaffold grpc server
-	var stream proto.SkaffoldService_EventLogClient
-	var err error
-	for i := 0; i < readRetries; i++ {
-		stream, err = client.EventLog(context.Background())
-		if err != nil {
-			t.Logf("waiting for connection...")
-			time.Sleep(waitTime)
-			continue
-		}
-	}
+	stream, err := readEventAPIStream(client, t)
 	if stream == nil {
 		t.Fatalf("error retrieving event log: %v\n", err)
 	}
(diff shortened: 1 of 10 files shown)


balopat:
I wouldn't add a warning - if someone sees the issue, we should just ask them to upgrade to 1.11.4+

PR #2573, opened 2019-07-30
docker auth: use GetAllCredentials() to use credHelpers

### LINKED ISSUE #239: Private base images don't work with skaffold properly
Skaffold cannot pull from private registry. Docker daemon can.
OS and docker daemon:
```
root@debbie:~# uname -a
Linux debbie 4.14.0-3-amd64 #1 SMP Debian 4.14.17-1 (2018-02-14) x86_64 GNU/Linux
root@debbie:~# lsb_release  -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux testing (buster)
Release:        testing
Codename:       buster
root@debbie:~# docker -v
Docker version 17.12.1-ce, build 7390fc6
root@debbie:~# 
```
Skaffold:
```
root@debbie:~# skaffold version
v0.2.0
root@debbie:~# 
```

Skaffold output:
```
dcherniv@debbie:~REDACTED$ skaffold run
Starting build...
Found minikube or Docker for Desktop context, using local docker daemon.
WARN[0000] Could not get minikube docker env, falling back to local docker daemon 
Sending build context to Docker daemon  814.1kB
Step 1/22 : FROM REGISTRY/REDACTED/REDACTED:latest
ERRO[0000] run: running skaffold steps: run: build step: running build: unauthorized: The client does not have permission for manifest 
```
Docker output:
```
dcherniv@debbie:~REDACTED$ docker pull REGISTRY/REDACTED/REDACTED:latest
latest: Pulling from REGISTRY/REDACTED/REDACTED
6d987f6f4279: Pull complete 
dab216b6afad: Pull complete 
c9580cbb00a7: Pull complete 
Digest: sha256:340d53a95a520fb94b4eda1718780bd841a7d8ae484181e700163da75dfd742b
Status: Downloaded newer image for REGISTRY/REDACTED/REDACTED:latest
dcherniv@debbie:~/REDACTED$ 
```
After the image is downloaded to local, skaffold run succeeds. Output ommited for brevity.

What gives? my docker config is in ~/.docker/config.json

Auth is base64 encoded:
```
        "auths": {
                "https://REGISTRY": {
                        "auth": "REDACTEDBASE64",
                        "email": "blah@nah.id"
                },
          [....]
        }
```
In private docker registry logs i see requests from skaffold coming in as anonymous/unauthenticated, where with docker pull i see the request having a proper username.
Private registry is running on artifactory.

mattnworb:
The current mechanism for building the map of `AuthConfig` to pass to
the docker daemon when building the image bypasses any credsHelpers in
the user's docker config json file.

`cf.GetCredentialsStore("").GetAll()` ends up returning the `auths`
section from the config file, and does not use or execute any configured
`credHelpers`.

There is a comment in the source code here for a commented-out call to
GetAllCredentials() that it is too slow, but I think it is worth noting
that this is the same function used by a regular `docker build`:
https://github.com/docker/cli/blob/e06530297ddd4d0a4baf4361fc094f4afb36de92/cli/command/image/build.go#L403

While it might be slow to execute each of the configured credHelpers to
be able to build a full map of `AuthConfigs`, this seems like the only
way to pass all of the user's auth configs from their config file to the
build command - otherwise any image pulls needed to build the image will
fail if they require authentication.

Fixes #239.

### DIFF
--- a/pkg/skaffold/docker/auth.go
+++ b/pkg/skaffold/docker/auth.go
@@ -77,9 +77,7 @@ func (credsHelper) GetAllAuthConfigs() (map[string]types.AuthConfig, error) {
 		return nil, errors.Wrap(err, "docker config")
 	}
 
-	// TODO(dgageot): this is really slow because it has to run all the credential helpers.
-	// return cf.GetAllCredentials()
-	return cf.GetCredentialsStore("").GetAll()
+	return cf.GetAllCredentials()
 }
 
 func (l *localDaemon) encodedRegistryAuth(ctx context.Context, a AuthConfigHelper, image string) (string, error) {


dgageot:
@mattnworb I'll see what the details are. I hate this behaviour of sending all credentials to docker build no matter what it requires. We had the plan to fix that at Docker a few years back. I'll check if we can do better

mattnworb:
@dgageot yeah I agree it is unfortunate, especially considering that the `docker/cli` code used by skaffold doesn't allow for any caching of tokens, etc. I'm surprised that the CLI doesn't do anything smarter, like try to figure out which images are needed by the Dockerfile, but I would imagine that would be quite a stretch for skaffold to do if the underlying docker/cli code depended on here does not support it.

pkg/skaffold/docker/auth.go
```diff
@@ -77,9 +77,7 @@ func (credsHelper) GetAllAuthConfigs() (map[string]types.AuthConfig, error) {
 		return nil, errors.Wrap(err, "docker config")
 	}
 
-	// TODO(dgageot): this is really slow because it has to run all the credential helpers.
-	// return cf.GetAllCredentials()
```
tejal29: i think the reason why this was commented out is because it is slow.

mattnworb: I mentioned this in the PR summary - while it is slower than not getting all credentials, it would also mean that the image build is not broken if it needs to pull from a private repo. It would arguably be better to be slower but functional for more use cases. 

Additionally it should be no slower than a ‘docker build’.

tejal29:
Merging this in and opened a TODO #2625

PR #2579, opened 2019-07-31
[linters] Use vendored dependencies. Don’t download them.

dgageot:
This behaviour can’t be set on the command line. We have to use a config file.
Also, removed the `--skip-dirs vendor` because it’s done by default by golangci-lint.


dgageot:
FYI, I'm going to self merge this one to fix `master` branch

PR #2582, opened 2019-07-31
refactor: Add upgrade utility to handle all pipelines in a SkaffoldConfig

### LINKED ISSUE #2469: Schema upgrade mechanism fails to upgrade profiles by default
Currently, the schema upgrade mechanism works by implementing the `VersionedConfig` interface:
```go
type VersionedConfig interface {
	GetVersion() string
	Upgrade() (VersionedConfig, error)
}
```
The responsibility of `Upgrade` is to migrate the whole config from one version to the next. This opens the possibility for a class of bugs, where the main pipeline (`SkaffoldConfig.Pipeline`) is migrated to a new version, but none of the pipeline configs in the profiles `SkaffoldConfig.Profiles[*].Pipeline` (see for example #2415). If there are use-cases where profile-pipelines need to be upgraded differently than the main pipeline, the following suggestion is void.

Suggestion
----

Add a new interface
```go
type SkaffoldPipeline interface {
	Upgrade() (SkaffoldPipeline, error)
}
```
which upgrades a given pipeline to the next version. In addition, add an upgrade helper which migrates the main pipeline _and_ all profile pipelines to the next version.
This will be backwards compatible with skaffold schema versions before the `Pipeline` structure was introduced.

corneliusweig:
Before, the main pipeline and the profile pipelines had to be upgraded
manually, which was easy to forget.

In contrast to what is outlined in #2469, I didn't introduce a new interface. This is mainly because I want to avoid double json-cloning of the same objects. Instead, I opted for reflection which has the main drawback that it requires the exact same go-naming for `Pipeline` and `Profiles[*].Pipeline`. Given that this naming scheme is rather stable now, I think this is affordable.

Fixes #2469

### DIFF
--- a/pkg/skaffold/schema/util/upgrade_pipelines.go
+++ b/pkg/skaffold/schema/util/upgrade_pipelines.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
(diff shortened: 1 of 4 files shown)


corneliusweig:
@nkubala Thanks for reviewing. My thinking was that this code is tested implicitly by the existing config upgrade tests, but it's better to make this explicit :). Do you think what I added is enough for that?

PR #2589, opened 2019-07-31
Fix handling of whitelisted directories in dockerignore

### LINKED ISSUE #2576: .dockerignore whitelist rules not respected
### Expected behavior

Skaffold calls `docker build` directly so that .dockerignore file is applied as expected, including whitelist rules like

```
**

!/.gitmodules
!/CMakeLists.txt
!/etc/**
!/CPlusPlusCode/**
!/submodules/**
```

which is the recommended approach – only sync what belongs to the build to avoid rebuilds.

### Actual behavior

In the above .dockerignore example, only

```
.gitmodules
CMakeLists.txt
Dockerfile
```

get transmitted in the Docker context. Far from expectation.

### Information

- Skaffold version: 0.34.1
- Operating system: macOS
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta13
kind: Config

build:
  artifacts:
    - image: skaffold-dockerignore-issue
      docker: {}

  tagPolicy:
    gitCommit:
      variant: AbbrevCommitSha
```

See full example project at: https://github.com/AndiDog/skaffold-dockerignore-issue

### Why this is important

Because builds are different in Skaffold vs. raw Docker CLI, and the good-practice whitelisting approach in .dockerignore is fully unusable.

corneliusweig:
This PR fixes handling of whitelisted directories in `.dockerignore`.

The code closely follows the reference implementation from
https://github.com/docker/cli/blame/f019bdcace678c71690a99f6c5f30303c33534bc/vendor/github.com/docker/docker/pkg/archive/archive.go#L832  Does that already mandate a proper attribution?

Fixes #2576

### DIFF
--- a/pkg/skaffold/docker/dependencies_test.go
+++ b/pkg/skaffold/docker/dependencies_test.go
@@ -357,7 +357,7 @@ func TestGetDependencies(t *testing.T) {
 			description: "dockerignore with context in parent directory",
 			dockerfile:  copyDirectory,
 			workspace:   "docker/..",
-			ignore:      "bar\ndocker/*\n*.go",
+			ignore:      "bar\ndocker\n*.go",
 			expected:    []string{".dot", "Dockerfile", "file", "test.conf"},
 		},
 		{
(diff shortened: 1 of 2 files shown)


priyawadhwa:
Thanks for tackling this @corneliusweig ! Did you get a chance to try the fix on the sample repo provided in the bug?

corneliusweig:
@priyawadhwa Yes, it runs without error. I also added test-cases to cover these scenarios.

pkg/skaffold/docker/dependencies_test.go
```diff
@@ -357,7 +357,7 @@ func TestGetDependencies(t *testing.T) {
 			description: "dockerignore with context in parent directory",
 			dockerfile:  copyDirectory,
 			workspace:   "docker/..",
-			ignore:      "bar\ndocker/*\n*.go",
+			ignore:      "bar\ndocker\n*.go",
```
corneliusweig: This change means the same, but increases test coverage.

Issue #2503, opened 2019-07-19
Refactor GetAvailablePort to PortBroker


priyawadhwa:
Currently, we use [GetAvailablePort](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/util/port.go#L37) in two places:
1. To get local ports for port-forwarding
2. To get a port for the event API ([here](https://github.com/GoogleContainerTools/skaffold/blob/299c21389bb9cf413e494e334fb499cb7c543a74/pkg/skaffold/server/server.go#L57))

We pass in a generic map interface to work with both of these, but this is unnecessary since `server.go` passes in an empty map as a placeholder.

We should refactor this function and create a type `PortBroker` which will handle getting and storing local ports for all of skaffold, so that multiple parts of the code using local ports won't step on each others toes.

This should also fix the issue of copied unit tests from #2593 in `entry_manager_test.go` and `port_test.go`

tstromberg:
I'm closing this issue as it hasn't seen activity in awhile, and it's unclear if this issue still exists. If this issue does continue to exist in the most recent release of Skaffold, please feel free to add a comment and we'll re-open it.

If someone sees a similar issue, please create a new issue, but do include a link to this issue if possible.

Thank you for opening the issue!

PR #2593, opened 2019-08-01
Cleaner kubectl `port-forward` retry logic

balopat:
This clears up the port forwarding logic on the kubectlForwarder level.  
I got rid of the two polling wait loops, instead there is one single loop that 
- runs until the port forward entry gets cancelled by skaffold itself. 
- it monitors the kubectl port-forward logs in one separate go routine and kills the process in case of known errors - new known error introduced: `error upgrading connection` 
- It reruns `kubectl port-forward` in case the process exits on failure (or gets killed by the log monitoring go routine) 
- pod-running-timeout is now more aggressive, 1 second for more responsive UX

Also, added logic to handle port collision right before kicking off the port forwarding process. If that happens, no new port is being brokered, in the hope that the user will kill the (assumed) external process. In case the colliding process dies, the port forwarding will resume (with a nice recovery message to the user). 

Also, we found a bug in the `portForwarder` map `LoadOrStore` function was not storing anything. This is fixed now with a copy paste of the unit test that was testing the port brokering logic on `sync.Map`. This copy paste should go away when we implement #2503. 

~Fixed the retry logic for the `TestDevPortForwardGKELoadBalancer` integration test, plus fixed the lingering portforwarding processes by sending SIGTERM to the `skaffold dev` process instead of just killing it.~

### DIFF
--- a/pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
+++ b/pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
@@ -55,10 +55,13 @@ func (k *KubectlForwarder) Forward(parentCtx context.Context, pfe *portForwardEn
 func (k *KubectlForwarder) forward(parentCtx context.Context, pfe *portForwardEntry) {
 	var notifiedUser bool
 	for {
-		if parentCtx.Err() == context.Canceled {
-			logrus.Debugf("port forwarding %v cancelled...", pfe)
+		pfe.terminationLock.Lock()
+		if pfe.terminated {
+			logrus.Debugf("port forwarding %v was cancelled...", pfe)
 			return
 		}
+		pfe.terminationLock.Unlock()
+
 		if !util.IsPortFree(pfe.localPort) {
 			//assuming that Skaffold brokered ports don't overlap, this has to be an external process that started
 			//since the dev loop kicked off. We are notifying the user in the hope that they can fix it
@@ -79,6 +82,7 @@ func (k *KubectlForwarder) forward(parentCtx context.Context, pfe *portForwardEn
 			pfe.cancel()
 		}
 		pfe.cancel = cancel
+
 		cmd := k.kubectl.Command(ctx,
 			"port-forward",
 			"--pod-running-timeout", "1s",
@@ -119,9 +123,13 @@ func (k *KubectlForwarder) forward(parentCtx context.Context, pfe *portForwardEn
 func (*KubectlForwarder) Terminate(p *portForwardEntry) {
 	logrus.Debugf("Terminating port-forward %v", p)
 
+	p.terminationLock.Lock()
+	defer p.terminationLock.Unlock()
+
 	if p.cancel != nil {
 		p.cancel()
 	}
+	p.terminated = true
 }
 
 // Monitor monitors the logs for a kubectl port forward command
--- a/pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
+++ b/pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
@@ -20,22 +20,19 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"net"
 	"strings"
 	"time"
 
 	"github.com/pkg/errors"
+
 	"github.com/sirupsen/logrus"
-	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubectl"
-	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 )
 
 type EntryForwarder interface {
-	Forward(parentCtx context.Context, pfe *portForwardEntry) error
+	Forward(parentCtx context.Context, pfe *portForwardEntry, retryFunc func())
 	Terminate(p *portForwardEntry)
-	Monitor(*portForwardEntry, func())
 }
 
 type KubectlForwarder struct {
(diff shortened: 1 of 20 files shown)


balopat:
I am missing unit tests + there is a scenario that is not checked now (not sure if it was covered before): kubectl portforward does not fail if the port is already bound :( 
I managed to create a little adversarial scenario where a competing process takes the port and kubectl port forward just runs in the background happily not throwing any errors.

pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
```diff
@@ -107,18 +127,29 @@ func (*KubectlForwarder) Terminate(p *portForwardEntry) {
...
+			return
+		default:
+			time.Sleep(1 * time.Second)
+			s, _ := buf.ReadString(byte('\n'))
+			if s != "" {
+				logrus.Tracef("[port-forward] %s", s)
+
+				if strings.Contains(s, "error forwarding port") ||
+					strings.Contains(s, "unable to forward") ||
+					strings.Contains(s, "error upgrading connection") {
```
briandealwis: Would it make sense to log a warning or something so the user knows the retry is happening and why?

balopat: it is there in trace mode

pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
```diff
@@ -20,105 +20,144 @@ import (
...
+// It kills the command on errors in the kubectl port-forward log
+// It restarts the command if it was not cancelled by skaffold
+// It retries in case the port is taken
+func (k *KubectlForwarder) Forward(parentCtx context.Context, pfe *portForwardEntry) {
+	go k.forward(parentCtx, pfe)
+}
+
+func (k *KubectlForwarder) forward(parentCtx context.Context, pfe *portForwardEntry) {
+	var notifiedUser bool
+	for {
```
tejal29: when reading this, it looks like this `forward` runs forever?

balopat: it does, until the pfe gets cancelled (there are 3 return statements in the body, all around cancellation!)

pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
balopat: I keep thinking about what you mentioned: "error upgrading connection" is from port forwarding in client-go: https://github.com/kubernetes/client-go/blob/master/tools/portforward/portforward.go#L194 - I _think_ it is fine to retry on it...

pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
```diff
@@ -20,105 +20,144 @@ import (
...
+			if s != "" {
+				logrus.Tracef("[port-forward] %s", s)
+
+				if strings.Contains(s, "error forwarding port") ||
+					strings.Contains(s, "unable to forward") ||
+					strings.Contains(s, "error upgrading connection") {
+					// kubectl is having an error. retry the command
+					logrus.Tracef("killing port forwarding %v", p)
+					if err := cmd.Process.Kill(); err != nil {
+						logrus.Tracef("failed to kill port forwarding %v, err: %s", p, err)
```
tejal29: Can we also print the pid for the process that did not get killed?
Users can kill it themselves.

pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
tejal29: got lost in reading and saw only 1 return statement.
I still feel, all the 3 return statements are waiting for conditions that are
1. pfe.terminated. -  not sure when that happens
2. & 3. user enter ^C in `skaffold dev` .

Will it keep trying forever untill above two conditions are satisfied?

pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
balopat: the typical case that I saw this error happen is when the process is already killed...it is really just a trace for the sake of completeness, I don't really think that the PID is necessary.

pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
balopat: well, if things are going well, it is going to wait in `cmd.Wait()` as long as kubectl portforward is running ...

balopat: but yes, it does retry unless it is an explicit cancel from the skaffold process...

balopat: Something like this: 

image

balopat: btw this was generated on planttext.com :) 
``` 
@startuml

title Port Forward State Model
[*] --> CheckForTermination
CheckForTermination -down-> Cancelled
CheckForTermination -down-> TryCmdStart
TryCmdStart -down-> Cancelled
TryCmdStart --> CheckForTermination
TryCmdStart -down-> LogMonitoring

Cancelled -down-> [*]
state "kubectl port-forward logs are monitored" as LogMonitoring {
  TryCmdWait -down-> Cancelled
  TryCmdWait -up-> CheckForTermination
}

@enduml

```

balopat: I wish I could put labels on the arrows - but the key thing is non-cancellation error scenarios go back to start, cancellation exists, no error continues to next step.

priyawadhwa:
cc @balopat looks like there was a prune error now that we are cleaning up on `TestEvent` -- logs [here](https://gist.github.com/priyawadhwa/c5adb77efef7ec3f793469b1f85d1027).

I think the issue is that we first run `TestEvent/v1/event_log` which [builds](https://gist.github.com/priyawadhwa/c5adb77efef7ec3f793469b1f85d1027#file-testevent-prune-error-L20) image with digest `77ae371cff1b`. This is then [cleaned up](https://gist.github.com/priyawadhwa/c5adb77efef7ec3f793469b1f85d1027#file-testevent-prune-error-L43) by prune. We then try to run `TestEvent/v1/events`, which builds the same image `77ae371cff1b`, [claims it's in the cache](https://gist.github.com/priyawadhwa/c5adb77efef7ec3f793469b1f85d1027#file-testevent-prune-error-L95-L97), but then actually fails to push the image because it doesn't exist. 

I added `--no-prune=true` to get around this.

balopat:
Merging this to release branch.

pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go
briandealwis: I'm pretty sure Cloud Code doesn't show trace — it would be good for users to at least be able to discover that there was a failure.  

I think restarting on error makes a lot of sense: it's essentially turning `kubectl port-forward` into a single-use attempt.  Perhaps this code could look for the more general error header (`portforward.go:.*error occurred`) and log the error but suppress when one of these strings (known issues)?

PR #2594, opened 2019-08-01
Integration test command

daddz:
I have put some more work into this as mentioned in #2561.

For demonstration purposes I created another repository that makes use of these changes over here: https://github.com/daddz/skaffold-integration

When running the `integrationtest` command it will try to find the first pod that matches the defined `podSelector` and then executes the specified `testCommand` via `kubectl exec` in that pod.


daddz:
@tejal29 thanks for the feedback!
I addressed your comments and create a design proposal: https://github.com/GoogleContainerTools/skaffold/pull/2671

PR #1451, opened 2019-01-11
Use kubectl to read the manifests

dgageot:
+ Adds support for json manifests
 + Should help when kustomize is built into kubectl

Fixes #921

Signed-off-by: David Gageot <david@gageot.net>

PR #2595, opened 2019-08-01
Remove obsolete `remoteManifests` field

corneliusweig:
I looks like `remoteManifests` has been combined with `manifests` in #1451 (commit 160cfa11fce8cc70537f5beb294932079b8f7bee).
The field is not used anywhere, so we can break nobody by finally removing it.


corneliusweig:
Sorry, wasn't aware of #2160 and #2258

PR #2608, opened 2019-08-02
merging back release/v0.35.0

balopat:


balopat:
We can ignore the cancelled `push` check. See #2607 for explanation. It shows failing build as I cancelled that build.

PR #2614, opened 2019-08-06
Watch all artifact workspaces, including those outside of the working directory

### LINKED ISSUE #2613: Watcher doesn't detect changes when artifacts are outside the directory from where skaffold is run
### Expected behavior

When a file is changed in a workspace (directory specified in `build.artifacts[].context` option), a rebuild should be triggered automatically, even if the workspace is located outside the directory from where `skaffold dev` is run.

### Actual behavior

File change doesn't trigger a rebuild in this case. Note that this used to work but not sure in which version of Skaffold this got broken.

### Information

- Skaffold version: v0.35.0
- Operating system: macOS 10.14.6
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta13
kind: Config
build:
  artifacts:
  - image: myapplication
    context: ../../../../myapplication
deploy:
  kubectl:
    manifests:
    - ../../apps/myapplication.yaml
```

Run `skaffold dev` from the directory which has the `skaffold.yaml` file above. Note that the artifact context is located in a different directory which is not under the directory which contains `skaffold.yaml`.

Note that if I instead run skaffold from the `myapplication` directory, pointing to the `skaffold.yaml` using the `-f` option (and changing the paths accordingly in the config), file watching works fine.

However that's not an appropriate workaround as I have a lot of different artifacts that I want to be able to build and watch, while maintaining the `skaffold.yaml` config in a different directory.

And as I mentioned above, this used to work but got broken in some later version of Skaffold.

priyawadhwa:
This should fix #2613: Watcher doesn't detect changes when artifacts are outside the directory from where skaffold is run

### DIFF
--- a/pkg/skaffold/trigger/triggers.go
+++ b/pkg/skaffold/trigger/triggers.go
@@ -48,7 +48,7 @@ func NewTrigger(runctx *runcontext.RunContext) (Trigger, error) {
 			Interval: time.Duration(runctx.Opts.WatchPollInterval) * time.Millisecond,
 		}, nil
 	case "notify":
-		return newFSNotifyTrigger(runctx)
+		return newFSNotifyTrigger(runctx), nil
 	case "manual":
 		return &manualTrigger{}, nil
 	default:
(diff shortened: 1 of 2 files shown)


priyawadhwa:
thanks @nkubala , done

PR #2622, opened 2019-08-08
Fix kaniko permissions with generate-pipeline command

marlon-gamez:
Changes functions used by generate-pipeline command to only add kaniko information to profile artifacts when necessary. Also adds GOOGLE_APPLICATION_CREDENTIALS env variable to tekton tasks when an artifact is built with kaniko.

Updated integration tests to match the new pipeline output.

### DIFF
--- a/pkg/skaffold/generate_pipeline/constants.go
+++ b/pkg/skaffold/generate_pipeline/constants.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package generatepipeline
 
 const (
--- a/pkg/skaffold/generate_pipeline/constants.go
+++ b/pkg/skaffold/generate_pipeline/constants.go
@@ -1,0 +1,5 @@
+package generatepipeline
+
+const (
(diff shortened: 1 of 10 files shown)


pkg/skaffold/generate_pipeline/generate_pipeline.go
```diff
@@ -139,7 +139,33 @@ func generateBuildTask(buildConfig latest.BuildConfig) (*tekton.Task, error) {
 		},
 	}
 
-	return pipeline.NewTask("skaffold-build", resources, steps), nil
+	var volumes []corev1.Volume
+	if buildConfig.Artifacts[0].KanikoArtifact != nil {
+		volumes = []corev1.Volume{
+			{
+				Name: "kaniko-secret",
```
tejal29: lets use constants everywhere for kaniko.
Define constant in the local pkg scope.

pkg/skaffold/generate_pipeline/profile.go
```diff
@@ -32,55 +32,49 @@ import (
...
-	if profileExists {
-		color.Default.Fprintln(out, "profile \"oncluster\" found!")
-		return nil
-	}
-
 confirmLoop:
 	for {
 		color.Default.Fprintf(out, "No profile \"oncluster\" found. Create one? [y/n]: ")
 		response, err := reader.ReadString('\n')
 		if err != nil {
```
tejal29: Not in this PR, but later you can add a func to re-try getting user input 3 times before failing.
It would be great if you create an issue in github and assign it your self with priority P2/

pkg/skaffold/generate_pipeline/constants.go
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package generatepipeline
+
+const (
+	defaultKanikoSecretName = "kaniko-secret"
```
tejal29: can we just rename this to `kanikoSecretName` since "default" suggest there is way to override. Which i don't think it is, right?

Issue #2596, opened 2019-08-01
Add port forwarding unit & integration tests


priyawadhwa:
Add integration tests for #2593 to cover:

1. (**Done with #2623**) deleting a pod after port forwarding has occurred
1. changing the skaffold config after a dev loop
1. (**Done with #2543**)making sure port forwarding works on the second dev loop

**(Done with - #2648)** Also, add constructor for `portForwardEntry`

Add unit tests for:
1. **(Done with #2661)** kubectl forwarder
1. **(Done with #2649)** loadorstore

PR #2623, opened 2019-08-09
Add port forwarding integration test

priyawadhwa:
This tests makes sure that port forwarding works as expected. It then
deletes all pods, and makes sure that port forwarding comes back
upon the same port.

First of a few tests for #2596


integration/port_forward_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubectl"
+	kubectx "github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/context"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/portforward"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner/runcontext"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+)
+
+func TestPortForward(t *testing.T) {
```
priyawadhwa: This test was copied and pasted from the deleted file `white_box_test.go`

integration/port_forward_test.go
```diff
@@ -0,0 +1,137 @@
...
+		"delete",
+		"pods", "--all",
+		"-n", ns.Name,
+	)
+
+	if output, err := killPodsCmd.CombinedOutput(); err != nil {
+		t.Fatalf("error deleting all pods: %v \n %s", err, string(output))
+	}
+	// port forwarding should come up again on the same port
+	assertResponseFromPort(t, localPort, constants.LeeroyAppResponse)
```
tejal29: Can you run this test 100 times and make sure there are no flakes. 

```
"INTEGRATION_TEST_ARGS="-test.run  ^TestPortForwardDeletePod --test.count 100" make integration

File: integration/dev_test.go [hunk 3]
```diff
@@ -270,43 +272,62 @@ func TestDevPortForwardGKELoadBalancer(t *testing.T) {
...
+func assertResponseFromPort(t *testing.T, port int, expected string) {
+	ctx, cancelTimeout := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancelTimeout()
+
+	for {
+		select {
+		case <-ctx.Done():
+			t.Fatalf("Timed out waiting for response from port %d", port)
+
+		default:
```
tejal29: you could also change this too
```
select {
		case <-ctx.Done():
                     t.Fatalf("Timed out waiting for response from port %d", port)
                case <- time.After(1* time.Sleep):

```
You can also create a timer and re-use (by resetting) the same one to avoid creating new channel `time.After`

integration/port_forward_test.go
priyawadhwa: Good idea -- for some reason my laptop gets really slow when running tests, so I ran 20 (~20 min) and it passed against Minikube.

PR #2638, opened 2019-08-12
Link task resources in generate-pipeline output

marlon-gamez:
Adds connection of resources from build to deploy task. Removes the need for runAfter. Instead, the task order is determined resource dependency. This allows the deploy task to use the build.out file that is generated by the build task.


marlon-gamez:
close & reopened to get travis to rerun a flaky test

PR #2640, opened 2019-08-12
Collect namespaces of deployed resources.

tejal29:
In #2495, @corneliusweig noticed for a skaffold run without `--namespace` flag, we make a call for deployment health-check and port-forwarding 
```
	})
```
and `ns` is `""`.
Similarly, during port-forwarding,  we do the same
```
services, err := clientset.CoreV1().Services("").List(metav1.ListOptions{
		LabelSelector: label,
	})
```
In `context.RunContext`, we already gather namespaces set in the kubecontext  as well as on command line. 
This PR is collect all the namespaces from `kubectl manifests` and `helm releases `if they are set individually in the manifests.

This PR is not enough for the bug fix for #2495. We need to followup with a PR where if runcontext.Namespaces is empty, we need to use "default" namespace.


pkg/skaffold/deploy/kubectl/namespaces.go
```diff
@@ -0,0 +1,70 @@
...
+// CollectNamespaces returns all the namespaces in the manifests.
+func (l *ManifestList) CollectNamespaces() ([]string, error) {
+	replacer := newNamespaceCollector()
+
+	if _, err := l.Visit(replacer); err != nil {
+		return nil, errors.Wrap(err, "collecting namespaces")
+	}
+	namespaces := make([]string, len(replacer.namespaces))
+	i := 0
+	for ns := range replacer.namespaces {
```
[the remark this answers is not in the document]
tejal29: this is map so its a key value pair.
I will use `append`. I first thought it allocates a new memory every time and will be inefficient. 
But turns out, it only allocates a new memory if the size is less than what was allocated. 
```
namespaces := make([]string, len(replacer.namespaces))

for  ns := range replacer.namespaces {
   namespaces = append(namespaces, ns)
}
```

pkg/skaffold/deploy/helm.go
```diff
@@ -70,18 +71,25 @@ func (h *HelmDeployer) Labels() map[string]string {
 	}
 }
 
-func (h *HelmDeployer) Deploy(ctx context.Context, out io.Writer, builds []build.Artifact, labellers []Labeller) error {
+func (h *HelmDeployer) Deploy(ctx context.Context, out io.Writer, builds []build.Artifact, labellers []Labeller) *Result {
 	var dRes []Artifact
 
 	event.DeployInProgress()
+	nsMap := map[string]bool{}
```
priyawadhwa: I think `map[string]struct{}` is the recommended way to have a set in Go, since `struct{}` doesn't take up any memory!

pkg/skaffold/deploy/deploy.go
```diff
@@ -39,3 +39,28 @@ type Deployer interface {
...
+
+func NewDeploySuccessResult(namespaces []string) *Result {
+	return &Result{namespaces: namespaces}
+}
+
+func (d *Result) Namespaces() []string {
+	return d.namespaces
+}
+
+func (d *Result) IsError() (bool, error) {
```
priyawadhwa: nit: I feel like I typically see "OK bools" returned as the second argument-- 

for example, with maps:
```go
val, ok := myMap[key]
```

PR #2651, opened 2019-08-14
Don't look up services in all namespaces.

### LINKED ISSUE #2495: Skaffold dev fails when user cannot list deployments at cluster scope

tejal29:
fixes #2495, 

In #2640, i made changes to collect all namespaces for deployed resources.
In PR, we go through all namespaces in the context and then look for services in each namespace.

### DIFF
--- a/pkg/skaffold/kubernetes/portforward/resource_forwarder_test.go
+++ b/pkg/skaffold/kubernetes/portforward/resource_forwarder_test.go
@@ -326,6 +326,20 @@ func TestRetrieveServices(t *testing.T) {
 					Spec: v1.ServiceSpec{Ports: []v1.ServicePort{{Port: 8080}}},
 				},
 			},
+		}, {
+			description: "services present but does not expose any port",
+			namespaces:  []string{"test"},
+			services: []*v1.Service{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "svc1",
+						Namespace: "test",
+						Labels: map[string]string{
+							deploy.RunIDLabel: "9876-6789",
+						},
+					},
+				},
+			},
 		},
 	}
 
--- a/pkg/skaffold/kubernetes/portforward/resource_forwarder_test.go
+++ b/pkg/skaffold/kubernetes/portforward/resource_forwarder_test.go
@@ -350,7 +350,7 @@ func TestRetrieveServices(t *testing.T) {
 				objs[i] = s
 			}
 			client := fakekubeclientset.NewSimpleClientset(objs...)
-			t.Override(&kClientSet, mockClient(client))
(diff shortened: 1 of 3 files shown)


tejal29:
hmm, for user defined port forwarding, why do we need to fetch it? We should be able to get the specific resource right?

priyawadhwa:
ah that's true, nevermind!

pkg/skaffold/kubernetes/portforward/resource_forwarder_test.go
```diff
@@ -252,3 +259,101 @@ func TestUserDefinedResources(t *testing.T) {
...
+						Namespace: "test",
+						Labels: map[string]string{
+							deploy.RunIDLabel: "9876-6789",
+						},
+					},
+					Spec: v1.ServiceSpec{Ports: []v1.ServicePort{{Port: 8080}}},
+				},
+			},
+		}, {
+			description: "services present but does not expose any port",
```
priyawadhwa: is this possible for a service?

tejal29: you wont normally do it. But yes, you can not have any ports defined in the Spec and the service is created.

```
$ kubectl apply -f microservices/leeroy-app/kubernetes/
service/leeroy-app created
deployment.apps/leeroy-app created

$ git diff microservices/leeroy-app/kubernetes/
diff --git a/examples/microservices/leeroy-app/kubernetes/deployment.yaml b/examples/microservices/leeroy-app/kubernetes/deployment.yaml
index 56ef7152..46feab6d 100644
--- a/examples/microservices/leeroy-app/kubernetes/deployment.yaml
+++ b/examples/microservices/leeroy-app/kubernetes/deployment.yaml
@@ -6,9 +6,6 @@ metadata:
     app: leeroy-app
 spec:
   clusterIP: None
-  ports:
-    - port: 50051
-      name: leeroy-app
   selector:
     app: leeroy-app
 ---
tejaldesai@@examples (move_resolve_namespace)$ 
```

PR #2685, opened 2019-08-19
Log a warning and rebuild if needed when caching fails

### LINKED ISSUE #2677: --cache-artifacts fails when there is no local docker daemon
I am trying to run skaffold debug mode but it fails with port-forwarding failure. I am using latest Cloud code plugin and a sample java application, mvn is builder used, Intellij version 19.x.x

### Expected behavior
Should be able to put breakpoints in live application
### Actual behavior
Not able to put breakpoints
### Information

- Skaffold version: 35( will uprade to latest and test)
- Operating system: Windows
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta11
kind: Config
build:
  artifacts:
  - image: ....../kube/test/jserver
    jibMaven:
      args: ["-f", "java-hello-world/pom.xml"]
    context: .
# defines the Kubernetes manifests to deploy on each run
deploy:
  kubectl:
    manifests:
    - C:\Users\prashant\cloud-code-samples-master\java\java-hello-world\kubernetes-manifests\hello.deployment.yaml

```

deployment yaml
```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: simplehttp
  namespace: 103000-pra-dev
spec:
  selector:
    matchLabels:
      app: simplehttp
  replicas: 1
  template:
    metadata:
      labels:
        app: simplehttp
    spec:
      containers:
      - name: simplehttp
        image: ......kube/test/jserver:latest
        ports:
        - containerPort: 8080
        env:
        - name: PORT
          value: "8080"
      securityContext:
        runAsUser: 199
      imagePullSecrets:
      - name: regcred
---
apiVersion: v1
kind: Service
metadata:
  name: simplehttp
  namespace: ns-pra-dev
  labels:
    app: simplehttp
spec:
  ports:
  - port: 80
    targetPort: 8080
    protocol: TCP
    name: http
  selector:
    app: simplehttp

---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    ingress.kubernetes.io/ssl-redirect: "true"
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
  generation: 1
  labels:
    app: simplehttp
  name: skaffold-example-ingress
  namespace: ns-pra-dev
  resourceVersion: "14651061"
spec:
  rules:
  - host: skaffold-example1.apps.net
    http:
      paths:
      - backend:
          serviceName: simplehttp
          servicePort: 8080
        path: /

  tls:
  - hosts:
    - skaffold-example1.apps.net
status:
  loadBalancer:
    ingress:
    - {}

```

1. C:\Users\prashant\go\src\github.com\GoogleContainerTools\skaffold.exe debug --filename java-hello-world/skaffold.yaml
--label ide=idea --label ideVersion=2019.1.3.0.0 --label ijPluginVersion=unknown --port-forward=true -v=debug

priyawadhwa:
We should never error out if caching fails. Instead, log a warning and rebuild the artifact if it still needs to be rebuilt.

Should fix #2677 

cc @balopat @briandealwis @prary

### DIFF
--- a/pkg/skaffold/build/cache/retrieve.go
+++ b/pkg/skaffold/build/cache/retrieve.go
@@ -64,12 +64,12 @@ func (c *cache) Build(ctx context.Context, out io.Writer, tags tag.ImageTags, ar
 		switch result := result.(type) {
 		case failed:
 			logrus.Warnf("error checking cache, caching may not work as expected: %v", result.err)
-			color.Red.Fprintln(out, "Error checking cache. Rebuilding.")
+			color.Yellow.Fprintln(out, "Error checking cache. Rebuilding.")
 			needToBuild = append(needToBuild, artifact)
 			continue
 
 		case needsBuilding:
-			color.Red.Fprintln(out, "Not found. Building")
+			color.Yellow.Fprintln(out, "Not found. Building")
 			hashByName[artifact.ImageName] = result.hash
 			needToBuild = append(needToBuild, artifact)
 			continue
--- a/pkg/skaffold/build/cache/retrieve.go
+++ b/pkg/skaffold/build/cache/retrieve.go
@@ -28,6 +28,7 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/event"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/skaffold/build/cache/retrieve.go
```diff
@@ -62,7 +63,10 @@ func (c *cache) Build(ctx context.Context, out io.Writer, tags tag.ImageTags, ar
 		result := results[i]
 		switch result := result.(type) {
 		case failed:
-			return nil, errors.Wrap(result.err, "checking cache")
+			logrus.Warnf("error checking cache, caching may not work as expected: %v", result.err)
+			color.Red.Fprintln(out, "Error checking cache. Rebuilding.")
```
balopat: Maybe not red just orange? 
```suggestion
			color.Orange.Fprintln(out, "Error checking cache. Rebuilding.")
```

priyawadhwa: We use red when we can't [find](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/cache/retrieve.go#L68) the image and rebuild, so I figured it would look better to do the same here too. WDYT?

briandealwis: Since caching is enabled by default, and it doesn't result in an error, red doesn't seem to be warranted.

priyawadhwa: SGTM

PR #2702, opened 2019-08-22
prototyping web UI for Event & Control API

balopat:
As a method to test Event/Control API - this PR adds a prototype UI for browsing events + state and links for the portforwarded services and the Control API. It is reachable on http://localhost:50052/dash (or whatever port you define with `--rpc-http-port`). 

To develop locally, build with `env DEV_DASH=true make install` and then the changes of the filesystem will be reflected immediately. The site is completely client side, using the Event/Control API json data as backend. The site is located under `pkg/skaffold/server/public`.

I decided to temporarily rely on my fork of rakyll/statik to do the go embedding - until https://github.com/rakyll/statik/pull/86/files gets resolved.


balopat:
I'm going to close this for now and rework it later.

PR #2703, opened 2019-08-22
Fix call to newPortForwardEntry constructor in kubectl_forwarder_test

apaz037:
### Objective:
- Fix broken kubectl_forwarder_test

Parameters for constructor were missing ownerReference.
added empty string as owner reference in calls to fix failing tests.


apaz037:
@googlebot I signed it!

PR #2711, opened 2019-08-22
Update cache-artifacts option usage language to reflect new default

demisx:
Updated `--cache-artifacts` option usage language to be more accurate since now artifact caching is on by default.

### DIFF
--- a/docs/content/en/docs/references/cli/_index.md
+++ b/docs/content/en/docs/references/cli/_index.md
@@ -116,7 +116,7 @@ Examples:
 
 Options:
   -b, --build-image=[]: Choose which artifacts to build. Artifacts with image names that contain the expression will be built only. Default is to build sources for all artifacts
-      --cache-artifacts=false: Set to false to disable default caching of artifacts
+      --cache-artifacts=true: Set to true to enable caching of artifacts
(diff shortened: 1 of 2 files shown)


docs/content/en/docs/references/cli/_index.md
```diff
@@ -116,7 +116,7 @@ Examples:
 
 Options:
   -b, --build-image=[]: Choose which artifacts to build. Artifacts with image names that contain the expression will be built only. Default is to build sources for all artifacts
-      --cache-artifacts=true: Set to true to enable caching of artifacts
+      --cache-artifacts=false: Set to false to disable default caching of artifacts
```
balopat: 1. this file is generated using `hack/generate-man.sh`, so please revert the manual changes, run that script and commit the results! 

2. Also, the default value is put in `--cache-artifacts=true` which is `true` currently. I would still want to leave that there.

demisx: Got it. I kinda had a feeling that I was doing something silly. Anyway, I've followed your instructions and seems to me that I've ended up with the same diff at the end. Maybe I am missing something. I made a separate commit after each step. Please review.

PR #2722, opened 2019-08-26
Allow users to specify --namespace flag for tasks generated by generate-pipeline

marlon-gamez:
A few of these commits were already added by #2692 

This PR allows users to use the --namespace flag with generate-pipeline to then have the tasks generated by the command also run skaffold commands in the specified namespace.


pkg/skaffold/generate_pipeline/tasks.go
```diff
@@ -0,0 +1,166 @@
...
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/pipeline"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner/runcontext"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/version"
+
+	tekton "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func generateBuildTasks(runCtx *runcontext.RunContext, configFiles []*ConfigFile) ([]*tekton.Task, error) {
```
tejal29: If you are only using namesapce can you change this to 
```suggestion
func generateBuildTasks(ns string, configFiles []*ConfigFile) ([]*tekton.Task, error) {
```

tejal29: please add a test case and make sure namespace is present in the generated config.

PR #2734, opened 2019-08-27
fix clientset in-cluster

### LINKED ISSUE #2732: k8s clientset doesn't work when skaffold is ran from an pod
### Expected behavior

When skaffold is executed from a pod, it should be able to build within the same cluster using kaniko (if the right credentials are provided). 

### Actual behavior

Skaffold fails to build at first build with 

```
FATA[0002] exiting dev mode because first build failed: build failed: build failed: setting up pull secret: getting kubernetes client: getting client config for kubernetes client: error creating REST client config: invalid configuration: no configuration has been provided 
```

### Information

- Skaffold version: 0.36
- Operating system: Linux (from pod)
- Contents of skaffold.yaml: examples/kaniko

1.  `kubectl run -ti --image gcr.io/k8s-skaffold/skaffold  -n kube-system --command=true debug3 bash`
2. `git clone https://github.com/GoogleContainerTools/skaffold` 
3. `cd skaffold/examples/kaniko`
4. `skaffold build` 

```
root@debug3-68985f4cdf-hwtt7:/skaffold/examples/kaniko# skaffold build
Generating tags...
 - gcr.io/k8s-skaffold/skaffold-example -> gcr.io/k8s-skaffold/skaffold-example:v0.36.0-76-gfb2107f
Tags generated in 17.920901ms
Checking cache...
 - gcr.io/k8s-skaffold/skaffold-example: Not found. Building
Cache check complete in 1.905331497s
Starting build...
Creating kaniko secret [e2esecret]...
FATA[0001] build failed: build failed: setting up pull secret: getting kubernetes client: getting client config for kubernetes client: error creating REST client config: invalid configuration: no configuration has been provided 
```

I believe the reason is that we changed the config creation in this commit: https://github.com/GoogleContainerTools/skaffold/commit/7b2c6478f8efd34967084471dbacac619abd6f3a

This works outside the cluster but in-cluster creates (and caches) an invalid config. 

This also breaks our PR docs previewer as it reuses the same `pkg/skaffold/kubernetes.GetClientset()`.

balopat:
Fixes #2732.


integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+	t.Logf("create-build-step logs: \n%s", logs)
+	defer func() {
+		if output, err := skaffold.Delete("-p", "create-build-step").InNs(namespace).InDir("./testdata/skaffold-in-cluster").RunWithCombinedOutput(t); err != nil {
+			t.Logf("failed to cleanup skaffold-in-cluster: %s, output: %s", err, output)
+		}
+	}()
+
+	client, err := kubernetes.Client()
+	if err != nil {
+		t.Errorf("failed to get k8s client: %s", err)
```
dgageot: Maybe use `Fatalf` to avoid using `FailNow`

dgageot: (same in other places)

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+		t.FailNow()
+	}
+
+	if _, err := f.WriteString(kustomization); err != nil {
+		t.Errorf("failed writing kustomziation: %s", err)
+		t.FailNow()
+	}
+	f.Sync()
+	return func() {
+		f.Close()
```
dgageot: Can it be closed as soon as it is written?

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+		`nameSuffix: -%s
+resources:
+ - k8s-job.yaml`, suffix)
+	f, err := os.OpenFile("./testdata/skaffold-in-cluster/build-step/kustomization.yaml", os.O_WRONLY|os.O_CREATE, 0666)
+	if err != nil {
+		t.Errorf("failed opening kustomziation file for writing: %s", err)
+		t.FailNow()
+	}
+
+	if _, err := f.WriteString(kustomization); err != nil {
```
dgageot: Could you use `ioutil.WriteFile`?

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+}
+
+func setupSuffixKustomization(suffix string, t *testing.T) func() {
+	kustomization := fmt.Sprintf(
+		`nameSuffix: -%s
+resources:
+ - k8s-job.yaml`, suffix)
+	f, err := os.OpenFile("./testdata/skaffold-in-cluster/build-step/kustomization.yaml", os.O_WRONLY|os.O_CREATE, 0666)
+	if err != nil {
+		t.Errorf("failed opening kustomziation file for writing: %s", err)
```
corneliusweig:
```suggestion
		t.Fatalf("failed opening kustomziation file for writing: %s", err)
```
Would save the following `FailNow()`

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+		t.FailNow()
+	}
+
+	if _, err := f.WriteString(kustomization); err != nil {
+		t.Errorf("failed writing kustomziation: %s", err)
```
corneliusweig:
```suggestion
		t.Fatalf("failed writing kustomziation: %s", err)
```

integration/build_test.go
corneliusweig:
```suggestion
		t.Fatalf("failed to get k8s client: %s", err)
```

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+		t.Errorf("failed to get k8s client: %s", err)
+		t.FailNow()
+	}
+	podsClient := client.CoreV1().Pods(namespace)
+
+	if err := kubernetes.WaitForPodSucceeded(context.TODO(), podsClient, podName, 2*time.Minute); err != nil {
+		t.Errorf("in-cluster build pod failed: %s", err)
+		logs, err := podsClient.GetLogs(podName, &corev1.PodLogOptions{}).DoRaw()
+		if err != nil {
+			t.Errorf("error getting logs for pod: %s", err)
```
corneliusweig:
```suggestion
			t.Fatalf("error getting logs for pod: %s", err)
```

integration/testdata/skaffold-in-cluster/README.md
```diff
@@ -0,0 +1,21 @@
+# Skaffold in-cluster test data 
+
+Skaffold aims to be used as a buildstep in CI/CD pipelines. 
+What if this pipeline step is running inside a K8s cluster?
+ 
+In that case Skaffold needs to be able to work with an in-cluster k8s context to setup the secret and to create the pod for a kaniko build. 
+This test case is testing that flow.
+ 
+The `skaffold.yaml` describes _both_ the creation of an imaginary buildstep.
+The buildstep is implemented with a k8s Job under  `build-step` and an image,
```
dgageot: nit: two spaces

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+}
+
+func copySkaffoldBinary(t *testing.T) func() {
+	// copy the skaffold binary to the test case folder
+	// this is geared towards the in-docker setup: the fresh built binary is here
+	// for manual testing, we can override this temporarily
+	skaffoldSrc := "/usr/bin/skaffold"
+	skaffoldDst := "./testdata/skaffold-in-cluster/skaffold"
+	if written, err := fileutils.CopyFile(skaffoldSrc, skaffoldDst); written <= 0 || err != nil {
+		t.Errorf("failed to copy skaffold binary for test case: %s", err)
```
corneliusweig:
```suggestion
		t.Fatalf("failed to copy skaffold binary for test case: %s", err)
```

pkg/skaffold/kubernetes/context/context_test.go
```diff
@@ -49,14 +55,35 @@ users:
...
+
 func TestCurrentContext(t *testing.T) {
 	testutil.Run(t, "valid context", func(t *testutil.T) {
 		resetKubeConfig(t, validKubeConfig)
 
 		config, err := CurrentConfig()
 
 		t.CheckNoError(err)
-		t.CheckDeepEqual("cluster-foo", config.CurrentContext)
+
```
dgageot: nit: remove empty line

integration/build_test.go
dgageot: bit:typo

integration/build_test.go
dgageot: nit:typo

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+		}
+		t.Errorf("logs: %s", logs)
+		t.Fail()
+	}
+}
+
+func copySkaffoldBinary(t *testing.T) func() {
+	// copy the skaffold binary to the test case folder
+	// this is geared towards the in-docker setup: the fresh built binary is here
+	// for manual testing, we can override this temporarily
```
dgageot: Maybe use `exec.LookPath("skaffold")`?

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+	}
+	if !ShouldRunGCPOnlyTests() {
+		t.Skip("skipping test that is gcp only")
+	}
+
+	cleanupSkaffoldBinary := copySkaffoldBinary(t)
+	defer cleanupSkaffoldBinary()
+
+	suffix := uuid.New().String()
+	podName := fmt.Sprintf("skaffold-in-cluster-%s", suffix)
```
dgageot: maybe move this line next to the pod creation

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+	}
+}
+
+func copySkaffoldBinary(t *testing.T) func() {
+	// copy the skaffold binary to the test case folder
+	// this is geared towards the in-docker setup: the fresh built binary is here
+	// for manual testing, we can override this temporarily
+	skaffoldSrc := "/usr/bin/skaffold"
+	skaffoldDst := "./testdata/skaffold-in-cluster/skaffold"
+	if written, err := fileutils.CopyFile(skaffoldSrc, skaffoldDst); written <= 0 || err != nil {
```
dgageot: We should avoid using docker/docker code to copy a file

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+	if !ShouldRunGCPOnlyTests() {
+		t.Skip("skipping test that is gcp only")
+	}
+
+	cleanupSkaffoldBinary := copySkaffoldBinary(t)
+	defer cleanupSkaffoldBinary()
+
+	suffix := uuid.New().String()
+	podName := fmt.Sprintf("skaffold-in-cluster-%s", suffix)
+	cleanupKustomization := setupSuffixKustomization(suffix, t)
```
dgageot: I'm not sure I understand why we use a suffix here? Could this be solved by creating a namespace like in other tests?

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+
+	logs := skaffold.Run("-p", "create-build-step", "--cache-artifacts=true").InDir("./testdata/skaffold-in-cluster").InNs(namespace).RunOrFailOutput(t)
+	t.Logf("create-build-step logs: \n%s", logs)
+	defer func() {
+		if output, err := skaffold.Delete("-p", "create-build-step").InNs(namespace).InDir("./testdata/skaffold-in-cluster").RunWithCombinedOutput(t); err != nil {
+			t.Logf("failed to cleanup skaffold-in-cluster: %s, output: %s", err, output)
+		}
+	}()
+
+	client, err := kubernetes.Client()
```
dgageot: Other tests use `SetupNamespace` to create a ns and get a namespaced k8s client. Can this be done here?

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("failed to get k8s client: %s", err)
+		t.FailNow()
+	}
+	podsClient := client.CoreV1().Pods(namespace)
+
+	if err := kubernetes.WaitForPodSucceeded(context.TODO(), podsClient, podName, 2*time.Minute); err != nil {
```
dgageot: It would be nice to use something like `client.WaitForPodsReady()`

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+	podName := fmt.Sprintf("skaffold-in-cluster-%s", suffix)
+	cleanupKustomization := setupSuffixKustomization(suffix, t)
+	defer cleanupKustomization()
+
+	const namespace = "default"
+
+	logs := skaffold.Run("-p", "create-build-step", "--cache-artifacts=true").InDir("./testdata/skaffold-in-cluster").InNs(namespace).RunOrFailOutput(t)
+	t.Logf("create-build-step logs: \n%s", logs)
+	defer func() {
+		if output, err := skaffold.Delete("-p", "create-build-step").InNs(namespace).InDir("./testdata/skaffold-in-cluster").RunWithCombinedOutput(t); err != nil {
```
dgageot: If we can use a one-shot namespace, this can be removed.

pkg/skaffold/kubernetes/context/context.go
```diff
@@ -50,12 +50,7 @@ func UseKubeContext(overrideKubeContext string) {
 // If UseKubeContext was called before, the CurrentContext will be overridden.
 // The result will be cached after the first call.
 func GetRestClientConfig() (*restclient.Config, error) {
-	rawConfig, err := getRawKubeConfig()
-	if err != nil {
-		return nil, err
-	}
-	clientConfig := clientcmd.NewNonInteractiveClientConfig(rawConfig, kubeContext, &clientcmd.ConfigOverrides{CurrentContext: kubeContext}, nil)
```
corneliusweig: The reason for the detour to the raw kubeconfig was that the `DeferredLoadingClientConfig` caches the (REST)`clientConfig` with the current kubecontext (see vendor/k8s.io/client-go/tools/clientcmd/merged_client_builder.go:62). This means that the RESTClientConfig used by Skaffold will always work with the original kube-context, regardless of whether that was changed by Skaffold. Thus, the k8s REST calls would work with a different context than direct `kubectl` invocations (via `CurrentConfig` pkg/skaffold/kubernetes/context/context.go:59). This is even confusing to explain -- did I make it understandable?

So in essence, without the detour via the raw kube-config, there can be a misalignment of REST and `kubectl` calls which we should avoid.

---

Also see new test-case "change context after first execution"

pkg/skaffold/kubernetes/context/context.go
```diff
@@ -70,15 +65,19 @@ func getCurrentConfig() (clientcmdapi.Config, error) {
 	return cfg, err
 }
 
-// getRawKubeConfig retrieves and caches the raw kubeConfig. The cache ensures that Skaffold always works with the identical kubeconfig,
-// even if it was changed on disk.
-func getRawKubeConfig() (clientcmdapi.Config, error) {
+func getKubeConfig() clientcmd.ClientConfig {
```
corneliusweig: If this will be kept, this should have a clarifying comment similar to `getRawKubeConfig`.

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+	defer cleanupSkaffoldBinary()
+
+	suffix := uuid.New().String()
+	podName := fmt.Sprintf("skaffold-in-cluster-%s", suffix)
+	cleanupKustomization := setupSuffixKustomization(suffix, t)
+	defer cleanupKustomization()
+
+	const namespace = "default"
+
+	logs := skaffold.Run("-p", "create-build-step", "--cache-artifacts=true").InDir("./testdata/skaffold-in-cluster").InNs(namespace).RunOrFailOutput(t)
```
dgageot: `RunOrFail` will now print the output if there's an error

pkg/skaffold/kubernetes/context/context_test.go
```diff
@@ -105,6 +132,28 @@ func TestGetRestClientConfig(t *testing.T) {
 
 		t.CheckError(true, err)
 	})
+
+	testutil.Run(t, "context immutability", func(t *testutil.T) {
```
corneliusweig:
```suggestion
	testutil.Run(t, "kube-config immutability", func(t *testutil.T) {
```

pkg/skaffold/kubernetes/context/context_test.go
```diff
@@ -66,7 +93,7 @@ func TestCurrentContext(t *testing.T) {
 		config, err := CurrentConfig()
 
 		t.CheckNoError(err)
-		t.CheckDeepEqual("cluster-bar", config.CurrentContext)
+		t.CheckDeepEqual(clusterBarContext, config.CurrentContext)
 	})
 
 	testutil.Run(t, "invalid context", func(t *testutil.T) {
```
corneliusweig: This test fails now (and didn't beforehand). Should have included that case...
```go
testutil.Run(t, "change context after first execution", func(t *testutil.T) {
		resetKubeConfig(t, validKubeConfig)

		cfg, err := GetRestClientConfig()
		kubeContext = clusterBarContext

		t.CheckNoError(err)
		t.CheckDeepEqual("https://bar.com", cfg.Host)
	})
```

integration/build_test.go
```diff
@@ -108,6 +115,92 @@ func TestBuild(t *testing.T) {
...
+
+	cleanupSkaffoldBinary := copySkaffoldBinary(t)
+	defer cleanupSkaffoldBinary()
+
+	suffix := uuid.New().String()
+	podName := fmt.Sprintf("skaffold-in-cluster-%s", suffix)
+	cleanupKustomization := setupSuffixKustomization(suffix, t)
+	defer cleanupKustomization()
+
+	const namespace = "default"
```
corneliusweig: I guess for this exceptional case where additional static resources are needed, using the default namespace is fine.

balopat:
Thank you so much both for reviewing! 
Besides the nits, the two big issues: 

1. why `default` namespace and name suffixing? 
2. why changing back to `DeferredLoadingClientConfig`? 

I'll address those comments here: 

1. as there is no way to pass `skaffold build` which namespace the kaniko build should run in (maybe we should fix that with a new flag or reusing `--namespace`?) kaniko will always run in default - or whatever is `build.cluster.namespace` setup in the `skaffold.yaml`. I could templatize the skaffold.yaml...?

2. I will have a look at your testcase @corneliusweig - I have to understand a bit more the flows where skaffold changes kubeContext - I am surprised that that's something we allow at all. But even if we do this caching via RawConfig - we'll need to cater somehow for inCluster logic that we completely lost by avoiding the [`DeferredLoadingClientConfig.ClientConfig()` logic for in cluster path](https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/merged_client_builder.go#L120-L124)

integration/build_test.go
```diff
@@ -18,12 +18,21 @@ package integration
...
 	"context"
+	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/GoogleContainerTools/skaffold/testutil"
```
dgageot: nit: format those imports

testutil/util.go
```diff
@@ -61,6 +64,30 @@ func (t *T) Override(dest, tmp interface{}) {
...
 }
 
+func (t *T) CopyFile(src, dst string) {
+	from, err := os.Open(src)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer from.Close()
+
+	to, err := os.OpenFile(dst, os.O_RDWR|os.O_CREATE, 0666)
```
dgageot: Maybe use `ioutil.ReadFile` and `ioutil.WriteFile` would be produce shorter code.

testutil/util.go
```diff
@@ -61,6 +64,30 @@ func (t *T) Override(dest, tmp interface{}) {
...
+	if err != nil {
+		log.Fatal(err)
```
dgageot: t.Fatal?

pkg/skaffold/kubernetes/context/context_test.go
balopat: This testcase fails on `master` too. 
I think you wanted 

```
	testutil.Run(t, "change context after first execution", func(t *testutil.T) {
		resetKubeConfig(t, validKubeConfig)

		cfg, err := GetRestClientConfig()
		kubeContext = clusterBarContext
		cfg, err = GetRestClientConfig()

		t.CheckNoError(err)
		t.CheckDeepEqual("https://bar.com", cfg.Host)
	})
```

This requirement is future looking - currently we deal with a single kubeContext per skaffold execution. Your argument around kubectl vs client-go becoming inconsistent does not apply in the current setting: we call `UseKubeContext()` at the very beginning of the skaffold process, and kubectl CLI is instantiated from the `RunContext.KubeContext`. I would even close down the `UseKubeContext()` method with a `sync.Once` to allow only a single modification of that private field. 

Maybe in the future we could have a situation where we need to interact with multiple contexts in multiple deployments. Assuming that the kubeconfig is the same in this case and is cached similarly to what we have now - we just use different kubeContexts from it - in that case we should be able to create both CLIs (which is easy currently) and REST clients that points to the same context in the kubeconfig...but then we wouldn't drive it through _changing_ the global kubeContext variable, but rather by creating  a `GetRestClientConfig(kubeContext string)` method. And that would be broken, given my solution points to the `DeferredLoading...ClientConfig` which caches the actual kubeconfig after first call and does not allow for overrides - i.e. we can cache the kubeconfig at the beginning of the process but we won't be able to pick kubecontexts from it. 

Now your original solution works with picking contexts from the cached kubeconfig. However it lacks the logic for in-cluster situations. Now, thinking about it - we should think about he in-cluster config as _another context_. Thus here's what I'll implement: 

- use the override logic you're using with the current kubecontext (which might be empty)
- if kubecontext is empty and the resulting kubeconfig is empty - try the in-cluster rest-config

pkg/skaffold/kubernetes/context/context.go
```diff
@@ -48,15 +49,28 @@ func UseKubeContext(overrideKubeContext string) {
...
+	clientConfig := clientcmd.NewNonInteractiveClientConfig(rawConfig, kctx, &clientcmd.ConfigOverrides{CurrentContext: kctx}, nil)
 	restConfig, err := clientConfig.ClientConfig()
-	return restConfig, errors.Wrap(err, "error creating REST client config")
+	if kctx == "" && clientcmd.IsEmptyConfig(err) {
+		logrus.Debug("no kube-context set and no kubeConfig found, attempting in-cluster config")
+		restConfig, err := restclient.InClusterConfig()
+		return restConfig, errors.Wrap(err, "error creating REST client config in-cluster")
+	}
+
+	return restConfig, errors.Wrapf(err,"error creating REST client config for kubeContext '%s'", kctx)
```
corneliusweig:
```suggestion
	return restConfig, errors.Wrapf(err, "error creating REST client config for kubeContext '%s'", kctx)
```
nit: missing space

integration/testdata/skaffold-in-cluster/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM gcr.io/gcp-runtimes/ubuntu_16_0_4
+
+COPY ./skaffold /usr/bin/skaffold
```
corneliusweig: Maybe move the test-setup `COPY` commands after the block of commands for `kubectl` to improve layer caching?

pkg/skaffold/kubernetes/context/context_test.go
corneliusweig: > This testcase fails on master too.

:sweat_smile:

balopat:
> Oh, I was not aware, that `DeferredLoadingClientConfig` falls back to an in-cluster config and this breaks if doing the detour via kube-config (makes absolute sense though). So your solution looks like the correct thing to do in this situation.
> 
> As for the reason why this is necessary at all: when we will add the kubeContext configuration via `skaffold.yaml`, we need to look at the current kube-context during profile activation. However, profiles will also be able to override the kube-context, so I think we _must_ be able to change the kubecontext after the first call to `GetRestClientConfig` (and related utils).

Good point! 

> On the other hand, you say
> 
> > I would even close down the UseKubeContext() method with a sync.Once to allow only a single modification of that private field.
> 
> That also sounds like something we should implement. This way we would be protected, when somebody changes the `skaffold.yaml` kubeContext setting, such that Skaffold would suddenly not clean up in the old context. We only need to make sure that every execution path just calls `UseKubeContext()` once.

Agreed, we should think this through thoroughly in your PR introducing kubecontext to skaffold.yaml! 

> Btw, I really enjoyed this inception-test-case :)
hehe 👍

pkg/skaffold/kubernetes/context/context.go
```diff
@@ -48,15 +50,28 @@ func UseKubeContext(overrideKubeContext string) {
 
 // GetRestClientConfig returns a REST client config for API calls against the Kubernetes API.
 // If UseKubeContext was called before, the CurrentContext will be overridden.
-// The result will be cached after the first call.
+// The kubeconfig used will be cached for the life of the skaffold process after the first call.
+// If the CurrentContext is empty and the resulting config is empty, this method attempts to
+// create a RESTClient with an in-cluster config.
```
tejal29: +1 nice

Issue #2687, opened 2019-08-20
add option to override labels / not to apply labels


giladsh1:
### Expected behavior
The latest Skaffold release (v0.36.0) has set artifact caching on by default (awesome feature by the way!).
This means I don't need to repeat previous builds when changing / fixing k8s configs.
However, since Skaffold applies custom labels for each run, including the `skaffold.dev/run-id` label, this means that deployments will forced to be configured each run, and they won't be idempotent. 
For example, please look at the following mongoDB deployment (statefullset) - 
```
Name:               mongo-0
Namespace:          riscale-test
Priority:           0
PriorityClassName:  <none>
Node:               gke-gke-dev-europe-we-persistent-pool-43a650d2-3nlv/10.132.0.27
Start Time:         Tue, 20 Aug 2019 16:34:41 +0300
Labels:             app=mongo
                    app.kubernetes.io/managed-by=skaffold-v0.36.0
                    controller-revision-hash=mongo-d4d79d7bc
                    env=test
                    skaffold.dev/builder=google-cloud-build
                    skaffold.dev/cleanup=true
                    skaffold.dev/deployer=kubectl
                    skaffold.dev/profiles=riscale-test
                    **_skaffold.dev/run-id=bfae5d05-c01c-40a8-8516-663dafd4a419_**
                    skaffold.dev/tag-policy=git-commit
                    skaffold.dev/tail=true
                    statefulset.kubernetes.io/pod-name=mongo-0
```
And the following deployment is the exact same, other than the skaffold.dev/run-id tag - 
```
Name:               mongo-0
Namespace:          riscale-test
Priority:           0
PriorityClassName:  <none>
Node:               gke-gke-dev-europe-we-persistent-pool-43a650d2-3nlv/10.132.0.27
Start Time:         Tue, 20 Aug 2019 16:34:41 +0300
Labels:             app=mongo
                    app.kubernetes.io/managed-by=skaffold-v0.36.0
                    controller-revision-hash=mongo-d4d79d7bc
                    env=test
                    skaffold.dev/builder=google-cloud-build
                    skaffold.dev/cleanup=true
                    skaffold.dev/deployer=kubectl
                    skaffold.dev/profiles=riscale-test
                    **_skaffold.dev/run-id=dc524cde-21f0-42dd-aa52-8fcdcf0890c7_**
                    skaffold.dev/tag-policy=git-commit
                    skaffold.dev/tail=true
[...]

PR #2752, opened 2019-08-29
Allow users to override the run-id

### LINKED ISSUE #2745: Skaffold run is not idempotent anymore
Since runId was introduced now k8s resources are recreated. 
We should think about carefully how to enable idempotent behavior again.

dgageot:
Skaffold can be made idempotent again by using `skaffold run -l skaffold.dev/run-id=SOMETHING`
Or by setting the `SKAFFOLD_LABELS=skaffold.dev/run-id=dgageot` env variable globally.

Fixes #2745 and #2687

### DIFF
--- a/integration/run_test.go
+++ b/integration/run_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package integration
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/GoogleContainerTools/skaffold/integration/skaffold"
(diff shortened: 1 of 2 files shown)


paultiplady:
Since this issue has a lot of references pointing to it, I want to note here that I believe overriding the `run-id` only makes skaffold idempotent _most_ of the time. If you upgrade skaffold in between commands, or the docker api, then other labels are going to change. (e.g. see #3171)

PR #2759, opened 2019-08-29
Indent deploy phase output

dgageot:


dgageot:
@nkubala oh man, exactly, what about helm...
I don't want to touch it until I understand how it works. Which is exactly what I'm trying to do right now. In fact, it's such a pain that I came here to check on my PRs to forget helm's hell for a sec.

Issue #2463, opened 2019-07-13
`skaffold init ; skaffold build/dev/run/...` fails if Kubernetes manifest references a tagged image


jpetazzo:
I run `skaffold init` with the following files:

`---8<---deployment.yaml---`
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: skatest
  name: skatest
spec:
  replicas: 1
  selector:
    matchLabels:
      app: skatest
  template:
    metadata:
      labels:
        app: skatest
    spec:
      containers:
      - image: foo/bar:baz
        name: skatest
```

`---8<---Dockerfile---`
```dockerfile
FROM nginx
```

This generates the following skaffold file:

```yaml
apiVersion: skaffold/v1beta12
kind: Config
build:
  artifacts:
  - image: foo/bar:baz
deploy:
  kubectl:
    manifests:
    - deployment.yaml
```

Then I run `skaffold dev`.

### Expected behavior

Skaffold builds my image and stuff happens.

### Actual behavior

I get the following output:
```
Generating tags...
 - foo/bar:baz -> WARN[0000] Unable to find git commit: Running [git describe --tags --always]: stdout , stderr: fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
, err: exit status 128: exit status 128 
foo/bar:baz:dirty
Tags generated in 1.774232ms
Starting build...
Building [foo/bar:baz]...
Sending build context to Docker daemon  2.048kB
Pruning images...
Image prune complete in 18.783µs
FATA[0000] exiting dev mode because first build failed: build failed: building [foo/bar:baz]: build artifact: docker build: Error response from daemon: invalid reference format 
```

It is trying to build the image `foo/bar:baz:dirty`, which fails.

If I remove the image tag `:baz` from the `skaffold.yaml` file, the build succeeds.

### Extra notes

I suspect that this is a classic case of "user doesn't read the documentation and expects things to work regardless", but it looks like Skaffold could warn me or scream at me if my Kubernetes resources reference tagged images. I like my onboarding to be smooth!

(This is particularly confusing because the output says `Building [foo/bar:baz]...`, which we know to be LIES ALL LIES!!!)

dgageot:
This PR #2773 should fix the issue

PR #2773, opened 2019-08-30
Improve `skaffold init` behaviour when tags are used in manifests

### LINKED ISSUE #834: Ignore tag in skaffold.yaml and kubernetes yaml
I think this worked once upon a time, but it doesn't anymore.

We should allow users to specify a tag in the skaffold.yaml or their kubernetes.yaml and

1. If we built the image, ignore it and replace it with our intermediate tag
2. If we didn't build the image, don't do anything.

dgageot:
+ Skaffold init now ignores images that can't be parsed
+ It also ignores images references by digest in the manifests
+ Skaffold run/dev ignore tags (not digests) in manifests

Fixes #834 and #2463

### DIFF
--- a/integration/init_test.go
+++ b/integration/init_test.go
@@ -40,6 +40,10 @@ func TestInit(t *testing.T) {
 			dir:  "testdata/init/hello",
 		},
 		{
+			name: "ignore existing tags",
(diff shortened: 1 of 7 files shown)


dgageot:
@nkubala This should be good to go

PR #2782, opened 2019-09-02
Warn helm users when image tags are not used

dgageot:


pkg/skaffold/deploy/helm.go
```diff
@@ -326,9 +330,20 @@ func (h *HelmDeployer) deployRelease(ctx context.Context, out io.Writer, r lates
...
+		valuesSet[v] = true
 		args = append(args, "--set", fmt.Sprintf("%s=%s", k, v))
 	}
 
+	// Let's make sure that every image tag is set with `--set`.
+	// Otherwise, templates have no way to use the images that were built.
+	for _, build := range builds {
+		if !valuesSet[build.Tag] {
+			warnings.Printf("image [%s] is not used.", build.Tag)
+			warnings.Printf("image [%s] is used instead.", build.ImageName)
```
balopat: I'm wondering if new lines or spaces should be used somewhere here?

dgageot: There's actually a new line between those too warnings.

PR #2786, opened 2019-09-02
Limit concurrent builds on Cluster and GCB builders

dgageot:
Users can now configure the maximum number of concurrent GCB/In cluster builds. I thought about using a flag and since I hate flags, I chose yaml.


dgageot:
@balopat I'll fix that!

Issue #2562, opened 2019-07-29
Fail-fast on environment variable not set


dkirrane:
### Expected behavior

Fail fast if an environment variable is not set.

### Actual behavior

**\<no value\>** because `VERSION` environment variable is not set.

```
$ skaffold run
?[34mGenerating tags...?[0m
?[34m - eu.gcr.io/oceana-team/ip-office-poc -> ?[0meu.gcr.io/my-project/my-app:<no value>
```

_sometime later ... after jib build:_

**Invalid image reference: eu.gcr.io/my-project/my-app:<no value>**
```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:1.4.0:build (default-cli) on project my-app: Invalid image reference eu.gcr.io/my-project/my-app:<no value>, perhaps you should check that the reference is formatted correctly according to https://docs.docker.com/engine/reference/commandline/tag/#extended-description
[ERROR] For example, slash-separated name components cannot have uppercase letters: Invalid image reference: eu.gcr.io/my-project/my-app:<no value>
```

### Information

- Skaffold version: v0.34.1
- Operating system: Windows 10 64-bit PC
- Contents of skaffold.yaml: 

```yaml
apiVersion: skaffold/v1beta13
kind: Config
build:
  artifacts:
  - image: eu.gcr.io/my-project/my-app
    context: ./app
    jibMaven:
      args:
      - -Djib.httpTimeout=300000
      - -Dmaven.test.skip=true
  tagPolicy:
    envTemplate:
      template: '{{.IMAGE_NAME}}:{{.VERSION}}'
```

### Steps to reproduce the behavior

1. `skaffold run`

balopat:
I think this is great to fail fast in general, and this would be relatively simple to implement. 
However if we have users who want to have "empty var" in certain environments, that can be problematic on Windows for example where creating an empty variable means deleting the variable as far as I know.

balopat:
We discussed this - let's get a design proposal in for this - there are a bunch of options how we can implement this. Our current thinking is that this should be optional behavior - some questions to explore:
- should this apply to all templated fields (https://skaffold.dev/docs/how-tos/templating/) or just this template.  
- what should be the name of the field: `failFast:true` or `allowEmpty:false` or `verifyTemplate: true` or similar 
[...]

PR #2793, opened 2019-09-03
[EnvTemplate] Fail fast

dgageot:
+ Fail fast when an env variable is used but not set
 + Fail when deprecated placeholders are used


dgageot:
@balopat currently, if an env variable is not defined, it'll replace it with `<no value>` so I think it's ok to break that. wdyt?

balopat:
> @balopat currently, if an env variable is not defined, it'll replace it with `<no value>` so I think it's ok to break that. wdyt?

That is pretty ugly and yes, good guess that people don't rely on that anyway. Awesome. Let's go with this. 

Fixes #2562.

PR #2794, opened 2019-09-03
Fix spurious logs when dev loop is restarted

### LINKED ISSUE #2791: [skaffold dev] changing skaffold.yaml reprints all the logs
While in a `skaffold dev` loop, I run `touch skaffold.yaml`. This resets the dev loop with no change and it should carry on with the logs where it left them.

Instead:
 + It reprints all the logs for every container involved in the deployment.
 + It print `WARN[0019] signal: killed` for every running pod (I assume it's the kubectl logs being killed)
 + It often prints `<Container was Terminated>` for each container. Even when they are not terminated.

dgageot:
Fix #2791

### DIFF
--- a/pkg/skaffold/kubernetes/log.go
+++ b/pkg/skaffold/kubernetes/log.go
@@ -152,7 +152,11 @@ func (a *LogAggregator) streamContainerLogs(ctx context.Context, pod *v1.Pod, co
 	tr, tw := io.Pipe()
 	go func() {
 		if err := a.kubectlcli.Run(ctx, nil, tw, "logs", sinceSeconds, "-f", pod.Name, "-c", container.Name, "--namespace", pod.Namespace); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/skaffold/kubernetes/log.go
```diff
@@ -193,7 +197,10 @@ func (a *LogAggregator) streamRequest(ctx context.Context, headerColor color.Col
 			// Read up to newline
 			line, err := r.ReadString('\n')
 			if err == io.EOF {
-				a.printLogLine(headerColor, prefix, "<Container was Terminated>\n")
+				// Unless the context was interrupted, this means that the container was stopped.
+				if ctx.Err() != context.Canceled {
```
balopat: another time I see this happen is when I run `skaffold dev` after `skaffold run` - the "tail end" of the container logs gets picked up.

Issue #2775, opened 2019-08-30
Better config versioning management


balopat:
Specs: https://docs.google.com/document/d/e/2PACX-1vRBN1vHjSGbs3ZmdvyzzhBn48NjeuP4OiTTG5mec_YrRzNDL63pFa1Xy1rgCUNiQXTjAE1QyvCr5yxX/pub

Based on the above, I have some recommendations around what to do with regards to version management, I believe these would simplify our lives: 

TODO: 
- [x] fail new_version script if current version is unreleased
- [x] we should mark a version released/unreleased, by a comment on top the config.go file maintained - explaining how to do config changes 
- [x] on every PR we should check for _structural changes_ in schemas. Fail if there is a structural change in a released version.
-  [x] in our release.sh script: if the latest version is unreleased, mark the latest version as released (in comment) 
- [ ] convert the above doc to a markdown / page on our skaffold.dev / DEVELOPMENT.md

PR #2798, opened 2019-09-04
Improve versioning

balopat:
This is regarding #2775.

- hack/new_version now checks via the Github API whether the current version is released or not - if not then fails 
- hack/new_version.sh now takes the new version as an argument for easier testing, e.g, try it out with `hack/new_version.sh v1beta15`!

next PRs will cater for :
- add a PR check that uses this release logic to ensure that no released version is mutated

### DIFF
--- a/hack/new_config_version/version.go
+++ b/hack/new_config_version/version.go
@@ -18,11 +18,20 @@ package main
 
 import (
 	"bufio"
+	"context"
+	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
+
+	"github.com/google/go-github/github"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema"
 )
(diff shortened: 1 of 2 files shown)


briandealwis:
Could the `release.sh` script compare `.../latest/config.go` — or the generated schema? — since the last tagged release?

The last tagged release is available from `git describe --match 'v[0-9]*' --abbrev 0` ([source](https://stackoverflow.com/questions/3867619/how-to-get-last-git-tag-matching-regex-criteria#comment26823639_3867811)).

balopat:
> I'm not sure I like the usage of an alias for `latest`. We've been there and it's harder to use. For example, you don't get auto import from the IDE when you write `latest.SkaffoldConfig{}`

The autoimport is a good point. I think the `latest` pointer is not a vital thing, I can introduce the `Released` flag without it. I'll split out the timeout increase!

balopat:
> Could the `release.sh` script compare `.../latest/config.go` — or the generated schema? — since the last tagged release?
> 
> The last tagged release is available from `git describe --match 'v[0-9]*' --abbrev 0` ([source](https://stackoverflow.com/questions/3867619/how-to-get-last-git-tag-matching-regex-criteria#comment26823639_3867811)).

That could work too, thanks for the suggestion! 
Pros: it can't be spoofed + no need for an extra bool flag. 
Cons: now the contributor won't have a visual confirmation before trying to add a new field - but we can  still add a comment before the config version though in the release script.

hack/new_version.sh
```diff
@@ -14,11 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-echo "Current skaffold config version: " $(sed -n 's;.*Version.*"skaffold/\(.*\)";\1;p' pkg/skaffold/schema/latest/config.go)
```
dgageot: Isn't this message useful when a version is not provided as an argument?

hack/new_version.sh
```diff
@@ -14,11 +14,14 @@
...
+set -e
 
-echo
+NEW_VERSION=$1
+
+if [[ "" == "$NEW_VERSION" ]]; then
```
dgageot: `if [[ -z "$NEW_VERSION" ]]`

hack/new_config_version/version.go
```diff
@@ -18,21 +18,62 @@ package main
...
 )
 
 // Before: prev -> current (latest)
 // After:  prev -> current -> new (latest)
 func main() {
+	logrus.SetLevel(logrus.DebugLevel)
 	new := os.Args[1]
-	current := strings.TrimPrefix(schema.SchemaVersions[len(schema.SchemaVersions)-1].APIVersion, "skaffold/")
+	current := latest.Version
+	fmt.Printf("current version: %s", current)
```
dgageot: `logrus.Infof`?

hack/new_config_version/version.go
```diff
@@ -18,21 +18,62 @@ package main
...
+	"context"
+	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
 
+	"github.com/google/go-github/github"
```
dgageot: goimports

hack/new_version.sh
balopat: I'll add it in `version.go`

PR #2808, opened 2019-09-04
Unify jibMaven and jibGradle blocks into jib with auto detection

### LINKED ISSUE #1134: Jib builder should autodetect Maven vs Gradle
From #1096:

@dgageot:
> Is there a way to add a single type of artifact called Jib that auto-detects if it's a gradle or a maven build? That would make the config easier to read/write. I know the maven and gradle configs don't have the same parameters though.

@coollog:
> Yep, we can definitely auto-detect Gradle/Maven (via detection of pom.xml or build.gradle or the respective wrappers). However, the concern is which priority to detect these in. For example, someone could have both pom.xml and build.gradle and those files could be named differently as well. The wrappers could also both be present. In any case, we would also need to provide a user configuration for Maven or Gradle as a fallback in case the auto-detection was incorrect. This detection could probably be specified via a buildFile parameter where passing in an .xml file would mean Maven and .gradle file would be Gradle.

@dgageot:
> Maybe have a jib artifact with a type property. Type would be auto-detected unless the user specifies maven or gradle. Once again, it can be changed later but I think it makes the configuration simpler.

@briandealwis:
> For the Maven case we need to look for a pom.xml or a .mvn/ directory — Maven now supports a polyglot mode (e.g., using groovy or Scala or other formats for expressing the project model), so pom.xml may not exist.

@loosebazooka:
> Maven and Gradle will probably have different configuration parameters. We might need to make it nested to handle a single jib artifact.
> ```
> jib {
>   maven {
>     profile = "asdf"
>     ...
>   }
> }
> 
> // or
> 
> jib {
>   gradle {
>     ...
>   }
> }
> ```

briandealwis:
Combines the `JibMavenArtifact` and `JibGradleArtifact` (and corresponding `jibMaven:` and `jibGradle:` blocks) into a single `JibArtifact` and `jib:` block.  Jib now guesses the underlying Jib plugin type based on the file contents within the project.

`JibArtifact` actually has a `Type` field — a simple int — for explicitly specifying the plugin type, but it is not current exported as it's only been used for tests.  We could make this field explicit later, but we'd need to teach the schema generator how to translate enums.

Fixes #1134


docs/content/en/schemas/v1beta14.json
```diff
@@ -1229,67 +1194,39 @@
...
-          "type": "array",
-          "description": "additional build flags passed to Maven.",
-          "x-intellij-html-description": "additional build flags passed to Maven.",
-          "default": "[]",
-          "examples": [
-            "[\"-x\", \"-DskipTests\"]"
-          ]
-        },
-        "module": {
+        "profile": {
```
loosebazooka: I feel like we should get rid of this and just translate it to `-Pprofile` in the `args` section?

briandealwis: 👍

pkg/skaffold/jib/jib.go
```diff
@@ -88,6 +88,45 @@ type filesLists struct {
...
+	for _, gradleFile := range []string{"build.gradle", "gradle.properties", "settings.gradle", "gradlew", "gradlew.bat", "gradlew.cmd"} {
+		if util.IsFile(filepath.Join(workspace, gradleFile)) {
+			return JibGradle, nil
+		}
+	}
+	// check for typical maven files; .mvn is a directory used for polyglot maven
+	if util.IsFile(filepath.Join(workspace, "pom.xml")) || util.IsDir(filepath.Join(workspace, ".mvn")) {
+		return JibMaven, nil
+	}
+	return -1, errors.Errorf("Unable to determine Jib plugin type for %s", workspace)
```
chanseokoh: Didn't you say you would use 0 for unknown?

pkg/skaffold/build/gcb/jib.go
```diff
@@ -21,26 +21,35 @@ import (
...
-	return cloudbuild.Build{
-		Steps: []*cloudbuild.BuildStep{{
-			Name:       b.MavenImage,
-			Entrypoint: "sh",
-			Args:       fixHome("mvn", jib.GenerateMavenArgs("build", tag, artifact, b.skipTests, b.insecureRegistries)),
-		}},
+func (b *Builder) jibBuildSpec(artifact *latest.Artifact, tag string) (cloudbuild.Build, error) {
+	t, err := jib.DeterminePluginType(artifact.Workspace, artifact.JibArtifact)
+	if err != nil {
+		return cloudbuild.Build{}, err
```
chanseokoh: Q: why an empty `cloudbuild.Build{}` instead of nil?

docs/content/en/docs/how-tos/builders/_index.md
```diff
@@ -208,7 +204,7 @@ no longer required with Jib 1.4.0 and should be removed.
 #### Gradle
 
 To build a multi-module project with Gradle, specify each sub-project as a separate
```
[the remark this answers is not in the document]
briandealwis: Good point — I rewrote the Gradle section to match the style of the Maven section.

integration/examples/jib-multimodule/README.adoc
```diff
@@ -18,14 +18,14 @@ build:
      - image: gcr.io/k8s-skaffold/skaffold-jib-1
        # context is the root of the multi-module project
        context: .
-       jibMaven:
-          # module is either the relative location within the project (i.e.,
+       jib:
+          # project is either the relative location within the project (i.e.,
           # relative to `context`) or :artifactId or groupId:artifactId
```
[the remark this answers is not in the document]
briandealwis: Good catch.

pkg/skaffold/jib/jib_maven.go
```diff
@@ -32,26 +32,26 @@ const MinimumJibMavenVersion = "1.4.0"
 // MavenCommand stores Maven executable and wrapper name
 var MavenCommand = util.CommandWrapper{Executable: "mvn", Wrapper: "mvnw"}
 
-// GetDependenciesMaven finds the source dependencies for the given jib-maven artifact.
+// getDependenciesMaven finds the source dependencies for the given jib-maven artifact.
 // All paths are absolute.
-func GetDependenciesMaven(ctx context.Context, workspace string, a *latest.JibMavenArtifact) ([]string, error) {
-	deps, err := getDependencies(workspace, getCommandMaven(ctx, workspace, a), a.Module)
+func getDependenciesMaven(ctx context.Context, workspace string, a *latest.JibArtifact) ([]string, error) {
```
loosebazooka: Doesn't this still need to get exported? It's used in `local.go` and `types.go`. I also don't know how anything works in go.

pkg/skaffold/build/gcb/jib.go
briandealwis: It's a value and not a reference, so can't return nil.

pkg/skaffold/schema/v1beta13/upgrade_test.go
```diff
@@ -35,6 +35,13 @@ build:
   - image: gcr.io/k8s-skaffold/bazel
     bazel:
       target: //mytarget
+  - image: gcr.io/k8s-skaffold/jib-maven
+    jibMaven:
+      args: ['-v']
```
loosebazooka: should add in a test with `module` -> `project` in jibMaven?

pkg/skaffold/jib/jib.go
briandealwis: Because PluginType is an integer, its default value is 0 (e.g., when creating a new instance or when unmarshalling a YAML structure), and is effectively means "guess".  Whereas at this point, we have guessed and failed.  So I think something non-zero is better.

pkg/skaffold/jib/jib_maven.go
briandealwis: They were changed to use `GetDependencies()`.

pkg/skaffold/jib/jib.go
```diff
@@ -88,6 +88,45 @@ type filesLists struct {
...
+func DeterminePluginType(workspace string, artifact *latest.JibArtifact) (PluginType, error) {
+	// check if explicitly specified
+	if artifact != nil {
+		if t := PluginType(artifact.Type); t.IsKnown() {
+			return t, nil
+		}
+	}
+
+	// check for typical gradle files
+	for _, gradleFile := range []string{"build.gradle", "gradle.properties", "settings.gradle", "gradlew", "gradlew.bat", "gradlew.cmd"} {
```
loosebazooka: wonder if we should be searching for kotlin config files: `build.gradle.kts`?

pkg/skaffold/schema/v1beta13/upgrade_test.go
briandealwis: 👍

pkg/skaffold/jib/jib.go
briandealwis: That's a good point.  Opened #2820 to address it separately as `skaffold init` support needs to be updated too.

briandealwis:
> How does https://skaffold.dev/docs/references/yaml/ get updated?

It uses [JS to slurp the schema from the site](https://github.com/GoogleContainerTools/skaffold/blob/master/docs/content/en/docs/references/yaml/main.js#L9)

Makefile
```diff
@@ -105,8 +105,7 @@ test: $(BUILD_DIR)
 .PHONY: coverage
 coverage: $(BUILD_DIR)
 	@ ./hack/test.sh
-	@ curl -s https://codecov.io/bash > $(BUILD_DIR)/upload_coverage
-	@ bash $(BUILD_DIR)/upload_coverage
+	@- curl -s https://codecov.io/bash > $(BUILD_DIR)/upload_coverage && bash $(BUILD_DIR)/upload_coverage
```
balopat: can we split this out to a separate PR?

deprecation-policy.md
```diff
@@ -54,8 +54,7 @@ However the **behavior** of individual component might suffer breaking changes d
 - Artifact types:
   - Dockerfile: beta
   - Bazel: beta
-  - jibMaven: alpha
-  - jibGradle: alpha
+  - Jib: alpha
```
balopat: I think we can mark jib beta safely at this point

Makefile
briandealwis: #2841

docs/content/en/docs/how-tos/builders/_index.md
```diff
@@ -190,13 +186,13 @@ each produce a separate container image.
 
 #### Maven
```
balopat: I would love to see the resolution logic described here.

PR #2817, opened 2019-09-05
remove unused value in helm example

medyagh:
image value is not used, as the image is built from Dockerfile.
also changed the service type to NodePort so we can access it by `minikube service svc-name` command.


medyagh:
@dgageot closing this in favor of https://github.com/GoogleContainerTools/skaffold/pull/2819

PR #2822, opened 2019-09-05
add example for skaffold generate-pipeline

marlon-gamez:
Add a simple example to give users an idea of how to use the command


marlon-gamez:
@tejal29 done!

Issue #2670, opened 2019-08-16
Suggestion: Continuous Fuzzing


yevgenypats:
Hi, I'm Yevgeny Pats Founder of [Fuzzit](https://fuzzit.dev) - Continuous fuzzing as a service platform.

We have a free plan for OSS and I would be happy to contribute a PR if that's interesting.
The PR will include the following
- [go-fuzz](https://github.com/dvyukov/go-fuzz) fuzzers (This is generic step not-connected to fuzzit)
- Continuous Fuzzing of master branch which will generate new corpus and look for new crashes
- Regression on every PR that will run the fuzzers through all the generated corpus and fixed crashes from previous step. This will prevent new or old bugs from crippling into master.

You can see our basic example [fuzzitdev/example-go](https://github.com/fuzzitdev/example-go) and you can see an example of "in the wild" integration [google/syzkaller](https://github.com/google/syzkaller).

Let me know if this is something worth working on.

Cheers,
Yevgeny

balopat:
Hi Yevgeny, thank you for reaching out! This sounds interesting, would be cool to catch some bugs with it. First ideas where we could use: one obvious part would be the config parser. We also have a Control API that takes requests. There might be other places too, but this would be a good start to experiment with.

yevgenypats:
hi @balopat the config parser sounds like a good candidate indeed. I think I'll start with that.

Also, I need to mention that we have a [reward program](https://fuzzit.dev/2019/08/12/announcing-rewards-for-go-rust-oss-projects/) in case you have time/want to implement this yourself I would be happy to reward you as well as get unbiased feedback on how smooth the integration was. Let me know what you prefer.

balopat:
Thank you, good to know - for me finding a bug would be a sufficient reward - it's my job to make skaffold better. I don't have time to jump in right now but it looks interesting.

However, someone from the community might be motivated more by the reward program. I'm open to contributions either from you or others. Feel free to post this issue on our slack channel.

yevgenypats:
Sounds good, can you send me an invite to slack to yp [at] fuzzit.dev

balopat:
I see you managed to find it and post! Let me know if you need any help!

yevgenypats:
@balopat help is appreciated, looks like there were more projects [interested](https://github.com/fuzzitdev/fuzzit) then I assumed and I'm working now on unrelated fuzzit release. if you have some spare time then this is this will be quicker. 

[...]

PR #2835, opened 2019-09-08
Automated fuzzing on Fuzzit

bookmoons:
Integrates with [Fuzzit](https://fuzzit.dev/) for automated fuzzing. Proposed in #2670.

Under this setup Travis runs fuzzing on every build. It runs a fuzz regression test locally of past crashes and submits to Fuzzit for a full fuzzing run. The build ends as soon as the fuzzing run is in the queue. Fuzzit emails when it finds crashes.

This tries to fuzz the suggested targets plus some other identified parsers. Targets are:
* `control-api-http` - HTTP requests to the control API server.
* `control-api-tcp` - Binary requests to the control API server.
* `parse-config` - Fuzz the config parser.
* `parse-reference` - Fuzz the Docker image reference parser.

There's a JDWP specification parser in `pkg/skaffold/debug` that seems suitable for fuzzing. It's tough to get at because it's a private function. If you think it's worth fuzzing exposing it would make it possible.

A `skaffold` org is preconfigured on Fuzzit. If whoever will be administering [signs in](https://fuzzit.dev/) to create an account, I can get you added to the org.

Thank you for considering.


bookmoons:
@googlebot signed it

bookmoons:
@dominikh Hi, I built on your good work for this PR. The googlebot is asking for consent to have your commits merged. Would you mind looking at it?

bookmoons:
I see there's a strange error in the build. A type error even though the types are identical.

```
/home/travis/gopath/src/skaffold/pkg/skaffold/docker/auth.go:69:39: cannot use cf (variable of type *configfile.ConfigFile) as *configfile.ConfigFile value in argument to gcp.AutoConfigureGCRCredentialHelper
```

Not sure what that's about, it works locally. Will be looking into this.

dominikh:
@googlebot I consent.

bookmoons:
Thanks @dominikh.

bookmoons:
Applied the Go modules approach from https://github.com/dvyukov/go-fuzz/issues/195#issuecomment-523526736 to get this working. Thanks @mvdan for the pattern.

This is ready for review.

balopat:
This looks good but I'm hesitant to add 8 more minutes to our feedback loop. Can you please change it to a job instead of a stage so at least it can run in parallel with the other jobs?

bookmoons:
Will do, thanks for looking at it.

bookmoons:
Rebased this and converted the fuzz stages to jobs. I'll check back to see how the build goes.

If you'd like to [sign in](https://fuzzit.dev/) to Fuzzit to make an account I can add you to the org. Then final setup is like this:
* In Fuzzit, switch to the skaffold org with the dropdown at the top.
* In settings grab an API key.
* In repo settings in [Travis](https://travis-ci.org/) paste it to envvar `FUZZIT_API_KEY`.

dgageot:
@balopat do you think we should merge this (when it's rebased)?

balopat:
I'm not convinced. One thing is the slowness of the test, the other is that most of the test code seems to be HTTP specific instead of Skaffold's test code. It feels like this kind of code should live in a fuzz-http library or similar...I'm not sure I want to maintain this.

dgageot:
@balopat I agree. I'm sorry @bookmoons but I think maintaining that would cost us more that it brings.

balopat:
Sounds good. Apologies for dragging this out this long. Closing.

Issue #2542, opened 2019-07-24
Allow kustomization folders as source for skaffold config


botwhytho:
This is a follow up to the work done for https://github.com/GoogleContainerTools/skaffold/issues/2200

We are using `kustomize` to merge multiple `skaffold.yaml` files with great results. We have centralized remote bases with profiles activated by `kubeContext` which centralizes and simplifies some CI logic. We then apply patches at the repo level to specify things that are more repo-specific: build details, image name, number of artifacts, etc.

Right now in our CI, we have steps like the following:
```bash
kustomize build deploy/skaffold/ > skaffold.yaml
skaffold run
```

Ideally we'd like to only have to run:
```bash
skaffold run -f deploy/skaffold/
```
In this case skaffold would test if the string provided to the `-f|--filename` flag is either a compatible skaffold config or a folder that contains a kustomization, if the latter it would use the output of the kustomization as the skaffold config.

Not necessary but would also be useful if the same functionality works even if no `-f|--filename` flag is provided, so skaffold would first test a `skaffold.yaml` file in the current directory, then if none exists would test if the current directory contains a kustomization and proceed if the output is a valid skaffold config.

Considering a remote skaffold config can be provided by specifying a URL to the `-f` flag, skaffold should be flexible enough to accept kustomizations as a source for it's config.

wstrange:
@botwhytho Would be interested in seeing a sample of how you combine multiple skaffold.yaml files with kustomize.

balopat:
This is very interesting, but I'm not convinced yet that we can tackle everything with this. Also profiles seem to collide a bit with this direction. Can you share a little bit more about the structure of your project and examples of what your overlays/kustomizations are?

botwhytho:
Hi, the idea behind the feature request is centralizing best practices and 'shared config' while keeping project/repo specific details where it belongs (in a specific repo). I think this works in tandem with profiles (and now yaml anchors, as I saw the merged PR), and not against it.

Currently we have a central repo that has some kustomize and scaffold config. We might have a base skaffold file similar to this one:
```yaml
apiVersion: skaffold/v1beta10
kind: Config
metadata:
  name: base-config
build:
  tagPolicy:
    envTemplate:
      template: "{{.IMAGE_NAME}}:{{.ARTIFACT_PREFIX}}_{{.VERSION}}"

profiles:
- name: local-dev
  deploy:
    kustomize:
      path: kustomize/overlays/local

- name: dev
  activation:
[...]

PR #2836, opened 2019-09-08
Support YAML anchors in skaffold.yaml (key must start with a dot)

AndiDog:
Example which doesn't work with current Skaffold version:

```
apiVersion: skaffold/v1beta12
kind: Config

.build_common: &build_common
  local:
    useDockerCLI: yes

  tagPolicy:
    gitCommit:
      variant: AbbrevCommitSha

.build_common_activation: &build_common_activation
  # Empty `ONLY_MODULE` builds all modules (activate all profiles)
  env: ONLY_MODULE=

profiles:
  - name: client
    activation:
      - *build_common_activation
      - env: ONLY_MODULE=client
    build:
      <<: *build_common
      artifacts:
        - image: client
          docker:
            buildArgs:
              ONLY_MODULE: client

  - name: server
    activation:
      - *build_common_activation
      - env: ONLY_MODULE=server
    build:
      <<: *build_common
      artifacts:
        - image: server
          docker:
            buildArgs:
              ONLY_MODULE: server
```

The environment variable approach is a real-world use case I have (a script uses that variable to create the manifest for only one module of the monorepo, or all modules).

This change discards all top-level keys of the YAML dictionary before the strict parsing step. The yaml.v2 package already takes care to resolve aliases correctly, so only the keys need to be removed, since they cannot be mapped to config struct fields.

The concept is the same as [supported by .gitlab-ci.yml](https://docs.gitlab.com/ee/ci/yaml/#anchors), for instance: using `.` as prefix.

### DIFF
--- a/pkg/skaffold/schema/profiles_test.go
+++ b/pkg/skaffold/schema/profiles_test.go
@@ -552,6 +552,55 @@ func TestActivatedProfiles(t *testing.T) {
 	}
 }
 
+func TestYamlAlias(t *testing.T) {
+	config := `
+.activation_common: &activation_common
+  env: ABC=common
+build:
+  artifacts:
+  - image: example
+profiles:
+- name: simple1
+  activation:
+  - *activation_common
+  - env: ABC=1
+  build:
+    artifacts:
+    - image: simpleimage1
+- name: simple2
+  activation:
+  - *activation_common
+  - env: ABC=2
+  build:
+    artifacts:
+    - image: simpleimage2
+`
+
+	testutil.Run(t, "", func(t *testutil.T) {
+		tmpDir := t.NewTempDir().
+			Write("skaffold.yaml", addVersion(config))
+
+		parsed, err := ParseConfig(tmpDir.Path("skaffold.yaml"), false)
+		t.RequireNoError(err)
+
+		skaffoldConfig := parsed.(*latest.SkaffoldConfig)
+
+		t.CheckDeepEqual(2, len(skaffoldConfig.Profiles))
+		t.CheckDeepEqual("simple1", skaffoldConfig.Profiles[0].Name)
+		t.CheckDeepEqual([]latest.Activation{{Env: "ABC=common"}, {Env: "ABC=1"}}, skaffoldConfig.Profiles[0].Activation)
+		t.CheckDeepEqual("simple2", skaffoldConfig.Profiles[1].Name)
+		t.CheckDeepEqual([]latest.Activation{{Env: "ABC=common"}, {Env: "ABC=2"}}, skaffoldConfig.Profiles[1].Activation)
+
+		err = ApplyProfiles(skaffoldConfig, cfg.SkaffoldOptions{
+			Profiles: []string{"simple1"},
+		})
+		t.CheckNoError(err)
+
+		t.CheckDeepEqual(1, len(skaffoldConfig.Build.Artifacts))
+		t.CheckDeepEqual(latest.Artifact{ImageName: "simpleimage1"}, *skaffoldConfig.Build.Artifacts[0])
+	})
+}
+
 func str(value string) *interface{} {
 	var v interface{} = value
 	return &v
--- a/docs/content/en/docs/references/yaml/_index.html
+++ b/docs/content/en/docs/references/yaml/_index.html
@@ -8,3 +8,9 @@ weight: 120
 <link rel="stylesheet" type="text/css" href="main.css">
 <script type="module" src="main.js"></script>
 <table id="table" data-version="{{% skaffold-version %}}"></table>
+
+<h3>YAML anchors</h3>
+<p>
+    Anchors can be defined by having top-level keys starting with a dot, e.g. <code>.common_stuff: &alias_name</code>.
+    You can then reuse the value using <code>*alias_name</code>.
+</p>
--- a/pkg/skaffold/schema/versions.go
+++ b/pkg/skaffold/schema/versions.go
@@ -18,6 +18,7 @@ package schema
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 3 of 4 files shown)


AndiDog:
Rebased onto master

pkg/skaffold/schema/profiles_test.go
```diff
@@ -464,6 +464,49 @@ func TestActivatedProfiles(t *testing.T) {
...
+  - env: ABC=2
+`
+
+	testutil.Run(t, "", func(t *testutil.T) {
+		tmpDir := t.NewTempDir().
+			Write("skaffold.yaml", addVersion(config))
+
+		parsed, err := ParseConfig(tmpDir.Path("skaffold.yaml"), false)
+		t.CheckNoError(err)
+		if parsed == nil || err != nil {
```
dgageot: t.CheckNoError(err) should be enough here to test that there's no error

pkg/skaffold/schema/profiles_test.go
```diff
@@ -464,6 +464,49 @@ func TestActivatedProfiles(t *testing.T) {
...
+
+		t.CheckDeepEqual([]latest.Profile{
+			{Name: "simple1", Activation: []latest.Activation{{Env: "ABC=common"}, {Env: "ABC=1"}}},
+			{Name: "simple2", Activation: []latest.Activation{{Env: "ABC=common"}, {Env: "ABC=2"}}},
+		}, skaffoldConfig.Profiles)
+
+		err = ApplyProfiles(skaffoldConfig, cfg.SkaffoldOptions{
+			Profiles: []string{"simple1"},
+		})
+		t.CheckNoError(err)
```
dgageot: Should you add assertions on the resulting config?

pkg/skaffold/schema/profiles_test.go
AndiDog: It would surely fail the test as expected, but the issue is that the test code keeps running after `CheckNoError`. Comment out the `if parsed == nil || err != nil {` _here_ and the "ignore dotted keys" feature, and you will get failure output from the line `skaffoldConfig := parsed.(*latest.SkaffoldConfig)` which should not even run:

```
Running tool: /usr/local/Cellar/go/1.12.9/libexec/bin/go test -timeout 30s github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema -run ^(TestYamlAlias)$

panic: interface conversion: util.VersionedConfig is nil, not *latest.SkaffoldConfig [recovered]
	panic: interface conversion: util.VersionedConfig is nil, not *latest.SkaffoldConfig

goroutine 23 [running]:
testing.tRunner.func1(0xc0003f6300)
	/usr/local/Cellar/go/1.12.9/libexec/src/testing/testing.go:830 +0x392
panic(0x18f4f00, 0xc0001ce060)
	/usr/local/Cellar/go/1.12.9/libexec/src/runtime/panic.go:522 +0x1b5
github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema.TestYamlAlias.func1(0xc00015c120)
	/Users/asommer/dev/extern/skaffold/pkg/skaffold/schema/profiles_test.go:496 +0x5b4
github.com/GoogleContainerTools/skaffold/testutil.Run.func1(0xc0003f6300)
	/Users/asommer/dev/extern/skaffold/testutil/util.go:122 +0x89
testing.tRunner(0xc0003f6300, 0xc0001e00f0)
	/usr/local/Cellar/go/1.12.9/libexec/src/testing/testing.go:865 +0xc0
created by testing.(*T).Run
	/usr/local/Cellar/go/1.12.9/libexec/src/testing/testing.go:916 +0x35a
FAIL	github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema	0.034s
Error: Tests failed.
```

Since that's confusing for test developers, I suggest introducing a new helper `RequireNoError` which panics immediately (other test frameworks also distinguish "CHECK_*" and "REQUIRE_*" in this way).

pkg/skaffold/schema/profiles_test.go
AndiDog: Done

pkg/skaffold/schema/profiles_test.go
balopat: I agree that sometimes we want that behavior - `testing.T` has `FailNow`, which we could call - I don't think panic is a good idea - that stops the whole testsuite

AndiDog: Added such a function (using `t.FailNow()`), and rebased onto master

balopat:
One worry I have with this approach is - if we (hypothetically) ever wanted to embrace kustomize as a first class approach to manipulate and reuse pieces of skaffold.yaml - would that approach be incompatible?

AndiDog:
> One worry I have with this approach is - if we (hypothetically) ever wanted to embrace kustomize as a first class approach to manipulate and reuse pieces of skaffold.yaml - would that approach be incompatible?

kustomize patches parsed manifests, with YAML aliases already resolved (more exactly it can only patch JSON, I believe). Ignoring the anchors should not make a difference. I'm questioning the usefulness, though – if the use case is to support multiple environments, that should be done in the manifest generation step (e.g. `ENVIRONMENT=prod ./generate-manifest.sh && skaffold build --file-output skaffold-build.json`). Going off-topic... maybe you can point me to the related issue to add to the discussion with experiences I made.

AndiDog:
@balopat Should we finish this PR? I tried to address your worry in the above comment, and resolved all review requests.

balopat:
Thank you for your patience @AndiDog - this was the feature request I had in mind: https://github.com/GoogleContainerTools/skaffold/issues/2542. It is not the deploy-time kustomize usage that I thought about - but kustomize for the skaffold.yaml! WDYT?

AndiDog:
> Thank you for your patience @AndiDog - this was the feature request I had in mind: #2542. It is not the deploy-time kustomize usage that I thought about - but kustomize for the skaffold.yaml! WDYT?

The use cases would be interesting. But whether it gets implemented one or the other way, or not – the YAML anchors patch should still work because it's the parser which resolves the aliases already, so all Skaffold has to do is to discard dotted keys (which, if they are anchors, already got copied to where they were referenced) before passing the YAML on to `yaml.UnmarshalStrict(buf, cfg)` for parsing into struct fields. Steps like kustomize would come before that in order to produce a raw YAML to parse, or as the feature request says, to verify whether it's a valid Skaffold config at all.

AndiDog:
Can this be reviewed?

dgageot:
@balopat could you take a look too?

PR #2850, opened 2019-09-10
Ignore Jib project name when skaffold init only finds one module

### LINKED ISSUE #2844: "skaffold init" generates incorrect config for Jib project
In #2843, the config for a Jib project has `module: demo`

```yaml
    jibMaven:
      module: demo
```

However, the project was a simple Spring Boot demo generated from the [Initializr](https://start.spring.io/) and not a multi-module project. I had to remove `module:` to make it work correctly. Seems like a bug.

@GoogleContainerTools/java-tools-build

TadCordle:
Fixes #2844 

@GoogleContainerTools/java-tools-build

### DIFF
--- a/pkg/skaffold/jib/jib_init.go
+++ b/pkg/skaffold/jib/jib_init.go
@@ -156,7 +156,11 @@ func ValidateJibConfig(path string) []Jib {
 			return nil
 		}
 
-		results[i] = Jib{BuilderName: builderType.Name(), Image: parsedJSON.Image, FilePath: path, Project: parsedJSON.Project}
+		results[i] = Jib{BuilderName: builderType.Name(), Image: parsedJSON.Image, FilePath: path}
+		if len(matches) > 1 {
+			// Only apply project if there are multiple modules
+			results[i].Project = parsedJSON.Project
+		}
 	}
 	return results
 }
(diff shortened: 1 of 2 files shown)


pkg/skaffold/jib/jib_init.go
```diff
@@ -156,7 +156,11 @@ func ValidateJibConfig(path string) []Jib {
 			return nil
 		}
 
-		results[i] = Jib{BuilderName: builderType.Name(), Image: parsedJSON.Image, FilePath: path, Project: parsedJSON.Project}
+		results[i] = Jib{BuilderName: builderType.Name(), Image: parsedJSON.Image, FilePath: path}
```
chanseokoh: What happens if a multi-module project has only one sub-module? That is, a root `pom.xml` (`<packaging>pom`) and a single sub-directory for one module. Does `jib:skaffold-init` return one `BEGIN JIB JSON`?

TadCordle: I can double check, but I think it returns one for the root project and one for the module.

TadCordle: I'm beginning to think this is best fixed Jib-side. I'm going to mess around with Jib and see if I can get somewhere close, and if so then I'll close this.

Issue #176, opened 2018-03-09
Wait for deployed resources to reach their intended state 


viglesiasce:
In the case of `skaffold run` there are times where you will want to ensure that the resources that were deployed have reached their intended state. For example, after applying a change to a deployment you will want to check the rollout status and ensure that it completed successfully. This would most likely be the mode you would want when running in a CD pipeline.

Place1:
I'd be happy to put up a PR that adds `rollout status` support for the kubectl deployer

balopat:
cc @tejal29

tejal29:
@Place1 Thank you for helping us out. I am working on a PR #2750 which does implement kubectl rollout

Place1:
Oh fantastic. Looks like your pr is doing what’s needed. Nice one.

tstromberg:
Closing as I believe this to be fixed.

PR #2851, opened 2019-09-10
Add Resource.Status object and remove sync.Map

tejal29:
Relates to #176 

Add a `resource.Status` struct which stores the status of the deployment in `Deployment.status` field.

Remove `sync.Map` which was used to deployments status.

Output changes:
* No output changes.

Next PRs:
- [ ] Add `updated` and `reported` booleans to `resource.Status`.
- [ ] Update Status only if its changed instead of updating every single time.
- [ ] Finally merge in #2813

### DIFF
--- a/pkg/skaffold/deploy/resource/deployment.go
+++ b/pkg/skaffold/deploy/resource/deployment.go
@@ -30,6 +30,7 @@ type Deployment struct {
 	namespace string
 	rType     string
 	deadline  time.Duration
+	status    *Status
 }
 
 func (d *Deployment) String() string {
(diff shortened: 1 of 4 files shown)


pkg/skaffold/deploy/resource/deployment.go
```diff
@@ -44,11 +45,27 @@ func (d *Deployment) Deadline() time.Duration {
 	return d.deadline
 }
 
+func (d *Deployment) Status() *Status {
```
balopat: If I look at this PR in isolation - I wonder, if we have `Status` returning the pointer to `resource.status` - then we don't really need it. Also, UpdateStatus is superflous, we can just Status public? 

I guess you will introduce an interface? Maybe we should add that now?

tejal29: The reason why `status` is private is because `UpdateStatus` should the only way to update a new status for a resource.  I think it makes more sense to make it `Status` not stored as a pointer.

PR #2854, opened 2019-09-10
Always add image flag to jib builders in skaffold init

### LINKED ISSUE #2852: Skaffold init uses incorrect image with single jib config + single k8s manifest
I have a single-module maven project setup with a k8s yaml that points to `gcr.io/k8s-skaffold/project1`:
```
apiVersion: apps/v1
kind: Deployment
...
      containers:
      - name: web1
        image: gcr.io/k8s-skaffold/project1
        ports:
          - containerPort: 8080
...
```

I have Jib setup to point to a different image, `gcr.io/tcordle-test/abc`:
```
...
<configuration>
  <to><image>gcr.io/tcordle-test/abc</image></to>
</configuration>
...
```

When I run `skaffold init --XXenableJibInit`, the prompt is skipped because there's only one image and one builder. However, the output has Jib building to its own target image, instead of what's defined in the k8s manifest.

```
$ skaffold init --XXenableJibInit
apiVersion: skaffold/v1beta14
kind: Config
build:
  artifacts:
  - image: gcr.io/tcordle-test/abc
    jibMaven: {}
deploy:
  kubectl:
    manifests:
    - k8s/web.yaml
```

The expected behavior is that the `jibMaven` block will contain an arg (`-Dimage=gcr.io/k8s-skaffold/project1`) that will make sure it tags the image correctly.

TadCordle:
Fixes #2852 

`skaffold init` used to use Jib's configured `to.image` to auto-pair Jib projects with k8s images. However, the skaffold.yaml that was generated prioritized the `to.image` configuration over the k8s images, possibly resulting in a skaffold.yaml that points a Jib builder to an image that isn't defined in a k8s manifest. This PR fixes that by only using k8s images in the generated artifacts, and also always explicitly passing a `-Dimage=` flag to Jib to make sure it builds the correct image, despite its configuration.

@GoogleContainerTools/java-tools-build

### DIFF
--- a/pkg/skaffold/jib/jib_init.go
+++ b/pkg/skaffold/jib/jib_init.go
@@ -60,10 +60,7 @@ func (j Jib) Describe() string {
 func (j Jib) CreateArtifact(manifestImage string) *latest.Artifact {
 	workspace := filepath.Dir(j.FilePath)
 
-	a := &latest.Artifact{ImageName: j.Image}
-	if j.Image == "" {
-		a.ImageName = manifestImage
-	}
+	a := &latest.Artifact{ImageName: manifestImage}
 
 	if workspace != "." {
 		a.Workspace = workspace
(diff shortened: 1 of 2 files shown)


TadCordle:
@chanseokoh Yes, for the most part. If a match can be determined, it will still be automatic (i.e. Jib's `to.image` matches an image in a k8s yaml), but if there are multiple builders that can't be determined automatically then it will resort to a prompt. The only automatic pairing that happens with a mismatched image is if exactly one builder and one k8s image are found.

PR #2060, opened 2019-05-02
Prepare to Add functionality to Replacer interface to restrict setting labels on certain kinds.

tejal29:
Fixes #1737 and other deploy issues  where labels get added to wrong CRDs.

In this PR,  i added a Matcher Interface which is used by replacer to replace objects on a manifest key if it is present.

```
// Matcher is used by Replacer to match object blob to replace based on  a manifest key in the Manifest
type Matcher interface {
	IsMatchKey(key string) bool
	Matches(v interface{}) bool
}
```

If the manifest object blob contains a match-key and its value is not matched, then we move to the the next Manifest object blob.
In this PR,  i have introduced a new interface.  This does not change in functionality.
Currently, all replacers `namespaceCollector`, `labelsSetter`, `imageSaver` and `imageReplacer` use the `anyMatcher` which match everything.

The next steps for this feature  would be
- [ ] Add a `KindMatcher` which will match key `kind` with values `Pod`, `Deployment` and `Services` (deployed Resources)
- [ ] Use the `KindMatcher` for `labelSetter`

PR #2859, opened 2019-09-11
Add kind Matcher which accepts a blacklist of values for `Kind` key

tejal29:
Relates to #1737 

Should go after #2862
Addressed subsequent work from #2060 

In this PR, 

**Output Changed**:
Add a debug line when "Kind" key has a non string value. 
For an invalid  manifest with "kind" value of not string type
```
apiVersion: v1
kind: {}
metadata:

```
Output line will look this in debug mode.
```
DEBU[0000] {} is type struct {} but type string expected for key `kind`. skipping value match.
```

**Next Changes:**
 - [ ]  Use the KindMatcher for labelSetter 
Something like this:
```
type labelsSetter struct {
        matcher matcher.Kind
	labels map[string]string
}

func newLabelsSetter(labels map[string]string) *labelsSetter {
	return &labelsSetter{
		labels: labels,
+                matcher: matcher.New([]string{'CustomResourceDefinition'})
	}
}

func (r *labelsSetter) ObjMatcher() Matcher {
	return r.matcher
}
```

### DIFF
--- a/pkg/skaffold/deploy/kubectl/matcher/kind.go
+++ b/pkg/skaffold/deploy/kubectl/matcher/kind.go
@@ -1,0 +1,64 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package matcher
+
+import (
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	kindKey = "kind"
+)
+
+// Kind matches an object with "Kind" which is not listed in the forbiddenValues.
+type Kind struct {
+	forbiddenValues map[string]struct{}
+}
+
+func (k *Kind) Matches(v interface{}) bool {
+	if str, ok := k.getValue(v); ok {
+		return k.notInBlacklist(str)
+	}
+	logrus.Debugf("%v is type %T but type string expected for key `%s`. skipping value match.", v, v, kindKey)
+	return true
+}
+
+func (k *Kind) getValue(v interface{}) (string, bool) {
+	value, ok := v.(string)
+	return value, ok
+}
+
+func (k *Kind) notInBlacklist(s string) bool {
+	_, ok := k.forbiddenValues[s]
+	return !ok
+}
+
+func (k *Kind) IsMatchKey(key string) bool {
+	return strings.ToLower(key) == kindKey
+}
+
+func New(values []string) *Kind {
+	m := map[string]struct{}{}
+	for _, v := range values {
+		m[v] = struct{}{}
+	}
+	return &Kind{
+		forbiddenValues: m,
+	}
+}
--- a/pkg/skaffold/deploy/kubectl/matcher/kind.go
+++ b/pkg/skaffold/deploy/kubectl/matcher/kind.go
@@ -26,7 +26,7 @@ const (
 	kindKey = "kind"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/skaffold/deploy/kubectl/matcher/kind.go
```diff
@@ -0,0 +1,64 @@
...
+func (k *Kind) notInBlacklist(s string) bool {
+	_, ok := k.forbiddenValues[s]
+	return !ok
+}
+
+func (k *Kind) IsMatchKey(key string) bool {
+	return strings.ToLower(key) == kindKey
+}
+
+func New(values []string) *Kind {
```
balopat: this could be just `string...`, a bit less code is required

pkg/skaffold/deploy/kubectl/matcher/kind.go
```diff
@@ -0,0 +1,64 @@
...
+	value, ok := v.(string)
+	return value, ok
+}
+
+func (k *Kind) notInBlacklist(s string) bool {
+	_, ok := k.forbiddenValues[s]
+	return !ok
+}
+
+func (k *Kind) IsMatchKey(key string) bool {
```
balopat:
```suggestion
func (k *Kind) MatchesKey(key string) bool {
```

pkg/skaffold/deploy/kubectl/matcher/kind.go
```diff
@@ -0,0 +1,64 @@
...
+const (
+	kindKey = "kind"
+)
+
+// Kind matches the value of "Kind" key with values not listed in the forbiddenValues.
+type Kind struct {
+	forbiddenValues map[string]struct{}
+}
+
+func (k *Kind) Matches(v interface{}) bool {
```
balopat:
```suggestion
func (k *Kind) MatchesValue(v interface{}) bool {
```

pkg/skaffold/deploy/kubectl/matcher/kind.go
```diff
@@ -0,0 +1,64 @@
...
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	kindKey = "kind"
+)
+
+// Kind matches the value of "Kind" key with values not listed in the forbiddenValues.
```
balopat:
```suggestion
// FieldMatcher tests wether a field matches the key with values not listed in the forbiddenValues.
```

pkg/skaffold/deploy/kubectl/matcher/kind.go
```diff
@@ -0,0 +1,64 @@
...
+type Kind struct {
```
balopat: I would rename this to `FieldMatcher` and pass in the key instead of making it only for `kind`

pkg/skaffold/deploy/kubectl/matcher/kind.go
tejal29: Accepted.

pkg/skaffold/deploy/kubectl/matcher/kind.go
tejal29: This is a interface defines in https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/deploy/kubectl/matcher.go#L22
```
// Matcher is used by Replacer to match object blob to replace based
// on a manifest key in the Manifest
// Note: If the manifest key is not present, the replacer will replace.
type Matcher interface {
	MatchesKey(key string) bool
	Matches(v interface{}) bool
}
```

pkg/skaffold/deploy/kubectl/matcher/kind.go
tejal29: agreed. we can do it. The reason i didn't make it generic was
- For it be truly generic, we will have to define the list of ignored value as `interface{}` to support others.
  where as for kind, we know the value of the kind is `string`
```
type GenericIgnoreField struct {
key string
values []interface
}
```
We have an option to only support keys with values of type `string`. 
I would like to keep it simple and support only key with `string` values for now. 

.

pkg/skaffold/deploy/kubectl/matcher/kind.go
tejal29: If we have to do this, we need to write a `Ignorer` interface.
WDYT?

PR #2886, opened 2019-09-15
Add graceful termination for custom builders

### LINKED ISSUE #2656: Graceful termination for custom builder
### Expected behavior
When cancelling skaffold build with a custom builder, it would be helpful if the build command process had time to clean up. For example, first sending SIGTERM, and then waiting before sending SIGKILL.

### Actual behavior
Cancelling CommandContext sends SIGKILL immediately, preventing the custom builder from cleaning up. An example of something that might need to be cleaned up is a pod from a cluster builder.

### Information

- Skaffold version: v0.34.1
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta13
kind: Config
build:
  artifacts:
  - image: ...
    custom:
      buildCommand: ./build.sh
```

1. `skaffold build` with any `build.sh`
2. Ctrl-C before `build.sh` completes

cedrickring:
Fixes #2656

I added a 2 seconds grace period for the process to complete after receiving SIGINT.

### DIFF
--- a/pkg/skaffold/build/custom/custom.go
+++ b/pkg/skaffold/build/custom/custom.go
@@ -70,6 +70,8 @@ func (b *ArtifactBuilder) Build(ctx context.Context, out io.Writer, a *latest.Ar
 
 func (b *ArtifactBuilder) handleGracefulTermination(ctx context.Context, cmd *exec.Cmd) error {
 	done := make(chan struct{})
+	defer close(done)
+
 	go func() {
 		select {
 		case <-ctx.Done():
@@ -92,14 +94,14 @@ func (b *ArtifactBuilder) handleGracefulTermination(ctx context.Context, cmd *ex
 				// forcefully kill process after 2 seconds grace period
 				cmd.Process.Kill()
 			case <-done:
+				return
 			}
 		case <-done:
+			return
 		}
 	}()
 
-	err := cmd.Wait()
-	close(done)
-	return err
+	return cmd.Wait()
 }
 
 func (b *ArtifactBuilder) retrieveCmd(out io.Writer, a *latest.Artifact, tag string) (*exec.Cmd, error) {
--- a/pkg/skaffold/build/custom/custom.go
+++ b/pkg/skaffold/build/custom/custom.go
@@ -81,11 +81,11 @@ func (b *ArtifactBuilder) Build(ctx context.Context, out io.Writer, a *latest.Ar
 				cmd.Process.Kill()
 			}
 
-			// wait 30 seconds or wait for the process to complete
+			// wait 2 seconds or wait for the process to complete
 			select {
-			case <-time.After(30 * time.Second):
+			case <-time.After(2 * time.Second):
 				logrus.Debugf("Killing process %v\n", cmd.Process.Pid)
-				// forcefully kill process after 30 seconds grace period
+				// forcefully kill process after 2 seconds grace period
 				cmd.Process.Kill()
 			case <-done:
 			}
--- a/pkg/skaffold/build/custom/custom.go
+++ b/pkg/skaffold/build/custom/custom.go
@@ -37,6 +37,7 @@ import (
 var (
 	// For testing
 	buildContext = retrieveBuildContext
+	buildCmdFunc = buildCmd
 )
 
 // ArtifactBuilder is a builder for custom artifacts
@@ -65,6 +66,10 @@ func (b *ArtifactBuilder) Build(ctx context.Context, out io.Writer, a *latest.Ar
 		return errors.Wrap(err, "starting cmd")
 	}
 
+	return b.handleGracefulTermination(ctx, cmd)
+}
+
+func (b *ArtifactBuilder) handleGracefulTermination(ctx context.Context, cmd *exec.Cmd) error {
 	done := make(chan struct{})
 	go func() {
 		select {
@@ -93,24 +98,15 @@ func (b *ArtifactBuilder) Build(ctx context.Context, out io.Writer, a *latest.Ar
 		}
 	}()
 
-	err = cmd.Wait()
+	err := cmd.Wait()
 	close(done)
 	return err
 }
 
 func (b *ArtifactBuilder) retrieveCmd(out io.Writer, a *latest.Artifact, tag string) (*exec.Cmd, error) {
 	artifact := a.CustomArtifact
-	split := strings.Split(artifact.BuildCommand, " ")
 
-	cmd := exec.Command(split[0], split[1:]...)
-
-	// special case for inline (ba)sh scripts
-	if (split[0] == "bash" || split[0] == "sh") && split[1] == "-c" {
-		args := strings.Join(split[2:], " ")
-
-		cmd = exec.Command(split[0], split[1], args[1:len(args)-1])
-	}
-
+	cmd := buildCmdFunc(artifact.BuildCommand)
 	cmd.Stdout = out
 	cmd.Stderr = out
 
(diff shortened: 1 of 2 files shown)


tejal29:
@cedrickring, i feel the better way to do this is by defining a different cleanUpCommand.

```
apiVersion: skaffold/v1beta13
kind: Config
build:
  artifacts:
  - image: ...
    custom:
      buildCommand: ./build.sh
      cleanUpCommand: ./clean-up.sh
```
And then, skaffold to execute the `cleanUpCommand` and not wait for a constant time to cleanup.
What do you think?

cedrickring:
Hey @tejal29, thanks for the feedback.

The (up to) 30 seconds wait is not exactly for cleanup imo, but for the command to exit gracefully. Right now the process gets killed immediately after pressing Ctrl+C and doesn't have time to exit normally (temp files aren't cleaned up etc). Usually a process shouldn't take the whole 30 seconds to terminate.
 
I think a cleanup script could be an additional option for this, but not a replacement.

pkg/skaffold/build/custom/custom_test.go
```diff
@@ -125,8 +130,98 @@ func TestRetrieveCmd(t *testing.T) {
...
+		{
+			description: "terminate gracefully and kill process",
+			artifact: &latest.Artifact{
+				ArtifactType: latest.ArtifactType{
+					CustomArtifact: &latest.CustomArtifact{
+						BuildCommand: fmt.Sprintf("%s --test.run=TestHelperGracefulBuildKill", os.Args[0]),
+					},
+				},
+			},
+			testEnv:   []string{"GO_STARTED_FROM_TEST=1"},
```
tejal29: this should unset in a `defer`.

pkg/skaffold/build/custom/custom_test.go
```diff
@@ -125,8 +130,98 @@ func TestRetrieveCmd(t *testing.T) {
...
+			description: "terminate gracefully and exit 0",
+			artifact: &latest.Artifact{
+				ArtifactType: latest.ArtifactType{
+					CustomArtifact: &latest.CustomArtifact{
+						BuildCommand: fmt.Sprintf("%s --test.run=TestHelperGracefulBuildCancel", os.Args[0]),
+					},
+				},
+			},
+			testEnv:   []string{"GO_STARTED_FROM_TEST=1"},
+			shouldErr: false,
```
tejal29: not needed as default is `false`

tejal29:
@cedrickring, thanks for clarifying. I see, you are implementing what https://github.com/golang/go/issues/22757 proposes. 

I think, if we implement this solution, we should reduce the timeout for 2 seconds like 
https://github.com/mvdan/sh/pull/170/files#diff-b071a7aa58a1e42f9c5278440e585c77R49

pkg/skaffold/build/custom/custom_test.go
```diff
@@ -125,8 +130,98 @@ func TestRetrieveCmd(t *testing.T) {
...
+
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, os.Interrupt)
+
+	go func() {
+		<-signals
+		fmt.Println("Got interrupt")
+	}()
+
+	fmt.Println("Sleeping 25 seconds")
```
tejal29: i am little bit concerned with these tests adding sleeps. Can you just add a bash command like `sleep 5` in the build script?

Also, run it multiple times (100) to make sure its reliable.

cedrickring: Imo 100 repeats are a bit much for a unit test. Also inlining a bash script required some extra changes in code.

cedrickring: Also the tests require the interrupt signal to be trapped. Otherwise they will always error.

tejal29: sorry, i meant running it 100 times via `go run -n -test.run <>`. I will do it myself before reviewing.

pkg/skaffold/build/custom/custom_test.go
```diff
@@ -125,8 +139,57 @@ func TestRetrieveCmd(t *testing.T) {
...
+			},
+			shouldErr: true,
+		},
+	}
+
+	for _, test := range tests {
+		testutil.Run(t, test.description, func(t *testutil.T) {
+			builder := NewArtifactBuilder(false, test.testEnv)
+
+			for i := 0; i < 10; i++ {
```
tejal29: please remove 10 iterations. Sorry I meant running it locally to make sure its not flaky.

pkg/skaffold/build/custom/custom.go
```diff
@@ -52,19 +56,61 @@ func NewArtifactBuilder(pushImages bool, additionalEnv []string) *ArtifactBuilde
...
+	return err
 }
 
-func (b *ArtifactBuilder) retrieveCmd(ctx context.Context, out io.Writer, a *latest.Artifact, tag string) (*exec.Cmd, error) {
+func (b *ArtifactBuilder) retrieveCmd(out io.Writer, a *latest.Artifact, tag string) (*exec.Cmd, error) {
 	artifact := a.CustomArtifact
 	split := strings.Split(artifact.BuildCommand, " ")
 
-	cmd := exec.CommandContext(ctx, split[0], split[1:]...)
+	cmd := exec.Command(split[0], split[1:]...)
```
tejal29: We don't want to support running arbitrary `bash scripts`.
Can we instead refactor `retriveCmd` and add a func `func buildCmd(string) *exec.Cmd`?

``` 
// for testing
var (
buildCmdFunc = buildCmd
)

func buildCmd(s string) *exec.Cmd {
    split := strings.Split(s " ")
     return exec.CommandContext(split[0], split[1:]...)
}

```

And in tests

```
func  mockBuildCmd(s string) (*exec.Cmd, error) {
split := strings.Split(s, " ")
 args := strings.Join(split[2:], " ")
 return  exec.Command(split[0], split[1], args[1:len(args)-1])
}




func TestGracefulBuildCancel(t *testing.T) {
...
 for _, test := range(tests) {
   testutil.Override(&retrieveCmdFunc, &mockBuildCmd)
  ....
}
```

pkg/skaffold/build/custom/custom.go
```diff
@@ -52,19 +56,61 @@ func NewArtifactBuilder(pushImages bool, additionalEnv []string) *ArtifactBuilde
...
 	}
 
-	return cmd.Run()
+	if err := cmd.Start(); err != nil {
+		return errors.Wrap(err, "starting cmd")
+	}
+
+	done := make(chan struct{})
+	go func() {
+		select {
```
tejal29: This code block is really long. Can we refactor this in to `handleGracefulTermination`

pkg/skaffold/build/custom/custom.go
cedrickring: This is obsolete after refactoring

pkg/skaffold/build/custom/custom.go
```diff
@@ -52,19 +56,57 @@ func NewArtifactBuilder(pushImages bool, additionalEnv []string) *ArtifactBuilde
...
-	return cmd.Run()
+	if err := cmd.Start(); err != nil {
+		return errors.Wrap(err, "starting cmd")
+	}
+
+	return b.handleGracefulTermination(ctx, cmd)
+}
+
+func (b *ArtifactBuilder) handleGracefulTermination(ctx context.Context, cmd *exec.Cmd) error {
+	done := make(chan struct{})
```
tejal29: This looks really great. Just one question, i am assuming you have the done channel to not send kill signal if the process is already done? Lets put a defer close as soon as you create the channel

We can also shorten this code block further

```
func (b *ArtifactBuilder) handleGracefulTermination(ctx context.Context, cmd *exec.Cmd) error {
  done := make(chan struct{})
 defer(close(done))
  select {
  case <- ctx.Done():
    // Send interrupt signal if not windows.
  case <- time.After(termination*time.Second):
   // Send kill signal
  case <-done
 }
 err := cmd.Wait()
 return err
}


 }
```

I think this is it. And then we can merge this in!

Thank you for your patience.

cedrickring: According to your suggestion, the process will only run for `termination` seconds and then be killed. i think the nested `select` is still necessary to wait for the context to be cancelled...

tejal29: Sounds good.  Running kokoro tests now

PR #2893, opened 2019-09-16
Add buildpacks example to be referenced in buildpacks docs

priyawadhwa:
Add the same buildpacks example to `examples/buildpacks` and `integration/examples/buildpacks` so it can be referenced in the buildpacks tutorial in #2879


tejal29:
Can we also add them this in the integration test?

integration/run_test.go
```diff
@@ -131,9 +132,15 @@ func TestRun(t *testing.T) {
...
 			dir:         "testdata/custom",
 			pods:        []string{"bazel"},
+		}, {
+			description: "buildpacks",
+			dir:         "examples/buildpacks",
+			pods:        []string{"getting-started"},
 		},
 	}
 	for _, test := range tests {
+		setBuildpacksDefaultBuilder(t)
```
tejal29: This happens on every test. Should we do this before the `for` loop once?

balopat:
This looks great! 
One major feedback - I would move the tutorial part (which is great btw) into the example folder's README.md itself instead of having it in the `How-to` section,  to make it consistent with the rest of the sections.

PR #2901, opened 2019-09-17
Initial auto sync support design doc

loosebazooka:
Design doc for how we might try to solve the auto synchronization of files that require a build first.

Useful if want to support java servers that can reload classes and resources -- and perhaps many other user cases!

Staged Doc: https://github.com/GoogleContainerTools/skaffold/blob/sync-auto-proposal/docs/design_proposals/sync-auto-support.md


loosebazooka:
@corneliusweig FYI since I think you worked on the sync stuff recently?

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,150 @@
+# Title
+
+* Author(s): Appu Goundan (@loosebazooka)
+* Design Shepherd: \<skaffold-core-team-member\>
```
tejal29:
```suggestion
* Design Shepherd: Tejal Desai
```

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,150 @@
...
+5. *skaffold syncs files*
+
+which is far from ideal.
+
+#### Fix it?
+
+What we might want, is an API that can accept configuration that looks like:
+
+```
+sync: auto {
```
tejal29: We can drop `auto` and just add another mode `generated` along with existing `manual` and `infer`?
```suggestion+21
sync: 
  generated:
  - command: "./gradlew jib classes"
    inputs:
      - src/main/java
      - src/main/resources
    files:
      - src: "target/classes"
        dest: "/app/classes"
      - src: "target/resources"
        dest: "/app/resources"
  - command: "compile_c_stuff.sh"
    inputs:
    - asdf.c
    files:
     - src: "asdf.o"
       dest: "/libs/asdf.o"
  manual:
    - src: "local/computer/extrafiles"
      dest: "/extrafiles"
  includeTools: true
```

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,150 @@
...
+1. `sync: auto` - just a name for the mode, can be anything
+2. `indirect` - this means, you can't sync files directly, you need a command to
+   tranform them before sync. Indirect is also a list that can handle multiple
+   indirect types.
+     1. `command` - the command to run to generate syncable files
+     2. `inputs` - the inputs to watch to trigger `command`, if left empty, this
+        should just be considered the full list of inputs that are normally
+        watched for this build minus anything in `direct` (below).
+     3. `files` - the built files to sync (and where to sync them)
+3. `direct` - same as manual sync, just copy them directly, don't wait on some
```
tejal29: Would this replace manual? or another definition in `auto` ?

Currently, we don't support `manual` and `infer` sync rules together. 
https://skaffold.dev/docs/how-tos/filesync/#limitations
As per this proposal, do we plan to support `auto` as well `manual` sync modes?

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,150 @@
...
+   command to transform them
+4. `includeTools` - useful for containers that don't contain, for example, `tar`
```
tejal29: would this for cases where a generated file is bundled up in the container?
So the sync will now, 
- unzip the tar
- sync file 
- zip again?

corneliusweig:
@loosebazooka hey thanks for looping me in. I'll take a look at it!

docs/design_proposals/sync-auto-support.md
tejal29: how do specify which tools to include? Is there a separate DSL for that?

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Background
+
+Currently skaffold does not support `sync` for files that are generated
+during a build. For example when syncing java files to a container, one
+would normally expect `.class` files to be sync'd, but skaffold is
+really only aware of `.java` files in a build.
+
+1. Why is this required?
+  - Having a sync mode that has knowledge of files that need to be built before
```
tejal29: Not your fault, may be we should add it to our design doc proposal, 
Can we list some use cases we want to support. On the top of my mind

1. Users can sync files generated files
2. Users can sync generated files which are bundled inside a tar and inside a container.
3. Users can sync files outside of their current workspace? ( for now users can hook up a bash script which can take input ../../)
4.  Users can sync remote files?

With that in mind, what is the contract between Skaffold and `command` to generate these files.
Would skaffold run the command when it sees a change in the `command.inputs` files? In which case, `inputs` should be non-empty.

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,150 @@
...
+   and similar to debug, we can use some init container to inject the necessary
+   tooling for sync.
+
+### Open Issues/Questions
+
+Please list any open questions here in the following format:
+
+**How does a tool, like Jib, configure this**
+
+Jib will export the right data, similar to how `skaffoldFiles` currently exports
```
tejal29: This kind of answers my earlier comment https://github.com/GoogleContainerTools/skaffold/pull/2901#pullrequestreview-289605330
Can we lock down what this exported data looks like?

docs/design_proposals/sync-auto-support.md
priyawadhwa: You can add me as a design shepherd as well!

docs/design_proposals/sync-auto-support.md
loosebazooka: Given that sync does not currently support multiple sync configurations, this is designed is that this is all one large block that does many things. A builder can have some files that need to be compiled, and others that don't -- so if I was configuring this from jib, I could separate them out.

However this design may be flawed in that it redoes the work that other sync options already do.

docs/design_proposals/sync-auto-support.md
loosebazooka: I think `includeTools` maybe shouldn't be part of this design. While it's useful specifically for the default base image jib chooses (distroless/java), it could benefit be it's own design that answers these questions.

But to answer them anyway, since I'm here, I think for now it would include `tar` -- since that's what we need to do a sync. If the requirements of sync change, it could include whatever else `sync` needs. (that's why naming would be `includeTools` instead of `includeTar`)

corneliusweig: Having an option to add `tar` into the target container is a great idea. I also think that this is an idea worth being explored on its own. For example, it could solve https://github.com/GoogleContainerTools/skaffold/issues/1814. Consequently, I would see this option outside of the `auto` block and right under `sync`, so that other sync modes can also benefit from it.

docs/design_proposals/sync-auto-support.md
corneliusweig: I like this variant for the `yaml` syntax. As I understand, the existing `sync` specs would keep their semantics, but the `generated` is a new build hook to be triggered before sync. (Maybe `buildHooks` is also a more natural name than `generated`). In other words:

* if `generated` is present, execute the build hooks when changes matching `inputs` are detected
* if `generated` is absent -> current behavior.

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,150 @@
...
+  - command: "./gradlew jib classes"
+    inputs:
+      - src/main/java
+      - src/main/resources
+    files:
+      - src: "target/classes"
+        dest: "/app/classes",
+      - src: "target/resources.
+        dest: "/app/resources",
+  - command: "compile_c_stuff.sh"
```
corneliusweig: Have you thought about dependency loops here? What if the build hook is triggered by a change in file `A`, the build hook runs and updates `A` again. It's a case which can hardly be detected upfront as an error, right?

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,150 @@
...
+    inputs:
+    - asdf.c
+    files:
+     - src: "asdf.o"
+       dest: "/libs/asdf.o"
+  direct:
```
corneliusweig: Maybe I haven't thought enough about the edge cases. But is it really necessary to separate out the `direct` sync cases? What would be wrong about a sync rule that applies to both generated artifacts and direct source files?

docs/design_proposals/sync-auto-support.md
loosebazooka: Have not thought about this, I mostly work in the jib context where I don't expect that to happen. But detection of that should throw an error.

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,150 @@
...
+which is far from ideal.
+
+#### Fix it?
+
+What we might want, is an API that can accept configuration that looks like:
+
+```
+sync: auto {
+  indirect:
+  - command: "./gradlew jib classes"
```
corneliusweig: From a usability perspective, it is still quite some intellectual overhead to specify the build commands in the right way and define the proper sync rules. My gut feeling is that this procedure is a lot different from what most developers are used to, especially for java (if you do SpringBoot, you never care about `.class` files). Do you plan a further simplification so that the sync works out of the box for, say, SpringBoot?

docs/design_proposals/sync-auto-support.md
corneliusweig: But the question is: can we detect it at all?

docs/design_proposals/sync-auto-support.md
loosebazooka: I think the process of doing a build when it shouldn't could be costly?
And if the build system isn't smart, it could trigger sync'ing of files that didn't need to be (and this could cause a server reload when there didn't need to be?)

docs/design_proposals/sync-auto-support.md
loosebazooka: I think this config is mostly a manifestation of what the API would be for a builder like jib to define all this - not necessarily a user. So `jib` would just tell the sync system what the config is, and it would go about doing what needs to be done.

I'm not opposed to it, but I agree, it's fairly involved for a user to configure this.

docs/design_proposals/sync-auto-support.md
corneliusweig: My understanding was that the build hook is only triggered, if one of its input files changed (and Skaffold takes care to detect this). The syncing could be considered as a different step, where everything that changed is sync'ed. And if the build hook didn't run, none of its outputs would have changed. Or is that not how you planned it?

loosebazooka: Oh I see. I planned for it all to exist in the `auto` block in this example as a single sync rule. Naming is a little weird at the moment.

docs/design_proposals/sync-auto-support.md
loosebazooka: Yeah that's a hard problem :\

docs/design_proposals/sync-auto-support.md
priyawadhwa: If a file is present in both `generated` and `manual` we would either have to give one preference or throw an error. Personally I'm leaning towards throwing an error, so that user's don't end up with unexpected sync stuff happening.

docs/design_proposals/sync-auto-support.md
tejal29: hmm. i am still confused on it makes sense for me as a user writing this config.
```
sync:
   inlcudeTar: true
  generated:
   ....
```
If this is true,  skaffold would now install tar inside all containers built ?
If this is false, skaffold sync will fail potentially #1814 ?

If above is true; then 
As a user, if i have a  jib builder, i need to set this `true`.

I feel the users are might not be able to identify this. Can we instead, add an interface to builder
```
type Builder interface {
   SyncTools() []tools
}
```
And then skaffold in dev mode if sync enabled, will install these tools in the built container?

loosebazooka: It would make tar available to that specific artifacts container. Not all jib users *have* to do this, but the `distroless` collection of images do not have tar (or any tools really) included by default.

I would think, instead of asking the user to change their base image to enable sync, we just make it easier for them by making the tools available automatically. I assume it would a mechanism similar to how debugging tools are available to a pod right now using init containers (@briandealwis ?).

Again, perhaps I should take this out of this design doc and move it to a new one?

chanseokoh: Perhaps for simplicity, this doc could assume `tar` exists in the container or declares that this works only when `tar` exists. Installing other builds tools could be another doc, independent of this?

docs/design_proposals/sync-auto-support.md
chanseokoh: I'm still confused. Will the user be able to configure all these in `skaffold.yaml`, or is it that this doc just explains a proposed internal sync API in the form? In either case, what if the user configures other current sync options in their `skaffold.yaml`? Will they conflict?

docs/design_proposals/sync-auto-support.md
tejal29: ah understood, so maybe skaffold can be a little smart about if `tar` is included in the built image. 
Untill then, asking users to specify via config is easy way.
Skaffold can also suggest adding it to their config if the error is regarding `tar` not found.

docs/design_proposals/sync-auto-support.md
corneliusweig: In much the same vein, I'd recommend to list a few other use cases for the proposed yaml changes beside better Jib support. It sounds as if your main goal is to achieve better Jib integration, but that could be done with less user-facing changes. I'm a bit worried that the current proposal is very powerful, but also rather complex (both for users and maintenance).
If the sole intended use for this is Jib, there could be a simpler option without so many user-facing changes.

loosebazooka: sure, I just needed a way to describe the API in a way that could be read. There may not be any need at all to expose this beyond an internal API. It also helps explain how this might be extended beyond just jib usage (though that might be the only usecase for now).

loosebazooka:
It looks like I should rewrite this to reflect that this isn't really exposed to the user.

loosebazooka:
Updated this to reflect what is internal and external.

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,206 @@
...
+      dest: "/app/classes"
+      strip: "build/classes/java/main/"
+```
+
+A user's devloop then looks like this:
+
+1. run `./gradlew classes copyDependencies`
+2. run `skaffold dev`
+3. *make changes to some java file*
+4. run `./gradlew classes`
```
chanseokoh: I think a similar question was asked, but what happens now when you change some Java source files is that it runs the Jib builder and rebuilds everything including `.class` files. So, modifying a source file triggers a full rebuild which in turn will trigger the proposed sync. Somehow, we should be able to determine if we should forgo a full build and go for the sync route only.

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,206 @@
...
+    dockerbuild. Intermediate build artifacts might still be possible to
+    determine, however that would require an extra mechanism to do a partial
+    docker build and sync files from a built container to a running container --
+    something we do not intend to cover here.
+
+#### Problems with current API/config
+The current `sync` system has the following problems:
+1. No way to trigger local external processes - for example, in jib, to build a
+   container, one would run `./gradlew jib`, but to update class files so the
+   system may sync them, one would only be required to run `./gradlew classes`.
```
chanseokoh: Does Cloud Code (or IntelliJ in general) puts build outputs into the usual `build/` or `target/` directory? (Eclipse doesn't; it outputs to a separate directory that doesn't interfere with command-line Gradle/Maven.) Also, does IntelliJ actually runs command-line Gradle/Maven by default when the user modifies a Java source file inside the IDE? (I'm guessing no for the latter question.) I'm asking because IDEs will compile `.class` files on-the-fly when the source files are modified inside the IDE and places into the default build output directory, the user won't actually need to run `./gradlew classes`.

loosebazooka: Yeah, so I don't think we'll be able to solve the ide/build-system cooperation problem here. What we expect is that saved source changes will trigger a rebuild+sync through skaffold-jib. What the ide does might be independent/irrelevant.

example: 
- user changes MyClass.java
  - ide compiles and outputs to its own output directory
    - nothing further happens related to the IDE
  - skaffold sees `MyClass.java` has changed
    - skaffold triggers a `gradle classes` to build classes
    - skaffold syncs classes with remote container.

docs/design_proposals/sync-auto-support.md
loosebazooka: Yeah the assumption is that, at first, we can sync everything. Anything that jib defines as a file in the container will be sync'd when it changes.

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,212 @@
...
+```
+
+
+### Open Issues/Questions
+
+Please list any open questions here in the following format:
+
+** What happens if the build definition changes? **
+I'm not sure if we should just allow the user to go nuts, or if we should
+explicitly disallow this, because changing the build definition might result in
```
tejal29: Current behavior is when a user changes `skaffold.yaml` when a `skaffold dev` loop is running, a build and deploy is triggered.
So, this should be fine?

loosebazooka: Yeah, I guess we can just pick a mode for now and define it, change it later if needed? If the java build definition changes, then that triggers a full rebuild too.

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,213 @@
...
+a post-sync state that is broken.
+
+** What about files that have specific permissions? **
+Jib allows users to customize file permissions, for instance a file on the
+container can be configured to be 0x456 or something. One option instead of
+dealing with this, is to just make all sync'd files 777? Or we allow passthrough
+of permissions from the build system to the sync system.
+
+**Should we allow the user to configure the auto block?**
```
chanseokoh: This is an interesting question. Just like the user may want to ignore some files handled by "auto", they may want to sync files not synced. But if we don't allow any, "auto" becomes quite restricted, and the user is left with no good option (i.e., the status quo) for the dev loop scenario we are talking about if they wanted to add or remove extra sync rules.

At the same time, it would be difficult for the user to know what files a builder auto-syncs to where unless they are well documented, so manually adding extra "ignored" or "syncables" may not be straightforward.

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,255 @@
...
+build:
+  artifacts:
+  - image: ...
+    context: jib-project
+    jib: {}
+    sync:
+      auto: {}
+```
+
+#### Option 1: Delegate generation of sync tar to builder
```
loosebazooka: @corneliusweig @tejal29 added in this section about the builder being responsible for the syncable tar...

I can see the problem that it kind of ties down the sync implementation to tar.

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,262 @@
...
+    context: jib-project
+    jib: {}
+    sync:
+      auto: {}
+```
+
+
+#### Get necessary information from the builder
+
+Skaffold can expose an API that can accept a complex configuration on how
```
tejal29: looks good. so builder's will implement a new interface AutoSync [here](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/build.go#L37)

```
type Builder interface {
        SyncMap(ctx context.Context, artifact *latest.Artifact) (map[string][]string, error)
       // for given artifact, give the auto sync files. 
        AutoSyncMap(ctx, artifact *latest.Artifact)(AutoSync, error)

}

```

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,292 @@
+# Title
+
+* Author(s): Appu Goundan (@loosebazooka)
+* Design Shepherd: Tejal Desai (@tejal29), Priya Wadhwa (@priyawadhwa)
```
balopat:
```suggestion
* Design Shepherds: Tejal Desai (@tejal29), Balint Pato (@balopat)
```
I'm taking over this effort for this quarter :)

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,292 @@
...
+newSyncState = buildAndCalculateNewSyncState("this project")
+syncStates["this project"] = newSyncState
+
+diff = diff(lastSyncState, newSyncState)
+return SYNC{files in diff}
+```
+
+## Integration test plan
+
+*TBD*
```
balopat:
```suggestion

This can be a Kind based test, we don't need GCP secrets. 
Hence: 
- need to install gradle or maven on travis integration test jobs outside the cluster for testing the build script
- the test would do the following steps: 
    - spin up a simple java app built by jib with skaffold dev + autosync enabled
    - get pod name
    - change a .java file (should trigger generation of a class file) 
    - check with kubectl inside the pod (if pod not found as rebuild was triggered, fail) that the class file is copied over (compare content with external version) 
```

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,292 @@
...
+The following implementation had a few problems and that were potential deal
+breakers, this implementation is left in here for information purposes, no
+attempt was made to implement this. It has the following issues:
+1. Depends on `sync` using `tar` which should be an implementation detail
+2. Doesn't really provide a good base for exposing the `auto` sync configuration
+   block. As skaffold takes care of fewer things, the user must be responsible
+   for implementing it.
+
+#### Delegate generation of sync tar to builder
+
```
balopat: This works for solutions where we can influence the development of the build system (i.e. jib). But it can't be the only solution, as if we can't influence the build system, we need to implement the solution in skaffold.

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,292 @@
+# Title
+
+* Author(s): Appu Goundan (@loosebazooka)
+* Design Shepherds: Tejal Desai (@tejal29), Balint Pato (@balopat)
+* Date: 09/17/2019
+* Status: New
```
balopat:
```suggestion
* Status: Implementation in progress
```

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,292 @@
...
+would normally expect `.class` files to be sync'd, but skaffold is
+really only aware of `.java` files in a build.
+
+1. Why is this required?
+  - We would like to support the following features
+      - Skaffold can sync files that are not watched (these files are not
+        considered inputs for the container builder)
+      - Skaffold can sync files that are generated by a buildscript
+      - And one or both of the following:
+          - Skaffold can be configured to run a custom script to when certain files change
```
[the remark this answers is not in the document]
loosebazooka: The default build script isn't necessarily the one you want to use to generate intermediate assets, for example 
- jib full container build: `./gradlew jib`
- jib build requirements: `./gradlew classes`

docs/design_proposals/sync-auto-support.md
```diff
@@ -0,0 +1,292 @@
...
+I would expect to add a task like `_jibSkaffoldSyncMap` that will produce
+json output for the skaffold-jib intergration to consume and forward to the sync
+system. And example output might look like:
+
+```
+BEGIN JIB JSON: SYNCMAP/1
+{
+  "generated": [
+    {
+      src: “target/classes”
```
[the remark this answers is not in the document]
loosebazooka: I think the problem is the jib stuff is baked in, whereas exposing this externally would expose some of those external features a little more? (inputs/command)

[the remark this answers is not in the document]
balopat: I see, gotcha.

balopat:
Merged with admin rights - CI build is nonsensical here.

PR #2897, opened 2019-09-17
Optional secret

prary:
pull secret should be optional
Resolving issue #2888

Issue #2888, opened 2019-09-16
PullSecretName is never null


prary:
https://github.com/GoogleContainerTools/skaffold/blob/a892d7910f74fab9ead9930da3e3c26b74771757/pkg/skaffold/build/cluster/secret.go#L33

PullSecretName is set to DefaultKanikoSecretName in case it is empty.

https://github.com/GoogleContainerTools/skaffold/blob/a892d7910f74fab9ead9930da3e3c26b74771757/pkg/skaffold/schema/defaults/defaults.go#L213

balopat:
This is resolved in #2910.

PR #2910, opened 2019-09-19
Optional pull secret

prary:
Dropping PR #2897 
pull secret should be optional
Resolving issue #2888


pkg/skaffold/build/cluster/sources/sources.go
```diff
@@ -82,28 +82,29 @@ func podTemplate(clusterDetails *latest.ClusterDetails, artifact *latest.KanikoA
...
-				VolumeSource: v1.VolumeSource{
-					Secret: &v1.SecretVolumeSource{
-						SecretName: clusterDetails.PullSecretName,
-					},
-				},
-			},
-			},
 		},
 	}
+	if clusterDetails.PullSecretName != "" {
```
tejal29: this function is becoming really long, can we refactor this into 

```suggestion+15
	if clusterDetails.PullSecretName != "" {
	  vms, vs := getVolumeMountsAndVolumes()
	  pod.Spec.Containers[0].VolumeMounts = vms
	  pod.Spec.Volumes = vs  
```

priyawadhwa:
@prary do you mind if I add a test to this PR?

pkg/skaffold/build/cluster/sources/sources.go
prary: ya sure

prary:
@priyawadhwa that would be great :D

priyawadhwa:
Looks like current tests actually cover these changes! I've just refactored the code a little and it should be good to go.

PR #2917, opened 2019-09-20
adding release comment management to all config.go

### LINKED ISSUE #2775: Better config versioning management

balopat:
Related to #2775.

**Description**

This "master PR" adds automation around managing comments on the config.go files that inform the user about whether the latest `config.go` is safe to modify or not (depending on whether it's released or not). 

The plan is to break this up to smaller PRs. However feedback is welcome here, and it is a working version that you can test for UX. 

**User facing changes**

Only contributor facing changes. 

**Before**

Currently the hack/check-schema-changes.sh script is the only way to know if the config that you're modifying is released or not (or checking manually). 

**After**

This PR adds: 

- a helper comment on all the currently released config.go files  
```
// !!! WARNING !!! This config version is already released, please DO NOT MODIFY the structs in this file.
```
- a helper comment on the current latest (v1beta15) that it is unreleased: 
```
// This config version is not yet released, it is SAFE TO MODIFY the structs in this file.
```
- a script that updates these comments in bulk: `go run hack/versions/cmd/update_comments/main.go`
- a script that marks the latest version released - and it is called from `release.sh` so that we flip the latest version to released
- update to the new_config script to make sure the new config release has the unreleased comment
- fixed go_diff.go to add `interface{}` type support that is in `v1beta5`

**Next PRs.**

to break this up to smaller PRs: 
- [x] (2 files) fix go_diff.go to add `interface{}` type support that is in `v1beta5`
- [ ] (6 files) the new comment management logic and hooking it up to `release` and `new_config` 
- [x] (20 files) update of the comments historically on all config.go files 

To finish #2775: 

- [ ] document the process in DEVELOPMENT.md

**Submitter Checklist**

review them:

**Reviewer Notes**

- [ ] Unit test added.

### DIFF
--- a/hack/versions/pkg/schema/comment.go
+++ b/hack/versions/pkg/schema/comment.go
@@ -1,0 +1,137 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package schema
+
+import (
+	"bytes"
+	"go/ast"
+	"go/parser"
+	"go/printer"
+	"go/token"
+	"io/ioutil"
+	"os"
+	"os/exec"
+)
+
+const releasedComment = `// !!! WARNING !!! This config version is already released, please DO NOT MODIFY the structs in this file.`
+const unreleasedComment = `// This config version is not yet released, it is SAFE TO MODIFY the structs in this file.`
+
+// historicalVersionComments is used to recognize whether an existing comment is a "release comment" or not.
+// If you want to change a (say releasedComment) comment historically on all files, then:
+// 1.) add the current version to `historicalVersionComments`
+// 2.) change the text of releasedComment (or unreleasedComment)
+// 3.) run `go run hack/versions/cmd/update_comments/main.go`
+// 4.) remove the updated version from `historicalVersionComments`
+var historicalVersionComments = []string{
+	releasedComment,
+	unreleasedComment,
+}
+
+func UpdateVersionComment(origFile string, released bool) error {
+	info, err := os.Stat(origFile)
+	if err != nil {
+		return err
+	}
+	content, err := updateVersionComment(origFile, released)
+	if err != nil {
+		return err
+	}
+
+	if err := ioutil.WriteFile(origFile, content, info.Mode()); err != nil {
+		return err
+	}
+
+	cmd := exec.Command("go", "fmt", origFile)
+	return cmd.Run()
+}
+
+func updateVersionComment(origFile string, released bool) ([]byte, error) {
+	fset := token.NewFileSet()
+
+	var commentString string
+	if released {
+		commentString = releasedComment
+	} else {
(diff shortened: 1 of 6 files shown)


hack/versions/pkg/diff/godiff_test.go
```diff
@@ -59,6 +59,25 @@ func TestCmpGoStructs(t *testing.T) {
 `,
 			b: `package a
 //a different comment
+`,
+			same:      true,
+			shouldErr: false,
+		},
+		{
+			name: "supported types",
```
tejal29: I see you added support for `interface` Type? is this test case for that that? 
```suggestion
			name: "supported interface types",
```

hack/versions/pkg/schema/comment.go
```diff
@@ -0,0 +1,137 @@
...
+	"os"
+	"os/exec"
+)
+
+const releasedComment = `// !!! WARNING !!! This config version is already released, please DO NOT MODIFY the structs in this file.`
+const unreleasedComment = `// This config version is not yet released, it is SAFE TO MODIFY the structs in this file.`
+
+// historicalVersionComments is used to recognize whether an existing comment is a "release comment" or not.
+// If you want to change a (say releasedComment) comment historically on all files, then:
+// 1.) add the current version to `historicalVersionComments`
```
tejal29: shd the order be preserved. i. e `historicalVersionComments[0]`  is old release comment and 1 is unreleasedComment?

hack/versions/pkg/schema/comment_test.go
```diff
@@ -0,0 +1,118 @@
...
+func TestAddComment(t *testing.T) {
+
+	tcs := []struct {
+		name     string
+		orig     string
+		expected string
+		released bool
+	}{
+		{
+			name:     "unreleased comment added on file",
```
tejal29: It would be easy to also cover `historicalVersionComments`  here by overriding it. 
The test can serve as a good documentation on how to use it.

hack/versions/pkg/schema/comment.go
```diff
@@ -0,0 +1,137 @@
...
+// 2.) change the text of releasedComment (or unreleasedComment)
+// 3.) run `go run hack/versions/cmd/update_comments/main.go`
+// 4.) remove the updated version from `historicalVersionComments`
```
priyawadhwa: sorry if I'm missing something, should this say "remove the current version from `historicalVersionComment`, since that's what was added in step 1?

balopat: oh yes, that's right

hack/versions/pkg/schema/comment.go
balopat: order doesn't matter!

hack/versions/pkg/schema/comment_test.go
balopat: As I'm thinking about this - I should rename this to `knownComments` and we should fail on a file if it doesn't have a `knownComment` as the first comment.

PR #2926, opened 2019-09-23
Include build args in cache hash generation

### LINKED ISSUE #2922: Cache hash does not change when using buildArgs
### Expected behavior
When I run `skaffold dev`, I expect to get unique cache hashes, when my `buildArgs` change.

### Actual behavior
The build is using the same cache key, regardless of the `buildArgs`

### Information

- Skaffold version: v0.38.0
- Operating system: Ubuntu 19.04
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta14
kind: Config
profiles:
- name: reproduce
  build:
    local:
      useBuildkit: true
    artifacts:
    - image: reproduce
      docker:
        buildArgs:
          SOME_ARGUMENT: "{{.SOME_ARGUMENT}}"
```

1. Create a dockerfile to demonstrate that the cache is not unique per buildArg:
```
FROM alpine
ARG SOME_ARGUMENT
RUN echo ${SOME_ARGUMENT}
```
2. Run `SOME_ARGUMENT=without-cache skaffold dev --profile=reproduce`
```
Listing files to watch...
 - reproduce
List generated in 932.731µs
Generating tags...
 - reproduce -> reproduce:3178031-dirty
Tags generated in 2.94554ms
Checking cache...
 - reproduce: Not found. Building
Cache check complete in 389.237µs
Starting build...
Building [reproduce]...
[+] Building 0.4s (6/6) FINISHED
 => [internal] load build definition from Dockerfile
 => => transferring dockerfile: 99B
 => [internal] load .dockerignore
 => => transferring context: 2B
 => [internal] load metadata for docker.io/library/alpine:latest
 => CACHED [1/2] FROM docker.io/library/alpine
 => [2/2] RUN echo without-cache
 => exporting to image
 => => exporting layers
 => => writing image sha256:7544cf3257de0eebbe5bcaf8136619fe76a4e21d9d4c7424740ccd94b583f375
 => => naming to reproduce:3178031-dirty
The push refers to repository [reproduce]
7a45fe20619c: Preparing
03901b4a2ea8: Preparing
03901b4a2ea8: Layer already exists
7a45fe20619c: Pushed
3178031-dirty: digest: sha256:b1307ed2a65d50c875d503a2d236fdfedc86df5a937712e9b8b9c5e77b5684e6 size: 734
Build complete in 567.474661ms
Starting test...
Test complete in 4.307µs
Tags used in deployment:
 - reproduce -> reproduce:3178031-dirty@sha256:b1307ed2a65d50c875d503a2d236fdfedc86df5a937712e9b8b9c5e77b5684e6
Starting deploy...
```
3. Run `SOME_ARGUMENT=with-cache skaffold dev --profile=reproduce`
```
Listing files to watch...
 - reproduce
List generated in 2.337227ms
Generating tags...
 - reproduce -> reproduce:3178031-dirty
Tags generated in 6.120024ms
Checking cache...
 - reproduce: Found
Cache check complete in 52.519435ms
Tags used in deployment:
 - reproduce -> reproduce:3178031-dirty@sha256:b1307ed2a65d50c875d503a2d236fdfedc86df5a937712e9b8b9c5e77b5684e6
Starting deploy...
```

priyawadhwa:
Fixes #2922

**Description**

Include build args in cache hash generation. This should fix the bug
described in #2922, where changing buildArgs resulted in the same
cache key, causing incorrect deployments.

**Submitter Checklist**

review them:

**Reviewer Notes**

- [x] The code flow looks good. 
- [x] Unit test added.
- [x] User facing changes look good.

**Release Notes**

Describe any user facing changes here so maintainer can include it in the release notes, or delete this block.

Include build args in cache hash generation. This should fix the bug
described in #2922, where changing buildArgs resulted in the same
cache key, causing incorrect deployments.

### DIFF
--- a/pkg/skaffold/build/cache/hash_test.go
+++ b/pkg/skaffold/build/cache/hash_test.go
@@ -189,10 +189,9 @@ func TestBuildArgsEnvSubstitution(t *testing.T) {
 		t.Override(&artifactConfigFunction, fakeArtifactConfig)
 
 		depLister := &stubDependencyLister{dependencies: []string{"dep"}}
-		actual, err := getHashForArtifact(context.Background(), depLister, artifact)
+		hash1, err := getHashForArtifact(context.Background(), depLister, artifact)
 
 		t.CheckNoError(err)
-		t.CheckDeepEqual("afc018228939a12195994275d6a24ef8f0eb7fa5019ae0582b2daf9e4c353656", actual)
 
 		// Make sure hash is different with a new env
 
@@ -200,10 +199,12 @@ func TestBuildArgsEnvSubstitution(t *testing.T) {
 			return []string{"FOO=baz"}
 		}
 
-		actual, err = getHashForArtifact(context.Background(), depLister, artifact)
+		hash2, err := getHashForArtifact(context.Background(), depLister, artifact)
 
 		t.CheckNoError(err)
-		t.CheckDeepEqual("f698ab606ce86ad1c6b7842890a6573060dfaa770e6df0913076f83f14ac32ae", actual)
+		if hash1 == hash2 {
+			t.Fatal("hashes are the same even though build arg changed")
+		}
 	})
 }
 
--- a/pkg/skaffold/build/cache/hash_test.go
+++ b/pkg/skaffold/build/cache/hash_test.go
@@ -88,6 +88,36 @@ func TestGetHashForArtifact(t *testing.T) {
 				},
 			},
 			expected: "09b366c764d0e39f942283cc081d5522b9dde52e725376661808054e3ed0177f",
+		}, {
+			description: "build args",
+			artifact: &latest.Artifact{
+				ArtifactType: latest.ArtifactType{
+					DockerArtifact: &latest.DockerArtifact{
+						BuildArgs: map[string]*string{"one": stringPointer("1"), "two": stringPointer("2")},
+					},
+				},
+			},
+			expected: "f5b610f4fea07461411b2ea0e2cddfd2ffc28d1baed49180f5d3acee5a18f9e7",
+		}, {
+			description: "build args in different order",
+			artifact: &latest.Artifact{
+				ArtifactType: latest.ArtifactType{
+					DockerArtifact: &latest.DockerArtifact{
+						BuildArgs: map[string]*string{"two": stringPointer("2"), "one": stringPointer("1")},
+					},
+				},
+			},
+			expected: "f5b610f4fea07461411b2ea0e2cddfd2ffc28d1baed49180f5d3acee5a18f9e7",
+		}, {
+			description: "different build args",
+			artifact: &latest.Artifact{
+				ArtifactType: latest.ArtifactType{
+					DockerArtifact: &latest.DockerArtifact{
+						BuildArgs: map[string]*string{"one": stringPointer("")},
+					},
+				},
+			},
+			expected: "961582bfb8d159de1129f89fa12852308afce65816b5f3c521ee57cf9ec524d7",
 		},
 	}
 	for _, test := range tests {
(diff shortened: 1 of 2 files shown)


pkg/skaffold/build/cache/hash_test.go
```diff
@@ -195,3 +225,109 @@ func TestCacheHasher(t *testing.T) {
...
+				DockerArtifact: &latest.DockerArtifact{},
+			},
+		}, {
+			description: "kaniko artifact with build args",
+			artifactType: latest.ArtifactType{
+				KanikoArtifact: &latest.KanikoArtifact{
+					BuildArgs: map[string]*string{},
+				},
+			},
+			expected: map[string]*string{},
```
tejal29: Can we add some build args and test the expected values?

pkg/skaffold/build/cache/hash_test.go
```diff
@@ -88,6 +88,36 @@ func TestGetHashForArtifact(t *testing.T) {
 				},
 			},
 			expected: "09b366c764d0e39f942283cc081d5522b9dde52e725376661808054e3ed0177f",
+		}, {
```
tejal29: We should add tests to make sure the hash is not equal for a artifacts with different build args and  Two identical artifact should yield the same hash.

pkg/skaffold/build/cache/hash_test.go
priyawadhwa: I do that it `TestGetHashForArtifact`, this is just to make sure we get the build args if they are defined.

pkg/skaffold/build/cache/hash_test.go
priyawadhwa: The current tests should cover both of those:

`build args`
`build args in different order`
`different build args`

Multiply:
Would it make sense to cover build artifacts that use things like environment variables as well?

priyawadhwa:
Hey @Multiply so build args actually don't support environment variable substitution right now in the skaffold config.

Multiply:
@priyawadhwa I'm using them locally at the moment. (using buildkit(

priyawadhwa:
@Multiply ah gotcha, I'll add that in

pkg/skaffold/build/cache/hash_test.go
tejal29: umm no, we don't explicitly test if the artifact has is different if everything else is same. We are only checking for value. 
In future, if someone changes code here (and by mistake disregards build args), they might just change `expected` to match what was actual. It does not guard against breaking this in future.

Can we move all these cases to `TestArtifactConfig` where you explicitly compute hash on 2 artifacts and check if artifact hash is not equal when artifacts is same type but 
1. Build args different
2. Build args are same. 
3. Build args are same but different order

tejal29:
Lets not do 2 things in one PR. Can we please move the environment substitution to another PR?

Thanks 
Tejal

pkg/skaffold/build/cache/hash_test.go
priyawadhwa: Yup, ended up writing a new test, `TestBuildArgs`. Thanks!

priyawadhwa:
So we need to 

1. do the environment substitution on the build args
2. pass in those build args to the hasher

I don't think it makes sense to split these up, because I can't write tests for the former without the code for the latter.

tejal29:
> So we need to
> 
> 1. do the environment substitution on the build args
> 2. pass in those build args to the hasher
> 
> I don't think it makes sense to split these up, because I can't write tests for the former without the code for the latter (since I'm testing that the hash at the end is correct. The environment substitution itself is a function in a different package which I'm reusing)

Ah fair enough, you haven't written any extra logic for testing the environment solution. just 1 line change
```
buildArgs, err := docker.EvaluateBuildArgs(buildArgs)
```

pkg/skaffold/build/cache/hash_test.go
```diff
@@ -160,6 +161,44 @@ func TestArtifactConfig(t *testing.T) {
...
+		// Make sure hash is different with a new env
+
+		util.OSEnviron = func() []string {
+			return []string{"FOO=baz"}
+		}
+
+		actual, err = getHashForArtifact(context.Background(), depLister, artifact)
+
+		t.CheckNoError(err)
+		t.CheckDeepEqual("f698ab606ce86ad1c6b7842890a6573060dfaa770e6df0913076f83f14ac32ae", actual)
```
tejal29: Would it make sense to change the test to just compare the hashes are not equal after you change the env variable value?
That ways 
- we are checking the functionality that artifact hashes are not equal for different env values in build args..
- test does not rely on changes in the hashing logic which is irrelevant to this test case.

priyawadhwa: cool, done

tejal29:
Failed due to `TestGetSetFileValues/multiple_value` out of order.

PR #2929, opened 2019-09-24
handle StatusCheck Events implementation logic.

tejal29:
Relates to #172

**Description**

**User facing changes**

n/a

**Before**

n/a
**After**

n/a

**Next PRs.**

- Actually call these events in pkg/skaffold/deploy/status_check.go

**Submitter Checklist**

review them:

- [ X] Includes [unit tests](../DEVELOPMENT.md#creating-a-pr)
- [n/a ] Mentions any output changes.
- [ n/a] Adds documentation as needed: user docs, YAML reference, CLI reference.
- [ na/] Adds integration tests if needed.

**Reviewer Notes**

- [ ] Unit test added.

**Release Notes**

Describe any user facing changes here so maintainer can include it in the release notes, or delete this block.


pkg/skaffold/event/event.go
```diff
@@ -33,6 +33,8 @@ const (
 	Complete   = "Complete"
 	Failed     = "Failed"
 	Info       = "Information"
+	Started    = "Started"
```
priyawadhwa: We could use "In Progress" instead of "Started" (since In Progress implies that a process has Started) and "Complete" instead of "Succeeded"? That way we have consistency with other events. WDYT?

tejal29: So, I had it  "completed" instead of "succeeded" befor to be consistent. But one of the comments from IDE team on the internal doc was to change it to "succeeded" so it goes well with "Failed"

We shd change "completed" to "succeeded" which I can do in another PR.

Regarding "started" to "in progress",  there is a delay between when the status check started va when we first a first update.  Hence I added Started for the first event and subsequent events are "in Progress" with a text update on how many resources are pending

tejal29: The other alternative is to have a In progress with message "2/2 deployment are still pending" instead of strated

priyawadhwa: Sounds good, could we open an issue to change "completed" to "succeeded"?

Personally I'm leaning towards having "In Progress" with the message you said above, just to match, but I don't feel super strongly about it.

PR #2946, opened 2019-09-26
Random secret postfixes for kaniko

### LINKED ISSUE #2672: docker-cfg secret should be suffixed with uuid
Scenario 1
Sometime skaffold run fails after creating a docker-cfg secret when skaffold is re-run then it find the already created docker-cfg secret and hence fails. This can be avoided by generating random secret name and using that further.

Scenario 2 
Skaffold run parallel doing different deployment in same namespace. In such scenario one out of two run fails since it find secret already created in the namespace.

### Expected behavior
In normal scenrio skaffold does the clean up of dokcer-cfg secret but when it fails due to some reason then it fails to delete the docker-cfg secret.
### Actual behavior
Either we should use the previously created docker-cfg secret or should always created secret with uuid attach to docker-cfg
### Information

- Skaffold version: Latest
- Operating system: any
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta11
kind: Config
build:
  artifacts:
  - image: ......./..../..../pythonserver
    kaniko:
      buildContext:
        localDir: {}
      cache: {}
  cluster:
    namespace: ns-ns-ns
    dockerConfig:
      path: ~/.docker/config.json
```

1. do a parallel run of skaffold in same ns 
2. alternatively do a skaffold run and kill it after docker-cfg secret is created and re-run it.

prary:
Fixes #2672. Adds configurable random postfixes to the docker and pull secrets for kaniko builds.


pkg/skaffold/build/cluster/secret.go
```diff
@@ -27,104 +27,79 @@ import (
...
 
-	color.Default.Fprintf(out, "Creating kaniko secret [%s/%s]...\n", b.Namespace, b.PullSecretName)
+func (b *Builder) setupDockerConfigSecret(out io.Writer) (func(), error) {
+	if b.DockerConfig == nil {
+		return func() {}, nil
+	}
+	return reCreateSecret(out, b.Namespace, b.DockerConfig.SecretName, b.DockerConfig.Path, "config.json")
+}
+
+func reCreateSecret(out io.Writer, ns string, secretName string, secretPath string, secretkey string) (func(), error) {
```
priyawadhwa: nit: let's call this function `recreateSecret`

prary:
hi @priya,

I was about to introduce a bug thanks for pointing out. I was trying to delete the exisiting secret and recreate it which could not be cleaned in earlier run. 

Right now I am bit confused as how it should behave. Here's how I am planning to solution it, help choosing the right one.

```func (b *Builder) setupPullSecret(out io.Writer) (func(), error) {
	if b.PullSecret == "" && b.PullSecretName == "" { // user is not using secret
		return func() {}, nil
	}
	if b.PullSecret != "" {
          Solution 1 -- Check if previously created secret exist and reuse it. Skaffold should not worry about old token, let user handle it.
          Solution 2 -- re-create secret  to replace old secrets which could not be cleaned up due to ungraceful shutdown or may be it contain expire token. 
	}
	// this could be inferred as 'user wants to use existing secret'
	checkIfSecretExist(b.PullSecretName) return func{}, nil
}
```

prary:
hi @Priya,

I was about to introduce a bug thanks for pointing out. I was trying to delete the exisiting secret and recreate it which could not be cleaned in earlier run.

Right now I am bit confused as how it should behave. Here's how I am planning to solution it, help in choosing the right one.
```
func (b *Builder) setupPullSecret(out io.Writer) (func(), error) {
	if b.PullSecret == "" && b.PullSecretName == "" { // user is not using secret
		return func() {}, nil
	}
	if b.PullSecret != "" {
          Solution 1 -- Check if previously created secret exist and reuse it. 
                              Skaffold should not worry about old token, let user handle it.
          Solution 2 -- re-create secret  to replace old secrets which could not be cleaned up
                              due to ungraceful shutdown or may be it contain expire token. 
	}
	// this could be inferred as 'user wants to use existing secret'
	checkIfSecretExist(b.PullSecretName) return func{}, nil
}
```

prary:
Hi @tejal29 any reason why make checks is failing?

prary:
hi @tejal29 @priyawadhwa 

I over estimated the changes for uuid secret PTAL.

priyawadhwa:
Hey @prary, we discussed this in an internal meeting. Instead of using UUID, let's add a field to the skaffold config:

```yaml
cluster:
    secretPolicy: recreateIfExists
```

the exact syntax can be different. But the idea is, if a user sets this, then we always delete an old secret if it exists and recreate it.

WDYT @tejal29 ?

prary:
With uuid we can give ability to use skaffold as a pipeline where multiple user can run skaffold build with there own config.json

priyawadhwa:
@prary just to clarify, your use case is that you have mulitple users, each with a different config.json, trying to build with kaniko in the same namespace?

prary:
@priyawadhwa you are absolutely right

pkg/skaffold/schema/defaults/defaults.go
```diff
@@ -217,8 +219,8 @@ func setDefaultClusterDockerConfigSecret(cluster *latest.ClusterDetails) error {
 	if cluster.DockerConfig == nil {
 		return nil
 	}
-
-	cluster.DockerConfig.SecretName = valueOrDefault(cluster.DockerConfig.SecretName, constants.DefaultKanikoDockerConfigSecretName)
+	uuid, err := uuid.NewUUID()
```
tejal29: This should be guarded by a flag and not happen everytime.

prary: you mean

```
if cluster.DockerConfig.SecretName== nil {
       uuid, err := uuid.NewUUID()
       cluster.DockerConfig.SecretName = valueOrDefault(cluster.DockerConfig.SecretName, constants.DefaultKanikoDockerConfigSecretName+uuid.String())		
}
```

prary: Do you mean adding field in skaffold.yaml like
```
cluster:
  randomSecret: True
```

tejal29: yes! sounds good.

dgageot:
@prary Could you please rebase your PR?

balopat:
@prary, I fixed the tests and rebased, I hope you're okay with these changes.

PR #2964, opened 2019-10-01
Allow configuring `jib` plugin type

### LINKED ISSUE #2963: Jib configuration block should allow explicitly setting type
With #2808 we merged the `jibMaven` and `jibGradle` blocks into a single `jib` block where we determine the plugin-type based on the presence of [key files like `build.gradle`](https://github.com/GoogleContainerTools/skaffold/blob/4145d76cbccd0ea46ee31541f5f550368e257fa3/pkg/skaffold/jib/jib.go#L110-L129).  This isn't fool-proof and we should allow setting the value explicitly.

briandealwis:
Fixes #2963.  

Note that this is a somewhat unorthodox change as it exposes an internal and previously ignored field from the `v1beta14`'s `jib` config.

**Description**

In #2808 we unified the `jibMaven` and `jibGradle` configuration blocks into a single `jib` block.  The Skaffold-Jib builder looks for certain files to determine the Jib plugin type (Maven or Gradle).  We've had a report where Skaffold-Jib was unable to guess the Jib plugin type.  This PR adds the ability to explicitly configure the `jib` block to specify the plugin type.

The unified [`JibArtifact` already has a `Type` field](https://github.com/GoogleContainerTools/skaffold/blob/567a0800792804933941fb44ae8698a8afe1b68e/pkg/skaffold/schema/latest/config.go#L781-L782) which allows explicitly specifying the builder type.  This field is internal and not marshalled.  The intent was to expose this field to encode a Go enum (with corresponding `MarshalYAML()` and `UnmarshalYAML()`) but I punted at the time as it required teaching the schema generator to output [JSON schema `enum`](https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values).  But I realized that the `pkg/skaffold/schema/validation` approach was easier, but it requires changing the type of `Type`.

**User facing changes**
This PR changes the type of `JibArtifact`'s `Type` field and exposes it.  As a result:

- `v1beta14` and `v1beta15`'s `jib` now have optional `type` fields
- `skaffold fix` on `v1beta13` `jibMaven` and `jibGradle` configurations will now have a `type` field

**Reviewer Notes**

- [ ] Unit test added.

**Release Notes**

`jib` artifacts can explicitly specify the plugin type as `maven` or `gradle`


pkg/skaffold/schema/validation/validation.go
```diff
@@ -187,3 +188,18 @@ func validatePortForwardResources(pfrs []*latest.PortForwardResource) []error {
...
+func validateJibPluginTypes(artifacts []*latest.Artifact) (errs []error) {
+	for _, a := range artifacts {
+		if a.JibArtifact == nil || a.JibArtifact.Type == "" {
+			continue
+		}
+		t := strings.ToLower(a.JibArtifact.Type)
+		if t == "maven" || t == "gradle" {
+			continue
+		}
+		errs = append(errs, fmt.Errorf("artifact %s has invalid plugin type '%s'", a.ImageName, t))
```
chanseokoh: Is it possible that `a.ImageName` is empty?

"... has invalid Jib plugin type..." ?

pkg/skaffold/schema/validation/validation.go
```diff
@@ -187,3 +188,18 @@ func validatePortForwardResources(pfrs []*latest.PortForwardResource) []error {
...
+	}
+	return
```
chanseokoh: `return errs`

briandealwis: It's a named return variable (this was a copy-n-pasta of `validateDockerNetworkMode` above

pkg/skaffold/schema/validation/validation.go
briandealwis: Image names are validated further above in `validateImageNames()`.  There's not much we can do at this point.

pkg/skaffold/schema/latest/config.go
```diff
@@ -778,6 +778,8 @@ type JibArtifact struct {
 	// For example: `["--no-build-cache"]`.
 	Flags []string `yaml:"args,omitempty"`
 
-	// Type the Jib builder type (internal: see jib.PluginType)
-	Type int `yaml:"-"`
```
balopat: This should be fine as it actually helps making the upgrade process to not break when the detection mechanism doesn't work. It is a conservative change. I will use admin rights to merge this PR.

briandealwis:
Sorry @balopat, this totally slipped through the cracks.

briandealwis:
I need to update this in light of #3041

briandealwis:
I hacked up the schema generator to emit JSON schema `enum`s.  PTAL!

balopat:
This is awesome, thanks for putting the enum together as well!

balopat:
Merged with admin privileges for the historical version change as it actually helps with upgrades.

PR #2975, opened 2019-10-02
Add PullSecretMountPath to ClusterDetails

### LINKED ISSUE #731: Skaffold/Kaniko with ECR

priyawadhwa:
Relates to _in case of new feature, this should point to issue/(s) which describes the feature_

Fixes #731 

**Description**

This will give users the option to specify where the pull secret should
be mounted within the container. This should fix #731 and enable ECR
support.

**User facing changes**

n/a

**Next PRs.**

n/a

**Submitter Checklist**

review them:

**Reviewer Notes**

- [ ] Unit test added.

### DIFF
--- a/pkg/skaffold/schema/defaults/defaults.go
+++ b/pkg/skaffold/schema/defaults/defaults.go
@@ -217,6 +217,7 @@ func setDefaultClusterPullSecret(cluster *latest.ClusterDetails) error {
 		}
 		cluster.PullSecret = absPath
 		cluster.PullSecretName = valueOrDefault(cluster.PullSecretName, constants.DefaultKanikoSecretName)
+		cluster.PullSecretMountPath = valueOrDefault(cluster.PullSecretMountPath, constants.DefaultKanikoSecretMountPath)
 		return nil
 	}
 	return nil
--- a/pkg/skaffold/schema/defaults/defaults.go
+++ b/pkg/skaffold/schema/defaults/defaults.go
@@ -210,6 +210,7 @@ func setDefaultClusterTimeout(cluster *latest.ClusterDetails) error {
 }
 
 func setDefaultClusterPullSecret(cluster *latest.ClusterDetails) error {
+	cluster.PullSecretMountPath = valueOrDefault(cluster.PullSecretMountPath, constants.DefaultKanikoSecretMountPath)
 	if cluster.PullSecret != "" {
 		absPath, err := homedir.Expand(cluster.PullSecret)
 		if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/skaffold/schema/defaults/defaults.go
```diff
@@ -210,6 +210,7 @@ func setDefaultClusterTimeout(cluster *latest.ClusterDetails) error {
 }
 
 func setDefaultClusterPullSecret(cluster *latest.ClusterDetails) error {
+	cluster.PullSecretMountPath = valueOrDefault(cluster.PullSecretMountPath, constants.DefaultKanikoSecretMountPath)
```
tejal29: kaniko secret can be optional now. In that case, we need to move this down i guess?

pkg/skaffold/schema/defaults/defaults_test.go
```diff
@@ -111,6 +111,27 @@ func TestSetDefaultsOnCluster(t *testing.T) {
 
 		t.CheckNoError(err)
 		t.CheckDeepEqual(constants.DefaultKanikoSecretName, cfg.Build.Cluster.PullSecretName)
+		t.CheckDeepEqual(constants.DefaultKanikoSecretMountPath, cfg.Build.Cluster.PullSecretMountPath)
```
tejal29: Add a test for when the `secret` is optional and make sure `PullSecretMountPath` is not null.

pkg/skaffold/schema/defaults/defaults.go
priyawadhwa: This needs to be set for both `PullSecret` and `PullSecretName`, so I put it at the top. If neither of those are set, this will never be used, so it's ok to set it up here.

pkg/skaffold/schema/defaults/defaults_test.go
priyawadhwa: By "when the secret is optional" do you mean when `PullSecret` and `PullSecretName` are not set?

tejal29: yes i think that what @prary accomplished in #2910.

pkg/skaffold/schema/defaults/defaults.go
tejal29: umm, so we end up setting defaults for unused field.
Maybe, we should step back.. so, i looked at the bug you pointed, here. I will grab you offline.

PR #2984, opened 2019-10-04
link validation in docs

balopat:
Relates to _in case of new feature, this should point to issue/(s) which describes the feature_

**Description**

Enables link validation for relative links within our docs. This is useful as guards against mistakes. 
However, due to https://github.com/gohugoio/hugo/issues/2424 this means that links are now all opened in the same tab, even external ones. I believe that the validation is useful enough to take the hit on the external link openings. Opinions welcome :) 

**User facing changes**

External links are now opened in the same tab. (sideeffect of https://github.com/gohugoio/hugo/issues/2424) 

**Before**

External links are currently opened in a new tab (`target="_blank"`)

**After**

External links are now opened in the same tab. 

**Reviewer Notes**

### DIFF
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -17,6 +17,10 @@ defaultContentLanguageInSubdir = false
 # Useful when translating.
 enableMissingTranslationPlaceholders = true
 
+# validate links and fail if not found
+refLinksErrorLevel="ERROR"
+relativeURLs=true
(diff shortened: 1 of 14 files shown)


balopat:
oops, the doc previewer ran out of IPs (quota was 8, I requested an increase):
image

PR #3007, opened 2019-10-07
Reuse gcr.io tokens

dgageot:
Also auto-configure gcloud as a credential helper
when that makes sense.

### DIFF
--- a/pkg/skaffold/docker/authenticators.go
+++ b/pkg/skaffold/docker/authenticators.go
@@ -1,0 +1,106 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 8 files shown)


pkg/skaffold/docker/authenticators.go
```diff
@@ -0,0 +1,106 @@
...
+
+var authenticators = Authenticators{}
+
+// Authenticators stores an authenticator per registry.
+type Authenticators struct {
+	byRegistry map[string]*lockedAuthenticator
+	lock       sync.Mutex
+}
+
+// For retrieves the authentiator for a given image reference.
```
jonjohnsonjr: This reuse is something we could reasonably do in `google.Keychain`

dgageot: Go for it! It'll make it easier for Skaffold

jonjohnsonjr:
@dgageot what do you think about https://github.com/google/go-containerregistry/pull/603

dgageot:
I'm going to merge this one. When go-containerregistry has the code required to do the caching, I'll change Skaffold.

jonjohnsonjr:
It's merged now :smile:

PR #3014, opened 2019-10-08
Pass the context

dgageot:

### DIFF
--- a/pkg/skaffold/runner/diagnose.go
+++ b/pkg/skaffold/runner/diagnose.go
@@ -32,9 +32,7 @@ import (
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/skaffold/runner/diagnose.go
```diff
@@ -32,9 +32,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (r *SkaffoldRunner) DiagnoseArtifacts(out io.Writer) error {
```
tejal29: Does this mean, `DiagnoseArtifacts` now runs in foreground?

tejal29: We spoke offline, i misunderstood the change

PR #3016, opened 2019-10-08
add a resourceCounter to track pods

tejal29:
Relates to #176 

**Description**
In this PR add a `resourceCounter` which tracks `pods` and `deployments`
- The `pods` counter is initialized to 0 and used anywhere.

**User facing changes**

n/a

**Next PRs.**
- Change `getDeployStatus` and `printResourceStatus` to use the `resourceCounter.pod`
- Fetch pods and add them in the check

**Submitter Checklist**

review them:

- [ X ] Includes [unit tests](../DEVELOPMENT.md#creating-a-pr)
- n/a Mentions any output changes.
- n/a Adds documentation as needed: user docs, YAML reference, CLI reference.
- n/a Adds integration tests if needed.

**Reviewer Notes**

- [ ] Unit test added.

**Release Notes**


pkg/skaffold/deploy/status_check_test.go
```diff
@@ -413,3 +414,77 @@ func withStatus(d *resource.Deployment, details string, err error) *resource.Dep
...
+			c:           newCounter(10),
+			expected:    newCounter(10),
+		},
+		{
+			description: "counter with updated pending is copied correctly",
+			c:           &counter{total: 10, pending: 2},
+			expected:    &counter{total: 10, pending: 2},
+		},
+		{
+			description: "counter with updated failed and pendingis copied correctly",
```
balopat:
```suggestion
			description: "counter with updated failed and pendings copied correctly",
```

pkg/skaffold/deploy/status_check.go
```diff
@@ -212,9 +217,34 @@ func newCounter(i int) *counter {
...
+func (c *counter) copy() *counter {
+	if c == nil {
+		return nil
+	}
+	return &counter{
+		total:   c.total,
+		pending: c.pending,
+		failed:  c.failed,
+	}
+}
```
balopat: nit: newline

pkg/skaffold/deploy/status_check.go
```diff
@@ -212,9 +217,35 @@ func newCounter(i int) *counter {
...
+func (c *counter) markProcessed(err error) *counter {
 	if err != nil {
 		atomic.AddInt32(&c.failed, 1)
 	}
-	return int(atomic.AddInt32(&c.pending, -1))
+	atomic.AddInt32(&c.pending, -1)
+	return c.copy()
+}
+
+func (c *counter) copy() *counter {
```
balopat: why are we returning a pointer here? 
I think by value should be sufficient - we don't want to sound like that we are returning a mutable structure here -  also  `c == nil` will never evaluate to true!!

tejal29: make sense. Copy should return a non mutable value.

pkg/skaffold/deploy/status_check.go
```diff
@@ -220,9 +225,34 @@ func newCounter(i int) *counter {
 	}
 }
 
-func (c *counter) markProcessed(err error) int {
+func (c *counter) markProcessed(err error) counter {
```
balopat: what do you think about moving all these instance methods instead into functions, like the style of `append`? 
`func markProcessed(c *counter, err error) counter`? that will emphasize it even more that these are not state mutations.

PR #3040, opened 2019-10-12
Support port-forward address bind with skaffold dev

### LINKED ISSUE #1358: Support port-forward address bind with skaffold dev

feloy:
Fixes #1358. 

**Description**

A field `Address` has been added to the `PortForwardResource` config structure, which defaults to 127.0.0.1.

This value is then passed to `kubectl port-forward`

**User facing changes**

Ths user can configure the port forwarding on a specific address:

skaffold.yaml:
```
portForward:
- resourceName: go-hello-world-external
  resourceType: Service
  namespace: doc
  address: 0.0.0.0
  port: 80
  localPort: 8091
```

When port-forwarding, skaffold CLI displays the address to which the forward is bound:

```
Port forwarding service/go-hello-world-external in namespace test, remote port 80 -> address  0.0.0.0 port 4503
```

**Submitter Checklist**

review them:

**Reviewer Notes**

- [ ] Unit test added.

**Release Notes**

```
*  Support port-forward address bind with skaffold dev
A field `Address` has been added to the `PortForwardResource` config structure, which defaults to 127.0.0.1. This value is then passed to `kubectl port-forward`
```


feloy:
I cannot find a way to test this new behavior, as it is not possble to define a port forward to find a resource in the *current* namespace (and the integration tests are running in a random namespace). I opened an issue #3098 and a change in #3099.

Can you give me your opinion?

pkg/skaffold/util/port.go
```diff
@@ -71,17 +71,17 @@ func GetAvailablePort(port int, forwardedPorts ForwardedPorts) int {
...
 	}
 
-	return IsPortFree(p)
+	return IsPortFree(address, p)
 }
 
-func IsPortFree(p int) bool {
-	l, err := net.Listen("tcp", fmt.Sprintf("%s:%d", Loopback, p))
+func IsPortFree(address string, p int) bool {
+	l, err := net.Listen("tcp", fmt.Sprintf("%s:%d", address, p))
```
tejal29: if the address is not localhost, it means will be making a call to remote? I am not sure, what the user experinece would be then.
I feel, if the address is not `Loopback`, we should skip the check and return immediately.

pkg/skaffold/kubernetes/portforward/entry_manager_test.go
```diff
@@ -155,7 +155,7 @@ func AssertCompetingProcessesCanSucceed(ports util.ForwardedPorts, t *testing.T)
 	wg.Add(N)
 	for i := 0; i < N; i++ {
 		go func() {
-			port := util.GetAvailablePort(4503, ports)
+			port := util.GetAvailablePort("127.0.0.1", 4503, ports)
```
tejal29: why is this defaulted to `127.0.0.1` and not the address in the config?

feloy: I would like not to change the behaviour of this test, and create othe specific tests with non-loczlhodt addresses

pkg/skaffold/util/port.go
feloy: I think we should distinguish if the address is attached to the local host, or is a distant address.  For example, making nginx listen on a distant address will make it fail with an error '"Cannot assign requested address"

feloy:
> Thank you for your PR!.
> 
> Please explain, where we are passing the newly added `address` config.

Sorry, I'm not sure to understand. At which level would you like I explain this? In this user doc? In ths ources? In this tcket?

balopat:
I think this looks great. Can you please rebase?

feloy:
> I think this looks great. Can you please rebase?

Thanks, I rebased

PR #3043, opened 2019-10-14
Move the buildpacks/custom script sample to another folder

dgageot:
This is to prepare for the native support of buildpacks.


dgageot:
@balopat Makes sense! I've updated the PR

dgageot:
ok. It's too much of a pain. The CI now complains that there's a sample that's in `examples` but not in `integration/examples`. It is right to complain. So I'm going to close the PR for now

PR #3055, opened 2019-10-14
Delegate release notes to external tool

corneliusweig:
**Description**

This PR delegates the "heavy" lifting for creating the release changelog to an external tool hosted in my github space. I intend to maintain this tool and not break the formatting. @balopat also has write access to this repository and was involved in factoring out this piece of code.

The `hack/release_notes.sh` script installs the external helper program as `hack/release-notes`, if not present.

**User facing changes**

n/a

**Next PRs.**

n/a

**Reviewer Notes**

- [ ] Unit test added.


hack/release.sh
```diff
@@ -14,12 +14,30 @@
...
+
 DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 EXAMPLES_DIR=${DIR}/../examples
 INTEGRATION_EXAMPLES_DIR=${DIR}/../integration/examples
 
+install_release_notes_helper() {
+  release_notes_workdir="$(mktemp -d)"
+  trap 'rm -rf -- ${release_notes_workdir}' RETURN
+
+  # See https://stackoverflow.com/questions/56842385/using-go-get-to-download-binaries-without-adding-them-to-go-mod for this workaround
```
balopat: I would just add the tool to go mods!

corneliusweig: Can you explain a little how to do that? If I just add it with `go get .../release-notes`, then it will be lost on the next `go mod tidy`. Besides, I don't think this would install the tool in `$GOPATH/bin` or so.

dgageot: @corneliusweig One way to do it is to create a top level `tools` folder that is a sub-package, with it's own `go.mod`. In that folder, you add a `tools.go` file that imports `xxx/release-notes`. You can then run `go mod vendor` in that folder.
In the makefile, you'd run `cd tools; go install -mod=vendor xxx/release-notes`

dgageot: Here's a sample with `golangci-lint`

corneliusweig: @dgageot oh.. that's a nice way. I tried to stay away from vendor mode though, it seems to work ok either way.

dgageot: Why did you avoid vendoring? The advantage is that there's nothing to download on the Travis. It's a bit quicker and more resilient.

corneliusweig: It was adding 11MB of vendored code. Besides, it is only run on developer machines, right? So will not affect Travis anyway.

I know that you are not a big fan of go modules, so if you have a strong opinion in this case, I won't argue any longer :)

dgageot: Not a big fan indeed. In that case, it's actually very convenient. You are right that it doesn't run on Travis so, yes, vendoring is not so useful. Good point!

PR #3062, opened 2019-10-15
Improve skaffold init file traversal

### LINKED ISSUE #3060: Skaffold init with Jib init behaves differently in analyze mode
Skaffold v0.40

Project has Dockerfile and skaffold.yaml already present. We assume the generation of the Skaffold YAML is not necessary at this point. However:

`skaffold init --analyze`:
```
FATA[0000] pre-existing skaffold.yaml found
``` 
(expected)

`skaffold init --analyze --XXenableJibInit`:
```
{"builders":[{"name":"Docker","payload":{"path":"Dockerfile"}},{"name":"Jib Maven Plugin","payload":{"path":"pom.xml"}}],"images":[{"name":"java-hello-world","foundMatch":false}]}
```
(unexpected, we would notify user about skaffold YAML generation possibility although skaffold.yaml already present).

TadCordle:
Fixes #3060. 

**Description**

Currently, `skaffold init` ends directory traversal early when a Jib config is detected, so as to not accidentally traverse sub-modules, detect more Jib configurations, and end up with a list of redundant builders. However, the directory traversal had issues that could cause other important files to be missed, such as k8s configs, existing skaffold.yamls, or other build files in the Jib config directory (see before/after descriptions below). This PR changes `skaffold init` to fully walk the project directory, using a flag to ignore builders in nested directories instead of ignoring the directories altogether. This lets us find k8s manifests/skaffold.yamls regardless of where they are relative to Jib build files, while letting us ignore redundant builder configs.

**User facing changes**

`skaffold init` will now detect important files more consistently.

**Before**

Given the example project directory below:

```
Dockerfile
build.gradle  <-- Jib configured here
kubernetes-manifests/
pom.xml
skaffold.yaml
src/
target/
```

`skaffold init` would traverse the directory in alphabetical order, finding the dockerfile first, then finding Jib configured in the `build.gradle` first, causing it to stop and return to the parent directory. This causes it to miss another potential Jib config in the same directory (`pom.xml`), the kubernetes manifests, and the existing `skaffold.yaml`.

**After**

Now, `skaffold init` traverses all the files in a directory first before traversing sub-directories (solving the problem of missed builders/`skaffold.yaml`). If a Jib config is found, then it will not continue to search for builders in the sub-directories; however it will continue to traverse deeper to find k8s manifests (e.g. in `kubernetes-manifests/`).

**Submitter Checklist**

**Reviewer Notes**

- [ ] Unit test added.

**Release Notes**

Fix `skaffold init` ending file traversal early when Jib config is detected.

### DIFF
--- a/pkg/skaffold/initializer/init.go
+++ b/pkg/skaffold/initializer/init.go
@@ -30,6 +30,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/karrick/godirwalk"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"gopkg.in/AlecAivazis/survey.v1"
@@ -541,35 +542,69 @@ func printAnalyzeJSON(out io.Writer, skipBuild bool, pairs []builderImagePair, u
 func walk(dir string, force, enableJibInit bool, validateBuildFile func(bool, string) ([]InitBuilder, error)) ([]string, []InitBuilder, error) {
 	var potentialConfigs []string
 	var foundBuilders []InitBuilder
-	err := filepath.Walk(dir, func(path string, f os.FileInfo, e error) error {
-		if f.IsDir() && util.IsHiddenDir(f.Name()) {
-			logrus.Debugf("skip walking hidden dir %s", f.Name())
-			return filepath.SkipDir
-		}
-		if f.IsDir() || util.IsHiddenFile(f.Name()) {
-			return nil
-		}
-		if IsSkaffoldConfig(path) {
-			if !force {
-				return fmt.Errorf("pre-existing %s found", path)
-			}
-			logrus.Debugf("%s is a valid skaffold configuration: continuing since --force=true", path)
-			return nil
-		}
-		if IsSupportedKubernetesFileExtension(path) {
-			potentialConfigs = append(potentialConfigs, path)
-			return nil
-		}
-		// try and parse build file
-		if builderConfigs, err := validateBuildFile(enableJibInit, path); builderConfigs != nil {
-			for _, buildConfig := range builderConfigs {
-				logrus.Infof("existing builder found: %s", buildConfig.Describe())
-				foundBuilders = append(foundBuilders, buildConfig)
-			}
+
+	var dirCallback func(path string, findBuilders bool) error
+	dirCallback = func(path string, findBuilders bool) error {
+		dirents, err := godirwalk.ReadDirents(path, nil)
+		if err != nil {
 			return err
 		}
+
+		var directories []*godirwalk.Dirent
+		findBuildersInDirectories := true
+		for _, file := range dirents {
+			// If we found a directory, keep track of it until we've gone through all the files first
+			if file.IsDir() {
+				directories = append(directories, file)
+				continue
+			}
+
+			if util.IsHiddenFile(file.Name()) {
+				continue
+			}
+			filePath := filepath.Join(path, file.Name())
+			if IsSkaffoldConfig(filePath) {
+				if !force {
+					return fmt.Errorf("pre-existing %s found", filePath)
+				}
+				logrus.Debugf("%s is a valid skaffold configuration: continuing since --force=true", filePath)
+				continue
+			}
+			if IsSupportedKubernetesFileExtension(filePath) {
+				potentialConfigs = append(potentialConfigs, filePath)
+				continue
+			}
+			// try and parse build file
+			if findBuilders {
+				if builderConfigs, err := validateBuildFile(enableJibInit, filePath); builderConfigs != nil {
+					for _, buildConfig := range builderConfigs {
+						logrus.Infof("existing builder found: %s", buildConfig.Describe())
+						foundBuilders = append(foundBuilders, buildConfig)
+					}
+					if err == filepath.SkipDir {
+						findBuildersInDirectories = false
+					}
+				}
+			}
+		}
+
+		// Otherwise traverse deeper
+		for _, dir := range directories {
+			if util.IsHiddenDir(dir.Name()) {
+				logrus.Debugf("skip walking hidden dir %s", dir.Name())
+				continue
+			}
+
+			err = dirCallback(filepath.Join(path, dir.Name()), findBuildersInDirectories)
+			if err != nil {
+				return err
+			}
+		}
+
 		return nil
-	})
+	}
+
+	err := dirCallback(dir, true)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/pkg/skaffold/initializer/init.go
+++ b/pkg/skaffold/initializer/init.go
@@ -553,6 +553,8 @@ func walk(dir string, force, enableJibInit bool, validateBuildFile func(bool, st
 		var directories []*godirwalk.Dirent
 		findBuildersInDirectories := true
 		sort.Sort(dirents)
+
+		// Traverse files
 		for _, file := range dirents {
 			// If we found a directory, keep track of it until we've gone through all the files first
 			if file.IsDir() {
@@ -575,6 +577,7 @@ func walk(dir string, force, enableJibInit bool, validateBuildFile func(bool, st
 				potentialConfigs = append(potentialConfigs, filePath)
 				continue
 			}
+
 			// try and parse build file
 			if findBuilders {
 				if builderConfigs, err := validateBuildFile(enableJibInit, filePath); builderConfigs != nil {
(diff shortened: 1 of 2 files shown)


pkg/skaffold/initializer/init.go
```diff
@@ -541,35 +542,73 @@ func printAnalyzeJSON(out io.Writer, skipBuild bool, pairs []builderImagePair, u
...
+		if err != nil {
+			return err
 		}
-		if IsSkaffoldConfig(path) {
-			if !force {
-				return fmt.Errorf("pre-existing %s found", path)
+
+		var directories []*godirwalk.Dirent
+		findBuildersInDirectories := true
+		sort.Sort(dirents)
```
tejal29: is this necessary?

TadCordle: Tests seemed unpredictable otherwise (passed on my machine, failed on travis due to elements being in a different order).

pkg/skaffold/initializer/init.go
```diff
@@ -541,35 +542,73 @@ func printAnalyzeJSON(out io.Writer, skipBuild bool, pairs []builderImagePair, u
...
+
+		// Traverse files
+		for _, file := range dirents {
```
tejal29: this func is too long and hard to read. Can you reduce the number of branches or break it down further into smaller functions?.

TadCordle: I'm trying now, but so far every attempt to break it apart just leads to passing lots of arguments between the functions, and it ends up being harder to follow. I'll keep trying, but so far inlining it seems the best in my opinion.

pkg/skaffold/initializer/init.go
```diff
@@ -242,7 +243,10 @@ func autoSelectBuilders(builderConfigs []InitBuilder, images []string) ([]builde
 	return pairs, builderConfigs, unresolvedImages
 }
 
-func detectBuilders(enableJibInit bool, path string) ([]InitBuilder, error) {
+// detectBuilders checks if a path is a builder config, and if it is, returns the InitBuilders representing the
+// configs. Also returns a boolean marking whether search completion for subdirectories (true = subdirectories should
```
chanseokoh: "marking ~~whether~~ search completion", I think.

TadCordle:
@tejal29 I broke the long function down, LMK if it's an improvement.

pkg/skaffold/initializer/init.go
```diff
@@ -541,35 +542,73 @@ func printAnalyzeJSON(out io.Writer, skipBuild bool, pairs []builderImagePair, u
...
-	err := filepath.Walk(dir, func(path string, f os.FileInfo, e error) error {
-		if f.IsDir() && util.IsHiddenDir(f.Name()) {
-			logrus.Debugf("skip walking hidden dir %s", f.Name())
-			return filepath.SkipDir
-		}
-		if f.IsDir() || util.IsHiddenFile(f.Name()) {
-			return nil
+
+	var dirCallback func(path string, findBuilders bool) error
+	dirCallback = func(path string, findBuilders bool) error {
```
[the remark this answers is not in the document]
TadCordle: The function is recursive, so I had to declare it before instantiating it.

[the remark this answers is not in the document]
chanseokoh: Is the name "callback" appropriate? This just seems like an ordinary function with recursive calls. Doesn't look like some kind of function passed to someone else to be called later in reaction by them.

pkg/skaffold/initializer/init.go
```diff
@@ -538,40 +542,80 @@ func printAnalyzeJSON(out io.Writer, skipBuild bool, pairs []builderImagePair, u
...
 		}
-		// try and parse build file
-		if builderConfigs, err := validateBuildFile(enableJibInit, path); builderConfigs != nil {
-			for _, buildConfig := range builderConfigs {
-				logrus.Infof("existing builder found: %s", buildConfig.Describe())
-				foundBuilders = append(foundBuilders, buildConfig)
+
+		// Traverse into subdirectories
+		for _, dir := range directories {
+			err = dirCallback(filepath.Join(path, dir.Name()), findBuildersInDirectories)
```
chanseokoh: Eventually, you end up traversing all the sub-directories that you think you should, so I think it doesn't matter if you do BFS or DFS (if I'm not mistaken about this). Then I think doing DFS would simplify the code, removing the `directories` array. No?

TadCordle: I think it does matter that we do a BFS, because we want to search all the files in a directory before deciding whether to search subdirectories. For example, in a multi-module maven project, if we have the sub modules in subdirectories, we want to find the parent pom.xml before we enter those directories, or we risk running the `jib:_skaffoldInit` goal on a sub module instead of the parent.

pkg/skaffold/initializer/init.go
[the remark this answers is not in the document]
TadCordle: Hmm, you're probably right.

pkg/skaffold/initializer/init.go
chanseokoh: Oh, yeah, my misunderstanding.

pkg/skaffold/initializer/init.go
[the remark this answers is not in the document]
nkubala: @TadCordle ah ok thanks for clarifying. I agree with @chanseokoh about the naming.

pkg/skaffold/initializer/init.go
```diff
@@ -538,40 +542,80 @@ func printAnalyzeJSON(out io.Writer, skipBuild bool, pairs []builderImagePair, u
...
+			}
+
+			// If we found a directory, keep track of it until we've gone through all the files first
+			if file.IsDir() {
+				directories = append(directories, file)
+				continue
+			}
+
+			filePath := filepath.Join(path, file.Name())
+			if findBuildersInDirectories, err = checkFile(filePath, force, findBuilders, enableJibInit, &potentialConfigs, &foundBuilders); err != nil {
```
chanseokoh: I think `findBuildersInDirectories` should be some sort of accumulation that assignment. It can't be that you just overwrite the previous value with the most recent, no?

TadCordle: In practice that will never happen in an unintended way, but I can try to make the code more clear about it.

chanseokoh: Hmm... I think the problem I found is like this. Let's say we have two files in the root directory:

`/fileA`: `detectBuilders` inside `checkFile` will return true for `findBuildersInSubdirectories`.
`/fileB`: `detectBuilders` will return false.

Let's assume `findBuilders` is true at the root loop. Note this `findBuilders` doesn't change inside the loop, always being true.

So, if we check the files in the order A -> B in the loop, `findBuildersInSubdirectories` will end up being false after the loop terminates. OTOH, if the loop goes B -> A, it will end up being true, because the last `checkFile` assignment will win.

And this was the problem that I had originally pointed out. Then I told you later offline that this seemed okay, but looking at this again, now I think this problem still stands for both before and after your last commit.

TadCordle: I think I finally see what you mean by accumulation now, and I think you're right. Ugh. I will fix it soon.

pkg/skaffold/initializer/init.go
```diff
@@ -538,40 +542,80 @@ func printAnalyzeJSON(out io.Writer, skipBuild bool, pairs []builderImagePair, u
...
+		dirents, err := godirwalk.ReadDirents(path, nil)
+		if err != nil {
+			return err
 		}
-		if IsSkaffoldConfig(path) {
-			if !force {
-				return fmt.Errorf("pre-existing %s found", path)
+
+		var directories []*godirwalk.Dirent
+		continueSearching := findBuilders
```
chanseokoh: I like the name `searching`. How about the other way around? Like

`searchBuildersInSubdirectories := findBuilders`

here and

`builderConfigs, continueSearchingBuilders := detectBuilders(...)`
`searchBuildersInSubdirectories = searchBuildersInSubdirectories && continueSearchingBuilders`

below? Because `continueSearching` as of now could give the impression that it is for not breaking out some searching inside the loop. And since we have `foundBuilders`, I find `search` gives more distinguishing feeling for booleans vs builders found. That said, we have `foundBuilders` and `foundConfig`, but the latter is a boolean, so I find it unfortunately too. But not sure what to do about it.

And for `var directories []*godirwalk.Dirent`, how about `subdirectories`?

And for `entryDirectory`, I think something like `searchConfigsAndBuilders` as a recursive search function.

But these are small naming suggestions that I don't feel strongly about, so feel free to go with what you feel comfortable.

TadCordle:
@tejal29 @nkubala Would you mind giving another review? @chanseokoh's approval isn't enough since this is outside of Jib files.

PR #3092, opened 2019-10-18
Clarify debug docs for deprecated Workload APIs

briandealwis:
Minor doc modification to expand on `debug` not handling deprecated workload objects.


briandealwis:
Unsure why but `{{% alert %}}` doesn't seem to render the content as markdown.  Moved this warning into the _Limitations_ section.  Which is where it really belonged.

Issue #3098, opened 2019-10-19
Default namespace of port forwarding


feloy:
### Expected behavior

When I define a port forward in my skaffold.yaml file without defining a namespace on which to find the resource to port forward, I expect that skaffold look in the *current* namespace.

### Actual behavior

If the namespace is not specified, the "default" namespace is used.

### Discussion

This behavior seems useless to me, because it is easy for the user to specify the "default" namespace if he wants to work on this namespace. Also, it is practically impossible that "default" is the expected namespace, as it is recommended not to use this "default" namespace.

In the other hand, it could be interesting that different users are able to use the same skaffold.yaml file in different namespaces. With the actual behavior, it is impossible without editing the skaffold.yaml file. It would be possible by letting skaffold select the current namespace by default.

### Information

- Skaffold version: master
- Operating system: Linux
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta17
kind: Config
build:
  artifacts:
    - image: gcr.io/k8s-skaffold/leeroy-web
      context: ./leeroy-web/
    - image: gcr.io/k8s-skaffold/leeroy-app
      context: ./leeroy-app/
deploy:
  kubectl:
    manifests:
      - ./leeroy-web/kubernetes/*
      - ./leeroy-app/kubernetes/*
portForward:
  - resourceType: deployment
    resourceName: leeroy-web
    port: 8080
    localPort: 9000
```

tejal29:
Thanks @feloy for the PR, team was busy with Release 1.0.0. We will take a look at this soon now!

feloy:
The PR #3099  is marged, I close the issue. Thanks

PR #3099, opened 2019-10-19
Get current namespace as default namespace for port forwards

feloy:
Fixes `#3098`.

**Description**

When declaring a port forward without declaring the namespace, the **current** namespace is used, instead of the namespace named "default".

**Before**

If the namespace is not specified  for a port forward, the "default" namespace is used.

**After**

If the namespace is not specified  for a port forward, the current namespace is used.

**Submitter Checklist**

**Reviewer Notes**

- [ ] Unit test added.

**Release Notes**

When declaring a port forward without declaring the namespace, the current namespace is used, insted of the namespace named "default"

### DIFF
--- a/pkg/skaffold/schema/defaults/defaults_test.go
+++ b/pkg/skaffold/schema/defaults/defaults_test.go
@@ -17,11 +17,13 @@ limitations under the License.
 package defaults
 
 import (
+	"errors"
 	"testing"
 
 	"k8s.io/client-go/tools/clientcmd/api"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
+	kubectx "github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/context"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 	"github.com/GoogleContainerTools/skaffold/testutil"
 )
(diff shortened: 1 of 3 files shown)


pkg/skaffold/schema/defaults/defaults_test.go
```diff
@@ -254,7 +254,7 @@ func TestSetDefaultPortForwardNamespace(t *testing.T) {
 	err := Set(cfg)
 	testutil.CheckError(t, false, err)
 	testutil.CheckDeepEqual(t, "mynamespace", cfg.PortForward[0].Namespace)
-	testutil.CheckDeepEqual(t, constants.DefaultPortForwardNamespace, cfg.PortForward[1].Namespace)
+	testutil.CheckDeepEqual(t, "ns", cfg.PortForward[1].Namespace)
```
tejal29: umm, shdn't this be 
1. default namepsace when no currentNamespace is specified.
2. whatever is the current namespace when currentnamespace is specified.

Can you  override `currentNamespace` and mock it in test `**TestSetDefaultPortForwardNamespace**`?
Add 2 more test cases covering above 2 scenarios.

feloy: The SetupFakeKubernetesContext function has a behaviour I do not understand, as It seems not possible to change the current context by executing it a second time with a different context. Do I miss something?

feloy:
```
func getRawKubeConfig() (clientcmdapi.Config, error) {
	kubeConfigOnce.Do(func() {
```

In fact, it is not possible to update the kubeconfig file during the tests execution to override the currentContext, because the kubeconfig file is cached.

@tejal29 can you advise me how to handle this?

feloy: @tejal29 thanks, I've added the tests

PR #3120, opened 2019-10-23
Add heuristics to speed up Jib check in skaffold init

TadCordle:
With this PR, skaffold first checks for Jib-related strings in the build file before attempting to run the task/goal, rather than attempting to call into Jib any time a build.gradle or pom.xml is found. This eliminates a lot of false positives and speeds up `skaffold init` considerably on projects that don't have Jib configured.

Here are the timings for running `skaffold init --XXenableJibInit --analyze` on a 2-module maven project with no Jib configurations, before and after this PR:
* Before: **5.593s**
* After: **0.246s**

Timings before and after are about the same if Jib is configured on all modules. This PR seems to save around 2-3 seconds for every module Jib is not configured.

It still may be a good idea to consider automatically applying Jib on Java projects where it isn't already configured, but in the meantime this seems like a good improvement to have.

**Release Notes**

Speed up Jib checks in `skaffold init` on Java projects.

### DIFF
--- a/pkg/skaffold/jib/jib_init.go
+++ b/pkg/skaffold/jib/jib_init.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os/exec"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


TadCordle:
Assuming I tested everything correctly, my findings:

**Maven**
* Parent only: `jib:_skaffold-init` runs on the root project and lists all modules.
* Module only: Parent pom is skipped, and `jib:_skaffold-init` runs on the module directly.

**Gradle**
* Parent only: If jib isn't applied to the subprojects, `_jibSkaffoldInit` won't list them.
* Module only: Parent is skipped and `_jibSkaffoldInit` runs on the module directly.

chanseokoh:
Are your findings before or after this quick check? Doesn't this PR change any behavior? I think I better talk to you offline.

loosebazooka:
Does this work if jib is configured further down the tree in a multimodule project?

PR #3123, opened 2019-10-24
faster quicktests

balopat:
passes `$GOMAXPROCS` with the full number of processors to quickest. On my linux workstation this brings down from 140s -> 20 s `make quicktest` - should help with Mac / Win test speeds as well potentially a little bit on Travis. 

Why not test.sh? Maybe that could work too - it didn't seem to matter much (maybe because of `-cover` or `-race` wouldn't allow parallel runs?), so I didn't complicate things - we might need to look into that in more detail.


balopat:
OK, I'll run again the tests, maybe I made a mistake

PR #3138, opened 2019-10-29
skaffold credits

balopat:
<!-- 🎉🎉🎉 Thank you for the PR!!! 🎉🎉🎉 -->

**Description**

This will be a new command `skaffold credits`:

```
skaffold help
...

Other Commands:
  completion        Output shell completion for the given shell (bash or zsh)
  config            Interact with the Skaffold configuration
  credits           Export third party notices to given path (current dir by default)
  diagnose          Run a diagnostic on Skaffold
  version           Print the version information
```
```
skaffold credits --help                                                                                                                                                                                                                    4604ms  Thu 31 Oct 2019 01:20:30 PM PDT
Export third party notices to given path (./skaffold-credits by default)

Examples:
  # export third party licenses to ~/skaffold-credits
  skaffold credits -d ~/skaffold-credits

Options:
  -d, --dir='./skaffold-credits': destination directory to place third party licenses

Usage:
  skaffold credits [flags] [options]

Use "skaffold options" for a list of global command-line options (applies to all commands).
```

This is using two new tools: 
- licenses script from github.com/google/trillian to automatically list the licenses 
- rakyll/statik for embedding the notices file 

The flow includes now generating the sources for embedding before every binary creation command. 

**User facing changes**

**Before**

No `skaffold credits` command. 

**After**

See new command above. 

**Next PRs.**

none. 

**Submitter Checklist**

review them:

**Reviewer Notes**

- [ ] Unit test added.


**Release Notes**

```
`skaffold credits` - skaffold now exposes third party licenses recursively  that you can export in the form of a tgz file. 
```


cmd/skaffold/app/cmd/credits_release.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/rakyll/statik/fs"
+
+	//required for rakyll/statik embedded content
+	_ "github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/cmd/credits/statik"
+)
+
+func exportCredits(out io.Writer) error {
+	statikFS, err := fs.New()
+	if err != nil {
+		log.Fatalf("error opening embedded filesystem: %s", err)
```
tejal29: This would probably happen when you have not run a licenses.sh script right? Can we print some more helpful message like "maybe you are running from ./hack/licenses.sh"

cmd/skaffold/app/cmd/credits_release.go
```diff
@@ -0,0 +1,86 @@
...
+				return err
+			}
+		}
+		if !fileInfo.IsDir() {
+			file, err := statikFS.Open(filePath)
+			if err != nil {
+				log.Fatalf("error opening %s in embedded filesystem: %s", filePath, err)
+				return err
+			}
+			size := int(fileInfo.Size())
```
tejal29: Why not use ioutil.ReadFile() here?

balopat: I can use `ioutil.ReadAll()`, I'll change it!

cmd/skaffold/app/cmd/credits_release.go
balopat: actually in that case the code wouldn't compile - this happens in some other obscure case - IO can't think of a way the user can recover from it...

PR #3146, opened 2019-10-31
Skaffold image credits

balopat:
After #3138 is merged. 
This adds a /THIRD_PARTY_LICENSES to the `gcr.io/k8s-skaffold/skaffold` image.
Also vendors the files required by tools so that the build speed is not impacted.


balopat:
I merged this myself because it fixes the bug in the previous PR, also because it is high priority.

PR #3147, opened 2019-11-01
Fix splash

balopat:
Fixes buttons alignment, colors and footer padding.


balopat:
interestingly both the docs previewer and our integration test on kaniko failed with the same getting external ip issue: 
```
--- FAIL: TestBuildKanikoInsecureRegistry (313.35s)
    build_test.go:389: error getting external ip: timed out waiting for the condition
```
I'll restart again and see if it persisted since last night or not...

PR #3157, opened 2019-11-02
fix kaniko cache-dir

### LINKED ISSUE #3156: kaniko hostpath not correctly defined
### Expected behavior
kaniko argument "config-dir" is using the same value as the mount point

```yaml
containers:
  - args:
    ...
    - --cache-dir
    - /cache
    ...
    volumeMounts:
    - mountPath: /cache
      name: kaniko-cache
  ...
volumes:
  - hostPath:
      path: /data/docker/kaniko/cache
      type: ""
    name: kaniko-cache
```

### Actual behavior
"config-dir" is using the cache.hostpath value and not the mount point

```yaml
containers:
  - args:
    ...
    - --cache-dir
    - /data/docker/kaniko/cache
    ...
    volumeMounts:
    - mountPath: /cache
      name: kaniko-cache
  ...
volumes:
  - hostPath:
      path: /data/docker/kaniko/cache
      type: ""
    name: kaniko-cache
```

### Information

- Skaffold version: v0.41.0
- Operating system: fedora; kaniko runing into k3s
- Contents of skaffold.yaml:

```yaml
build:
      artifacts:
        - image: xxxx/yyyy
          context: .
          kaniko:
            dockerfile: build/package/Skaffold
            buildContext:
              localDir: {}
            cache:
              hostPath: /data/docker/kaniko/cache
```
This behaviour has been introduced by the commit 84b196baccac929bd2c758200582f11d2116abec

run "skaffold dev"

golgoth31:
Fixes #3156.

**Description**

Fix kaniko caching by using the right cache-dir

**User facing changes**

n/a

**Before**

INFO[0000] Downloading base image alpine:latest         
INFO[0002] Error while retrieving image from cache: getting file info: stat /data/docker/kaniko/cache/sha256:e4355b66995c96b4b468159fc5c7e3540fcef961189ca13fee877798649f531a: no such file or directory

**After**

INFO[0000] Downloading base image alpine:latest         
INFO[0002] Found sha256:e4355b66995c96b4b468159fc5c7e3540fcef961189ca13fee877798649f531a in local cache 
INFO[0002] Found manifest at /cache/sha256:e4355b66995c96b4b468159fc5c7e3540fcef961189ca13fee877798649f531a.json

**Next PRs.**

n/a

**Submitter Checklist**

review them:

**Reviewer Notes**

- [ ] Unit test added.

### DIFF
--- a/pkg/skaffold/build/cluster/kaniko.go
+++ b/pkg/skaffold/build/cluster/kaniko.go
@@ -28,6 +28,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/cluster/sources"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
@@ -133,7 +134,7 @@ func args(artifact *latest.KanikoArtifact, context, tag string, insecureRegistri
 			args = append(args, "--cache-repo", artifact.Cache.Repo)
 		}
 		if artifact.Cache.HostPath != "" {
-			args = append(args, "--cache-dir", artifact.Cache.HostPath)
+			args = append(args, "--cache-dir", constants.DefaultKanikoCacheDirMountPath)
 		}
 	}
 
--- a/pkg/skaffold/build/cluster/kaniko_test.go
+++ b/pkg/skaffold/build/cluster/kaniko_test.go
@@ -63,7 +63,7 @@ func TestArgs(t *testing.T) {
 			artifact: &latest.KanikoArtifact{
 				DockerfilePath: "Dockerfile",
 				Cache: &latest.KanikoCache{
-					HostPath: "/cache",
+					HostPath: "/host/cache",
 				},
 			},
 			expectedArgs: []string{"--cache=true", "--cache-dir", "/cache"},


pkg/skaffold/build/cluster/kaniko.go
```diff
@@ -133,7 +134,7 @@ func args(artifact *latest.KanikoArtifact, context, tag string, insecureRegistri
 			args = append(args, "--cache-repo", artifact.Cache.Repo)
 		}
 		if artifact.Cache.HostPath != "" {
-			args = append(args, "--cache-dir", artifact.Cache.HostPath)
+			args = append(args, "--cache-dir", constants.DefaultKanikoCacheDirMountPath)
```
balopat: I don't understand this change here - the logic is - if the user overrode the cache path - we should use it, and you are changing it to "if the user overrode the cache path, ignore it and keep it the default"

golgoth31: Hi,
I think there's a misunderstanding of this variable usage, IMHO, the "hostPath" variable should represent the path on the host where the kaniko pod is running. So the cache-dir kaniko's command represent the path inside the pod, this one doesn't really need to be modified.
As mentioned in the issue #3156 , with the current usage of the hostpath variable, kaniko is unable to find its cache; the current resulting kaniko's yaml file is as follow:

```yaml
containers:
  - args:
    ...
    - --cache-dir
    - /data/docker/kaniko/cache
    ...
    volumeMounts:
    - mountPath: /cache
      name: kaniko-cache
  ...
volumes:
  - hostPath:
      path: /data/docker/kaniko/cache
      type: ""
    name: kaniko-cache
```

As shown, the cache is mounted at "/cache" by kubernetes but search at "$hostPath" (/data/docker/kaniko/cache here) by kaniko. This is the reason of the patch. I have compiled skaffold with this patch and kaniko finds its cache perfectly.
Thank you for your time

priyawadhwa: That makes sense to me. Assuming the cache exists on the node, users can specify exactly where it is via `hostPath` and then we can automatically mount it to some default location (I would use `/kaniko/<some dir>` as the mount point in the container.

golgoth31: Hi,
The "/cache" is the default for kaniko and you defined the constant I use here to "/cache". So, I can also update the constant to be "/kaniko/cache" but I think it's less confusing to let at "/cache".

balopat: Gotcha, thanks for explaining - I also read the docs :), it makes sense now.

PR #3185, opened 2019-11-06
Fix master branch

dgageot:


dgageot:
Merged with admin privileges for the historical version change.

PR #3189, opened 2019-11-06
[Buildpacks] Support env variables

dgageot:
It's now possible to pass custom env variables to the buildpacks.
This is required to pass some configuration to the buildpacks.

### DIFF
--- a/docs/content/en/docs/pipeline-stages/builders/buildpacks.md
+++ b/docs/content/en/docs/pipeline-stages/builders/buildpacks.md
@@ -32,9 +32,10 @@ The following options can optionally be configured:
 
 **User defined environment variables**
 
-`env` makes it possible to pass arbitrary environment variables to the buildpacks.
-This is in general used to adjust the detection and the build phases. Many buildpacks
-use environment variables to select specific versions of language runtimes.
+`env` makes it possible to configure specific environment variables for buildpacks.
+Many buildpacks use environment variables to adjust their detection and the build phases,
+such as selecting specific versions of language runtimes.
+Note that user's current environment is not passed through to buildpacks.
 
 **Example**
 
--- a/docs/content/en/docs/pipeline-stages/builders/buildpacks.md
+++ b/docs/content/en/docs/pipeline-stages/builders/buildpacks.md
@@ -25,8 +25,16 @@ The following options can optionally be configured:
 
 {{< schema root="BuildpackArtifact" >}}
 
+**Builder**
+
 `builder` is *required* and tells Skaffold which
 [Builder](https://buildpacks.io/docs/app-developer-guide/build-an-app/) to use.
+
+**User defined environment variables**
+
+`env` makes it possible to pass arbitrary environment variables to the buildpacks.
+This is in general used to adjust the detection and the build phases. Many buildpacks
+use environment variables to select specific versions of language runtimes.
 
 **Example**
 
--- a/integration/examples/buildpacks/skaffold.yaml
+++ b/integration/examples/buildpacks/skaffold.yaml
@@ -1,4 +1,4 @@
-apiVersion: skaffold/v1
+apiVersion: skaffold/v2alpha1
 kind: Config
 build:
   artifacts:
--- a/integration/examples/buildpacks/skaffold.yaml
+++ b/integration/examples/buildpacks/skaffold.yaml
@@ -7,3 +7,4 @@ build:
       builder: "heroku/buildpacks"
       env:
       - GOVERSION=1.13.1
+      - GOPROXY={{.GOPROXY}}
--- a/integration/examples/buildpacks/skaffold.yaml
+++ b/integration/examples/buildpacks/skaffold.yaml
@@ -5,3 +5,5 @@ build:
   - image: gcr.io/k8s-skaffold/skaffold-buildpacks
     buildpack:
       builder: "heroku/buildpacks"
+      env:
+      - GOVERSION=1.13.1
--- a/pkg/skaffold/build/cache/hash.go
+++ b/pkg/skaffold/build/cache/hash.go
@@ -30,6 +30,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/misc"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 )
@@ -80,6 +81,15 @@ func getHashForArtifact(ctx context.Context, depLister DependencyLister, a *late
 		inputs = append(inputs, args...)
 	}
 
+	// add env variables for the artifact if specified
+	if env := retrieveEnv(a); len(env) > 0 {
+		evaluatedEnv, err := misc.EvaluateEnv(env)
+		if err != nil {
+			return "", errors.Wrap(err, "evaluating build args")
+		}
+		inputs = append(inputs, evaluatedEnv...)
+	}
+
 	// get a key for the hashes
 	hasher := sha256.New()
 	enc := json.NewEncoder(hasher)
@@ -109,6 +119,16 @@ func retrieveBuildArgs(artifact *latest.Artifact) map[string]*string {
 
 	case artifact.CustomArtifact != nil && artifact.CustomArtifact.Dependencies.Dockerfile != nil:
 		return artifact.CustomArtifact.Dependencies.Dockerfile.BuildArgs
+
+	default:
+		return nil
+	}
+}
+
+func retrieveEnv(artifact *latest.Artifact) []string {
+	switch {
+	case artifact.BuildpackArtifact != nil:
+		return artifact.BuildpackArtifact.Env
 
 	default:
 		return nil
--- a/pkg/skaffold/build/cache/hash.go
+++ b/pkg/skaffold/build/cache/hash.go
@@ -126,13 +126,10 @@ func retrieveBuildArgs(artifact *latest.Artifact) map[string]*string {
 }
 
 func retrieveEnv(artifact *latest.Artifact) []string {
-	switch {
-	case artifact.BuildpackArtifact != nil:
+	if artifact.BuildpackArtifact != nil {
 		return artifact.BuildpackArtifact.Env
-
-	default:
-		return nil
 	}
+	return nil
 }
 
 func convertBuildArgsToStringArray(buildArgs map[string]*string) []string {
--- a/pkg/skaffold/build/misc/env_test.go
+++ b/pkg/skaffold/build/misc/env_test.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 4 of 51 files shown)


dgageot:
Sure! I'm planning on adding some doc for the buildpacks support anyways

docs/content/en/docs/pipeline-stages/builders/buildpacks.md
```diff
@@ -25,9 +25,17 @@ The following options can optionally be configured:
...
+**Builder**
+
 `builder` is *required* and tells Skaffold which
 [Builder](https://buildpacks.io/docs/app-developer-guide/build-an-app/) to use.
 
+**User defined environment variables**
+
+`env` makes it possible to pass arbitrary environment variables to the buildpacks.
+This is in general used to adjust the detection and the build phases. For eg., most
+builders let you change runtime versions with environment variables.
```
briandealwis: I think it’s buildpacks, not builders, that use the variables to restrict runtimes versions.

Wdyt: _Many buildpacks use environment variables to select specific versions of language runtimes._

docs/content/en/docs/pipeline-stages/builders/custom.md
```diff
@@ -9,17 +9,19 @@ Custom build scripts allow skaffold users the flexibility to build artifacts wit
...
+| User-defined environment variables | Environment variables specified in a `key=value` form, in the `env` section of the `custom` configuration. | |
 | $IMAGE     | The fully qualified image name. For example, "gcr.io/image1:tag" | The custom build script is expected to build this image and tag it with the name provided in $IMAGE. The image should also be pushed if `$PUSH_IMAGE=true`. | 
 | $PUSH_IMAGE      | Set to true if the image in `$IMAGE` is expected to exist in a remote registry. Set to false if the image is expected to exist locally.      |   The custom build script will push the image `$IMAGE` if `$PUSH_IMAGE=true` | 
-| $BUILD_CONTEXT  | An absolute path to the directory this artifact is meant to be built from. Specified by artifact `context` in the skaffold.yaml.      | None. | 
-| Local environment variables | The current state of the local environment (e.g. `$HOST`, `$PATH)`. Determined by the golang [os.Environ](https://golang.org/pkg/os#Environ) function.| None. |
+| $BUILD_CONTEXT  | An absolute path to the directory this artifact is meant to be built from. Specified by artifact `context` in the skaffold.yaml.      | |
```
briandealwis: This change in order seems to indicate the precedence?  Might be worth explicitly calling that out?

pkg/skaffold/build/custom/custom.go
```diff
@@ -89,20 +89,27 @@ func (b *ArtifactBuilder) retrieveCmd(out io.Writer, a *latest.Artifact, tag str
...
-	envs := []string{
+	evaluatedEnv, err := misc.EvaluateEnv(a.Env)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to evaluate env variables")
+	}
+
+	var envs []string
+	envs = append(envs, util.OSEnviron()...)
+	envs = append(envs, b.additionalEnv...)
+	envs = append(envs, evaluatedEnv...)
```
briandealwis: Is there an implied order here?  I'd think that later values will override earlier values in the array, such that this means that values in the `skaffold.yaml` override those on the command-line?  Which seems unusual.

(I wonder if this function might not be better converting to a map and then back to a string array?)

pkg/skaffold/build/custom/custom.go
```diff
@@ -89,20 +89,27 @@ func (b *ArtifactBuilder) retrieveCmd(out io.Writer, a *latest.Artifact, tag str
 	return cmd, nil
 }
 
-func (b *ArtifactBuilder) retrieveEnv(a *latest.Artifact, tag string) ([]string, error) {
-	buildContext, err := buildContext(a.Workspace)
+func (b *ArtifactBuilder) retrieveEnv(workspace string, a *latest.CustomArtifact, tag string) ([]string, error) {
```
briandealwis: This name is at odds with `hash.go`'s `retrieveEnv` that literally just retrieves the appropriate env section.  This `retrieveEnv` doesn't seem quite right since there is generation happening here too.  `builderEnv()`?

pkg/skaffold/build/custom/custom_test.go
```diff
@@ -30,36 +30,61 @@ func TestRetrieveEnv(t *testing.T) {
...
+			tag:      "gcr.io/image/push:tag",
+			expected: []string{"KEY1=VALUE1", "IMAGE=gcr.io/image/push:tag", "IMAGES=gcr.io/image/push:tag", "PUSH_IMAGE=false", "BUILD_CONTEXT="},
+		},
+		{
+			description: "invalid custom env",
+			artifact: &latest.CustomArtifact{
+				Env: []string{"INVALID"},
+			},
+			tag:       "gcr.io/image/push:tag",
+			shouldErr: true,
```
briandealwis: I'd like to see a test with overrides, if that's possible.

pkg/skaffold/build/misc/env_test.go
```diff
@@ -0,0 +1,61 @@
...
+func TestEvaluateEnv(t *testing.T) {
+	tests := []struct {
+		description string
+		env         []string
+		want        []string
+		shouldErr   bool
+	}{
+		{
+			description: "env variables",
+			env:         []string{"key1=value1", "key2=", "key3={{.FOO}}"},
```
briandealwis: Should variables be able to refer to other variables in the same level?  For example, `key4={{.key3}}`?

docs/content/en/docs/pipeline-stages/builders/buildpacks.md
dgageot: Thanks. I'll take your text

docs/content/en/docs/pipeline-stages/builders/buildpacks.md
```diff
@@ -61,6 +69,10 @@ buildpack:
     - vendor/**
 ```
 
+### File Sync
+
+Syncable files must be included in both the `paths` section of `dependencies`, so that the skaffold file watcher knows to watch them, and the `sync` section, so that skaffold knows how to sync them.
```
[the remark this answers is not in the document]
dgageot: It really depends on the buildpack. Not a lot we can do to help I guess.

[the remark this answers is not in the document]
briandealwis: Buildpacks might be a good candidate for the "smart sync".  Maybe we could have our buildpack record sync info in the buildpack metadata.

docs/content/en/docs/pipeline-stages/builders/custom.md
dgageot: I've re-focused this PR on the buildpacks builder only

pkg/skaffold/build/custom/custom.go
dgageot: I've re-focused this PR on the buildpacks builder only

pkg/skaffold/build/custom/custom_test.go
dgageot: I've re-focused this PR on the buildpacks builder only

pkg/skaffold/build/misc/env_test.go
dgageot: Any example why we'd want to do that?

dgageot:
@balopat @briandealwis Thanks for your feedback. I've added the doc and removed the code related to the custom builder. I prefer to keep this PR focused on the buildpacks builder only.

docs/content/en/docs/pipeline-stages/builders/buildpacks.md
```diff
@@ -25,9 +25,17 @@ The following options can optionally be configured:
...
 {{< schema root="BuildpackArtifact" >}}
 
+**Builder**
+
 `builder` is *required* and tells Skaffold which
 [Builder](https://buildpacks.io/docs/app-developer-guide/build-an-app/) to use.
 
+**User defined environment variables**
+
+`env` makes it possible to pass arbitrary environment variables to the buildpacks.
```
briandealwis: _pass_ makes it sounds like we expose variables in the user's current environment. This is more constrained, right?
```suggestion
`env` makes it possible to configure specific environment variables for buildpacks.
Many buildpacks use environment variables to adjust their detection and the build phases,
such as selecting specific versions of language runtimes.
Note that user's current environment is not passed through to buildpacks.
```

pkg/skaffold/build/misc/env_test.go
briandealwis: I think go templates can be nested? https://golang.org/pkg/text/template/#hdr-Nested_template_definitions We should document any differences from normal usage of go templates, especially for the go-n00bs (like me!) who don't really know the limitations and how these uses would differ from other uses of go templates.

docs/content/en/docs/pipeline-stages/builders/buildpacks.md
```diff
@@ -25,9 +25,17 @@ The following options can optionally be configured:
...
+This is in general used to adjust the detection and the build phases. Many buildpacks
+use environment variables to select specific versions of language runtimes.
+
 **Example**
 
 The following `build` section, instructs Skaffold to build a
```
briandealwis: Could you also add an example of `env` — maybe with a go-template? — to the `samples/builders/buildpacks.yaml`?  (Is it possible to expose a value from the user's environment with this mechanism? `USER={{.USER}}`?)

(I likely don't know what I'm talking about here: I know only enough go templating to process `docker images —format`.)

docs/content/en/docs/pipeline-stages/builders/buildpacks.md
dgageot: I'll fo that

docs/content/en/docs/pipeline-stages/builders/buildpacks.md
dgageot: That's a good idea to add such value to the sample. But I don't see any value that would really be useful. Any idea?

dgageot: Sorry, with regards to nested templates, I still don't understand what your point is.

briandealwis: `NPM_CONFIG_LOGLEVEL`?

pkg/skaffold/build/cache/hash.go
```diff
@@ -115,6 +125,16 @@ func retrieveBuildArgs(artifact *latest.Artifact) map[string]*string {
...
 
+func retrieveEnv(artifact *latest.Artifact) []string {
+	switch {
+	case artifact.BuildpackArtifact != nil:
+		return artifact.BuildpackArtifact.Env
+
+	default:
+		return nil
+	}
+}
```
nkubala: why use a switch statement here?
```golang
if artifact.BuildpackArtifact != nil {
  return artifact.BuildpackArtifact.Env
}
return nil
```

docs/content/en/docs/pipeline-stages/builders/buildpacks.md
dgageot: Add GOPROXY

pkg/skaffold/build/cache/hash.go
dgageot: I'll fix that

pkg/skaffold/build/misc/env_test.go
dgageot: I'd rather keep it simple and not go into too much details about go templates.
I added an example to the `buildpacks` sample so that users understand the basic usage.

dgageot:
@balopat @nkubala @briandealwis It should be ready for a new review.

integration/examples/buildpacks/skaffold.yaml
```diff
@@ -7,3 +7,4 @@ build:
       builder: "heroku/buildpacks"
       env:
       - GOVERSION=1.13.1
+      - GOPROXY={{.GOPROXY}}
```
briandealwis: Cool!  I didn't realize it was that easy!

PR #3195, opened 2019-11-06
`debug` should replace existing ports or environment values

### LINKED ISSUE #3192: Cloud Code debug adds second definition of JAVA_TOOL_OPTIONS
(issue, including original description, migrated from https://github.com/GoogleCloudPlatform/cloud-code-intellij/issues/2669)

- Cloud Code for IntelliJ version: 19.0.1
- OS: macosx 

**What did you do?**

tried cloud code debug, noticed it added a second JAVA_TOOL_OPTIONS env variable.

We use JAVA_TOOL_OPTIONS to pass in memory settings and other JVM flags. Skaffold.yaml builds using jib-maven-plugin, and deploys plain yaml manifests for deployment and service. 

**What did you expect to see?**

jdwp agent config appended to existing JAVA_TOOL_OPTIONS

**What did you see instead?**

duplicate definition of JAVA_TOOL_OPTIONS env variable: 

snippet of output from `kubectl get deployment MYDEPLOYMENT -o yaml`

```
      - env:
        - name: JAVA_TOOL_OPTIONS
          value: -XX:TieredStopAtLevel=1 -XX:+UseSerialGC -Duser.timezone=America/New_York
            -Xms64m -Xmx128m
        - name: JAVA_TOOL_OPTIONS
          value: -agentlib:jdwp=transport=dt_socket,server=y,address=5005,suspend=n,quiet=y

```

briandealwis:
Fixes #3192 

**Description**

Certain language runtimes require defining or adding to specific environment variables to enable debugging in some form.  `debug` was not checking if there was an existing variable and could result in multiple entries.

This change:
  - provides some helpers for adding or replacing existing environment variables or port definitions
  - changes the existing debug transforms to use these helpers
  - removes some todos to retrieve debug ports from existing `containerPorts` — why bother if they're not actually hooked up


pkg/skaffold/debug/transform.go
```diff
@@ -311,3 +311,37 @@ func describe(obj runtime.Object) (group, version, kind, description string) {
 	}
 	return
 }
+
+// exposePort adds a `ContainerPort` instance or amends an existing entry with the same port.
+func exposePort(entries []v1.ContainerPort, portName string, port int32) []v1.ContainerPort {
+	for i := range entries {
+		// ports and names must be unique so rewrite an existing entry if found
+		if entries[i].ContainerPort == port || entries[i].Name == portName {
```
balopat: so this might be still ambiguous, depending on the order of the entries..., is that what we want? 

i.e. 
```
portName = "a", port= 80

entries1 = [ 
{a, 100}, ---> we are stopping here
{b,80}
]

entries1 = [ 
{b,80}, --->  we are stopping here
{a, 100}

]
```

briandealwis: Good catch!

briandealwis: Note that this shouldn't happen since we try to find unallocated ports.  But we may have other language runtimes where that isn't the case.

briandealwis:
PTAL.

pkg/skaffold/debug/transform.go
```diff
@@ -311,3 +311,48 @@ func describe(obj runtime.Object) (group, version, kind, description string) {
...
+
+// exposePort adds a `ContainerPort` instance or amends an existing entry with the same port.
+func exposePort(entries []v1.ContainerPort, portName string, port int32) []v1.ContainerPort {
+	found := false
+	for i := 0; i < len(entries); {
+		switch {
+		case entries[i].Name == portName:
+			// Ports and names must be unique so rewrite an existing entry if found
+			entries[i].Name = portName
+			entries[i].ContainerPort = port
```
balopat: can we add a log warning here when there is a mismatch between the port and the existing ContainerPort? It would be useful to see that in case it happens.

pkg/skaffold/debug/transform.go
```diff
@@ -311,3 +311,48 @@ func describe(obj runtime.Object) (group, version, kind, description string) {
...
+			found = true
+			i++
+		case entries[i].ContainerPort == port:
+			// Cut any entries with a clashing port
+			entries = append(entries[:i], entries[i+1:]...)
```
balopat: and the same here, putting a warning log that we are cutting a clashing exposed port with mismatching name. 
`"skaffold debug needs to expose port %d for %s. Removing clashing port definition %d (%s)", port, portName, entries[i].ContainerPort, entries[i].Name`

PR #3233, opened 2019-11-13
Simplify code to get GCS/GCB clients

dgageot:
+ Remove duplication
 + Remove indirections

### DIFF
--- a/pkg/skaffold/gcp/client.go
+++ b/pkg/skaffold/gcp/client.go
@@ -17,60 +17,22 @@ limitations under the License.
 package gcp
 
 import (
-	"context"
-	"sync"
+	"google.golang.org/api/option"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/version"
-
-	cstorage "cloud.google.com/go/storage"
-	"github.com/pkg/errors"
-	cloudbuild "google.golang.org/api/cloudbuild/v1"
-	"google.golang.org/api/option"
 )
 
-var (
-	cbclient *cloudbuild.Service
-	cbOnce   sync.Once
-	cbErr    error
-)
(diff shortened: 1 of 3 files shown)


pkg/skaffold/gcp/client.go
```diff
@@ -17,60 +17,22 @@ limitations under the License.
...
-var (
-	cbclient *cloudbuild.Service
-	cbOnce   sync.Once
-	cbErr    error
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/version"
 )
 
-// CloudBuildClient returns an authenticated client for interacting with
-// the Google Cloud Build API. This client is created once and cached
-// for repeated use in Skaffold.
```
briandealwis: So I think we'll now be creating a new client per build?

dgageot: yes. I didn't see the value of caching the client. can only create more problems.

pkg/skaffold/gcp/client.go
```diff
@@ -17,60 +17,22 @@ limitations under the License.
...
-func CloudBuildClient() (*cloudbuild.Service, error) {
-	cbOnce.Do(func() {
-		var options []option.ClientOption
-		var err error
-		creds, cErr := activeUserCredentials()
```
balopat: why are we removing this? This is an important piece of logic that we implemented to fix https://github.com/GoogleContainerTools/skaffold/issues/1966 - unless I'm missing something and the client libraries include this logic now?

dgageot: This is not removed, just moved to `ClientOptions()`

balopat: gotcha, I missed that

PR #3244, opened 2019-11-14
Fix schema

dgageot:
This is to fix the master branch


dgageot:
Force merging to fix master branch

PR #3245, opened 2019-11-14
Fix auto-project selection for GCB and Kaniko

dgageot:
The GCP project should be deduced from the image name being pushed, after the default repo is applied.

### DIFF
--- a/integration/build_test.go
+++ b/integration/build_test.go
@@ -390,8 +390,6 @@ func getExternalIP(t *testing.T, c *NSKubernetesClient, ns string) string {
 	return ip
 }
 
-// Make sure GCB project is deduced from artifact's image name,
-// AFTER the default repo is applied.
 func TestBuildGCBWithDefaultRepo(t *testing.T) {
 	if testing.Short() {
 		t.Skip("skipping integration test")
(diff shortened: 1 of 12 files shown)


integration/build_test.go
```diff
@@ -390,3 +389,29 @@ func getExternalIP(t *testing.T, c *NSKubernetesClient, ns string) string {
...
+// AFTER the default repo is applied.
+func TestBuildGCBWithDefaultRepo(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+	if !ShouldRunGCPOnlyTests() {
+		t.Skip("skipping test that is gcp only")
+	}
+
+	skaffold.Build("-d", "gcr.io/k8s-skaffold").InDir("testdata/gcb-default-repo").RunOrFail(t)
```
balopat: At least a comment here would be nice that the reason we do not assert anything in this test is that it is assumed that the only way this build can succeed is if the right project id is determined for GCB. 

Now, having said that I would be most comfortable with explicitly test for the behavior you are introducing instead of indirectly testing it with integration tests. 

Still better than no tests :D

dgageot: Thanks, I'll add the comments. I tried to add some unit tests but it was a pain. I'll see what we can do to make that easier in the future.

PR #2710, opened 2019-08-22
fix development guide link

tejal29:
Some on slack channel mentioned the link is broken

PR #3251, opened 2019-11-14
Add alias for old /docs/getting-started/

briandealwis:
Fix #2710 external documents that link to the old getting-started page.

As noted by @sageprice, there are external documents that link to https://skaffold.dev/docs/getting-started/#installing-skaffold and are breaking.  This PR uses a [Hugo alias](https://gohugo.io/content-management/urls/#aliases) to redirect to the new install page.

(Verified with `make preview-docs` on `localhost:1313/docs/getting-started/#installing-skaffold`

### DIFF
--- a/docs/content/en/docs/install/_index.md
+++ b/docs/content/en/docs/install/_index.md
@@ -2,6 +2,7 @@
 title: "Installing Skaffold"
 linkTitle: "Installing Skaffold"
 weight: 10
+aliases: [/docs/getting-started/]
 ---
 
 {{< alert title="Note" >}}


sageprice:
Thanks @briandealwis for this change. For reference, docs in question were Cloud Code docs for IntelliJ and VS Code: https://cloud.google.com/code/docs/vscode/install, https://cloud.google.com/code/docs/intellij/install. They're in the process of being updated as well, though good to do this in case there are any docs we don't own with the same issue.

briandealwis:
34.94.206.169:1313/docs/getting-started/#installing-skaffold

PR #3260, opened 2019-11-15
Upgrade Jib to 1.8.0

chanseokoh:
Jib 1.8.0 released and available.


chanseokoh:
A Gradle integration test is failing, as GCB runs an old Gradle.

As of now, the published [`gcr.io/cloud-builders/gradle` has the recent Gradle 5.6.2](https://github.com/GoogleCloudPlatform/cloud-builders/pull/452#issuecomment-550460760), but GCB is still out of sync to run the [cached Gradle 4.6 image](https://github.com/GoogleCloudPlatform/cloud-builders/pull/452#issuecomment-550485095). As @ImJasonH said in the comment, I was expecting GCB will be synced this week, but looks like it didn't happen. But eventually GCB should be synced to pick up 5.6.2, so I'll leave this PR open until then.

```
            Already have image (with digest): gcr.io/cloud-builders/gradle
...
            FAILURE: Build failed with an exception.
...
            * What went wrong:
            An exception occurred applying plugin request [id: 'com.google.cloud.tools.jib', version: '1.8.0']
            > Failed to apply plugin [id 'com.google.cloud.tools.jib']
               > Detected Gradle 4.6, but jib requires Gradle 4.9 or higher. You can upgrade by running 'gradle wrapper --gradle-version=4.9'.
```

chanseokoh:
Okay, so here's a problem. We released Jib 1.8 last week, which now requires Gradle >= 4.9, up from 4.6. (We think 4.9 is a low bar, as Gradle 6 is now out.)

Users have been complaining that GCB uses Gradle 4.6. It's even reported it breaks samples provided on GCP. But surprisingly, `gcr.io/cloud-builders/gradle:latest` actually runs Gradle 5.6.2. Turns out, it's GCB caching that makes GCB stuck with 4.6. They tried to purge the cache and make GCB synced recently, but they failed and rolled back the plan. And now due to the holiday season freeze, they said GCB will be stuck with 4.6 at least until next year. https://github.com/GoogleCloudPlatform/cloud-builders/pull/452#issuecomment-557202989

The workaround is to specify `gcr.io/cloud-builders/gradle:5.6.2-jdk-8` instead of `:latest`. (`:latest` currently points to `gcr.io/cloud-builders/gradle:5.6.2-jdk-8`; they are same.) But this means Skaffold will have to hardcode this tag for a while. So I'm thinking this may not be a good idea, since if GCB can refresh the image next January, it'd be just a couple-month wait only for the users of Jib 1.8+Skaffold+GCB+Gradle, which is pretty narrow. So to me, it seems acceptable to just wait for GCB to refresh next year.

dgageot:
@chanseokoh yeah, it sure can wait. If people want to use jib 1.8 with grade and GCB, they still can do it, they just have to change which grade image is used.

ImJasonH:
Can Skaffold use the `gradle` image on Dockerhub instead? These get updates more quickly and get cached more quickly in GCB. If not it would be useful to me to better understand why not.

dgageot:
@ImJasonH Would those images be pre-pulled? Because that's currently the only reason for us to use official builders

dgageot:
@chanseokoh can we close this PR and reopen it when all the pre-conditions are met?

chanseokoh:
Yeah, GCB is still using Gradle 4.6. Let me close this and reopen later.

PR #3278, opened 2019-11-21
Fix two issues with profiles

### LINKED ISSUE #3261: Setting specific profile with flag should override any activations (which are 'implicit')
### Expected behavior

We have some profiles with kubecontext activation in our skaffold configs but have other profiles ('testing' for example) that are unable to run even when running `skaffold run -p test` because the kubecontext activation seems to override that profile.

It is a common pattern for runtime flags to take precedence of more 'implicitly set' values like a config file or the default values. Skaffold seems to break this assumption/pattern

### Actual behavior

At least with kubeContext activations (haven't tested with others), the profile with that activation will be triggered as opposed to the profile explicitly set with `-p/--profile [PROFILE NAME]`. Also, documentation doesn't make it clear if two profiles would run at once if activation would trigger both of them (I am aware that one can explicitly run more than one profile, not sure if this works with activations as well).

Finally skaffold should probably print to stdout what profile is using as this would aid a lot in troubleshooting such issues.

### Information

- Skaffold version: v0.36.0
- Operating system: both observed in macOS 10.14.6 and Debian Stretch docker container

I cannot post specific kubeContext values here but a skaffold.yaml with two profiles should reproduce this easily:
* One profile ('profile-a') can deploy service a and have a `kubeContext` activation
* The other one ('profile-b') can deploy service b and have no activation
* Running `skaffold deploy -p profile-b` with the Kubernetes context of profile a as the current context will deploy service a and NOT service b

### LINKED ISSUE #3277: Skaffold generates strange warnings when running with many profiles
### Expected behavior
I expect Skaffold to run in `dev`-mode without errors and warnings, when configured correctly
### Actual behavior
We use Skaffold to develop locally a big platform, which consists from many services.
We use Helm as our deployer.
We have separate profiles for different services in `skaffold.yaml`, therefore our  skaffold command looks like following:

```
skaffold dev -p=profile-one,profile-two,profile-three,profile-n --cache-artifacts=true --auto-build=false --auto-deploy=false --rpc-http-port=50070
```

However, after Helm completes deploy, skaffold starts to generate errors like following:
```
WARN[0096] error adding label to runtime object: patching resource default/some-secret: Secret "some-secret" is invalid: metadata.labels: Invalid value: "profile-one__profile-two__profile-three__profile-n": must be no more than 63 characters 
```

This warnings are generated for **every** single resource we have: configmaps, secrets, roles, role bindings and so on, and the label value is the names of our skaffold profiles, concatenated with `__` (to underscores), in order, in which they appear in our `skaffold.yaml`. 

### Information

- Skaffold version: 1.0.1
- Operating system: MacOS Catalina
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta16
kind: Config
metadata:
  name: some-name
profiles:
- name: profile-one
  build:
    local:
      push: false
    artifacts:
    - image: some/image
      context: some-submodule
      docker:
        dockerfile: Dockerfile
    - image: some-other/image
      context: some-other-submodule
      docker:
        dockerfile: Dockerfile
  deploy:
    helm:
      releases:
      - name: release-name
        chartPath: charts/our-chart
        valuesFiles:
        - charts/our-chart/values/value-file-one.yaml
        - charts/our-chart/values/value-file-two.yaml

# Supplemental profiles
- name: profile-two
  patches:
  - op: add
    path: /deploy/helm/releases/0/valuesFiles/-
    value: charts/our-chart/values/yet-one-file.yaml
- name: profile-three
  patches:
  - op: add
    path: /deploy/helm/releases/0/valuesFiles/-
    value: charts/our-chart/values/yet-another-file.yaml
```

Skaffold command: `skaffold dev -p=profile-one,profile-two,profile-three,profile-n --cache-artifacts=true --auto-build=false --auto-deploy=false --rpc-http-port=50070`

_Actual number of profiles, specified in command, is bigger than three, of course._

dgageot:
Fix #3277
Fix #3261

### DIFF
--- a/pkg/skaffold/schema/profiles_test.go
+++ b/pkg/skaffold/schema/profiles_test.go
@@ -611,6 +611,17 @@ func TestActivatedProfiles(t *testing.T) {
 			},
 			expected: []string{"one", "one-as-well", "not-two"},
 		},
+		{
+			description: "Profiles on the command line are activated after auto-activatd profiles",
+			opts: cfg.SkaffoldOptions{
+				Command:  "run",
+				Profiles: []string{"activated", "also-activated"},
+			},
+			profiles: []latest.Profile{
+				{Name: "run-profile", Activation: []latest.Activation{{Command: "run"}}},
+			},
+			expected: []string{"run-profile", "activated", "also-activated"},
+		},
 	}
 
 	for _, test := range tests {
--- a/pkg/skaffold/config/options.go
+++ b/pkg/skaffold/config/options.go
@@ -17,6 +17,7 @@ limitations under the License.
 package config
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
@@ -81,8 +82,9 @@ func (opts *SkaffoldOptions) Labels() map[string]string {
 	if opts.Namespace != "" {
 		labels["skaffold.dev/namespace"] = opts.Namespace
 	}
-	if len(opts.Profiles) > 0 {
-		labels["skaffold.dev/profiles"] = strings.Join(opts.Profiles, "__")
+	for i, profile := range opts.Profiles {
+		key := fmt.Sprintf("skaffold.dev/profile.%d", i)
+		labels[key] = profile
 	}
 	for _, cl := range opts.CustomLabels {
 		l := strings.SplitN(cl, "=", 2)
--- a/pkg/skaffold/config/options_test.go
+++ b/pkg/skaffold/config/options_test.go
@@ -52,14 +52,15 @@ func TestLabels(t *testing.T) {
 			description: "profile",
 			options:     SkaffoldOptions{Profiles: []string{"profile"}},
 			expectedLabels: map[string]string{
-				"skaffold.dev/profiles": "profile",
+				"skaffold.dev/profile.0": "profile",
 			},
 		},
 		{
 			description: "profiles",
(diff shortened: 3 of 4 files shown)


petr-buchin:
@dgageot Thank you very much for a fix of my issue!
However, what do you think - maybe it's better to create some hash of profiles string, instead of trimming it to 62 characters and adding z ?

Hash from the whole profiles string will actually change with the profiles string, but trimmed string will not, if changes were made after "trim point".

dgageot:
Well, you're right, I was a bit "lazy" on this one ;-)
I'm not sure I like the solution with a digest, it's not something users can really decipher.
 - We could add one label per profile (profile.1=..., profile2=...)
 - Or we could remove that altogether. I'm not sure it brings a lot of value.

@balopat wdyt?

dgageot:
OK, I decided to replace my ugly hack and use a key per profile instead.
(It doesn't run any check on profile names, though)

pkg/skaffold/schema/profiles_test.go
```diff
@@ -611,6 +611,17 @@ func TestActivatedProfiles(t *testing.T) {
 			},
 			expected: []string{"one", "one-as-well", "not-two"},
 		},
+		{
+			description: "Profiles on the command lines are activated after auto-activatd profiles",
```
balopat:
```suggestion
			description: "Profiles on the command lines are activated after auto-activated profiles",
```

PR #3282, opened 2019-11-22
Default to gcloud auth

dgageot:
Make sure that `gcloud` doesn't need to be configured as a helper to be used
to push to/pull from `*.gcr.io`

This used to be the case but using `GetAllAuthConfigs()` over `GetAuthConfig(registry string)` broke this behaviour.

Also use the gcloud auth code from `go-containerregistry`

### DIFF
--- a/pkg/skaffold/docker/auth.go
+++ b/pkg/skaffold/docker/auth.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 
 	"github.com/docker/cli/cli/config"
+	"github.com/docker/cli/cli/config/configfile"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/pkg/homedir"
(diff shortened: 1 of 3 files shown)


balopat:
This is failing with gcloud auth errors on integration tests. PTAL @dgageot

dgageot:
@balopat I cut the PR in two pieces. The other part needs some more work. This part passes the tests

pkg/skaffold/docker/auth.go
```diff
@@ -61,13 +62,22 @@ type AuthConfigHelper interface {
...
-	gcp.AutoConfigureGCRCredentialHelper(cf, registry)
+	gcp.AutoConfigureGCRCredentialHelper(cf)
+
+	return cf, nil
+}
+
+func (credsHelper) GetAuthConfig(registry string) (types.AuthConfig, error) {
+	cf, err := loadDockerConfig()
+	if err != nil {
+		return types.AuthConfig{}, err
```
briandealwis: This means a broken config.json will cause skaffold to abort.  Maybe we should log an error, but continue on?

dgageot: 1/ docker-credential-gcloud should be favored. See https://github.com/GoogleCloudPlatform/docker-credential-gcr "For normal development setups, users are encouraged to use gcloud auth configure-docker, instead"
2/ In the docker build code, we ignore the error, like the Docker CLI does it.
3/ We can't just configure it for `gcr.io` when `GetAllAuthConfigs()` is used. When `GetAuthConfig` is used, configuring the helper for *.gcr.io has the exact same effect

PR #3291, opened 2019-11-25
fix table on api page

balopat:
https://skaffold.dev/docs/design/api/#control-api is not rendering correctly: 

image


balopat:
note, this is still not a fix, I'll fix it after lunch :)

balopat:
http://35.236.6.114:1313/docs/design/api/#control-api now it looks good

PR #3299, opened 2019-11-27
Improve error output when kompose fails

dgageot:
+ Improve the output of `util.RunCmdOut()` in case of error.
 + Use that improved output when kompose fails

Before:
```
$ skaffold init --compose-file docker-compose.yaml
FATA[0000] running kompose: exit status 1
```

After:
```
$ skaffold init --compose-file docker-compose.yaml
FATA[0000] Running [kompose convert -f docker-compose.yaml]
 - stdout:
 - stderr: FATA ima_ge Additional property ima_ge is not allowed
: exit status 1
```

### DIFF
--- a/pkg/skaffold/initializer/init.go
+++ b/pkg/skaffold/initializer/init.go
@@ -117,11 +117,8 @@ func DoInit(ctx context.Context, out io.Writer, c Config) error {
 	rootDir := "."
 
 	if c.ComposeFile != "" {
-		// run kompose first to generate k8s manifests, then run skaffold init
-		logrus.Infof("running 'kompose convert' for file %s", c.ComposeFile)
-		komposeCmd := exec.CommandContext(ctx, "kompose", "convert", "-f", c.ComposeFile)
-		if err := util.RunCmd(komposeCmd); err != nil {
-			return errors.Wrap(err, "running kompose")
+		if err := runKompose(ctx, c.ComposeFile); err != nil {
+			return err
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/skaffold/util/cmd.go
```diff
@@ -74,7 +74,7 @@ func (*Commander) RunCmdOut(cmd *exec.Cmd) ([]byte, error) {
 
 	err = cmd.Wait()
 	if err != nil {
-		return stdout, errors.Wrapf(err, "Running %s: stdout %s, stderr: %s, err: %v", cmd.Args, stdout, stderr, err)
+		return stdout, errors.Wrapf(err, "Running %s\n - stdout: %s\n - stderr: %s", cmd.Args, stdout, stderr)
```
briandealwis: Embedding newlines doesn't seem great.  It feels like this should be an object.

pkg/skaffold/initializer/init.go
```diff
@@ -228,6 +225,18 @@ func DoInit(ctx context.Context, out io.Writer, c Config) error {
 	return nil
 }
 
+// runKompose runs the `kompose` CLI before running skaffold init
+func runKompose(ctx context.Context, composeFile string) error {
+	if _, err := os.Stat(composeFile); os.IsNotExist(err) {
+		return err
```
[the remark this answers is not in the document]
dgageot: It'll show the filename

pkg/skaffold/util/cmd.go
dgageot: We could do that. And it would have an Error() function that embeds newlines...
I'd rather keep this simple form

pkg/skaffold/initializer/init.go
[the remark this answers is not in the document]
dgageot: It'll show the filename as is

pkg/skaffold/util/cmd.go
dgageot: And that object would have an Error() function with that string... I'd rather keep this simple form

PR #3300, opened 2019-11-27
Fix typo in PR template

dgageot:


dgageot:
@briandealwis I used `behavior` because it was written this way on the next line.

PR #3308, opened 2019-11-27
Use test helper

dgageot:


dgageot:
I'd love to find a way to enforce that

PR #3326, opened 2019-11-29
Cause debug to skip unmarshallable yamls

### LINKED ISSUE #3281: skaffold debug fails to recognize CRD
### Expected behavior
`skaffold debug` recognizes my openfaas CRD and deploys it (even if it's not debuggable)
### Actual behavior
`skaffold debug` fails because a CRD is being used in one of the manifests under `deploy.kubectl.manifests`:
```
skaffold debug -v=debug
INFO[0000] Skaffold &{Version:v1.0.0-98-g4c2ba6fa ConfigVersion:skaffold/v2alpha1 GitVersion: GitCommit:4c2ba6fa2e0912f8b1302f1f736dcf702ad795c5 GitTreeState:clean BuildDate:2019-11-21T20:25:53Z GoVersion:go1.12.7 Compiler:gc Platform:linux/amd64}
DEBU[0000] config version (skaffold/v1) out of date: upgrading to latest (skaffold/v2alpha1)
DEBU[0000] Defaulting build type to local build
DEBU[0000] validating yamltags of struct SkaffoldConfig
DEBU[0000] validating yamltags of struct Metadata
DEBU[0000] validating yamltags of struct Pipeline
DEBU[0000] validating yamltags of struct BuildConfig
DEBU[0000] validating yamltags of struct TagPolicy
DEBU[0000] validating yamltags of struct GitTagger
DEBU[0000] validating yamltags of struct BuildType
DEBU[0000] validating yamltags of struct LocalBuild
DEBU[0000] validating yamltags of struct DeployConfig
DEBU[0000] validating yamltags of struct DeployType
DEBU[0000] validating yamltags of struct KubectlDeploy
DEBU[0000] validating yamltags of struct KubectlFlags
INFO[0000] Using kubectl context: minikube
DEBU[0000] Using builder: local
DEBU[0000] Running command: [minikube docker-env --shell none]
DEBU[0000] Command output: [DOCKER_TLS_VERIFY=1
DOCKER_HOST=tcp://192.168.39.52:2376
DOCKER_CERT_PATH=/home/myuser/.minikube/certs
]
DEBU[0000] setting Docker user agent to skaffold-v1.0.0-98-g4c2ba6fa
DEBU[0000] push value not present, defaulting to false because localCluster is true
Listing files to watch...
INFO[0000] List generated in 38.544µs
Generating tags...
INFO[0000] Tags generated in 8.607µs
Checking cache...
INFO[0000] Cache check complete in 17.34µs
Tags used in deployment:
Starting deploy...
DEBU[0000] Running command: [kubectl version --client -ojson]
DEBU[0000] Command output: [{
  "clientVersion": {
    "major": "1",
    "minor": "15",
    "gitVersion": "v1.15.2",
    "gitCommit": "f6278300bebbb750328ac16ee6dd3aa7d3549568",
    "gitTreeState": "clean",
    "buildDate": "2019-08-05T09:23:26Z",
    "goVersion": "go1.12.5",
    "compiler": "gc",
    "platform": "linux/amd64"
  }
}
]
DEBU[0000] Running command: [kubectl --context minikube create --dry-run -oyaml -f /home/myuser/skaffoldtest/function.yaml]
DEBU[0000] Command output: [apiVersion: openfaas.com/v1alpha2
kind: Function
metadata:
  name: myfunction
  namespace: openfaas-fn
spec:
  image: myfunction
  name: myfunction
]
DEBU[0000] manifests with tagged images: apiVersion: openfaas.com/v1alpha2
kind: Function
metadata:
  name: myfunction
  namespace: openfaas-fn
spec:
  image: myfunction
  name: myfunction
Cleaning up...
DEBU[0000] Running command: [kubectl --context minikube create --dry-run -oyaml -f /home/myuser/skaffoldtest/function.yaml]
DEBU[0001] Command output: [apiVersion: openfaas.com/v1alpha2
kind: Function
metadata:
  name: myfunction
  namespace: openfaas-fn
spec:
  image: myfunction
  name: myfunction
]
DEBU[0001] Running command: [kubectl --context minikube delete --ignore-not-found=true -f -]
INFO[0001] Cleanup complete in 225.070878ms
FATA[0001] exiting dev mode because first deploy failed: unable to transform manifests: reading Kubernetes YAML: no kind "Function" is registered for version "openfaas.com/v1alpha2" in scheme "vendor/k8s.io/client-go/kubernetes/scheme/register.go:65"
```

### Information

- Skaffold version: v1.0.0-98-g4c2ba6fa
- Operating system: Ubuntu 18.04.3 LTS
- Contents of skaffold.yaml:

```yaml
---
apiVersion: skaffold/v1
kind: Config
deploy:
  kubectl:
    manifests:
      - ./function.yaml

```

- Contents of function.yaml:

```yaml
---
apiVersion: openfaas.com/v1alpha2
kind: Function
metadata:
  name: myfunction
  namespace: openfaas-fn
spec:
  name: myfunction
  image: myfunction

```

1. add the yamls from above
2. `skaffold debug`

briandealwis:
Fixes #3281

**Description**

Change `debug` to skip processing any manifests that cannot be unmarshalled.

**User facing changes**

A debug-level message is now output (if level = debug):
```
time="2019-11-29T14:48:53-05:00" level=debug msg="Unable to load Kubernetes YAML: no kind \"Function\" is registered for version \"openfaas.com/v1alpha2\" in scheme \"pkg/runtime/scheme.go:101\"\n"
```

**Before**

`skaffold debug` would exit on an unhandled manifest

**After**

`skaffold debug` skips unhandled manifests.

**Next PRs.**

n/a

**Submitter Checklist**

review them:

- [ ] ~Adds documentation as needed: user docs, YAML reference, CLI reference.~

**Reviewer Notes**

- [ ] Unit test added.

**Release Notes**

n/a

### DIFF
--- a/pkg/skaffold/debug/debug.go
+++ b/pkg/skaffold/debug/debug.go
@@ -67,7 +67,7 @@ func applyDebuggingTransforms(l kubectl.ManifestList, retriever configurationRet
 	for _, manifest := range l {
 		obj, _, err := decodeFromYaml(manifest, nil, nil)
 		if err != nil {
-			logrus.Debugf("Unable to load Kubernetes YAML: %v\n", err)
+			logrus.Debugf("Unable to interpret manifest for debugging: %v\n", err)
 		} else if transformManifest(obj, retriever) {
 			manifest, err = encodeAsYaml(obj)
 			if err != nil {
--- a/pkg/skaffold/debug/debug.go
+++ b/pkg/skaffold/debug/debug.go
@@ -67,10 +67,8 @@ func applyDebuggingTransforms(l kubectl.ManifestList, retriever configurationRet
 	for _, manifest := range l {
 		obj, _, err := decodeFromYaml(manifest, nil, nil)
 		if err != nil {
-			return nil, errors.Wrap(err, "reading Kubernetes YAML")
-		}
-
-		if transformManifest(obj, retriever) {
+			logrus.Debugf("Unable to load Kubernetes YAML: %v\n", err)
+		} else if transformManifest(obj, retriever) {
 			manifest, err = encodeAsYaml(obj)
 			if err != nil {
 				return nil, errors.Wrap(err, "marshalling yaml")
--- a/pkg/skaffold/debug/debug_test.go
+++ b/pkg/skaffold/debug/debug_test.go
@@ -485,7 +485,7 @@ status:
 		},
 		{
 			description: "skip unhandled yamls like crds",
-			shouldErr: false,
+			shouldErr:   false,
 			in: `---
 apiVersion: openfaas.com/v1alpha2
 kind: Function
--- a/pkg/skaffold/debug/debug_test.go
+++ b/pkg/skaffold/debug/debug_test.go
@@ -483,6 +483,28 @@ spec:
 status:
   replicas: 0`,
 		},
+		{
+			description: "skip unhandled yamls like crds",
+			shouldErr: false,
+			in: `---
+apiVersion: openfaas.com/v1alpha2
+kind: Function
+metadata:
+  name: myfunction
+  namespace: openfaas-fn
+spec:
+  name: myfunction
+  image: myfunction`,
+			out: `---
+apiVersion: openfaas.com/v1alpha2
+kind: Function
+metadata:
+  name: myfunction
+  namespace: openfaas-fn
+spec:
+  name: myfunction
+  image: myfunction`,
+		},
 	}
 	for _, test := range tests {
 		testutil.Run(t, test.description, func(t *testutil.T) {


pkg/skaffold/debug/debug.go
```diff
@@ -67,10 +67,8 @@ func applyDebuggingTransforms(l kubectl.ManifestList, retriever configurationRet
 	for _, manifest := range l {
 		obj, _, err := decodeFromYaml(manifest, nil, nil)
 		if err != nil {
-			return nil, errors.Wrap(err, "reading Kubernetes YAML")
-		}
-
-		if transformManifest(obj, retriever) {
+			logrus.Debugf("Unable to load Kubernetes YAML: %v\n", err)
```
balopat:
```suggestion
			logrus.Debugf("Unable to interpret manifest for debugging: %v\n", err)
```

pkg/skaffold/debug/debug.go
```diff
@@ -67,10 +67,8 @@ func applyDebuggingTransforms(l kubectl.ManifestList, retriever configurationRet
 	for _, manifest := range l {
 		obj, _, err := decodeFromYaml(manifest, nil, nil)
 		if err != nil {
```
balopat: we need to be more specific to match the `no kind XYZ is registered for version ABC in scheme ` type of errors

briandealwis: I'll respectfully disagree: any such validation should happen elsewhere.  `debug`'s current behaviour is to tweak the workload resources that it knows about, and skip any other resources.  We should extend that skipping behaviour to any resources that we (or our libraries) are unable to interpret too.

bmcstdio:
@balopat @briandealwis won't this cause any manifests that cannot be decoded to be skipped _entirely_ (i.e., to not actually be deployed)? IIUC that is the case, and I'd argue that's not as desirable as continuing to deploy such manifests but without applying any debug transformations—which is what #3326 does.

pkg/skaffold/debug/debug.go
balopat: I think what I meant is that there should be 3 branches: 

1) `err ~= no kind XYZ is registered for version ABC in scheme` -> continue 
2) `err != nil` but it's something else than unregistered schema -> return error 
3) ` err == nil` transform. 

WDYT?

balopat: Hm...on second thought - my thinking was that  problem is that we don't know what the error is - if it is a valid yaml but with unknown kind - great let's skip it. if it's an invalid yaml though, I don't want to skip it. -- but you are right, if we want to do upfront YAML validation, it could/should be elsewhere, not necessarily in debug.

balopat:
> @balopat @briandealwis won't this cause any manifests that cannot be decoded to be skipped _entirely_ (i.e., to not actually be deployed)? IIUC that is the case, 

It is not the case, the https://github.com/GoogleContainerTools/skaffold/pull/3326/files#diff-370debde40bf1d20477fd2e0467bbdf6R80 line ensures that untransformed objects are added as well. 

> and I'd argue that's not as desirable as continuing to deploy such manifests but without applying any debug transformations—which is what #3326 does.

The difference between your PR (#3327) and this one is that yours is implementing  the three branches mentioned here: https://github.com/GoogleContainerTools/skaffold/pull/3326#discussion_r359364430 However I do agree with @briandealwis that this validation is not the responsibility of the debug transformer. We should think about the debug transformer as `DebugTransformer(manifests) = transformedManifests + non-transformable manifests`. The error handling is now deferred to a later stage, when the actual deploy happens.

briandealwis:
Odd, macOS tests failed in `TestGracefulBuildCancel`

PR #3341, opened 2019-12-03
Cache expensive Go compilation and linting

dgageot:
This will significantly reduce "unit tests" and "checks" jobs on TravisCI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,16 +15,28 @@ jobs:
   include:
     - os: linux
       name: "checks"
+      env:
+        - STATICCHECK_CACHE=$HOME/linting
       script:
         - make checks
+      cache:
+        directories:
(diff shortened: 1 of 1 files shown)


tejal29:
I dont' like the idea of using cache for unit test running. 
In my past team, we have seen some cases for [`pants test`](https://github.com/pantsbuild/pants)       where some ci runs were false positive due to them being not invalidated. 

But again, our team wrote pants cache, and developers write bugs :) 
How is your confidence in `go tests` cache ?

I am ok with trying this and then reverting in case we see false positives :)

dgageot:
@tejal29 On linux, the test cache is disabled (with --count=1). So I thought the risk would only be that a mac/win test could be flaky and we wouldn't detect that.

PR #3366, opened 2019-12-11
Open in cloud code

dgageot:
Add "Open in Cloud Code" button to every example.


dgageot:
@nkubala This feature is not yet available with IntelliJ

PR #3369, opened 2019-12-12
Jib Auto Sync - core + maven implementation

loosebazooka:
Just putting this up here for now, if it feels like I'm messing too much with the core system maybe we can restructure that. I'll clean up the code some more and add tests if we can kinda agree on the changes in here?

1. So most of this is baked into the jib package, but I don't think anything in it necessarily HAS to be jib specific. If we want to extend the model out for other systems we can. But it mostly
feels like a java problem anyway?

2. Added an InitSync before first build in dev mode to initalize sync of the files - requires a build to get the sync data

3. I feel like there's some weird cases I haven't handled yet

4. logging is wrong, probably need to fix that

### DIFF
--- a/pkg/skaffold/build/jib/jib_sync.go
+++ b/pkg/skaffold/build/jib/jib_sync.go
@@ -1,0 +1,213 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package jib
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"time"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/filemon"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+	"github.com/pkg/errors"
+)
+
+var syncLists = make(map[string]SyncMap)
+
+type SyncMap struct {
+	Direct    []SyncEntry `json:"direct"`
+	Generated []SyncEntry `json:"generated"`
+}
+
+type SyncEntry struct {
+	Src  string `json:"src"`
+	Dest string `json:"dest"`
+
+	filetime time.Time
+}
+
+func InitSync(ctx context.Context, workspace string, artifact *latest.JibArtifact) error {
+	syncMap, err := getSyncMap(ctx, workspace, artifact)
+	if (err != nil) {
+		return err
+	}
+	syncLists[artifact.Project] = *syncMap
+	return nil
+}
+
+// returns toCopy, toDelete, error
+func GetSyncDiff(ctx context.Context, workspace string, artifact *latest.JibArtifact, e filemon.Events) (map[string][]string, map[string][]string, error) {
+
+	// if anything that was modified was a buildfile, do NOT sync, do a rebuild
+	buildFiles := GetBuildDefinitions(artifact)
+	for _, f := range e.Modified {
+		if !filepath.IsAbs(f) {
+			if ff, err := filepath.Abs(f); err != nil{
+				return nil, nil, err
+			} else {
+				f = ff
+			}
+		}
+		for _, bf := range buildFiles {
+			if f == bf {
+				return nil, nil, nil
+			}
+		}
+	}
+
+	// it seems less than efficient to keep the original JSON structure when doing look ups, so maybe we should only use the json objects for serialization
+	// and store the sync data in a better type?
+	oldSyncMap := syncLists[artifact.Project]
+
+
+	// if all files are modified and direct, we don't need to build anything
+	if len(e.Deleted) == 0 || len(e.Added) == 0 {
+		matches := make(map[string][]string)
+		for _, f := range e.Modified {
+			for _, se := range oldSyncMap.Direct {
+				// filemon.Events doesn't seem to make any guarantee about the paths,
+				// so convert them to absolute paths (that's what jib provides)
+				if !filepath.IsAbs(f) {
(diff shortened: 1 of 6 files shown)


pkg/skaffold/build/jib/maven.go
```diff
@@ -85,8 +85,24 @@ func getCommandMaven(ctx context.Context, workspace string, a *latest.JibArtifac
 	return MavenCommand.CreateCommand(ctx, workspace, args)
 }
 
+func getSyncMapCommandMaven(ctx context.Context, workspace string, a *latest.JibArtifact) *exec.Cmd {
+	cmd := MavenCommand.CreateCommand(ctx, workspace, mavenBuildArgs("_skaffold-sync-map", a, true))
```
chanseokoh: `mavenBuildArgs` append either `package` or `prepare-package`. Does that make sense? And I don't really think `_skaffold-sync-map` will work. It should be `jib:_skaffold-sync-map`?

Also we pass `--quiet` for `_skaffold-files_v2`. But I'm not sure if `--quiet` helps, because I didn't see any Maven output in your short demo.

pkg/skaffold/build/jib/jib_sync.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func runAndParseSyncMap(cmd *exec.Cmd, sm *SyncMap) error {
+	stdout, err := util.RunCmdOut(cmd)
+	if err != nil {
+		return errors.Wrap(err, "failed to get Jib sync map")
+	}
+
+	// To parse the output, search for "BEGIN JIB JSON", then unmarshal the next line into the pathMap struct.
+	matches := regexp.MustCompile(`BEGIN JIB JSON\r?\n({.*})`).FindSubmatch(stdout)
```
chanseokoh: I liked @briandealwis's [idea](https://github.com/GoogleContainerTools/jib/pull/2192#issuecomment-563303984) to have this extensible. Something we should keep in mind to make things match.

pkg/skaffold/build/jib/jib_sync.go
```diff
@@ -0,0 +1,213 @@
...
+
+	filetime time.Time
+}
+
+func InitSync(ctx context.Context, workspace string, artifact *latest.JibArtifact) error {
+	syncMap, err := getSyncMap(ctx, workspace, artifact)
+	if (err != nil) {
+		return err
+	}
+	syncLists[artifact.Project] = *syncMap
```
chanseokoh: Is it possible that multiple Java projects have the same "JibArtifact.Project" name as a key to the map? The map is a global map, isn't it?

And isn't `artifact.Project == ""` for a single-module project setup?

pkg/skaffold/build/jib/jib_sync.go
```diff
@@ -0,0 +1,213 @@
...
+type SyncMap struct {
+	Direct    []SyncEntry `json:"direct"`
+	Generated []SyncEntry `json:"generated"`
+}
+
+type SyncEntry struct {
+	Src  string `json:"src"`
+	Dest string `json:"dest"`
+
+	filetime time.Time
```
chanseokoh: Does an entry always refer to a file and not a directory? Just to understand what the modification time of a directory may mean, if it can be a directory. But I kind of remember that our sync map will list every single file and not directories?

And I always forget. For a Java file, will the entry be a `.java` file or a `.class` file?

pkg/skaffold/build/jib/jib_sync.go
```diff
@@ -0,0 +1,213 @@
...
+		if !filepath.IsAbs(f) {
+			if ff, err := filepath.Abs(f); err != nil{
+				return nil, nil, err
+			} else {
+				f = ff
+			}
+		}
+		for _, bf := range buildFiles {
+			if f == bf {
+				return nil, nil, nil
```
chanseokoh: Considering the method name, I think it's worth adding a comment `nil, nil` means a full rebuild.

pkg/skaffold/build/jib/jib_sync.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	if len(e.Deleted) != 0 {
+		// change into logging
+		fmt.Println("Deletions are not supported by jib auto sync at the moment")
+		return nil, nil, nil;
+	}
+
+	// we need to do another build and get a new sync map
+	newSyncMap, err := getSyncMap(ctx, workspace, artifact)
```
chanseokoh: Oh, so `getSyncMap` is really supposed to trigger a build itself (i.e., doing `package`)?

I'd like to understand how this could potentially interfere (race condition) with normal file watching, as we talked about before.

pkg/skaffold/build/jib/jib_sync.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	line := bytes.Replace(matches[1], []byte(`\`), []byte(`\\`), -1)
+	return json.Unmarshal(line, &sm)
+}
+
+func updateModTime(se []SyncEntry) error {
+	for i, _ := range se {
+		e := &se[i]
+		if info, err := os.Stat(e.Src); err != nil {
```
chanseokoh: I hope this (and the operation to compute absolute path above) is cheap to do. The sync map will inherently be small, so we should probably good?

pkg/skaffold/build/jib/maven.go
chanseokoh: Sorry, forget prepending `jib:_`. I see it's added in `mavenBuildArgs`.

pkg/skaffold/build/jib/jib_sync.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+		for _, bf := range buildFiles {
+			if f == bf {
+				return nil, nil, nil
+			}
+		}
+	}
+
+	// it seems less than efficient to keep the original JSON structure when doing look ups, so maybe we should only use the json objects for serialization
+	// and store the sync data in a better type?
```
briandealwis: The original format leads to O(n) checks

pkg/skaffold/sync/sync.go
```diff
@@ -260,3 +287,18 @@ func Perform(ctx context.Context, image string, files syncMap, cmdFn func(contex
...
 	return errs.Wait()
 }
+
+func Init(ctx context.Context, artifacts []*latest.Artifact) error {
+	for i := range artifacts {
+		a := artifacts[i]
+		if a.Sync != nil && a.Sync.Auto != nil{
+			if a.JibArtifact != nil {
+				if err := jib2.InitSync(ctx, a.Workspace, a.JibArtifact); err != nil {
+					// maybe should just disable sync here, instead of dead?
```
briandealwis:
```suggestion
					// maybe should just disable sync here, instead of dying?
```

loosebazooka:
this is covered by a bunch of new PRs

PR #3391, opened 2019-12-16
Add forwarding deployer muxer to enable multiple deployers

corneliusweig:
Relates to #2875, #3292

Should merge before : #3392 

**Description**

Add a deployer implementation (`DeployerMux`) that multiplexes the deployer actions with multiple deployer instances. This new deployer is simply a `[]deploy.Deployer` which forwards the commands of the `Deployer` interface to the instances in the list. When encountering an error, `DeployerMux` aborts and returns this error. Otherwise it collects the outputs and returns all results in bulk.

This is a pre-requisite for enabling multiple deployers in `skaffold.yaml`.

**User facing changes**

n/a

This PR only adds the functionality but does not enable anything.

**Next PRs.**

- adapt `skaffold.yaml` to accept multiple deployer configurations in the `deploy` field.

**Submitter Checklist**

review them:

**Reviewer Notes**

- [ ] Unit test added.

### DIFF
--- a/pkg/skaffold/deploy/deploy_mux.go
+++ b/pkg/skaffold/deploy/deploy_mux.go
@@ -20,9 +20,8 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"sort"
 	"strings"
-
-	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
 )
@@ -32,7 +31,31 @@ import (
 // it collects the results and returns it in bulk.
 type DeployerMux []Deployer
 
-var _ Deployer = DeployerMux{}
+type unit struct{}
+
+// stringSet helps to de-duplicate a set of strings.
+type stringSet map[string]unit
+
+func newStringSet() stringSet {
+	return make(map[string]unit)
+}
+
+// insert adds strings to the set.
+func (s stringSet) insert(strings ...string) {
+	for _, item := range strings {
+		s[item] = unit{}
+	}
+}
+
+// toList returns the sorted list of inserted strings.
+func (s stringSet) toList() []string {
+	res := make([]string, 0, len(s))
+	for item := range s {
+		res = append(res, item)
+	}
+	sort.Strings(res)
+	return res
+}
 
 func (m DeployerMux) Labels() map[string]string {
 	labels := make(map[string]string)
@@ -43,27 +66,27 @@ func (m DeployerMux) Labels() map[string]string {
 }
 
 func (m DeployerMux) Deploy(ctx context.Context, w io.Writer, as []build.Artifact, ls []Labeller) *Result {
-	seenNamespaces := sets.String{}
+	seenNamespaces := newStringSet()
 	for _, deployer := range m {
 		result := deployer.Deploy(ctx, w, as, ls)
 		if result.err != nil {
 			return result
 		}
-		seenNamespaces.Insert(result.Namespaces()...)
+		seenNamespaces.insert(result.Namespaces()...)
 	}
-	return NewDeploySuccessResult(seenNamespaces.List())
+	return NewDeploySuccessResult(seenNamespaces.toList())
 }
 
 func (m DeployerMux) Dependencies() ([]string, error) {
-	deps := sets.String{}
+	deps := newStringSet()
 	for _, deployer := range m {
 		result, err := deployer.Dependencies()
 		if err != nil {
 			return nil, err
 		}
-		deps.Insert(result...)
+		deps.insert(result...)
 	}
-	return deps.List(), nil
+	return deps.toList(), nil
 }
 
 func (m DeployerMux) Cleanup(ctx context.Context, w io.Writer) error {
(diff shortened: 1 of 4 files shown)


pkg/skaffold/deploy/deploy_mux.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
+)
+
+// DeployerMux forwards all method calls to the deployers it contains.
+// When encountering an error, it aborts and returns the error. Otherwise,
+// it collects the results and returns it in bulk.
+type DeployerMux []Deployer
+
+var _ Deployer = DeployerMux{}
```
balopat: What does this line do?

balopat: gotcha, it is just a compile time check to make sure that `DeployerMux` implements the `Deployer` interface.

balopat: a comment would be nice?

pkg/skaffold/deploy/deploy_mux.go
```diff
@@ -0,0 +1,90 @@
...
+func (m DeployerMux) Labels() map[string]string {
+	labels := make(map[string]string)
+	for _, deployer := range m {
+		copyMap(labels, deployer.Labels())
+	}
+	return labels
+}
+
+func (m DeployerMux) Deploy(ctx context.Context, w io.Writer, as []build.Artifact, ls []Labeller) *Result {
+	seenNamespaces := sets.String{}
```
balopat: this is cool, but kinda scares me that it's in kubernetes. Haven't made up my mind yet about it. 
The risk of it disappearing / changing is minimal as skaffold is for kubernetes. But in general we tend to avoid depending on utilities from dependencies - unless the dependency is specifically for that utility. E.g. docker has some great stuff in their ioutils package but we decided not to use them.

pkg/skaffold/deploy/deploy_mux.go
```diff
@@ -0,0 +1,90 @@
...
+		if result.err != nil {
+			return result
+		}
+		seenNamespaces.Insert(result.Namespaces()...)
+	}
+	return NewDeploySuccessResult(seenNamespaces.List())
+}
+
+func (m DeployerMux) Dependencies() ([]string, error) {
+	deps := sets.String{}
```
balopat: same here, the set semantics are correct - but not sure I would rely on the k8s version of it.

pkg/skaffold/deploy/deploy_mux.go
corneliusweig: Exactly, this was my local compile check that I have implemented the interface. As this pattern seems to be unused in Skaffold, I opted to remove the line instead.

pkg/skaffold/deploy/deploy_mux.go
```diff
@@ -0,0 +1,90 @@
...
+// When encountering an error, it aborts and returns the error. Otherwise,
+// it collects the results and returns it in bulk.
+type DeployerMux []Deployer
+
+var _ Deployer = DeployerMux{}
+
+func (m DeployerMux) Labels() map[string]string {
+	labels := make(map[string]string)
+	for _, deployer := range m {
+		copyMap(labels, deployer.Labels())
```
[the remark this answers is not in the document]
corneliusweig: Good catch. This was already addressed with #3390, where the deployer labels are removed from the global list of labellers and are applied just in the specific deployer instead.
So I think the semantics of this line is correct.

pkg/skaffold/deploy/deploy_mux.go
corneliusweig: Yeah that totally makes sense. The de-duplication is easily implemented so there's no need to add a dependency here.
I think there are a few places where Skaffold code could benefit from a StringSet util, but this is not the right PR to make that change :). So I've inlined a small helper in this file.

corneliusweig:
Sorry for my confusion. I accidentally committed the review comments to the wrong branch 2 weeks ago :man_facepalming:
Please have another look :)

pkg/skaffold/deploy/deploy_mux.go
```diff
@@ -0,0 +1,113 @@
...
+func (m DeployerMux) Dependencies() ([]string, error) {
+	deps := newStringSet()
+	for _, deployer := range m {
+		result, err := deployer.Dependencies()
+		if err != nil {
+			return nil, err
+		}
+		deps.insert(result...)
+	}
+	return deps.toList(), nil
```
dgageot: Won't that change the order of dependencies?

corneliusweig: Well, yes it does. But `deps` is a set, so there is no defined order and I can just choose to output sorted, right?

At least the helm deployer also sorts its dependencies, so that should be ok.

dgageot: I think, it's better to sort them here.
What we can do latter is change `Dependencies()` so that it returns a sorted string set.

corneliusweig: So you mean move the sorting step from `toList` after the `deps.toList()` call? (sorry if I'm asking stupid questions)

dgageot: I'm the stupid here, I did not see the cal too sort in toList(). Forget everything I said.

corneliusweig: :sweat_smile:

PR #3394, opened 2019-12-17
Skaffold init recognises nodeJS projects built with Buildpacks

dgageot:
This is the beginning of `skaffold init`'s support for Cloud Native Buildpacks.

It's still a work in progress and currently only recognizes nodejs projects that have a package.json.
Since it's activated by a hidden flag, I'd like to get this merged before adding support for more languages.

### DIFF
--- a/pkg/skaffold/build/buildpacks/init.go
+++ b/pkg/skaffold/build/buildpacks/init.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


pkg/skaffold/build/buildpacks/init.go
```diff
@@ -0,0 +1,82 @@
...
+var (
+	ValidateConfig = validateConfig
+)
+
+// Name is the name of the Buildpack builder
+var Name = "Buildpacks"
+
+// Buildpack holds information about a Buildpack project
+type Buildpacks struct {
+	File string `json:"path,omitempty"`
```
briandealwis: Would it make sense to name this `Path` in keeping with the JSON key?

pkg/skaffold/initializer/init.go
```diff
@@ -654,8 +675,10 @@ func checkConfigFile(filePath string, force bool, potentialConfigs *[]string) (b
 	}
 
 	if IsSupportedKubernetesFileExtension(filePath) {
-		*potentialConfigs = append(*potentialConfigs, filePath)
-		return true, nil
+		if filepath.Base(filePath) != "package.json" { // Not a valid k8s manifest
```
briandealwis: This doesn't seem scalable and there must be dozens of JSON and YAML files that caught up in this!

dgageot: Yeah, it was a quick hack. I think the verification of k8s manifests has to be improved

PR #3395, opened 2019-12-17
Auto sync with Skaffold

dgageot:
This is WIP to introduce auto-sync configuration to Buildpacks artifacts.

This feature requires collaboration from the buildpacks that have to export the sync metadata and also that need to be able to build a container image used just for dev.

!!This is very early stage. Many things need to be improved!!


loosebazooka:
@dgageot I think there might be some conflicts between this and #3382. Also, I'm curious about autosync being on by default (for jib at least).

PR #3406, opened 2019-12-19
Simplify skaffold init code

### LINKED ISSUE #3405: `skaffold init` can produce an erroneous config with non-default Dockerfile name
Run `skaffold init` on a folder that looks like this:
```
./front/main.go
./front/DockerfileProd
./k8s.yaml
```

It should produce this config:
```yaml
apiVersion: skaffold/v2alpha1
kind: Config
metadata:
  name: getting-started
build:
  artifacts:
  - image: gcr.io/k8s-skaffold/skaffold-example
    context: front
    docker:
      dockerfile: DockerfileProd
  tagPolicy:
    gitCommit: {}
  local: {}
deploy:
  kubectl:
    manifests:
    - k8s.yaml
```

It instead produces this config:
```yaml
apiVersion: skaffold/v2alpha1
kind: Config
metadata:
  name: getting-started
build:
  artifacts:
  - image: gcr.io/k8s-skaffold/skaffold-example
    context: front
    docker:
      dockerfile: front/DockerfileProd
  tagPolicy:
    gitCommit: {}
  local: {}
deploy:
  kubectl:
    manifests:
    - k8s.yaml
```

dgageot:
Each commit is a simplification. I didn't merge them to ease the review.
Also fixes #3405

### DIFF
--- a/pkg/skaffold/build/buildpacks/init.go
+++ b/pkg/skaffold/build/buildpacks/init.go
@@ -57,7 +57,7 @@ func (c ArtifactConfig) UpdateArtifact(a *latest.Artifact) {
 
 // ConfiguredImage returns the target image configured by the builder, or empty string if no image is configured
 func (c ArtifactConfig) ConfiguredImage() string {
-	// Target image is not configured in dockerfiles
+	// Target image is not configured in buildpacks
 	return ""
 }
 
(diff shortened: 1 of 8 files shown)


dgageot:
Good catch @balopat!

It used to be that only a few defaults were explicit.
Now they all are. I don't really like that,
I was thinking of having none...

Anyways, I think it’s best to add a non regression test and make sure we produce the same yaml as before. Wdyt?

dgageot:
I changed a few things to make sure we produce the same yaml as before.

PR #3410, opened 2019-12-20
Integration test: check the output of `skaffold init`

dgageot:
Compare the output with the expected output.

### DIFF
--- a/integration/init_test.go
+++ b/integration/init_test.go
@@ -63,10 +63,12 @@ func TestInit(t *testing.T) {
 	for _, test := range tests {
 		testutil.Run(t, test.name, func(t *testutil.T) {
 			initArgs := append([]string{"--force"}, test.args...)
+
 			skaffold.Init(initArgs...).InDir(test.dir).WithConfig("skaffold.yaml.out").RunOrFail(t.T)
 
 			checkGeneratedConfig(t, test.dir)
 
+			// Make sure the skaffold yaml can be parsed
 			skaffold.Diagnose().InDir(test.dir).WithConfig("skaffold.yaml.out").RunOrFail(t.T)
 		})
 	}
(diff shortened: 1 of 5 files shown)


integration/init_test.go
```diff
@@ -56,21 +59,55 @@ func TestInit(t *testing.T) {
...
 			defer deleteNs()
 
 			initArgs := append([]string{"--force"}, test.args...)
-			skaffold.Init(initArgs...).InDir(test.dir).WithConfig("skaffold.yaml.out").RunOrFail(t)
+			skaffold.Init(initArgs...).InDir(test.dir).WithConfig("skaffold.yaml.out").RunOrFail(t.T)
+
+			checkGeneratedConfig(t, test.dir)
 
-			skaffold.Run().InDir(test.dir).WithConfig("skaffold.yaml.out").InNs(ns.Name).RunOrFail(t)
+			skaffold.Run().InDir(test.dir).WithConfig("skaffold.yaml.out").InNs(ns.Name).RunOrFail(t.T)
```
balopat: if you changed this to Diagnose above, we could do this here too, right?

dgageot: This sample also generates kubernetes yaml, so it's better to do a full run. I'll add a comment

integration/init_test.go
```diff
@@ -56,21 +59,55 @@ func TestInit(t *testing.T) {
...
 	}
 }
+
+func checkGeneratedConfig(t *testutil.T, dir string) {
+	expectedOutput, err := ioutil.ReadFile(filepath.Join(dir, "skaffold.yaml"))
+	t.CheckNoError(err)
+
+	output, err := ioutil.ReadFile(filepath.Join(dir, "skaffold.yaml.out"))
+	t.CheckNoError(err)
+	t.CheckDeepEqual(string(expectedOutput), string(output))
```
balopat: This might become a bit brittle over time, we could add some guidance to make maintenance a bit less painful in the comment here or as part of the error message that: "if you see failure here, inspect the changes. In case the changes in config are expected (e.g. version change), run `find integration -name skaffold.yaml.out -execdir mv skaffold.yaml.out skaffold.yaml \;` to override all the skaffold.yamls to the latest version.

balopat: on the long run we might want to make this a bit smarter to ignore guaranteed cases of string changes that are okay - e.g. version change.

dgageot: The version change should be made automatically by our script

PR #3413, opened 2019-12-20
Setup kind and build the docker image in //

dgageot:
By creating the kind cluster in parallel of building the docker image, 
this could make the integration tests 2 minutes faster.


balopat:
I approved it, but let's see if we run into any issues on Travis machines :D

dgageot:
yeah, I'm also not sure it's really going to be faster on Travis.

dgageot:
Seems to be ok. Integration phase is down to 17minutes. I've that number in the recent past but rarely.

PR #3414, opened 2019-12-20
fix wait logic in TestWaitForPodSucceeded

balopat:


balopat:
The logic is this for each test: 

main go routine: `modify pod state -> sleep(update_time) -> modify pod state -> sleep(update_time) -> ...`

waitfor pod go routine: `wait for (pod to be in <state>) or timeout after (timeout_time)` 

if `(update_time == timeout_time)` (like before), then this will be flaky with a high probability of a race. It's like a "super impatient client that waits for the pod to get into a phase". Given in the first two test cases the desired update happens after the first sleep. Not sure how this ever worked consistently. 

if `update_time << timeout_time`, this fixes it, it makes the "client more patient". 

The third test case is a bit superfluous, we shouldn't need to "update" the state 5 times to the same thing, one initial update should be sufficient and the "client" will timeout anyway.

PR #3421, opened 2019-12-20
Diagnose TestDiagnose

dgageot:


dgageot:
It looks like it takes about 30s to download gradle and about 15s to download micronaut and deps...
Need to think about how to improve that

PR #3423, opened 2019-12-20
Upgrade a few CLI tools

dgageot:


dgageot:
Let's remove the helm upgrade because it requires upgrading the test cluster too

PR #3424, opened 2019-12-20
Try bionic on Travis

dgageot:


dgageot:
Let's close this one. It's not faster than xenial

PR #3432, opened 2019-12-23
Use ko instead of buildpacks for the custom builder

dgageot:
Since buildpacks are natively supported, it’s better
to use another example for the custom build script.


balopat:
@nkubala want to have a look too?

PR #3437, opened 2019-12-24
Commit annoying go.sum

dgageot:


dgageot:
I'm force merging this fil. It really annoying to have it modified locally by the go toolchain for whatever reason

Issue #3399, opened 2019-12-18
Kaniko build on Google Cloud Build doesn't include --target parameter


iamarcel:
### Expected behavior
When scheduling a Kaniko build on GCB, the artifact's Target parameter gets passed as an argument to the Kaniko executor on GCB, just like when building it in-cluster.

### Actual behavior
The Kaniko build spec function [doesn't include code to create a `--target` parameter](https://github.com/GoogleContainerTools/skaffold/blob/42330f2/pkg/skaffold/build/gcb/kaniko.go#L30-L63) (while the respective method for in-cluster builds [does](https://github.com/GoogleContainerTools/skaffold/blob/42330f2/pkg/skaffold/build/cluster/kaniko.go#L129-L131)).

The tests for the GCB Kaniko build spec [don't include the Target parameter](https://github.com/GoogleContainerTools/skaffold/blob/42330f20d0a4a02e9d4386d2002b1b4f2111ecd8/pkg/skaffold/build/gcb/kaniko_test.go#L29) either.

### Information

- Skaffold version: v1.0.1
- Operating system: Fedora 31
- Relevant contents of skaffold.yaml:

```yaml
build:
  googleCloudBuild:
    projectId: my-google-project
  artifacts:
  - image: eu.gcr.io/my-google-project/build
    kaniko:
      target: build
      cache: {}
      buildContext:
        gcsBucket: my-kaniko-bucket
```

The GCB build details shows these arguments:

```
--destination eu.gcr.io/my-google-project/build:f4c58a6 --dockerfile Dockerfile --cache
```

balopat:
Thank you @iamarcel - I agree that this is missing, and you are pointing to the right places. PRs are welcome!

hwanjin-jeong:
I contribute to this issue
Please see that you want to check the resolved code: #3462

dgageot:
Fixed with https://github.com/GoogleContainerTools/skaffold/pull/3462
Thanks a lot @hwanjin-jeong !

PR #3462, opened 2019-12-31
Add --target parameter with kaniko on Google Cloud Build

hwanjin-jeong:
Fixes `#3399`

### DIFF
--- a/pkg/skaffold/build/gcb/kaniko.go
+++ b/pkg/skaffold/build/gcb/kaniko.go
@@ -36,6 +36,7 @@ func (b *Builder) kanikoBuildSpec(artifact *latest.KanikoArtifact, tag string) (
 	kanikoArgs := []string{
 		"--destination", tag,
 		"--dockerfile", artifact.DockerfilePath,
+		"--target", artifact.Target,
 	}
 	kanikoArgs = append(kanikoArgs, buildArgs...)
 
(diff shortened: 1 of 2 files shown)


pkg/skaffold/build/gcb/kaniko.go
```diff
@@ -36,6 +36,7 @@ func (b *Builder) kanikoBuildSpec(artifact *latest.KanikoArtifact, tag string) (
 	kanikoArgs := []string{
 		"--destination", tag,
 		"--dockerfile", artifact.DockerfilePath,
+		"--target", artifact.Target,
```
dgageot: Although it works well with an empty target, I'd rather add the `--target` only if the target is set. That feels better. You'd have to test both cases then.

pkg/skaffold/build/gcb/kaniko_test.go
```diff
@@ -62,10 +62,10 @@ func TestKanikoBuildSpec(t *testing.T) {
 			Args: []string{
 				"--destination", "nginx",
 				"--dockerfile", "Dockerfile",
-				"--target", "builder",
 				"--build-arg", "arg1=value1",
 				"--build-arg", "arg2",
 				"--cache",
+				"--target", "builder",
```
dgageot: Could you test for both cases?

pkg/skaffold/build/gcb/kaniko.go
hwanjin-jeong: thanks for your advice.

pkg/skaffold/build/gcb/kaniko_test.go
hwanjin-jeong: sorry, i late for refactoring.
do you have any suggestions?

PR #3463, opened 2019-12-31
Refactor the code that deals with Docker API

dgageot:
+ Remove as much as possible the need for t.Override() to mock the Docker API
 + Create a single docker.API object and inject it into every object that deals with the Docker API (local or remote)


pkg/skaffold/docker/types.go
```diff
@@ -0,0 +1,111 @@
+/*
+Copyright 2019 The Skaffold Authors
```
tejal29: 2020

pkg/skaffold/docker/build_args.go
```diff
@@ -0,0 +1,101 @@
+/*
+Copyright 2019 The Skaffold Authors
```
tejal29: 2020

pkg/skaffold/runner/new.go
```diff
@@ -43,12 +44,14 @@ import (
 
 // NewForConfig returns a new SkaffoldRunner for a SkaffoldConfig
 func NewForConfig(runCtx *runcontext.RunContext) (*SkaffoldRunner, error) {
+	docker := docker.NewAPI(runCtx)
```
tejal29: This now happens irrespective of the what builder you are using. 
e.g. if i am skaffold user who does not have docker installed locally and all my builds are running on my cluster shd this be still initialized?

Would this not cause an error here https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/docker/client.go#L79 for skaffold user who don't have docker installed?

dgageot:
The idea is that it should behave like before. I might have missed something though. The api is lazy. It won’t connect until it’s used.

dgageot:
I'm going to close this one. I think it's important to make that change but it can wait.

PR #3483, opened 2020-01-08
Use unique key for jib caches

loosebazooka:
If two jib projects had clashing Project fields (even empty), they would trigger refreshing dependencies on every cycle of the skaffold dev loop. This creates a projectKey from the workspace and the project and uses it as the key. 

While nothing was broken before, this provides the functionality we actually expected.

### DIFF
--- a/pkg/skaffold/build/jib/jib.go
+++ b/pkg/skaffold/build/jib/jib.go
@@ -86,10 +86,16 @@ type filesLists struct {
 }
 
 // watchedFiles maps from project name to watched files
-var watchedFiles = map[string]filesLists{}
+var watchedFiles = map[projectKey]filesLists{}
 
-func GetBuildDefinitions(a *latest.JibArtifact) []string {
-	return watchedFiles[a.Project].BuildDefinitions
+type projectKey string
+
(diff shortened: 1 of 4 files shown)


chanseokoh:
FTR: https://github.com/GoogleContainerTools/skaffold/pull/3369#discussion_r357238030 and https://github.com/GoogleContainerTools/skaffold/pull/3369#discussion_r359414517

pkg/skaffold/build/jib/jib.go
```diff
@@ -86,10 +86,16 @@ type filesLists struct {
...
 // watchedFiles maps from project name to watched files
-var watchedFiles = map[string]filesLists{}
+var watchedFiles = map[projectKey]filesLists{}
 
-func GetBuildDefinitions(a *latest.JibArtifact) []string {
-	return watchedFiles[a.Project].BuildDefinitions
+type projectKey string
+
+func getProjectKey(workspace string, a *latest.JibArtifact) projectKey {
+	return projectKey(workspace + "+" + a.Project)
```
chanseokoh: As discussed, the fix in this PR make sense. I'll just let other Go folks to approve this particular implementation.

PR #3488, opened 2020-01-09
Clarify GCP service account and secret creation

wstrange:
I got tripped up on the name of the secret (my key name was incorrect). This might help


wstrange:
Looking further into the gcpBucket - it is not clear to me that skaffold even uses this feature. It seems to ignore it and upload directly to the pod.  Are the docs wrong?

dgageot:
@wstrange I'll merge and submit another PR to remove that `gcsBucket` part

PR #3489, opened 2020-01-09
Add common jib sync code for processing syncmap

loosebazooka:
~~Introduces code to initialize sync caches, update them and check
for differences when watched files change~~
~~It does not allow actually configuring sync yet. It just sets up the necessary underlying code to make calls out to jib and process the diffs~~

Converted this into two separate PRs

This one now just calls out to jib to get the syncmap and coverts json to a usable object.

### DIFF
--- a/pkg/skaffold/build/jib/sync.go
+++ b/pkg/skaffold/build/jib/sync.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Skaffold Authors
+Copyright 2020 The Skaffold Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/skaffold/build/jib/sync.go
+++ b/pkg/skaffold/build/jib/sync.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package jib
+
+import (
+	"bytes"
+	"encoding/json"
+	"os"
+	"os/exec"
+	"regexp"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+)
+
+type SyncMap map[string]SyncEntry
+
+type SyncEntry struct {
+	Dest     []string
+	FileTime time.Time
+	IsDirect bool
+}
+
+type JSONSyncMap struct {
+	Direct    []JSONSyncEntry `json:"direct"`
+	Generated []JSONSyncEntry `json:"generated"`
+}
+
+type JSONSyncEntry struct {
+	Src  string `json:"src"`
+	Dest string `json:"dest"`
+}
+
+func getSyncMapFromSystem(cmd *exec.Cmd) (*SyncMap, error) {
+	jsm := JSONSyncMap{}
+	stdout, err := util.RunCmdOut(cmd)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get Jib sync map")
+	}
+
+	// To parse the output, search for "BEGIN JIB JSON", then unmarshal the next line into the pathMap struct.
+	// Syncmap is transitioning to "BEGIN JIB JSON: SYNCMAP/1" starting in jib 2.0.0
+	// perhaps this feature should only be included from 2.0.0 onwards? And we generally avoid this?
+	matches := regexp.MustCompile(`BEGIN JIB JSON(?:: SYNCMAP/1)?\r?\n({.*})`).FindSubmatch(stdout)
+	if len(matches) == 0 {
+		return nil, errors.New("failed to get Jib Sync data")
+	}
+
+	line := bytes.Replace(matches[1], []byte(`\`), []byte(`\\`), -1)
+	if err := json.Unmarshal(line, &jsm); err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	sm := make(SyncMap)
+	for _, de := range jsm.Direct {
+		info, err := os.Stat(de.Src)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not obtain file mod time")
+		}
+		sm[de.Src] = SyncEntry{
+			Dest:     []string{de.Dest},
+			FileTime: info.ModTime(),
+			IsDirect: true,
+		}
+	}
+	for _, ge := range jsm.Generated {
+		info, err := os.Stat(ge.Src)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not obtain file mod time")
+		}
+		sm[ge.Src] = SyncEntry{
+			Dest:     []string{ge.Dest},
+			FileTime: info.ModTime(),
+			IsDirect: false,
+		}
+	}
+	return &sm, nil
+}
--- a/pkg/skaffold/build/jib/sync.go
+++ b/pkg/skaffold/build/jib/sync.go
@@ -68,27 +68,26 @@ func getSyncMapFromSystem(cmd *exec.Cmd) (*SyncMap, error) {
 	}
 
 	sm := make(SyncMap)
-	for _, de := range jsm.Direct {
-		info, err := os.Stat(de.Src)
-		if err != nil {
-			return nil, errors.Wrap(err, "could not obtain file mod time")
-		}
-		sm[de.Src] = SyncEntry{
-			Dest:     []string{de.Dest},
-			FileTime: info.ModTime(),
-			IsDirect: true,
-		}
+	if err := sm.addEntries(jsm.Direct, true); err != nil {
+		return nil, errors.WithStack(err)
 	}
-	for _, ge := range jsm.Generated {
-		info, err := os.Stat(ge.Src)
-		if err != nil {
-			return nil, errors.Wrap(err, "could not obtain file mod time")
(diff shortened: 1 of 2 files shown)


pkg/skaffold/build/jib/sync.go
```diff
@@ -0,0 +1,94 @@
...
+	Direct    []JSONSyncEntry `json:"direct"`
+	Generated []JSONSyncEntry `json:"generated"`
+}
+
+type JSONSyncEntry struct {
+	Src  string `json:"src"`
+	Dest string `json:"dest"`
+}
+
+func getSyncMapFromSystem(cmd *exec.Cmd) (*SyncMap, error) {
```
balopat: nit: why `FromSystem`?

pkg/skaffold/build/jib/sync.go
```diff
@@ -0,0 +1,94 @@
...
+	jsm := JSONSyncMap{}
+	stdout, err := util.RunCmdOut(cmd)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get Jib sync map")
+	}
+
+	// To parse the output, search for "BEGIN JIB JSON", then unmarshal the next line into the pathMap struct.
+	// Syncmap is transitioning to "BEGIN JIB JSON: SYNCMAP/1" starting in jib 2.0.0
+	// perhaps this feature should only be included from 2.0.0 onwards? And we generally avoid this?
```
balopat: are there other differences in the sync output or it's the same?

pkg/skaffold/build/jib/sync_test.go
```diff
@@ -0,0 +1,134 @@
...
+		{
+			description: "bad marker",
+			stdout:      "BEGIN JIB JSON: BAD/1\n{}",
+			shouldErr:   true,
+			expected:    nil,
+		},
+		{
+			description: "direct only",
+			stdout: "BEGIN JIB JSON: SYNCMAP/1\n" +
+				fmt.Sprintf("{\"direct\":[{\"src\":\"%s\",\"dest\":\"%s\"}]}", dep1, dep1Target),
```
balopat: you could use backticks `` `{"direct":[{"src":"%s","dest":"%s"}]}` `` in all these

pkg/skaffold/build/jib/sync.go
```diff
@@ -0,0 +1,94 @@
...
+	matches := regexp.MustCompile(`BEGIN JIB JSON(?:: SYNCMAP/1)?\r?\n({.*})`).FindSubmatch(stdout)
+	if len(matches) == 0 {
+		return nil, errors.New("failed to get Jib Sync data")
+	}
+
+	line := bytes.Replace(matches[1], []byte(`\`), []byte(`\\`), -1)
```
balopat: What is this for? Windows support? Also don't see it covered in testcases unless I'm missing something.

pkg/skaffold/build/jib/sync.go
```diff
@@ -0,0 +1,94 @@
...
+	if err := json.Unmarshal(line, &jsm); err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	sm := make(SyncMap)
+	for _, de := range jsm.Direct {
```
balopat: Do you foresee a lot of differences between Direct and Generated besides isDirect: true/false? If not - maybe we could extract a local function here and just pass in a `[]JSONSyncEntry` and `isDirect: bool` to it and reduce the duplication here?

pkg/skaffold/build/jib/sync.go
balopat: maybe just `getSyncMap`?  in the end we are under jib anyways

pkg/skaffold/build/jib/sync.go
loosebazooka: The output is the same, we decided we wanted some more matching data. All our json is prefixed with "BEGIN JIB JSON" so if we were run two of these in one execution (for example: getDependencies and getSyncMap) we wouldn't be able to know which json is for which easily.

pkg/skaffold/build/jib/sync.go
loosebazooka: Yeah that makes sense.

pkg/skaffold/build/jib/sync.go
loosebazooka: Good questions, this is copied from another jib-json section, I'll verify what it's doing.

chanseokoh: https://github.com/GoogleContainerTools/skaffold/pull/2276#discussion_r304542768

pkg/skaffold/build/jib/sync.go
loosebazooka: The difference is in that 1.8.0 we already release syncmap generation with the prefix "BEGIN JIB JSON", and we're correcting that with 2.0.0 -> "BEGIN JIB JSON: SYNCMAP/1"

pkg/skaffold/build/jib/sync.go
loosebazooka: there is actually another `getSyncMap` in this series of PRs, I guess since I split them out, this naming is a little weird.

pkg/skaffold/build/jib/sync.go
balopat: @chanseokoh great minds...

It's not going to be easy to cover this with integration tests as we don't have Windows integration tests just yet. We _could_ (and should:)) take the output of jib on a Windows env and paste it in our testcases, similarly how we put unix outputs in there.

loosebazooka: yeah working on that, might need to restructure a few things to actually allow it.

pkg/skaffold/build/jib/sync.go
balopat: What does the other `getSyncMap` do?

pkg/skaffold/build/jib/sync.go
balopat: Sounds good. This should be fine then to include from 1.8.0, I see no issue there. Do you see any risk there? I guess it would make this expression cleaner, but other than that there is no advantage to limit to 2.0.0+, right?

pkg/skaffold/build/jib/sync.go
loosebazooka: I dug into this some more. I think it's an artifact from previous jib outputs, it just happened to work because multiple back slashes are ignored. I think the reality is that we don't need to do this, I have a new commit that removes it, will push tomorrow.

pkg/skaffold/build/jib/sync.go
loosebazooka: I think we can limit sync to 2.0.0+ and just remove this code. 2.0.0 should go out next week

pkg/skaffold/build/jib/sync.go
loosebazooka: It checks if gradle/maven and then collects the right cmd and then calls out to this.

pkg/skaffold/build/jib/sync.go
loosebazooka: It does kind of need to be here for now for this to "work" while in development.

pkg/skaffold/build/jib/sync.go
loosebazooka: done.

pkg/skaffold/build/jib/sync.go
```diff
@@ -0,0 +1,91 @@
...
+	jsm := JSONSyncMap{}
+	stdout, err := util.RunCmdOut(cmd)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get Jib sync map")
+	}
+
+	// To parse the output, search for "BEGIN JIB JSON", then unmarshal the next line into the pathMap struct.
+	// Syncmap is transitioning to "BEGIN JIB JSON: SYNCMAP/1" starting in jib 2.0.0
+	// perhaps this feature should only be included from 2.0.0 onwards? And we generally avoid this?
+	matches := regexp.MustCompile(`BEGIN JIB JSON(?:: SYNCMAP/1)?\r?\n({.*})`).FindSubmatch(stdout)
```
chanseokoh: Do you think it's better to fix the version `/1`? For example, if the user users a newer Jib version returning `/2`, the user has to upgrade Skaffold too. Of course, that can be our strategy, but the error message below when there's no match doesn't help to the user. The user will just see "failed to get Jib Sync data" after upgrading Jib.

And perhaps remove the following comment? Weird to leave it in the code.

``// perhaps this feature should only be included from 2.0.0 onwards? And we generally avoid this?``

loosebazooka: I want to leave it in while we are still doing both types of parsing. (comment will be removed and code cleaned up after we 2.0.0 and we have a way to variably set the value of: https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/jib/maven.go#L39 -- which will happen in a followup)

So doing `SYNCMAP/version` is just providing us some metadata for futureproofing. We don't actually use this number in anyway, but for instance if we do create a new version of skaffold that accepts both 1 and 2, we can make this value different. Currently, if the API changes, the parser will die anyway.

chanseokoh: I guess we are not on the same page...

For example, in 2020, imagine we release Skaffold version A having this code that looks for `SYNCMAP/1`. In 2021, we release Skaffold version B that can parse both `/1` and `/2` and a new Jib version that returns `SYNCMAP/2`. Now, at some point, the user upgrades Jib to a new version (returning `/2`), but the user is still using Skaffold version A. Then, my concern is that `skaffold dev` will just die saying

```
error: failed to get Jib Sync data
```

And the user has no idea why.

chanseokoh: But I guess we'll have some sort of problem either way, so I'm not actually against settling with this policy.

loosebazooka: Right, I think at least this way it possible to make it a little more explicit.

pkg/skaffold/build/jib/sync.go
balopat: gotcha

pkg/skaffold/build/jib/sync.go
balopat: we can just remove it once 2.0.0 is out, let's merge it.

PR #3495, opened 2020-01-10
Demonstrate inferred sync

dgageot:


briandealwis:
I think this is obsoleted with the auto-inferred sync in #3496

dgageot:
@briandealwis I think #3496 might not be good.

PR #3499, opened 2020-01-10
Dockerfile detector will only check files containing "Dockerfile" in the name

### LINKED ISSUE #3484: skaffold init thinks some python files are Dockerfiles
image

content of: `src/messages/tests.py`
```
from django.test import TestCase

# Create your tests here.
```

spectrox:
Fixes #3484.

**Description**

Initializer tries to find all available builders by checking each file in the tree for compatible with Dockerfile syntax.
This syntax is not unique and python file can be taken for Dockerfile:
```
from django.test import TestCase
```

This fix adds additional check to the file name which should contain "Dockerfile".

I am not sure if that is appropriate solution, because it can affect existing users.
But as far as I know, developers tend to keep `Dockerfile` in the name even when using multiple files in the same folder.

**User facing changes**

**Before**

When running `skaffold init` with python files like the one mentioned above you can see them in the list of available builders.

**After**

No python files should be listed in the survey.

**Submitter Checklist**

review them:

**Reviewer Notes**

- [ ] Unit test added.

### DIFF
--- a/pkg/skaffold/initializer/init.go
+++ b/pkg/skaffold/initializer/init.go
@@ -288,9 +288,12 @@ func detectBuilders(enableJibInit, enableBuildpackInit bool, path string) ([]Ini
 	}
 
 	// Check for Dockerfile
-	if docker.Validate(path) {
-		results := []InitBuilder{docker.ArtifactConfig{File: path}}
-		return results, true
+	base := filepath.Base(path)
+	if strings.Contains(base, "Dockerfile") {
+		if docker.Validate(path) {
+			results := []InitBuilder{docker.ArtifactConfig{File: path}}
+			return results, true
+		}
 	}
 
 	// TODO: Remove backwards compatibility if statement (not entire block)
--- a/pkg/skaffold/initializer/init.go
+++ b/pkg/skaffold/initializer/init.go
@@ -289,7 +289,7 @@ func detectBuilders(enableJibInit, enableBuildpackInit bool, path string) ([]Ini
 
 	// Check for Dockerfile
 	base := filepath.Base(path)
-	if strings.Contains(base, "Dockerfile") {
+	if strings.Contains(strings.ToLower(base), "dockerfile") {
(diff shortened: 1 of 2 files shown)


spectrox:
@googlebot I signed it!

pkg/skaffold/initializer/init.go
```diff
@@ -288,9 +288,12 @@ func detectBuilders(enableJibInit, enableBuildpackInit bool, path string) ([]Ini
 	}
 
 	// Check for Dockerfile
-	if docker.Validate(path) {
-		results := []InitBuilder{docker.ArtifactConfig{File: path}}
-		return results, true
+	base := filepath.Base(path)
+	if strings.Contains(base, "Dockerfile") {
```
dgageot: FWIW, I've also seen a lot of `*.dockerfile` out there

pkg/skaffold/initializer/init.go
```diff
@@ -288,9 +288,12 @@ func detectBuilders(enableJibInit, enableBuildpackInit bool, path string) ([]Ini
...
+		if docker.Validate(path) {
```
cedrickring: We could also check in `docker.Validate(path)` if there is some python-ish code after `from ...` (like `import <name>`). 

WDYT @dgageot

pkg/skaffold/initializer/init.go
spectrox: so, it would be better to make it case insensitive?

pkg/skaffold/initializer/init.go
spectrox: I don't think that Dockerfile validator should be aware of python-ish code constructs.
It's responsibility to validate Dockerfile and it does that by validating commands. Validating arguments would be too much overhead IMO.

pkg/skaffold/initializer/init.go
spectrox: I've made it case insensitive and updated test to cover this case.

PR #3501, opened 2020-01-11
Use origin/master as baseline for schema version check

corneliusweig:
**Description**

This is just a small teak to improve the usability of the schema version test.

Before it used the `master` branch as the reference version. However, this branch is not automatically updated by git when fetching changes from remote. On the other hand, the `origin/master` branch is reset to the newest version on every pull/fetch from origin. Hence, the reference version automatically updates even if the user never updates his `master` branch.

This assumes that users have a standard setup for remote repository names. But I think this is as widespread as the name `master`.

**User facing changes**

n/a

**Next PRs.**

n/a

**Submitter Checklist**

n/a

**Reviewer Notes**

- [ ] Unit test added.

### DIFF
--- a/hack/check-schema-changes.sh
+++ b/hack/check-schema-changes.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 
-# This check checks whether the PR compared to master contains any changes
+# This check checks whether the PR compared to origin/master contains any changes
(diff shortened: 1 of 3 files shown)


corneliusweig:
Hmm.. the failing kaniko test seems unrelated to me.

dgageot:
yeah, let's ignore that

PR #3503, opened 2020-01-13
Use pack CLI to build on GCB

dgageot:
Instead of using our own implementation of the Buildpacks lifecycle, it's more robust to use `pack` CLI.

### DIFF
--- a/deploy/buildpacks/publish.sh
+++ b/deploy/buildpacks/publish.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Skaffold Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+PACK_VERSION=v0.6.0
+
+docker build . --build-arg PACK_VERSION=${PACK_VERSION} -t gcr.io/k8s-skaffold/pack:${PACK_VERSION} -t gcr.io/k8s-skaffold/pack:latest
+docker push gcr.io/k8s-skaffold/pack:${PACK_VERSION}
+docker push gcr.io/k8s-skaffold/pack:latest
--- a/pkg/skaffold/constants/constants.go
+++ b/pkg/skaffold/constants/constants.go
@@ -62,6 +62,7 @@ const (
 	DefaultCloudBuildMavenImage  = "gcr.io/cloud-builders/mvn"
 	DefaultCloudBuildGradleImage = "gcr.io/cloud-builders/gradle"
 	DefaultCloudBuildKanikoImage = "gcr.io/kaniko-project/executor"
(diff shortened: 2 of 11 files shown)


examples/buildpacks/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM alpine:3.10
+ENV PACK_VERSION 0.6.0
+ENV PACK_URL https://github.com/buildpack/pack/releases/download/v${PACK_VERSION}/pack-v${PACK_VERSION}-linux.tgz
+RUN wget -O pack.tgz "${PACK_URL}"
+RUN tar -zxf pack.tgz
+RUN mv pack /usr/local/bin
```
briandealwis: This file seems to be here in error

dgageot: Let me fix that

pkg/skaffold/constants/constants.go
```diff
@@ -62,6 +62,7 @@ const (
 	DefaultCloudBuildMavenImage  = "gcr.io/cloud-builders/mvn"
 	DefaultCloudBuildGradleImage = "gcr.io/cloud-builders/gradle"
 	DefaultCloudBuildKanikoImage = "gcr.io/kaniko-project/executor"
+	DefaultCloudBuildPackImage   = "gcr.io/k8s-skaffold/pack"
```
tejal29: nit: can we move this gcb/buildpacks.go as a  local constant if there are cyclic dependencies issues?

deploy/buildpacks/publish.sh
```diff
@@ -0,0 +1,23 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+PACK_VERSION=0.6.0
```
tejal29: usually, folks end up copying the version string from `pack --version` output
```
 ~/Downloads/pack  --version
v0.6.0 (git sha: 109b629d388cec0ed3836b9fed6717727a9187c1)
tejaldesai-macbookpro2:~ tejaldesai$ 
```
Do you want to add the prefix `v` to the version number?

tejal29: Also follow the same `v0.0.6` version convention for the skaffold gcb image tag?

deploy/buildpacks/publish.sh
```diff
@@ -0,0 +1,23 @@
...
+
+docker build . --build-arg PACK_VERSION=${PACK_VERSION} -t gcr.io/k8s-skaffold/pack:${PACK_VERSION} -t gcr.io/k8s-skaffold/pack:latest
+docker push gcr.io/k8s-skaffold/pack:${PACK_VERSION}
+docker push gcr.io/k8s-skaffold/pack:latest
```
tejal29: Does heroku publish a container image that we can re-use or any plans in future to do so? Or else, skaffold team will have to keep the track of publishing the new image everytime pack is upgraded (which is fine for now)

deploy/buildpacks/publish.sh
dgageot: I'm fixing that. Thank you!

deploy/buildpacks/publish.sh
dgageot: I haven't seen such plan.

pkg/skaffold/constants/constants.go
dgageot: I'll move all those default image names next to where they are used in another PR

PR #3511, opened 2020-01-15
Allow 2020 copyright year

loosebazooka:
Allows `2019|2020`.

Previously when only 2019 was used, it was a hard requirement that all files use 2019, so any new file added would require a 2019 copyright header.

With this change, there will not be a way to ensure that new files use 2020 except by human :eyes: but we preserve the original copyright year on files that were created earlier than 2020.


balopat:
Thanks for this :) yeah, I think that's fine / I can modify the script to ensure year of creation = today's year. 
Last year I kinda messed this up and updated the copyright year to 2019 on all files, I might fix that with this change, as I learned that the policy is to keep the copyright year the creation year.

PR #3514, opened 2020-01-16
[doc] There’s no `gcsBucket` config anymore

dgageot:


dgageot:
@balopat it's removed in a compatible way. Nothing will fail if people use it, it'll just sync from local files instead.

PR #3521, opened 2020-01-17
breaking apart init logic

balopat:
This PR breaks apart the currently 700 LOC file to smaller files by responsibility area. 
There is no functional change or refactoring, well, only a single error message change: 

Instead of 
```
FATA[0000] pre-existing deploy/webhook/skaffold.yaml found 
```

Now the error will suggest `--force`

```
FATA[0000] pre-existing deploy/webhook/skaffold.yaml found (you may continue with --force)
```


balopat:
I decided to work a bit more on the feature I'm trying to do and then break it into smaller PRs.

balopat:
oh, @dgageot thanks for the approval - do you think this in itself is worthwhile?

dgageot:
@balopat Yes, this piece of code is really hard to reason about

balopat:
cool, I'll keep refactoring it a bit more then, I agree

PR #3527, opened 2020-01-17
Update CLI tools in docker image

nicknovitski:
Hopefully this is a fairly self-explanatory change.

### DIFF
--- a/deploy/skaffold/Dockerfile.deps
+++ b/deploy/skaffold/Dockerfile.deps
@@ -14,7 +14,7 @@
 
 # Download kubectl
 FROM alpine:3.10 as download-kubectl
-ENV KUBECTL_VERSION v1.12.8
+ENV KUBECTL_VERSION v1.14.10
 ENV KUBECTL_URL https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl
 RUN wget -O kubectl "${KUBECTL_URL}"
 RUN chmod +x kubectl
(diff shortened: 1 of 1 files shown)


nicknovitski:
@googlebot I signed it!

nicknovitski:
Is the kokoro check supposed to finish in finite time?

deploy/skaffold/Dockerfile
```diff
@@ -14,35 +14,35 @@
...
-ENV KUBECTL_VERSION v1.12.8
+ENV KUBECTL_VERSION v1.17.1
 ENV KUBECTL_URL https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl
 RUN wget -O kubectl "${KUBECTL_URL}"
 RUN chmod +x kubectl
 
 # Download helm
 FROM alpine:3.10 as download-helm
-ENV HELM_VERSION v2.12.0
+ENV HELM_VERSION v2.16.1
```
dgageot: Our integration tests are not prepared for this change yet. Could you not update helm in this PR until we fix our test infrastructure?

deploy/skaffold/Dockerfile
```diff
@@ -14,35 +14,35 @@
 
 # Download kubectl
 FROM alpine:3.10 as download-kubectl
-ENV KUBECTL_VERSION v1.12.8
+ENV KUBECTL_VERSION v1.17.1
```
dgageot: I don't think we want to jump directly to the latest version of kubectl. Only a few users will have this version installed and it's not guaranteed to be compatible with every older version of k8s. ping @nkubala @balopat

nicknovitski: That makes sense.  How about 1.14?

It's also possible to install kubectl as a google cloud sdk component, which will provide a consistent version without needing it to be specified separately from the cloud sdk version.  Currently the sdk provides version 1.14.x of `kubectl`.  It also includes binaries like `kubectl.1.1{3,4,5}`, but I don't see a way for skaffold to use those.

deploy/skaffold/Dockerfile
nicknovitski: Sure thing.

nicknovitski:
Ah, #3433 was just merged, which puts these deps into a separate image that has to be manually built, pushed, and referenced.
```
# This base image has to be updated manually after running `make build_deps`
FROM gcr.io/k8s-skaffold/build_deps:6f7f8095b663b0c9c18873855f55fee20aa06a7e
``` 
I can't push the image, so I can't really complete this contribution. :/

dgageot:
If that’s fine by you, I can cherry pick your commits and push a new PR.

nicknovitski:
Thank you, that would be fine with me.

dgageot:
closing in favor of https://github.com/GoogleContainerTools/skaffold/pull/3553

balopat:
@nicknovitski do you mind writing a note on the #3553 PR to confirm your consent on @dgageot cherry picking your commits into that PR?

PR #3534, opened 2020-01-20
Update golangci-lint

dgageot:
Also make sure we alway use the version we’ve just installed.


hack/golangci.yml
```diff
@@ -7,16 +7,22 @@ linters:
   enable:
     - bodyclose
     - deadcode
+    - dogsled
```
loosebazooka: Dogsled twice

hack/golangci.yml
```diff
@@ -7,16 +7,22 @@ linters:
...
     - goimports
     - golint
+    - goprintffuncname
     - gosimple
     - govet
     - ineffassign
     - interfacer
     - maligned
     - misspell
+    - nakedret
```
loosebazooka: Nakedret twice

PR #3536, opened 2020-01-20
Be pedantic with imports

dgageot:


hack/pedantic-imports.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Skaffold Authors
```
loosebazooka: 2020

balopat: great catch - I still owe with the updated script :D

PR #3548, opened 2020-01-21
Add verbosity flag to go tests on travis

loosebazooka:


tejal29:
The flag only got applied to windows tests. Is there a reason for adding the `-v` flag ?

loosebazooka:
AFAIK, the `-v` flag prints out the reason the test failed (rather than just saying the test failed), this is nice for people who cannot mimic the test environment easy (ex: I do not have a mac to test mac failures on).

tejal29:
> AFAIK, the `-v` flag prints out the reason the test failed (rather than just saying the test failed), this is nice for people who cannot mimic the test environment easy (ex: I do not have a mac to test mac failures on).
> 
> It should apply to all test environments.

Ahh got it. Thanks!

tejal29:
The linux job still does not have `-v` flag
https://travis-ci.com/GoogleContainerTools/skaffold/jobs/278134461#L236
Rest all (windows, OSX and integration test) have `-v` flag.

tejal29:
Can you also add `-v` to coverage target used for linux profile.
https://github.com/GoogleContainerTools/skaffold/blob/master/Makefile#L111

loosebazooka:
Do we want it on coverage right? I put it on https://github.com/GoogleContainerTools/skaffold/blob/master/Makefile#L106 instead.

loosebazooka:
oh no you're right, needs to go on coverage.

PR #3542, opened 2020-01-20
[init refactor] organize tests

balopat:
This should come after #3538 is merged, it's built on top of that. 

- It reorganizes the tests from init_test.go to the relevant file's tests.
- adds tests for `stripTags` that is a new method introduced as I'm starting to take `DoInit` apart

I know...this is a bit bigger but I am close to getting to a sensible organization of the code. 
Also, after merging #3538 this should become smaller.

PR #3550, opened 2020-01-21
[init refactor] doinit tests

balopat:
should come after #3542. 

Converted (non-kompose) init integration tests to unit tests with benefits of
 * more insight into coverage of `DoInit` 
 * faster feedback (make test)
 * makes our test pyramid healthier - there should always be more unit tests than integration tests
 * adds Windows and Mac testing for `skaffold init`


pkg/skaffold/initializer/kubectl.go
```diff
@@ -46,6 +47,8 @@ type kubectlAnalyzer struct {
 
 func (a *kubectlAnalyzer) analyzeFile(filePath string) error {
 	if IsKubernetesManifest(filePath) && !isSkaffoldConfig(filePath) {
+		// to make skaffold.yaml more portable across OS-es we should generate always / based filePaths
+		filePath = strings.ReplaceAll(filePath, string(os.PathSeparator), "/")
```
tejal29: This is causing the failure on windows since the tests expect the `\` as file path separator.
How will this work when a user on window uses `skaffold deploy` and the generated path has `/` instead of expected os.PathSeparator?

tejal29: like we spoke in person. "/" works on windows. resolving this comment

PR #3555, opened 2020-01-22
Auto sync with Buildpacks

dgageot:
This adds automatic sync configuration for buildpacks.

Buildpacks can set special metadata on the image to indicate which files need to be synced.

More work is needed on this feature but I'd like this first step to be reviewed and merged.

### DIFF
--- a/pkg/skaffold/build/buildpacks/types.go
+++ b/pkg/skaffold/build/buildpacks/types.go
@@ -22,12 +22,14 @@ import "github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker"
 type Builder struct {
 	localDocker docker.LocalDaemon
 	pushImages  bool
+	devMode     bool
 }
 
 // NewArtifactBuilder returns a new buildpack artifact builder
-func NewArtifactBuilder(localDocker docker.LocalDaemon, pushImages bool) *Builder {
+func NewArtifactBuilder(localDocker docker.LocalDaemon, pushImages, devMode bool) *Builder {
 	return &Builder{
 		localDocker: localDocker,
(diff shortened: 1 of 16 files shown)


pkg/skaffold/build/buildpacks/types.go
```diff
@@ -22,12 +22,14 @@ import "github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker"
 type Builder struct {
 	localDocker docker.LocalDaemon
 	pushImages  bool
+	devMode     bool
```
balopat: why did you decide to put devMode live on the Builder and not on the artifact configuration?

balopat: Hmmm... I got confused and thought that "devMode" is specific to buildpacks, however it just signifies that we started `skaffold dev`. You can ignore my comment. Something around how we wire through our code, when we introduce new flags / options from runcontext still rubs me the wrong way.

pkg/skaffold/runner/runcontext/context.go
```diff
@@ -70,13 +71,17 @@ func GetRunContext(opts config.SkaffoldOptions, cfg latest.Pipeline) (*RunContex
 		insecureRegistries[r] = true
 	}
 
+	// TODO(dgageot): what about debug?
```
balopat: currently debug does not do sync. Also during debugging not sure how the restart would impact the session - I guess they would reconnect...but my gut feeling is that we can safely ignore this for now for debug.

dgageot: Thanks, I'll remove the comment!

PR #3556, opened 2020-01-22
Prepare for support for other types of pod

dgageot:
I'm only renaming a few functions


dgageot:
Let's self merge this one. I only renamed a few functions. No risk of breaking anything

PR #3562, opened 2020-01-22
golangci-lint 1.23.1 + autoupgrade

balopat:


hack/linter.sh
```diff
@@ -18,9 +18,16 @@ set -e -o pipefail
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+VERSION=1.23.1
+
+if [[ $(golangci-lint --version | grep $VERSION | wc -l) -eq 0 ]]; then
```
dgageot: This fails when `golangci-lint` is not yet installed

balopat: 🤦‍♂

dgageot: `golangci-lint --version | grep -c " $VERSION "` is more compact and it checks for the exact version, not a substring because of the spaces around `$VERSION`

balopat: Thank you for the suggestion! done!

PR #3564, opened 2020-01-22
Encode artifact image-name and container WORKDIR in container debug info

briandealwis:
Relates to #3122 

Should merge before: xxx

**Description**

`debug` annotates podspecs with debugging configuration information about its debuggable containers.  This debug configuration information was being passed around as map, but it was becoming a bit unwieldy. This PR:
  1. Turns this map into an object (`ContainerDebugConfiguration`).
  2. Turns the debugging ports into a proper map, instead of just being inlined to the config map.
  3. Extracts and encodes the corresponding artifact's image-name and the container working directory into this new `ContainerDebugConfiguration`.

One side-effect is that the serialized JSON doesn't sort the same.

The image-name and working-directory is information requested by the IDEs in #3122. 

**User facing changes**

The pod annotation format for the debug configuration information has changed slightly.  The IDEs do not currently use this information.  Otherwise there should be no visible changes.

**Next PRs.**

**Submitter Checklist**

review them:

**Reviewer Notes**

- [ ] Unit test added.

**Release Notes**

I don't think this change is worth mentioning since there are no known consumers.

### DIFF
--- a/pkg/skaffold/debug/debug_test.go
+++ b/pkg/skaffold/debug/debug_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package debug
 
 import (
+	"strings"
 	"testing"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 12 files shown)


pkg/skaffold/debug/debug_test.go
```diff
@@ -114,14 +115,14 @@ func (t testTransformer) RuntimeSupportImage() string {
 	return ""
 }
 
-func (t testTransformer) Apply(container *v1.Container, config imageConfiguration, portAlloc portAllocator) map[string]interface{} {
+func (t testTransformer) Apply(container *v1.Container, config imageConfiguration, portAlloc portAllocator) *ContainerDebugConfiguration {
```
dgageot: Can you return a value rather than a pointer?

briandealwis: I think this is a vestige prior to introducing `IsApplicable()`.  I'll do that as a separate PR if that's ok.

PR #3576, opened 2020-01-24
Automatically handle —no-pull option on `pack`.

dgageot:
Instead of having the user handle that, Skaffold will always pull images on the first build and on each subsequent builds, pack will be told to not pull.


pkg/skaffold/build/buildpacks/pulled_images.go
```diff
@@ -0,0 +1,53 @@
...
+	defer p.lock.Unlock()
+
+	if p.images == nil {
+		p.images = map[builderRunnerPair]bool{}
+	}
+
+	p.images[builderRunnerPair{
+		builder: builder,
+		runner:  runImage,
+	}] = true
```
briandealwis: This was a bit hard to parse!

dgageot:
kokoro is failing for unrelated reason (TestBuildKanikoInsecureRegistry)

pkg/skaffold/build/buildpacks/pulled_images.go
dgageot: I made that clearer

briandealwis:
One side-effect is that it makes it difficult to try out builder changes locally.

PR #3579, opened 2020-01-24
init/manifest flag

balopat:
This should come after #3577. 

This PR introduces the `--kubernetes-manifest` flag on `skaffold init`. 
This flag: 
- turns off k8s manifest detection 
- creates the deploy section with the list of manifests passed in to this flag

It is meant to be the counterpart of `--artifact` that similarly enables the user to specify a basic pair of image name + builder config. Also this is required by the IDEs for a generation flow, where the k8s manifest is going to be created by another tool - the IDE should be able to generate a skaffold.yaml that has the builders specified in `--artifact` and the manifests would point to a not-yet existing file.


cmd/skaffold/app/cmd/init.go
```diff
@@ -52,6 +53,7 @@ func NewCmdInit() *cobra.Command {
 			f.BoolVar(&force, "force", false, "Force the generation of the Skaffold config")
 			f.StringVar(&composeFile, "compose-file", "", "Initialize from a docker-compose file")
 			f.StringArrayVarP(&cliArtifacts, "artifact", "a", nil, "'='-delimited Dockerfile/image pair, or JSON string, to generate build artifact\n(example: --artifact='{\"builder\":\"Docker\",\"payload\":{\"path\":\"/web/Dockerfile.web\"},\"image\":\"gcr.io/web-project/image\"}')")
+			f.StringArrayVarP(&cliKubectlManifests, "kubectl-manifest", "k", nil, "list of predefined kubectl manifests (overrides detection)")
```
sujit-kamireddy: Any particular reason why its called kubectl-manifest instead of kubernetes-manifest?

balopat: Because the deployer is called kubectl - there could be `kustomization-file` and `helm-releases`... But I agree that this might be confusing / looks janky off the bet. I think I am okay to change it to kubernetes-manifest.

cmd/skaffold/app/cmd/init.go
```diff
@@ -52,6 +53,7 @@ func NewCmdInit() *cobra.Command {
 			f.BoolVar(&force, "force", false, "Force the generation of the Skaffold config")
 			f.StringVar(&composeFile, "compose-file", "", "Initialize from a docker-compose file")
 			f.StringArrayVarP(&cliArtifacts, "artifact", "a", nil, "'='-delimited Dockerfile/image pair, or JSON string, to generate build artifact\n(example: --artifact='{\"builder\":\"Docker\",\"payload\":{\"path\":\"/web/Dockerfile.web\"},\"image\":\"gcr.io/web-project/image\"}')")
+			f.StringArrayVarP(&cliKubernetesManifests, "kubernetes-manifest", "k", nil, "list of predefined kubectl manifests (overrides detection)")
```
tejal29: Can we add more explanation on if it can accept a file glob, or a directory ?
Do they need to be valid manifests?
```
list of predefined valid kubectl manifests (overrides detection) e.g. deploy/*.yaml, k8s/ or k8s/pod.yaml"
```

pkg/skaffold/initializer/deploy_init.go
```diff
@@ -0,0 +1,55 @@
...
+
+type cliDeployInit struct {
+	cliKubernetesManifests []string
+}
+
+func (c *cliDeployInit) deployConfig() latest.DeployConfig {
+	return latest.DeployConfig{
+		DeployType: latest.DeployType{
+			KubectlDeploy: &latest.KubectlDeploy{
+				Manifests: c.cliKubernetesManifests,
```
tejal29: Maybe we should validate if these exists before generating the yaml

cmd/skaffold/app/cmd/init.go
balopat: I'll add more explanation. No, they don't have to be valid kubectl manifests, they can be just a placeholder that is created later.

pkg/skaffold/initializer/deploy_init.go
balopat: I don't want to validate, I am thinking about these as they can be just placeholders...

PR #3589, opened 2020-01-27
Fix invalid package comments

dgageot:


dgageot:
Self merging this trivial PR

PR #3594, opened 2020-01-27
Prevent nil pointer dereference

dgageot:
Not sure how this can happen but this has happened.

I will try to reproduce and add tests around that code.


dgageot:
@balopat yes, you're right.

dgageot:
Closing in favour of #3597

PR #3600, opened 2020-01-28
Update pack dependency

dgageot:


go.sum
```diff
@@ -13,16 +13,22 @@ cloud.google.com/go v0.46.3/go.mod h1:a6bKKbmY7er1mI7TEI4lsAkts/mkhTSZK8w33B4RAg
 cloud.google.com/go v0.49.0 h1:CH+lkubJzcPYB1Ggupcq0+k8Ni2ILdG2lYjDIgavDBQ=
 cloud.google.com/go v0.49.0/go.mod h1:hGvAdzcWNbyuxS3nWhD7H2cIJxjRRTRLQVB0bdputVY=
 cloud.google.com/go/bigquery v1.0.1/go.mod h1:i/xbL2UlR5RvWAURpBYZTtm/cXjCha9lbfbpx4poX+o=
+cloud.google.com/go/bigquery v1.2.0 h1:5wdyDJgqmqlAqzIhN6hY9E2vqE9M545UVLU1rrSr/x4=
```
briandealwis: There seem to be a lot of seemingly-spurious changes here.  Any idea why these weren't pulled in previously?

(And are we really pulling in bigquery v1.0.1 and v1.2.0?  I wonder what pulls in v1.2.0, as it's marked as indirect but I don't see any go.mods pulling it in.)

dgageot: That's the magic of `go.sum`. It contains multiple versions of the same dep. Also sometimes, deps appear in there just like that. It has of course no impact on what's vendored.
This is hopeless to try to understand.

briandealwis: 😭

PR #3607, opened 2020-01-29
Better output for Docker commands

dgageot:
Better output for docker build|push|pull|load


dgageot:
@balopat sorry. It's supposed to behave like the docker CLI does. Instead of printing information on new lines, it'll update lines in place. If the output is not a terminal, then it'll behave like before.

I remember trying to activate that a couple of times and it wasn't working on Matt's machine and your machine. The code I used was a bit different. I hope it works now. Can you try pushing an image?

balopat:
Nice! 

Instead of this: 

image

I have the arrows flowing:

image

PR #3628, opened 2020-01-31
customizable jib feature minimum requirements

loosebazooka:
also updates the jib sync search string

### DIFF
--- a/pkg/skaffold/build/jib/gradle.go
+++ b/pkg/skaffold/build/jib/gradle.go
@@ -38,6 +38,7 @@ var (
 
 // Skaffold-Jib depends on functionality introduced with Jib-Gradle 1.4.0
 const MinimumJibGradleVersion = "1.4.0"
+const MinimumJibGradleVersionForSync = "2.0.0"
 
 // GradleCommand stores Gradle executable and wrapper name
 var GradleCommand = util.CommandWrapper{Executable: "gradle", Wrapper: "gradlew"}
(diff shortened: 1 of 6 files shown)


pkg/skaffold/build/jib/gradle.go
```diff
@@ -87,18 +88,18 @@ func getDependenciesGradle(ctx context.Context, workspace string, a *latest.JibA
...
 
 func getCommandGradle(ctx context.Context, workspace string, a *latest.JibArtifact) exec.Cmd {
-	args := append(gradleArgsFunc(a, "_jibSkaffoldFilesV2"), "-q")
+	args := append(gradleArgsFunc(a, "_jibSkaffoldFilesV2", MinimumJibGradleVersion), "-q")
 	return GradleCommand.CreateCommand(ctx, workspace, args)
 }
 
 func getSyncMapCommandGradle(ctx context.Context, workspace string, a *latest.JibArtifact) *exec.Cmd {
-	cmd := GradleCommand.CreateCommand(ctx, workspace, gradleBuildArgsFunc("_jibSkaffoldSyncMap", a, true))
+	cmd := GradleCommand.CreateCommand(ctx, workspace, gradleBuildArgsFunc("_jibSkaffoldSyncMap", a, true, MinimumJibMavenVersionForSync))
```
tejal29: would this mean,  a user will use 2 different `jib-maven` or `jib-gradle` version for in one skaffold run if they have a sync configured?

Can we instead, 
1. Check if sync is configured for any jib artifacts. 
2. If yes, Use `MinimumJibGradleVersionForSync` for all actions for all jib artifacts in one skaffold run.
3. If not use `1.4.0` version?

loosebazooka: It just means, that if they use sync features, they will be asked to use 2.0.0+, otherwise it will only require 1.4.0+

loosebazooka: this is on a per artifact basis.

loosebazooka: so old configs will continue to work (we wont force users to move to 2.0.0)

tejal29: sure. 

My concern is - (also due to lack of understanding of maven)
1. Can you two version of `jib-maven` plugin installed locally?
If yes, 
say on my local machine i have 1.4.1 and 2.0.0 installed, 
1. `getCommandGradle` will specify mininum version as `1.4.*` when calculating dependencies 
2. `getSyncMapCommandGradle` will use 2.0.0 when syncing files. 
In this scenario would two separate version of `jib` be used for the same artifact?
Can we instead force the same minimum gradle version?
In that case, an artifact with `sync` configured won't build at all until user upgrades to 2.0.0.

loosebazooka: Oh I see what you're asking: This doesn't actually set the version of jib being used. It just enforces a minimum version. The user configures the jib version in their build file.

PR #3640, opened 2020-02-03
Fix bug related to new kustomization format. Fixes #2909

### LINKED ISSUE #2909: `skaffold dev` breaks on kustomize new `patches:` field format
### Actual behavior

Running `skaffold dev` with a kustomization.yaml file containing the `patches:` field to target multiple objects (https://github.com/kubernetes-sigs/kustomize/blob/master/docs/plugins/builtins.md#field-name-patches) fails with:
```
FATA[0000] watching files for deployer: listing files: listing files: yaml: unmarshal errors:
  line xx: cannot unmarshal !!map into string
```
In the skaffold kustomize example (https://github.com/GoogleContainerTools/skaffold/blob/master/examples/kustomize/kustomization.yaml) it is clear that skaffold expects the "old" format of defining patches just by filename:

``` yaml
patches: 
- file.yml
- another-file.yml
```

The new format is more elaborate defining targets and allows for inline patches and can look like this:
```yaml
# from https://github.com/kubernetes-sigs/kustomize/blob/master/docs/plugins/builtins.md#field-name-patches
patches:
- path: patch.yaml
  target:
    group: apps
    version: v1
    kind: Deployment
    name: deploy.*
    labelSelector: "env=dev"
    annotationSelector: "zone=west"
- patch: |-
    - op: replace
      path: /some/existing/path
      value: new value
  target:
    kind: MyKind
    labelSelector: "env=dev"       
```

Running `skaffold run` or `skaffold deploy` strangely **works** and produces valid yaml.
Running `kustomize` directly works and produces valid yaml as well.

### Expected behavior

`skaffold dev` with a kustomization target is expected not to break on usage of the new patches field format. 

### Information

- Skaffold version: v0.38.0
- Kustomize version: v3.2.0
- Operating system: macOS

zasgar:
Should go after #3635 
Fixes #2909

**Description**
The new kustomization format has file paths listed under "path", instead of directly under "patches".

**User facing changes**

Older format will cause an error. 
**Submitter Checklist**

review them:

**Reviewer Notes**

- [ ] Unit test added.

**Release Notes**

- Support for new Kustomization "patches" format.


zasgar:
@tejal29 will it be helpful if we support both the old/new formats for Kustomize? It's pretty straightforward to do and I can the patch to work with both cases.

tejal29:
@zasgar #3663 is in and this is no longer required.

PR #3651, opened 2020-02-06
Disable all colors in Buildpacks’s output when not in a terminal

dgageot:
+ Make sure the output from pack and from buildpacks goes to stdout
 + Make sure all colors are disabled when not printing to a terminal

### DIFF
--- a/pkg/skaffold/build/buildpacks/lifecycle.go
+++ b/pkg/skaffold/build/buildpacks/lifecycle.go
@@ -22,14 +22,12 @@ import (
 	"strings"
 
 	"github.com/buildpacks/pack"
-	"github.com/heroku/color"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/misc"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
-	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 )
 
 // For testing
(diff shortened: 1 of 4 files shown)


pkg/skaffold/build/buildpacks/logger.go
```diff
@@ -30,12 +33,27 @@ type logger struct {
...
 
 func NewLogger(out io.Writer) logging.Logger {
+	// If out is not a terminal, let's make sure no colors are printed.
+	if _, isTerm := util.IsTerminal(out); !isTerm {
+		out = colorable.NewNonColorable(out)
+	}
+
+	l := logrus.New()
+	l.SetOutput(out)
+	l.SetFormatter(new(plainFormatter))
```
briandealwis: I'd appreciate a comment here as to why this formatter is required

pkg/skaffold/build/buildpacks/lifecycle.go
```diff
@@ -85,12 +83,6 @@ func (b *Builder) build(ctx context.Context, out io.Writer, a *latest.Artifact,
 }
 
 func runPackBuild(ctx context.Context, out io.Writer, localDocker docker.LocalDaemon, opts pack.BuildOptions) error {
-	// If out is not a terminal, let's make sure pack doesn't output with colors.
-	if _, isTerm := util.IsTerminal(out); !isTerm {
-		// pack uses heroku/color under the hood.
-		color.Disable(true)
```
briandealwis: Oh, I guess a problem here is that we never re-enable color.

pkg/skaffold/build/buildpacks/logger.go
dgageot: Added

pkg/skaffold/build/buildpacks/lifecycle.go
dgageot: Yes, the new code is more robust

PR #3660, opened 2020-02-06
a hidden flag for simpler access to new init format

### LINKED ISSUE #3655: FR: Flag for new skaffold init --analyze format
I'd love to be able to opt into the new `skaffold init --analyze` format _without_ opting into either jib or buildpacks support. Right now, migrating to the new format is coupled with adding new behavior, which makes staging these changes tough.

## Current behavior
```
$ skaffold init --analyze | jq .
{
  "dockerfiles": [
    "Dockerfile"
  ],
  "images": [
    "nodejs-hello-world"
  ]
}
```

```
$ SKAFFOLD_XXENABLEJIBINIT=1 skaffold init --analyze | jq .
{
  "builders": [
    {
      "name": "Docker",
      "payload": {
        "path": "Dockerfile"
      }
    }
  ],
  "images": [
    {
      "name": "nodejs-hello-world",
      "foundMatch": false
    }
  ]
}
```

## Desired behavior

Something like --XXenableNewInit would enable this:
```
$ SKAFFOLD_XXENABLENEWINIT=1 skaffold init --analyze | jq .
{
  "builders": [
    {
      "name": "Docker",
      "payload": {
        "path": "Dockerfile"
      }
    }
  ],
  "images": [
    {
      "name": "nodejs-hello-world",
      "foundMatch": false
    }
  ]
}
```

balopat:
Fixes #3655.

### DIFF
--- a/pkg/skaffold/initializer/init.go
+++ b/pkg/skaffold/initializer/init.go
@@ -61,9 +61,9 @@ type Config struct {
 	Analyze                bool
 	EnableJibInit          bool // TODO: Remove this parameter
 	EnableBuildpacksInit   bool
+	EnableNewInitFormat    bool
 	BuildpacksBuilder      string
 	Opts                   config.SkaffoldOptions
-	EnableNewInitFormat    bool
 }
 
 // builderImagePair defines a builder and the image it builds
--- a/pkg/skaffold/initializer/init.go
+++ b/pkg/skaffold/initializer/init.go
@@ -63,6 +63,7 @@ type Config struct {
 	EnableBuildpacksInit   bool
 	BuildpacksBuilder      string
 	Opts                   config.SkaffoldOptions
+	EnableNewInitFormat    bool
 }
 
 // builderImagePair defines a builder and the image it builds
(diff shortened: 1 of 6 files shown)


pkg/skaffold/initializer/init.go
```diff
@@ -63,6 +63,7 @@ type Config struct {
 	EnableBuildpacksInit   bool
 	BuildpacksBuilder      string
 	Opts                   config.SkaffoldOptions
+	EnableNewInitFormat    bool
```
jonjohnsonjr: > > pkg/skaffold/initializer/init.go:54:13: struct of size 400 bytes could be of size 392 bytes (maligned)
type Config struct {

This linter is pedantic! I love it! You can save 4 bytes by moving this up with the other `bool`s.

balopat: oh yeah, right, this always gets me

tejal29:
In future, do you have plans to mark `XXenableJibInit` and `XXenableBuildpacksInit` as deprecated and use the new flag instead and convey it to IDEs?

balopat:
> In future, do you have plans to mark `XXenableJibInit` and `XXenableBuildpacksInit` as deprecated and use the new flag instead and convey it to IDEs?
> 
> Edit:
> Looks like they are used later in `detectBuilder` for other purpose. There are too many flags now and we should condense them.

yeah, they are used separately. 
I expect these all to disappear soon!

Issue #1950, opened 2019-04-12
Skaffold sync failure


prary:
I am running skaffold 24 version. I am using skaffold sync(with sync command same as in version 26), its failing with exit status 2.

### Expected behavior
I should be able to directly sync the modified file to running pod.

### Actual behavior
Returning exit code 2.

level=warning msg="Skipping deploy due to sync error: copying files: exit status 2"

### Information

- Skaffold version: 24
- Operating system: Linux
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta6
kind: Config
build:
 artifacts:
 - image: someurl.com/kube/pythonserver
 context: .
 sync:
 '*.py': .
 kaniko:
 buildContext:
 localDir: {}
 namespace: prashant
 image: someurl.com/gkp/kaniko-executor:latest dockerConfig:
 path: C:\Users\prashant\.docker\config.json
deploy:
 kubectl:
 manifests:
 - deploy.yml
```

### Steps to reproduce the behavior

1. Make a sample hello world python application
2. Create dockerfile for the same, build and deploy the application using kaniko and skaffold in dev mode.
3. Change Hello world to Hello New world in python application.
4. Skaffold would catch the changes made and would try to copy the file to running pod.
5. Skaffold fails to copy the changed file.

Here is the Dockerfile for my application

```
FROM someurl.net/lrh7:latest
RUN yum install tar -y
COPY app.py .
EXPOSE 8080
CMD [ "python", "app.py" ]

```

https://groups.google.com/d/msg/skaffold-users/nGPcn6uypq8/oBciDgE5CQAJ

briandealwis:
@prary please try a later version of Skaffold: I think you're hitting #1721 which was fixed with #1722 for v0.25.0.

On the mailing list, @prary showed:
```
time="2019-04-11T10:26:34Z" level=debug msg="Running command: [kubectl exec simplehttp-6bffcd784c-dt8xq --namespace prashant -c simplehttp -i -- tar xmf - --no-same-owner]"
time="2019-04-11T10:26:36Z" level=warning msg="Skipping deploy due to sync error: copying files: exit status 2"
```

Note that the `tar` is missing `-C /`.

prary:
Hi brain,
I have added that missing change also but still facing the issue, in some time from now I will share the screenshot for the same.

prary:
Hi @briandealwis 
here are the running logs attached.


[...]

PR #3680, opened 2020-02-12
Allow Sync for non-root containers-hotreload example

ibidani:
When deploying the hot-reload example to non-root containers, skaffold dev sync functionality fails with "permissions denied"
There has been multiple open issues referring to this error: #2041 #1950 
This change will ensure examples will be compatible with more clusters with higher security requirements.

**Description**
chowning the target folder with the proper runtime UID so that tar xmf (Sync) command will not fail

**User facing changes**

Write n/a if not output or log lines changed and no behavior is changed

**Before**
The following error cases the Sync on save to break
```
WARN[0024] Skipping deploy due to sync error: copying files: Running [kubectl --context <context> exec node-56846577b4-c878r --namespace <namespace> -c node -i -- tar xmf - -C / --no-same-owner]
 - stdout: 
 - stderr: tar: removing leading '/' from member names
tar: can't remove old file app/index.js: Permission denied
command terminated with exit code 1
: exit status 1 
```
**After**
skaffold dev sync is working as expected on both node and python examples

**Next PRs.**

**Submitter Checklist**

review them:

**Reviewer Notes**

**Release Notes**
- hot-reload example skaffold dev sync working on non-root containers

### DIFF
--- a/examples/hot-reload/k8s/node.yaml
+++ b/examples/hot-reload/k8s/node.yaml
@@ -22,6 +22,8 @@ spec:
       labels:
         app: node
     spec:
+      securityContext:
+        runAsUser: 1000
       containers:
       - name: node
         image: node-example
--- a/examples/hot-reload/k8s/python.yaml
+++ b/examples/hot-reload/k8s/python.yaml
@@ -22,6 +22,8 @@ spec:
       labels:
         app: python
     spec:
+      securityContext:
+        runAsUser: 1000
       containers:
       - name: python
         image: python-reload
--- a/examples/hot-reload/node/Dockerfile
+++ b/examples/hot-reload/node/Dockerfile
@@ -5,5 +5,5 @@ EXPOSE 3000
 CMD ["npm", "run", "dev"]
 
 COPY package* ./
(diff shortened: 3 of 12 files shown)


ibidani:
Oops, I'll update the integration files today

ibidani:
I updated the integration files
Thanks @dgageot

ibidani:
Thanks Good suggestion, I'll update the code tonight

ibidani:
@dgageot I just realized this PR inspired you to open #3683 which fixes this issue. 
Should I close this PR?

dgageot:
I don’t think so. I’d love to see all our examples better represent real projects. I didn’t want to step on your feet and changed a different sample.

ibidani:
I've backported the great changes in #3683 into hot-reload example, per my tests on unprivileged cluster securityContext was necessary for the hot-load to succeed without the "tar: can't remove old file app/index.js: Permission denied" error mentioned above.
I've learned so much, thanks @dgageot 
```
securityContext:
    runAsUser: 1000
```

PR #3691, opened 2020-02-15
hidden --minikube-profile flag

balopat:
This is a tactical move until we fix properly https://github.com/GoogleContainerTools/skaffold/issues/3668. 

It adds a hidden --minikube-profile flag that forces skaffold to use the given profile's docker daemon for building, and hence will default to not pushing.  

(Apologies for no tests.)

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -261,6 +261,15 @@ var FlagRegistry = []Flag{
 		FlagAddMethod: "StringVar",
 		DefinedOn:     []string{"build", "debug", "dev", "run"},
 	},
+	{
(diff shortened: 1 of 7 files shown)


cmd/skaffold/app/cmd/flags.go
```diff
@@ -261,6 +261,15 @@ var FlagRegistry = []Flag{
 		FlagAddMethod: "StringVar",
 		DefinedOn:     []string{"build", "debug", "dev", "run"},
 	},
+	{
+		Name:          "minikube-profile",
+		Shorthand:     "m",
```
dgageot: nit: Do we really need a shorthand for a hidden flag?

PR #3724, opened 2020-02-24
Faster make v2

dgageot:
This solves multiple slowness issues with the current Makefile:
 + Some variables are eagerly evaluated even if they are not used by the current target.
 + `make generate-statik` should make its inputs and outputs  explicit so that we can avoid the cost of regenerating

The net result is that `make` and `make install` will be faster to find out that nothing needs to be rebuilt and will be faster to rebuild changes to the Go code.

@nkubala FYI, this is a fixed version of #3706

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,9 +46,6 @@ endif
 export GO111MODULE = on
 export GOFLAGS = -mod=vendor
 
-GO_GCFLAGS = "all=-trimpath=$(CURDIR)"
-GO_ASMFLAGS = "all=-trimpath=$(CURDIR)"
-
 LDFLAGS_linux = -static
 LDFLAGS_darwin =
 LDFLAGS_windows =
@@ -69,7 +66,7 @@ GO_LDFLAGS_linux =" $(GO_LDFLAGS)  -extldflags \"$(LDFLAGS_linux)\""
 
 # Build for local development.
 $(BUILD_DIR)/$(PROJECT): generate-statik $(GO_FILES) $(BUILD_DIR)
-	GOOS=$(GOOS) GOARCH=$(GOARCH) CGO_ENABLED=1 go build -tags $(GO_BUILD_TAGS_$(GOOS)) -ldflags $(GO_LDFLAGS_$(GOOS)) -gcflags $(GO_GCFLAGS) -asmflags $(GO_ASMFLAGS) -o $@ $(BUILD_PACKAGE)
+	GOOS=$(GOOS) GOARCH=$(GOARCH) CGO_ENABLED=1 go build -tags $(GO_BUILD_TAGS_$(GOOS)) -ldflags $(GO_LDFLAGS_$(GOOS)) -o $@ $(BUILD_PACKAGE)
 
 .PHONY: install
 install: $(BUILD_DIR)/$(PROJECT)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -28,10 +28,11 @@ GCP_PROJECT ?= k8s-skaffold
 GKE_CLUSTER_NAME ?= integration-tests
 GKE_ZONE ?= us-central1-a
 
-SUPPORTED_PLATFORMS := linux-$(GOARCH) darwin-$(GOARCH) windows-$(GOARCH).exe
+SUPPORTED_PLATFORMS = linux-$(GOARCH) darwin-$(GOARCH) windows-$(GOARCH).exe
```
tejal29: After going through http://www.gnu.org/software/make/manual/make.html#Flavors, 
I feel simply expanded variables and make more sense in this Makefile. 

Is there a reason why you changed to "recursively expanded" ?

Makefile
```diff
@@ -243,6 +240,7 @@ build-docs-preview:
 generate-schemas:
 	go run hack/schemas/main.go
 
-.PHONY: generate-statik
-generate-statik:
+# static files
+
+$(STATIK_FILES): go.mod docs/content/en/schemas/*
```
tejal29: Why do we need add the dependency to go.mod and docs/ here? These files are checked in the repository and would always be present

tejal29:
Testing notes:
Make is definitely faster
```
tejaldesai@@skaffold (faster-make-v2)$ time make
hack/generate-statik.sh
Collecting licenses
Collecting schemas
GOOS=linux GOARCH=amd64 CGO_ENABLED=1 go build -tags "osusergo netgo static_build release" -ldflags " -X github.com/GoogleContainerTools/skaffold/pkg/skaffold/version.version=v1.4.0-11-g15c669938-dirty -X github.com/GoogleContainerTools/skaffold/pkg/skaffold/version.buildDate=2020-02-27T15:22:04Z -X github.com/GoogleContainerTools/skaffold/pkg/skaffold/version.gitCommit=15c6699381e1dca6e5f1faa939df350576912668 -X github.com/GoogleContainerTools/skaffold/pkg/skaffold/version.gitTreeState=dirty -s -w  -extldflags \"-static\"" -o out/skaffold github.com/GoogleContainerTools/skaffold/cmd/skaffold

real	0m26.161s
user	0m48.223s
sys	0m1.393s
tejaldesai@@skaffold (faster-make-v2)$ time make
make: 'out/skaffold' is up to date.

real	0m0.356s
user	0m0.290s
sys	0m0.071s
tejaldesai@@skaffold (faster-make-v2)$ 

```

Tested the cloudbuild.yaml.
```
$git clean -fdx cmd/
Removing cmd/skaffold/app/cmd/statik/statik.go

$gcloud builds submit --config deploy/cloudbuild.yaml --substitutions=_RELEASE_BUCKET=tejal-test,COMMIT_SHA=$(git rev-parse HEAD) --project tejal-test
...
Success
```

```
TAG_NAME=vtest3724 gcloud builds submit --config deploy/cloudbuild-release.yaml --substitutions=_RELEASE_BUCKET=tejal-test --project tejal-test
..

SUCCESS

```

Makefile
dgageot: The dependency tells Make which inputs it should look at to decide if the output needs to be rebuilt.

tejal29: Make sense. So statik would only run if any of those inputs to `hack/generate_static.go` change

PR #3725, opened 2020-02-24
WIP: helm3 experimentations

tstromberg:
This is intended to be split into two PR's:

- One to introduce robust tests for what commands are generated by helm
- One to add support for the minor command-line differences in helm3:
  - `delete` drops the `--purge` flag
  - `install` drops the `--name` flag
  - `get` adds a `<type>` argument

IMHO, having minor flag switches like this is preferable to introducing a helm3 deployer, as it:

- does not require users to modify their configs based on the local binary version installed
- handles future minor releases (3.1) that will have their own behavioral change
- allows maximum code sharing among the 1200 existing lines that are still fully funcional


tejal29:
Can we close this now?

PR #3731, opened 2020-02-25
Add support for parsing helm version

tstromberg:
To prepare for helm3. The version is only logged in this PR. In a subsequent one, it will affect the flags generated, and will thus require tests.

#2142


pkg/skaffold/deploy/helm.go
```diff
@@ -76,6 +80,13 @@ func (h *HelmDeployer) Labels() map[string]string {
 func (h *HelmDeployer) Deploy(ctx context.Context, out io.Writer, builds []build.Artifact, labellers []Labeller) *Result {
 	event.DeployInProgress()
 
+	hv, err := h.binVer(ctx)
```
tejal29: Consider moving this to the constructor`NewHelmDeployer` to warn or fail fast in future when skaffold runner is initializing run config `runner.NewForConfig`

I feel we should move this to `debug` for now. Right now, users might not understand why the warning message is important. I am assuming in future, we would fallback to helm 2 or helm 3  and then warning might make more sense.
e.g.
```
failed to parse binary version: 3rc1. Assuming helm-2.X deployer.
```

pkg/skaffold/deploy/helm.go
```diff
@@ -512,6 +523,38 @@ func (h *HelmDeployer) Render(context.Context, io.Writer, []build.Artifact, []La
...
+	var b bytes.Buffer
+	if err := h.helm(ctx, &b, false, "version", "--short", "-c"); err != nil {
+		return semver.Version{}, errors.Wrap(err, "helm version")
+	}
+	bs := b.Bytes()
+
+	// raw for 3.1: "v3.1.0+gb29d20b"
+	// raw for 2.15: "Client: v2.15.1+gcf1de4f"
+	raw := string(bs)
+	idx := strings.Index(raw, "v")
```
tejal29: Please add tests for the L542 to L551 for different helm version strings.

pkg/skaffold/deploy/helm.go
tstromberg: I had considered this, but constructors should not generate I/O events, as there could be extremely adverse performance impacts when doing so.

pkg/skaffold/deploy/helm.go
tstromberg: Will do. Until the follow-up PR is submitted, this will unfortunately will mean that we have to test a private function, as this change is advisory only, and has no side-effects to the public functions.

tstromberg: This now has unit tests.

pkg/skaffold/deploy/helm.go
```diff
@@ -78,6 +82,13 @@ func (h *HelmDeployer) Labels() map[string]string {
 func (h *HelmDeployer) Deploy(ctx context.Context, out io.Writer, builds []build.Artifact, labellers []Labeller) *Result {
 	event.DeployInProgress()
 
+	hv, err := h.binVer(ctx)
+	if err != nil {
+		logrus.Infof("failed to parse binary version: %v", err)
```
tejal29:
```suggestion
		logrus.Debugf("failed to parse binary version: %v", err)
```

pkg/skaffold/deploy/helm.go
```diff
@@ -78,6 +82,13 @@ func (h *HelmDeployer) Labels() map[string]string {
...
+	} else {
+		logrus.Infof("deploying with helm version %v", hv)
```
tejal29:
```suggestion
		logrus.Debugf("deploying with helm version %v", hv)
```

pkg/skaffold/deploy/helm_test.go
```diff
@@ -565,6 +586,60 @@ func TestPackageHelmChart(t *testing.T) {
 
 }
 
+// TODO(tstromberg): Remove private function test once there are side-effects in a public function
+func TestBinVer(t *testing.T) {
+	tests := []struct {
```
tejal29: For completeness can we add a function where incorrect versions throw error as expected

tstromberg: Done.

PR #3759, opened 2020-02-28
This code is only used by tests

dgageot:

### DIFF
--- a/pkg/skaffold/initializer/config_test.go
+++ b/pkg/skaffold/initializer/config_test.go
@@ -22,7 +22,6 @@ import (
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


pkg/skaffold/initializer/config_test.go
```diff
@@ -239,3 +186,24 @@ func Test_canonicalizeName(t *testing.T) {
 		})
 	}
 }
+
+func artifacts(pairs []build.BuilderImagePair) []*latest.Artifact {
```
tejal29: This function is very copy of  https://github.com/GoogleContainerTools/skaffold/blob/45a55e159b46b5a3ba0c0bdd94f6376f1997eb7d/pkg/skaffold/initializer/build/util.go#L84 and can be removed too.

PR #3761, opened 2020-02-28
Add profile option to RunBuilder in test helper

loosebazooka:
Something I needed for #3382, see example here: https://github.com/GoogleContainerTools/skaffold/pull/3382/files#diff-74521f5953df4ff0bacd9d520bc395feR111

### DIFF
--- a/integration/skaffold/helper.go
+++ b/integration/skaffold/helper.go
@@ -41,6 +41,7 @@ type RunBuilder struct {
 	dir        string
 	ns         string
 	repo       string
+	profiles   []string
 	args       []string
 	env        []string
 	stdin      []byte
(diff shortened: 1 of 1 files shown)


integration/skaffold/helper.go
```diff
@@ -262,6 +269,9 @@ func (b *RunBuilder) cmd(ctx context.Context) *exec.Cmd {
 	if b.repo != "" && command.Flags().Lookup("default-repo") != nil {
 		args = append(args, "--default-repo", b.repo)
 	}
+	if b.profiles != nil && len(b.profiles) > 0 && command.Flags().Lookup("profile") != nil {
```
dgageot: You can simplify the test len(nil) is 0 in Go.
```suggestion
	if len(b.profiles) > 0 && command.Flags().Lookup("profile") != nil {
```

PR #3786, opened 2020-03-05
cleanup common flags + better -f description

balopat:
I wanted to rename `-f` flag's description to `Path or URL to the Skaffold config file`. 
However I found that we had a bunch of duplication - so I cleaned up that duplication. 
In the process, I had to add `CommonFlags` to `diagnose`, `fix` and `init`.  It turns out we wanted `diagnose` to take `-c` as well, so I left it there.

### DIFF
--- a/cmd/skaffold/app/cmd/diagnose.go
+++ b/cmd/skaffold/app/cmd/diagnose.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
 	yaml "gopkg.in/yaml.v2"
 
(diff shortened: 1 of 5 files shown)


briandealwis:
My $0.02: `init` and `fix` would benefit from a different `-f` text with an indication that the file will be changed -- the other cases the file is only read.

balopat:
> My $0.02: `init` and `fix` would benefit from a different `-f` text with an indication that the file will be changed -- the other cases the file is only read.

I think that's a fair point, I did think about that too. I think for `fix` we are good - as it's an input primarily - but for `init` we are generating it.

briandealwis:
Agreed with `fix`: `--overwrite` does say that it overwrites the original with fixed content.

PR #3804, opened 2020-03-10
Also clean up statik files

dgageot:


dgageot:
Self merging this minor fix to the `make clean` target

PR #3805, opened 2020-03-10
Fix ko sample

dgageot:
The previous code used to modify the artifact’s go.mod file


dgageot:
Self merging small fix to ko samples

PR #3806, opened 2020-03-10
Upgrade kompose to 1.21.0

dgageot:
Properly generates deployments of type `apps/v1` instead of deprecated type `extensions/v1beta2`


dgageot:
Self merging version bump

PR #3815, opened 2020-03-11
Use new t.Cleanup() to simplify tests

dgageot:


loosebazooka:
I think we need to update  https://github.com/GoogleContainerTools/skaffold/blob/master/integration/sync_test.go#L86 ?

dgageot:
@nkubala done

PR #3823, opened 2020-03-12
`skaffold build-image` command

dgageot:
Preliminary implementation of `skaffold build-image`.

The idea of this new command is to be able to build a Docker image out of any folder. Skaffolf will auto detect which builder to use among Docker, Jib and Buildpacks.

For example in `examples/getting-started` folder:

```
skaffold build-image --name my-image
```

Should build an image with `docker build` .

### DIFF
--- a/cmd/skaffold/app/cmd/buildimage.go
+++ b/cmd/skaffold/app/cmd/buildimage.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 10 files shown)


balopat:
> Preliminary implementation of `skaffold build-image`.
> 
> The idea of this new command is to be able to build a Docker image out of any folder. Skaffolf will auto detect which builder to use among Docker, Jib and Buildpacks.
> 
> For example in `examples/getting-started` folder:
> 
> ```
> skaffold build-image --name my-image
> ```
> 
> Should build an image with `docker build` .

Why aren't we doing this with an implicit `skaffold init` instead? I am puzzled why we need a separate command for this, why isn't `skaffold build` sufficient? We could just enhance that too.

dgageot:
I'm going to close this one for now

PR #3825, opened 2020-03-13
Simplify code that finds the artifact's type

dgageot:

### DIFF
--- a/pkg/skaffold/initializer/build/util.go
+++ b/pkg/skaffold/initializer/build/util.go
@@ -88,7 +88,8 @@ func Artifacts(pairs []BuilderImagePair) []*latest.Artifact {
 
 	for _, pair := range pairs {
 		artifact := &latest.Artifact{
-			ImageName: pair.ImageName,
+			ImageName:    pair.ImageName,
+			ArtifactType: pair.Builder.ArtifactType(),
 		}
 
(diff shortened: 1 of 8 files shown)


pkg/skaffold/initializer/build/util.go
```diff
@@ -97,8 +98,6 @@ func Artifacts(pairs []BuilderImagePair) []*latest.Artifact {
 			artifact.Workspace = workspace
 		}
 
-		pair.Builder.UpdateArtifact(artifact)
```
briandealwis: My only concern here is that lines 94–98 may update the artifact's `Workspace`, which could be relevant to `UpdateArtifact`.  I suppose since that workspace is from the same `InitBuilder` then `ArtifactType()` should have that same information.

Would it be worth moving the Artifact creation entirely into the InitBuilders?

dgageot: I which case do you think it would be useful for `UpdateArtifact` to know the `Workspace`? I wanted to simplify the code but I'm ok to go the opposite way if there's a concrete need for it

briandealwis: I think it's a definite improvement.  If we need the workspace, we can amend the function to pass in the workspace.

PR #3837, opened 2020-03-17
Support Dockerfile.dockerignore

agrinh:
**Fixes**: Did not create an issue for this, please let me know if I should.

**Description**
Add support for getting dependencies from Dockerfile.dockerignore files. See e.g. https://github.com/moby/buildkit/pull/901 . Since the same code produces dependencies for Kaniko I expect this will also make Skaffold align better with Kaniko (https://github.com/GoogleContainerTools/kaniko/pull/801).

**User facing changes**
Users with multiple Dockerfiles will now be able to have different `.dockerignore` files. For example, this is the behavior with the following context:
```
.
├── .dockerignore
├── Dockerfile
├── Dockerfile.dev
└── Dockerfile.dev.dockerignore
```
**_before_:** `docker` would respect `Dockerfile.dev.dockerignore` but `skaffold` would not take it into account when producing dependencies, and syncing could break e.g. if `.dockerignore` hid files used by `Dockerfile.dev.dockerignore`, leading to unexpected behavior.

**_after_:** `skaffold` will make `Dockerfile.dev.dockerignore` take precedence over `.dockerignore` (will fall back to `.dockerignore` if it doesn't exist).

### DIFF
--- a/pkg/skaffold/docker/dependencies.go
+++ b/pkg/skaffold/docker/dependencies.go
@@ -54,7 +54,7 @@ func GetDependencies(ctx context.Context, workspace string, dockerfilePath strin
 		return nil, err
 	}
 
-	excludes, err := readDockerignore(workspace)
+	excludes, err := readDockerignore(workspace, absDockerfilePath)
 	if err != nil {
 		return nil, errors.Wrap(err, "reading .dockerignore")
 	}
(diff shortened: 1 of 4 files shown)


agrinh:
@googlebot I signed it!

pkg/skaffold/docker/dependencies.go
```diff
@@ -89,19 +89,25 @@ func GetDependencies(ctx context.Context, workspace string, dockerfilePath strin
 }
 
 // readDockerignore reads patterns to ignore
-func readDockerignore(workspace string) ([]string, error) {
+func readDockerignore(workspace string, dockerfilePath string) ([]string, error) {
```
dgageot: Should it be called `absDockerfilePath` to show that it's not relative to the `workspace`?

pkg/skaffold/docker/dependencies.go
```diff
@@ -89,19 +89,25 @@ func GetDependencies(ctx context.Context, workspace string, dockerfilePath strin
...
-		if err != nil {
-			return nil, err
+			excludes, err = dockerignore.ReadAll(r)
+			if err != nil {
+				return nil, err
+			}
+			return excludes, nil
 		}
 	}
 	return excludes, nil
```
dgageot: Should it return `nil, nil`?

pkg/skaffold/docker/dependencies.go
agrinh: Yes, makes sense. Done.

pkg/skaffold/docker/dependencies.go
agrinh: Good point, that would be clearer. Done.

agrinh:
Weird, integration tests failed at first but they succeeded after restarting.

PR #3854, opened 2020-03-21
upgrade hugo + small fixes

balopat:
This upgrades hugo to the latest version - not the default behavior of including HTML changed, hence the `unsafe=true` for the goldmark renderer (when I manually pushed yesterday the site, this resulted in a bunch of pages to not work including the install page - you can still see that here: https://skaffold-latest.firebaseapp.com/docs/install/). Also Privacy settings link was not working.


dgageot:
The site looks good to me

PR #3869, opened 2020-03-25
Fix possible nil dereference

dgageot:


dgageot:
Minor change to the tests. Self merging

PR #3877, opened 2020-03-26
Wait for the logs to be printed

dgageot:
Previous fix removed the `wc.Wait()` by error


dgageot:
Minor change. Self merging

PR #3885, opened 2020-03-27
Refactoring on filepath.Walk

dgageot:
This is a first step in the refactoring of `filepath.Walk` and `godirwalk.Walk` usage.

The idea is to simplify the code that walks directories by extracting as much common code as possible.
More simplifications are coming but this first step is already bringing many simplifications.

### DIFF
--- a/pkg/skaffold/build/jib/jib.go
+++ b/pkg/skaffold/build/jib/jib.go
@@ -31,12 +31,12 @@ import (
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
-	"github.com/karrick/godirwalk"
 	"github.com/sirupsen/logrus"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/walk"
 )
 
 const (
@@ -250,22 +250,26 @@ func walkFiles(workspace string, watchedFiles []string, ignoredFiles []string, c
 		}
 
 		// Process directory
-		if err = godirwalk.Walk(dep, &godirwalk.Options{
-			Unsorted: true,
-			Callback: func(path string, _ *godirwalk.Dirent) error {
-				if isIgnored(path, ignoredFiles) {
-					return filepath.SkipDir
-				}
-				if info, err := os.Stat(path); err == nil && !info.IsDir() {
-					// try to relativize the path: an error indicates that the file cannot
-					// be made relative to the roots, and so we just use the full path
-					if relative, err := relativize(path, workspaceRoots...); err == nil {
-						path = relative
-					}
-					return callback(path, info)
-				}
+		if err = walk.From(dep).Unsorted().Do(func(path string, info walk.Dirent) error {
+			if isIgnored(path, ignoredFiles) {
+				return filepath.SkipDir
+			}
+
+			if info.IsDir() {
 				return nil
-			},
+			}
+
+			stat, err := os.Stat(path)
+			if err != nil {
+				return nil // Ignore
+			}
+
+			// try to relativize the path: an error indicates that the file cannot
+			// be made relative to the roots, and so we just use the full path
+			if relative, err := relativize(path, workspaceRoots...); err == nil {
+				path = relative
+			}
+			return callback(path, stat)
 		}); err != nil {
 			return fmt.Errorf("filepath walk: %w", err)
 		}
--- a/pkg/skaffold/build/jib/jib.go
+++ b/pkg/skaffold/build/jib/jib.go
@@ -249,16 +249,16 @@ func walkFiles(workspace string, watchedFiles []string, ignoredFiles []string, c
 			continue
 		}
 
-		// Process directory
-		if err = walk.From(dep).Unsorted().Do(func(path string, info walk.Dirent) error {
+		notIgnored := func(path string, info walk.Dirent) (bool, error) {
 			if isIgnored(path, ignoredFiles) {
-				return filepath.SkipDir
+				return false, filepath.SkipDir
 			}
 
-			if info.IsDir() {
-				return nil
-			}
+			return true, nil
+		}
 
+		// Process directory
+		if err = walk.From(dep).Unsorted().When(notIgnored).WhenIsFile().Do(func(path string, info walk.Dirent) error {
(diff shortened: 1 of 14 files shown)


pkg/skaffold/deploy/helm.go
```diff
@@ -162,25 +163,16 @@ func (h *HelmDeployer) Dependencies() ([]string, error) {
...
-					// if we are not running the helm dep build phase, as that modifies files inside
-					// the chartDepsDir and results in an infinite build loop.
-					deps = append(deps, path)
-				}
-			}
 
-			return nil
-		})
+		// We can always add a dependency if it is not contained in our chartDepsDir.
+		// However, if the file is in  our chartDepsDir, we can only include the file
```
briandealwis:
```suggestion
		// However, if the file is in our chartDepsDir, we can only include the file
```

pkg/skaffold/walk/walk.go
```diff
@@ -0,0 +1,165 @@
...
+				}
+			}
+			return nil
+		},
+	})
+}
+
+func (w *builder) MustDo(action Action) {
+	if err := w.Do(action); err != nil {
+		panic("unable to list files")
```
briandealwis: I don't feel strongly about this but go/go-read#dont-panic suggests using `log.Fatal()` rather than panicking, but I can see this would be useful.  Regardless, we should include the error here.

pkg/skaffold/walk/walk.go
```diff
@@ -0,0 +1,165 @@
...
+	WhenIsDir() Builder
+	WhenIsFile() Builder
+	WhenHasName(string) Builder
+
+	// Actions
+	Do(Action) error
+	MustDo(Action)
+	AppendPaths(*[]string) error
+	CollectPaths() ([]string, error)
+}
```
briandealwis: Wouldn't it make sense to incorporate `Excluder` in the `Builder` interface?

pkg/skaffold/docker/syncmap.go
```diff
@@ -75,42 +74,39 @@ func walkWorkspaceWithDestinations(workspace string, excludes []string, fts []fr
...
 		switch mode := fi.Mode(); {
 		case mode.IsDir():
-			if err := godirwalk.Walk(absFrom, &godirwalk.Options{
-				Unsorted: true,
-				Callback: func(fpath string, info *godirwalk.Dirent) error {
-					if fpath == absFrom {
-						return nil
-					}
+			if err := walk.From(absFrom).Unsorted().Do(func(fpath string, info walk.Dirent) error {
+				if fpath == absFrom {
```
briandealwis: It feels like this excluding logic should be provided to the walker, something like:
```
walk.From(absFrom).Unsorted().Excluding(excluder).Do(...)
```
WDYT?

dgageot: It should indeed. I've just got started working on the Excluder. When it's more clearly defined, I'll add it to the walk API

pkg/skaffold/walk/walk.go
```diff
@@ -0,0 +1,165 @@
...
+)
+
+type Dirent interface {
+	IsDir() bool
+	Name() string
+}
+
+type Predicate func(string, Dirent) bool
+
+type Action func(string, Dirent) error
```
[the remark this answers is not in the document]
dgageot: I kept the name that filepath.Walk and godirwalk.Walk are using but it's true that they are a bit obscure. Will try to change that.

pkg/skaffold/walk/walk.go
```diff
@@ -0,0 +1,165 @@
...
+
+type Excluder interface {
+	Exclude(relPath string) (bool, error)
+	SkipDir(relPath string) bool
+}
```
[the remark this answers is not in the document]
dgageot: It's an early implementation. If I can, I'd like to have only one function but not yet sure how.

pkg/skaffold/walk/walk.go
dgageot: yes, I will in a later PR

pkg/skaffold/walk/walk.go
dgageot: `panic` is how it's done in the stdlib. See `regexp.MustCompile` for example. I'll add the error.

dgageot:
@briandealwis could you take a second (third?) look

pkg/skaffold/walk/walk.go
```diff
@@ -0,0 +1,172 @@
...
+
+// Predicate represents a predicate on file system entries.
+// Given a file's absolute path and information, it returns `true`
+// when the predicate is matched. It can also return a `filepath.SkipDir`
+// error to skip a directory and its children altogether.
+type Predicate func(absPath string, info Dirent) (bool, error)
+
+// Action is a function that takes a file's absolute path and information,
+// and optionally returns an error.
+type Action func(absPath string, info Dirent) error
```
briandealwis, line 39: Confirmed that `godirwalk.Walk()` doesn't return absolute files unless an absolute directory was passed in.  So these should just be  `path` and skip any mention of absolute.

pkg/skaffold/walk/walk.go
```diff
@@ -0,0 +1,172 @@
...
+			return err
+		}
+
+		return action(w.dir, info)
+	}
+
+	return godirwalk.Walk(w.dir, &godirwalk.Options{
+		Unsorted: w.unsorted,
+		Callback: func(path string, info *godirwalk.Dirent) error {
+			match, err := w.predicate(path, info)
```
briandealwis, line 122: It seems most of the uses of `Walk()` relativize the path against the dir.  I wonder if it would make sense to relativize the path here and pass in the relative path into the predicate and action.  Or maybe it can be controlled by an option on the builder? `walk.From(xxx).RelativePaths().When(yyy).Do(zzz)`.

pkg/skaffold/build/jib/jib.go
```diff
@@ -250,22 +250,26 @@ func walkFiles(workspace string, watchedFiles []string, ignoredFiles []string, c
...
-						path = relative
-					}
-					return callback(path, info)
-				}
+		if err = walk.From(dep).Unsorted().Do(func(path string, info walk.Dirent) error {
+			if isIgnored(path, ignoredFiles) {
+				return filepath.SkipDir
+			}
+
+			if info.IsDir() {
```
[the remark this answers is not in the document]
dgageot: Doing it. Thanks

pkg/skaffold/walk/walk.go
dgageot: Yeah, I've seen that too. Maybe in a next PR though?

PR #3888, opened 2020-03-30
Use less memory for linting

dgageot:


dgageot:
Self-merging. Fixing master branch

Issue #3886, opened 2020-03-29
Skaffold dev on a new machine does not download images from registry.


arlyon:
### Expected behavior

I have some microservices. I run `skaffold build` on my machine (or in CI), the images are uploaded, and then when others who depend on them (ie. frontend team) rebases, they can run `skaffold dev`, it pulls the images I pushed, and they are able to keep working immediately.

### Actual behavior

Images are 'not found' and built instead. It appears that skaffold doesn't check docker hub and as such, everyone who rebases needs to either manually pull the updated images from docker or wait for it to build. This means rebases are quite expensive (taking around an hour if building).

### Information

- Skaffold version: 1.6.0
- Operating system: images built and pushed on macOS, other machines (that have to rebuild) are windows and centOS

<details><summary>skaffold.yaml</summary>
<p>

```yaml
apiVersion: skaffold/v2beta1
kind: Config
metadata:
  name: workout
deploy:
  kustomize:
    paths: 
    - kubernetes/overlays/development
build:
  artifacts:
  - image: arlyon/workout-server
    context: ./server/
  - image: arlyon/workout-learn
    context: ./learn/
  local:
    push: true
profiles:
- name: production
  deploy:
    kustomize:
      paths: 
      - kubernetes/overlays/production
  activation:
  - kubeContext: google-cloud
- name: traefik
  deploy:
    kustomize:
      paths:
      - kubernetes/overlays/development
      - kubernetes/overlays/traefik
portForward:
- resourceType: service
  resourceName: server
  address: 0.0.0.0
  port: 8080
```

</p>
</details>

### Steps to reproduce the behavior

I don't think this is unique to my project. I'm just not sure why this wouldn't happen. I would expect the cache first check locally, then to fall back and check docker hub, and finally resort to building but it seems like that step is being missed out. Note that after deleting the images on the machine that built them, they _are_ 'found remotely' from docker hub, but not on other machines

arlyon:
PR is in progress (link above).

In the mean time, a workaround is to use a shared skaffold cache (ie. checked into the source code) coupled with [this script](https://gist.github.com/arlyon/2a146dd97976f3c9916c9fa258f28fb6) that will attempt to pull the images that match your current commit.

tstromberg:
This seems reasonable - and I'm glad to see you've provided a PR. Thank you!

tejal29:
Closing the issue as the fix got merged!

PR #3891, opened 2020-03-31
Try to import docker images before falling back to building

arlyon:
**Fixes**: #3886

**Description**

When perfoming a lookup of artifacts, this change will attempt to import artifacts from docker running locally as well as whatever remote registry is configured. This enables the worklow outlined in the issue above.

This PR is not done because there needs to be some discussion as it changes the behaviour of builds. Should the functionality be placed behind a flag? Since the contents of images aren't verified, this could for example cause problems if a malicious (or unaware) user uploads images tagged with an incorrect commit however for those who are aware of the (imo limited) drawbacks the productivity boost far-outweighs the risk. That said, I would argue to make it opt-in.

**User facing changes**

This would cause anyone running `skaffold build` to have the images be imported from docker, if possible, 'short circuiting' some builds. This is particularly useful when sharing images across teams (say someone needs to check out your branch to help diagnose an issue).

**Follow-up work**

Once the questions above are answered, then the PR needs to be adapted slightly and the docs updated.

### DIFF
--- a/pkg/skaffold/build/cache/cache.go
+++ b/pkg/skaffold/build/cache/cache.go
@@ -49,6 +49,7 @@ type cache struct {
 	insecureRegistries map[string]bool
 	cacheFile          string
 	imagesAreLocal     bool
+	tryImportMissing   bool
 	hashForArtifact    func(ctx context.Context, a *latest.Artifact) (string, error)
 }
 
@@ -64,7 +65,7 @@ type Config interface {
 }
 
 // NewCache returns the current state of the cache
-func NewCache(cfg Config, imagesAreLocal bool, dependencies DependencyLister) (Cache, error) {
+func NewCache(cfg Config, imagesAreLocal bool, tryImportMissing bool, dependencies DependencyLister) (Cache, error) {
 	if !cfg.CacheArtifacts() {
 		return &noCache{}, nil
 	}
@@ -92,6 +93,7 @@ func NewCache(cfg Config, imagesAreLocal bool, dependencies DependencyLister) (C
 		insecureRegistries: cfg.GetInsecureRegistries(),
 		cacheFile:          cacheFile,
 		imagesAreLocal:     imagesAreLocal,
+		tryImportMissing:   tryImportMissing,
 		hashForArtifact: func(ctx context.Context, a *latest.Artifact) (string, error) {
 			return getHashForArtifact(ctx, dependencies, a, cfg.Mode())
 		},
--- a/pkg/skaffold/build/cache/cache.go
+++ b/pkg/skaffold/build/cache/cache.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"sync"
 
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/sirupsen/logrus"
@@ -45,6 +46,7 @@ type ArtifactCache map[string]ImageDetails
 // cache holds any data necessary for accessing the cache
 type cache struct {
 	artifactCache      ArtifactCache
+	cacheMutex         sync.RWMutex
 	client             docker.LocalDaemon
 	insecureRegistries map[string]bool
 	cacheFile          string
--- a/pkg/skaffold/build/cache/details.go
+++ b/pkg/skaffold/build/cache/details.go
@@ -111,8 +111,10 @@ func (d needsPushing) Push(ctx context.Context, out io.Writer, c *cache) error {
 	}
 
 	// Update cache
+	c.cacheMutex.Lock()
 	e := c.artifactCache[d.hash]
 	e.Digest = digest
 	c.artifactCache[d.hash] = e
+	c.cacheMutex.Unlock()
 	return nil
 }
--- a/pkg/skaffold/build/cache/lookup.go
+++ b/pkg/skaffold/build/cache/lookup.go
@@ -53,7 +53,9 @@ func (c *cache) lookup(ctx context.Context, a *latest.Artifact, tag string) cach
 		return failed{err: fmt.Errorf("getting hash for artifact %q: %s", a.ImageName, err)}
 	}
 
+	c.cacheMutex.RLock()
 	entry, cacheHit := c.artifactCache[hash]
+	c.cacheMutex.RUnlock()
 	if !cacheHit {
 		if entry, err = c.tryImport(ctx, a, tag, hash); err != nil {
 			logrus.Debugf("Could not import artifact from Docker, building instead (%s)", err)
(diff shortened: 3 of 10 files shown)


arlyon:
@googlebot I signed it!

pkg/skaffold/build/cache/lookup.go
```diff
@@ -59,7 +61,10 @@ func (c *cache) lookup(ctx context.Context, a *latest.Artifact, tag string) cach
 
 	entry, cacheHit := c.artifactCache[hash]
 	if !cacheHit {
-		return needsBuilding{hash: hash}
+		entry, err = c.tryImport(ctx, a, tag, hash)
+		if err != nil {
+			return needsBuilding{hash: hash}
```
tstromberg, line 66: It would be nice to log the error that caused it to still need building.

Otherwise, PR seems reasonable to me. Anything required to remove the 'Draft' label?

arlyon: Nothing (aside from docs changes), as long as you're happy to make this the default behaviour, as opposed to opt in. I'll have a scan through the docs and see what needs updating.

arlyon:
@tstromberg Just fixed the failing test and added the debug log!

dgageot:
@arlyon I think it's best to make it opt-in or to make it possible to get the current behaviour back. I've always found it super important in a fully local mode to be able to never connect to a remote registry. For example, you are in a train, with no network, assuming all the layers where already downloaded, you are able to iterate quickly without reaching out to the network.

arlyon:
> @arlyon I think it's best to make it opt-in or to make it possible to get the current behaviour back. I've always found it super important in a fully local mode to be able to never connect to a remote registry. For example, you are in a train, with no network, assuming all the layers where already downloaded, you are able to iterate quickly without reaching out to the network.

You're right. For now I implemented it as a config option under local builds:

```yaml
build:
  artifacts:
  - image: arlyon/workout-server
    context: ./server/
  - image: arlyon/workout-learn
    context: ./learn/
  local:
    tryImportMissing: true
    push: true
```

pkg/skaffold/schema/latest/config.go
```diff
@@ -191,6 +191,10 @@ type LocalBuild struct {
 	// connects to a remote cluster.
 	Push *bool `yaml:"push,omitempty"`
 
+	// TryImportMissing should attempt to import this artifact from
+	// Docker (either locally or a remote registry) if not in cache.
+	TryImportMissing bool `yaml:"tryImportMissing,omitempty"`
```
tejal29, line 196: Can you please make this a pointer like other?
And then you would need to set to it false in `defaultToLocalBuild` in `pkg/skaffold/schemadefaults/`

tejal29:
@arlyon are you still working on this?

nkubala:
@arlyon this PR is getting a bit stale but I'd love to help you get it in. let me know if you need any help with it! i'm still seeing a test failing on a data race in the cache somewhere, can you check that out?

arlyon:
Hey, haven't been using skaffold at my day job recently but I have some time today so I'll rebase and see if I can get the requested changes in today.

arlyon:
I have updated with the requested changes and fixed the potential data race issues by introducing a RWMutex. `make test` fails complaining that schema `v2beta5` is already released but I'm not sure what to do about that.

gsquared94:
You can rebase on master now. A new schema version has already been created. That should help with the tests

arlyon:
Hate to ping because I know you're all busy but would appreciate a merge @gsquared94 @tejal29  :)

dgageot:
I can TAL tomorrow. Added to m'y TODO

Le ven. 24 juil. 2020 à 17:00, Alexander Lyon <notifications@github.com> a
écrit :

pkg/skaffold/build/local/types.go
```diff
@@ -75,12 +76,18 @@ func NewBuilder(runCtx *runcontext.RunContext) (*Builder, error) {
 		pushImages = *runCtx.Cfg.Build.LocalBuild.Push
 	}
 
+	tryImportMissing := false
+	if runCtx.Cfg.Build.LocalBuild.TryImportMissing != nil {
+		tryImportMissing = *runCtx.Cfg.Build.LocalBuild.TryImportMissing
```
dgageot, line 81: `TryImportMissing` should probably just be of type bool since it defaults to `false`

pkg/skaffold/build/cache/lookup.go
```diff
@@ -108,3 +116,39 @@ func (c *cache) lookupRemote(ctx context.Context, hash, tag string, entry ImageD
...
+	}
+
+	if digest, err := docker.RemoteDigest(tag, c.insecureRegistries); err == nil {
+		logrus.Debugf("Added digest for %s to cache entry", tag)
+		entry.Digest = digest
+	}
+	c.cacheMutex.Lock()
+	defer c.cacheMutex.Unlock()
+	c.artifactCache[hash] = entry
+	return entry, nil
```
dgageot, line 153:
```suggestion
	c.cacheMutex.Lock()
	c.artifactCache[hash] = entry
	c.cacheMutex.Unlock()
	return entry, nil
```

pkg/skaffold/build/cache/cache.go
```diff
@@ -80,10 +83,12 @@ func NewCache(runCtx *runcontext.RunContext, imagesAreLocal bool, dependencies D
 
 	return &cache{
 		artifactCache:      artifactCache,
+		cacheMutex:         sync.RWMutex{},
```
dgageot, line 86: This line is not needed. The default value for `sync.RWMutex` works.

pkg/skaffold/build/cache/details.go
```diff
@@ -111,6 +111,8 @@ func (d needsPushing) Push(ctx context.Context, out io.Writer, c *cache) error {
 	}
 
 	// Update cache
+	c.cacheMutex.Lock()
+	defer c.cacheMutex.Unlock()
 	e := c.artifactCache[d.hash]
 	e.Digest = digest
 	c.artifactCache[d.hash] = e
```
dgageot, line 118:
```suggestion
	c.cacheMutex.Lock()
	e := c.artifactCache[d.hash]
	e.Digest = digest
	c.artifactCache[d.hash] = e
	c.cacheMutex.Unlock()
```

pkg/skaffold/build/local/types.go
briandealwis: @tejal29 asked him to turn it into a pointer.  A pointer at least allows us to later check if the value was explicitly specified.

dgageot: What for?

briandealwis: If we change the default?

dgageot: I see. Not sure we should take such future changes into account. pointers on booleans are the worst

nkubala:
@arlyon this PR is getting a bit stale. do you still have the bandwidth to try and push it through?

arlyon:
Hi @nkubala sorry for the delay I though I'd pushed the requested changes but clearly not. I have rebased on master now!

MarlonGamez (approved):
@arlyon This looks good to me! Thanks for making all of the requested changes :)

PR #3894, opened 2020-04-01
Use new name for the linter’s cache

dgageot:


dgageot:
Self merging. This fixes the Out of Memory error on most of the builds

PR #3914, opened 2020-04-02
fix: correctly set namespace when checking for an existing helm release via skaffold deploy

mrparkers:
**Fixes**: #3838 

**Description**

When using `skaffold deploy` with `helm`, skaffold was not honoring the `--namespace` command-line flag when checking if a helm release already exists.  You can reproduce this error using the code within the `examples/helm-deployment` folder:

Before this PR:

```bash
cd examples/helm-deployment
kubectl create ns skaffold-test
skaffold build -q | skaffold deploy --namespace skaffold-test --build-artifacts - # results in a successful deployment
skaffold build -q | skaffold deploy --namespace skaffold-test --build-artifacts - # results in the error "cannot re-use a name that is still in use"
```

After this PR:

```bash
cd examples/helm-deployment
kubectl create ns skaffold-test
skaffold build -q | skaffold deploy --namespace skaffold-test --build-artifacts - # results in a successful deployment
skaffold build -q | skaffold deploy --namespace skaffold-test --build-artifacts - # results in a successful helm upgrade
```

### DIFF
--- a/pkg/skaffold/deploy/helm.go
+++ b/pkg/skaffold/deploy/helm.go
@@ -257,17 +257,17 @@ func (h *HelmDeployer) deployRelease(ctx context.Context, out io.Writer, r lates
 		helmVersion: helmVersion,
 	}
 
-	if err := h.exec(ctx, ioutil.Discard, false, getArgs(helmVersion, releaseName, r.Namespace)...); err != nil {
-		color.Yellow.Fprintf(out, "Helm release %s not installed. Installing...\n", releaseName)
-
-		opts.upgrade = false
-		opts.flags = h.Flags.Install
-	}
-
 	if h.namespace != "" {
 		opts.namespace = h.namespace
 	} else if r.Namespace != "" {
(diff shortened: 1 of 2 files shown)


mrparkers:
cc @clement-buchart

mrparkers:
side note: please correct me if I'm wrong, but it looked like a lot of the assertions for the existing unit tests around the helm deploy code were incorrect - most of them were asserting that the `--namespace` flag was not used in the first `helm get`, but was used in the following `helm upgrade`.  this didn't make sense to me, I think that we should be expecting that the `--namespace` flag is used in both cases or in neither case.

I went ahead and fixed these tests and added a few more that used a namespaced skaffold context to assert that the `--namespace` flag is being used when the skaffold context contains a namespace, even if the helm release does not contain a namespace.

please let me know if I misunderstood something here and if I have to revert it back to the way it was.

dgageot:
Awesome! Thanks a lot @mrparkers

Issue #3843, opened 2020-03-19
helm: delete doesn't respect namespace


no1melman:
### Expected behavior

Run

```
skaffold delete -n mynamespace
```

Everything should be cleaned up

### Actual behavior

When running with custom namespace parameter, the `delete` command ignores it.
This doesn't help if you used `run -n` to put into a different namespace

Output from trace logs show command being run:

```
DEBU[0000] Running command: [helm --kube-context docker-for-desktop delete myrelease --namespace othernamespace] 
```

### Information

- Skaffold version: 1.5.0
- Operating system: MacOS Mojave
- Contents of skaffold.yaml:

```yaml
deploy:
  helm:
    releases:
    - name: myrelease
      chartPath: k8s/client
      values:
        image.repository: myrelease
      namespace: othernamespace
```

### Steps to reproduce the behavior

Run

```
skaffold run -n mynamespace
skaffold delete -n mynamespace
```

dgageot:
@no1melman are you using helm3 or a previous version?

no1melman:
Helm 3

no1melman:
@balopat do you require any more evidence?

erpel:
Can confirm:

`skaffold -v debug delete --namespace amqtest`
...
`DEBU[0000] Running command: [helm --kube-context zephyr-infra delete activemq]`

`$ helm version
version.BuildInfo{Version:"v3.1.2", GitCommit:"d878d4d45863e42fd5cff6743294a11d28a9abce", GitTreeState:"clean", GoVersion:"go1.13.8"}
$ skaffold version
v1.6.0`

abh:
This might be fixed in #3929

tejal29:
We are prioritizing reviewing the PR and verifying this is Fixed in #3929

tstromberg:
I looked through #3929 and don't see how it would fix this issue?

tstromberg:
I tested this at head, but and it seems to be fixed now:

`cd examples/helm-deployment-dependencies`
`skaffold run -n triage-party`
`skaffold delete -n triage-party`

Please add a comment if you are still seeing this with the most recent Skaffold release.

PR #3929, opened 2020-04-09
Initial implementation of Helm Renderer

tete17:
**Related**:https://github.com/GoogleContainerTools/skaffold/issues/1937

**Description**
This PR should add Render capabilities to the Helm deployer.
I have added support for the following feature: 
 * valuesFiles
 * setValueTemplates
 * values
 * Namespaced charts

I hope this is enough, If I have missed some that are consider a must please let me know. I have also tried having a good support between Helm v3 and v2.

The only user facing change (except the obvious addition of `skaffold render` for helm, can bee seen [here](https://github.com/GoogleContainerTools/skaffold/compare/master...tete17:Add-Render-Capabilities-to-Helm-Deployer?expand=1#diff-5c13a6850c742c6b4fa07cf1f6aeb088L611).

I have changed this behaviour because it was assuming all members of the `values` map were supposed to be images built by skaffold. This is use to make skaffold replace the images by the combination of the image:tag in the manifest.

I have modified so it only looks for the appropiate images and doesn't fail if you add something like
```yaml
values: 
   image: "skaffold-helm"
   replicas: 3
```
This lead me to modify certain unit tests accordingly. I am not entirely sure about this change I am by no means and expert in helm so feedback is highly welcomed.

It is also my first contribution to a go project so please be gentle :smile:. I hope I followed best practice I have a feeling I am duplicating some code so please let me know if someone else also has that feeling.

### DIFF
--- a/pkg/skaffold/deploy/helm.go
+++ b/pkg/skaffold/deploy/helm.go
@@ -20,7 +20,6 @@ import (
 	"bufio"
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -53,10 +52,13 @@ import (
 
 var (
 	// versionRegex extracts version from "helm version --client", for instance: "2.14.0-rc.2"
-	versionRegex = regexp.MustCompile(`\"v(\d[\w\.\-\.]+)`)
+	versionRegex = regexp.MustCompile(`v(\d[\w.\-]+)`)
 
 	// helm3Version represents the version cut-off for helm3 behavior
 	helm3Version = semver.MustParse("3.0.0-beta.0")
+
+	// error to throw when helm version can't be determined
+	versionErrorString = "failed to determine binary version: %w"
 )
 
 // HelmDeployer deploys workflows using the helm CLI
@@ -99,7 +101,7 @@ func (h *HelmDeployer) Deploy(ctx context.Context, out io.Writer, builds []build
 
 	hv, err := h.binVer(ctx)
 	if err != nil {
-		return NewDeployErrorResult(fmt.Errorf("failed to determine binary version: %w", err))
+		return NewDeployErrorResult(fmt.Errorf(versionErrorString, err))
 	}
 	logrus.Infof("Deploying with helm v%s ...", hv)
 
@@ -129,10 +131,10 @@ func (h *HelmDeployer) Deploy(ctx context.Context, out io.Writer, builds []build
 
 	// Let's make sure that every image tag is set with `--set`.
 	// Otherwise, templates have no way to use the images that were built.
-	for _, build := range builds {
-		if !valuesSet[build.Tag] {
-			warnings.Printf("image [%s] is not used.", build.Tag)
-			warnings.Printf("image [%s] is used instead.", build.ImageName)
+	for _, b := range builds {
+		if !valuesSet[b.Tag] {
+			warnings.Printf("image [%s] is not used.", b.Tag)
+			warnings.Printf("image [%s] is used instead.", b.ImageName)
 			warnings.Printf("See helm sample for how to replace image names with their actual tags: https://github.com/GoogleContainerTools/skaffold/blob/master/examples/helm-deployment/skaffold.yaml")
 		}
 	}
@@ -215,7 +217,7 @@ func (h *HelmDeployer) Dependencies() ([]string, error) {
 func (h *HelmDeployer) Cleanup(ctx context.Context, out io.Writer) error {
 	hv, err := h.binVer(ctx)
 	if err != nil {
-		return fmt.Errorf("binary version: %w", err)
+		return fmt.Errorf(versionErrorString, err)
 	}
 
 	for _, r := range h.Releases {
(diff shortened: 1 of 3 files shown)


tete17:
@googlebot I signed it!

abh:
At a glance this also fixes a bug where the namespace specified with `skaffold -n ...` isn't actually used.  Yay @tete17  :-)

pkg/skaffold/deploy/helm.go
```diff
@@ -214,8 +213,83 @@ func (h *HelmDeployer) Cleanup(ctx context.Context, out io.Writer) error {
 }
 
 // Render generates the Kubernetes manifests and writes them out
-func (h *HelmDeployer) Render(context.Context, io.Writer, []build.Artifact, []Labeller, string) error {
-	return errors.New("not yet implemented")
+func (h *HelmDeployer) Render(ctx context.Context, out io.Writer, builds []build.Artifact, labellers []Labeller, filepath string) error {
```
tstromberg, line 216: I'm not well-versed in the renderer, but doesn't it seems very wrong for a function to accept both a `io.Writer` and a `filepath` that overrides the `io.Writer` argument.

 Is this just an API quirk you are trying to deal with?

tete17: I am no versed enough in go to know if it is a good pattern or not.

All I can say is that the `filepath` parameter is used to support the `--output` option.
That decides whether you want your rendered manifest into a file or to the standard output.

You can see a similar usage as my PR [here](https://github.com/GoogleContainerTools/skaffold/blob/bb7807edd302bb6a354434037684a218b248f27e/pkg/skaffold/deploy/util.go#L86) and it usages in the `kubectl` renderer and [here](https://github.com/GoogleContainerTools/skaffold/blob/bb7807edd302bb6a354434037684a218b248f27e/pkg/skaffold/deploy/kustomize.go#L200) for the `kustomize` renderer.

tete17:
I rebased the branch to the latest master

pkg/skaffold/deploy/helm.go
```diff
@@ -214,8 +213,83 @@ func (h *HelmDeployer) Cleanup(ctx context.Context, out io.Writer) error {
...
+	hv, err := h.binVer(ctx)
+	if err != nil {
+		return fmt.Errorf("binary version: %w", err)
```
nkubala, line 219: let's make this a bit more descriptive, and match with our other errors. could you pull out the error string that we're throwing when checking binary version in `Deploy()` into a constant, and then use that both there and here?

pkg/skaffold/deploy/helm.go
```diff
@@ -607,12 +676,10 @@ func pairParamsToArtifacts(builds []build.Artifact, params map[string]string) (m
 
 	for param, imageName := range params {
 		b, ok := imageToBuildResult[imageName]
-		if !ok {
-			return nil, fmt.Errorf("no build present for %s", imageName)
+		if ok {
+			paramToBuildResult[param] = b
```
nkubala, line 680: was this error not valid? it seems like we do still want to make sure all build artifacts are present.

pkg/skaffold/deploy/helm.go
```diff
@@ -51,7 +50,7 @@ import (
 
 var (
 	// versionRegex extracts version from "helm version --client", for instance: "2.14.0-rc.2"
-	versionRegex = regexp.MustCompile(`\"v(\d[\w\.\-\.]+)`)
+	versionRegex = regexp.MustCompile(`v(\d[\w.\-]+)`)
```
nkubala, line 53: was this regex mismatching before? I don't doubt your change here, just want to understand.

pkg/skaffold/deploy/helm.go
```diff
@@ -214,8 +213,83 @@ func (h *HelmDeployer) Cleanup(ctx context.Context, out io.Writer) error {
...
-	return errors.New("not yet implemented")
+func (h *HelmDeployer) Render(ctx context.Context, out io.Writer, builds []build.Artifact, labellers []Labeller, filepath string) error {
+	hv, err := h.binVer(ctx)
+	if err != nil {
+		return fmt.Errorf("binary version: %w", err)
+	}
+
+	var output io.Writer
+
+	if filepath != "" {
```
nkubala, line 224: this looks really similar to `dumpToFileOrWriter()` in `pkg/skaffold/deploy/util.go`. would you be able to reuse that function here?

pkg/skaffold/deploy/helm.go
```diff
@@ -214,8 +213,83 @@ func (h *HelmDeployer) Cleanup(ctx context.Context, out io.Writer) error {
...
+			v, err := util.ExpandEnvTemplate(r.SetValueTemplates[key], nil)
+			if err != nil {
+				return err
+			}
+			args = append(args, "--set", fmt.Sprintf("%s=%s", key, v))
+		}
+
+		if r.Namespace != "" {
+			args = append(args, "--namespace", r.Namespace)
+		}
```
nkubala, line 285: is this all that was required for fixing https://github.com/GoogleContainerTools/skaffold/issues/3843?

in general we'd prefer splitting this out into a separate PR so we can track them separately, but I suppose it's fine to keep it here for now.

pkg/skaffold/deploy/helm.go
tete17: I think an Image is worth more than a thousands words.
dummy

LSS my IDE recommended. I can roll it back if you want

pkg/skaffold/deploy/helm.go
tete17: I doubt #3843 is fixed with this MR since I haven't touched any of the Deploy/Cleanup code.

@no1melman are you sure this PR fixes you issue?

pkg/skaffold/deploy/helm.go
tete17: Applied

pkg/skaffold/deploy/helm.go
tete17: Reused it, thanks :)

pkg/skaffold/deploy/helm.go
tete17: So this was confusing to me so I will do my best to explain it.

They way this function was written before meant that you could only have images name in your `values:` map.
Example:
```yaml
deploy:
  helm:
    releases:
    - name: myrelease
      chartPath: k8s/client
      values:
        image: "my/image/name/built/&/pushed/by/skaffold
        image2. "second/image"
```

If you  were to add a different value just to specify maybe a replica count this function would throw an error saying not all build artifacts would have been used. I think that should not be the behaviour and instead let you add more values if you desire but detect when a value corresponds to and artifact and only then replace it by the full value including the tag.

It is true that with this change you wouldn't be able to make sure all build artifacts are used but I don't think that should be a problem. I believe in the kustomize render something similar happens in that if an artifact is not being used it is simply ignored,

I may be wrong in the things stated here. If you please correct me and I will find a different way to inject the full image names (including tags) into the render function.

tete17:
@nkubala Thank you very much for the review.

I hope I didn't gave the impression of desperate :laughing: by asking for a review.

I went ahead and I think addressed all of your comment, as well as rebasing to the latest master.

Took me a bit of time but I also added some integrations tests. Let me know if you think we can add more content to them.

pkg/skaffold/deploy/helm.go
balopat: These seem fine to me, they are redundant - the `"` is not really required as `v` identifies the start of the string (although you could argue that potentially strings starting with v but not immediately after a `"` will be matched too - shouldn't matter in this output though).  Also, in this case. as we are in a character set (`[...]`), the two `\.`s are superfluous and in a character set you don't need to escape `.`.

pkg/skaffold/deploy/helm.go
nkubala: ahh ok I see what's going on here. this is an issue that people have complained about before, and one I certainly think we should address: https://github.com/GoogleContainerTools/skaffold/issues/3316.

however, I'd like to do this in a separate PR because I think we need more tests for this, and since this is addressing a separate issue than this PR is meant to. I think what you have here is a good place to start, since it would allow the deploy to at least go through when values are provided that are not images. would you mind taking this out of this PR, and submitting it in a follow up so we can track it separately? 🙏

tete17: Right on the spot!

This was exactly what I was trying to fix. I will move outside of this MR and try to tackle it separately.

Thanks for bringing it up

tete17:
I used this chance to rebase once more and cleanup the commits to make a nice git history.

Take it from here @nkubala

tete17:
@nkubala can you review?

pkg/skaffold/deploy/helm.go
```diff
@@ -51,10 +50,13 @@ import (
...
 var (
 	// versionRegex extracts version from "helm version --client", for instance: "2.14.0-rc.2"
-	versionRegex = regexp.MustCompile(`\"v(\d[\w\.\-\.]+)`)
+	versionRegex = regexp.MustCompile(`v(\d[\w.\-]+)`)
 
 	// helm3Version represents the version cut-off for helm3 behavior
 	helm3Version = semver.MustParse("3.0.0-beta.0")
+
+	// error to throw when helm version can't be determined
+	versionErrorString = "failed to determine binary version: %w"
```
tstromberg, line 59: optional: I personally prefer to see errors repeated for clarity than made into variables, but I don't care enough to demand it here.

tete17: It was an explicit request from @nkubala

tete17:
I went ahead and rebased once more to fix the conflicts introduced in https://github.com/GoogleContainerTools/skaffold/pull/4169

tete17:
My pleasure thanks to @nkubala & @tstromberg for the constant reviews.

I also wonder is we could use this chance to also move `skaffold render` out of `alpha` and into `beta` since all deployers are supported now?

PR #3971, opened 2020-04-17
Upgrade Jib to 2.2.0

chanseokoh:


loosebazooka:
I think there's some more in test-data potentially? like sync test

chanseokoh:
Good catch. I wasn't in sync with master.

PR #3992, opened 2020-04-21
install docs: use "install" and "choco -y"

### LINKED ISSUE #2948: add -y to `choco install skaffold`
`choco install skaffold -y` works better as it doesn't wait for a prompt, which can cause choco to hang.

tstromberg:
Use install(1) because it's purpose is more straight-forward. 

Use `-y` for choco to avoid hanging prompt.

Fixes #2948

**Fixes**: _Tracking issues that this PR will close_

**Description**


dgageot:
TIL!

PR #3926, opened 2020-04-07
fix a nilpointer when running status_check_Test

balopat:
**Description**

Fixes the error below by initializing the event handler. 

Running ` go test -v github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy -run "TestGetDeployStatus" ` currently fails with 

```
=== RUN   TestGetDeployStatus
=== RUN   TestGetDeployStatus/one_error
=== RUN   TestGetDeployStatus/no_error
=== RUN   TestGetDeployStatus/multiple_errors
=== RUN   TestGetDeployStatus/0_deployments
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1934520]

goroutine 12 [running]:
github.com/GoogleContainerTools/skaffold/pkg/skaffold/event.(*eventHandler).handle(0x2f6f740, 0xc0003ef110)
        /Users/balintp/skaffold/pkg/skaffold/event/event.go:419 +0xd10
created by github.com/GoogleContainerTools/skaffold/pkg/skaffold/event.(*eventHandler).handleStatusCheckEvent
        /Users/balintp/skaffold/pkg/skaffold/event/event.go:329 +0xa8
FAIL    github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy    1.049s
```

**User facing changes (remove if N/A)**
none 

**Follow-up Work (remove if N/A)**
none

PR #3993, opened 2020-04-21
fix a nilpointer when running status_check_Test

dgageot:
This is a direct copy of #3926, with fixed imports


dgageot:
Self merging small fix to a test

PR #3998, opened 2020-04-21
Fail fast with clear message if Docker is not started

dgageot:
All of the local builders will rely on a local Docker:
 + Either to build the image,
+ Or to docker load it.

Let's fail fast if Docker is not available

### DIFF
--- a/pkg/skaffold/build/local/local.go
+++ b/pkg/skaffold/build/local/local.go
@@ -74,6 +74,16 @@ func (b *Builder) buildArtifact(ctx context.Context, out io.Writer, artifact *la
 }
 
 func (b *Builder) runBuildForArtifact(ctx context.Context, out io.Writer, artifact *latest.Artifact, tag string) (string, error) {
+	if !b.pushImages {
+		// All of the builders will rely on a local Docker:
+		// + Either to build the image,
+		// + Or to docker load it.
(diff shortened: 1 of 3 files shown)


pkg/skaffold/build/local/local.go
```diff
@@ -74,6 +74,16 @@ func (b *Builder) buildArtifact(ctx context.Context, out io.Writer, artifact *la
 }
 
 func (b *Builder) runBuildForArtifact(ctx context.Context, out io.Writer, artifact *latest.Artifact, tag string) (string, error) {
+	if !b.pushImages {
+		// All of the builders will rely on a local Docker:
+		// + Either to build the image,
+		// + Or to docker load it.
+		// Let's fail fast if Docker is not available
```
briandealwis, line 81: This premise isn't true for Jib or Bazel… though Skaffold seems to fail as it tries to check the tags.

dgageot: I'm not sure I understand:
 + For Bazel, Skaffold runs a `docker load` from the tarball produced by Bazel.
 + For Jib, the local docker is required for `mvn jib:dockerBuild` to succeed

dgageot:
@briandealwis Could you take another look?

PR #3999, opened 2020-04-21
Fail fast if the Dockerfile can’t be found

dgageot:


pkg/skaffold/docker/dependencies.go
```diff
@@ -49,6 +49,14 @@ func GetDependencies(ctx context.Context, workspace string, dockerfilePath strin
 		return nil, fmt.Errorf("normalizing dockerfile path: %w", err)
 	}
 
+	// If the Dockerfile doesn't exist, we can't compute the dependencies.
+	// But since we know the Dockerfile is a dependency, let's return a list
+	// with only that file. It make errors down the line more actionnable
```
briandealwis, line 54:
```suggestion
	// with only that file. It makes errors down the line more actionable
```

PR #4004, opened 2020-04-22
Mock test to make it 4s faster

dgageot:


dgageot:
Self merging this test improvement

PR #4012, opened 2020-04-23
Add codes for error types and detect terminated containers

tejal29:
**Related** #3952 
In this change, Diagnostic library will now return an error code for each error type e. g "ImgPullErr" or "RunContainerErr"
Skaffold pod health check will consume this error code and propogate it to IDEs in `ResourceStatusCheckEvent` for them to determine what next actions could be taken.

**Description**

None.

Add `ErrorCode` to `ResourceStatusCheckEvent` proto
Hook up pod health check in `pkg/diag` with skaffold.


proto/skaffold.proto
```diff
@@ -231,4 +231,25 @@ enum ErrorCode {
 //    COULD_NOT_GET_DOCKER_CLIENT = 7;
 //    COULD_NOT_GET_LOCAL_CLUSTER = 8;
 //    BUILDER_CLEANUP = 9;
+
```
balopat, line 234: I think we could follow grouping by first digit similar to HTTP status codes (https://www.rfc-editor.org/rfc/rfc7231.txt) 
I like the container errors being 3xx, and k8s infra errors 4xx, but then NO_ERROR should be 2xx, unknown 5xx, e.g. (unknown 500, unknown unsched 501) WDYT?

proto/skaffold.proto
```diff
@@ -231,4 +231,25 @@ enum ErrorCode {
...
+    STATUS_CHECK_NO_ERROR = 300;
+    STATUS_CHECK_UNKNOWN = 301;
+    STATUS_CHECK_UNKNOWN_UNSCHEDULABLE = 302;
+
```
balopat, line 238: let's add docs for each of these, so they show up nicely in the generated docs http://34.94.185.45:1313/docs/references/api/grpc/#errorcode

proto/skaffold.proto
tejal29: sounds good. i have tried to group them, but not to the best!

PR #3933, opened 2020-04-10
nodejs example: Update nodemod dependency to v2.0.3

medyagh:
running ```npm audit fix ```  to satisfy the github security warnings

PR #4015, opened 2020-04-23
Fix nodemon versions

dgageot:
Adding more fixes to #3933


dgageot:
Self merging this small fix

PR #4016, opened 2020-04-23
Support `skaffold fix —version skaffold/v1`

### LINKED ISSUE #3613: Can `skaffold fix` take a flag to specify which schema version to transform to?
We have a few `skaffold.yaml` files at version `skaffold/v1beta15`. We want to transform them to `skaffold/v1`, but there is currently no way to do this with `skaffold fix` since it will currently transform the files to `skaffold/v2alpha2` (latest).

Perhaps something along the lines of:
`skaffold fix --filename=skaffold.yaml --version=skaffold/v1`

dgageot:
Fixes #3613

Also took the opportunity to:
+ improve the code a bit
+ add more unit tests
+ add examples of usage for `skaffold fix`
+ improve error messages

### DIFF
--- a/pkg/skaffold/schema/versions_test.go
+++ b/pkg/skaffold/schema/versions_test.go
@@ -149,7 +149,7 @@ deploy:
 	}
 }
 
-func TestParseConfig(t *testing.T) {
+func TestParseConfigAndUpgrade(t *testing.T) {
 	tests := []struct {
 		apiVersion  string
 		description string
@@ -273,7 +273,7 @@ func TestParseConfig(t *testing.T) {
 			tmpDir := t.NewTempDir().
 				Write("skaffold.yaml", fmt.Sprintf("apiVersion: %s\nkind: Config\n%s", test.apiVersion, test.config))
 
-			cfg, err := ParseConfig(tmpDir.Path("skaffold.yaml"), true)
+			cfg, err := ParseConfigAndUpgrade(tmpDir.Path("skaffold.yaml"), latest.Version)
 			if cfg != nil {
 				config := cfg.(*latest.SkaffoldConfig)
 				err := defaults.Set(config)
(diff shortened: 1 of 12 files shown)


pkg/skaffold/schema/versions_test.go
```diff
@@ -472,3 +472,27 @@ func TestCantUpgradeFromLatestVersion(t *testing.T) {
...
+			Write("skaffold.yaml", fmt.Sprintf("apiVersion: %s\nkind: Config\n%s", latest.Version, minimalConfig)).
+			Chdir()
+
+		_, err := ParseConfigAndUpgrade("skaffold.yaml", "unknown")
+
+		t.CheckErrorContains(`unknown api version: "unknown"`, err)
+	})
+}
+
+func TestParseConfigAndUpgradeToOlderVersion(t *testing.T) {
```
tejal29, line 488: maybe i misunderstood. We would to provide an ability to fix skaffold config to version older than the current latest right?
e.g. if skaffold-v1.2, the config version shipped with it is `skaffold/v1alpha2`. However users want to be `skaffold/v1`
did i get it right?

dgageot: yes, they can choose a version that's older than `latest` but not a version that's older than the version used in the skaffold.yaml. That would be a downgrade.

tejal29: hmm that would mean, if a user
1. Upgrades the version to latest using `skaffold fix` and then realizes, they updated the config to `skaffold/v1aplha2`. 
2. Now they want to go back to `skaffold/v1` version since they don't want to rely on `alpha` config. 
It is currently not possible right because `skaffold/v1alpha2` is greater than `skaffold/v1`

tejal29: i think for now, i am fine with this. But in practice to come out of this situation, users will have to
1. revert back to old config, 
2. try `skaffold fix --version`

dgageot: indeed. Supporting downgrades would be a lot of work

PR #4026, opened 2020-04-26
Enable debugging of Buildpacks

briandealwis:
**Description**

This PR adds support for Cloud Native Buildpacks for `skaffold debug`.  

At a high level, there are two fundamental changes:
  1. `pkg/skaffold/debug/transform*.go` adds support for configuring a set of _entrypoint launchers_: these are known `ENTRYPOINT`s that effectively launch the CMD value as the command-line, such that the `ENTRYPOINT` can be ignored.  The CNB launcher (`/cnb/lifecycle/launcher`) is such a launcher.  The `node` images also use a `docker-entrypoint.sh` script, which is also an entrypoint launcher. 
  2. `pkg/skaffold/debug/cnb.go` performs some resolution with the CNB image metadata to identify the command-line that will be executed, which is then passed through the normal transformations.

CNB images are normally configured with 1 or more _processes_: each process has a separate _type_ or name, and define a command and arguments.  Each process can also be marked as being _direct_, or whether it needs to be executed as an inline shell script.  If not specified, there is a default process type, `web`. CNB images specifically a launcher (`/cnb/lifecycle/launcher`) to identify and run the right command line.

`skaffold debug` has experimental support for debugging images produced by Cloud Native Buildpacks.

- Incorporate the node-wrapper from container-debug-support

### DIFF
--- a/pkg/skaffold/debug/cnb.go
+++ b/pkg/skaffold/debug/cnb.go
@@ -41,7 +41,7 @@ func init() {
 //
 //   1. _predefined processes_ are named sets of command+arguments.  There are two types:
 //      - direct: these are passed uninterpreted to os.exec; the command is resolved in PATH
-//      - script: the command is treated as a shell script and passed to `sh -c`, the remaing
+//      - script: the command is treated as a shell script and passed to `sh -c`, the remaining
 //        arguments are added to the shell, and so available as positional arguments.
 //        For example: `sh -c 'echo $0 $1 $2 $3' arg0 arg1 arg2 arg3` => `arg0 arg1 arg2 arg3`.
 //        (https://github.com/buildpacks/lifecycle/issues/218#issuecomment-567091462).
@@ -58,11 +58,11 @@ func updateForCNBImage(container *v1.Container, ic imageConfiguration, transform
 	// image was built with an older lifecycle.
 	m := cnb.BuildMetadata{}
 	// buildpacks/lifecycle 0.6.0 now embeds processes into special image label
-	if metadataJSON, found := ic.labels["io.buildpacks.build.metadata"]; !found {
+	metadataJSON, found := ic.labels["io.buildpacks.build.metadata"]
+	if !found {
 		return ContainerDebugConfiguration{}, "", fmt.Errorf("image is missing buildpacks metadata; perhaps built with older lifecycle?")
-	} else {
-		json.Unmarshal([]byte(metadataJSON), &m)
 	}
+	json.Unmarshal([]byte(metadataJSON), &m)
 	if len(m.Processes) == 0 {
 		return ContainerDebugConfiguration{}, "", fmt.Errorf("buildpacks metadata has no processes")
 	}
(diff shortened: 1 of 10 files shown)


pkg/skaffold/debug/cnb.go
```diff
@@ -0,0 +1,144 @@
...
+			var rewriter func(transformed []string) []string
+			if p.Direct {
+				// p.Command is the command and p.Args are the arguments
+				ic.arguments = append([]string{p.Command}, p.Args...)
+				rewriter = func(transformed []string) []string {
+					return append([]string{"--"}, transformed...)
+				}
+			} else {
+				// Script type: split p.Command, pass it through the transformer, and then reassemble in the rewriter.
+				if args, err := shell.Split(p.Command); err == nil {
```
dgageot, line 112: nit: handle the `err != nil` case first

pkg/skaffold/debug/cnb.go
```diff
@@ -0,0 +1,144 @@
...
+	if len(ic.arguments) == 0 {
+		// indicates an image mis-configuration as we should have resolved the the
+		// CNB_PROCESS_TYPE (if specified) or `web`.
+		logrus.Warnf("no CNB launch found for %s/%s", ic.artifact, processType)
+		return ic, nil
+	}
+
+	// ic.arguments[0] is a shell script:  split it, pass it through the transformer, and then reassemble in the rewriter.
+	var rewriter func(transformed []string) []string
+	if args, err := shell.Split(ic.arguments[0]); err == nil {
```
dgageot, line 135: What about the error case?

pkg/skaffold/debug/cnb.go
```diff
@@ -0,0 +1,144 @@
...
+		remnants := ic.arguments[1:]
+		ic.arguments = args
+		rewriter = func(transformed []string) []string {
+			// reassemble back into a script with arguments
+			return append([]string{shJoin(transformed)}, remnants...)
+		}
+	}
+	return ic, rewriter
```
dgageot, line 143: The rewriter is `nil` here. Is that ok?

pkg/skaffold/debug/transform_go.go
```diff
@@ -61,10 +61,27 @@ func isLaunchingDlv(args []string) bool {
...
 			return true
 		}
 	}
-	if len(config.entrypoint) > 0 {
+
+	// FIXME: as there is currently no way to identify a buildpacks-produced image as holding a Go binary,
+	// nor to cause certain environment variables to be defined in the resulting image, look at the image's
+	// CNB metadata to see if any well-known Go-related buildpacks had been involved.
+	knownGoBuildpackIds := []string{
+		"google.go.runtime", "google.go.build", "google.go.gopath", "google.go.gomod", // GCP Buildpacks
```
dgageot, line 73: `google.go.build` is always on, all the other on that line are optional ans can't run without `google.go.build`

pkg/skaffold/debug/cnb.go
briandealwis: There's not much we can do, but I'll add a comment.

pkg/skaffold/debug/cnb.go
briandealwis: Yeah, it's checked in the caller.

briandealwis:
PTAL

briandealwis:
PTAL @dgageot

briandealwis:
(Last two commits were to add some debug logging for TestEventsRPC failures.)

briandealwis:
Committed with the event logging in TestEventsRPC as it was useful for catching the errors, and is hidden on successful tests.

PR #4027, opened 2020-04-26
Skip podspecs that already have a debug.cloud.google.com/config annotation

briandealwis:
Skip transforming a pod spec if it already has our debug annotation.  This ability is useful if someone wants to explicitly configure debugging for a container image such as to customize the `dlv` command-line used for a Go app.


pkg/skaffold/debug/transform.go
```diff
@@ -178,6 +178,11 @@ func transformManifest(obj runtime.Object, retrieveImageConfiguration configurat
 // transformPodSpec attempts to configure a podspec for debugging.
 // Returns true if changed, false otherwise.
 func transformPodSpec(metadata *metav1.ObjectMeta, podSpec *v1.PodSpec, retrieveImageConfiguration configurationRetriever) bool {
+	// skip annotated podspecs — allows users to customize their own image
```
tejal29, line 181: Shd we also add this to debug docs that users can use the "debug.cloud.google.com/config" to skip skaffold transforming pods?

briandealwis: Good idea.

briandealwis: I'll do this separately as there are a few other tweaks to the debug docs.

PR #4053, opened 2020-04-29
Handle port forwarding rules in sequence

### LINKED ISSUE #3563: Enable only User Defined Port Forwarding

dgageot:
First, user defined rules.
Then, automatic rules.

Fixes #3563

### DIFF
--- a/pkg/skaffold/kubernetes/portforward/resource_forwarder.go
+++ b/pkg/skaffold/kubernetes/portforward/resource_forwarder.go
@@ -66,12 +66,11 @@ func (p *ResourceForwarder) Start(ctx context.Context) error {
 
 // Port forward each resource individually in a goroutine
 func (p *ResourceForwarder) portForwardResources(ctx context.Context, resources []*latest.PortForwardResource) {
-	for _, r := range resources {
-		r := r
-		go func() {
+	go func() {
+		for _, r := range resources {
 			p.portForwardResource(ctx, *r)
-		}()
-	}
+		}
+	}()
 }
 
 func (p *ResourceForwarder) portForwardResource(ctx context.Context, resource latest.PortForwardResource) {


pkg/skaffold/kubernetes/portforward/resource_forwarder.go
```diff
@@ -66,12 +66,11 @@ func (p *ResourceForwarder) Start(ctx context.Context) error {
 
 // Port forward each resource individually in a goroutine
 func (p *ResourceForwarder) portForwardResources(ctx context.Context, resources []*latest.PortForwardResource) {
-	for _, r := range resources {
-		r := r
-		go func() {
+	go func() {
```
tejal29, line 69: Not sure, if we are slowing down skaffold portforward phase due to sequentially forwarding resources.

Other alternative would be at L63, we first port forward `uderDefinedResources` in parallel. Wait for it to complete and then portforward serviceResources.

I don't mind this approach either.

dgageot: Thanks for the idea @tejal29. You know what, let's go with correctness and if we have an issue, we'll improve the perf.

Issue #2170, opened 2019-05-24
`skaffold debug`: support npm-nodemon


briandealwis:
#2141 has one caveat:
> Note: npm run-scripts that use `nodemon` (a node-based script) instead of node will not work as `nodemon`'s node will interpret the `NODE_OPTIONS` and configure itself for debugging.

This npm-nodemon style of launch seems useful, but it's unclear if `nodemon` makes sense with use with a debugger as DevTools seems to be able to upload and install file changes on the server.

If we do need to support npm invoking nodemon, I think we'll need to take a different approach where we would wrap `node` with a script that would determine whether it is launching an _application script_ or a _library script_ from `node_modules` such as `nodemon`.  If a library script, then it should not run with `--inspect` and somehow defer it to a child node.

briandealwis:
I've attached [a proof-of-concept `node`-wrapper script](https://github.com/GoogleContainerTools/skaffold/files/3219061/node-wrapper.sh.txt) here that propagates an `--inspect`-style argument to a child `node` that invokes an application script, skipping intermediate `node` executions of library scripts like `nodemon`.  The `--inspect` argument is propagated via a `NODE_DEBUG` environment variable.

This `node` wrapper parses the command-line arguments and `NODE_OPTIONS` (set by `npm --node-options="--inspect" ...`) for an `--inspect` argument and the script to be executed.
  - if the script is an application script then we invoke the real `node` with the `--inspect` argument as conveyed through `NODE_DEBUG` (if any)
  - if invoked with `--inspect` argument then set `NODE_DEBUG` to the value and strip it from the command-line
  - if the provided script is `.../node_modules/.../nodemon` then we run `nodemon $NODE_DEBUG`
      - we need to do this as `nodemon --inspect` will ensure that it uses `spawn` to invoke the child `node`, which will pick up our wrapped node.  Otherwise it uses `fork` to launch the actual application script file directly, which circumvents the use of the `node` wrapper.
  - Otherwise the provided script is in `node_modules` and we invoke the real `node` with the provided script

(I'm pretty sure this script does not properly handle arguments with spaces; I wanted to avoid assuming bash.)

briandealwis:
This affects Skaffold's own [`examples/nodejs`](https://github.com/GoogleContainerTools/skaffold/tree/master/examples/nodejs).

[...]

PR #4086, opened 2020-05-03
Use `node` wrapper to debug NodeJS apps

briandealwis:
Fixes: #2170

**Description**

Cause NodeJS apps to use the newly-introduced `node` wrapper.

Many applications use NodeJS-based tools as part of their launch, like `npm`, rather than
invoke `node` directly.  These intermediate `node` instances may interpret the
`--inspect` arguments.  GoogleContainerTools/container-debug-support#34 introduced a `node` wrapper that only passes `--inspect` when provided an application script, and otherwise propagates `--inspect` via an environment variable.

One other minor change: this now sorts the environment variables when populating the k8s `Container` objects to make comparisons stable in the tests.

**User facing changes **

NodeJS apps that previously seemed undebuggable should now respond to debugging treatment.


briandealwis:
We live at `latest`.  There are tests in [integration/testdata/debug](https://github.com/GoogleContainerTools/skaffold/tree/master/integration/testdata/debug) for `nodejs` and `npm`.  They wait for the deployments to be stable.  Hmm, I should define some readiness probes there.

PR #4089, opened 2020-05-03
Enable file-watching for `debug`

### LINKED ISSUE #2351: Enable filesync with debug
`debug` currently disables the Skaffold watcher to prevent Skaffold from redeploying artifacts on file changes: having a debug session ripped out when in progress is a suboptimal user experience!  And many runtimes support installing changes, normally handled by the IDE.

But a side-effect is that this also disables the filesync functionality.  We should be able to forego artifact redeployment while still supporting filesync.

### LINKED ISSUE #3272: 'skaffold debug' should support build triggers
`skaffold debug` is awesome, but as far as I understand it doesn't support "redeploy-on-change"—meaning that to debug a change one has to stop and restart `skaffold debug`, which may be a bit awkward. It would be great if `skaffold debug` would behave a bit more like `skaffold dev`. Thoughts?

### LINKED ISSUE #4129: WithFlags() change should defer applying default values until the command is to be used
#4089 seeks to add `dev`'s `--auto-{build,deploy,sync}` options to `skaffold debug` but [with different default values](https://github.com/GoogleContainerTools/skaffold/pull/4089/files#diff-e38cfd6902a46a3295cb4067d8158878R37-R49).  But the obvious (or naive? :grinning:) attempt fails as Skaffold's commands framework immediately applies the function passed to `WithFlags()`:

https://github.com/GoogleContainerTools/skaffold/blob/86b4fde68f9b1530f09fc304e52841ca3efdb522/cmd/skaffold/app/cmd/commands.go#L78-L80

Since `NewCmdDebug` is processed after `NewCmdDev`:

https://github.com/GoogleContainerTools/skaffold/blob/86b4fde68f9b1530f09fc304e52841ca3efdb522/cmd/skaffold/app/cmd/cmd.go#L124-L128

the `debug` defaults win and so the `dev` integration tests fail as auto-{build/deploy/sync} are disabled.

For now I'll change `CmdNewDebug` to instead record these values locally and then overwrite the `opts.Auto*` values in the `doDebug()`, but it seems awkward.

briandealwis:
Fixes #2351
Fixes #3272 
Fixes #4129  

**Description**

Pulls in the `dev`-like arguments to control image rebuilding and redeploying such as:
  - `--watch-image` to restrict the images to be monitored
  - `--trigger` — for this I kept the same "notify" default, but I wonder if manual would make more sense
  - `--auto-build`, `--auto-deploy`, and `--auto-sync` but defaulting to `false`.

This at least allows users to enable rebuilding and redeploying on change, which is useful for Go in particular.

### User facing changes

`skaffold debug` now supports listing images to watch and a watch-trigger, though they are essentially non-functional and not worth mentioning since `--auto-build`, `--auto-deploy`, and `--auto-sync` default to `false`, and are hidden.


briandealwis:
PTAL @dgageot

cmd/skaffold/app/cmd/debug.go
```diff
@@ -21,26 +21,51 @@ import (
...
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
 	debugging "github.com/GoogleContainerTools/skaffold/pkg/skaffold/debug"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy"
 )
 
 // NewCmdDebug describes the CLI command to run a pipeline in debug mode.
+// Unlike `dev`, `debug` defaults `auto-build` and `auto-deploy` to `false`.
```
dgageot, line 31: I am not sure that same flag can have different values based on the command. Could you make sure it really works?

briandealwis:
Yeah, it works because it uses a different temporary variable.  When I start `scaffold debug` in `examples/getting-started`, and then make a change to `main.go`, there are no rebuilds.

But you're right — I should make a test here.

briandealwis:
PTAL @dgageot — added some tests to verify that the arguments are set and are independent of each other.  I de-aliased `dev` too so that `NewCmdDebug` and `NewCmdDev` can be called independently of each other.

briandealwis:
It turns out that our runner creation depends on the implicit initialization of the trigger type from `Dev()`:
```
TestDeploy: helper.go:194: skaffold deploy: exit status 1, 
    TestDeploy: panic.go:617: creating runner: creating watch trigger: unsupported trigger: 
```

cmd/skaffold/app/cmd/debug.go
```diff
@@ -21,26 +21,54 @@ import (
...
+	var targetImages []string
+	var watchPollInterval int
+
 	return NewCmd("debug").
 		WithDescription("[beta] Run a pipeline in debug mode").
 		WithLongDescription("Similar to `dev`, but configures the pipeline for debugging.").
 		WithCommonFlags().
-		NoArgs(doDebug)
+		WithFlags(func(f *pflag.FlagSet) {
+			f.StringVar(&trigger, "trigger", "notify", "How is change detection triggered? (polling, notify, or manual)")
```
dgageot, line 47: Can those flags be added to the common flags list, now that they are used by 2 commands?

cmd/skaffold/app/cmd/debug_test.go
```diff
@@ -37,3 +41,47 @@ func TestNewCmdDebug(t *testing.T) {
...
+// https://github.com/GoogleContainerTools/skaffold/issues/4129
+// https://github.com/spf13/pflag/issues/257
+func TestDebugIndependentFromDev(t *testing.T) {
+	mockRunner := &mockDevRunner{}
+	testutil.Run(t, "DevDebug", func(t *testutil.T) {
+		t.Override(&createRunner, func(config.SkaffoldOptions) (runner.Runner, *latest.SkaffoldConfig, error) {
+			return mockRunner, &latest.SkaffoldConfig{}, nil
+		})
+		t.Override(&opts, config.SkaffoldOptions{})
+		t.Override(&doDev, func(_ context.Context, _ io.Writer) error {
```
dgageot, line 55:
```suggestion
		t.Override(&doDev, func(context.Context, io.Writer) error {
```

cmd/skaffold/app/cmd/debug_test.go
```diff
@@ -37,3 +41,47 @@ func TestNewCmdDebug(t *testing.T) {
...
+			}
+			if !opts.AutoDeploy {
+				t.Error("opts.AutoDeploy should be true for dev")
+			}
+			if !opts.AutoSync {
+				t.Error("opts.AutoSync should be true for dev")
+			}
+			return nil
+		})
+		t.Override(&doDebug, func(_ context.Context, _ io.Writer) error {
```
dgageot, line 67:
```suggestion
		t.Override(&doDebug, func(context.Context, io.Writer) error {
```

cmd/skaffold/app/cmd/debug.go
briandealwis: oh right 🤦

briandealwis:
Using the common flags is much nicer.

PR #4091, opened 2020-05-04
Rename "Builder" message to “ImageBuilder” in proto

### LINKED ISSUE #4056: Do not use "Builder" as a message name for Skaffold proto
When we consume Skaffold proto file in Cloud Code IJ (and compile it into Java code), we now have a conflict with `Builder` name being already used in any protoc-compiled class. Java generated code heavily uses `Builder` pattern and each structure will have its own `Builder` associated to it. We suggest renaming to something like `ContainerBuilder`:

```
...
message BuildMetadata {
  message ContainerBuilder {
      BuilderType type = 1;
      int32 count = 2;
  }
...
```

dgageot:
Fixes #4056


dgageot:
ping @ivanporty @balopat

PR #4103, opened 2020-05-05
Verify `debug` integration tests using readiness and liveness probes

briandealwis:
Motivated in part by the following comment https://github.com/GoogleContainerTools/skaffold/pull/4086#issuecomment-623797576, this PR updates the `integration/testdata/debug` projects to define startup and liveness probes to verify that the underlying runtime debugger endpoints and application endpoints are established.  These probes use Kubernetes to probe the pod to ensure the debugger runtime is setup and that the application is running and serving.

The `tcpSocket` startup probe for the Jib test causes the JVM to emit a _handshake failed_ message which is expected and harmless.

### DIFF
--- a/integration/testdata/debug/go/k8s/pod.yaml
+++ b/integration/testdata/debug/go/k8s/pod.yaml
@@ -8,3 +8,16 @@ spec:
     image: skaffold-debug-go
     ports:
     - containerPort: 8080
+    # connect to the dlv APIv2 port
+    startupProbe:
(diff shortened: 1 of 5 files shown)


briandealwis:
Oh our integration tests are against Kubernetes 1.13.  No startupProbes.

briandealwis:
Really odd: the probers are erroring getting connection refused when trying to access the container ports, but I can `kubectl port-forward` with no problem.

PR #4118, opened 2020-05-06
Better error message when image can’t be pushed

dgageot:


dgageot:
@briandealwis I don't care either way. FYIW, Google's guidelines is to use contractions to sound more friendly to the user :-)

PR #4141, opened 2020-05-11
Run the linters in their own TravisCI job

dgageot:
Let's try to see how the linter behaves as the only task in a CI job, without caching.


dgageot:
I'm going to self merge this PR because it makes the CI more stable

Issue #2806, opened 2019-09-04
Pause continuous deployment


balopat:
Add the ability to change the state of `autoBuild/autoDeploy/autoSync` with a single Control API call.

PR #4145, opened 2020-05-11
Add control API to pause and resume autoBuild, autoDeploy and autoSync

gsquared94:
Fixes: #2806  
**Description**

Add control API to pause and resume autoBuild, autoDeploy and autoSync

**API Contract**
image


gsquared94:
> I tried this out locally, and I'm seeing strange behavior with the `microservices` example. when I set `auto-deploy` to false with `curl -X POST http://localhost:50052/v1/auto_execute -d '{"deploy": false}'`, changes aren't triggering a rebuild when they should be. at first glance I don't see any copy paste errors, but can you try this example out and see if you're seeing the same issue?
> 
> also, it would be great if we could expose the state of each "switch" in the exposed `state` object. sometimes I lose track of where I am, and it would be nice to have a way to tell. this would also be useful for the IDEs to integrate this so they wouldn't have to keep track in the client (though they may anyway). this could come in a follow up PR though

I think that's because `{"deploy": false}` is equivalent to `{"build": false, "sync": false, "deploy": false}` since the missing parameters in rpc request is set to default value AFAIK. That's something I wanted your opinion on. Either we use struct params (instead of boolean) in `Intent` otherwise we'll need 3 API's for the three phases separately. Which do you think is better?

gsquared94:
> also, it would be great if we could expose the state of each "switch" in the exposed `state` object. sometimes I lose track of where I am, and it would be nice to have a way to tell. this would also be useful for the IDEs to integrate this so they wouldn't have to keep track in the client (though they may anyway). this could come in a follow up PR though

It should be there as `AutoTrigger` property in `BuildState`, `DeployState` and `FileSyncState` already. The API doesn't return params if they are their default value. So if you ran with `auto-build=false` then the state api response won't show `AutoTrigger` param for `BuildState`.

PR #4162, opened 2020-05-12
improve deployment waiting logic in integration tests

balopat:
- Wait was too long, in `WaitForDeploymentsToStabilize` - it took 5 mins to fail a test, now it's 30 seconds
- Failure was hard to trace back to the test, as there were missing t.Helper() invocations in the chain
- `WaitForDeploymentsToStabilize` was not checking availability - `TestBuildAndDeploy` never really worked in kind - `skaffold build` first builds (but doesn't load to Kind! or push!) then we call `skaffold deploy` which works, but the pods always fail with `ErrImgPull` - we just didn't see it fail as `WaitForDeploymentsToStabilize` was only checking replicacount, and not available replica count.

### DIFF
--- a/integration/deploy_test.go
+++ b/integration/deploy_test.go
@@ -26,8 +26,8 @@ import (
 )
 
 func TestBuildDeploy(t *testing.T) {
-	if testing.Short() || RunOnGCP() {
-		t.Skip("skipping kind integration test")
+	if testing.Short() || !RunOnGCP() {
+		t.Skip("skipping GCP integration test")
 	}
 
 	ns, client := SetupNamespace(t)
--- a/integration/dev_test.go
+++ b/integration/dev_test.go
@@ -26,13 +26,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/clientcmd"
+
 	"github.com/GoogleContainerTools/skaffold/integration/skaffold"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
 	"github.com/GoogleContainerTools/skaffold/proto"
 	"github.com/GoogleContainerTools/skaffold/testutil"
-	"github.com/sirupsen/logrus"
-	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/tools/clientcmd"
 )
 
 func TestDevNotification(t *testing.T) {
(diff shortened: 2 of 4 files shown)


integration/dev_test.go
```diff
@@ -26,16 +26,16 @@ import (
...
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/tools/clientcmd"
-
 	"github.com/GoogleContainerTools/skaffold/integration/skaffold"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
 	"github.com/GoogleContainerTools/skaffold/proto"
 	"github.com/GoogleContainerTools/skaffold/testutil"
+	"github.com/sirupsen/logrus"
```
dgageot, line 33: This breaks the linter

integration/deploy_test.go
```diff
@@ -26,8 +26,8 @@ import (
 )
 
 func TestBuildDeploy(t *testing.T) {
-	if testing.Short() || RunOnGCP() {
-		t.Skip("skipping kind integration test")
+	if testing.Short() || !RunOnGCP() {
```
dgageot, line 29: Should we fix something in `skaffold deploy` with kind, instead?

balopat: skaffold deploy needs a registry to pull the images from, but I don't want to introduce credentials - as that causes all sorts of issues in travis (contributors can't run these tests, only maintainers). We can run a registry inside kind and that is an interesting use case that we should work out as well, but it will be a bit painful to setup, if possible at all today - how would we "tag" images that use the same registry address from inside the cluster and outside of it? Typically outside would be `localhost:xyz/image`, while from within `registry/image`. I'm actually going to open an issue for this. But not sure we want to tackle this in this PR. 
A third option could be to spin up a local registry on the node - that is possible but feels finicky as well, again, not sure how the cluster will be able to access it. 
A fourth option could be running a public insecure registry :) 
A fifth could be having an open credentials situation.
I'm not a fan of 4/5 for security/abuse protection reasons. 
Let me know if I missed something.

integration/dev_test.go
balopat: ah, yeah, it's time to setup in my new IDE the filewatchers for `goimports -w -local github.com/GoogleContainerTools/skaffold` :)

integration/deploy_test.go
balopat: Another group of ideas could be to load into kind from skaffold build and enable that functionality - again a bit bigger in scope than what I wanted to achieve here.

PR #4171, opened 2020-05-13
move global args to be applied to version call

ryanhagen:
**Description**
Executing `helm version` will check the server side namespace `kube-system`. In my particular use case of skaffold, I will not have access to `kube-system`, therefore, global variable `--tiller-namespace` will have to be applied to the `helm version` command, hence this PR.


ryanhagen:
@googlebot I fixed it.

pkg/skaffold/deploy/helm.go
```diff
@@ -220,9 +220,9 @@ func (h *HelmDeployer) Render(context.Context, io.Writer, []build.Artifact, []La
 
 // exec executes the helm command, writing combined stdout/stderr to the provided writer
 func (h *HelmDeployer) exec(ctx context.Context, out io.Writer, useSecrets bool, args ...string) error {
+	args = append(args, h.Flags.Global...)
```
tejal29, line 223: @ryanhagen please add a another test case where flags is not empty here https://github.com/GoogleContainerTools/skaffold/blob/49888e0abdd6249c18fdb311970a1d7734f48b1a/pkg/skaffold/deploy/helm_test.go#L734

ryanhagen: See below commit

tejal29:
@ryanhagen the travis job is failing due to lint errors. can you please fix ?

briandealwis:
I don't think this PR is necessary as #4294 causes Skaffold to now use `helm version --client` to avoid contacting the cluster.

tejal29:
@ryanhagen Can we close this PR since #4294  is in as @briandealwis mentioned?

ryanhagen:
@tejal29 sounds good to me.

PR #4181, opened 2020-05-14
Exclude tmpcharts folder and helm generated .lock files from list of watched files

gsquared94:
Exclude `tmpcharts` folder and helm generated `.lock` files from list of watched files

Fixes: #3565  
**Description**

`helm dep up` makes changes to the `tmpcharts` in addition to `charts` subdirectory within a chart package. It also modifies `requirements.lock` file (helm 2) and `Chart.lock` file (helm 3). So we exclude these file from the file change monitor when `skipBuildDependencies` is not set to `true`

This is consistent with existing behavior of excluding `charts` subdirectory similarly.

### DIFF
--- a/pkg/skaffold/deploy/helm.go
+++ b/pkg/skaffold/deploy/helm.go
@@ -162,14 +162,28 @@ func (h *HelmDeployer) Dependencies() ([]string, error) {
 			continue
 		}
 
-		chartDepsDir := filepath.Join(release.ChartPath, "charts")
+		chartDepsDirs := []string{
+			"charts",
+			"tmpcharts",
+		}
 
-		// We can always add a dependency if it is not contained in our chartDepsDir.
+		// We can always add a dependency if it is not contained in our chartDepsDirs.
 		// However, if the file is in our chartDepsDir, we can only include the file
 		// if we are not running the helm dep build phase, as that modifies files inside
 		// the chartDepsDir and results in an infinite build loop.
 		isDep := func(path string, info walk.Dirent) (bool, error) {
-			return !info.IsDir() && (!strings.HasPrefix(path, chartDepsDir) || r.SkipBuildDependencies), nil
+			if info.IsDir() {
+				return false, nil
+			}
+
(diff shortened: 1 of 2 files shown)


pkg/skaffold/deploy/helm.go
```diff
@@ -162,14 +162,43 @@ func (h *HelmDeployer) Dependencies() ([]string, error) {
...
 		isDep := func(path string, info walk.Dirent) (bool, error) {
-			return !info.IsDir() && (!strings.HasPrefix(path, chartDepsDir) || r.SkipBuildDependencies), nil
+			if info.IsDir() {
+				return false, nil
+			}
+			if r.SkipBuildDependencies {
+				return true, nil
+			}
+
+			for _, v := range chartDepsDirs {
```
tejal29, line 189: Would this miss any other `requirements.lock` say in a python app?

pkg/skaffold/deploy/helm.go
```diff
@@ -162,14 +162,43 @@ func (h *HelmDeployer) Dependencies() ([]string, error) {
...
+				if strings.HasPrefix(path, filepath.Join(release.ChartPath, v)) {
+					return false, nil
+				}
+			}
+
+			for _, v := range lockFiles {
+				if strings.EqualFold(info.Name(), v) {
```
tejal29, line 196: Would this miss any other requirements.lock say in a python app? 
Shd we look for `reuirements.lock` in the chartPath?

tejal29: Can we also a test case for this?

gsquared94: are you saying that I might miss some `.lock` file or that I'll erroneously include it?
I'm currently removing the `.lock` files by name without depending on the path and AFAIK `Chart.lock` and `requirements.lock` are colocated with the respective `.yaml` files, which is always in `chartPath

gsquared94: discussed offline

PR #4188, opened 2020-05-15
Remove duplication

dgageot:


pkg/skaffold/runner/build_deploy.go
```diff
@@ -127,11 +120,20 @@ func (r *SkaffoldRunner) DeployAndLog(ctx context.Context, out io.Writer, artifa
 		}
 	}
 
-	<-ctx.Done()
+	if r.runCtx.Opts.Tail || r.runCtx.Opts.PortForward.Enabled {
```
balopat, line 123: why do we need this if?

dgageot: Because, without logs or port forwarding, we don't wait

dgageot:
@balopat Can you look again?

PR #4200, opened 2020-05-16
Common flags: simpler code and no init() function

dgageot:


dgageot:
Self merging this small one

PR #4204, opened 2020-05-16
A default deployer is guaranteed to be set

dgageot:
This is guaranteed by the defaults mechanism.


dgageot:
Self merging minor change

PR #4205, opened 2020-05-16
Improve `skaffold debug` integration tests

briandealwis:
This was supposed to be a simple change: add readiness and liveness checks for the `debug` integration tests to ensure:

  1. the debugged app was launched and responds to normal HTTP fetches,
  2. the underlying debugger was started and could be connected to.

Java was a bit of a pain as a simple TCP prober to the JDWP port results in an handshake error message being logged, which feels a bit disconcerting.  So we instead connect with `jdb`, the command-line debugger.  Which doesn't have useful output and always exits with 0 regardless of whether it was able to attach.  So we wrap `jdb` too

This task became more complicated as `skaffold debug` was defaulting to configuring the debuggers to listen on `localhost`.  But this meant the probers could not connect as they come from outside of the pod.  So this PR also changes the debuggers to listen on INADDR_ANY / 0.0.0.0, such that the probers can connect.


briandealwis:
@kschaab fyi

PR #4206, opened 2020-05-18
Show suggestions for every command

dgageot:
Show suggestions for every command, not just those who don't take args.


dgageot:
ping @tejal29. This is to make your code applies to every type of command. Is that ok?

PR #4212, opened 2020-05-18
Retrieve the proper kind cluster name

### LINKED ISSUE #4198: Kind image loading is broken in TestDev_WithKubecontextOverride
Kind loading is broken in `TestDev_WithKubecontextOverride` - it cannot be called on its own. 

```
--- FAIL: TestDev_WithKubecontextOverride (0.91s)
/integration/TestDev_WithKubecontextOverride/skaffold_run_with_kubecontext_override
time="2020-05-15T18:02:39-07:00" level=info msg="Namespace: skaffoldbkrlk"
time="2020-05-15T18:02:39-07:00" level=info msg="[skaffold run --default-repo gcr.io/k8s-skaffold --kube-context kind-modified-context]"
    TestDev_WithKubecontextOverride/skaffold_run_with_kubecontext_override: helper.go:194: skaffold run: exit status 1, Generating tags...
         - skaffold-example -> gcr.io/k8s-skaffold/skaffold-example:v0.36.0-2130-g78b1b9fc4
        Checking cache...
         - skaffold-example: Found. Tagging
        Tags used in deployment:
         - skaffold-example -> gcr.io/k8s-skaffold/skaffold-example:fe2e30a7f7395ff45b7ece353b2ebdf137541dc7a64f3547e998cfc5fb8cf6ed
        Loading images into kind cluster nodes...
         - gcr.io/k8s-skaffold/skaffold-example:fe2e30a7f7395ff45b7ece353b2ebdf137541dc7a64f3547e998cfc5fb8cf6ed -> Failed
    TestDev_WithKubecontextOverride/skaffold_run_with_kubecontext_override: panic.go:617: loading images into kind nodes: unable to load image with kind "gcr.io/k8s-skaffold/skaffold-example:fe2e30a7f7395ff45b7ece353b2ebdf137541dc7a64f3547e998cfc5fb8cf6ed": running [kind load docker-image --name modified-context gcr.io/k8s-skaffold/skaffold-example:fe2e30a7f7395ff45b7ece353b2ebdf137541dc7a64f3547e998cfc5fb8cf6ed]
         - stdout: ""
         - stderr: "Error: unknown cluster \"modified-context\"\n"
         - cause: exit status 1, 
    --- FAIL: TestDev_WithKubecontextOverride/skaffold_run_with_kubecontext_override (0.91s)
```

I'm thoroughly puzzled still why it can pass with other tests around / before it. 
My best bet is that somehow artifact caching makes it skip the `kind load` - but I couldn't reliably reproduce it yet locally.

dgageot:
Fixes #4198

### DIFF
--- a/integration/dev_test.go
+++ b/integration/dev_test.go
@@ -266,7 +266,6 @@ func replaceInFile(target, replacement, filepath string) ([]byte, os.FileMode, e
 }
 
 func TestDev_WithKubecontextOverride(t *testing.T) {
-	t.Skip("Skipping due to error in this logic: see https://github.com/GoogleContainerTools/skaffold/issues/4198")
 	MarkIntegrationTest(t, CanRunWithoutGcp)
 
 	testutil.Run(t, "skaffold run with kubecontext override", func(t *testutil.T) {
@@ -295,7 +294,7 @@ func createModifiedKubeconfig(namespace string) ([]byte, string, error) {
 	}
 
 	contextName := "modified-context"
-	if isKind, _ := config.IsKindCluster(kubeConfig.CurrentContext); isKind {
+	if config.IsKindCluster(kubeConfig.CurrentContext) {
 		contextName = "kind-" + contextName
 	}
 
--- a/pkg/skaffold/config/util.go
+++ b/pkg/skaffold/config/util.go
@@ -199,31 +199,50 @@ func isDefaultLocal(kubeContext string) bool {
 		return true
 	}
 
-	isKind, _ := IsKindCluster(kubeContext)
-	return isKind
+	return IsKindCluster(kubeContext)
 }
 
 // IsKindCluster checks that the given `kubeContext` is talking to `kind`.
-// It also returns the name of the `kind` cluster.
-func IsKindCluster(kubeContext string) (bool, string) {
+func IsKindCluster(kubeContext string) bool {
 	switch {
 	// With kind version < 0.6.0, the k8s context
 	// is `[CLUSTER NAME]@kind`.
 	// For eg: `cluster@kind`
 	// the default name is `kind@kind`
 	case strings.HasSuffix(kubeContext, "@kind"):
-		return true, strings.TrimSuffix(kubeContext, "@kind")
+		return true
 
 	// With kind version >= 0.6.0, the k8s context
 	// is `kind-[CLUSTER NAME]`.
 	// For eg: `kind-cluster`
 	// the default name is `kind-kind`
 	case strings.HasPrefix(kubeContext, "kind-"):
-		return true, strings.TrimPrefix(kubeContext, "kind-")
+		return true
 
 	default:
-		return false, ""
+		return false
 	}
+}
+
+// KindClusterName returns the internal kind name of a kubernetes cluster.
+func KindClusterName(clusterName string) string {
+	switch {
+	// With kind version < 0.6.0, the k8s context
+	// is `[CLUSTER NAME]@kind`.
+	// For eg: `cluster@kind`
+	// the default name is `kind@kind`
+	case strings.HasSuffix(clusterName, "@kind"):
+		return strings.TrimSuffix(clusterName, "@kind")
+
+	// With kind version >= 0.6.0, the k8s context
+	// is `kind-[CLUSTER NAME]`.
+	// For eg: `kind-cluster`
+	// the default name is `kind-kind`
+	case strings.HasPrefix(clusterName, "kind-"):
+		return strings.TrimPrefix(clusterName, "kind-")
+	}
+
+	return clusterName
 }
 
 func IsUpdateCheckEnabled(configfile string) bool {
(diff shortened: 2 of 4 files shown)


integration/dev_test.go
```diff
@@ -295,7 +294,7 @@ func createModifiedKubeconfig(namespace string) ([]byte, string, error) {
 	}
 
 	contextName := "modified-context"
-	if isKind, _ := config.IsKindCluster(kubeConfig.CurrentContext); isKind {
+	if config.IsKindCluster(kubeConfig.CurrentContext) {
```
balopat, line 297: I think this logic is just broken. 
Our kind cluster will be `modified-context` that leads to the error that can't load the image to `modified-context` cluster as it does not exist. Maybe we should rename the kind cluster to `modified-context`?

balopat:
wow not sure what happened with the integration tests, suddenly all of them wanted to push after build and failed with auth errors! I bounced "partition 2" to see if it goes away or not

integration/dev_test.go
dgageot: I don't think that's what's happening:
+ Before we duplicate the kube context, we have a context that's usually `kind-kind` that references a cluster with the same name. To load images into kind, we have to get rid of the `kind-` prefix or suffix.
+ After we duplicate, the context is `kind-modified-context` but the cluster is still `kind-kind`. But the previous code never retrieved the name of the cluster. It assumed it was the same as the context. Which is the case by default but can be changed by the user.
+ What this code does is retrieve the actual cluster name, removed its prefix/suffix in order to get the *internal* name of the cluster

balopat: L222 did exactly that prefix removal: `return true, strings.TrimPrefix(kubeContext, "kind-")` I thought you just refactored things around but did not change the logic - am I missing something?

balopat: > After we duplicate, the context is `kind-modified-context` but the cluster is still `kind-kind`. But the previous code never retrieved the name of the cluster. It assumed it was the same as the context. Which is the case by default but can be changed by the user.

This code does not retrieve the name of the cluster either (in the test). 
The production code still uses the same prefix trimming logic - thus will still "think" the cluster is named `modified-context` instead of `kind` - unless

pkg/skaffold/config/util_test.go
```diff
@@ -291,22 +291,40 @@ func TestIsDefaultLocal(t *testing.T) {
...
+func TestKindClusterName(t *testing.T) {
+	tests := []struct {
+		kubeCluster  string
+		expectedName string
+	}{
+		{kubeCluster: "kind", expectedName: "kind"},
+		{kubeCluster: "kind-kind", expectedName: "kind"},
+		{kubeCluster: "kind-other", expectedName: "other"},
+		{kubeCluster: "kind@kind", expectedName: "kind"},
+		{kubeCluster: "other@kind", expectedName: "other"},
```
balopat, line 321: we should add a test for the `default` case - where we return the context as the cluster name

pkg/skaffold/config/util.go
```diff
@@ -199,33 +199,52 @@ func isDefaultLocal(kubeContext string) bool {
...
+		return true
 
 	default:
-		return false, ""
+		return false
 	}
 }
 
+// KindClusterName returns the internal kind name of a kubernetes cluster.
+func KindClusterName(clusterName string) string {
```
balopat, line 228:
```suggestion
func KindClusterName(context string) string {
```
wdyt?

pkg/skaffold/config/util_test.go
dgageot: The first line does that: `kind` -> `kind`. This is for example what will happen with kind < 0.6. The cluster name in the kubeconfig is by default `kind`

pkg/skaffold/config/util.go
dgageot: No. The name of the cluster is not the name of the context. That's the whole point of this fix.

integration/dev_test.go
dgageot: You are missing the fact that the cluster name is not the context name.  Also, the kind internal cluster name is not the cluster name.

Here's an example of what can happen in the test:

Before we create a new context:
 + context name: `kind-kind`
 + cluster name: `kind-kind`
 + kind name: `kind`

After we create a new context:
 + context name: `kind-modified-context`
 + cluster name: `kind-kind` (This one doesn't change, that's where the bug comes from)
 + kind name: `kind`

This issue shows the case where users rename their context but keep the cluster name as is. We have to remove the prefix from the cluster name, not the context name.

pkg/skaffold/runner/deploy.go
```diff
@@ -55,7 +56,19 @@ See https://skaffold.dev/docs/pipeline-stages/taggers/#how-tagging-works`)
...
+		if err != nil {
+			return fmt.Errorf("unable to get kubernetes config: %w", err)
+		}
+
+		currentContext, present := currentCfg.Contexts[r.runCtx.KubeContext]
+		if !present {
+			return fmt.Errorf("unable to get current kubernetes context: %w", err)
+		}
+
+		kindCluster := config.KindClusterName(currentContext.Cluster)
```
balopat, line 70: ahhhhhh got it!!!! :D

PR #4222, opened 2020-05-19
move field reported to changed

tejal29:
**Description**
I am working on hooking up pod statuses into skaffold.
With pod statuses now coming in, we need a variable to track if a pod status which is part of deployment is changed to print out deployment and pod statuses together like this.
```
- deployment/leeroy-web is ready. [1/2 deployment(s) still pending]
 - deployment/leeroy-app: waiting for rollout to finish: 1 old replicas are pending termination...
    - pod/leeroy-app-85dfb77b44-md2dd: container leeroy-app is waiting to start: leeroy-app1 can't be pulled
 - deployment/leeroy-app: waiting for rollout to finish: 1 old replicas are pending termination...
    - pod/leeroy-app-85dfb77b44-wemde: container leeroy-app in error: Back-off pulling image "leeroy-app1"
```
Repurposing the previous `reported` to indicate if a deployment status  was changed because a new pod came status was available or new pod came up.

### DIFF
--- a/pkg/skaffold/deploy/resource/deployment.go
+++ b/pkg/skaffold/deploy/resource/deployment.go
@@ -54,11 +54,14 @@ func (d *Deployment) Deadline() time.Duration {
 
 func (d *Deployment) UpdateStatus(details string, err error) {
 	updated := newStatus(details, err)
-	if !d.status.Equal(updated) {
-		d.status = updated
-		if strings.Contains(details, rollOutSuccess) || isErrAndNotRetryAble(err) {
-			d.done = true
-		}
+	if d.status.Equal(updated) {
+		d.status.changed = false
+		return
+	}
+	d.status = updated
+	d.status.changed = true
+	if strings.Contains(details, rollOutSuccess) || isErrAndNotRetryAble(err) {
+		d.done = true
 	}
 }
 
@@ -111,10 +114,9 @@ func (d *Deployment) IsStatusCheckComplete() bool {
 }
 
 func (d *Deployment) ReportSinceLastUpdated() string {
-	if d.status.reported {
+	if !d.status.changed {
 		return ""
 	}
-	d.status.reported = true
 	return fmt.Sprintf("%s: %s", d, d.status)
 }
 
--- a/pkg/skaffold/deploy/resource/deployment.go
+++ b/pkg/skaffold/deploy/resource/deployment.go
@@ -114,9 +114,10 @@ func (d *Deployment) IsStatusCheckComplete() bool {
 }
 
 func (d *Deployment) ReportSinceLastUpdated() string {
-	if !d.status.changed {
+	if d.status.reported && !d.status.changed {
 		return ""
 	}
+	d.status.reported = true
 	return fmt.Sprintf("%s: %s", d, d.status)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/skaffold/deploy/resource/deployment.go
```diff
@@ -54,11 +54,14 @@ func (d *Deployment) Deadline() time.Duration {
 
 func (d *Deployment) UpdateStatus(details string, err error) {
 	updated := newStatus(details, err)
-	if !d.status.Equal(updated) {
-		d.status = updated
-		if strings.Contains(details, rollOutSuccess) || isErrAndNotRetryAble(err) {
-			d.done = true
-		}
+	if d.status.Equal(updated) {
+		d.status.changed = false
```
briandealwis, line 58: Is it possible to have several status updates before the status has been reported in `ReportSinceLastUpdated()`?  Consider the following:
```
d.UpdateStatus("status1")
//  status.changed now true
d.UpdateStatus("status2")
// status.changed now true
d.UpdateStatus("status2")
// status.changed now false
```

If something checks `ReportSinceLastUpdated()` only now, then nothing will be output.

tejal29: Currently, we call `ReportSinceLastUpdated` every 500 ms and deployment statuses are polled every 100ms. So it is possible to have 3-4 updates in that time.

tejal29: i will add this change back in.

pkg/skaffold/deploy/resource/deployment.go
```diff
@@ -111,7 +114,7 @@ func (d *Deployment) IsStatusCheckComplete() bool {
 }
 
 func (d *Deployment) ReportSinceLastUpdated() string {
-	if d.status.reported {
+	if d.status.reported && !d.status.changed {
```
briandealwis, line 117: This doesn't seem right as `.reported` is only ever changed here and it's always set to `true`.

I think you just want to track the last-reported status and the current status, and compare them here.

pkg/skaffold/deploy/resource/deployment.go
```diff
@@ -54,11 +54,14 @@ func (d *Deployment) Deadline() time.Duration {
...
 func (d *Deployment) UpdateStatus(details string, err error) {
 	updated := newStatus(details, err)
-	if !d.status.Equal(updated) {
-		d.status = updated
-		if strings.Contains(details, rollOutSuccess) || isErrAndNotRetryAble(err) {
-			d.done = true
-		}
+	if d.status.Equal(updated) {
+		d.status.changed = false
+		return
```
briandealwis, line 59: Is this return appropriate here?  You still need to check the `isErrAndNotRetryable(err)` to possibly set `done`?

pkg/skaffold/deploy/resource/deployment.go
tejal29: @briandealwis This works correctly because, for every new status seen, we create a status with constructor `newStatus`. 
The `.reported` set to false. 
See deployment.go L57

pkg/skaffold/deploy/resource/deployment.go
tejal29: umm no. the status is already same as previous state. We don't need to perform any checks.

pkg/skaffold/deploy/resource/deployment.go
briandealwis: Ah ok 👍

Issue #2753, opened 2019-08-29
Add support for k3d


strikeout:
### Use case
Im trying to leverage `k3d` for efficient local k8s development. k3d requires an extra command to import the build images into the k3d containerd file-system (`k3d import-images $images`) so I need to use the custom build spec - but just on the local profile. The number of artifacts is quite large so it is unfeasible to patch every single one by itself. What would be the best approach here? 

### Actual behavior
Having n artifacts requires n patches inside each profile as the build specifics are set on an artifact level. Its not possible to "patch all artifacts at the same time" with json-patch (?) or apply the build spec to a level above the artifacts for inheritance.

### Expected behavior
Adding a `custom:` build spec to a profile should make all artifacts use said build spec.
Alternative: support local k3d cluster natively

```
build:
  artifacts:
    - image: app-one
      context: app/one
    - image: app-two
      context: app/two
    - image: app-three
      context: app/three

...

profiles:
  - name: local
    activation:
      - env: ENV=local
    build:
      custom: 
         # expected, but not possible
         buildCommand: ./build.sh
  - name: production
    activation:
      - env: ENV=prod
    ...

```

balopat:
Thanks for opening the issue! 
The bulk artifact type change doesn't sound like something we would support, not sure if there is a good generic way to do that. 
To add k3d as a target platform - we could be convinced but we would need more evidence that a lot of users would benefit from it.  Also, I am not sure what k3d provides beyond `kind` that works seemingly similarly (docker based lightweight k8s cluster with image loading ..) and we do support kind out of the box: https://github.com/kubernetes-sigs/kind.

rjshrjndrn:
for me, k3d is blazingly fast than kind, lightweight

tstromberg:
Leaving this one open, but AFAIK, no one is currently working on this. PR's welcome!

eox-dev:
From a WSL2 perspective, would be very interested in this given `minikube` [effectively doesn't work](https://github.com/kubernetes/minikube/issues/7879) but `k3d` does.

PR #4216, opened 2020-05-18
Add support for k3d (#2753)

hassomehide:
Fixes: #2753 

**Description**
This adds support for k3d ([k3s in docker](https://github.com/rancher/k3d)), which is very similar to kind in terms of cluster creation and image loading.

Tested with k3d v3.0.0 (currently in beta). Support could be added to older versions (<=1.7.0) as well, however clusters created with those do not have a distinguishable prefix in their name - any ideas how to solve this?

PR #4224, opened 2020-05-19
Add support for k3d

hassomehide:
Fixes: #2753 
Previous PR: #4216 (created a new one to change commit author email)

**Description**
This adds support for k3d (k3s in docker), which is very similar to kind in terms of cluster creation and image loading.

Tested with k3d v3.0.0 (currently in beta). Older versions (<=1.7.0) have a similar image loading mechanism, however clusters created with those do not have any distinguishable prefix (k3d-) in their name.


hassomehide:
Fixed some conflicts that occurred after #4212 was merged, also tried to bring those changes to the new k3d code so kind and k3d would share the same logic regarding context name vs cluster name.

hassomehide:
Sounds good, will take a look at it - in another PR then.

PR #4265, opened 2020-05-27
[buildpacks] Custom project toml

dgageot:
The path to the `project.toml` can be specified in the `skaffold.yaml`

### DIFF
--- a/pkg/skaffold/schema/latest/config.go
+++ b/pkg/skaffold/schema/latest/config.go
@@ -753,6 +753,10 @@ type BuildpackArtifact struct {
 	// TrustBuilder indicates that the builder should be trusted.
 	TrustBuilder bool `yaml:"trustBuilder,omitempty"`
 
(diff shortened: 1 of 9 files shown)


pkg/skaffold/schema/latest/config.go
```diff
@@ -750,6 +750,10 @@ type BuildpackArtifact struct {
 	// Order matters.
 	Buildpacks []string `yaml:"buildpacks,omitempty"`
 
+	// ProjectDescriptor is the path to the toml project descriptor.
+	// Defaults to `project.toml`.
```
briandealwis, line 754: I don't think the toml is necessary since we have it below.  And it is a file.  (This is similar what `pack` shows.)
```suggestion
	// ProjectDescriptor is the path to the project descriptor file.
	// Defaults to `project.toml` if it exists.
```

PR #4269, opened 2020-05-27
Collapse ImagePullBackOff and ErrImagePullBackOff together

tejal29:
Relates to: #3917 

Both [ErrImagePullBackOff](https://github.com/kubernetes/kubernetes/blob/d0183703cbe715c879cb42db375c7373b7f2b6a1/pkg/kubelet/kubelet_test.go#L910) and [ImagePullBackOff](https://github.com/kubernetes/kubernetes/blob/d24fe8a801748953a5c34fd34faa8005c6ad1770/pkg/kubelet/images/types.go#L28) could appear in the `Reason` string.

Hence combining them together.

### DIFF
--- a/pkg/diag/validator/pod.go
+++ b/pkg/diag/validator/pod.go
@@ -30,17 +30,17 @@ import (
 )
 
 const (
-	success             = "Succeeded"
-	running             = "Running"
-	actionableMessage   = `could not determine pod status. Try kubectl describe -n %s po/%s`
-	errorPrefix         = `(?P<Prefix>)(?P<DaemonLog>Error response from daemon\:)(?P<Error>.*)`
-	taintsExp           = `\{(?P<taint>.*?):.*?}`
-	crashLoopBackOff    = "CrashLoopBackOff"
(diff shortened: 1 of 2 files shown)


pkg/diag/validator/pod.go
```diff
@@ -202,14 +202,18 @@ func (p *podStatus) String() string {
 }
 
 func extractErrorMessageFromWaitingContainerStatus(c v1.ContainerStatus) (proto.StatusCode, error) {
+	r := c.State.Waiting.Reason
+	if strings.HasSuffix(r, imagePullBackOff) {
+		r = imagePullBackOff
```
briandealwis, line 207: Are you saying that multiple reasons could occur, or that we have "ImagePullBackOff" or "ErrImagePullBackOff"?  I'm leery with trying to transform error codes like this.  It doesn't seem hard to have both and change the case statement below for the additional condition.

tejal29: sounds good. i will add another constant for `ErrImagePullBackOff`

Issue #4279, opened 2020-05-29
`skaffold delete` ignoring namespace argument using Helm


MoogyG:
I am using skaffold 1.8.0.

I want a default namespace for my helm chart in development, so I use 
```
deploy:
  helm:
    releases:
    - name: my-release
      namespace: my-namespace
```

Everything goes well, the chart is uninstalled if I exit `skaffold dev` or running `skaffold run` and then `skaffold delete`.

But in my CI, I need a namespace for every branch so I use `skaffold run -n my-namespace-ci-6gh81g` (the suffix is the commit sha), and the chart is created in this namespace `my-namespace-ci-6gh81g` succesfully, overriding the default `my-namespace` namespace.

But if I use `skaffold delete -n my-namespace-ci-6gh81g`, the namespace argument is ignored and I got this error : `Error: uninstall: Release not loaded: my-release: release: not found`.
It seems to be seeking the chart in the namespace `my-namespace`.
The same error appears when using `skaffold dev -n my-namespace-ci-6gh81g` and then exiting the command.

Here is my `skaffold.yaml`

```
apiVersion: skaffold/v2beta2
kind: Config
metadata:
  name:	skaffold-test
deploy:
  helm:
    releases:
    - name: my-release
      chartPath: k8s
      namespace: my-namespace
      values:
        image: nginx
    flags:
      install:
        - "--create-namespace" 
```

briandealwis:
Odd as #3914 ensured `--namespace` was added to the `helm delete` for Skaffold 1.8.0.  And I see Skaffold 1.8.0 passing it in my testing.

Could you include the logs of running `skaffold ... -v debug`?
And could you please try with the latest release (1.10.1)?

MoogyG:
Here is the debug for the command `skaffold dev -n my-namespace-ci-test -v debug` using skaffold 1.8.0 :

```
INFO[0000] starting gRPC server on port 50051           
INFO[0000] starting gRPC HTTP server on port 50052      
INFO[0000] Skaffold &{Version:v1.8.0 ConfigVersion:skaffold/v2beta2 GitVersion: GitCommit:bd280192092e28067f0f52584c8bcb4f4dc480e4 GitTreeState:clean BuildDate:2020-04-17T23:01:25Z GoVersion:go1.14 Compiler:gc Platform:linux/amd64} 
INFO[0000] applying profile: dev                        
DEBU[0000] overlaying profile on config for field Build 
[...]

PR #4281, opened 2020-05-31
apply namespace from command first when cleaning up helm release

GwonsooLee:
Fixes: #4279 
**Related**: #3914 

**Description**
After #3914 is released `--namespace` is added when cleaning up the helm release. According to the code, when deploying helm release `opt.namespace` is checked first.
https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/deploy/helm.go#L363-L367

But when cleaning up releases, the namespace of the release is applied. (I'm not sure if this is intended..)
https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/deploy/helm.go#L229-L234

--namespace is working well if I use `build & deploy`, but it is not working with `dev` command.

### DIFF
--- a/pkg/skaffold/deploy/helm.go
+++ b/pkg/skaffold/deploy/helm.go
@@ -227,10 +227,10 @@ func (h *HelmDeployer) Cleanup(ctx context.Context, out io.Writer) error {
 		}
 
 		var namespace string
-		if r.Namespace != "" {
-			namespace = r.Namespace
-		} else if h.namespace != "" {
+		if h.namespace != "" {
 			namespace = h.namespace
+		} else if r.Namespace != "" {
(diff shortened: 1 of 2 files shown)


GwonsooLee:
The reason why `build & deploy` works is that whey deploy the helm release it calls `deployRelease`function and it checks the namespace from command before execution. And I think Cleanup() function is not called..(please correct me if I am wrong...)

https://github.com/GoogleContainerTools/skaffold/blob/1b68b3bb938e241a02da639c3d07412a6324096b/pkg/skaffold/deploy/helm.go#L363-L382

briandealwis:
I'm going to add some tests to verify this change, and that we're using the right namespace in the other cases.

PR #4309, opened 2020-06-10
`Skaffold init` supports Python projects with Buildpacks

dgageot:


dgageot:
ping @nkubala @briandealwis

integration/examples/buildpacks-python/skaffold.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: skaffold/v2beta5
+kind: Config
+build:
+  artifacts:
+  - image: skaffold-buildpacks
+    buildpacks:
+      builder: "gcr.io/buildpacks/builder"
```
briandealwis, line 7: In light of #4313, this should be with `:v1`

pkg/skaffold/build/buildpacks/init.go
```diff
@@ -71,12 +71,30 @@ func (c ArtifactConfig) Path() string {
...
+
+	// Go.
+	case file == "go.mod":
+		valid = true
+
+	// Python.
+	case file == "requirements.txt":
+		if _, err := os.Stat(filepath.Join(filepath.Dir(path), "Procfile")); err == nil {
+			valid = true
+		}
```
briandealwis, line 94: The boolean return doesn't quite capture enough.  We should be able to return output a message here along the lines of "...$path seems to be a Python project, but Buildpacks requires a Procfile".  Or maybe this framework needs the ability to register a follow-up action (`REQUIRES_PROCFILE`) so that the CLI/UI/IDE can offer the user some detail as to why it failed.

A todo for now is fine.

PR #4313, opened 2020-06-11
Update GCP Buildpacks builder image references to :v1

briandealwis:
**Description**

GCP Buildpacks has changed to use a versioning scheme via an image tag to avoid breaking consumers from breaking changes.  This PR changes all references to `gcr.io/buildpacks/builder` to add the new `:v1` image tag.  I updated the relevant `pkg/skaffold/schema/` update tests too, though they don't actually matter.

- all examples use `:v1`


briandealwis:
fyi @lukasberger

PR #4316, opened 2020-06-11
Skaffold deploy hydrated manifests

ChrisGe4:
Description
Proposal: go/skaffold-deploy-hydrated-manifests

Currently skaffold deploy renders the Kubernetes manifests first then deploys the final manifests . This PR modifies modifies `skaffold deploy` to solely deploy the hydrated Kubernetes manifests. 

In this PR, a new option --skip-render will be added to the deploy command. `skaffold deploy --skip-render` only deploy the kubernetes manifests without first rendering them. Deploy can also take manifests in GSC.


ChrisGe4:
@nkubala Please review when you get a chance.

ChrisGe4:
I run the failed test locally and they both succeeded. 

for the failed integration test, I run deploy --images index.docker.io/library/busybox:1 --default-repo=

I got the following output from `kubectl describe deployment kustomize-test`

```
Name:                   kustomize-test
Namespace:              default
CreationTimestamp:      Mon, 15 Jun 2020 01:46:56 -0400
Labels:                 app=kustomize-test
                        app.kubernetes.io/managed-by=skaffold-unknown
                        skaffold.dev/builder=local
                        skaffold.dev/cleanup=true
                        skaffold.dev/deployer=kustomize
                        skaffold.dev/docker-api-version=1.40
                        skaffold.dev/run-id=4493d96b-2fd5-4e5b-aea4-b1fb7e23cb22
                        skaffold.dev/tag-policy=git-commit
Annotations:            deployment.kubernetes.io/revision: 1
                        kubectl.kubernetes.io/last-applied-configuration:
                          {"apiVersion":"apps/v1","kind":"Deployment","metadata":{"annotations":{},"labels":{"app":"kustomize-test","app.kubernetes.io/managed-by":"...
Selector:               app=kustomize-test
Replicas:               1 desired | 1 updated | 1 total | 1 available | 0 unavailable
StrategyType:           RollingUpdate
MinReadySeconds:        0
RollingUpdateStrategy:  25% max unavailable, 25% max surge
Pod Template:
  Labels:  app=kustomize-test
           app.kubernetes.io/managed-by=skaffold-unknown
           skaffold.dev/builder=local
           skaffold.dev/cleanup=true
           skaffold.dev/deployer=kustomize
           skaffold.dev/docker-api-version=1.40
           skaffold.dev/run-id=4493d96b-2fd5-4e5b-aea4-b1fb7e23cb22
           skaffold.dev/tag-policy=git-commit
  Containers:
   kustomize-test:
    Image:      index.docker.io/library/busybox:1
    Port:       <none>
    Host Port:  <none>
    Command:
      sleep
      3600
    Environment:  <none>
    Mounts:       <none>
  Volumes:        <none>
Conditions:
  Type           Status  Reason
  ----           ------  ------
  Available      True    MinimumReplicasAvailable
  Progressing    True    NewReplicaSetAvailable
OldReplicaSets:  <none>
NewReplicaSet:   kustomize-test-6c88b65c94 (1/1 replicas created)
Events:
  Type    Reason             Age   From                   Message
  ----    ------             ----  ----                   -------
  Normal  ScalingReplicaSet  70s   deployment-controller  Scaled up replica set kustomize-test-6c88b65c94 to 1
```
Not sure why got 

```
   TestDeploy: deploy_test.go:81: string differ (-got, +want):   string(
        - 	"index.docker.io/library/busybox",
        + 	"index.docker.io/library/busybox:1",
          )
```
-----------------------------------------------------------

As to the failed unit test, it takes 72s to run it locally, but the timeout is set to 60s on the test server.  Could be the cause of the failure?

docs/content/en/docs/references/cli/_index.md
```diff
@@ -445,6 +445,9 @@ Examples:
   # Build the artifacts and then deploy them
   skaffold build -q | skaffold deploy --build-artifacts -
 
+  # Deploy without first rendering the manifests
+  skaffold skaffold deploy --skip-render
```
aboyett, line 449:
```suggestion
  skaffold deploy --skip-render
```

pkg/skaffold/deploy/kubectl_test.go
```diff
@@ -639,3 +640,55 @@ spec:
...
+	}{
+		{
+			description: "manifest from GCS",
+			cfg: &latest.KubectlDeploy{
+				Manifests: []string{"gs://dev/deployment.yaml"},
+			},
+			commands: testutil.
+				CmdRunOut(fmt.Sprintf("gsutil cp -r %s %s", "gs://dev/deployment.yaml", manifestTmpDir), "log").
+				AndRunOut("kubectl version --client -ojson", kubectlVersion112).
+				AndRunOut("kubectl --context kubecontext --namespace testNamespace create --dry-run -oyaml -f "+manifestTmpDir+"/deployment.yaml", deploymentWebYAML).
```
tejal29, line 660: please replace / with filepath.Join. I think your windows unit test is failing due to it.

cmd/skaffold/app/cmd/deploy.go
```diff
@@ -43,31 +43,35 @@ func NewCmdDeploy() *cobra.Command {
...
 	return withRunner(ctx, func(r runner.Runner, config *latest.SkaffoldConfig) error {
-		deployed, err := getArtifactsToDeploy(out, buildOutputFile.BuildArtifacts(), preBuiltImages.Artifacts(), config.Build.Artifacts)
-		if err != nil {
-			return err
-		}
-
-		for i := range deployed {
-			tag, err := r.ApplyDefaultRepo(deployed[i].Tag)
+		var deployed []build.Artifact
+		if !opts.SkipRender {
```
tejal29, line 60: you can simplify this by return first when `--skip-render=true`
```suggestion
		if opts.SkipRender {
		   return r.DeployAndLog(ctx, out, deployed)
		}
```

tejal29: as far as i understand, with `--skip-render=true` deploy should happen. However with your change, seems like deploy is skipped all together.
Am i reading this correct?

ChrisGe4: With your suggestion, `return r.DeployAndLog(ctx, out, deployed)` will show up twice. Should I duplicate that line of code?

deploy is not skipped as `return r.DeployAndLog(ctx, out, deployed)` is the code to do the deployment.

ChrisGe4:
I rerun those failed checks locally without seeing any error.

cmd/skaffold/app/cmd/deploy.go
tejal29: yes. however, `deployed` is empty in `opts.SkipRender=false` flow. So no artifacts will be deployed?
I will try to run this locally today.

tejal29: ah got. The `--skip-render` does not replace the built images and deploys the manifests as is.

tejal29: >With your suggestion, return r.DeployAndLog(ctx, out, deployed) will show up twice. Should I duplicate that line of code?

Yes, i think it will easier to read if you add the `opts.SkipRender` check as first line of the function and return and pass an empty build.Artifact array.
```
return r.DeployAndLog(ctx, out, []build.Artifact{})
``

File: pkg/skaffold/deploy/kubectl.go [hunk 4]
```diff
@@ -109,10 +118,14 @@ func (k *KubectlDeployer) Deploy(ctx context.Context, out io.Writer, builds []bu
 }
 
 func (k *KubectlDeployer) manifestFiles(manifests []string) ([]string, error) {
-	var nonURLManifests []string
+	var nonURLManifests, gcsManifests []string
 	for _, manifest := range manifests {
 		if !util.IsURL(manifest) {
```
tejal29, line 123: You can break this in a switch case and do nothing when manifest is an url
e.g.  https://play.golang.org/p/QaInR52gYCG
```suggestion
                switch {
                case util.IsURL(manifest):
                   // do nothing.
                case strings.HasPrefix(manifest, "gs://"):
                        gcsManifests = append(gcsManifests, manifest)
                default:
                         nonURLManifests = append(nonURLManifests, manifest)
              }                  
```

pkg/skaffold/deploy/kubectl.go
```diff
@@ -143,6 +169,8 @@ func (k *KubectlDeployer) readManifests(ctx context.Context, offline bool) (depl
 	if err != nil {
 		return nil, fmt.Errorf("listing manifests: %w", err)
 	}
+	// Clean the temporary directory that holds the manifests downloaded from GCS
+	defer os.RemoveAll(manifestTmpDir)
```
tejal29, line 173: move defer up to L143. go coding standards mentions to call defer immediately after object creation.

e.g. the way is written right now, if `util.ExpandPathsGlob` fails, this temporary dir wont be cleaned up.

pkg/skaffold/deploy/util.go
```diff
@@ -133,6 +136,23 @@ func dumpToFile(renderedManifests string, filepath string) error {
 	return err
 }
 
+// Returns relative path pointing to the GCS temp dir
+func downloadManifestsFromGCS(manifests []string) (string, error) {
+	if err := os.MkdirAll(manifestTmpDir, os.ModePerm); err != nil {
```
tejal29, line 141: Can you please use create a `manifestTmpDir` in this function instead of  a variable using 
```
manifestTmpDir := ioutil.TempDir("", manifestsStagingFolder)
```

pkg/skaffold/deploy/util.go
```diff
@@ -133,6 +136,23 @@ func dumpToFile(renderedManifests string, filepath string) error {
...
+		return "", fmt.Errorf("failed to create the tmp directory: %w", err)
+	}
+	for _, manifest := range manifests {
+		if manifest == "" || !strings.HasPrefix(manifest, gcsPrefix) {
+			return "", fmt.Errorf("%v is not a valid GCS path", manifest)
```
tejal29, line 146: please cleanup the manifestTmpDir here using `defer os.RemoveAll(manifestTmpDir)`

pkg/skaffold/deploy/util.go
```diff
@@ -133,6 +136,23 @@ func dumpToFile(renderedManifests string, filepath string) error {
...
+		}
+		gcs := util.Gsutil{}
+		if err := gcs.Copy(context.Background(), manifest, manifestTmpDir, true); err != nil {
+			return "", fmt.Errorf("failed to download manifests fom GCS: %w", err)
```
tejal29, line 150: please cleanup the manifestTmpDir here using defer os.RemoveAll(manifestTmpDir)

pkg/skaffold/deploy/kubectl.go
tejal29: In your code, `downloadManifestsFromGCS` creates the temporary dir. 
With this approach, you have to clean up the temporary dir multiple places. 
I would suggest,  you change the signature of `downloadManifestsFromGCS` to 
`func downloadManifestsFromGCS(dest string, gcsManifests []string) error`
In this case, your flow would be
```
tempDir, err := ioutil.TempDir("", manifestsStagingFolder)
 if err != nil {			
  return fmt.Errorf("failed to create tmp directory: %w", err)	
}
defer os.RemoveAll(tempDir)
if err :=downloadManifestsFromGCS(tempDir, gcsManifests)  != nil {
	return nil, fmt.Errorf("downloading from GCS: %w", err)
}
l, err := util.ExpandPathsGlob(tmpDir, []string{"*"})
if err != nil {
  return nil, fmt.Errorf("expanding kubectl manifest paths: %w", err)
}
list = append(list, l...)
```

ChrisGe4: I will use this comment to reply all the comments regarding the temporary directory. `func Dependencies` only list the manifests but the files were read and processed in `func readManifests`, which means cleanup has to happen in `readManifests`, which also means I can't use `ioutil.TempDir` to create a randomly named temp dir(since the deploy is synchronized, this should be safe ). 
You are right about
> e.g. the way is written right now, if util.ExpandPathsGlob fails, this temporary dir wont 
> be cleaned up.
I will move the defer to L170.

pkg/skaffold/deploy/util.go
ChrisGe4: See the comment above.

pkg/skaffold/deploy/util.go
ChrisGe4: See the comment above.

pkg/skaffold/deploy/util.go
ChrisGe4: See the comment above.

pkg/skaffold/deploy/kubectl.go
tejal29: sounds good.

PR #4318, opened 2020-06-12
`skaffold init` supports Java/Buildpacks projects

dgageot:


pkg/skaffold/build/buildpacks/init.go
```diff
@@ -73,24 +73,27 @@ func (c ArtifactConfig) Path() string {
...
+	case "package.json":
 		valid = true
 
 	// Go.
-	case file == "go.mod":
+	case "go.mod":
+		valid = true
+
+	// Java.
+	case "pom.xml", "build.gradle", "build.gradle.kts":
```
briandealwis, line 90: Could also add [`.mvn/extensions.xml`](http://maven.apache.org/configure.html#mvn-folder) (and jvm.config and maven.config?)

dgageot: Not sure those files are detected by our buildpacks. Do you know?

briandealwis: They weren't, but … GoogleCloudPlatform/buildpacks#50 should fix that.  I realize that only `extensions.xml` is applicable here as `jvm.config` and `maven.config` don't actually create the project object.

(We can add these separately — I don't think that PR should hold this up.)

PR #4319, opened 2020-06-13
Support mktemp on older Macs

breun:
**Description**

I maintain the `skaffold` port in MacPorts and was notified of the fact that [building Skaffold fails on Mac OS X 10.10 and older](https://trac.macports.org/ticket/60634) with the following error:

```
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix 
make: *** [cmd/skaffold/app/cmd/statik/statik.go] Error 1
```

`mktemp` requires either a template with `X`s (`template`) or a prefix (`-t prefix). This usage is still in the latest `man mktemp`, although on current Macs `mktemp` defaults to `-t tmp` if no arguments are passed or if only the `-d` flag is passed.

This pull request adds `-t tmp` to `mktemp` calls in Skaffold's scripts to add support for Mac OS X and older. This should not affect later versions of Mac OS X (or macOS).


breun:
Instead of `tmp` also a prefix like `skaffold` could be used to more easily identify a temporary directory as being used by Skaffold.

breun:
Hm, apparently Travis builds Skaffold on Ubuntu 16.04, which has different behavior: it expects the prefix to contain `X`s, while `mktemp` on macOS doesn't. :(

Issue #4322, opened 2020-06-15
Render does not fully overwrite output file if it already exists


alewis001:
### Expected behavior
The contents of the file `<path>/<file>.yaml` reflects the exact output of the command `skaffold render --output <path>/<file>.yaml`. If the file already exists, the file's content is fully overwritten.

### Actual behavior
The `render` command is only overwriting the bytes inside the file, not the entire file. 

I.e. If the current file is bigger (X bytes) and the output from `render` results in less bytes (Y bytes); then Y bytes in the file are overwritten but the remaining X-Y bytes in the file are left untouched. 

This issue affects Kubectl, Kustomize and Helm deployers. 

I think the issue is with the use of `os.OpenFile` instead of `os.Create` but I could be way off the mark as I'm very new to Go.

Helm and Kubectl both use `outputRenderedManifests`, which in turn uses `dumpToFile`, which uses `os.OpenFile`.

Kustomize uses `os.OpenFile` but inside `kustomize.go` on line 215 rather than `util.go`.

I tried changing the affected lines above to use `os.Create` and that appears to fix the issue; however, whether that's the appropriate fix I wouldn't like to say. I'm happy to make that change and submit a PR if that's the way to go.

### Information

- Skaffold version: 1.11.0
- Operating system: macos
- Deployer: Kubectl, Kustomize, Helm

### Steps to reproduce the behavior

1. Get the `skaffold/examples/helm-deployment` example
1. Modify the skaffold.yaml to make it a local build only
1. Run: `skaffold render --output render.txt`
1. Edit render.txt, add some additional text/YAML to the end of the file.
1. Run: `skaffold render --output render.txt`
1. Check contents of render.txt; the additional text added above is not removed.

E.g. Edited render.txt I.e."before"
```
---
# Source: skaffold-helm/templates/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: skaffold-helm
  labels:
    app: skaffold-helm
spec:
  replicas: 1
  selector:
    matchLabels:
      app: skaffold-helm
  replicas: 2
  template:
    metadata:
      labels:
        app: skaffold-helm
    spec:
      containers:
      - name: skaffold-helm
        image: skaffold-helm:2ffd66737879da07ff3de2326f8e8c9d82f8748b5333ee4ae37033952ea269b0
add:
  some:
    additional: yaml
```
After another render to the same file:
```
---
# Source: skaffold-helm/templates/deployment.yaml
[...]

PR #4323, opened 2020-06-15
#4322: Fix render not fully overwriting output files.

alewis001:
If the output file already existed, it was not being truncated to the
new output. If the existing content was larger than the new output, the
remainder of the existing output was left in the file corrupting the
contents.

Fixes: #4322  

**Description**

The changes are:

* Change the `os.OpenFile` call to `os.Create` in `util.go`
* Change `kustomise.go` to be consistent with the other deployers and use the `outputRenderedManifests` function of `util.go`

### DIFF
--- a/pkg/skaffold/deploy/kustomize.go
+++ b/pkg/skaffold/deploy/kustomize.go
@@ -210,19 +210,7 @@ func (k *KustomizeDeployer) Render(ctx context.Context, out io.Writer, builds []
 		return err
 	}
 
-	manifestOut := out
-	if filepath != "" {
-		f, err := os.OpenFile(filepath, os.O_RDWR|os.O_CREATE, 0666)
-		if err != nil {
-			return fmt.Errorf("opening file for writing manifests: %w", err)
-		}
-		defer f.Close()
-		f.WriteString(manifests.String() + "\n")
-		return nil
-	}
-
-	fmt.Fprintln(manifestOut, manifests.String())
-	return nil
+	return outputRenderedManifests(manifests.String(), filepath, out)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/skaffold/deploy/kustomize.go
```diff
@@ -210,19 +210,7 @@ func (k *KustomizeDeployer) Render(ctx context.Context, out io.Writer, builds []
...
-		f, err := os.OpenFile(filepath, os.O_RDWR|os.O_CREATE, 0666)
-		if err != nil {
-			return fmt.Errorf("opening file for writing manifests: %w", err)
-		}
-		defer f.Close()
-		f.WriteString(manifests.String() + "\n")
-		return nil
-	}
-
-	fmt.Fprintln(manifestOut, manifests.String())
```
briandealwis, line 224: This change isn't equivalent as when `filepath == ""`, this code would have printed the manifests to `out`.

alewis001: Thanks @briandealwis. In `outputRenderedManifests(renderedManifests string, output string, manifestOut io.Writer)` of `util.go`, if `out == ""` it calls fmt.Fprintln (see Line 104 of `util.go`). So, in this case if `filepath` is "" I think that should do what you're asking doesn't it?

briandealwis: My apologies — I could have sworn the `outputRenderedManifests()` was very different.

alewis001: No worries, just glad it does the right thing :)

PR #4325, opened 2020-06-15
Automate release steps post changelog PR merge.

gsquared94:
Fixes: #4324  

**Description**

Current release has a lot of manual actions that can be broadly categorized into two steps:
- Create Release Changelog PR
- Create and Publish Github Release

This PR automates the second step. When the Changelog PR is merged it automatically creates the Github Release(as a draft) with all text and attachments. We will only need to hit publish button once we're ok with the draft. In time this manual verification can be skipped too.

This requires that the Changelog PR is labelled with `draft-release` *before* being merged. Also the PR title should be `Release vX.x.x` or `vX.x.x release` since it parses the PR title for the tag semVer. 

In a subsequent PR I'll automate the first step of creating the Release Changelog PR.

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -24,6 +24,13 @@ jobs:
         echo ::set-env name=TAG_NAME::"v${t}"
         echo ::set-env name=RELEASE_NAME::"v${t} Release"
 
+    - name: Create and push Tag
+      run: |
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub Action"
+        git tag -a ${{ env.TAG_NAME }} -m "${{ env.RELEASE_NAME }}"
+        git push origin ${{ env.TAG_NAME }}
+
     - name: Install gsutil
       run: |
         curl -Lo $HOME/gsutil.tar.gz https://storage.googleapis.com/pub/gsutil.tar.gz
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -1,0 +1,57 @@
+name: create draft release
+
+on:
+  pull_request:
+    types: [closed]
+jobs:
+  build:
+    if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'draft-release')
+    name: Create Tag and Draft Release
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Calculate Tag and release names
+      run: |
+        t=$(echo ${{ github.event.pull_request.title }} | sed -ne 's/.*\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/p')
+        if [ -z "$t" ]; then
+        echo Malformed title for PR; failed to extract semVer tag
+        return 1
+        fi
+        echo ::set-env name=TAG_NAME::"v${t}"
+        echo ::set-env name=RELEASE_NAME::"v${t} Release"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/draft-release.yml
```diff
@@ -0,0 +1,57 @@
...
+        echo ::set-env name=TAG_NAME::"v${t}"
+        echo ::set-env name=RELEASE_NAME::"v${t} Release"
+
+    - name: Install gsutil
+      run: |
+        curl -Lo $HOME/gsutil.tar.gz https://storage.googleapis.com/pub/gsutil.tar.gz
+        tar xfz $HOME/gsutil.tar.gz -C $HOME
+        echo ::add-path::$HOME/gsutil
+
+    - name: Download release artifacts
```
tejal29, line 33: What happens when this step fails tp download anything or one of the binaries is missing?

.github/workflows/draft-release.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    - name: Create Release
+      shell: bash
+      run: |
+        curl -fsSL https://github.com/github/hub/raw/master/script/get | bash -s 2.14.1
+        body=$(git log -p --follow -1 CHANGELOG.md | grep '^\+' | cut -c 2- | tail -n +2)
+        assets=()
+        for asset in $HOME/${{ env.TAG_NAME }}/*; do
+        assets+=("-a" "$asset")
+        done
```
tejal29, line 54: shd we assert here the number of binaries?

.github/workflows/draft-release.yml
gsquared94: The pipeline overall should fail as a result

.github/workflows/draft-release.yml
gsquared94: The number of artifacts is unfortunately already hardcoded on line 38: `$found -lt 10`

.github/workflows/draft-release.yml
gsquared94: image

.github/workflows/draft-release.yml
gsquared94: The `hub` command does log the number of artifacts that it uploads
image

gsquared94:
> To test this, shd i create a PR with the label mentioned?

You should be able to test this by resetting your fork master with this branch and raising a PR that has changes to the `CHANGELOG.md` file; title like `Release v1.11.0` and the label `draft-release`. It'll only work if you try on an already released version since running it in your fork won't trigger the cloud build pipeline that creates the artifacts. You should delete the tag and release from the Github release page to verify that it recreates both. Or if you want to you can make it point to your own bucket and release whatever version you want.

tejal29:
Thanks, i was able to test of the workflow got triggered 
image

nkubala:
> This requires that the Changelog PR is labelled with draft-release before being merged

what happens if we forget to do this? inevitably I think it will happen at some point so we should know what our recourse is.

tejal29:
This is great!
image

tejal29:
and it created a draft release here 
image

gsquared94:
> 
> 
> > This requires that the Changelog PR is labelled with draft-release before being merged
> 
> what happens if we forget to do this? inevitably I think it will happen at some point so we should know what our recourse is.

@nkubala ya I didn't want it to be automatically figured like just detecting changes to the Changelog file or a specific PR title as those could result in unwanted deployments. So it's explicit using a label. Unfortunately once the PR has been merged without the label, adding it to the closed PR won't help and Github Actions doesn't support manually triggering the workflow. The resolution would be another PR with an empty commit (sigh).

PR #4330, opened 2020-06-16
change 0 code for unknown error

tejal29:
When going through success rates for Deploy and Debug, VS code success rate was too low. 
As per the proto3 docs

> In proto3, all fields are "optional" (in that it is not an error if the sender fails to set them). But, 
fields are no longer "nullable", in that there's no way to tell the difference between a field being 
explicitly set to its default value vs. not having been set at all.

It could be because, the default value for "ErrorCode" is 0 which means "Unknown_Error".
However, the "ErrorCode" could be 0 because it was not set.

Distinguishing that in this PR.


proto/skaffold.proto
```diff
@@ -359,8 +359,8 @@ enum ClusterType {
 // These error codes are prepended by Phase Name e.g.
 // BUILD, DEPLOY, STATUSCHECK, DEVINIT
 enum StatusCode {
-    // Could not determine error and phase
-    UNKNOWN_ERROR = 0;
+    // Default StatusCode
```
briandealwis, line 362: Can we make this doc string clearer:
```suggestion
    // A default status code for events that do not have an associated phase.
    // Typically seen with the DevEndEvent event on success.
```

proto/skaffold.proto
```diff
@@ -359,8 +359,8 @@ enum ClusterType {
...
+    NO_ERROR = 0;
```
briandealwis, line 363: And we discussed separately to rename this to `OK`.

PR #4334, opened 2020-06-17
chore: use setValues not values in helm docs example

steve148:
**Description**

Noticed when going through the docs recently that the example listed on the helm deployers page referenced the `values` field. If I understand if correctly, the field has been renamed to `setValues` now. While the example code folders in the repo seem to be aligned with the newer field name, the example given in the helm docs page was not, so I created this incredibly difficult, super arduous fix.

I didn't create an issue since I thought it was a fairly small change / edit, but if you'd like I can create an issue first to make sure this is a correct change.

Before

Screenshot from 2020-06-16 23-34-24

After

Screenshot from 2020-06-16 23-34-46

Don't think this section applies here.

### DIFF
--- a/docs/content/en/docs/pipeline-stages/deployers/helm.md
+++ b/docs/content/en/docs/pipeline-stages/deployers/helm.md
@@ -54,7 +54,7 @@ deploy:
   helm:
     releases:
     - name: my-release
-      setValues:
+      artifactOverrides:
         image: gcr.io/my-project/my-image # no tag present!
         # Skaffold continuously tags your image, so no need to put one here.
 ```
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/pipeline-stages/deployers/helm.md
```diff
@@ -54,7 +54,7 @@ deploy:
   helm:
     releases:
     - name: my-release
-      values:
+      setValues:
```
tejal29, line 57: `values` is now renamed to `artifactOverrides`.
```suggestion
      artifactOverrides:
```

PR #4337, opened 2020-06-17
Fix propagation of buildpacks working directory

briandealwis:
**Description**

I backed out a critical commit to change the working directory in #4326 that was masked by a different change to the CNB tests.

More deeply: `debug` works by examining the container image run configuration (specifically the entrypoint + command, environment).  CNB images use a launcher process that retrieves the actual command from configuration files in the image; this information is also available as annotations on the image.  The CNB debug support wraps the normal debug transformation process, providing an image run configuration information that was rewritten from the CNB information.  It also needs to unwrap some of the changes.  One aspect that was missed here was to unwrap the reported working directory.

### DIFF
--- a/pkg/skaffold/debug/cnb.go
+++ b/pkg/skaffold/debug/cnb.go
@@ -80,6 +80,10 @@ func updateForCNBImage(container *v1.Container, ic imageConfiguration, transform
 	}
 
 	c, img, err := transformer(container, ic)
+	// must explicitly modify the working dir as the imageConfig is lost after we return
+	if c.WorkingDir == "" {
+		c.WorkingDir = ic.workingDir
+	}
 
 	// Only rewrite the container.Args if set: some transforms only alter env vars,
 	// and the image's arguments are not changed.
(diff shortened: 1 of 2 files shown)


briandealwis:
cc: @etanshaul @redborian

pkg/skaffold/debug/cnb.go
```diff
@@ -80,6 +80,10 @@ func updateForCNBImage(container *v1.Container, ic imageConfiguration, transform
 	}
 
 	c, img, err := transformer(container, ic)
+	// must explicitly modify the working dir as the imageConfig is lost after we return
+	if c.WorkingDir == "" {
```
gsquared94, line 84: Seems a bit weird that you'd have to modify the object immediately after creating it. Is this only happening when `err != nil`? Should the default value be rather set in `transform.go` line 344:
```
return ContainerDebugConfiguration{ WorkingDir: config.workingDir }, "", fmt.Errorf("unable to determine runtime for %q", container.Name)
```

gsquared94: nevermind, it got merged while I was looking at it 😄

gsquared94: Also I totally skipped the description which explains it 😅

Issue #4305, opened 2020-06-09
Global Configuration


dspitzer:
After reading the "Global Configuration" documentation I'm confused:

- What format should that global configuration file have? Yaml?
- How do I specify for which context it is valid?
- Are only the options listed in the table valid for this config file?
- Why is the example a cli statement?

tejal29:
@dspitzer Thank you for your feedback. 
Would it help if we add a sample config in the docs https://skaffold.dev/docs/design/global-config/

To answer your question:
1. Global config is a yaml file typically placed at ~/.skaffold/config
2. Ideally, you should not write the config but use the command  `skaffold config set` to set config value.
`skaffold config set --global local-cluster true`

tejal29:
![image](https://user-images.githubusercontent.com/1534539/84951427-457af580-b0a5-11ea-86ff-239e518a3a45.png)

tejal29:
@dspitzer To answer your questions, 
* What format should that global configuration file have? Yaml? 
yes. This is a yaml file

* How do I specify for which context it is valid?
context here refers to your `kubect context` which is usually created using `kubectl config` command. See this for more information https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/


* Are only the options listed in the table valid for this config file?
Yes these are the only valid configs. 

* Why is the example a cli statement?
The example is a CLI statement since the way to interact with config is way CLI

PR #4341, opened 2020-06-17
[Docs] add an example for global config

tejal29:
Fixes: #4305 
A user reported from reading global config docs, it wasnt clear to create or add global config file. 
Add a sample global config and specify this gets created when user runs `skaffold config set`


tejal29:
Docs changes https://github.com/GoogleContainerTools/skaffold/issues/4305#issuecomment-645629943

PR #4345, opened 2020-06-18
[buildpacks] `debug` detect direct processes with `/bin/sh -c ...`

briandealwis:
**Description**

GCP Buildpacks turn Procfile processes into direct processes that invoke `/bin/bash` with arguments `-c procscript`.  Although this seems odd, it's really a special case of the script-type process launch.

### DIFF
--- a/pkg/skaffold/debug/cnb.go
+++ b/pkg/skaffold/debug/cnb.go
@@ -41,12 +41,13 @@ func init() {
 //
 //   1. _predefined processes_ are named sets of command+arguments.  There are two types:
 //      - direct: these are passed uninterpreted to os.exec; the command is resolved in PATH
+//        Note that these may actually be configured to execute `/bin/sh -c ...`
 //      - script: the command is treated as a shell script and passed to `sh -c`, the remaining
 //        arguments are added to the shell, and so available as positional arguments.
 //        For example: `sh -c 'echo $0 $1 $2 $3' arg0 arg1 arg2 arg3` => `arg0 arg1 arg2 arg3`.
 //        (https://github.com/buildpacks/lifecycle/issues/218#issuecomment-567091462).
 //   2. _direct execs_ where the container's arg[0] == `--` are treated like direct processes
-//   3. _shell scripts_ where the container's arg[0] is the script and treated are like indirect processes.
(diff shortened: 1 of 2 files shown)


pkg/skaffold/debug/cnb.go
```diff
@@ -117,11 +118,18 @@ func adjustCommandLine(m cnb.BuildMetadata, ic imageConfiguration) (imageConfigu
...
 		if p.Type == processType {
+			// Direct: p.Command is the command and p.Args are the arguments
 			if p.Direct {
-				// p.Command is the command and p.Args are the arguments
-				ic.arguments = append([]string{p.Command}, p.Args...)
-				return ic, func(transformed []string) []string {
-					return append([]string{"--"}, transformed...)
+				// Detect and unwrap `/bin/sh -c ...`-style command lines; GCP Buildpacks turn Procfiles into `/bin/bash -c ...`
+				if (p.Command == "/bin/sh" || p.Command == "/bin/bash") && len(p.Args) >= 2 && p.Args[0] == "-c" {
+					p.Command = p.Args[1]
```
tejal29, line 125: is the command is `Command: "/bin/sh", Args: []string{"-c", "command arg1 arg2"}` as in your test case, then 
`p.Command = strings.Split(p.Args[2], " ")[0]` and
`p.Args = strings.Split(p.Args[2], " ")[1:]`

or is it the case at this point, p.Args is already a space separated array of all args?

briandealwis: So `p.Args[1]` is the script provided to `sh -c`.  p.Args[2:] are then mapped to arguments for that shell script.  So it's very rarely done:
```console
$ sh -c 'echo "0 - $0"; i=0; for arg in $*; do i=$(($i + 1)); echo "$i - $arg"; done' hi there how are you
0 - hi
1 - there
2 - how
3 - are
4 - you
```

Note that `strings.Split(..., " ")` doesn't take into account quoted strings, etc.

briandealwis: (It's entirely possible that they provide `/bin/sh -x -c '...'` but I'm fine with giving up and saying "🙅‍♂️ _no debug for you_".)

tejal29: lol.

PR #4352, opened 2020-06-19
Extend `skaffold debug` integration tests to buildpacks

briandealwis:
**Description**

Extends integration tests for `debug` to test building with GCP Buildpacks using both Procfiles and the new project.toml.  We reuse the existing test projects rather than define new projects.  Like the existing tests, it continues to check that the language debugging runtimes are enabled and responding.

These tests verify that pods and deployments are annotated with the debug configuration metadata and contains the working-directory and runtime.  These annotations are used to drive the `DebuggingContainerEvent` event.

Other changes:
- Ensures all images have a working-directory so that the field is set in the debug configuration metadata.
- Renames the `jib` tests to `java` since the buildpacks mode doesn't use Jib.
- Causes the `check-jdwp.sh` script to be placed in `/workspace/scripts` to be in the same place with both Jib and Buildpacks images.


integration/testdata/debug/README.md
```diff
@@ -0,0 +1,9 @@
+# Integration Tests for `skaffold debug`
+
+These are a set of test projects for `skaffold debug`.  There are two
+configurations:
+
+  - `skaffold.yaml` configures docker- and jib-based builders
+  - `skaffold-bp.yaml` configures buildpacks-based builders
```
dgageot, line 7: Could you use two profiles?

integration/debug_test.go
```diff
@@ -29,42 +31,70 @@ func TestDebug(t *testing.T) {
 
 	tests := []struct {
 		description string
-		dir         string
+		config      string
 		args        []string
 		deployments []string
 		pods        []string
+
```
dgageot, line 38: linter complains on the new line

Issue #4175, opened 2020-05-13
volume structures (mountPath, persistentVolumeClaim) must be lower-case to avoid unmarshal error


angry-cellophane:
Follow-up to the issue reported in the comments of this question - https://github.com/GoogleContainerTools/skaffold/issues/3883

### Expected behavior
Skaffold parses the config and builds the image.

### Actual behavior
Fails to parse the config

```
parsing skaffold config: unable to parse config: yaml: unmarshal errors:
  line 7: field mountPath not found in type v1.VolumeMount
  line 12: field persistentVolumeClaim not found in type v1.Volume
```

### Information

- Skaffold version: v1.9.0-25-gd7d0e8713-dirty
- Operating system: Ubuntu 18.04.4 LTS
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta3
kind: Config
build:
  artifacts:
  - image: gcr.io/k8s-skaffold/example
    kaniko:
      volumeMounts:
      - name: pvc1
        mountPath: /path
  cluster:
    volumes:
    - name: pvc1
      persistentVolumeClaim:
        claimName: kaniko-workspace
```

### Steps to reproduce the behavior

1. `skaffold build`

I'm not a big expert in golang and learned while debugging this issue so any help would be appreciated. So far it looks like [this issue](https://github.com/go-yaml/yaml/issues/424) with go-yaml  library.

Volume-ish structs come from the k8s api and doesn't have yaml tags.
for exampe,
```go
type VolumeMount struct {
	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
	ReadOnly bool `json:"readOnly,omitempty" protobuf:"varint,2,opt,name=readOnly"`
	MountPath string `json:"mountPath" protobuf:"bytes,3,opt,name=mountPath"`
	SubPath string `json:"subPath,omitempty" protobuf:"bytes,4,opt,name=subPath"`
	MountPropagation *MountPropagationMode `json:"mountPropagation,omitempty" protobuf:"bytes,5,opt,name=mountPropagation,casttype=MountPropagationMode"`
	SubPathExpr string `json:"subPathExpr,omitempty" protobuf:"bytes,6,opt,name=subPathExpr"`
}
```

whereas all skaffold structs have the yaml tag
```go
type KanikoArtifact struct {
	AdditionalFlags []string `yaml:"flags,omitempty"`
	DockerfilePath string `yaml:"dockerfile,omitempty"`
[...]

PR #4362, opened 2020-06-19
add docs on how to define kaniko volume mount and update secret instructions

tejal29:
Fixes: #4175

Add docs on 
1. How to configure volume mounts for kaniko pod
2. Better secret docs.

### DIFF
--- a/docs/content/en/docs/pipeline-stages/builders/docker.md
+++ b/docs/content/en/docs/pipeline-stages/builders/docker.md
@@ -95,7 +95,7 @@ build:
 ```
 Skaffold will create a secret from the service account key file. Skaffold will delete the secret from your cluster at the end of the build.
 
-(**Note**: Please do not check this service account key file in your git repository)
+(**Note**: Do not check this service account key file in your git repository)
 
 
 Similarly, when pushing to a docker registry:
@@ -141,8 +141,7 @@ cluster:
     claimname: kaniko-cache-claim
 ```
 
-**Note:** Due to [go-yaml/yaml#564](https://github.com/go-yaml/yaml/pull/564), all keys under `kaniko.VolumeMounts` and
-`cluster.Volumes` section should be in lower case. 
+**Note:** All keys under `kaniko.VolumeMounts` and `cluster.Volumes` section must be in lower case. For details, please see [skaffold#4175](https://github.com/GoogleContainerTools/skaffold/issues/4175).
 
 ## Dockerfile remotely with Google Cloud Build
 
--- a/docs/content/en/docs/pipeline-stages/builders/docker.md
+++ b/docs/content/en/docs/pipeline-stages/builders/docker.md
@@ -67,15 +67,37 @@ These credentials are configured in the `cluster` section with the following opt
 
 {{< schema root="ClusterDetails" >}}
 
-To set up the credentials for Kaniko refer to the [kaniko docs](https://github.com/GoogleContainerTools/kaniko#kubernetes-secret) (**Note**: Rename the downloaded JSON key to *kaniko-secret* without appending *.json*).
-Alternatively, the path to a credentials file can be set with the `pullSecret` option:
+### Configure Kaniko Credentials 
+
+To set up the credentials for Kaniko refer to the [kaniko docs](https://github.com/GoogleContainerTools/kaniko#kubernetes-secret)
+
+(**Note**: Rename the downloaded JSON key to *kaniko-secret* without appending *.json*).
+
+Alternatively, the path to credentials file can be set with the `pullSecretPath` option:
 ```yaml
 build:
   cluster:
     pullSecretName: pull-secret-in-kubernetes
-    # OR
-    pullSecret: path-to-service-account-key-file
+    pullSecretPath: path-to-service-account-key-file-within-secret
+  
 ```
+
+Skaffold can also set up credentials if a secret does not exist in your cluster. This is usually the case when you are performing a 
+build in your personal cluster. 
+
+First, [create a service account](https://cloud.google.com/iam/docs/creating-managing-service-accounts#creating) 
+in the Google Cloud Console project you want to push the final image to with Storage Admin permissions. 
+Download a JSON key for this service account at a convenient location.
+```yaml
+build:
+  cluster:
+    pullSecretPath: path-to-local-service-account-key-file
+```
+Skaffold will create a secret from the service account key file. Skaffold will delete the secret from your cluster at the end of the build.
+
+(**Note**: Please do not check this service account key file in your git repository)
+
+
 Similarly, when pushing to a docker registry:
 ```yaml
 build:
@@ -93,6 +115,34 @@ The following `build` section, instructs Skaffold to build a
 Docker image `gcr.io/k8s-skaffold/example` with Kaniko:
 
 {{% readfile file="samples/builders/kaniko.yaml" %}}
+
+### Configure Kaniko Volume Mounts
+
+You might need to configure volume mounts for a Kaniko pod either to 
+1. Mount a secret or
+2. Set up a [persistent cache](https://github.com/GoogleContainerTools/kaniko/blob/master/examples/kaniko-test.yaml#L27).
+
+To set up volume mounts, configure the `volumeMount` key in `cluster` section like this
+```yaml
+build:
+  artifacts:
+  - image: getting-started
+    kaniko:
+      cache:
+        repo: getting-started
+      volumeMounts:
+      - name: kaniko-cache
+        mountpath: /cache
+        readonly: true
+cluster:
+  volumes:
+  - name: kaniko-cache
+    persistentvolumevlaim:
+    claimname: kaniko-cache-claim
+```
+
+**Note:** Due to [go-yaml/yaml#564](https://github.com/go-yaml/yaml/pull/564), all keys under `kaniko.VolumeMounts` and
+`cluster.Volumes` section should be in lower case. 
 
 ## Dockerfile remotely with Google Cloud Build
 
--- a/integration/examples/kaniko/skaffold.yaml
+++ b/integration/examples/kaniko/skaffold.yaml
@@ -7,7 +7,7 @@ build:
         cache: {}
   cluster:
     pullSecretName: e2esecret-with-path
-    pullSecretPath:     kaniko-secret.json
+    pullSecretPath: kaniko-secret.json
     namespace: default
 deploy:
   kubectl:


docs/content/en/docs/pipeline-stages/builders/docker.md
```diff
@@ -67,15 +67,37 @@ These credentials are configured in the `cluster` section with the following opt
...
+in the Google Cloud Console project you want to push the final image to with Storage Admin permissions. 
+Download a JSON key for this service account at a convenient location.
+```yaml
+build:
+  cluster:
+    pullSecretPath: path-to-local-service-account-key-file
 ```
+Skaffold will create a secret from the service account key file. Skaffold will delete the secret from your cluster at the end of the build.
+
+(**Note**: Please do not check this service account key file in your git repository)
```
briandealwis, line 98: The dev style guide [advocates against _please_](https://developers.google.com/style/tone#politeness-and-use-of-please).  Besides, this seems worthy of being an order :-)

```suggestion
(**Note**: Do not check this service account key file in your git repository)
```

docs/content/en/docs/pipeline-stages/builders/docker.md
```diff
@@ -67,15 +67,37 @@ These credentials are configured in the `cluster` section with the following opt
...
+(**Note**: Rename the downloaded JSON key to *kaniko-secret* without appending *.json*).
+
+Alternatively, the path to credentials file can be set with the `pullSecretPath` option:
 ```yaml
 build:
   cluster:
     pullSecretName: pull-secret-in-kubernetes
-    # OR
-    pullSecret: path-to-service-account-key-file
+    pullSecretPath: path-to-service-account-key-file-within-secret
```
briandealwis, line 81: `within`?  `with` might be better?

docs/content/en/docs/pipeline-stages/builders/docker.md
```diff
@@ -67,15 +67,37 @@ These credentials are configured in the `cluster` section with the following opt
 
 {{< schema root="ClusterDetails" >}}
 
-To set up the credentials for Kaniko refer to the [kaniko docs](https://github.com/GoogleContainerTools/kaniko#kubernetes-secret) (**Note**: Rename the downloaded JSON key to *kaniko-secret* without appending *.json*).
-Alternatively, the path to a credentials file can be set with the `pullSecret` option:
+### Configure Kaniko Credentials 
+
+To set up the credentials for Kaniko refer to the [kaniko docs](https://github.com/GoogleContainerTools/kaniko#kubernetes-secret)
+
+(**Note**: Rename the downloaded JSON key to *kaniko-secret* without appending *.json*).
```
briandealwis, line 74: Should this be placed in the artifact's `context` directory?

docs/content/en/docs/pipeline-stages/builders/docker.md
```diff
@@ -67,15 +67,37 @@ These credentials are configured in the `cluster` section with the following opt
...
   cluster:
     pullSecretName: pull-secret-in-kubernetes
-    # OR
-    pullSecret: path-to-service-account-key-file
+    pullSecretPath: path-to-service-account-key-file-within-secret
+  
+```
+
+Skaffold can also set up credentials if a secret does not exist in your cluster. This is usually the case when you are performing a 
+build in your personal cluster.
```
briandealwis, line 86: To me, `personal cluster` suggests Minikube or something similar.

docs/content/en/docs/pipeline-stages/builders/docker.md
```diff
@@ -67,15 +67,37 @@ These credentials are configured in the `cluster` section with the following opt
...
+Skaffold can also set up credentials if a secret does not exist in your cluster. This is usually the case when you are performing a 
+build in your personal cluster. 
+
+First, [create a service account](https://cloud.google.com/iam/docs/creating-managing-service-accounts#creating) 
+in the Google Cloud Console project you want to push the final image to with Storage Admin permissions. 
+Download a JSON key for this service account at a convenient location.
+```yaml
+build:
+  cluster:
+    pullSecretPath: path-to-local-service-account-key-file
```
briandealwis, line 94: It's unfortunate this is called `pull`?

docs/content/en/docs/pipeline-stages/builders/docker.md
tejal29: no. i added a note down that please do not check in secret in the directory. This secret could be at any location.

docs/content/en/docs/pipeline-stages/builders/docker.md
tejal29: actually it is within the secret. 
e.g.
```
kubectl create secret generic kaniko-secret --from-file=~/kaniko-secret.json`
```
The secret created will be loaded at path `kaniko-secret.json`.

docs/content/en/docs/pipeline-stages/builders/docker.md
tejal29: it could be your personal (local or remote) GKE cluster  In case developers are using. 
Is there an umbrella term for that?

docs/content/en/docs/pipeline-stages/builders/docker.md
tejal29: This secret is only used for pulling images and we construct `GoogleApplicationCredentials` from it. 
Hence it was named as pull secret originally. But yes something we could re-think

docs/content/en/docs/pipeline-stages/builders/docker.md
briandealwis: Sorry, I'm not very adept remote cluster builds with Kaniko, and I'm confused by both this section and the linked to Kaniko secrets.  The linked-to Kaniko document uses kubectl to create a secret wrapping the service-account key file to access GCR.

Would it make sense to just copy the relevant steps here?

It's also not clear to me where I'd put this `kaniko-secret` file.  It seems there's a standard place, given the "Alternatively, …" sentence below.

docs/content/en/docs/pipeline-stages/builders/docker.md
```diff
@@ -67,15 +67,37 @@ These credentials are configured in the `cluster` section with the following opt
...
 
-To set up the credentials for Kaniko refer to the [kaniko docs](https://github.com/GoogleContainerTools/kaniko#kubernetes-secret) (**Note**: Rename the downloaded JSON key to *kaniko-secret* without appending *.json*).
-Alternatively, the path to a credentials file can be set with the `pullSecret` option:
+### Configure Kaniko Credentials 
+
+To set up the credentials for Kaniko refer to the [kaniko docs](https://github.com/GoogleContainerTools/kaniko#kubernetes-secret)
+
+(**Note**: Rename the downloaded JSON key to *kaniko-secret* without appending *.json*).
+
+Alternatively, the path to credentials file can be set with the `pullSecretPath` option:
```
briandealwis, line 76: What is this alternative to?  The `(**Note**: …` or the `To set up the credentials …`?

docs/content/en/docs/pipeline-stages/builders/docker.md
briandealwis: I think it would be helpful to explicitly highlight that this is for GKE/GCR?

docs/content/en/docs/pipeline-stages/builders/docker.md
chanseokoh: No wonder it's confusing.

When creating a secret with
```
kubectl create secret generic kaniko-secret --from-file=<path to kaniko-secret.json>
```
the key name of the `data:` map in the created secret is set to the filename given by `--from-file`. So, if you follow the linked Kaniko doc, you'll end up having the following for the secret.
```yaml
data:
  kaniko-secret.json: <... secret data ...>
``` 
However, Skaffold by default [expects](https://github.com/GoogleContainerTools/skaffold/issues/3828#issuecomment-626965077) you to create the secret with the key name `kaniko-secret`. That's the reason for the instruction to rename the JSON key file to `kaniko-secret` without `.json`.

docs/content/en/docs/pipeline-stages/builders/docker.md
chanseokoh: The former (i.e., `**Note**`).

Recently `pullSecretPath` was converted to serve as a dual-purpose property (#4381) to address the issue that Skaffold uses the hard-coded secret key name `kaniko-secret`. So, instead of renaming the JSON key file to `kaniko-secret` (without `.json`), you can now say, e.g., `pullSecretPath: my-key.json` if you created your secret with
```
kubectl create secret generic kaniko-secret --from-file=.../my-key.json
```
Another alternative not documented here is to explicitly specify the key name with
```
kubectl create secret generic kaniko-secret --from-file=kaniko-secret=.../my-key.json
```

docs/content/en/docs/pipeline-stages/builders/docker.md
tejal29: > It's also not clear to me where I'd put this `kaniko-secret` file. It seems there's a standard place, given the "Alternatively, …" sentence below.

We could copy the instructions here instead of pointing to kaniko docs.

docs/content/en/docs/pipeline-stages/builders/docker.md
tejal29: Kaniko can also on EKS, Azure hence i mentioned personal cluster. 
What i want to focus on is, usually if you are in a shared cluster env, secrets will be installed by admins and developers don't have access to create them

However, when you are using a personal cluster to deploy or in personal namespace on a shared cluster, you would need to create you own secret.

docs/content/en/docs/pipeline-stages/builders/docker.md
MarlonGamez: I agree that we should copy the instructions to these docs. It makes for a much easier read through them

docs/content/en/docs/pipeline-stages/builders/docker.md
gsquared94: maybe reword to :
```
Skaffold can also set up credentials if a secret does not exist in your cluster. This is usually the case when you are not using a shared cluster where the cluster admin would have already configured it for you. 
```

docs/content/en/docs/pipeline-stages/builders/docker.md
```diff
@@ -94,6 +116,33 @@ Docker image `gcr.io/k8s-skaffold/example` with Kaniko:
...
+      cache:
+        repo: getting-started
+      volumeMounts:
+      - name: kaniko-cache
+        mountpath: /cache
+        readonly: true
+cluster:
+  volumes:
+  - name: kaniko-cache
+    persistentvolumevlaim:
```
briandealwis, line 140:
```suggestion
    persistentvolumeclaim:
```

briandealwis: I think we need to retrieve this from a file so that we can do some validation as I think this example need another level of indirection for `volumesource` and should actually be:

```
    volumes:
      - name: kaniko-cache
        volumesource:
          persistentvolumeclaim: 
            claimname: kaniko-cache-claim
```

PR #4367, opened 2020-06-22
Move jib sync testdata to `integration/examples`

loosebazooka:
- separate out maven/gradle skaffold files into skaffold-maven.yaml
  and skaffold-gradle.yaml
- add detailed README.md to example
- does NOT copy example to in `/examples` yet


integration/examples/jib-sync/README.md
```diff
@@ -0,0 +1,165 @@
...
+    jib: {}
+    sync: 
+      auto: {}
+```
+
+This example is designed around the functionality available in [Spring Boot Developer Tools](https://docs.spring.io/spring-boot/docs/current/reference/html/using-spring-boot.html#using-boot-devtools) for developing against running applications.
+
+Some additional steps in your java build are required for this to work:
+- Sync requires `tar` on the running container to copy files over. The default base image that Jib uses `gcr.io/distroless/java` does not include `tar` or any utilities. During development you must use a base image that includes `tar`, in this example we use the `debug` flavor of distroless: `gcr.io/distroless/java:debug` 
+`maven`
```
chanseokoh, line 68: `maven` is not properly rendered. Probably put a new line on it.

integration/examples/jib-sync/README.md
```diff
@@ -0,0 +1,165 @@
...
+```groovy
+jib {
+  ...
+  from {
+    image = "gcr.io/distroless/java:debug"
+  }
+}
+```
+
+- You must include the `spring-boot-devtools` dependency at the `compile/implementation` scope, which is contrary to the configuration outlined in the [official docs](https://docs.spring.io/spring-boot/docs/current/reference/html/using-spring-boot.html#using-boot-devtools). Because jib is unaware of any special spring only configuration in your builds, we recommend using profiles to turn on or off devtools support in your jib container builds.
```
chanseokoh, line 93: I think Jib currently includes `spring-boot-devtools` even if `<optional>true`. (And we decided to take care of this using an extension.) Should we still advise to set `<scope>compile`?

integration/examples/jib-sync/skaffold-gradle.yaml
```diff
@@ -0,0 +1,16 @@
...
+      type: gradle
+      args: 
+      - -Psync
+    sync:
+      auto: {}
+
+portForward:
+- resourceType: pod
+  resourceName: test-file-sync
+  port: 8080
```
chanseokoh, line 16: Interesting that this doesn't exist in `skaffold-maven.yaml`.

integration/examples/jib-sync/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+`skaffold.yaml`
+```
+build:
+  artifacts:
+  - image: skaffold-example
+    context: .
+    jib: 
+      args: 
+        - -Psync
```
chanseokoh, line 131: For consistency,

```suggestion
      args: 
      - -Psync
```

integration/examples/jib-sync/README.md
```diff
@@ -0,0 +1,165 @@
...
+  - op: add
+    # we want to activate sync on our skaffold artifact
+    path: /build/artifacts/0/sync
+    value:
+      - auto: {}
+  - op: add
+    # we activate the sync profile in our java builds
+    path: /build/artifacts/0/jib/args
+    value:
+      - -Psync
```
chanseokoh, line 158: For consistency,

```suggestion
    value:
    - -Psync
```

integration/examples/jib-sync/skaffold-gradle.yaml
loosebazooka: oops...

integration/examples/jib-sync/README.md
loosebazooka: I think we leave optional dependencies out. no?

loosebazooka: because it's not explicitly resolved to be a runtime dependency.

loosebazooka: oh about the scope, `<compile>` is default isn't it?

integration/examples/jib-sync/README.md
chanseokoh: This is still not correctly rendered. `maven` is on the same line with `gcr.io/distroless/java:debug`.

Screenshot:

image

integration/examples/jib-sync/README.md
chanseokoh: Ah, right, `<compile>` is already the default. The Spring Boot doc only adds `<optional>true`, so this comment doesn't seem to apply.

And I think we include `spring-boot-devtools` regardless of `<optional>true`. (Would be good to double-check.)

loosebazooka: yeah you're right. I can say

integration/examples/jib-sync/build.gradle
```diff
@@ -13,11 +13,11 @@ sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
 dependencies {
-  compile "org.springframework.boot:spring-boot-starter-web"
-  compile "org.springframework.boot:spring-boot-starter-actuator"
-  compile "org.springframework.boot:spring-boot-devtools"
+  implementation "org.springframework.boot:spring-boot-starter-web"
+  implementation "org.springframework.boot:spring-boot-starter-actuator"
+  implementation "org.springframework.boot:spring-boot-devtools"
```
chanseokoh, line 18: I see that you updated Maven `pom.xml` to selectively add the devtools dependency with a profile. Should we also apply this only when `-Psync` for Gradle?

loosebazooka: I thought I did, but I guess I didn't. :\

PR #4390, opened 2020-06-24
Refactor Add proto.ActionableErr to diag.Resource and deploy.Resource.Status

tejal29:
In this PR, I have added the actionable error message to the diagnostics package. 
For suggesting actionable errors for k8 infra errors, it would make sense to do it in the diag package.

Got rid of `details` in `skaffold.deploy.Resource.Status` to re-use `proto.ActionableErr.Message`.

Adding ActionableErr to daig.Resource package and propagating to deployment status in
skaffold.deploy.Resource.Status

Most of the code changes are in test.  Very minor changes in source code files.

Fixes: #nnn 

**Description**


pkg/diag/validator/pod.go
```diff
@@ -209,34 +209,35 @@ func processPodEvents(e corev1.EventInterface, pod v1.Pod, ps *podStatus) {
 	}
 	switch recentEvent.Reason {
 	case failedScheduling:
-		ps.statusCode = proto.StatusCode_STATUSCHECK_FAILED_SCHEDULING
-		ps.err = fmt.Errorf(recentEvent.Message)
+		ps.ae.ErrCode = proto.StatusCode_STATUSCHECK_FAILED_SCHEDULING
+		ps.ae.Message = recentEvent.Message
```
briandealwis, line 213: These two-step struct accesses kinda go against the Law of Demeter — basically you should try to only manipulate information one step away.  But this function would make sense as a method on `podStatus *`. WDYT?

pkg/diag/validator/pod.go
```diff
@@ -209,34 +209,35 @@ func processPodEvents(e corev1.EventInterface, pod v1.Pod, ps *podStatus) {
...
 func (p *podStatus) isStable() bool {
-	return p.phase == success || (p.phase == running && p.err == nil)
+	return p.phase == success || (p.phase == running && p.ae.Message == "")
 }
 
 func (p *podStatus) withErrAndLogs(errCode proto.StatusCode, l []string, err error) *podStatus {
-	p.err = err
-	p.statusCode = errCode
+	if err != nil {
+		p.ae.Message = err.Error()
```
briandealwis, line 238: Should we have an else clause to reset it to "" otherwise?

pkg/diag/validator/pod.go
```diff
@@ -209,34 +209,35 @@ func processPodEvents(e corev1.EventInterface, pod v1.Pod, ps *podStatus) {
...
+	name      string
+	namespace string
+	phase     string
+	logs      []string
+	ae        *proto.ActionableErr
 }
 
 func (p *podStatus) isStable() bool {
-	return p.phase == success || (p.phase == running && p.err == nil)
+	return p.phase == success || (p.phase == running && p.ae.Message == "")
```
briandealwis, line 233: Shouldn't this code check that `p.ae != nil`?

pkg/diag/validator/pod.go
```diff
@@ -275,10 +276,12 @@ func extractErrorMessageFromWaitingContainerStatus(po *v1.Pod, c v1.ContainerSta
...
 func newPodStatus(n string, ns string, p string) *podStatus {
 	return &podStatus{
-		name:       n,
-		namespace:  ns,
-		phase:      p,
-		statusCode: proto.StatusCode_STATUSCHECK_SUCCESS,
+		name:      n,
+		namespace: ns,
+		phase:     p,
+		ae: &proto.ActionableErr{
```
briandealwis, line 282: Does it need to be a pointer if it's always set?

pkg/diag/validator/resource.go
```diff
@@ -26,31 +26,35 @@ import (
...
 	if r.namespace == "default" {
 		return fmt.Sprintf("%s/%s", r.kind, r.name)
 	}
 	return fmt.Sprintf("%s:%s/%s", r.namespace, r.kind, r.name)
 }
+func (r Resource) ActionableError() *proto.ActionableErr {
+	return r.ae
+}
+func (r Resource) StatusUpdated(another Resource) bool {
+	return r.ae.ErrCode != another.ae.ErrCode
```
briandealwis, line 52: Should this code check if `r.ae != nil && another.ae != nil` ?

pkg/skaffold/event/event.go
```diff
@@ -217,9 +217,9 @@ func StatusCheckEventInProgress(s string) {
 	})
 }
 
-func ResourceStatusCheckEventCompleted(r string, statusCode proto.StatusCode, err error) {
-	if err != nil {
-		resourceStatusCheckEventFailed(r, statusCode, err)
+func ResourceStatusCheckEventCompleted(r string, ae *proto.ActionableErr) {
+	if ae.ErrCode != proto.StatusCode_STATUSCHECK_SUCCESS {
```
briandealwis, line 221: This needs to check `ae == nil`

pkg/skaffold/event/event.go
```diff
@@ -234,24 +234,23 @@ func resourceStatusCheckEventSucceeded(r string) {
...
-func ResourceStatusCheckEventUpdated(r string, statusCode proto.StatusCode, status string) {
+func ResourceStatusCheckEventUpdated(r string, ae *proto.ActionableErr) {
 	handler.handleResourceStatusCheckEvent(&proto.ResourceStatusCheckEvent{
-		Resource:   r,
-		Status:     InProgress,
-		Message:    status,
-		StatusCode: statusCode,
+		Resource:      r,
+		Status:        InProgress,
+		Message:       ae.Message,
```
briandealwis, line 251: What if `ae == nil`?

pkg/diag/validator/pod.go
tejal29: added this to be regular struct

pkg/skaffold/event/event.go
tejal29: not needed any more since its a struct

pkg/skaffold/event/event.go
tejal29: not needed any more since its a struct

pkg/diag/validator/pod.go
tejal29: with this changed to a struct, not needed anymore

pkg/diag/validator/pod.go
tejal29: with new code changes, it will be reset to ""

PR #4403, opened 2020-06-27
Automate Release PR creation

gsquared94:
Use Github Actions to create the Changelog PR required for creating a release. 
 

Fixes: #4324  

**Description**

 - Trigger: Creating a new release branch `release/vX.Y.Z`. 
 - Automated actions:
-- Push changes to this branch for CHANGELOG.md, skaffold config comment update, examples directory update.
-- PR created against master with label `draft-release`

After manually verifying the PR, on merging it it'll combine with the existing flow of creating a draft release.

To create a new release the only steps necessary will be:
- Create new branch `release/v**`
- Approve release notes Pull Request
- Approve Release Draft

This requires removing this [branch protection rule](https://github.com/GoogleContainerTools/skaffold/settings/branch_protection_rules/6022668) for `release/**` so that Github Actions can push changes directly to the release branch created.

This requires labelling Pull Requests with one of the following labels so that they're put in the corresponding section of the release notes changelog:
- `release/new-feature`, `release/fixes`, `release/refactor`, `release/doc-updates`

### DIFF
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -18,7 +18,7 @@ jobs:
       run: |
         t=$(echo ${{ github.ref }} | sed -ne 's/.*\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/p')
         if [ -z "$t" ]; then
-        echo Malformed release branch name; failed to extract semVer tag
+        echo "Malformed release branch name ${{ github.ref }}; failed to extract semVer tag"
         return 1
         fi
         echo ::set-env name=TAG_NAME::"v${t}"
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -32,7 +32,7 @@ jobs:
 
     - name: Generate Changelog
       run: |
-        release-notes GoogleContainerTools skaffold --since=v1.11.0 --token=${{ secrets.GITHUB_TOKEN }} > $HOME/change.txt
+        release-notes GoogleContainerTools skaffold --token=${{ secrets.GITHUB_TOKEN }} > $HOME/change.txt
 
     - name: Sync examples
       run: |
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -1,0 +1,69 @@
+name: Create Release PR
+on:
+  create:
+    branches:
+    - 'release/v**'
+
+jobs:
+  build:
+    name: Create Release PR
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Calculate Tag
+      run: |
+        t=$(echo ${{ github.ref }} | sed -ne 's/.*\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/p')
+        if [ -z "$t" ]; then
+        echo Malformed release branch name; failed to extract semVer tag
+        return 1
+        fi
+        echo ::set-env name=TAG_NAME::"v${t}"
+
+    - name: Install release-notes
+      run: |
+        mkdir -p $HOME/deps
+        curl -Lo $HOME/release-notes.tar.gz https://github.com/gsquared94/release-notes/releases/download/v0.1.2/release-notes-linux-amd64.tar.gz
+        tar xfz $HOME/release-notes.tar.gz -C $HOME/deps
+        echo ::add-path::$HOME/deps
+
+    - name: Generate Changelog
+      run: |
+        release-notes GoogleContainerTools skaffold --since=v1.11.0 --token=${{ secrets.GITHUB_TOKEN }} > $HOME/change.txt
+
+    - name: Sync examples
+      run: |
+        EXAMPLES_DIR=${PWD}/examples
+        INTEGRATION_EXAMPLES_DIR=${PWD}/integration/examples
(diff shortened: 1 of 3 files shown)


gsquared94:
To test this PR you can:
- Set the master branch of your fork to this commit.
- Either:
   - Label a few PRs merged since the last skaffold release with one of `release/new-feature`, `release/fixes`, `release/refactor`, `release/doc-updates`
Or
   - Add flag `--since=v1.11.0` to line 35 of `.github/workflows/release-pr.yml` which will ignore the newer releases while calculating the diff
- Create a branch `release/vX.Y.Z`
- Verify that the PR `Release vX.Y.Z` gets created with the correct file changes.

.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,69 @@
...
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Calculate Tag
+      run: |
+        t=$(echo ${{ github.ref }} | sed -ne 's/.*\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/p')
+        if [ -z "$t" ]; then
+        echo Malformed release branch name; failed to extract semVer tag
+        return 1
```
briandealwis, line 22: Should echo the branch name for diagnosing.  Tabbing seems off.  Oh, and you should quote the string as this will attempt to echo and then execute `failed` as a command.
```suggestion
          echo "Malformed release branch name '${{ github.ref }}': failed to extract semVer tag"
          return 1
```

(I hope the `${{ x }}` works within strings?)

.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,69 @@
...
+
+    - name: Generate Changelog
+      run: |
+        release-notes GoogleContainerTools skaffold --token=${{ secrets.GITHUB_TOKEN }} > $HOME/change.txt
+
+    - name: Sync examples
+      run: |
+        EXAMPLES_DIR=${PWD}/examples
+        INTEGRATION_EXAMPLES_DIR=${PWD}/integration/examples
+        rm -rf ${EXAMPLES_DIR} && rm -rf ${INTEGRATION_EXAMPLES_DIR}/bazel/bazel-* && cp -r ${INTEGRATION_EXAMPLES_DIR} ${EXAMPLES_DIR} && rm -rf ${EXAMPLES_DIR}/test-*
```
briandealwis, line 41: The following will wipe out the integration tests as part of the commit, won't it?  Shouldn't it be done in `EXAMPLES_DIR` after the copy?
```
rm -rf ${INTEGRATION_EXAMPLES_DIR}/bazel/bazel-*
```

gsquared94: This is the same command being run in `release.sh` today. It's probably for deleting these folders that get created when running this example locally:
image

That's redundant here since these folders don't exist on master today and are also added to `gitignore`. But I don't see any harm in leaving it here just to be sure.

.github/workflows/release-pr.yml
gsquared94: done

.github/workflows/release-pr.yml
briandealwis: Ohhh, that makes sense.

briandealwis: It's worth a comment though

.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,69 @@
...
+        if [ -z "$t" ]; then
+        echo "Malformed release branch name ${{ github.ref }}; failed to extract semVer tag"
+        return 1
+        fi
+        echo ::set-env name=TAG_NAME::"v${t}"
+
+    - name: Install release-notes
+      run: |
+        mkdir -p $HOME/deps
+        curl -Lo $HOME/release-notes.tar.gz https://github.com/gsquared94/release-notes/releases/download/v0.1.2/release-notes-linux-amd64.tar.gz
```
briandealwis, line 29: I have a mild concern about requiring a fork.

gsquared94:
> So I have a few concerns:
> 
> 1. _"This requires removing this branch protection rule for release/** so that Github Actions can push changes directly to the release branch created." — my reading of [OSPO's policies around Github Actions](https://opensource.google/docs/github/services/#actions) says we shouldn't do this (though the nebulous `and/or` makes it less clear).

From what I understood it specifies that these policies are necessary for `master`, `develop` or `feature` branches. If it's applicable to `release` branches too then we wouldn't be able to auto push changes to the release branches even with bots(?)

> 2. (mild) I have a mild preference to keep the bulk of this automation in `hack/release.sh` as it's easier to test locally, and it reduces the effort to switch to yet-another-releasing orchestrator.

I actually prefer this too but didn't want to modify the existing release script to keep the old way alive until this stabilized. I could create a `release-actions.sh` and push a chunk of the code there if that looks better?

> 3. The use of PR labels is going to require more automation, and it seems like there must be a better way than our creating our own bots to enforce such labels.

Agree. Your link to conventional commits seems a cleaner way of doing this. But that will:
 1. Require additional setup to enforce squash merges since Changelog would be generated from commit messages instead of PR titles.
 2. IMHO PR titles tend to be more descriptive and better formatted than commit messages but I could be wrong.

.github/workflows/release-pr.yml
gsquared94: I forked that project since I wanted to be able to categorize the changelog and that feature was missing there. Couldn't follow the concern part. Should I have done this differently?

tejal29:
IIUC, to automate the release, i will have to create a new release branch and push it to origin without any changes right?

gsquared94:
> IIUC, to automate the release, i will have to create a new release branch and push it to origin without any changes right?

@tejal29 I'm closing this PR for now. After discussion with @briandealwis and @nkubala, we're probably gonna rework the whole release flow around kokoro later. 

Also without this, the Release Draft automation doesn't really work. So we're back to manual releases until we sort this out.

tejal29:
Closing the PR

PR #4425, opened 2020-07-01
Use docker/cli code to send context to `docker build`

### LINKED ISSUE #4328: Skaffold complains about the docker file being outside the docker build context
### Expected behavior
That skaffold has no more restrictions on the location of the dockerfile than docker build has
### Actual behavior
Skaffold complains that the dockerfile is outside the build context 
### Information

```yaml
apiVersion: skaffold/v2beta4
kind: Config
metadata:
  name: hobbes
build:
  tagPolicy:
    sha256: {}
  artifacts:
    - image: az
      context: workers/az.worker/
      docker:
        dockerfile: ../../docker/Dockerfile.worker
```

any setup where the docker file is outside of the specified context

dgageot:
Instead of creating ourself the docker build context, we rely on code copied from https://github.com/docker/cli.

This has two advantages:
 + Dockerfiles that are out of the build context are now supported (Fixes #4328)
 + `skaffold build` and `docker build` from the same source generate the exact same image

This has one disadvantage:
 + `skaffold build` used to be a bit smarter than `docker build` and wouldn't send files that are not actually copied from the build context to the container. This is not the case anymore.

This new method or copying the build context is used for:
 + [x] local docker build
 + [x] kaniko builds
 + [ ] gcb builds 

### DIFF
--- a/pkg/skaffold/build/cluster/kaniko.go
+++ b/pkg/skaffold/build/cluster/kaniko.go
@@ -35,13 +35,18 @@ import (
 const initContainer = "kaniko-init-container"
 
 func (b *Builder) buildWithKaniko(ctx context.Context, out io.Writer, workspace string, artifact *latest.KanikoArtifact, tag string) (string, error) {
+	buildCtx, relDockerfile, err := docker.BuildContext(workspace, artifact.DockerfilePath)
+	if err != nil {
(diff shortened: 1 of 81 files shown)


dgageot:
Closing for now

Issue #4370, opened 2020-06-22
Skaffold helps build debuggable containers


briandealwis:
Builds from certain language runtimes would benefit from being informed of whether the build is for debugging or development vs production.  For example, Go builds for debug should be built with `-gcflags='-N -l'` to disable optimizations.

gsquared94:
I can take a stab at this

PR #4467, opened 2020-07-14
Add `ImageOptions` to artifact builder.

gsquared94:
Fixes: #4370  

**Description**

This PR introduces `ImageOptions` to the different builders and currently only partially implements it for docker builder and buildpacks.
It recognizes `Dev` and `Debug` as build configurations and passes different build args like `GO_GCFLAGS` when building the artifact for debug vs dev. It also updates hashcode formula for image caching so that images for `Dev` vs `Debug` are cached separately.

Dockerfile can expose `ARG SKAFFOLD_GO_GCFLAGS` and `ARG SKAFFOLD_GO_LDFLAGS` and expect Skaffold to supply meaningful values.
Buildpacks auto set `GOOGLE_GCFLAGS` and `GOOGLE_LDFLAGS` (so far).

Note: _The artifact cached during `skaffold dev` will have a different hashcode than `skaffold debug` for all builders going forward_

### DIFF
--- a/integration/examples/getting-started/Dockerfile
+++ b/integration/examples/getting-started/Dockerfile
@@ -1,11 +1,11 @@
 FROM golang:1.12.9-alpine3.10 as builder
 
 # These flag values are provided when building for debug
-ARG GO_GCFLAGS
-ARG GO_LDFLAGS
+ARG SKAFFOLD_GO_GCFLAGS
+ARG SKAFFOLD_GO_LDFLAGS
 
 COPY main.go .
-RUN eval go build --gcflags="${GO_GCFLAGS}" --ldflags="${GO_LDFLAGS}" -o /app main.go
+RUN eval go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -ldflags="${SKAFFOLD_GO_LDFLAGS}" -o /app main.go
 
 FROM alpine:3.10
 # Define GOTRACEBACK to mark this container as using the Go language runtime
--- a/integration/examples/getting-started/Dockerfile
+++ b/integration/examples/getting-started/Dockerfile
@@ -1,6 +1,11 @@
 FROM golang:1.12.9-alpine3.10 as builder
+
+# These flag values are provided when building for debug
+ARG GO_GCFLAGS
+ARG GO_LDFLAGS
+
 COPY main.go .
-RUN go build -o /app main.go
+RUN eval go build --gcflags="${GO_GCFLAGS}" --ldflags="${GO_LDFLAGS}" -o /app main.go
 
 FROM alpine:3.10
 # Define GOTRACEBACK to mark this container as using the Go language runtime
--- a/integration/testdata/debug/go/Dockerfile
+++ b/integration/testdata/debug/go/Dockerfile
@@ -1,9 +1,10 @@
 FROM golang:1.12 as builder
+ARG SKAFFOLD_GO_GCFLAGS
 
 COPY app.go .
-# Must use eval to handle GOGCFLAGS with spaces like `-gcflags='all=-N -l'`
-ARG GO_GCFLAGS
-RUN eval go build --gcflags="${GO_GCFLAGS}" -o /app .
+RUN echo "SKAFFOLD_GO_GCFLAGS: ${SKAFFOLD_GO_GCFLAGS}"
+# Must use eval to handle GOGCFLAGS with spaces like `all=-N -l`
+RUN eval go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -o /app .
 
 FROM gcr.io/distroless/base
 # `skaffold debug` uses GOTRACEBACK as an indicator of the Go runtime
--- a/integration/testdata/debug/go/Dockerfile
+++ b/integration/testdata/debug/go/Dockerfile
@@ -1,10 +1,10 @@
 FROM golang:1.12 as builder
-ARG SKAFFOLD_GO_GCFLAGS
 
 COPY app.go .
-RUN echo "SKAFFOLD_GO_GCFLAGS: ${SKAFFOLD_GO_GCFLAGS}"
-# Must use eval to handle GOGCFLAGS with spaces like `all=-N -l`
-RUN eval go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -o /app .
+ARG SKAFFOLD_GO_GCFLAGS
+ARG SKAFFOLD_LD_GCFLAGS
+# Must use eval to handle gcflags with spaces like `all=-N -l`
+RUN eval go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -ldflags="${SKAFFOLD_GO_LDFLAGS}" -o /app .
 
 FROM gcr.io/distroless/base
 # `skaffold debug` uses GOTRACEBACK as an indicator of the Go runtime
--- a/integration/testdata/debug/go/Dockerfile
+++ b/integration/testdata/debug/go/Dockerfile
@@ -2,8 +2,8 @@ FROM golang:1.12 as builder
 
 COPY app.go .
 # Must use eval to handle GOGCFLAGS with spaces like `-gcflags='all=-N -l'`
-ARG GOGCFLAGS
-RUN eval go build "${GOGCFLAGS}" -o /app .
+ARG GO_GCFLAGS
+RUN eval go build "${GO_GCFLAGS}" -o /app .
 
 FROM gcr.io/distroless/base
 # `skaffold debug` uses GOTRACEBACK as an indicator of the Go runtime
--- a/integration/testdata/debug/go/Dockerfile
+++ b/integration/testdata/debug/go/Dockerfile
@@ -3,7 +3,7 @@ FROM golang:1.12 as builder
 COPY app.go .
 # Must use eval to handle GOGCFLAGS with spaces like `-gcflags='all=-N -l'`
 ARG GO_GCFLAGS
-RUN eval go build "${GO_GCFLAGS}" -o /app .
+RUN eval go build --gcflags="${GO_GCFLAGS}" -o /app .
 
 FROM gcr.io/distroless/base
 # `skaffold debug` uses GOTRACEBACK as an indicator of the Go runtime
--- a/pkg/skaffold/build/cache/hash.go
+++ b/pkg/skaffold/build/cache/hash.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"sort"
 
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
 	"github.com/sirupsen/logrus"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/misc"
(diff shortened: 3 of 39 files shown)


integration/examples/getting-started/Dockerfile
```diff
@@ -1,6 +1,11 @@
 FROM golang:1.12.9-alpine3.10 as builder
+
+# These flag values are provided when building for debug
+ARG GO_GCFLAGS
```
gsquared94, line 4: `GO_GCFLAGS` and `GO_LDFLAGS` are explicitly added as dockerfile args and their values are provided by Skaffold

integration/testdata/debug/go/Dockerfile
```diff
@@ -2,8 +2,8 @@ FROM golang:1.12 as builder
 
 COPY app.go .
 # Must use eval to handle GOGCFLAGS with spaces like `-gcflags='all=-N -l'`
-ARG GOGCFLAGS
-RUN eval go build "${GOGCFLAGS}" -o /app .
+ARG GO_GCFLAGS
```
gsquared94, line 5: revert this change.

integration/testdata/debug/go/Dockerfile
```diff
@@ -2,8 +2,8 @@ FROM golang:1.12 as builder
...
+RUN eval go build "${GO_GCFLAGS}" -o /app .
```
gsquared94, line 6: fix command `--gcflags="${GO_GCFLAGS}"`

pkg/skaffold/build/cache/hash.go
```diff
@@ -89,6 +90,15 @@ func getHashForArtifact(ctx context.Context, depLister DependencyLister, a *late
 		inputs = append(inputs, evaluatedEnv...)
 	}
 
+	// add image options hash
```
gsquared94, line 93: cache takes `opts` hash value

pkg/skaffold/build/cluster/cluster.go
```diff
@@ -48,13 +48,13 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, tags tag.ImageTags,
 	return build.InParallel(ctx, out, tags, artifacts, b.buildArtifact, b.ClusterDetails.Concurrency)
 }
 
-func (b *Builder) buildArtifact(ctx context.Context, out io.Writer, artifact *latest.Artifact, tag string) (string, error) {
-	digest, err := b.runBuildForArtifact(ctx, out, artifact, tag)
+func (b *Builder) buildArtifact(ctx context.Context, out io.Writer, artifact *latest.Artifact, opts *build.ImageOptions) (string, error) {
```
gsquared94, line 51: all builders accept `build.ImageOptions`

pkg/skaffold/build/options.go
```diff
@@ -0,0 +1,85 @@
...
+// Configuration denotes build configuration for the artifact builder as either Release or Debug
+type Configuration int
+
+const (
+	Release = iota
+	Debug
+)
+
+// ImageOptions provides options for the artifact builder
+type ImageOptions struct {
```
gsquared94, line 37: `ImageOptions` encapsulates image tag along with the current configuration (Debug or Release) and a map of additional properties

gsquared94, line 37: `ImageOptions` encapsulates image tag along with the current configuration (Debug or Release) and a map of additional properties

pkg/skaffold/docker/image.go
```diff
@@ -190,7 +190,10 @@ func (l *localDaemon) Build(ctx context.Context, out io.Writer, workspace string
 		ForceRemove: l.forceRemove,
 		NetworkMode: strings.ToLower(a.NetworkMode),
 		NoCache:     a.NoCache,
-	})
+	}
+
+	imageOpts = opts.ApplyModifiers(imageOpts)
```
gsquared94, line 195: Applying docker builder options

pkg/skaffold/docker/options.go
```diff
@@ -0,0 +1,84 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package docker
+
+import "github.com/docker/docker/api/types"
```
gsquared94, line 19: Docker builder options implementation

pkg/skaffold/build/options.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+var CurrentConfiguration Configuration
+
+func CreateBuilderOptions(tag string) *ImageOptions {
+	return &ImageOptions{
+		Tag:           tag,
+		Configuration: CurrentConfiguration,
+	}
+}
```
briandealwis, line 51: We really should avoid global state.  [`SkaffoldOptions.Command`](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/config/options.go#L76) records the command used on the command-line, and offers [`IsDevMode()` and `IsDebugMode()`](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/config/options.go#L92-L98) to test.

I wonder if it would be worth broadening the scope of this structure to capture the build options, including the artifact, and have it be passed in and around.

pkg/skaffold/build/local/local.go
```diff
@@ -112,3 +113,11 @@ func (b *Builder) getImageIDForTag(ctx context.Context, tag string) (string, err
 	}
 	return insp.ID, nil
 }
+
+func ToDockerOpts(opts *build.ImageOptions) *docker.BuildOptions {
```
briandealwis, line 117: This should be done in the docker builder itself.

pkg/skaffold/build/local/local.go
```diff
@@ -86,19 +87,19 @@ func (b *Builder) runBuildForArtifact(ctx context.Context, out io.Writer, a *lat
 
 	switch {
 	case a.DockerArtifact != nil:
-		return b.buildDocker(ctx, out, a, tag)
+		return b.buildDocker(ctx, out, a, ToDockerOpts(opts))
```
briandealwis, line 90: It feels like this ImageOptions should be passed down into the builders.

pkg/skaffold/build/local/local.go
```diff
@@ -45,8 +46,8 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, tags tag.ImageTags,
 	return build.InParallel(ctx, out, tags, artifacts, b.buildArtifact, *b.cfg.Concurrency)
 }
 
-func (b *Builder) buildArtifact(ctx context.Context, out io.Writer, a *latest.Artifact, tag string) (string, error) {
-	digestOrImageID, err := b.runBuildForArtifact(ctx, out, a, tag)
+func (b *Builder) buildArtifact(ctx context.Context, out io.Writer, a *latest.Artifact, opts *build.ImageOptions) (string, error) {
```
briandealwis, line 49: Since we're passing in the destination image tag, opts shouldn't be a pointer.

pkg/skaffold/docker/options.go
```diff
@@ -0,0 +1,84 @@
...
+
+var buildArgsForDebug = map[string]string{
+	"GO_GCFLAGS": "'all=-N -l'", // disable build optimization for Golang
```
briandealwis, line 22: What do you think about prefixing these with `SKAFFOLD_`?

pkg/skaffold/build/options.go
gsquared94: Since it's in package `build` struct names shouldn't start with `Build-`(at least the IDE prompts so 🤷🏾‍♂️) which is why I named it so; but I'll think of something else.

pkg/skaffold/build/options.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"sort"
+	"strconv"
+)
+
+// Configuration denotes build configuration for the artifact builder as either Release or Debug
+type Configuration int
+
+const (
+	Release = iota
```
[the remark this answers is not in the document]
gsquared94: I've changed this to only have `Dev` and `Debug` since I also use this parameter in calculating cache hashcode. So having separate `Build` type would cause the image built by `skaffold build` not to be used for next `skaffold dev`

pkg/skaffold/build/local/local.go
gsquared94: Pushing `ImageOptions` into any of the actual builder implementations causes an import cycle (you had predicted this!). So I've defined `ImageOptions` as the generic builder-agnostic options object. The different builders define their own options object in their respective adapter package.

pkg/skaffold/build/local/local.go
gsquared94: > Copied from above

Pushing ImageOptions into any of the actual builder implementations causes an import cycle (you had predicted this!). So I've defined ImageOptions as the generic builder-agnostic options object. The different builders define their own options object in their respective adapter package.

pkg/skaffold/build/local/local.go
gsquared94: done

pkg/skaffold/docker/options.go
gsquared94: done

pkg/skaffold/build/options.go
gsquared94: Done

pkg/skaffold/build/options.go
```diff
@@ -0,0 +1,85 @@
...
+	Debug
+)
+
+// ImageOptions provides options for the artifact builder
+type ImageOptions struct {
+	// FIXME: Tag should be []string but we don't support multiple tags yet
+	Tag           string             // image tag
+	Configuration Configuration      // build image for release or debug
+	Args          map[string]*string // additional builder specific args
```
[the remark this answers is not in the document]
gsquared94: I've renamed it to `BuildArgs`. 
The way I see this being used is in differentiating parameters that could influence build between the different skaffold actions like Dev, Build or Debug. Since skaffold.yaml doesn't have sections for these in case there are different settings required those can be put in this generic map. Whichever builder understands that setting would utilize it. This way since the setting affects the built image it also participates in cache hashcode calculation. WDYT?

pkg/skaffold/build/options.go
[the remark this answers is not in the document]
gsquared94: I've changed this to only have Dev and Debug since I also use this parameter in calculating cache hashcode. So having separate Build type would cause the image built by skaffold build not to be used for next skaffold dev

pkg/skaffold/build/options.go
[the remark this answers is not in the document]
gsquared94: A concrete example could be to push the buildpacks builder image digest to this map so that if that changes it invalidates the cache. That'd solve this TODO:
image

integration/testdata/debug/go/Dockerfile
```diff
@@ -1,9 +1,10 @@
 FROM golang:1.12 as builder
 
 COPY app.go .
-# Must use eval to handle GOGCFLAGS with spaces like `-gcflags='all=-N -l'`
-ARG GOGCFLAGS
-RUN eval go build "${GOGCFLAGS}" -o /app .
+ARG SKAFFOLD_GO_GCFLAGS
+ARG SKAFFOLD_LD_GCFLAGS
```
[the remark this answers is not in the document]
gsquared94: 😅 🤦🏽‍♂️

gsquared94:
Closing in favor of #4606

Issue #4470, opened 2020-07-14
status-check diagnostics can be puzzling


briandealwis:
Trying with the Cloud Code samples, I sometimes see output like the following:
```
 - deployment/nodejs-hello-world: waiting for rollout to finish: 0 of 1 updated replicas are available...
    - pod/nodejs-hello-world-d644cb489-7qvxm: container install-nodejs-support in error: 
 - deployment/nodejs-hello-world: waiting for rollout to finish: 0 of 1 updated replicas are available...
    - pod/nodejs-hello-world-d644cb489-7qvxm: container install-nodejs-support terminated with exit code 0
[nodejs-hello-world-d644cb489-7qvxm install-nodejs-support] Installing runtime debugging support files in /dbg
[nodejs-hello-world-d644cb489-7qvxm install-nodejs-support] Installation complete
 - deployment/nodejs-hello-world is ready.
```

A few thoughts on improvements:
  - The `container install-nodejs-support in error:` doesn't include the error, perhaps because we clip the empty log?  But then the log didn't have anything to show.  We should describe what we determined to show the container is in error.
    - Aha: the error is from this condition: `&ContainerStateWaiting{Reason:PodInitializing,Message:,}`

  - The `terminated with exit code 0` doesn't indicate whether this is expected or significant.  Perhaps `exited normally (exit code 0)`.

  - The `install-nodejs-support` container is an init-container and should be marked as such.

Perhaps when we're unable to extract messages, we should log the objects at debug or trace level?

PR #4471, opened 2020-07-14
Map container status PodInitializing to STATUSCHECK_SUCCESS

briandealwis:
Fixes: #4470

WDYT @tejal29 ?

### DIFF
--- a/pkg/diag/validator/pod.go
+++ b/pkg/diag/validator/pod.go
@@ -46,6 +46,7 @@ const (
 	imagePullBackOff    = "ImagePullBackOff"
 	errImagePullBackOff = "ErrImagePullBackOff"
 	containerCreating   = "ContainerCreating"
+	podInitializing     = "PodInitializing"
 	podKind             = "pod"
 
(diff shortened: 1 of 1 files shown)


pkg/diag/validator/pod.go
```diff
@@ -283,8 +284,11 @@ func (p *podStatus) String() string {
 }
 
 func extractErrorMessageFromWaitingContainerStatus(po *v1.Pod, c v1.ContainerStatus) (proto.StatusCode, []string, error) {
-	switch c.State.Waiting.Reason {
 	// Extract meaning full error out of container statuses.
+	switch c.State.Waiting.Reason {
+	case podInitializing:
+		// container is waiting to run
+		return proto.StatusCode_STATUSCHECK_SUCCESS, nil, nil
```
tejal29, line 291: Reading this -> https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior
I am wondering if we should wait for all init containers to run successfully and pod to be running within in `statusCheckDeadline` or return success here like you have.

PR #4508, opened 2020-07-21
Disable update check and survey prompt in non-interactive mode

### LINKED ISSUE #4506: Hide survey link for embedded use cases
Related to #4500 - Skaffold should not show the survey link when run from a non-interactive terminal.

For a related implementation note, minikube has a `--interactive=false` flag that IDE's set to avoid prompts and other features that are not relevant for embedded use cases. You could also support an environment variable.

dgageot:
Fixes #4506

### DIFF
--- a/cmd/skaffold/app/cmd/cmd.go
+++ b/cmd/skaffold/app/cmd/cmd.go
@@ -41,9 +41,10 @@ import (
 var (
 	opts              config.SkaffoldOptions
 	v                 string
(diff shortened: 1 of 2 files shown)


dgageot:
Let me try to add a test

dgageot:
In fact, it's a part of the code that's hard to test and the test wouldn't really add much value. Let's merge like this

PR #4510, opened 2020-07-21
SKAFFOLD_UPDATE_CHECK should also be a global flag

dgageot:
This specific env var was introduced before we mirrored every command line flag with env vars.
Make sure it's now available either way and properly documented.


dgageot:
@briandealwis could you take a look at this one. It doesn't really change anything. Just that it adds the documentation for a env var that we support.

PR #4524, opened 2020-07-21
Add suppress-logs cli flag and suppressLogs field to schema

MarlonGamez:
Fixes: #4512  
**Related**: #4514, #4515, #4516

**Description**

This PR simply adds a --suppress-logs flag and a suppressLogs field to the skaffold schema. They currently have no functionality, but this will be addressed by #4514, #4515, and #4516.

The 3 previously mentioned issues will need to be addressed in order to give functionality to these options.


MarlonGamez:
Realized we can just expand/repurpose the --quiet flag that currently exists for build. No need to introduce a new flag/option if not necessary.

PR #4527, opened 2020-07-22
Update buildpacks.md to reflect change in schema

ojhughes:
**Description**
change YAML example in buildpack docs from `buildpacks:` to `buildpack:`


ojhughes:
@googlebot I signed it!

PR #4530, opened 2020-07-22
Add suppress-logs flag

MarlonGamez:
Fixes: #4512
Related: #4514, #4515, #4516

**Description**
This PR simply adds a --suppress-logs flag. It currently has no functionality, but this will be addressed by #4514, #4515, and #4516.

**Follow-up Work**
The 3 previously mentioned issues will need to be addressed in order to provide functionality.

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -370,7 +370,7 @@ var FlagRegistry = []Flag{
 	},
 	{
 		Name:          "suppress-logs",
-		Usage:         "Suppress logs for specified stages in pipeline (build, deploy, status-check)",
+		Usage:         "Suppress logs for specified stages in pipeline (build, deploy, status-check, none, all)",
 		Value:         &opts.SuppressLogs,
 		DefValue:      []string{},
 		FlagAddMethod: "StringSliceVar",
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -368,6 +368,14 @@ var FlagRegistry = []Flag{
 		FlagAddMethod: "BoolVar",
 		DefinedOn:     []string{"render"},
 	},
+	{
+		Name:          "suppress-logs",
+		Usage:         "Suppress logs for specified stages in pipeline (build, deploy, status-check)",
+		Value:         &opts.SuppressLogs,
+		DefValue:      []string{},
+		FlagAddMethod: "StringSliceVar",
+		DefinedOn:     []string{"dev", "run", "debug", "build", "deploy"},
+	},
 }
 
 func (fl *Flag) flag() *pflag.Flag {
--- a/docs/content/en/docs/references/cli/_index.md
+++ b/docs/content/en/docs/references/cli/_index.md
@@ -148,6 +148,7 @@ Options:
       --rpc-http-port=50052: tcp port to expose event REST API over HTTP
       --rpc-port=50051: tcp port to expose event API
       --skip-tests=false: Whether to skip the tests after building
+      --suppress-logs=[]: Suppress logs for specified stages in pipeline (build, deploy, status-check)
   -t, --tag='': The optional custom tag to use for images which overrides the current Tagger configuration
       --toot=false: Emit a terminal beep after the deploy is complete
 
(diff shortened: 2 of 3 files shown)


gsquared94:
I saw you mentioned somewhere that we can't repurpose `--quiet`. Not exactly sure why though.

MarlonGamez:
> I saw you mentioned somewhere that we can't repurpose `--quiet`. Not exactly sure why though.

I think our goal with this flag is to be able to use it to specify which stages we want to suppress the output of. For example, I would run `skaffold run --suppress-logs=deploy` or maybe `skaffold dev --suprress-logs=build`. 

Currently, `--quiet` exists just as a boolean flag for `skaffold build`, so if we repurpose it to function the way I mentioned above, we risk breaking users who are currently using `skaffold build --quiet`. They'd have to change their commands to use `skaffold build --quiet=build`. I talked to @nkubala about this and we decided it's probably best to introduce a new flag.

gsquared94:
Any place I can read up on what this is going to look like? Also saw that you mentioned that `quiet` flag can't be

> > I saw you mentioned somewhere that we can't repurpose `--quiet`. Not exactly sure why though.
> 
> I think our goal with this flag is to be able to use it to specify which stages we want to suppress the output of. For example, I would run `skaffold run --suppress-logs=deploy` or maybe `skaffold dev --suppress-logs=build,deploy`.
> 
> Currently, `--quiet` exists just as a boolean flag for `skaffold build`, so if we repurpose it to function the way I mentioned above, we risk breaking users who are currently using `skaffold build --quiet`. They'd have to change their commands to use `skaffold build --quiet=build`. I talked to @nkubala about this and we decided it's probably best to introduce a new flag.

If we give special meaning to `--quiet` with  to fallback to existing behavior of 

> > I saw you mentioned somewhere that we can't repurpose `--quiet`. Not exactly sure why though.
> 
> I think our goal with this flag is to be able to use it to specify which stages we want to suppress the output of. For example, I would run `skaffold run --suppress-logs=deploy` or maybe `skaffold dev --suppress-logs=build,deploy`.
> 
> Currently, `--quiet` exists just as a boolean flag for `skaffold build`, so if we repurpose it to function the way I mentioned above, we risk breaking users who are currently using `skaffold build --quiet`. They'd have to change their commands to use `skaffold build --quiet=build`. I talked to @nkubala about this and we decided it's probably best to introduce a new flag.

Ok. LGTM then.

cmd/skaffold/app/cmd/flags.go
```diff
@@ -368,6 +368,14 @@ var FlagRegistry = []Flag{
 		FlagAddMethod: "BoolVar",
 		DefinedOn:     []string{"render"},
 	},
+	{
+		Name:          "suppress-logs",
+		Usage:         "Suppress logs for specified stages in pipeline (build, deploy, status-check)",
```
dgageot, line 373: Do we want to suppress them or shorten them?
Could it just be a "--verbose-logs" that's on by default? I feel it's a bit overkill to be able to change the logs for each stage independently.

MarlonGamez: @dgageot I do think that shorten is probably a better word to use over suppress. As for being able to control each stage independently, do you think that it's overkill from a user's perspective or from an implementation perspective? If from a user's perspective, I think we could support them using `=none` or `=all` in order to simplify things. This would allow them to not have to deal with per stage settings

dgageot: I think it's overkill for the users. But I'm open to other opinions :-)

MarlonGamez: @GoogleContainerTools/skaffold-team anyone have thoughts on this?

briandealwis: We talked about this in the logging breakout yesterday and it does seem useful to be able to suppress logs from different phases when the phase is successful.  This would be helpful for the Cloud Run emulation situation where Skaffold is being used as the mechanism for local building and deploying, but ideally shouldn't be shown to the user.

briandealwis: In that situation, we might want to show the build logs but hide the deploy and status checks

dgageot: Let's do that then

PR #4531, opened 2020-07-22
Wait for pending deletions to complete before a deploy

### LINKED ISSUE #4519: Skaffold sometimes doesn't deploy all services
### Expected behavior
All services should deploy all the time

### Actual behavior
A fraction of the time, a service gets lost and doesn't deploy

### Information

- Skaffold version: version v1.12.0
- Operating system: Mac 10.15.5
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta15
kind: Config
build:
  tagPolicy:
    sha256: {}
  # defines where to find the code at build time and where to push the resulting image
  artifacts:
  - image: dotnet-guestbook-backend
    context: src/backend
  - image: dotnet-guestbook-frontend
    context: src/frontend
# defines the Kubernetes manifests to deploy on each run
deploy:
  kubectl:
    manifests:
    - ./kubernetes-manifests/*.service.yaml
    - ./kubernetes-manifests/*.deployment.yaml
profiles:
# use the cloudbuild profile to build images using Google Cloud Build
- name: cloudbuild
  build:
    googleCloudBuild: {}
```

1. Use either the Cloud Code VS Code or IntelliJ extension
2. Create a new application using the Java guestbook sample
3. Select run on Kubernetes
4. Observe that about 5-10% of the time, the frontend service doesn't deploy

#### VS Code
Screen Shot 2020-07-21 at 12 39 54 AM

#### IntelliJ
Screen Shot 2020-07-21 at 1 31 32 PM

dgageot:
When Skaffold tries to deploy resources, it now checks that none of those resources are marked for deletion. In case, there are at least one, it'll wait for the deletion to complete.
If it doesn't do that, it's going to update resources that are marked for deletion and which are going to disappear in the background.

Fixes #4519

### DIFF
--- a/pkg/skaffold/deploy/kubectl.go
+++ b/pkg/skaffold/deploy/kubectl.go
@@ -53,6 +53,7 @@ type KubectlDeployer struct {
 	insecureRegistries map[string]bool
 	labels             map[string]string
 	skipRender         bool
+	waitForDeletions   bool
 }
 
 // NewKubectlDeployer returns a new KubectlDeployer for a DeployConfig filled
@@ -70,6 +71,7 @@ func NewKubectlDeployer(runCtx *runcontext.RunContext, labels map[string]string)
 		},
 		insecureRegistries: runCtx.InsecureRegistries,
 		skipRender:         runCtx.Opts.SkipRender,
+		waitForDeletions:   runCtx.Opts.WaitForDeletions,
 		labels:             labels,
 	}
 }
@@ -100,6 +102,13 @@ func (k *KubectlDeployer) Deploy(ctx context.Context, out io.Writer, builds []bu
 	if err != nil {
 		event.DeployInfoEvent(fmt.Errorf("could not fetch deployed resource namespace. "+
 			"This might cause port-forward and deploy health-check to fail: %w", err))
+	}
+
+	if k.waitForDeletions {
+		if err := k.kubectl.WaitForDeletions(ctx, textio.NewPrefixWriter(out, " - "), manifests); err != nil {
+			event.DeployFailed(err)
+			return NewDeployErrorResult(err)
+		}
 	}
 
 	if err := k.kubectl.Apply(ctx, textio.NewPrefixWriter(out, " - "), manifests); err != nil {
--- a/pkg/skaffold/deploy/kubectl/cli.go
+++ b/pkg/skaffold/deploy/kubectl/cli.go
@@ -18,14 +18,25 @@ package kubectl
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
+	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
 	pkgkubectl "github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubectl"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 )
+
+var WaitDeletion = struct {
+	MaxRetry int
+	Delay    time.Duration
+}{
+	MaxRetry: 30,
+	Delay:    2 * time.Second,
+}
 
 // CLI holds parameters to run kubectl.
 type CLI struct {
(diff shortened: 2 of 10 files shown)


pkg/skaffold/deploy/kubectl/cli.go
```diff
@@ -73,6 +75,50 @@ func (c *CLI) Apply(ctx context.Context, out io.Writer, manifests ManifestList)
...
+	if err := json.Unmarshal(out.Bytes(), &result); err != nil {
+		return nil, err
+	}
+
+	var list []string
+
+	// Find resources with `metadata.deletionTimestamp` field.
+	for _, r := range result.Items {
+		if r.Metadata.DeletionTimestamp != "" {
+			list = append(list, r.Metadata.Name)
```
briandealwis, line 115: I think it's worth appending the group/kind too.  I think there's some code in the `pkg/diag` that can be used here.

pkg/skaffold/deploy/kubectl/cli.go
```diff
@@ -73,6 +75,50 @@ func (c *CLI) Apply(ctx context.Context, out io.Writer, manifests ManifestList)
...
+}
+
+type metadata struct {
+	Name              string `json:"name"`
+	DeletionTimestamp string `json:"deletionTimestamp"`
+}
+
+// GetPendingDeletions uses `kubectl get` to list resources that are marked for deletion.
+func (c *CLI) GetPendingDeletions(ctx context.Context, manifests ManifestList) ([]string, error) {
+	args := []string{"-f", "-", "--ignore-not-found=true", "-ojson"}
```
briandealwis, line 93: I wonder if kubectl can be tweaked to do the filtering...

pkg/skaffold/deploy/kubectl.go
```diff
@@ -109,6 +110,28 @@ func (k *KubectlDeployer) Deploy(ctx context.Context, out io.Writer, builds []bu
 			"This might cause port-forward and deploy health-check to fail: %w", err))
 	}
 
+	// Wait for ongoing deletions to complete.
+	previousList := ""
+	for {
```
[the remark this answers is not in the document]
dgageot: We should be able to use `kubectl wait for=delete` (or something similar) instead of this loop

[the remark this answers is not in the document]
dgageot: `kubectl wait for=delete` doesn't work as expected. I added a max number of retries

pkg/skaffold/deploy/kubectl/cli.go
dgageot: Only possible with `-ojsonPath` but then it doesn't output json anymore...

pkg/skaffold/deploy/kubectl/cli.go
```diff
@@ -73,6 +84,64 @@ func (c *CLI) Apply(ctx context.Context, out io.Writer, manifests ManifestList)
...
+			return nil
+		}
+
+		list := `"` + strings.Join(marked, `", "`) + `"`
+		logrus.Debugln("Resources are marked for deletion:", list)
+		if list != previousList {
+			if len(marked) == 1 {
+				fmt.Fprintf(out, "%s is marked for deletion, waiting for completion\n", list)
+			} else {
+				fmt.Fprintf(out, "Those resources are marked for deletion, waiting for completion: %s\n", list)
```
briandealwis, line 134:
```suggestion
				fmt.Fprintf(out, "%d resources are marked for deletion, waiting for completion: %s\n", len(list), list)
```

pkg/skaffold/deploy/kubectl/cli.go
```diff
@@ -73,6 +84,64 @@ func (c *CLI) Apply(ctx context.Context, out io.Writer, manifests ManifestList)
...
+			}
+			previousList = list
+		}
+
+		time.Sleep(WaitDeletion.Delay)
+	}
+
+	return fmt.Errorf("those resources need to complete their deletion before a new deployment: %s", previousList)
```
briandealwis, line 142:
```suggestion
	return fmt.Errorf("%d resources failed to complete their deletion before a new deployment: %s", len(previousList), previousList)
```

pkg/skaffold/deploy/kubectl/cli.go
```diff
@@ -18,15 +18,26 @@ package kubectl
...
 	pkgkubectl "github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubectl"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 )
 
+var WaitDeletion = struct {
+	MaxRetry int
+	Delay    time.Duration
+}{
+	MaxRetry: 30,
+	Delay:    2 * time.Second,
```
nkubala, line 38: is this something a user would ever want to configure? `maxDeletionWaitTime` or something?

dgageot: probably, yes. I can add that in a second PR

pkg/skaffold/deploy/kubectl/cli.go
```diff
@@ -73,6 +84,67 @@ func (c *CLI) Apply(ctx context.Context, out io.Writer, manifests ManifestList)
...
+		} `json:"metadata"`
+	} `json:"items"`
+}
+
+// WaitForDeletions waits for resource marked for deletion to complete their deletion.
+func (c *CLI) WaitForDeletions(ctx context.Context, out io.Writer, manifests ManifestList) error {
+	previousList := ""
+	previousCount := 0
+
+	for try := 0; try < WaitDeletion.MaxRetry; try++ {
```
briandealwis, line 101: Would it make sense to instead use a child context with a timeout?  (I'm really just concerned about infinite loops waiting for an object that will never die.)

PR #4532, opened 2020-07-22
Docs to accommodate the deprecation of envTemplate's use of {{.IMAGE_NAME}}

felixtran39:
**Description**
Changing the docs to accommodate the deprecation of envTemplate's use of {{.IMAGE_NAME}}. The link to that PR is: https://github.com/GoogleContainerTools/skaffold/pull/4533

Depends on PR #4533

### DIFF
--- a/docs/content/en/docs/pipeline-stages/taggers.md
+++ b/docs/content/en/docs/pipeline-stages/taggers.md
@@ -93,16 +93,15 @@ process.
 
 The following `build` section, for example, instructs Skaffold to build a
 Docker image `gcr.io/k8s-skaffold/example` with the `envTemplate`
-tag policy. The tag template is `{{.IMAGE_NAME}}:{{.FOO}}`; when Skaffold
+tag policy. The tag template is `{{.FOO}}`; when Skaffold
 finishes building the image, it will check the list of available environment
 variables in the system for the variable `FOO`, and use its value to tag the
 image.
 
 {{< alert >}}
-<b>Note</b><br>
+<b>Deprecated</b><br>
 
-`IMAGE_NAME` is a built-in variable whose value is the `imageName` field in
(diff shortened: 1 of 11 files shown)


yuwenma:
LGTM!

docs/content/en/docs/pipeline-stages/taggers.md
```diff
@@ -93,16 +93,15 @@ process.
...
 variables in the system for the variable `FOO`, and use its value to tag the
 image.
 
 {{< alert >}}
-<b>Note</b><br>
+<b>Deprecated</b><br>
 
-`IMAGE_NAME` is a built-in variable whose value is the `imageName` field in
-the `artifacts` part of the `build` section.
+The use of `IMAGE_NAME` as a built-in variable whose value is the `imageName` field in the `artifacts` part of the `build` section has been deprecated. Please use the envTemplate to express solely the tag value for the image.
```
briandealwis, line 104: @nkubala should we say when this was deprecated and when it will be removed?

(We can do this as a follow-up)

felixtran39: > Looks like TestBuild/envTemplate_tagger need to be fixed up?

I believe this is due to the fact that the code logic for the deprecation has not been merged yet. I removed "{{.IMAGE_NAME}}" from the template used in TestBuild/envTemplate_tagger. I can undo this change and add it to the other PR.

felixtran39: > Looks like TestBuild/envTemplate_tagger need to be fixed up?

I believe this error is because this test case is dependent on the code logic from the other PR. I can undo this change in this PR and add it to the other one. Also, I removed `t.Parallel()` from one of the tests in git_commit_test.go since that seemed to occasionally throw race detection errors.

Issue #4500, opened 2020-07-20
Skaffold should not output version update messages when run from gcloud


briandealwis:
Users using gcloud shouldn't be advised to download a new copy from our website.

briandealwis:
Removing from chill: message should be suppressed in the IDEs with #4508.

PR #4534, opened 2020-07-23
Do not report version updates if run from Google Cloud SDK

briandealwis:
Fixes: #4500
**Related**: #4506

**Description**

Do not output new-version-found messages when installed via the Google Cloud SDK since the user should be getting updates through the Cloud SDK.  We can still prompt for surveys.

No new-version updates when run from Cloud SDK.

### DIFF
--- a/cmd/skaffold/app/cmd/cmd.go
+++ b/cmd/skaffold/app/cmd/cmd.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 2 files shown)


tejal29:
LGTM. Can you please add some tests.

tejal29:
This looks good to me. 
We should look if there a way to set the update check config to be false when skaffold when installed via gcloud? that ways, we can avoid this PR.
https://skaffold.dev/docs/references/privacy/

cmd/skaffold/app/cmd/cmd.go
```diff
@@ -196,7 +197,7 @@ func updateCheck(ch chan string, configfile string) error {
 	if err != nil {
 		return fmt.Errorf("get latest and current Skaffold version: %w", err)
 	}
-	if latest.GT(current) {
```
dgageot, line 199: Would it make sense to move that test to `update.IsUpdateCheckEnabled()`?

dgageot: OK, I read your comment on the chat. You want to keep the ping

briandealwis:
I don't much like this approach and it needs a rethink.

PR #4536, opened 2020-07-23
Failing to override value should be fatal

dgageot:
Makes it easier to understand why a test is failing when a `t.Override()` is not valid


dgageot:
Self merging this small fix to the test framework

PR #4546, opened 2020-07-23
Terminate could be called when there’s no handle yet

dgageot:


pkg/skaffold/kubectl/exec_windows.go
```diff
@@ -42,54 +45,58 @@ func CommandContext(ctx context.Context, name string, arg ...string) *Cmd {
 }
 
 // Start starts the specified command in a job object but does not wait for it to complete
-func (c *Cmd) Start() (err error) {
-	var handle windows.Handle
-	handle, err = windows.CreateJobObject(nil, nil)
+func (c *Cmd) Start() error {
+	handle, err := windows.CreateJobObject(nil, nil)
```
briandealwis, line 49: Is this `handle` a lightweight object or does it consume system resources?  Does it need to be cleaned up with `windows.CloseHandle()` before we return with an error?

pkg/skaffold/kubectl/exec_windows.go
```diff
@@ -42,54 +45,58 @@ func CommandContext(ctx context.Context, name string, arg ...string) *Cmd {
...
+	}()
+
+	return nil
 }
 
 // Terminate closes the job object handle which kills all connected processes
 func (c *Cmd) Terminate() error {
-	return windows.CloseHandle(c.handle)
+	c.handleLock.Lock()
+	handle := c.handle
```
briandealwis, line 94: It looks like it's possible for `handle` to be unset.  I suppose `windows.CloseHandle()` will error in that case?

PR #4562, opened 2020-07-24
Wait for the process to be done

dgageot:


dgageot:
1 successful run

pkg/skaffold/kubectl/exec_windows.go
```diff
@@ -91,5 +91,9 @@ func (c *Cmd) Run() error {
 
 // Terminate closes the job object handle which kills all connected processes
 func (c *Cmd) Terminate() error {
-	return windows.CloseHandle(c.handle)
+	if err := windows.CloseHandle(c.handle); err != nil {
+		return err
+	}
+	c.Wait()
```
gsquared94, line 97: IMO `exec_windows.go` methods should match behavior with `exec.go`, so `Terminate()` there should also wait

Issue #4554, opened 2020-07-23
Global config should only support `local-context` on a per kube-context basis


briandealwis:
@russwolf  somehow had his ~/.skaffold/config sert to make all cluster be non-local which led to very puzzling errors when attempting to develop against local clusters. He’s unsure how this setting was configured. And this particular setting seems of questionable value as a blanket setting, and should instead be set on a per-kube-context basis.

We should revisit the actual global config definition as it allows invalid and questionable configuration combinations.  For example, it allows setting insecure registries and debug helper locations on a per-kube-context basis.

etanshaul:
for the IDEs, can Skaffold surface a better error message in this case (where the cluster is configured as non-local, but the user has no default-registry set)?

PR #4566, opened 2020-07-24
Log when values are taken from global config file

briandealwis:
Log when values are resolved from the Skaffold global config file (typically ~/.skaffold/config).  At least simplifies diagnosing situations like the motivation for #4554.

```
$ skaffold dev -v info
WARN[0000] starting gRPC server on port 50053. (50051 is already in use) 
WARN[0000] starting gRPC HTTP server on port 50054. (50052 is already in use) 
INFO[0000] Skaffold &{Version: ConfigVersion:skaffold/v2beta6 GitVersion: GitCommit: GitTreeState: BuildDate: GoVersion:go1.14.3 Compiler:gc Platform:darwin/amd64} 
INFO[0000] Loaded Skaffold defaults from "/Users/bdealwis/.skaffold/config"
INFO[0000] Using kubectl context: docker-desktop        
INFO[0000] Using local-cluster=false from config        
...
```

### DIFF
--- a/pkg/skaffold/config/util.go
+++ b/pkg/skaffold/config/util.go
@@ -73,6 +73,9 @@ func readConfigFileCached(filename string) (*GlobalConfig, error) {
 			return
 		}
 		configFile, configFileErr = ReadConfigFileNoCache(filenameOrDefault)
+		if configFileErr == nil {
+			logrus.Infof("Loaded Skaffold defaults from %q", filenameOrDefault)
+		}
 	})
 	return configFile, configFileErr
 }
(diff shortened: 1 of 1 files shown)


pkg/skaffold/config/util.go
```diff
@@ -73,6 +73,9 @@ func readConfigFileCached(filename string) (*GlobalConfig, error) {
 			return
 		}
 		configFile, configFileErr = ReadConfigFileNoCache(filenameOrDefault)
+		if configFileErr == nil {
+			logrus.Infof("Loaded Skaffold defaults from %q", filenameOrDefault)
```
tejal29, line 77: shd we move this to debug? I usually set my `default-repo` etc in global config when demoing so that the command line is as simple as `skaffold dev` on examples. 
This will add more logs statements.

briandealwis: Skaffold's default log-level is warning, so this will only appear if run with `-v info`.

PR #4572, opened 2020-07-27
Improve wait for deletions

dgageot:
+ Make sure the check doesn't block for more than a given duration
+ Extract some duplicated code to a constructor
+ Add 2 command line flags to configure the check's max duration and delay between two checks

### DIFF
--- a/pkg/skaffold/deploy/kubectl/cli.go
+++ b/pkg/skaffold/deploy/kubectl/cli.go
@@ -31,11 +31,11 @@ import (
 )
 
 var WaitDeletion = struct {
-	MaxRetry int
-	Delay    time.Duration
+	Max   time.Duration
+	Delay time.Duration
 }{
-	MaxRetry: 30,
-	Delay:    2 * time.Second,
+	Max:   60 * time.Second,
+	Delay: 2 * time.Second,
 }
 
 // CLI holds parameters to run kubectl.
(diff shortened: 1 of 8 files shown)


pkg/skaffold/deploy/kubectl/cli.go
```diff
@@ -26,25 +26,29 @@ import (
...
 	Flags latest.KubectlFlags
 
-	ForceDeploy   bool
-	previousApply ManifestList
+	forceDeploy      bool
+	waitForDeletions config.WaitForDeletions
+	previousApply    ManifestList
+}
+
+func NewCLI(runCtx *runcontext.RunContext, flags latest.KubectlFlags) CLI {
```
briandealwis, line 45: I'd prefer that this not take the RunContext and instead pass in the configuration values explicitly: this CLI is a utility library and using the RunContext ties it to the Skaffold configuration.  We're now mixing policy with the mechanism.

dgageot: Do you mind if I do it in another PR? I would then do the same for kubectl.NewFromRunContext

briandealwis: Oops, I thought this was in `pkg/skaffold/kubectl`, not `pkg/skaffold/deploy/kubectl`.  I retract my comment.

PR #4582, opened 2020-07-29
Remove duplicated replica count.

dgageot:
This leads to strange behaviour such as not being able to apply a label on resources and not being able to check the deployment's status.


dgageot:
Self merging this small fix

PR #4584, opened 2020-07-29
Fail when k8s client can’t be obtained

dgageot:


dgageot:
Self-merging this small fix

PR #4586, opened 2020-07-29
Log pods for which we deployed the owner

dgageot:
**This PR is an experiment. Many things are broken.**

This is how application logs work in Skaffold:
+ A logger watches for all the pod event in a set of namespaces
+ For each pod event, we look for the images used in the pod.
+ If any container in that pod uses an image that Skaffold built, we output the logs for that container

This is tricky because:
+ We might end up logging a pod that was deployed beforehand and uses the same image tags.
+ We miss the pods that Skaffold deployed but that use pre-built images

We tried to rely on the runID label to solve that issue but it also has issues:
+ The runID mechanism break idempotency: a `skaffold run` just after a `skaffold run` should do nothing. With the runID, every object is updated.
+ If we switch from a runID label to a runID annotation, it will not cascade down to the pods.

This PR introduces a new mechanism:
+ kubectl/kustomize deployer, returns the list of resources it applied, with their uid. We keep that.
+ helm deployer used to add a label to the resources it deployed. Instead, we'll collect all the uids.
+ the pod watcher, when it sees a pod event looks for the top level parent and checks that this is in the set of uids we deployed.


pkg/skaffold/deploy/labels.go
```diff
@@ -47,26 +48,27 @@ const (
...
 
 	client, err := kubernetes.Client()
 	if err != nil {
-		return fmt.Errorf("error getting Kubernetes client: %w", err)
+		return nil, fmt.Errorf("error getting Kubernetes client: %w", err)
 	}
 
+	var resources kubectl.Resources
+
 	for _, res := range results {
```
gsquared94, line 65: Since this loops `[]Artifacts` which only contains skaffold-built images, this still won't select containers for pre-built images which belong to a different Deployment altogether; it only improves upon init containers and multi-container pods, right?

dgageot: exactly. Not yet.

PR #4592, opened 2020-07-30
Use Dynamic Client for Top level owner

dgageot:
This makes it possible to support any kind/version of owner unlike the current code that has to specify a fixed list.

### DIFF
--- a/pkg/skaffold/kubernetes/owner.go
+++ b/pkg/skaffold/kubernetes/owner.go
@@ -20,47 +20,92 @@ import (
 	"fmt"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
 )
 
 // TopLevelOwnerKey returns a key associated with the top level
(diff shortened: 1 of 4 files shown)


pkg/skaffold/kubernetes/owner.go
```diff
@@ -19,50 +19,93 @@ package kubernetes
...
+		Object: obj,
+		kind:   kind,
 	}
+
+	topLevel, err := topLevel(child, discovery, dynClient)
+	if err != nil {
+		return "", err
+	}
+
+	id := fmt.Sprintf("%s-%s", topLevel.GetKind(), topLevel.GetName())
```
tejal29, line 52: is this `Kind` and `Name` enough to uniquely identify a object. I am thinking a case where there are workloads across namespace.

dgageot: this is a copy of what was there before. It's the part of a larger key that identifies one port forwarding entry

briandealwis: Each object is uniquely identified by its namespace, kind, and metadata.name ([source](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/)).  We should include the namespace here.

PR #4598, opened 2020-07-30
Introduce Config interfaces

dgageot:
This is an additional non-dangerous step towards getting rid of RunContext as much as possible.

It introduces a `Config` interface for each constructor that used to depend on a RunContext. This acts as a "facade" on the RunContext and can only access information by their getter, making them more read-only. This break some hard coupling that we have which is also nice.


gsquared94 (approved):
Looks great to me!

dgageot:
Let me remove those wip commits

gsquared94:
> Let me remove those wip commits

BTW we only squash merge now to not have to take care of these commit cleanups anymore.

PR #4604, opened 2020-07-31
Make the dev loop interruptible.

### LINKED ISSUE #916: skaffold dev: not cancelling the in-flight build on change

### LINKED ISSUE #3908: status check blocks rebuild/redeploy
I'm seeing this on microservices example: when both status check are pending, rebuilds and redeploys don't work on file changes.

### LINKED ISSUE #4225: dev should not error out on introduced error to skaffold.yaml
Follow-on to #4158: skaffold dev shouldn't error out on a typo to `skaffold.yaml`.  I made a mistake while uncommenting and forgot to uncomment a line to produce an invalid yaml:
```
Pruning images...
WARN[0217] builder cleanup: pruning images: could not remove image after 5 retries 
parsing skaffold config: unable to parse config: yaml: unmarshal errors:
  line 7: cannot unmarshal !!seq into v2beta3.KubectlDeploy
```

and Skaffold then exited.

dgageot:
Wether it’s during the build, the deploy or the status check, the dev loop can now be interrupted on a file change.

Fix #3908
Fix #916

If it’s the `skaffold.yaml` that’s modified, instead of restarting the dev loop right away, we first validate that the config is valid.

Fix #4225

### DIFF
--- a/pkg/skaffold/runner/listen.go
+++ b/pkg/skaffold/runner/listen.go
@@ -23,13 +23,14 @@ import (
 	"io"
 
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/filemon"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/trigger"
 )
 
 type Listener interface {
-	WatchForChanges(context.Context, io.Writer, func() error) error
+	WatchForChanges(context.Context, io.Writer, func(ctx context.Context) error) error
 	LogWatchToUser(io.Writer)
 }
 
@@ -45,10 +46,11 @@ func (l *SkaffoldListener) LogWatchToUser(out io.Writer) {
 
 // WatchForChanges listens to a trigger, and when one is received, computes file changes and
 // conditionally runs the dev loop.
-func (l *SkaffoldListener) WatchForChanges(ctx context.Context, out io.Writer, devLoop func() error) error {
-	ctxTrigger, cancelTrigger := context.WithCancel(ctx)
-	defer cancelTrigger()
-	trigger, err := trigger.StartTrigger(ctxTrigger, l.Trigger)
+func (l *SkaffoldListener) WatchForChanges(ctx context.Context, out io.Writer, devLoop func(ctx context.Context) error) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	trigger, err := trigger.StartTrigger(ctx, l.Trigger)
 	if err != nil {
 		return fmt.Errorf("unable to start trigger: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


dgageot:
Converting it to draft since there's obviously something not working well.

pkg/skaffold/runner/listen.go
```diff
@@ -60,29 +62,45 @@ func (l *SkaffoldListener) WatchForChanges(ctx context.Context, out io.Writer, d
...
 	for {
 		select {
-		case <-ctx.Done():
+		case <-egCtx.Done():
+			cancelDev()
 			return nil
 		case <-l.intentChan:
-			if err := l.do(devLoop); err != nil {
+			// TODO(dgageot): make sure a dev cycle triggered by an intent can be interrupted.
+			if err := l.do(devCtx, devLoop); err != nil {
```
briandealwis, line 74: Why are you calling this directly instead of through `eg.Go()` like you do below?

pkg/skaffold/runner/listen.go
```diff
@@ -60,29 +62,45 @@ func (l *SkaffoldListener) WatchForChanges(ctx context.Context, out io.Writer, d
 
 	l.LogWatchToUser(out)
 
+	eg, egCtx := errgroup.WithContext(ctx)
+	devCtx, cancelDev := context.WithCancel(egCtx)
 	for {
 		select {
-		case <-ctx.Done():
+		case <-egCtx.Done():
```
briandealwis, line 69: Since you don't actually call `eg.Go()` on the first pass, won't you immediately hit this case, call `cancelDev()`, and then return nil?

pkg/skaffold/runner/listen.go
dgageot: If I remember correctly, it didn't work just by using the same code than under `<-trigger`. So I decided to just add a TODO for later.

pkg/skaffold/runner/listen.go
gsquared94: I can reply since I've taken over this PR.

The `context` is valid right from the initialization of the `eg` struct so waiting on the `Done` channel should work fine. `eg.Go` actually sets up blocking `eg.Wait`. So calling `eg.Wait` before `eg.Go` would return immediately and close this `Done` channel but that doesn't happen here. 
You can look at the source [here](https://github.com/golang/sync/blob/43a5402ce75a/errgroup/errgroup.go#L32)

gsquared94:
Closing this for now. Will create a new one when I can get to it.

PR #4605, opened 2020-07-31
Introduce a StatusChecker type

dgageot:


tejal29:
not needed due to #4690.
you can directly implement the TODO.

Issue #3977, opened 2020-04-18
`debug` should determine runtime from source


briandealwis:
We should be able to walk an artifact's source to determine the runtime type.  This would remove the need for GO* environment variables and more.

There may be a chance of false positives, such as projects that use scripts as part of their builds. The explicit runtime mechanism #2203 might handle that situation.

This might be best developed as a distinct library that could also be used as a command-line tool.  It could be useful for IDEs for detecting different project roots.  Ideally we'd be able to leverage (or extract this from) `skaffold init`.

PR #4606, opened 2020-07-31
Build debuggable containers

### LINKED ISSUE #4370: Skaffold helps build debuggable containers

gsquared94:
Fixes #4370 
Add default build args for docker and buildpacks when building for Debug vs Dev. Also handle skaffold image cache updates.

Redo #4467. No need to extract out `ImageOptions` since #4598 introduces a `Config` interface anyways.

Dockerfile can expose `ARG SKAFFOLD_GO_GCFLAGS` and `ARG SKAFFOLD_GO_LDFLAGS` and expect Skaffold to supply meaningful values.
Buildpacks auto set `GOOGLE_GCFLAGS` and `GOOGLE_LDFLAGS` build options.


pkg/skaffold/build/cache/hash_test.go
```diff
@@ -39,46 +40,43 @@ var mockCacheHasher = func(s string) (string, error) {
...
-		devMode      bool
+		mode         config.RunMode
 		expected     string
 	}{
 		{
 			description:  "hash for artifact",
 			dependencies: []string{"a", "b"},
 			artifact:     &latest.Artifact{},
-			expected:     "1caa15f7ce87536bddbac30a39768e8e3b212bf591f9b64926fa50c40b614c66",
+			expected:     "d99ab295a682897269b4db0fe7c136ea1ecd542150fa224ee03155b4e3e995d9",
```
tejal29, line 62: Can you explain why the hash for the artifact changed?

gsquared94: That's because I modified the hash calculation function:
image

tejal29:
> Dockerfile can expose ARG SKAFFOLD_GO_GCFLAGS and ARG SKAFFOLD_GO_LDFLAGS and expect Skaffold to supply meaningful values.

Why is this change required? Setting "GO_GC_FLAGS" and "GO_LDFLAGS" in dockerfile is a common practice. Changing this to "SKAFFOLD_GO_GCFLAGS" might break existing user's dockerfile

pkg/skaffold/build/cache/hash_test.go
tejal29: Thanks. This would mean, if i run `skaffold dev` and then `skaffold debug` the artifact hash will change and a skaffold will the image again. 

This is not optimal.

gsquared94: you're right. I've messed this part up a bit. I'll close the PR temporarily until I rethink this.

briandealwis: This is what we want *if* the options are different (e.g., the Go GCFLAGS disable optimizations etc that we would want to keep for `build`).  But as is in this PR, we will do this for all images.

I'm thinking this may need to be tied into #3977.

pkg/skaffold/config/options.go
```diff
@@ -94,18 +94,32 @@ type SkaffoldOptions struct {
...
 }
 
-func (opts *SkaffoldOptions) IsDevMode() bool {
-	return opts.Command == "dev"
-}
-
-func (opts *SkaffoldOptions) IsDebugMode() bool {
-	return opts.Command == "debug"
+func (opts *SkaffoldOptions) Mode() RunMode {
+	return RunMode(opts.Command)
```
briandealwis, line 122: What should happen if we have a command with no corresponding mode?

gsquared94: Are you saying that if we introduce `skaffold foo` then what would happen? I think we'd like to keep the `config.RunModes` struct up-to-date. This was the best way to do an enum that I found in Go but let me know if there's a better way.
```
var RunModes = struct {
	Build  RunMode
	Dev    RunMode
	Debug  RunMode
	Run    RunMode
	Deploy RunMode
	Render RunMode
}{
	Build:  "build",
	Dev:    "dev",
	Debug:  "debug",
	Run:    "run",
	Deploy: "deploy",
	Render: "render",
}
```

pkg/skaffold/build/cache/hash_test.go
gsquared94: now I directly use the evaluated build args for the hash. This includes evaluated env for the buildpacks builder which already has the `GOOGLE_DEVMODE=1` so we hash different between dev and other modes and don't need this hack:
image

gsquared94:
> > Dockerfile can expose ARG SKAFFOLD_GO_GCFLAGS and ARG SKAFFOLD_GO_LDFLAGS and expect Skaffold to supply meaningful values.
> 
> Why is this change required? Setting "GO_GC_FLAGS" and "GO_LDFLAGS" in dockerfile is a common practice. Changing this to "SKAFFOLD_GO_GCFLAGS" might break existing user's dockerfile

I'm not very opinionated about this. @briandealwis had suggested prefixing `SKAFFOLD_`, I think as a way to distinguish that it's an environment variable that we pass in explicitly. The user may set regular `GO_LDFLAGS` and `GO_GCFLAGS` in his Dockerfile to any arbitrary value and ignore skaffold supplied values entirely. I'm not sure why you think that it would break them though.

pkg/skaffold/config/options.go
briandealwis: This seems like a worthy situation to panic if it doesn't match.

gsquared94: I don't think that it needs to panic. Even if the command is not found in the static `RunModes` it should still be able to function as a `RunMode` object though it might not be useful.

gsquared94: For instance I've left out `init`, `completion`, `survey` commands from `RunModes` struct.

gsquared94: Also `opts.Command` gets set much earlier and only allows the supported `Cobra` commands in the first place.

gsquared94:
@briandealwis @tejal29 @nkubala PTAL.

pkg/skaffold/docker/default_args.go
```diff
@@ -0,0 +1,50 @@
...
+	"SKAFFOLD_GO_GCFLAGS": "'all=-N -l'", // disable build optimization for Golang
+	// TODO: Add for other languages
+}
+
+var BuildArgsForDev = map[string]string{
+	"SKAFFOLD_GO_LDFLAGS": "-w", // omit debug information in build output for Golang
+	// TODO: Add for other languages
+}
+
+func AppendDefaultArgs(mode config.RunMode, existing map[string]*string) map[string]*string {
```
briandealwis, line 31: Would it be possible to adjust this somehow to only add args if they're requested in the Dockerfile?  Otherwise you get warning messages during the build like:
```
Step 11/11 : ENTRYPOINT ["/app/backend"]
 ---> Using cache
 ---> 10cd7f4085cc
[Warning] One or more build-args [SKAFFOLD_GO_LDFLAGS] were not consumed
```

pkg/skaffold/docker/default_args.go
```diff
@@ -0,0 +1,50 @@
...
+		args = BuildArgsForDebug
+	default:
+		args = BuildArgsForDev
+	}
+
+	if existing == nil {
+		existing = make(map[string]*string)
+	}
+
+	for k, v := range args {
```
briandealwis, line 44: I'd really like to expose the run-mode — maybe `SKAFFOLD_RUN_MODE`?  For example, in #4699 there is a [Dockerfile](https://github.com/GoogleContainerTools/skaffold/pull/4699/files#diff-b6056c17f26b1f12838359e3f6c9edc5R14) with commands like the following:
```
RUN dotnet build "HelloWorld.csproj" -c Debug -o /app/build
```

Would be nice if they could do something like:
```
RUN dotnet build "HelloWorld.csproj" -c $(if [ "$SKAFFOLD_RUN_MODE" = debug]; then echo Debug; else echo Release; fi) -o /app/build
```

PR #4626, opened 2020-08-04
Increase the duration between survey prompts

tstromberg:
* If `survey` command is run, prompt once a year
* If `survey` command is not run, prompt once every 2 months

Also use idiomatic Go layout for durations (integer * duration).

### DIFF
--- a/pkg/skaffold/config/util.go
+++ b/pkg/skaffold/config/util.go
@@ -309,7 +309,7 @@ func recentlyPromptedOrTaken(cfg *ContextConfig) bool {
 	if cfg == nil || cfg.Survey == nil {
 		return false
 	}
-	return lessThan(cfg.Survey.LastTaken, 365*24*time.Hours) || lessThan(cfg.Survey.LastPrompted, 60*24*time.Hours)
+	return lessThan(cfg.Survey.LastTaken, 365*24*time.Hour) || lessThan(cfg.Survey.LastPrompted, 60*24*time.Hour)
 }
 
 func lessThan(date string, duration time.Duration) bool {
--- a/pkg/skaffold/config/util.go
+++ b/pkg/skaffold/config/util.go
@@ -37,8 +37,6 @@ import (
 const (
 	defaultConfigDir  = ".skaffold"
 	defaultConfigFile = "config"
-	tenDays           = time.Hour * 24 * 10
-	threeMonths       = time.Hour * 24 * 90
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/skaffold/config/util.go
```diff
@@ -311,7 +309,7 @@ func recentlyPromptedOrTaken(cfg *ContextConfig) bool {
 	if cfg == nil || cfg.Survey == nil {
 		return false
 	}
-	return lessThan(cfg.Survey.LastTaken, threeMonths) || lessThan(cfg.Survey.LastPrompted, tenDays)
+	return lessThan(cfg.Survey.LastTaken, 365*24*time.Hours) || lessThan(cfg.Survey.LastPrompted, 60*24*time.Hours)
```
tejal29, line 312: you will need to update the unit tests.

tstromberg:
> Please update the tests here https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/config/util_test.go#L472

Done.

pkg/skaffold/config/util_test.go
```diff
@@ -498,11 +498,11 @@ func TestShouldDisplayPrompt(t *testing.T) {
 			},
 		},
 		{
-			description: "should display prompt when last prompted is before 2 weeks",
+			description: "should display prompt when last prompted is older than 3 months",
```
briandealwis, line 501: These test descriptions should be expressed in days since they don't properly account for months with 28, 29, 30, 31 days.

pkg/skaffold/config/util_test.go
```diff
@@ -498,11 +498,11 @@ func TestShouldDisplayPrompt(t *testing.T) {
...
 			cfg: &ContextConfig{
 				Survey: &SurveyConfig{
 					DisablePrompt: util.BoolPtr(false),
-					LastPrompted:  "2019-01-10T00:00:00Z",
+					LastPrompted:  "2019-09-10T00:00:00Z",
```
briandealwis, line 505: This date is *after* the current time (2019-01-30T12:04:05Z).
```suggestion
					LastPrompted:  "2018-09-10T00:00:00Z",
```

pkg/skaffold/config/util_test.go
tstromberg: I don't understand - these are only text descriptions, and the other tests use the term 'months' as well? Should I be changing all of the existing test descriptions to say 'days' instead? This feels somewhat arbitrary.

briandealwis: I believe you should be changing the test descriptions.  The test descriptions were being expressed in relation to the previous date offsets of 3 months and 10 days.  You're revising the date offsets, and it just happens that the chosen dates also work for your revised date offsets (which means these tests weren't testing boundary conditions very precisely).  Most notable is the next test "_ should display prompt when last taken is before than 3 months ago_" which from its description should fail since you've changed the last-taken period to be 365 days ago.  But it passes because the probed date was more than 365 days ago.

nkubala (approved):

PR #4635, opened 2020-08-05
Dependabot fix done explicitly

gsquared94:
Duplicate of #4612 which bumps elliptic from 6.5.2 to 6.5.3 in examples/react-reload  app. This PR also makes the same change in the integration/examples/react-reload app


tejal29:
LGTM, please add which packages were updated in the PR descirption

PR #4646, opened 2020-08-10
skip test TestHouseKeepingMessagesShownForDev until 1.14.0

MarlonGamez:
**Description**
Skips the TestHouseKeepingMessagesShownForDev test. Since we've disabled the survey message for now, this test is causing travis to fail.

Revert this change once we've reenabled the survey message


MarlonGamez:
@gsquared94 do you mean to pass it as a parameter to the Skip method for it to log?

gsquared94:
> @gsquared94 do you mean to pass it as a parameter to the Skip method for it to log?

yes, exactly that

Issue #4616, opened 2020-08-03
skaffold deploy (and even skaffold fix) validates existence of "build->artifacts->context"


lucasmaj:
### Expected behavior
Existence of "build->artifacts->context" path has no bearing on "deploy" (nor "fix", nor any other pipeline that does not need it, so "build" only???). When using "deploy" separately from "build", context path (which is used for "build") is not guaranteed  to exist during "deploy". Does "deploy" need context anyway?
### Actual behavior
Error thrown stating that context does not exist. 

> invalid skaffold config: image "some.azurecr.io/myproject" context "./publish/myproject/" does not exist

### Information

- Skaffold version: 1.13.0
- Operating system: Windows and Linux, others unsure
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1
kind: Config
build:
  artifacts:
  - image: some.azurecr.io/myproject
    context: "./publish/myproject/"
    docker:
      dockerfile: "./path/to/some/Dockerfile"
deploy:
  kubectl:
    manifests:
    - myproject/deployment.yaml
```

### Steps to reproduce the behavior

1. skaffold build (or skaffold fix) above skaffold.yaml
2. To see a working version just comment out "context" or provide existing path.

gsquared94:
Your expected behavior definitely sounds reasonable. I'll discuss in our weekly triage what, if anything, necessitated the existence of a build context always.

lucasmaj:
Also would like to point out that this behavior was introduced in 1.13.0. Our workaround right now is to use v1.12.1.

briandealwis:
This was introduced by #4492.

lucasmaj:
Confirming it is fixed in v1.13.2. Thank you!

Issue #4491, opened 2020-07-17
Skaffold should error on an invalid `context` directory


briandealwis:
### Expected behavior

Noticed by [_Rune FS_ on StackOverflow](https://stackoverflow.com/questions/62404426/skaffold-cant-find-image-it-has-just-build): Skaffold does not raise an error on startup if an artifact has an invalid context directory, and instead produces a difficult to decipher error.

### Actual behavior

```diff
--- examples/getting-started/skaffold.yaml
+++ examples/getting-started/skaffold.yaml
@@ -3,6 +3,7 @@ kind: Config
 build:
   artifacts:
   - image: skaffold-example
+    context: this/path/does/not/exist
 deploy:
   kubectl:
     manifests:
```

```console
$ skaffold build
Generating tags...
 - skaffold-example -> skaffold-example:latest
Some taggers failed. Rerun with -vdebug for errors.
Checking cache...
 - skaffold-example: Error checking cache.
getting hash for artifact "skaffold-example": getting dependencies for "skaffold-example": file pattern [main.go] must match at least one file
```

Try without the build-cache:
``` console
$ skaffold  build --cache-artifacts=false
Generating tags...
 - skaffold-example -> skaffold-example:latest
Some taggers failed. Rerun with -vdebug for errors.
Found [minikube] context, using local docker daemon.
Building [skaffold-example]...
Sending build context to Docker daemon 

couldn't build "skaffold-example": docker build: error during connect: Post "https://192.168.192.145:2376/v1.40/build?buildargs=null&cachefrom=null&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile&forcerm=1&labels=null&memory=0&memswap=0&networkmode=&rm=0&shmsize=0&t=skaffold-example%3Alatest&target=&ulimits=null&version=": creating docker context: getting relative tar paths: file pattern [main.go] must match at least one file
```

PR #4657, opened 2020-08-13
Move context validation to build phase so as to not interfere with deploy

briandealwis:
**Related**: #4491
Reverts: #4492
Fixes:  #4616

**Description**

Revert the validation for artifact contexts (workspaces) from introduced by #4492 for the reasons outlined in #4616.  Instead perform artifact context checks before undertaking a build.  

Since this error is unexpected, we just report the first artifact with an error rather than collect all errors.

```console
$ skaffold build
image "leeroy-web" context "foo-leeroy-web" does not exist
```


gsquared94:
#4616 seems to be a very fundamental scenario for deploy; didn't we have any integration test for that? Can we add one?

briandealwis:
Funnily enough it takes a bit of work to create an IT around this as you still need the kubernetes manifests to be deployed.

briandealwis:
PTAL @gsquared94

gsquared94 (approved):
LGTM!

Issue #4663, opened 2020-08-14
skaffold does not reload when a kustomize `component` is modified


zero-below:
skaffold only monitors kustomize `resources` and `bases` for changes, not `components`.

```
cd $(mktemp -d)
mkdir a b

cat <<'EOF' > a/kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

components:
- ../b/
EOF

cat <<'EOF' > b/kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1alpha1
kind: Component

resources:
- deployment.yaml
EOF

cat <<'EOF' > b/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test1
EOF

cat <<'EOF' > skaffold.yaml
apiVersion: skaffold/v2beta6
kind: Skaffold
deploy:
  kustomize:
    paths:
    - a/
EOF

skaffold dev --render-only=true
# In another window, `touch a/kustomizeation.yaml` -- skaffold reloads.
# `touch b/kustomization.yaml` -- skaffold does not reload.
```

- Skaffold version: v1.13.1
- Operating system: OSX Darwin Kernel Version 19.6.0

I believe this may be a small change at https://github.com/GoogleContainerTools/skaffold/blob/6994b45a9a02daf03deb739413071adc5bdbcb47/pkg/skaffold/deploy/kustomize.go#L259 -- but I haven't had a chance to dig further yet.

PR #4664, opened 2020-08-14
Add support for reloading kustomize components.

zero-below:
Fixes: #4663 

**Description**
Adds support for auto-reload of kustomize `components` files, in dev-mode. Simply adds the list of `components` to get parsed from yaml, and appended to the list of `candidates`.


zero-below:
@googlebot I fixed it.

Issue #3904, opened 2020-04-01
Support kpt as a deployer


bgrant0607:
https://github.com/GoogleContainerTools/kpt
https://googlecontainertools.github.io/kpt/

Similar to the kustomize+kubectl support:
https://skaffold.dev/docs/pipeline-stages/deployers/#deploying-with-kustomize

tstromberg:
+1 - we should do this.

jlewi:
How will skaffold.yaml and kpt setters interact if at all? For example, when skaffold builds/deploys a kpt package will it update setters to reflect the most recent values?

I often use skaffold and kpt together. I often use skaffold to build my images and then update my kpt package to use the newly built image.

I usually do this by having skaffold output the image and then using kpt + yq to set it; e.g

```
skaffold build --kube-context=${CONTEXT} -v debug --file-output=image.json
kpt cfg set -R . image $(shell yq r image.json builds[0].tag )
```

Any thought about whether this is something skaffold should support out of box?

yuwenma:
> How will skaffold.yaml and kpt setters interact if at all? For example, when skaffold builds/deploys a kpt package will it update setters to reflect the most recent values?
> 
> I often use skaffold and kpt together. I often use skaffold to build my images and then update my kpt package to use the newly built image.
> 
> I usually do this by having skaffold output the image and then using kpt + yq to set it; e.g
> 
> ```
> skaffold build --kube-context=${CONTEXT} -v debug --file-output=image.json
> kpt cfg set -R . image $(shell yq r image.json builds[0].tag )
> ```
> 
> Any thought about whether this is something skaffold should support out of box?

Hi Jeremy, skaffold and kpt setters are working independently right now, but we have plans to support what you need here. Basically, you would be able to define kpt setters as a config file and provide its path in skaffold.yaml kpt. For now, I think a bash script to automate the step would be most efficient.

yuwenma:
The kpt deployer is added to skaffold (stable version) and I think we can mark this issue done. Future changes related to render v2 is orthogonal to this request.

irvifa:
HI @yuwenma do we have documentation on how to use this through the Skaffold?

yuwenma:
> HI Yuwen Ma do we have documentation on how to use this through the Skaffold?

@irvifa You can try out these interactive codelabs: 
- [01 kpt validation](https://github.com/GoogleContainerTools/skaffold/tree/master/codelab/01_kpt-validate)
[...]

PR #4682, opened 2020-08-17
Kpt Deployer Skeleton

felixtran39:
**Related**: #3904

**Description**
This is the skeletion of the deployer interface for the upcoming kpt deployer feature. There are no user facing changes.

**Follow-up Work**
The follow up work to this will be the changes to the config because the deployer implementation depends on it. Afterward, there will be PRs to implement the functionality.


MarlonGamez:
This looks good to me, but I would maybe mention that this is just the skeleton of the kpt deployer, and in the follow up section you can mention that another pr needs to be created implementing the functionality

PR #4689, opened 2020-08-18
Kpt Deployer Config

felixtran39:
**Related**: #3904

**Description**
This is the config changes for the kpt deployer. The kpt deployer's implementation depends on this config.

Users will be able to specify using a kpt deployer in their skaffold.yaml. The functionality of the kpt deployer itself will not be implemented when this PR is merged. User facing docs not changed because we don't want users to use this deployer yet (functionality not implemented).

Here is what it will look like. 
```
deploy:
  kpt:
    dir: "foo"
    fn:
      fnPath: "foo"
      image: "foo"
    live:
      apply:
      - "foo"
      - "bar"
```

Functionality for the deployer coming next.

### DIFF
--- a/docs/content/en/schemas/v2beta7.json
+++ b/docs/content/en/schemas/v2beta7.json
@@ -1603,8 +1603,8 @@
       "properties": {
         "dir": {
           "type": "string",
-          "description": "path to the directory to run kpt functions against and deploy to the cluster.",
-          "x-intellij-html-description": "path to the directory to run kpt functions against and deploy to the cluster."
+          "description": "path to the directory to run kpt functions against.",
+          "x-intellij-html-description": "path to the directory to run kpt functions against."
         },
         "fn": {
           "$ref": "#/definitions/KptFn",
--- a/docs/content/en/schemas/v2beta7.json
+++ b/docs/content/en/schemas/v2beta7.json
@@ -828,6 +828,11 @@
           "description": "*beta* uses the `helm` CLI to apply the charts to the cluster.",
           "x-intellij-html-description": "<em>beta</em> uses the <code>helm</code> CLI to apply the charts to the cluster."
         },
+        "kpt": {
+          "$ref": "#/definitions/KptDeploy",
+          "description": "*beta* uses the `kpt` CLI to manage and deploy manifests.",
+          "x-intellij-html-description": "<em>beta</em> uses the <code>kpt</code> CLI to manage and deploy manifests."
+        },
         "kubeContext": {
           "type": "string",
           "description": "Kubernetes context that Skaffold should deploy to.",
@@ -861,6 +866,7 @@
         "helm",
         "kubectl",
         "kustomize",
+        "kpt",
         "statusCheckDeadlineSeconds",
         "kubeContext",
         "logs"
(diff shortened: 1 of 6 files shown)


briandealwis:
(Looked good to me too)

docs/content/en/schemas/v2beta7.json
```diff
@@ -861,6 +866,7 @@
         "helm",
         "kubectl",
         "kustomize",
```
yuwenma, line 868: Add kpt before kubectl or before kustomize?

briandealwis: This schema is auto-generated from the corresponding structure in config.go.  What matters more is the order in which they are handled in [getDeployer()](https://github.com/felixtran39/skaffold/blob/df42484b7c66635847c7bfc429c8ab2f9a360f10/pkg/skaffold/runner/new.go#L190-L217).  I think it makes sense to have kpt be last, since that way it has a chance to act upon the results of the other deployers?

pkg/skaffold/schema/latest/config.go
```diff
@@ -523,6 +526,34 @@ type KustomizeDeploy struct {
...
+	// Live adds additional configurations for `kpt live`.
+	Live KptLive `yaml:"live,omitempty"`
+}
+
+// KptFn adds additional configurations used when calling `kpt fn`.
+type KptFn struct {
+	// FnPath is the path to a kpt function or pipeline.
+	FnPath string `yaml:"fnPath,omitempty"`
+
+	// Image specifies an image to be run as a function.
```
yuwenma, line 546: ditto.

pkg/skaffold/schema/v2beta6/upgrade.go
```diff
@@ -24,6 +24,8 @@ import (
 
 // Upgrade upgrades a configuration to the next version.
 // Config changes from v2beta6 to v2beta7
+// 1. Additions:
+//		New KptDeploy, KptFn, and KptLive to support Kpt Deployer.
```
yuwenma, line 28: nit: remove the second ","?

docs/content/en/schemas/v2beta7.json
felixtran39: The order they are handled in `getDeployer()` shouldn't affect the result either. If alphabetical order is the convention I can change that.

pkg/skaffold/schema/v2beta6/upgrade.go
briandealwis: Not if you like oxford commas :-)

docs/content/en/schemas/v2beta7.json
```diff
@@ -1593,6 +1599,73 @@
...
+    "KptFn": {
+      "properties": {
+        "fnPath": {
+          "type": "string",
+          "description": "path to a kpt function or pipeline.",
+          "x-intellij-html-description": "path to a kpt function or pipeline."
+        },
+        "image": {
+          "type": "string",
+          "description": "specifies an image to be run as a function.",
```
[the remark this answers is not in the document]
briandealwis: You'd use a `yamltags:"oneOf= kpt"` to the appropriate fields.

docs/content/en/schemas/v2beta7.json
yuwenma: oh, what I mean is that strategically we'd like to recommend kpt over kustomize and kubectl to users. Since this specific field is "preferredOrder", we can let kpt comes up before the other less recommended ones. Agree that the order doesn't affect the real config operations. My comment is more from the user's perspective.

pkg/skaffold/schema/v2beta6/upgrade.go
yuwenma: to be honest, I didn't know this before. Thanks Brian and @felixtran39 😄

PR #4691, opened 2020-08-18
Kpt Deployer applyDir implementation and tests

felixtran39:
**Related**: #3904

**Description**
`getApplyDir` is a function which resolves the package to deploy to the cluster for kpt deployer. If unspecified, it will create a hidden directory .kpt-hydrated.

**User facing changes**
Users will have the ability to specify their own applyDir in skaffold.yaml.

**Follow-up Work**
Further implementation of kpt deployer will be next.

### DIFF
--- a/docs/content/en/schemas/v2beta7.json
+++ b/docs/content/en/schemas/v2beta7.json
@@ -1601,6 +1601,11 @@
     },
     "KptDeploy": {
       "properties": {
+        "applyDir": {
+          "type": "string",
+          "description": "path to the directory to deploy to the cluster.",
+          "x-intellij-html-description": "path to the directory to deploy to the cluster."
+        },
         "dir": {
           "type": "string",
           "description": "path to the directory to run kpt functions against.",
@@ -1619,6 +1624,7 @@
       },
       "preferredOrder": [
         "dir",
+        "applyDir",
         "fn",
         "live"
       ],
--- a/pkg/skaffold/deploy/kpt.go
+++ b/pkg/skaffold/deploy/kpt.go
@@ -18,18 +18,27 @@ package deploy
 
 import (
 	"context"
+	"fmt"
 	"io"
+	"os"
+	"os/exec"
+	"strings"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner/runcontext"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 )
 
 // KptDeployer deploys workflows with kpt CLI
 type KptDeployer struct {
+	*latest.KptDeploy
 }
 
 func NewKptDeployer(runCtx *runcontext.RunContext, labels map[string]string) *KptDeployer {
-	return &KptDeployer{}
+	return &KptDeployer{
+		KptDeploy: runCtx.Pipeline().Deploy.KptDeploy,
+	}
 }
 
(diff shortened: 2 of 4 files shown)


pkg/skaffold/deploy/kpt.go
```diff
@@ -47,3 +56,53 @@ func (k *KptDeployer) Cleanup(ctx context.Context, out io.Writer) error {
...
+// creates a hidden directory in place of applyDir.
+func (k *KptDeployer) getApplyDir(ctx context.Context) (string, error) {
+	if k.ApplyDir != "" {
+		if _, err := os.Stat(k.ApplyDir); os.IsNotExist(err) {
+			return "", err
+		}
+		return k.ApplyDir, nil
+	}
+
+	applyDir := ".kpt-hydrated"
```
yuwenma, line 70: We would want to check the existence of .kpt-hydrated befor creation.

pkg/skaffold/deploy/kpt.go
```diff
@@ -47,3 +56,53 @@ func (k *KptDeployer) Cleanup(ctx context.Context, out io.Writer) error {
...
+
+	// 0755 is a permission setting where the owner can read, write, and execute.
+	// Others can read and execute but not modify the file.
+	if err := os.MkdirAll(applyDir, 0755); err != nil {
+		return "", fmt.Errorf("applyDir was unspecified. creating applyDir: %w", err)
+	}
+
+	cmd := exec.CommandContext(ctx, "kpt", kptCommandArgs(applyDir, []string{"live", "init"}, nil, nil)...)
```
yuwenma, line 78: We only call `kpt live init` if the inventory-template.yaml doesn't exist in the applyDir.

docs/content/en/schemas/v2beta7.json
```diff
@@ -1601,6 +1601,11 @@
     },
     "KptDeploy": {
       "properties": {
+        "applyDir": {
+          "type": "string",
+          "description": "path to the directory to deploy to the cluster.",
+          "x-intellij-html-description": "path to the directory to deploy to the cluster."
+        },
```
yuwenma, line 1608: Is that possible to use v2beta8.json (if not, I'm fine with adding the applyDir here).

It's safer to surface the kpt options after all the backend supports are done.

pkg/skaffold/deploy/kpt.go
felixtran39: If .kpt-hydrated already exists, os.MkdirAll does nothing and returns nil.

docs/content/en/schemas/v2beta7.json
felixtran39: The schemas are autogenerated. It is possible to use v2beta8 but that would require having latest config = v2beta7 and then created a "next latest" config = v2beta8. This seems a bit awkward.

yuwenma: I see. Is v2beta7 the next new version or current version?

pkg/skaffold/deploy/kpt.go
felixtran39: I agree. The reason I was hesitant with checking if it doesn't exist first is because users can change the name of inventory-template.yaml and `kpt live init` will end up creating another template resource.

docs/content/en/schemas/v2beta7.json
MarlonGamez: v2beta7 hasn't been released yet, but is the version that is being developed and still allows for changes. Once we do another skaffold release, this version will become locked and someone will have to generate v2beta8

felixtran39: v2beta7 is the upcoming version and v2beta6 is the current version.

PR #4696, opened 2020-08-18
Kpt Deployer Dependencies() implementation and tests

felixtran39:
**Related**: #3904

**Description**
This is the implementation and unit tests for the dependencies method of the kpt deployer. Dependencies include the functions being ran, the files used by those functions, and kustomizations.

**Follow-up Work**
No follow up work for this particular method. Follow up work for the Kpt deployer will include the other methods that are yet to be implemented.

### DIFF
--- a/pkg/skaffold/deploy/kpt.go
+++ b/pkg/skaffold/deploy/kpt.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
@@ -47,7 +48,19 @@ func (k *KptDeployer) Deploy(ctx context.Context, out io.Writer, builds []build.
 }
 
 func (k *KptDeployer) Dependencies() ([]string, error) {
-	return nil, nil
+	deps := newStringSet()
+	if len(k.Fn.FnPath) > 0 {
+		deps.insert(k.Fn.FnPath)
+	}
+
+	configDeps, err := getResources(k.Dir)
+	if err != nil {
+		return nil, fmt.Errorf("finding dependencies in %s: %w", k.Dir, err)
+	}
+
+	deps.insert(configDeps...)
+
+	return deps.toList(), nil
 }
 
 func (k *KptDeployer) Cleanup(ctx context.Context, out io.Writer) error {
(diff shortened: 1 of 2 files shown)


pkg/skaffold/deploy/kpt.go
```diff
@@ -111,3 +124,35 @@ func kptCommandArgs(dir string, commands, flags, globalFlags []string) []string
...
+	var files []string
+
+	if _, err := os.Stat(root); os.IsNotExist(err) {
+		return nil, err
+	}
+
+	err := filepath.Walk(root, func(path string, info os.FileInfo, _ error) error {
+		isResource, err := regexp.MatchString(`\.ya?ml$`, filepath.Base(path))
+		if err != nil {
+			return fmt.Errorf("matching regex: %w", err)
```
yuwenma, line 140: nit: add the `path` to the err message.

pkg/skaffold/deploy/kpt.go
```diff
@@ -111,3 +124,35 @@ func kptCommandArgs(dir string, commands, flags, globalFlags []string) []string
...
+
+		if info.IsDir() {
+			depsForKustomization, err := dependenciesForKustomization(path)
+			if err != nil {
+				return err
+			}
+
+			files = append(files, depsForKustomization...)
+		} else if isResource {
+			files = append(files, strings.ReplaceAll(path, "\\", "/"))
```
yuwenma, line 151: Is this path replacement for windows? Suggest adding a comment about that.

pkg/skaffold/deploy/kpt_test.go
```diff
@@ -49,18 +49,124 @@ func TestKpt_Deploy(t *testing.T) {
...
+		{
+			description: "dir with subdirs",
+			cfg: &latest.KptDeploy{
+				Dir: ".",
+			},
+			createFiles: map[string]string{
+				"food.yml":          "",
+				"foo/bar.yaml":      "",
+				"foo/bat/bad.yml":   "",
+				"foo/bat/README.md": "",
```
yuwenma, line 92: suggest adding a path containing the "\\"

pkg/skaffold/deploy/kpt.go
```diff
@@ -111,3 +124,35 @@ func kptCommandArgs(dir string, commands, flags, globalFlags []string) []string
...
+// and all local kustomization dependencies under root.
+func getResources(root string) ([]string, error) {
+	var files []string
+
+	if _, err := os.Stat(root); os.IsNotExist(err) {
+		return nil, err
+	}
+
+	err := filepath.Walk(root, func(path string, info os.FileInfo, _ error) error {
+		isResource, err := regexp.MatchString(`\.ya?ml$`, filepath.Base(path))
```
yuwenma, line 138: This is fair enough for now. But it is actually not accurate because the content of the .yaml files should also be KRM based. Can you add a comment saying that we expect kpt to provide a more accurate functions to walk through a dir and pulls out the kustomize/kpt resources.

pkg/skaffold/deploy/kpt_test.go
felixtran39: Good idea. I will also add a path with `\`.

pkg/skaffold/deploy/kpt.go
felixtran39: Sure thing. I agree, this may be something that can be offered when we move away from consuming kpt as a CLI.

PR #4697, opened 2020-08-18
Kpt Deployer Cleanup() implementation and tests

felixtran39:
Related: #3904

Description
This is the implementation and unit tests for the Cleanup() method of the kpt deployer. This involves running `kpt live destroy /DIR` where `/DIR` is applyDir (contains the hydrated resources).

Follow-up Work
No follow up work for this particular method. Follow up work for the Kpt deployer will include the other methods that are yet to be implemented.

### DIFF
--- a/pkg/skaffold/deploy/kpt.go
+++ b/pkg/skaffold/deploy/kpt.go
@@ -50,7 +50,17 @@ func (k *KptDeployer) Dependencies() ([]string, error) {
 	return nil, nil
 }
 
-func (k *KptDeployer) Cleanup(ctx context.Context, out io.Writer) error {
+func (k *KptDeployer) Cleanup(ctx context.Context, _ io.Writer) error {
+	applyDir, err := k.getApplyDir(ctx)
+	if err != nil {
+		return fmt.Errorf("getting applyDir: %w", err)
+	}
+
+	cmd := exec.CommandContext(ctx, "kpt", kptCommandArgs(applyDir, []string{"live", "destroy"}, nil, nil)...)
+	if err = util.RunCmd(cmd); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/skaffold/deploy/kpt.go
```diff
@@ -50,7 +50,17 @@ func (k *KptDeployer) Dependencies() ([]string, error) {
 	return nil, nil
 }
 
-func (k *KptDeployer) Cleanup(ctx context.Context, out io.Writer) error {
+func (k *KptDeployer) Cleanup(ctx context.Context, _ io.Writer) error {
```
yuwenma, line 53: nit: Docstring.

pkg/skaffold/deploy/kpt.go
```diff
@@ -50,7 +50,17 @@ func (k *KptDeployer) Dependencies() ([]string, error) {
...
+	applyDir, err := k.getApplyDir(ctx)
+	if err != nil {
+		return fmt.Errorf("getting applyDir: %w", err)
+	}
+
+	cmd := exec.CommandContext(ctx, "kpt", kptCommandArgs(applyDir, []string{"live", "destroy"}, nil, nil)...)
+	if err = util.RunCmd(cmd); err != nil {
+		return fmt.Errorf("kpt live destroy: %w", err)
+	}
```
yuwenma, line 62: Can you double check that when `kpt live destroy` fails, if this can output the kpt error message? IIRC, the kpt errors would be treated as the skaffold stdout. If so, we would want the cmd to write the stdout (the real kpt error) to a buff and return the buff here.

MarlonGamez:
Thanks for addressing Yuwen's comments, lgtm

PR #4698, opened 2020-08-19
Remove remote rules

tejal29:
in order to iterate quickly over error recommendations and suggestions I added remote rules. 
However, after talking to the team, we felt this is not the right direction and we should target suggestions to be part of a release.

Removing these rules.


briandealwis:
LGTM

I didn't realize this existed :-)  What's nice about these external rules is they provide a way to accommodate front-end-specific actionable suggestions (e.g., if we have a `diag` command, or a different front-end for GCB etc.).

PR #4701, opened 2020-08-19
Identify minikube cluster for any profile name

### LINKED ISSUE #3668: enable skaffold to work on minikube profiles

gsquared94:
Fixes #3668

### Description
Currently we need to explicitly specify the minikube profile name as a skaffold flag or limit it to only 'minikube'. This PR introduces several checks for detecting a minikube cluster so that the flag will no longer be required and we can name the cluster anything.
We check:
- The context is 'minikube', or
- ~~The k8s nodes have a minikube label (refer https://github.com/kubernetes/minikube/pull/6717), or~~
- The cluster certificate path is within the `~/.minikube` directory (refer [this](https://github.com/GoogleContainerTools/skaffold/pull/4701#pullrequestreview-474795316) comment), or
- The `minikube profile list` returns a config name matching the context, and
- If the minikube driver is a VM then the node IP should also match the k8s api server url

### DIFF
--- a/pkg/skaffold/cluster/minikube.go
+++ b/pkg/skaffold/cluster/minikube.go
@@ -67,7 +67,7 @@ func (clientImpl) IsMinikube(kubeContext string) bool {
 	if kubeContext == constants.DefaultMinikubeContext {
 		return true
 	}
-	if _, err := minikubeBinaryFunc();  err != nil {
+	if _, err := minikubeBinaryFunc(); err != nil {
 		logrus.Tracef("Minikube cluster not detected: %v", err)
 		return false
 	}
--- a/pkg/skaffold/cluster/minikube.go
+++ b/pkg/skaffold/cluster/minikube.go
@@ -67,8 +67,7 @@ func (clientImpl) IsMinikube(kubeContext string) bool {
 	if kubeContext == constants.DefaultMinikubeContext {
 		return true
 	}
-	_, err := minikubeBinaryFunc()
-	if err != nil {
+	if _, err := minikubeBinaryFunc();  err != nil {
 		logrus.Tracef("Minikube cluster not detected: %v", err)
 		return false
 	}
--- a/pkg/skaffold/cluster/minikube.go
+++ b/pkg/skaffold/cluster/minikube.go
@@ -65,24 +65,24 @@ func (clientImpl) IsMinikube(kubeContext string) bool {
 	}
 	_, err := minikubeBinaryFunc()
 	if err != nil {
-		logrus.Debugf("Minikube cluster not detected: %v", err)
+		logrus.Tracef("Minikube cluster not detected: %v", err)
 		return false
 	}
 
 	if ok, err := matchClusterCertPath(kubeContext); err != nil {
-		logrus.Debugf("failed to match cluster certificate path: %v", err)
+		logrus.Tracef("failed to match cluster certificate path: %v", err)
 	} else if ok {
 		logrus.Debugf("Minikube cluster detected: cluster certificate for context %q found inside the minikube directory", kubeContext)
 		return true
 	}
 
 	if ok, err := matchProfileAndServerURL(kubeContext); err != nil {
-		logrus.Debugf("failed to match minikube profile: %v", err)
+		logrus.Tracef("failed to match minikube profile: %v", err)
 	} else if ok {
 		logrus.Debugf("Minikube cluster detected: context %q has matching profile name or server url", kubeContext)
 		return true
 	}
-	logrus.Debugf("Minikube cluster not detected for context %q", kubeContext)
+	logrus.Tracef("Minikube cluster not detected for context %q", kubeContext)
 	return false
 }
 
@@ -138,7 +138,7 @@ func matchProfileAndServerURL(kubeContext string) (bool, error) {
 		return false, fmt.Errorf("getting kubernetes server url: %w", err)
 	}
 
-	logrus.Debugf("kubernetes server url: %s", apiServerURL)
+	logrus.Tracef("kubernetes server url: %s", apiServerURL)
 
 	ok, err := matchServerURLFor(kubeContext, apiServerURL)
 	if err != nil {
--- a/pkg/skaffold/cluster/minikube.go
+++ b/pkg/skaffold/cluster/minikube.go
@@ -24,14 +24,14 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path/filepath"
 
 	"github.com/sirupsen/logrus"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/homedir"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
-	k8s "github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/client"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/context"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 )
@@ -42,6 +42,7 @@ var GetClient = getClient
 var (
 	minikubeBinaryFunc      = minikubeBinary
 	getRestClientConfigFunc = context.GetRestClientConfig
+	getClusterInfo          = context.GetClusterInfo
 )
 
 type Client interface {
@@ -65,13 +66,13 @@ func (clientImpl) IsMinikube(kubeContext string) bool {
 	_, err := minikubeBinaryFunc()
 	if err != nil {
 		logrus.Debugf("Minikube cluster not detected: %v", err)
-		return false // minikube binary not found
+		return false
 	}
 
-	if ok, err := matchNodeLabel(kubeContext); err != nil {
-		logrus.Debugf("failed to check minikube node labels: %v", err)
+	if ok, err := matchClusterCertPath(kubeContext); err != nil {
+		logrus.Debugf("failed to match cluster certificate path: %v", err)
 	} else if ok {
-		logrus.Debugf("Minikube cluster detected: context %q nodes have minikube labels", kubeContext)
+		logrus.Debugf("Minikube cluster detected: cluster certificate for context %q found inside the minikube directory", kubeContext)
 		return true
 	}
 
@@ -112,20 +113,16 @@ func minikubeBinary() (string, error) {
 	return filename, nil
 }
 
-func matchNodeLabel(kubeContext string) (bool, error) {
-	client, err := k8s.Client()
+// matchClusterCertPath checks if the cluster certificate for this context is from inside the minikube directory
+func matchClusterCertPath(kubeContext string) (bool, error) {
+	c, err := getClusterInfo(kubeContext)
 	if err != nil {
-		return false, fmt.Errorf("getting Kubernetes client: %w", err)
+		return false, fmt.Errorf("getting kubernetes config: %w", err)
 	}
-	opts := v1.ListOptions{
-		LabelSelector: fmt.Sprintf("minikube.k8s.io/name=%s", kubeContext),
-		Limit:         100,
+	if c.CertificateAuthority == "" {
+		return false, nil
 	}
-	l, err := client.CoreV1().Nodes().List(opts)
-	if err != nil {
-		return false, fmt.Errorf("listing nodes with matching label: %w", err)
-	}
-	return l != nil && len(l.Items) > 0, nil
+	return util.IsSubPath(minikubePath(), c.CertificateAuthority), nil
 }
 
 // matchProfileAndServerURL checks if kubecontext matches any valid minikube profile
@@ -183,6 +180,18 @@ func matchServerURLFor(kubeContext string, serverURL *url.URL) (bool, error) {
 		}
 	}
 	return false, nil
+}
+
+// minikubePath returns the path to the user's minikube dir
+func minikubePath() string {
+	minikubeHomeEnv := os.Getenv("MINIKUBE_HOME")
+	if minikubeHomeEnv == "" {
+		return filepath.Join(homedir.HomeDir(), ".minikube")
+	}
+	if filepath.Base(minikubeHomeEnv) == ".minikube" {
+		return minikubeHomeEnv
+	}
+	return filepath.Join(minikubeHomeEnv, ".minikube")
 }
 
 type data struct {
--- a/pkg/skaffold/cluster/minikube.go
+++ b/pkg/skaffold/cluster/minikube.go
@@ -20,7 +20,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"log"
 	"net/url"
 	"os"
 	"os/exec"
@@ -43,6 +42,11 @@ var (
 	minikubeBinaryFunc      = minikubeBinary
 	getRestClientConfigFunc = context.GetRestClientConfig
 	getClusterInfo          = context.GetClusterInfo
+)
+
+const (
+	VirtualBox = "virtualbox"
+	HyperKit   = "hyperkit"
 )
 
 type Client interface {
@@ -148,11 +152,7 @@ func matchProfileAndServerURL(kubeContext string) (bool, error) {
 }
 
 func matchServerURLFor(kubeContext string, serverURL *url.URL) (bool, error) {
-	cmd, err := minikubeExec("profile", "list", "-o", "json")
-	if err != nil {
-		return false, fmt.Errorf("executing minikube command: %w", err)
-	}
-
+	cmd, _ := minikubeExec("profile", "list", "-o", "json")
 	out, err := util.RunCmdOut(cmd)
 	if err != nil {
 		return false, fmt.Errorf("getting minikube profiles: %w", err)
@@ -160,7 +160,7 @@ func matchServerURLFor(kubeContext string, serverURL *url.URL) (bool, error) {
 
 	var data data
 	if err = json.Unmarshal(out, &data); err != nil {
-		log.Fatal(fmt.Errorf("failed to unmarshal data: %w", err))
+		return false, fmt.Errorf("failed to unmarshal data: %w", err)
 	}
 
 	for _, v := range data.Valid {
@@ -168,7 +168,7 @@ func matchServerURLFor(kubeContext string, serverURL *url.URL) (bool, error) {
 			continue
 		}
 
-		if v.Config.Driver != "hyperkit" && v.Config.Driver != "virtualbox" {
+		if v.Config.Driver != HyperKit && v.Config.Driver != VirtualBox {
 			// Since node IPs don't match server API for other drivers we assume profile name match is enough.
 			// TODO: Revisit once https://github.com/kubernetes/minikube/issues/6642 is fixed
 			return true, nil
--- a/pkg/skaffold/cluster/minikube.go
+++ b/pkg/skaffold/cluster/minikube.go
@@ -1,0 +1,206 @@
+/*
+Copyright 2020 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cluster
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log"
+	"net/url"
+	"os"
+	"os/exec"
+
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
+	k8s "github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/client"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/context"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+)
+
+var GetClient = getClient
+
+// To override during tests
+var (
+	minikubeBinaryFunc      = minikubeBinary
+	getRestClientConfigFunc = context.GetRestClientConfig
+)
+
+type Client interface {
+	// IsMinikube returns true if the given kubeContext maps to a minikube cluster
+	IsMinikube(kubeContext string) bool
+	// MinikubeExec returns the Cmd struct to execute minikube with given arguments
+	MinikubeExec(arg ...string) (*exec.Cmd, error)
+}
+
+type clientImpl struct{}
+
+func getClient() Client {
+	return clientImpl{}
+}
+
+func (clientImpl) IsMinikube(kubeContext string) bool {
+	// short circuit if context is 'minikube'
+	if kubeContext == constants.DefaultMinikubeContext {
+		return true
+	}
+	_, err := minikubeBinaryFunc()
+	if err != nil {
+		logrus.Debugf("Minikube cluster not detected: %v", err)
+		return false // minikube binary not found
+	}
+
+	if ok, err := matchNodeLabel(kubeContext); err != nil {
+		logrus.Debugf("failed to check minikube node labels: %v", err)
+	} else if ok {
+		logrus.Debugf("Minikube cluster detected: context %q nodes have minikube labels", kubeContext)
+		return true
+	}
+
+	if ok, err := matchProfileAndServerURL(kubeContext); err != nil {
+		logrus.Debugf("failed to match minikube profile: %v", err)
+	} else if ok {
+		logrus.Debugf("Minikube cluster detected: context %q has matching profile name or server url", kubeContext)
+		return true
+	}
+	logrus.Debugf("Minikube cluster not detected for context %q", kubeContext)
+	return false
+}
+
+func (clientImpl) MinikubeExec(arg ...string) (*exec.Cmd, error) {
+	return minikubeExec(arg...)
+}
+
+func minikubeExec(arg ...string) (*exec.Cmd, error) {
+	b, err := minikubeBinaryFunc()
+	if err != nil {
+		return nil, fmt.Errorf("getting minikube executable: %w", err)
+	}
+	return exec.Command(b, arg...), nil
+}
+
+func minikubeBinary() (string, error) {
+	execName := "minikube"
+	if found, _ := util.DetectWSL(); found {
+		execName = "minikube.exe"
+	}
+	filename, err := exec.LookPath(execName)
+	if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,204 @@
...
+	// short circuit if context is 'minikube'
+	if kubeContext == constants.DefaultMinikubeContext {
+		return true, nil
+	}
+	_, err := mkBinaryFunc()
+	if err != nil {
+		return false, nil // minikube binary not found
+	}
+
+	if ok, err := matchNodeLabel(kubeContext); err != nil {
```
tejal29, line 70: as per the description is matchNoede && matchProfileAndServerURL? or is it a OR?
```
(( The k8s nodes have a minikube label (refer kubernetes/minikube#6717), or
The minikube profile list returns a config name matching the context ) and
If the minikube driver is a VM then the node IP should also match the k8s api server url)
```

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,204 @@
...
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
+		return "", fmt.Errorf("unable to find minikube executable. File not found %s", filename)
+	}
+	return filename, nil
+}
+
+func matchNodeLabel(profileName string) (bool, error) {
+	client, err := k8s.Client()
+	if err != nil {
+		return false, fmt.Errorf("getting Kubernetes client: %w", err)
```
tejal29, line 116: can we just return false or true here and in `matchProfileAndServerURL`. 

That ways, we can simply check
```
def IsMinikube (kubeContext string ) bool {
    .... 
   return matchNodeLabel || matchProfileAndServerURL 
}
```

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,204 @@
...
+	MinikubeExec(arg ...string) (*exec.Cmd, error)
+}
+
+type clientImpl struct{}
+
+func getClient() Client {
+	return clientImpl{}
+}
+
+func (clientImpl) IsMinikube(kubeContext string) (bool, error) {
```
tejal29, line 60: Can we change this to just return bool and log error.

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,204 @@
...
+	logrus.Debugf("kubernetes server url: %s", apiServerURL)
+
+	ok, err := matchServerURLFor(kubeContext, apiServerURL)
+	if err != nil {
+		return false, fmt.Errorf("checking minikube node url: %w", err)
+	}
+	return ok, nil
+}
+
+func matchServerURLFor(profileName string, serverURL *url.URL) (bool, error) {
```
tejal29, line 151: Same here, log error and return only bool.

pkg/skaffold/cluster/minikube.go
gsquared94: either matching node label or matching profile name(and  server url for some driver types) suffices. The description has an `or` between them.

pkg/skaffold/cluster/minikube.go
gsquared94: done

pkg/skaffold/cluster/minikube.go
gsquared94: I'll end up having a bunch of log messages in these functions. I think I'll leave it as is and log only in the top method

pkg/skaffold/cluster/minikube.go
gsquared94: I'll end up having a bunch of log messages in these functions. I think I'll leave it as is and log only in the top method

gsquared94:
> few nit changes to simply the code

Hopefully addressed all of them.

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+func matchNodeLabel(profileName string) (bool, error) {
+	client, err := k8s.Client()
+	if err != nil {
+		return false, fmt.Errorf("getting Kubernetes client: %w", err)
+	}
+	opts := v1.ListOptions{
+		LabelSelector: fmt.Sprintf("minikube.k8s.io/name=%s", profileName),
+		Limit:         100,
+	}
+	l, err := client.CoreV1().Nodes().List(opts)
```
briandealwis, line 124: This may be a very expensive call if the cluster is far away.

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+func minikubeBinary() (string, error) {
+	execName := "minikube"
+	if found, _ := util.DetectWSL(); found {
+		execName = "minikube.exe"
+	}
+	filename, err := exec.LookPath(execName)
+	if err != nil {
+		return "", errors.New("unable to find minikube executable. Please add it to PATH environment variable")
+	}
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
```
briandealwis, line 109: I wonder when this would happen!

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+	b, err := mkBinaryFunc()
+	if err != nil {
+		return nil, fmt.Errorf("getting minikube executable: %w", err)
+	}
+	return exec.Command(b, arg...), nil
+}
+
+func minikubeBinary() (string, error) {
+	execName := "minikube"
+	if found, _ := util.DetectWSL(); found {
```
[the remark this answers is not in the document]
gsquared94: this is actually [existing code](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/docker/client.go#L181) that I relocated.

pkg/skaffold/cluster/minikube.go
gsquared94: this is actually [existing code](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/docker/client.go#L181) that I just relocated. So I didn't try to delve too deeply into the implementation

gsquared94: I guess it'd happen if you have the minikube binary in `$PWD` and not on `$PATH`.

gsquared94: ya, I think that's the case
```
❯ cat main.go
package main

import (
        "fmt"
        "os"
)

func main() {
        if _, err := os.Stat("minikube"); os.IsNotExist(err) {
                fmt.Println("Not found!")
        } else {
                fmt.Println("Found!")
        }
}
❯ go run main.go
Not found!
❯ touch minikube
❯ go run main.go
Found!
```

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return "", errors.New("unable to find minikube executable. Please add it to PATH environment variable")
+	}
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
+		return "", fmt.Errorf("unable to find minikube executable. File not found %s", filename)
+	}
+	return filename, nil
+}
+
+func matchNodeLabel(profileName string) (bool, error) {
```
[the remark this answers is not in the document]
gsquared94: I sort of used it interchangeably since creating a minikube profile `x` creates a kubeContext `x` too.

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+		return true
+	}
+	_, err := mkBinaryFunc()
+	if err != nil {
+		return false // minikube binary not found
+	}
+
+	if ok, err := matchNodeLabel(kubeContext); err != nil {
+		logrus.Debugf("failed to check minikube node labels: %v", err)
+		return false
```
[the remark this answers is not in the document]
gsquared94: yes, changed.

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+	} else if ok {
+		logrus.Debugf("Detected minikube cluster for context %s due to matched labels", kubeContext)
+		return true
+	}
+
+	if ok, err := matchProfileAndServerURL(kubeContext); err != nil {
+		logrus.Debugf("failed to match minikube profile: %v", err)
+		return false
```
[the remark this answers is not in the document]
gsquared94: yes, changed.

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+	} else if ok {
+		logrus.Debugf("Detected minikube cluster for context %s due to matched profile name or server url", kubeContext)
```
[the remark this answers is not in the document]
gsquared94: done

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/context"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+)
+
+var GetClient = getClient
+
+// To override during tests
+var (
+	mkBinaryFunc  = minikubeBinary
+	k8sConfigFunc = context.GetRestClientConfig
```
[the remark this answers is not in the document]
gsquared94: done

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+	_, err := mkBinaryFunc()
+	if err != nil {
+		return false // minikube binary not found
+	}
+
+	if ok, err := matchNodeLabel(kubeContext); err != nil {
+		logrus.Debugf("failed to check minikube node labels: %v", err)
+		return false
+	} else if ok {
+		logrus.Debugf("Detected minikube cluster for context %s due to matched labels", kubeContext)
```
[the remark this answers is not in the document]
gsquared94: done

pkg/skaffold/cluster/minikube.go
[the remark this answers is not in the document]
gsquared94: I get your point. Changed everywhere to keep consistent with what the parameter actually represents.

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+
+type clientImpl struct{}
+
+func getClient() Client {
+	return clientImpl{}
+}
+
+func (clientImpl) IsMinikube(kubeContext string) bool {
+	// short circuit if context is 'minikube'
+	if kubeContext == constants.DefaultMinikubeContext {
```
nkubala, line 62: i've always wondered if this should be

```golang
if strings.Contains(kubeContext, constants.DefaultMinikubeContext)
```

it wouldn't surprise me if people were creating multiple minikube instances, and naming them `minikube-2` or other random use cases like that. the tradeoff is that this will break users who have remote cluster contexts configured that contain `minikube` in the name....which seems pretty rare. WDYT?

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+	_, err := minikubeBinaryFunc()
+	if err != nil {
+		logrus.Debugf("Minikube cluster not detected: %v", err)
+		return false // minikube binary not found
+	}
+
+	if ok, err := matchNodeLabel(kubeContext); err != nil {
+		logrus.Debugf("failed to check minikube node labels: %v", err)
+	} else if ok {
+		logrus.Debugf("Minikube cluster detected: context %q nodes have minikube labels", kubeContext)
```
nkubala, line 74: wondering if these should be `trace` level logs. i think we might be abusing `debug` level logging in skaffold but i'm not really sure

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func matchNodeLabel(kubeContext string) (bool, error) {
+	client, err := k8s.Client()
+	if err != nil {
+		return false, fmt.Errorf("getting Kubernetes client: %w", err)
+	}
+	opts := v1.ListOptions{
+		LabelSelector: fmt.Sprintf("minikube.k8s.io/name=%s", kubeContext),
+		Limit:         100,
```
nkubala, line 122: why provide a limit here?

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return false, fmt.Errorf("getting kubernetes config: %w", err)
+	}
+	apiServerURL, _, err := rest.DefaultServerURL(config.Host, config.APIPath, schema.GroupVersion{}, false)
+
+	if err != nil {
+		return false, fmt.Errorf("getting kubernetes server url: %w", err)
+	}
+
+	logrus.Debugf("kubernetes server url: %s", apiServerURL)
```
nkubala, line 144: this definitely seems like `trace` level

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,206 @@
...
+	MinikubeExec(arg ...string) (*exec.Cmd, error)
+}
+
+type clientImpl struct{}
+
+func getClient() Client {
+	return clientImpl{}
+}
+
+func (clientImpl) IsMinikube(kubeContext string) bool {
```
nkubala, line 60: what if we cached this in the global config? we're already using kube contexts as keys in the config anyway: we could just add an `isMinikube` field and then check here to short circuit this whole logic.

pkg/skaffold/cluster/minikube.go
gsquared94: After my latest change, it's pretty cheap to check either from the certificate path or the profile names. So I'd rather have false negatives which get caught later than have false positives by making the comparison `contains` over `equals`.

gsquared94:
> Querying the API server is just too expensive — we have some users developing against a cluster on the other side of the world. I thought @tstromberg's suggestion was to load up the kubeconfig files and look at the corresponding cluster's `certificate-authority` to see if it was in ~/.minikube:
> 
> ```
>     certificate-authority: /Users/bdealwis/.minikube/ca.crt
> ```

@briandealwis implemented this approach and removed the label checker. Although the label checker was a very deterministic check and we could have cached the result to not have to run it multiple times. Maybe we can revisit if current set of heuristics don't prove sufficient.

pkg/skaffold/cluster/minikube.go
gsquared94: I changed most of the logs to `trace` level

pkg/skaffold/cluster/minikube.go
gsquared94: no-op since I removed the code.

pkg/skaffold/cluster/minikube.go
gsquared94: I changed most of the logs to trace level

pkg/skaffold/cluster/minikube.go
gsquared94: The checks are quite inexpensive now so we might not need a cache. I'm slightly leaning towards not adding this to the global config because:
- it could pollute the config if the user has a workflow where say they use a random named minikube cluster each time and delete that afterwards.
- it could get stale- if they create minikube `test` cluster, delete that and later create kind or gke cluster called `test`
### 🤷🏾‍♂️

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,215 @@
...
+			if serverURL.Host == fmt.Sprintf("%s:%d", n.IP, n.Port) {
+				return true, nil
+			}
+		}
+	}
+	return false, nil
+}
+
+// minikubePath returns the path to the user's minikube dir
+func minikubePath() string {
```
gsquared94, line 186: copied this function from [minikube](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/localpath/localpath.go#L41)

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,215 @@
...
+func (clientImpl) IsMinikube(kubeContext string) bool {
+	// short circuit if context is 'minikube'
+	if kubeContext == constants.DefaultMinikubeContext {
+		return true
+	}
+	_, err := minikubeBinaryFunc()
+	if err != nil {
+		logrus.Tracef("Minikube cluster not detected: %v", err)
+		return false
+	}
```
tejal29, line 70:
```suggestion
	if _, err := minikubeBinaryFunc();  err != nil {
		logrus.Tracef("Minikube cluster not detected: %v", err)
		return false
	}
```

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,215 @@
...
+
+	ok, err := matchServerURLFor(kubeContext, apiServerURL)
+	if err != nil {
+		return false, fmt.Errorf("checking minikube node url: %w", err)
+	}
+	return ok, nil
+}
+
+func matchServerURLFor(kubeContext string, serverURL *url.URL) (bool, error) {
+	cmd, err := minikubeExec("profile", "list", "-o", "json")
```
tejal29, line 151: the error here should never be encountered, since we first check is executable is present in L66.
I feel this function has too many branches. 
WDYT about removing "minikubeExec" from all `matchServerURLFor`, `matchProfileAndServerURL` and pass it in the function signature.
currently, all the above functions are not used anywhere else.

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		return false, fmt.Errorf("executing minikube command: %w", err)
```
tejal29, line 153: this err is because minikube binary could not be found. consider changing this to "could not fetch minikube binary ..."

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	out, err := util.RunCmdOut(cmd)
+	if err != nil {
+		return false, fmt.Errorf("getting minikube profiles: %w", err)
+	}
+
+	var data data
+	if err = json.Unmarshal(out, &data); err != nil {
+		log.Fatal(fmt.Errorf("failed to unmarshal data: %w", err))
```
tejal29, line 163: could not unmarshal `minikube profile list` output.
Fatal will cause the code to stop. 
Did you mean use to trace instead?

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	for _, v := range data.Valid {
+		if v.Config.Name != kubeContext {
+			continue
+		}
+
+		if v.Config.Driver != "hyperkit" && v.Config.Driver != "virtualbox" {
```
tejal29, line 171: convert this to constants

pkg/skaffold/cluster/minikube.go
gsquared94: my bad, definitely intended to use trace

pkg/skaffold/cluster/minikube.go
gsquared94: Done

pkg/skaffold/cluster/minikube.go
gsquared94: Done

pkg/skaffold/cluster/minikube.go
gsquared94: Done

pkg/skaffold/cluster/minikube.go
gsquared94: removed `err` object instead

pkg/skaffold/cluster/minikube.go
briandealwis: Maybe we should have a KubeContext type instead of passing around string.

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,214 @@
...
+// minikubePath returns the path to the user's minikube dir
+func minikubePath() string {
+	minikubeHomeEnv := os.Getenv("MINIKUBE_HOME")
+	if minikubeHomeEnv == "" {
+		return filepath.Join(homedir.HomeDir(), ".minikube")
+	}
+	if filepath.Base(minikubeHomeEnv) == ".minikube" {
+		return minikubeHomeEnv
+	}
+	return filepath.Join(minikubeHomeEnv, ".minikube")
```
briandealwis, line 193: This doesn't seem right: MINIKUBE_HOME doesn't have to contain a `.minikube` directory.  At least their code [uses patterns like the following to access data](https://github.com/kubernetes/minikube/blob/1a19f8f0a5368f4081a8b3f6f10b2635e043b690/cmd/minikube/cmd/ssh-key.go#L35):
```
filepath.Join(localpath.MiniPath(), "machines", cc.Name, "id_rsa")
```

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (clientImpl) IsMinikube(kubeContext string) bool {
+	// short circuit if context is 'minikube'
+	if kubeContext == constants.DefaultMinikubeContext {
+		return true
+	}
+	if _, err := minikubeBinaryFunc(); err != nil {
+		logrus.Tracef("Minikube cluster not detected: %v", err)
+		return false
+	}
```
briandealwis, line 73: Should this check be performed first?  If we can't find the minikube binary, then it doesn't really matter if `context == 'minikube'`, since we can't run `minikube docker-env`?

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,214 @@
...
+
+	if ok, err := matchClusterCertPath(kubeContext); err != nil {
```
briandealwis, line 75: Should we just fetch the cluster-info and pass it into `matchClusterCertPath()` and into `matchProfileAndServerURL()`?

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,214 @@
...
+		if v.Config.Driver != HyperKit && v.Config.Driver != VirtualBox {
+			// Since node IPs don't match server API for other drivers we assume profile name match is enough.
+			// TODO: Revisit once https://github.com/kubernetes/minikube/issues/6642 is fixed
+			return true, nil
+		}
+		for _, n := range v.Config.Nodes {
+			if serverURL.Host == fmt.Sprintf("%s:%d", n.IP, n.Port) {
+				return true, nil
+			}
+		}
```
briandealwis, line 179: It feels to me like this is inverted:
  - we should check if any of the nodes match the cluster server _regardless of the driver type_.  That way we're insulated from whenever kubernetes/minikube#6642 is fixed.  (At least, my understanding is that the cluster-info's server IP won't match the node IP — my minikube is 127.0.0.1:32780 but the profile lists 172.17.0.2)
  - then see if the minikube profile name matches the kubeContext — this is a heuristic, but unless the user has manually edited their kubeconfig files, it should be true

pkg/skaffold/cluster/minikube.go
gsquared94: It seems to be there to safeguard against someone erroneously adding it so they don't get `~/.minikube/.minikube/`.

pkg/skaffold/cluster/minikube.go
tejal29: Instead of removing the `err` object, can you pass the `minikubeExec` path found it in L66 and avoid calling `minikubeExec` at all.

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,214 @@
...
+
+	ok, err := matchServerURLFor(kubeContext, apiServerURL)
+	if err != nil {
+		return false, fmt.Errorf("checking minikube node url: %w", err)
+	}
+	return ok, nil
+}
+
+func matchServerURLFor(kubeContext string, serverURL *url.URL) (bool, error) {
+	cmd, _ := minikubeExec("profile", "list", "-o", "json")
```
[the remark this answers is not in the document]
gsquared94: yes, we already validate that the binary exists by this point. Creating a `Cmd` doesn't return any other error

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,214 @@
...
+	return clientImpl{}
+}
+
+func (clientImpl) IsMinikube(kubeContext string) bool {
+	// short circuit if context is 'minikube'
+	if kubeContext == constants.DefaultMinikubeContext {
+		return true
+	}
+	if _, err := minikubeBinaryFunc(); err != nil {
+		logrus.Tracef("Minikube cluster not detected: %v", err)
```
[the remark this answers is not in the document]
gsquared94: ya `minikubeBinaryFunc` errors out only when the binary isn't found and all errors are already prefaced with `unable to find minikube executable...`

pkg/skaffold/cluster/minikube.go
```diff
@@ -0,0 +1,214 @@
...
+		return false, fmt.Errorf("getting minikube profiles: %w", err)
+	}
+
+	var data data
+	if err = json.Unmarshal(out, &data); err != nil {
+		return false, fmt.Errorf("failed to unmarshal data: %w", err)
+	}
+
+	for _, v := range data.Valid {
+		if v.Config.Name != kubeContext {
```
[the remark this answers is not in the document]
gsquared94: I'd argue assuming `minikube` context matches a minikube cluster is a much bigger assumption. Since `minikube create cluster` is the obvious minikube cluster creation command which defaults to a `minikube` context we've made that assumption in the current code. So `minikube create cluster -p test` (which is the obvious command to create a named cluster) would default to a `test` context and we should be able to make the same assumption here?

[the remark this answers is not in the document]
gsquared94: Obviously your other comment is correct that I should make this check after checking for the URLs first.

pkg/skaffold/cluster/minikube.go
gsquared94: Fixed in https://github.com/GoogleContainerTools/skaffold/pull/4742

pkg/skaffold/cluster/minikube.go
gsquared94: Fixed in https://github.com/GoogleContainerTools/skaffold/pull/4742

pkg/skaffold/cluster/minikube.go
gsquared94: Fixed in https://github.com/GoogleContainerTools/skaffold/pull/4742

gsquared94:
> A I noticed a few issues that should be fixed.

Fixed them in follow up PR https://github.com/GoogleContainerTools/skaffold/pull/4742

pkg/skaffold/cluster/minikube.go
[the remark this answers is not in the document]
briandealwis: It's very unlikely that I might create a k3d or kind cluster called "minikube".  It's not unlikely that I might create a k3d or kind cluster named "test".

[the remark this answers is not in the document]
gsquared94: Valid point. I'll remove the name check entirely then and just leave the URL checks.

pkg/skaffold/cluster/minikube.go
gsquared94: created https://github.com/GoogleContainerTools/skaffold/issues/4751

Issue #4711, opened 2020-08-24
Skaffold does not work with Helm v3.3


bdudelsack:
Skaffold fails to parse the Helm binary version.

```
helm version --client                                                                                                                                                                                                            
version.BuildInfo{Version:"v3.3", GitCommit:"", GitTreeState:"", GoVersion:"go1.15"}
```

### Expected behavior

Skaffold should parse Helm version without a problem.

### Actual behavior

Skaffold fails to get Helm version because of missing patch version.

```
WARN[0022] deployer cleanup: failed to determine binary version: semver make "3.3": No Major.Minor.Patch elements found 
exiting dev mode because first deploy failed: failed to determine binary version: semver make "3.3": No Major.Minor.Patch elements found
```

### Information

- Skaffold version: v1.13.2
- Operating system: Manjaro Linux

### Steps to reproduce the behavior

1. create a skaffold.yaml with Helm deployer
2. run skaffold dev

PR #4712, opened 2020-08-24
Use ParseTolerant to parse Helm version because of missing patch version

bdudelsack:
Fixes: #4711 

**Description**

Use semVer ParseTolerant version to parse the Helm binary version because it does not follow all semver rules.


bdudelsack:
@googlebot I signed it!

bdudelsack:
Added the tests for helm v3.3. 

The official binaries from helm do output the semver compliant version number, but custom built do not. For example Manjaro has a custom buit helm version in their repository which outputs not semver compliant version number.

kudlatyamroth:
When we can expect release with this fix?

briandealwis:
We're aiming to have a release on Wednesday.  You can use the ["bleeding edge" binaries](https://skaffold.dev/docs/install/) — we typically cut our releases from HEAD.

PR #4708, opened 2020-08-20
Kpt Deployer Render() implementation and tests

felixtran39:
**Related**:  #3904

**Description**
This is the implementation and unit tests for the Render() method of the kpt deployer. This involves running `kpt fn run` on a specified function pipeline.

Along with this, there are also slight modifications to getApplyDir to provide better error handling.

**Follow-up Work**
Kustomization capabilities will come soon for kpt deployer's render stage.

PR #4719, opened 2020-08-26
Add Kustomize Hydration to Kpt Deployer's Render method

felixtran39:
**Related**:  #3904, #4708 

**Description**
This PR builds upon the existing render method for the Kpt Deployer in #4708 by adding the remaining steps for the desired hydration workflow. This includes reading config manifests from the source directory and running `kustomize build` when kustomization configs exist under the specified directory (k.Dir).

Currently, a temporary directory called .pipeline is used to store outputs between different stages of the rendering process. This directory will be replaced by sinkDir when specified by the user.

There is also a slight modification to the kptFnRun method for checking whether FnPath and Image are both provided. This was discussed in further detail in #4708.

### DIFF
--- a/pkg/skaffold/deploy/kpt.go
+++ b/pkg/skaffold/deploy/kpt.go
@@ -49,7 +49,6 @@ type KptDeployer struct {
 	insecureRegistries map[string]bool
 	labels             map[string]string
 	globalConfig       string
-	workingDir         string
 }
 
 func NewKptDeployer(runCtx *runcontext.RunContext, labels map[string]string) *KptDeployer {
@@ -58,7 +57,6 @@ func NewKptDeployer(runCtx *runcontext.RunContext, labels map[string]string) *Kp
 		insecureRegistries: runCtx.GetInsecureRegistries(),
 		labels:             labels,
 		globalConfig:       runCtx.GlobalConfig(),
-		workingDir:         runCtx.GetWorkingDir(),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/skaffold/deploy/kpt.go
```diff
@@ -136,30 +169,75 @@ func (k *KptDeployer) renderManifests(ctx context.Context, _ io.Writer, builds [
...
 	if len(k.Fn.Image) > 0 {
-		if specifiedFnPath {
-			return nil, errors.New("cannot specify both fn-path and image")
-		}
-
 		flags = append(flags, "--image", k.Fn.Image)
+		count++
+	}
+	if count > 1 {
+		return nil, errors.New("cannot specify both fn-path and image")
```
yuwenma, line 237: nit: Only one of "fn-path" and "image" configs can be specified at most.

pkg/skaffold/deploy/kpt.go
```diff
@@ -136,30 +169,75 @@ func (k *KptDeployer) renderManifests(ctx context.Context, _ io.Writer, builds [
...
+	if _, err := util.RunCmdOut(cmd); err != nil {
+		return err
+	}
+
+	deps, err := dependenciesForKustomization(k.Dir)
+	if err != nil {
+		return fmt.Errorf("finding kustomization dependencies: %w", err)
+	}
+
+	// Kustomize build outputs hydrated configs to .pipeline, so the dry configs must be removed.
```
yuwenma, line 207: yeah, i like this approach.

PR #4721, opened 2020-08-26
Move `DetectWSL` function into util package

gsquared94:
Moved `DetectWSL` function into util package. This is a prereq for #4701

### DIFF
--- a/pkg/skaffold/docker/client.go
+++ b/pkg/skaffold/docker/client.go
@@ -207,6 +207,7 @@ func getMinikubeDockerEnv(minikubeProfile string) (map[string]string, error) {
 	}
 
 	if found, _ := util.DetectWSL(); found {
+		// rewrite Unix path to Windows
 		cmd := exec.Command("wslpath", env["DOCKER_CERT_PATH"])
 		out, err := util.RunCmdOut(cmd)
 		if err == nil {
--- a/pkg/skaffold/docker/client.go
+++ b/pkg/skaffold/docker/client.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


pkg/skaffold/docker/client.go
```diff
@@ -221,7 +206,7 @@ func getMinikubeDockerEnv(minikubeProfile string) (map[string]string, error) {
 		env[kv[0]] = kv[1]
 	}
 
-	if found, _ := detectWsl(); found {
+	if found, _ := util.DetectWSL(); found {
 		cmd := exec.Command("wslpath", env["DOCKER_CERT_PATH"])
```
briandealwis, line 210: I know this isn't part of your change but could you just add a comment here to the effect of:
```
// rewrite Unix path to Windows
```

(at least, I'm assuming Minikube in WSL reports a unix-style path?)

pkg/skaffold/util/wsl.go
```diff
@@ -0,0 +1,39 @@
...
+
+func DetectWSL() (bool, error) {
+	if _, err := os.Stat("/proc/version"); err == nil {
+		b, err := ioutil.ReadFile("/proc/version")
+		if err != nil {
+			return false, fmt.Errorf("read /proc/version: %w", err)
+		}
+
+		str := strings.ToLower(string(b))
+		if strings.Contains(str, "microsoft") {
```
briandealwis, line 34: Maybe a comment here
```
// Microsoft changed the case between WSL1 and 2
```

pkg/skaffold/docker/client.go
gsquared94: done

pkg/skaffold/util/wsl.go
gsquared94: done

PR #4728, opened 2020-08-27
Change kpt deployer doc from Beta to Alpha

yuwenma:


yuwenma:
> actually, you need to update this here: [`master`/pkg/skaffold/schema/latest/config.go#L450](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/schema/latest/config.go#L450)
> 
> then run `make generate-schemas` and commit that.

Oh, skaffold generates the user reference from the schema comment. like it!

PR #4735, opened 2020-08-28
Add test for using helm setFiles

MarlonGamez:
This PR only adds a simple test to check that using setFiles in the helm deployer stanza adds the proper flag when running helm commands


MarlonGamez:
Codecov seems to be behaving oddly on this PR. @tejal29 pointed out that it has the same diffs as a PR that she merged earlier #4690

gsquared94 (approved):

PR #4736, opened 2020-08-28
Extending Workflow for Kpt Deployer (accepting additional arguments)

felixtran39:
**Related**:  #3904

**Description**
This contains the implementation and additional test cases allowing users to specify additional arguments when running `kpt fn` or `kpt live` commands with Skaffold.

**User facing changes**
Users will have the option to specify additional arguments in their skaffold.yaml.

**Follow-up Work**
It may be a good idea to add more validation to the config in the future such as validating the values of these arguments to fit certain patterns and catch the error before the command is called.

Before:
```
deploy:
  kpt:
    dir: ...
    fn:
      fnPath: ...
      image: ...
    live:
      apply:
      - ...
      - ...
```

After:
```
deploy:
  kpt:
    dir: ...
    fn:
      fnPath: str
      image: str
      network: str
      networkName: str
      globalScope: str
      mount: []str
    live:
      inventoryID: str
      inventoryNamespace: str
      apply:
        pollperiod: str
        prunePropagationPolicy: str
        pruneTimeout: str
        reconcileTimeout: str
      
```

### DIFF
--- a/docs/content/en/schemas/v2beta7.json
+++ b/docs/content/en/schemas/v2beta7.json
@@ -1639,11 +1639,16 @@
           "description": "a directory to read functions from instead of the configuration directory.",
           "x-intellij-html-description": "a directory to read functions from instead of the configuration directory."
         },
-        "image": {
+        "globalScope": {
           "type": "boolean",
           "description": "sets global scope for functions.",
           "x-intellij-html-description": "sets global scope for functions.",
           "default": "false"
+        },
+        "image": {
+          "type": "string",
+          "description": "an image to be run as a function in lieu of running functions from a directory.",
+          "x-intellij-html-description": "an image to be run as a function in lieu of running functions from a directory."
         },
         "mount": {
           "items": {
@@ -1671,7 +1676,7 @@
         "image",
         "networkName",
         "network",
-        "image",
+        "globalScope",
         "mount"
       ],
       "additionalProperties": false,
@@ -1680,7 +1685,7 @@
     },
     "KptLive": {
       "properties": {
-        "flags": {
+        "apply": {
           "$ref": "#/definitions/KptLiveApply",
           "description": "adds additional configurations for `kpt live apply` commands.",
           "x-intellij-html-description": "adds additional configurations for <code>kpt live apply</code> commands."
@@ -1699,7 +1704,7 @@
       "preferredOrder": [
         "inventoryID",
         "inventoryNamespace",
-        "flags"
+        "apply"
       ],
       "additionalProperties": false,
       "description": "adds additional configurations used when calling `kpt live`.",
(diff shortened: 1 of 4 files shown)


docs/content/en/schemas/v2beta7.json
```diff
@@ -1639,15 +1639,45 @@
...
+        "networkName": {
+          "type": "string",
+          "description": "docker network to run the container in (default \"bridge\").",
+          "x-intellij-html-description": "docker network to run the container in (default &quot;bridge&quot;)."
         }
       },
       "preferredOrder": [
         "fnPath",
-        "image"
+        "image",
```
yuwenma, line 1676: nit: always use alphabetical order in config list especially if the list is large. (if some fields are required, required should always come first).

pkg/skaffold/deploy/kpt.go
```diff
@@ -360,3 +348,82 @@ func getResources(root string) ([]string, error) {
 
 	return files, err
 }
+
+// getFlags returns a list of arguments that the user specified for the `kpt fn run` command.
```
yuwenma, line 352: typo: getKptFnRunArgs

pkg/skaffold/deploy/kpt.go
```diff
@@ -360,3 +348,82 @@ func getResources(root string) ([]string, error) {
...
+		count++
+	}
+
+	if len(k.Fn.Image) > 0 {
+		flags = append(flags, "--image", k.Fn.Image)
+		count++
+	}
+
+	if count > 1 {
+		return nil, errors.New("only one of `fn-path` or `image` configs can be specified at most")
```
yuwenma, line 387: sorry. can we change this to "only one of `fn-path` or `image` may be specified"

pkg/skaffold/deploy/kpt_test.go
```diff
@@ -117,6 +117,42 @@ spec:
...
+		},
+		{
+			description: "user specifies prune propagation policy and prune timeout",
+			cfg: &latest.KptDeploy{
+				Dir:      ".",
+				ApplyDir: "valid_path",
+				Live: latest.KptLive{
+					Apply: latest.KptLiveApply{
+						PrunePropagationPolicy: "Orphan",
+						PruneTimeout:           "2m",
```
yuwenma, line 146: we need two sad path checks here
1. bad timeout value
2. unknown policy

pkg/skaffold/deploy/kpt.go
felixtran39: of course!

docs/content/en/schemas/v2beta7.json
felixtran39: Thanks for the tip!

PR #4738, opened 2020-08-31
Fix #4725: Apply profiles and patches before skaffold config upgrade to latest

### LINKED ISSUE #4725: Cannot define patches for properties that have been changed in newer versions
From a user report on slack, consider the patch:
```
  patches:
     - op: add
        path: /deploy/helm/releases/0
        value:
          name: docs
          namespace: documentation
          chartPath: ./documentation/helm
          remote: false
          values:
            image: docs
```
This is consistent for skaffold API version `skaffold/v1` but in newer versions `value` property of the helm deployer has been renamed to `artifactOverride`. This patch fails to get applied as it's not upgraded to the latest version properly unlike the regular `Pipeline` objects [do](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/schema/util/upgrade_pipelines.go#L31).

gsquared94:
Fixes #4725

I'm applying the profiles **before** upgrading the config from current version to latest. Since the Profiles' `Pipeline` and `Patch` are always defined against the specified config version, this should always work.

Since `Profile` doesn't get upgraded we rely on a switch block to add a new case statement if the `Activation` and `JSONPatch` structs are updated in the future (pretty unlikely though).
```go
switch {
	// Custom activation not supported before v1beta4
	case version.LT(v1b4):
		isActive, isContextActivated = true, false
	// when modifying the `Activation` struct add a case condition here corresponding to the new version and a corresponding `checkActivation` function.
	default:
		var versionedActivation *v1beta4.Activation
		skutil.CloneThroughJSON(activation, &versionedActivation)
		isActive, isContextActivated, err = checkActivation(versionedActivation, opts)
	}
```

```go
switch {
	// Profile patches not supported before v1beta4
	case version.LT(v1b4):
		return fmt.Errorf("profile patches not supported in v%v", v1b4)
	// when modifying the `JSONPatch` struct add a case condition here corresponding to the new version and a corresponding `createPatchOperation` function.
	default:
		patch, err = createPatchOperation(profilePatches.Index(i), buf)
	}
```

### DIFF
--- a/cmd/skaffold/app/cmd/fix.go
+++ b/cmd/skaffold/app/cmd/fix.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"reflect"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -28,6 +29,7 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/validation"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/yaml"
 )
@@ -84,10 +86,25 @@ func fix(out io.Writer, configFile string, toVersion string, overwrite bool) err
 		if err := ioutil.WriteFile(configFile, newCfg, 0644); err != nil {
 			return fmt.Errorf("writing config file: %w", err)
 		}
-		color.Default.Fprintf(out, "New config at version %s generated and written to %s\n", cfg.GetVersion(), opts.ConfigurationFile)
+		color.Default.Fprintf(out, "New config at version %s generated and written to %s\n", cfg.GetVersion(), configFile)
+		warnForProfilePatchUpgrade(cfg, out)
 	} else {
 		out.Write(newCfg)
 	}
 
 	return nil
+}
+
+func warnForProfilePatchUpgrade(config util.VersionedConfig, out io.Writer) {
+	profiles := reflect.ValueOf(config).Elem().FieldByName("Profiles")
+	if !profiles.IsValid() {
+		return
+	}
+
+	for i := 0; i < profiles.Len(); i++ {
+		if !profiles.Index(i).FieldByName("Patches").IsNil() {
+			color.Red.Fprintln(out, "Patch definitions in profiles cannot be upgraded automatically. Please fix them manually.")
+			return
+		}
+	}
 }
--- a/cmd/skaffold/app/cmd/runner.go
+++ b/cmd/skaffold/app/cmd/runner.go
@@ -67,7 +67,7 @@ func createNewRunner(opts config.SkaffoldOptions) (runner.Runner, *latest.Skaffo
 }
 
 func runContext(opts config.SkaffoldOptions) (*runcontext.RunContext, *latest.SkaffoldConfig, error) {
-	parsed, err := schema.ParseConfigAndUpgrade(opts.ConfigurationFile, latest.Version)
+	parsed, err := schema.ParseConfig(opts.ConfigurationFile)
 	if err != nil {
 		if os.IsNotExist(errors.Unwrap(err)) {
 			return nil, nil, fmt.Errorf("skaffold config file %s not found - check your current working directory, or try running `skaffold init`", opts.ConfigurationFile)
@@ -80,12 +80,15 @@ func runContext(opts config.SkaffoldOptions) (*runcontext.RunContext, *latest.Sk
 		return nil, nil, fmt.Errorf("parsing skaffold config: %w", err)
 	}
 
-	config := parsed.(*latest.SkaffoldConfig)
-
-	if err = schema.ApplyProfiles(config, opts); err != nil {
+	if err = schema.ApplyProfiles(parsed, opts); err != nil {
 		return nil, nil, fmt.Errorf("applying profiles: %w", err)
 	}
 
+	if parsed, err = schema.UpgradeConfig(parsed, latest.Version); err != nil {
+		return nil, nil, fmt.Errorf("upgrading config: %w", err)
+	}
+
+	config := parsed.(*latest.SkaffoldConfig)
 	kubectx.ConfigureKubeConfig(opts.KubeConfig, opts.KubeContext, config.Deploy.KubeContext)
 
 	if err := defaults.Set(config); err != nil {
--- a/pkg/skaffold/schema/profiles.go
+++ b/pkg/skaffold/schema/profiles.go
@@ -22,13 +22,16 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/blang/semver"
 	yamlpatch "github.com/krishicks/yaml-patch"
 	"github.com/sirupsen/logrus"
 
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/apiversion"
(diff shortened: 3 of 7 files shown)


gsquared94:
> I think this is a nifty approach!
> 
> It seems like we should add some checks to `fix` to warn that profile patch (when found) will not be updated.
> 
> Changing everything to reference a v1beta4.Activation seems grotty. I'm a reflection-n00b but can you not use reflection to access the structure elements?

The reason I reference `v1beta4` version of activation is so that I can cast it to a `ProfileSelector`
```
// line 77 of pkg/skaffold/schema/profiles.go
s = append(s, activations.Index(j).Addr().Interface().(util.ProfileSelector))
```

If I don't do that then I'll have to define this function for each and every version of `Activation`
`func (a *Activation) IsActive(options map[string]string) (bool, bool, error) ` which seemed like a lot of duplicate code.

The reason I'm doing any of this and not just using reflection is because I don't want to assume that `Activation` struct will never change. I'd be hardcoding field names and such if I go the reflection route all the way.

pkg/skaffold/schema/latest/config.go
```diff
@@ -744,7 +745,7 @@ type Profile struct {
 	// Activation criteria by which a profile can be auto-activated.
 	// The profile is auto-activated if any one of the activations are triggered.
 	// An activation is triggered if all of the criteria (env, kubeContext, command) are triggered.
-	Activation []Activation `yaml:"activation,omitempty"`
+	Activation []v1beta4.Activation `yaml:"activation,omitempty"`
```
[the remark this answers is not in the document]
gsquared94: replied elsewhere what necessitates this change.

briandealwis:
This code is entirely ours, so do the right thing: have a single `IsActive` that uses reflection and add a test to fail detect when `latest.Activation` changes in an unanticipated way.

pkg/skaffold/schema/profiles.go
```diff
@@ -80,45 +90,78 @@ func checkKubeContextConsistency(contextSpecificProfiles []string, cliContext, e
...
+// checkActivations converts each profile `Activation` to a known version and applies the corresponding `checkActivation` function.
+func checkActivations(profiles reflect.Value, version semver.Version, opts cfg.SkaffoldOptions) ([]string, []string, error) {
+	if !opts.ProfileAutoActivation {
+		return nil, nil, nil
+	}
 
-				kubeContext, err := isKubeContext(cond.KubeContext, opts)
+	var activatedProfiles []string
+	var contextSpecificProfiles []string
+	v1b4, _ := apiversion.Parse(v1beta4.Version)
```
briandealwis, line 118: Can we extract this (and the later use) to a constant?  (Will need to add `apiversion.MustParse()`?)

pkg/skaffold/schema/profiles.go
```diff
@@ -80,45 +90,78 @@ func checkKubeContextConsistency(contextSpecificProfiles []string, cliContext, e
...
+				activation := activations.Index(j).Addr().Interface()
+				var isActive, isContextActivated bool
+				var err error
+				switch {
+				// Custom activation not supported before v1beta4
+				case version.LT(v1b4):
+					isActive, isContextActivated = true, false
+				// when modifying the `Activation` struct add a case condition here corresponding to the new version and a corresponding `checkActivation` function.
+				default:
+					var versionedActivation *v1beta4.Activation
```
briandealwis, line 133: Aren't we better off cloning this to `latest.Activation`?  What if v2beta12 adds a new field to `Activation`, or a different activation type?

(I experimented with changing this and other uses to `latest.Activation` and it seemed to work fine for me.)

cmd/skaffold/app/cmd/fix.go
```diff
@@ -84,10 +86,25 @@ func fix(out io.Writer, configFile string, toVersion string, overwrite bool) err
 		if err := ioutil.WriteFile(configFile, newCfg, 0644); err != nil {
 			return fmt.Errorf("writing config file: %w", err)
 		}
-		color.Default.Fprintf(out, "New config at version %s generated and written to %s\n", cfg.GetVersion(), opts.ConfigurationFile)
+		color.Default.Fprintf(out, "New config at version %s generated and written to %s\n", cfg.GetVersion(), configFile)
+		warnForProfilePatchUpgrade(cfg, out)
```
briandealwis, line 90: I think we should check for profile patches regardless of whether it's overwrite or send to stdout?

cmd/skaffold/app/cmd/runner.go
```diff
@@ -67,7 +67,7 @@ func createNewRunner(opts config.SkaffoldOptions) (runner.Runner, *latest.Skaffo
 }
 
 func runContext(opts config.SkaffoldOptions) (*runcontext.RunContext, *latest.SkaffoldConfig, error) {
-	parsed, err := schema.ParseConfigAndUpgrade(opts.ConfigurationFile, latest.Version)
+	parsed, err := schema.ParseConfig(opts.ConfigurationFile)
```
briandealwis, line 70: We should have a comment here to describe the separation.

briandealwis (commented):
I do wonder if we wouldn't be better off trying to:

  1. Parse
  2. Upgrade
  3. Apply Profiles
  4. If the ApplyProfiles failed, then return to the original parse config, apply profiles, and then upgrade, and issue a warning.

pkg/skaffold/schema/profiles.go
gsquared94: So the idea here is that if `v2beta12` changes `Activation` struct then this code would change to:

```go
switch {
	// Custom activation not supported before v1beta4
	case version.LT(v1b4):
		isActive, isContextActivated = true, false
	// when modifying the `Activation` struct add a case condition here corresponding to the new version and a corresponding `checkActivation` function.
	case version.LT(v1b12):
		var versionedActivation *v1beta4.Activation
		skutil.CloneThroughJSON(activation, &versionedActivation)
		isActive, isContextActivated, err = checkActivationv1b4(versionedActivation, opts)
	default:
		var versionedActivation *v2beta12.Activation
		skutil.CloneThroughJSON(activation, &versionedActivation)
		isActive, isContextActivated, err = checkActivationb2v12(versionedActivation, opts)
	}
```

gsquared94: If I changed this to `latest.Activation` like below:
```go
default:
	var versionedActivation *latest.Activation
	skutil.CloneThroughJSON(activation, &versionedActivation)
	isActive, isContextActivated, err = checkActivation(versionedActivation, opts)

func checkActivation(a *latest.Activation, opts cfg.SkaffoldOpts)
```

Then, the next person modifying `Activation` struct might erroneously just modify the `checkActivation` function without handling for the old versions. Doing it the current way also catches this error.

briandealwis: Just so I understand, your way will error if there are new fields or renamed fields, then `CloneThroughJSON` should panic.

gsquared94:
> I do wonder if we wouldn't be better off trying to:
> 
>     1. Parse
> 
>     2. Upgrade
> 
>     3. Apply Profiles
> 
>     4. If the ApplyProfiles failed, then return to the original parse config, apply profiles, and then upgrade, and issue a warning.

Since the patch definitions can be valid it's possible that profile upgrade can be wrong without any error (refer [this](https://kubernetes.slack.com/archives/CABQMSZA6/p1597687772018200) slack thread). Do you think we should do this instead?
- If profile contains patch definitions:
  - Parse
  - Apply Profiles
  - Upgrade
- else
  - Parse&Upgrade
  - Apply Profiles

cmd/skaffold/app/cmd/fix.go
gsquared94: discussed offline with @briandealwis. We can add this message as a comment to the upgraded config.

pkg/skaffold/schema/profiles.go
gsquared94: discussed offline with @briandealwis. This approach requires devs to be aware of the context for this issue, which makes it susceptible to break in the future. Will rethink approach.

PR #4742, opened 2020-08-31
Minikube cluster detection followup

gsquared94:
Fixes for comments on https://github.com/GoogleContainerTools/skaffold/pull/4701 after it got merged


pkg/skaffold/cluster/minikube.go
```diff
@@ -162,21 +130,28 @@ func matchServerURLFor(kubeContext string, serverURL *url.URL) (bool, error) {
 		return false, fmt.Errorf("failed to unmarshal data: %w", err)
 	}
 
+	vmTypeDrivers := map[string]bool{
```
briandealwis, line 133: This needs a comment.  I wonder if it's better to invert this condition to identify non-VM drivers?

pkg/skaffold/cluster/minikube.go
```diff
@@ -162,21 +130,28 @@ func matchServerURLFor(kubeContext string, serverURL *url.URL) (bool, error) {
...
-		if v.Config.Name != kubeContext {
-			continue
+		for _, n := range v.Config.Nodes {
+			if err == nil && serverURL.Host == fmt.Sprintf("%s:%d", n.IP, n.Port) {
+				return true, nil
+			}
 		}
 
-		if v.Config.Driver != HyperKit && v.Config.Driver != VirtualBox {
+		if !vmTypeDrivers[v.Config.Driver] && v.Config.Name == kubeContext {
```
briandealwis, line 150: I have to say that I'm not a fan of embedding rather esoteric knowledge of minikube VM drivers here.  On my personal machine, for example, I use the VMware driver and I *think* this case applies there too.

gsquared94: I've removed all this now

pkg/skaffold/cluster/minikube.go
gsquared94: Removed this too.

pkg/skaffold/cluster/minikube.go
```diff
@@ -162,18 +129,15 @@ func matchServerURLFor(kubeContext string, serverURL *url.URL) (bool, error) {
...
-			return true, nil
-		}
+	for _, v := range data.Valid {
 		for _, n := range v.Config.Nodes {
-			if serverURL.Host == fmt.Sprintf("%s:%d", n.IP, n.Port) {
+			if err == nil && serverURL.Host == fmt.Sprintf("%s:%d", n.IP, n.Port) {
+				// TODO: Revisit once https://github.com/kubernetes/minikube/issues/6642 is fixed
 				return true, nil
 			}
 		}
```
briandealwis, line 143: I think we still want the profileId == kubeContext heuristic:
```golang
 		if v.Config.Name != kubeContext {
 		 		logrus.Infof("Found minikube profile with same name as %q", kubeContext)
 		 		return true, nil
 		}
```

briandealwis, line 143: Nit below: it would be better to rename the `data` struct to `profileList`.

gsquared94: but by your own example: having GKE, kind and minikube clusters all called test, this heuristic would fail. Even though there's a workaround for that, it could break folks on release and as such doesn't seem that backwards compatible. Maybe it'd be better if minikube fixed https://github.com/kubernetes/minikube/issues/6642 so we wouldn't have to go around guessing.

gsquared94: done

Issue #4713, opened 2020-08-24
Support defining build artifact dependency 


gsquared94:
This is a tracking issue for supporting build artifact dependency as a precursor to supporting module dependencies as defined in https://tinyurl.com/skaffold-modules

gsquared94:
Completed

PR #4752, opened 2020-09-01
Support build artifact dependencies

gsquared94:
Related: #4713 

### Description:
- Added `ArtifactDependency` struct to `Artifact`. This will allow us to define a `requires` section in the skaffold config

```yaml
build:
 artifacts:
   - image: leeroy-app
     requires:
       - alias: BASE
         image: simple-go-app
```

- Added skaffold config validations:
   - Artifact dependencies don't result in a circular dependency
   - The same alias or image name aren't repeated in a single dependency list

- Replaced `InSequence` and `InParallel` build controllers with an `InOrder` function in `scheduler.go`
```go
// InOrder builds a list of artifacts in dependency order.
// `concurrency` specifies the max number of builds that can run at any one time. If concurrency is 0, then it's set to the length of the `artifacts` slice.
// Each artifact build runs in its own goroutine. It limits the max concurrency using a buffered channel like a semaphore.
// At the same time, it uses the `artifactChanModel` to model the artifacts dependency graph and to make each artifact build wait for its required artifacts' builds.
func InOrder(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact, buildArtifact ArtifactBuilder, concurrency int) ([]Artifact, error) {
``` 
 ### Next steps:
- Inject the image dependencies into the various builders. Will start with `docker` and `buildpacks` builder where it's obvious that this info will be passed in as `build args` and `env variable` respectively.


pkg/skaffold/schema/validation/validation_test.go
```diff
@@ -750,3 +752,123 @@ func TestValidateLogsConfig(t *testing.T) {
...
+		{
+			description: "artifacts with no dependency",
+			artifactLen: 5,
+		},
+		{
+			description: "artifacts with dependencies",
+			dependency: [][]int{
+				{1, 3, 4},
+				{3, 2},
+				{2, 4},
```
briandealwis, line 772: If I understand `setDependency()` correctly, the first index is the artifact index that you are creating dependencies from.  Why are you reversing 3 and 2?

pkg/skaffold/schema/validation/validation_test.go
```diff
@@ -750,3 +752,123 @@ func TestValidateLogsConfig(t *testing.T) {
...
+				{4, 5},
+			},
+			artifactLen: 5,
+		},
+		{
+			description: "artifacts with circular dependencies",
+			dependency: [][]int{
+				{1, 3, 4},
+				{3, 2},
+				{2, 1},
```
briandealwis, line 782: Ditto reversing 2 and 3

pkg/skaffold/schema/validation/validation.go
```diff
@@ -77,6 +78,81 @@ func validateImageNames(artifacts []*latest.Artifact) (errs []error) {
...
+	for _, artifact := range artifacts {
+		if err := dfs(artifact, visited, make(map[string]bool), m); err != nil {
+			errs = append(errs, err)
+			return
+		}
+	}
+	return
+}
+
+func dfs(artifact *latest.Artifact, visited, marked map[string]bool, artifacts map[string]*latest.Artifact) error {
```
briandealwis, line 103: Credit the algorithm source?

pkg/skaffold/build/scheduler_test.go
```diff
@@ -360,3 +407,18 @@ func setUpChannels(n int) []chan string {
...
+		Build: latest.BuildConfig{
+			BuildType: latest.BuildType{
+				LocalBuild: &latest.LocalBuild{},
+			},
+		},
+	},
+		"temp",
+		true,
+		true,
+		true)
```
briandealwis, line 423: This formatting is a bit weird.  Maybe assign the pipeline to a variable so that the `InitializeState` is one line?

pkg/skaffold/schema/validation/validation_test.go
```diff
@@ -750,3 +752,123 @@ func TestValidateLogsConfig(t *testing.T) {
...
+			ImageName: "artifact1",
+			Dependencies: []*latest.ArtifactDependency{
+				{
+					Alias:     "alias2",
+					ImageName: "artifact2a",
+				},
+				{
+					Alias:     "alias2",
+					ImageName: "artifact2b",
+				},
```
briandealwis, line 840: These take up a lot of space and could be shortened to just:
```suggestion
				{Alias:     "alias2", ImageName: "artifact2a"},
				{Alias:     "alias2", ImageName: "artifact2b"},
```

pkg/skaffold/schema/validation/validation_test.go
```diff
@@ -750,3 +752,123 @@ func TestValidateLogsConfig(t *testing.T) {
...
+			ImageName: "artifact2",
+			Dependencies: []*latest.ArtifactDependency{
+				{
+					Alias:     "alias1",
+					ImageName: "artifact1",
+				},
+				{
+					Alias:     "alias2",
+					ImageName: "artifact1",
+				},
```
briandealwis, line 853: Compress these too

pkg/skaffold/build/scheduler.go
```diff
@@ -56,24 +67,29 @@ func InParallel(ctx context.Context, out io.Writer, tags tag.ImageTags, artifact
...
 		// sync.Map
 		go func(i int) {
+			for _, dep := range awdSlice[i].requiredArtifactChans {
+				// wait for dependency to complete build
+				<-dep
+			}
 			sem <- true
-			runBuild(ctx, w, tags, artifacts[i], results, buildArtifact)
+			runBuild(ctx, w, tags, awdSlice[i].artifact, results, buildArtifact)
+			// closing channel notifies all listeners waiting for this build to complete
```
briandealwis: What happens if a dependency build failed?

pkg/skaffold/build/scheduler.go
```diff
@@ -35,18 +35,29 @@ type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.A
...
-		return nil, nil
+type artifactWithDeps struct {
+	artifact              *latest.Artifact
+	artifactChan          chan interface{}
+	requiredArtifactChans []chan interface{}
+}
+
+// InOrder builds a list of artifacts in dependency order within the specified max concurrency
+func InOrder(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact, buildArtifact ArtifactBuilder, concurrency int) ([]Artifact, error) {
+	chanMap := make(map[string]chan interface{})
```
briandealwis: Document how this works: modelling the dependency graph as a set of channels, etc.

pkg/skaffold/build/scheduler.go
```diff
@@ -35,45 +35,58 @@ type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.A
...
 		outputs[i] = make(chan string, buffSize)
 		r, w := io.Pipe()
 
 		// Run build and write output/logs to piped writer and store build result in
 		// sync.Map
 		go func(i int) {
+			awdSlice[i].waitForDependencies(ctx)
 			sem <- true
-			runBuild(ctx, w, tags, artifacts[i], results, buildArtifact)
+			runBuild(ctx, w, tags, awdSlice[i].artifact, results, buildArtifact)
```
briandealwis, line 88: So what happens if a dependency build failed?

pkg/skaffold/build/scheduler_test.go
```diff
@@ -291,39 +292,73 @@ func TestInParallelConcurrency(t *testing.T) {
...
+			results, err := InOrder(context.Background(), ioutil.Discard, tags, artifacts, builder, test.limit)
 
 			t.CheckNoError(err)
 			t.CheckDeepEqual(test.artifacts, len(results))
 		})
 	}
 }
 
-func TestInParallelForArgs(t *testing.T) {
+func TestInOrderForArgs(t *testing.T) {
```
briandealwis, line 303: Need to have a test where a dependency build fails.

briandealwis (commented):
Looking good!

pkg/skaffold/build/scheduler.go
```diff
@@ -35,45 +35,58 @@ type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.A
...
-		return runInSequence(ctx, out, tags, artifacts, buildArtifact)
+func (a artifactWithDeps) markComplete() {
+	// closing channel notifies all listeners waiting for this build to complete
+	close(a.artifactChan)
+}
+func (a artifactWithDeps) waitForDependencies(ctx context.Context) {
+	for _, dep := range a.requiredArtifactChans {
+		// wait for dependency to complete build
+		select {
+		case <-ctx.Done():
```
nkubala, line 55: do we need to catch `ctx.Cancelled()` here as well? or is that handled somewhere further down/up?

pkg/skaffold/build/scheduler.go
```diff
@@ -35,45 +35,58 @@ type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.A
...
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
 	results := new(sync.Map)
-	outputs := make([]chan string, len(artifacts))
+	outputs := make([]chan string, len(awdSlice))
 
 	if concurrency == 0 {
-		concurrency = len(artifacts)
+		concurrency = len(awdSlice)
```
nkubala, line 74: just trying to work out how this works with nested dependencies. at a certain point, we can't really build concurrently can we? i suppose all top-level dependencies for a given artifact can be built in parallel if (and only if) those don't have dependencies of their own right?

pkg/skaffold/build/scheduler.go
```diff
@@ -35,45 +35,58 @@ type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.A
...
+		// wait for dependency to complete build
+		select {
+		case <-ctx.Done():
+		case <-dep:
+		}
 	}
+}
 
+// InOrder builds a list of artifacts in dependency order within the specified max concurrency
+func InOrder(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact, buildArtifact ArtifactBuilder, concurrency int) ([]Artifact, error) {
```
nkubala, line 62: do we still have `build.InSequence()`? if so we might want to rethink the name since `InSequence` and `InOrder` sound very similar to me.

pkg/skaffold/build/scheduler_test.go
```diff
@@ -341,17 +376,29 @@ func TestInParallelForArgs(t *testing.T) {
...
 	}
 }
 
+func setDependencies(a []*latest.Artifact, d [][]int) {
+	for _, dep := range d {
+		for i := range dep {
+			if i == 0 {
+				continue
+			}
+			a[dep[0]-1].Dependencies = append(a[dep[0]-1].Dependencies, &latest.ArtifactDependency{
```
nkubala, line 395: just by pattern matching with the next line, should this be `a[dep[i]-1]`?

pkg/skaffold/schema/latest/config.go
```diff
@@ -851,6 +854,14 @@ type ArtifactType struct {
 	CustomArtifact *CustomArtifact `yaml:"custom,omitempty" yamltags:"oneOf=artifact"`
 }
 
+// ArtifactDependency describes a specific build dependency for an artifact.
+type ArtifactDependency struct {
+	// Alias is the alias for this image that can be referenced in the builder.
+	Alias string `yaml:"alias" yamltags:"required"`
+	// ImageName is the artifact image name.
+	ImageName string `yaml:"image" yamltags:"required"`
```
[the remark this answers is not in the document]
nkubala: hmm that's an interesting point. `Alias` was originally `Name` which we changed because that was confusing with `ImageName`. but can we reference modules by `ImageName`? if we don't allow the same image name to be used in two distinct artifacts then I don't see why not...

pkg/skaffold/schema/validation/validation.go
```diff
@@ -77,6 +78,81 @@ func validateImageNames(artifacts []*latest.Artifact) (errs []error) {
...
+func validateArtifactCircularDependencies(artifacts []*latest.Artifact) (errs []error) {
+	m := make(map[string]*latest.Artifact)
+	for _, artifact := range artifacts {
+		m[artifact.ImageName] = artifact
+	}
+	visited := make(map[string]bool)
+	for _, artifact := range artifacts {
+		if err := dfs(artifact, visited, make(map[string]bool), m); err != nil {
+			errs = append(errs, err)
+			return
```
nkubala, line 97: since this is part of a larger error collection block that finishes all checks and surfaces all errors, should we not fail fast here and let this run its course?

pkg/skaffold/schema/validation/validation.go
```diff
@@ -77,6 +78,81 @@ func validateImageNames(artifacts []*latest.Artifact) (errs []error) {
...
+				aliasMap[d.Alias] = recorded
+			} else if aliasMap[d.Alias] == unseen {
+				aliasMap[d.Alias] = seen
+			}
+			if namesMap[d.ImageName] == seen {
+				errs = append(errs, fmt.Errorf("invalid build dependency for artifact %q: image name %q repeated", a.ImageName, d.ImageName))
+				namesMap[d.ImageName] = recorded
+			} else if namesMap[d.ImageName] == unseen {
+				namesMap[d.ImageName] = seen
+			}
```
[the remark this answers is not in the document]
nkubala: I believe this is just validating that *for each artifact*, all dependency aliases/image names are unique. I don't think we want to allow two distinct dependencies which share an image name (and are then effectively the same dependency), right?

nkubala (commented):
a few comments but in general this is looking good

pkg/skaffold/schema/latest/config.go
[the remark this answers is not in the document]
gsquared94: I'm in favor of keeping them separate because `alias` is gonna mean an environment variable for docker builder and maybe others. As such there are different naming restrictions like no hyphens, etc. We probably don't want to add new restrictions to `ImageName` field.

pkg/skaffold/build/scheduler.go
gsquared94: So this PR is just parity with existing behavior. So even if a build dependency fails it doesn't fails the downstream builds since the artifacts aren't being cross-referenced anywhere yet. However in my local branch I'm maintaining additional state in this controller which causes subsequent builds to fail since a required artifact build was not found.

pkg/skaffold/build/scheduler.go
gsquared94: yes, I'll add some detail into how I modeled the DAG using a set of channels which allows the max number of parallel builds within `concurrency` for independent artifacts while maintaining required order.

pkg/skaffold/build/scheduler.go
gsquared94: There is no `InSequence` any more. This method simplifies to the `InSequence` builder as soon as `concurrency=1`

pkg/skaffold/schema/validation/validation.go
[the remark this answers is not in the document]
gsquared94: exactly what @nkubala said

pkg/skaffold/schema/validation/validation.go
gsquared94: It's just `Cycle Detection in a Directed Graph using Depth First Search`. I don't think this has any other name.

pkg/skaffold/build/scheduler.go
gsquared94: I'm not sure I follow. `context` here has already been setup with cancellation so when `cancel` gets called in the main go routine it'll close this done channel.
From docs `WithCancel arranges for Done to be closed when cancel is called;`

pkg/skaffold/schema/validation/validation_test.go
```diff
@@ -750,3 +752,123 @@ func TestValidateLogsConfig(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateArtifactCircularDependencies(t *testing.T) {
+	tests := []struct {
```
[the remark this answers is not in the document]
gsquared94: done

pkg/skaffold/schema/validation/validation.go
gsquared94: since this validation is for cycle detection, we'll get stuck on a specific cycle if it exists and it seemed non-trivial to exit that cycle and then try to find the other cycles. So we won't be able to collect all cycle violations at a go which is why I thought to just surface out the first violation and the user can fix it one by one. WDYT?

pkg/skaffold/schema/validation/validation.go
```diff
@@ -77,6 +78,81 @@ func validateImageNames(artifacts []*latest.Artifact) (errs []error) {
...
+		if err := dfs(d, visited, marked, artifacts); err != nil {
+			return err
+		}
+	}
+	marked[artifact.ImageName] = false
+	return nil
+}
+
+// validateArtifactDependencyUniqueness makes sure that artifact dependency aliases and image names are unique for each artifact
+func validateArtifactDependencyUniqueness(artifacts []*latest.Artifact) (errs []error) {
```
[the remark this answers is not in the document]
gsquared94: done

pkg/skaffold/schema/validation/validation.go
gsquared94: added a description with that

pkg/skaffold/build/scheduler.go
gsquared94: added description:
```
//artifactWithDeps models the artifact dependency graph using a set of channels.
// Each artifact has a channel that it closes once it completes building (either success or failure) by calling `markComplete`
// Additionally it has a list of channels for each of its dependencies.
// Calling `waitForDependencies` ensures that all required artifacts' channels have already been closed and as such have finished building.
// This model allows for composing any arbitrary function with dependency ordering.
```
So now additionally piping it through a `sem` buffered channel ensures that at most `sem` number of parallel builds can run if there are no dependencies. But the `waitForDependencies` ensures that all builds need to follow defined order.

PS. go channels are awesome. 🦦

pkg/skaffold/schema/latest/config.go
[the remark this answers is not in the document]
gsquared94: I've made `alias` field optional and default to the value of `image`. This way users can get away with just defining `image` when it'll work fine as an `alias` too.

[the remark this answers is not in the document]
gsquared94: Also reordered to have `ImageName` first as it's the only required field

pkg/skaffold/build/scheduler.go
gsquared94: Added a test case for failure. The current behavior is to just report on the first failed build and I haven't modified it.

pkg/skaffold/build/scheduler_test.go
gsquared94: improved the test

pkg/skaffold/build/scheduler_test.go
gsquared94: done

pkg/skaffold/build/scheduler.go
gsquared94: Done

pkg/skaffold/build/scheduler.go
gsquared94: replied elsewhere

gsquared94:
PTAL again @nkubala and @briandealwis

pkg/skaffold/build/scheduler.go
briandealwis: Maybe a few more details are required as there are two key numbers at play here: 
  1. _concurrency_ which is the maximum number of builds that should be running at any one time, and
  2. _len(artifacts)_ is the number of artifacts to be built

`awdSlice` is of size _len(artifacts)_.  This code creates a goroutine for each element in `awdSlice`.  Each goroutine waits on its dependencies to finish building.  Because our artifacts form a DAG, at least one of the goroutines should be able to start building.

`markComplete()` closes the artifact's channel **which wakes all of the listeners waiting on that channel**.

pkg/skaffold/build/scheduler.go
```diff
@@ -87,6 +105,23 @@ func InParallel(ctx context.Context, out io.Writer, tags tag.ImageTags, artifact
 	return collectResults(out, artifacts, results, outputs)
 }
 
+func artifactWithDepsFor(artifacts []*latest.Artifact) []artifactWithDeps {
```
briandealwis, line 108: This name isn't great — the second time I read this, I thought this might skip artifacts without dependencies.  `makeChannelModel`?

pkg/skaffold/build/scheduler.go
```diff
@@ -35,45 +35,63 @@ type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.A
...
+	buffSize = bufferedLinesPerArtifact
 )
 
-// InParallel builds a list of artifacts in parallel but prints the logs in sequential order.
-func InParallel(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact, buildArtifact ArtifactBuilder, concurrency int) ([]Artifact, error) {
-	if len(artifacts) == 0 {
-		return nil, nil
-	}
+// artifactWithDeps models the artifact dependency graph using a set of channels.
+// Each artifact has a channel that it closes once it completes building (either success or failure) by calling `markComplete`
```
briandealwis, line 42:
```suggestion
// Each artifact has a channel that it closes once it completes building (either success or failure) by calling `markComplete`.  This notifies *all* listeners waiting for this artifact.
```

briandealwis (commented):
What happens / should happen if there when one of the artifacts encounters a build error?

pkg/skaffold/build/scheduler.go
gsquared94: added all these details in either method description or inline comments.

pkg/skaffold/schema/validation/validation_test.go
```diff
@@ -750,3 +752,115 @@ func TestValidateLogsConfig(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+// setDependencies constructs a graph of artifact dependencies using the map as an adjacency list representation of indices in the artifacts array.
+// For example:
+// m = {
+// 	0 : {1, 2},
+//  2 : {3},
```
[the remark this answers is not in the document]
gsquared94: perils of intellij showing tabs and whitespaces with the same width. Fixed!

pkg/skaffold/build/scheduler_test.go
```diff
@@ -341,17 +397,34 @@ func TestInParallelForArgs(t *testing.T) {
...
+			t.CheckDeepEqual(test.err, err, cmp.Comparer(errorsComparer))
 		})
 	}
 }
 
+// setDependencies constructs a graph of artifact dependencies using the map as an adjacency list representation of indices in the artifacts array.
+// For example:
+// m = {
+// 	0 : {1, 2},
+//  2 : {3},
```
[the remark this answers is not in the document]
gsquared94: fixed!

pkg/skaffold/build/scheduler.go
gsquared94: changed struct to `artifactChanModel` and function to `makeArtifactChanModel`. Thanks!

pkg/skaffold/schema/validation/validation.go
gsquared94: @briandealwis WDYT?

briandealwis:
Handling dependency errors can be done in [`runBuild()`](https://github.com/GoogleContainerTools/skaffold/blob/95b9410e47cf351f306c6705df8b043ff17b3f41/pkg/skaffold/build/scheduler.go#L127).

pkg/skaffold/schema/validation/validation.go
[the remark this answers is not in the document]
briandealwis: ^^^ (Github didn't unresolve the conversation when I submitted this comment as part of a review)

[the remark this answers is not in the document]
gsquared94: @briandealwis I think I get your point now.
The dockerfile:
```
FROM $builder as builder1
RUN foo1
FROM $builder as builder2
RUN foo2
FROM bar
COPY --from=builder1 foobar
```
can be simplified into:
```
FROM $builder1
RUN foo1
FROM $builder2
RUN foo2
FROM bar
COPY --from=$builder1 foobar
```
where $builder1 and $builder2 both reference the same image that $builder referenced earlier.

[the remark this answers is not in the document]
gsquared94: There's no compelling reason for this validation, I just thought that it didn't seem valuable to permit multiple aliases referencing the same image. But I'll remove it anyway.

gsquared94:
> What happens / should happen if there when one of the artifacts encounters a build error?

Currently it just runs all other builds even if a required build has failed since I haven't introduced the dependency `alias`es into the builders yet. Once the builders actually read the `ArtifactDependency` and search for their images it'll fail with a `required image not found` sort of error. 

I'm thinking about changing the current behavior where we just report about the first failed build (see [here](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/parallel.go#L133)) and instead `string.Join(errors, "\n")` and report all the failed builds. If I did that it'd probably be even better to report the errors in the order that they were built. I'll add details about that in the follow up PR. WDYT?

gsquared94:
> Handling dependency errors can be done in [`runBuild()`](https://github.com/GoogleContainerTools/skaffold/blob/95b9410e47cf351f306c6705df8b043ff17b3f41/pkg/skaffold/build/scheduler.go#L127).

yes, that's the intention.

pkg/skaffold/schema/validation/validation.go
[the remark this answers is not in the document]
gsquared94: Removed this extra validation in latest commit

pkg/skaffold/schema/validation/validation.go
```diff
@@ -77,6 +78,75 @@ func validateImageNames(artifacts []*latest.Artifact) (errs []error) {
...
+		if err := dfs(artifact, visited, make(map[string]bool), m); err != nil {
+			errs = append(errs, err)
+			return
+		}
+	}
+	return
+}
+
+// dfs runs a Depth First Search algorithm for cycle detection in a directed graph
+func dfs(artifact *latest.Artifact, visited, marked map[string]bool, artifacts map[string]*latest.Artifact) error {
```
gsquared94, line 104: TODO: add more testcases for the cycle detection

gsquared94:
Found some issues with the `artifactChanModel`. Will reopen PR after testing for a bit.

pkg/skaffold/schema/validation/validation.go
gsquared94: done

gsquared94:
closing this for now, will reopen after incorporating changes from https://github.com/GoogleContainerTools/skaffold/pull/4794 when that gets approved.

pkg/skaffold/build/scheduler.go
gsquared94: I get it know.
```
case <-ctx.Done():
    if ctx.Err() == ctx.Cancelled():
      ...
```

I'll add that. Thanks!

PR #4761, opened 2020-09-03
Add pkg/skaffold/ui package

MarlonGamez:
**Description**
Adds functions to be used to create groups of progress spinners for skaffold's future UI changes

Follow up PR using this package to add progress indicators during skaffold executions

How this would be used:
The ui package uses a singleton pattern. For each section of skaffold, a new progress group would be made, replacing the current group. A progress group will hold 1 or more items that are acively showing progress to the user.

For example, in the build phase of a skaffold, a new progress group would be made with `ui.NewProgressGroup()`
Spinners can then be added to the group using `spinner := ui.AddNewSpinner(prefix, name, style)`. Giving `nil` as the style will use the default spinner style of `⠋`, `⠙`, `⠹`, `⠸`, `⠼`, `⠴`, `⠦`, `⠧`, `⠇`, `⠏`
To mark a spinner as complete, use `spinner.Increment()`
At the end of a skaffold phase, call `ui.Wait()`. This will signal the progress group to wait on all of the bars still in progress.

Example: 
```golang
ui.NewProgressGroup()
topLevel := ui.AddNewSpinner("", "Building...", nil)

skaffold.Build()

topLevel.Increment()
ui.Wait()
```

In skaffold.Build func:
```golang
for i, artifact := range artifacts {
  spinner := ui.AddNewSpinner("  ", artifact.ImageName, nil)
}
```

This would end up looking like 
```
⠋ Building...
  ✓ image1
  ⠋ image2
```
and when done:
```
✓ Building...
  ✓ image1
  ✓ image2
```


tejal29:
@MarlonGamez Can you please provide snapshots of how this will look on some fake data?

MarlonGamez:
@tejal29 I updated the PR with some more information detailed the usage and what it would end up looking like :)

hack/generate-statik.sh
```diff
@@ -38,11 +38,21 @@ elif ! [ -x "$(command -v ${LICENSES})" ]; then
     popd
 fi
 
+# hack for github.com/vbauerster/mpb begin
```
tejal29, line 41: This sounds a little fishy. if the script fails before L52 then file will not be renamed.
Look into deleting this file in  exit trap

briandealwis (commented):
So my concern here is that we shoudn't be baking in `pkg/skaffold/ui` into `pkg/skaffold/*`.  We also need to make this optional or disable-able as IntelliJ's console view doesn't support advanced ANSI sequences.

Here are my quick suggestions:
  - I think we should move `pkg/skaffold/ui` into `cmd/skaffold/ui`
  - I think we should create a `pkg/skaffold/progress` package with a UI-independent `ProgressManager` and `ProgressReporter` interfaces
     - `ProgressReporter` will offer methods like `SetTotal(int)`, `Increment()`, and `Done()` (and potentially `NewChild(name, int)` to carve off some amount of work for a new progress reporter, if we need it)
     - `ProgressManager` would offer methods like `NewChild(description string) ProgressReporter`
  - We'd then have mappings for `ProgressManager` and `ProgressReporter` that use `mpb`.
  - We'd then modify `cmd/skaffold/app/skaffold` to use these `mpb`-based `ProgressManager`.  We could also have null variants that do nothing.

The idea is that we should be able to swap out the implementations — even on the fly.

(There may be some other package that defines ProgressManager/Reporter-like interfaces.)

hack/generate-statik.sh
```diff
@@ -38,11 +38,21 @@ elif ! [ -x "$(command -v ${LICENSES})" ]; then
...
+if [ -f vendor/github.com/vbauerster/mpb/v5/UNLICENSE ]; then
+   mv vendor/github.com/vbauerster/mpb/v5/UNLICENSE vendor/github.com/vbauerster/mpb/v5/LICENSE
+fi
+
 echo "Collecting licenses"
 cd ${DIR}/..
 ${LICENSES} save github.com/GoogleContainerTools/skaffold/cmd/skaffold --save_path="${TMP_DIR}/skaffold-credits"
 chmod -R u+w "${TMP_DIR}/skaffold-credits"
 
+# hack for github.com/vbauerster/mpb end
```
tejal29, line 51: Can you check what is the output of skaffold credits on your branch?
I tried checking the output of skaffold credits and there is no license generation for "github.com/vbauerster" project.
If you don't have LICENSE file, does the tool error out?

pkg/skaffold/ui/spinner_filler.go
```diff
@@ -0,0 +1,48 @@
...
+	if len(style) == 0 {
+		style = mpb.DefaultSpinnerStyle
+	}
+	filler := &spinnerFiller{
+		frames: style,
+	}
+	return filler
+}
+
+// To fulfill the implementation of mpb.BarFiller interface
```
tejal29, line 42: Can you add describe what `mpb.BarFiller.Fill` merthod does.

pkg/skaffold/ui/spinner_filler.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/vbauerster/mpb/v5"
+	"github.com/vbauerster/mpb/v5/decor"
+)
+
+type spinnerFiller struct {
+	frames []string
+	count  uint
+}
+
+// NewSpinnerFiller constructs a mpb.BarFiller, to be used with *Progress.Add(...) method.
```
tejal29, line 31:
```suggestion
// NewSpinnerFiller creates a mpb.BarFiller, to be used with *Progress.Add(...) method.
```

pkg/skaffold/ui/spinner_filler.go
```diff
@@ -0,0 +1,48 @@
...
+package ui
+
+import (
+	"io"
+
+	"github.com/vbauerster/mpb/v5"
+	"github.com/vbauerster/mpb/v5/decor"
+)
+
+type spinnerFiller struct {
```
tejal29, line 26: Add description, `spinnerFiller implements a mpb.BarFiller interface` Can you also describe what a `barFiller` is used for?

pkg/skaffold/ui/spinner_filler.go
```diff
@@ -0,0 +1,48 @@
...
+	frames []string
+	count  uint
```
tejal29, line 28: what is this count?  the number of characters in output line or spin icon?

pkg/skaffold/ui/spinner_filler.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	filler := &spinnerFiller{
+		frames: style,
+	}
+	return filler
+}
+
+// To fulfill the implementation of mpb.BarFiller interface
+func (s *spinnerFiller) Fill(w io.Writer, _ int, _ decor.Statistics) {
+	frame := s.frames[s.count%uint(len(s.frames))]
```
tejal29, line 44: this will compute which spin icon to show right?

pkg/skaffold/ui/spinner_filler.go
```diff
@@ -0,0 +1,48 @@
...
+
+	io.WriteString(w, frame)
```
tejal29, line 46: This will write the spin icon on the writer?

pkg/skaffold/ui/spinner_filler.go
```diff
@@ -0,0 +1,48 @@
...
+		style = mpb.DefaultSpinnerStyle
+	}
+	filler := &spinnerFiller{
+		frames: style,
+	}
+	return filler
+}
+
+// To fulfill the implementation of mpb.BarFiller interface
+func (s *spinnerFiller) Fill(w io.Writer, _ int, _ decor.Statistics) {
```
tejal29, line 43: Please add comments to describe the Fill method. 
IIUC, it computes the spin frame and writes it to the output?

tejal29 (changes requested):
nit and some verification on the licenses part.

MarlonGamez:
@briandealwis This would for sure be optional/disabled by default, I was planning on hiding things behind a flag for now. 

Thanks for the suggestions, your ideas sound great :) I'll work on refactoring/implementing an interface like that

pkg/skaffold/ui/progress.go
```diff
@@ -0,0 +1,51 @@
...
+
+package ui
+
+import (
+	"github.com/vbauerster/mpb/v5"
+	"github.com/vbauerster/mpb/v5/decor"
+)
+
+var (
+	current *mpb.Progress
```
tejal29, line 25: Please make this in a class with `Increment` `Wait` as methods.

pkg/skaffold/ui/ui.go
```diff
@@ -0,0 +1,31 @@
...
+	"io"
+	"io/ioutil"
+)
+
+var (
+	out io.Writer = ioutil.Discard
+)
+
+// SetOutput sets the destination for messages from the ui package
+func SetOutput(w io.Writer) {
```
tejal29, line 29: How this ui package going to be used? Would there be a way to not get progress bars and raw skaffold output?
IDEs like VScode and intellij use skaffold stream as is. This might not render well on the IDE side untill they switch to constructing UI based on event apis.

```
ui.NewProgressGroup()
topLevel := ui.AddNewSpinner("", "Building...", nil)

skaffold.Build()

topLevel.Increment()
ui.Wait()
```

MarlonGamez: Under the changes that Brian suggested, a ProgressManager would be set at the beginning of the skaffold execution. By default this would be a null ProgressManager, and so added function calls would do nothing. If using the implementation that I'm adding here, then the function calls would spawn the progress indicators that I've made

pkg/skaffold/ui/spinner_filler.go
MarlonGamez: This count the current icon, I'll rename it for clarity 👍🏽

vendor/modules.txt
```diff
@@ -538,7 +550,7 @@ golang.org/x/oauth2/jwt
 # golang.org/x/sync v0.0.0-20200317015054-43a5402ce75a
 ## explicit
 golang.org/x/sync/errgroup
-# golang.org/x/sys v0.0.0-20200523222454-059865788121
+# golang.org/x/sys v0.0.0-20200810151505-1b9f1253b3ed
```
gsquared94, line 553: should we use a version of the new modules that work with our existing version of `golang.org/x/sys` to minimize changes?

MarlonGamez: I'll try and see if there is a version that supports the same version of `golang.org/x/sys` 👍🏼

MarlonGamez:
I'm closing this as I'm working on a new PR that will implement a lot of the feedback given by Brian

PR #4487, opened 2020-07-17
Clarify usage of ArtifactOverrides, ImageStrategy

TBBle:
**Related**: #500, specifically addressing [this comment](https://github.com/GoogleContainerTools/skaffold/issues/500#issuecomment-659705231)

**Description**
ArtifactOverrides was incorrectly claiming it needed "values file" syntax, i.e. structured YAML, but it actually takes `--set-string` syntax, i.e. dotted path.

Also tied `ArtifactOverrides` directly to `ImageStrategy`, since the two work together.

**Follow-up Work**

I didn't check that `ImageStrategy` is meaningfully documented, so it might need a closer look too.

PR #4774, opened 2020-09-11
Fix issue with schema doc changes

MarlonGamez:
**Description**
Just a fix for schema docs. Looks like an issue was created with the merging of #4487, as it modified latest/config.go when v2beta6 was unreleased, but was merged after v2beta8 was generated


tejal29:
is this a revert? we should be able to merge doc changes without make a schema release.

tejal29 (approved):

PR #4776, opened 2020-09-11
Bump http-proxy from 1.18.0 to 1.18.1 in examples/react-reload and integration/examples/react-reload

MarlonGamez:
**Description**
Replacing #4770, as it doesn't bump the version in integration/examples/react-reload

Bumps the version of http-proxy from 1.18.0 to 1.18.1

### DIFF
--- a/examples/react-reload/app/package-lock.json
+++ b/examples/react-reload/app/package-lock.json
@@ -4144,9 +4144,9 @@
       "dev": true
     },
     "http-proxy": {
-      "version": "1.18.0",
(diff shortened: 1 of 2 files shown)


briandealwis (approved):
I think if you incorporated #4770's commit into this PR, instead of re-creating it, then merging this PR would close #4770 too.

MarlonGamez:
@briandealwis I believe dependabot also automatically closes its PR when it notices that the version is up to date

briandealwis:
That's nifty.  If only dependant would were smart enough to patch both `examples/` and `integration/examples/` in the same PR…

Issue #4455, opened 2020-07-09
skaffold deploy to support -t for tags


matti:
skaffold.yaml:

```
apiVersion: skaffold/v2alpha2
kind: Config
build:
  artifacts:
    - image: first
      context: components/first
    - image: second
      context: components/second
deploy:
  kustomize:
    path: kustomize
```

If I have already built the images I need to provide `--image` for every artifact:

```
skaffold deploy \
  -d my.registry.example.com \
  --images "first:mytag" \
  --images "second:mytag"
```

So if `skaffold deploy` supported `-t` like other commands, then this would become just:

```
skaffold deploy \
  -d my.registry.example.com \
  -t mytag
```

and the `skaffold deploy ...` command wouldn't need to know the artifacts.

matti:
Somewhat related to #922

gsquared94:
This seems like a reasonable ask. However there seem to already be a few different ways of passing info from build to deploy. We'll probably need to be a bit more consistent.

tejal29:
We decided to add this feature. Removing the discuss label

PR #4778, opened 2020-09-13
skaffold deploy -t flag

anshlykov:
Fixes: #4455 

**Description**
This PR adds the `--tag` flag to the deploy command. If the '--tag' flag is present, then the tags for all artifacts are overridden. 

Perhaps the deploy command should only use the new tag for artifacts for which the image tag is not explicitly specified. Let me know if I need to change anything.

### DIFF
--- a/cmd/skaffold/app/cmd/deploy.go
+++ b/cmd/skaffold/app/cmd/deploy.go
@@ -27,6 +27,7 @@ import (
 	"github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/flags"
 	"github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/tips"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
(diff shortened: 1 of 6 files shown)


cmd/skaffold/app/cmd/deploy.go
```diff
@@ -88,6 +89,21 @@ func getArtifactsToDeploy(out io.Writer, fromFile, fromCLI []build.Artifact, art
 	deployed = build.MergeWithPreviousBuilds(fromCLI, deployed)
 	deployed = build.MergeWithPreviousBuilds(fromFile, deployed)
 
+	if opts.CustomTag != "" {
```
MarlonGamez, line 92: Can you leave a comment before this section just to clarify that this piece of code is applying the tags? Or move it into a function like `applyCustomTag(deployed)` whose name will clarify

MarlonGamez (changes requested):

MarlonGamez (approved):
Thanks for making the change :) we're having some issues with CI right now so I'm gonna approve this and I'll merge it once we get stuff working again

PR #4787, opened 2020-09-15
gotest.sh should use -race like `make test`

briandealwis:
**Description**
`make test` runs `go test -race`, but `hack/gotest.sh` (used to run our tests on Travis) does not.  We've had a race in FakeAPIClient for a while.


briandealwis:
We seem to be using `-race` a bit haphazardly.

PR #4799, opened 2020-09-17
[kpt deployer] Improve skaffold.yaml docs.

yuwenma:
**Description**
* kpt.Dir is required, adding the explanation on the default config.
* kpt.applyDir is removed from top-tier config.
* Rewrote the kpt.Live struct to clarify the config scope and which configs should be considered together.
* Improve the kpt option docstring for skaffold users.

Changed the doc in skaffold.yaml for kpt deployer.


MarlonGamez (approved):
LGTM!

PR #4800, opened 2020-09-17
Pin autoprefixer version at 9.8.6 for docs

MarlonGamez:
Our Travis builds have been broken because of a recent update to autoprefixer which isn't yet supported by postcss-cli (https://github.com/postcss/autoprefixer/issues/1359#issuecomment-694292709).
Pinning the version of autoprefixer to the version before will fix this

### DIFF
--- a/deploy/webhook/Dockerfile
+++ b/deploy/webhook/Dockerfile
@@ -40,7 +40,8 @@ FROM node:10.15.3-stretch as runtime_deps
 ENV FIREBASE_TOOLS_VERSION 7.13.1
 RUN npm install -g firebase-tools@${FIREBASE_TOOLS_VERSION} postcss-cli
 WORKDIR /app/docs
-RUN npm install autoprefixer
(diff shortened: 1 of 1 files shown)


briandealwis (approved):
I wonder why we install autoprefixer locally instead of globally like we do postcss-cli?

MarlonGamez:
@briandealwis I think [this article](https://nodejs.org/en/blog/npm/npm-1-0-global-vs-local-installation/#:~:text=globally%20%E2%80%94%2D%20This%20drops%20modules%20in,in%20the%20current%20working%20directory.) gives some insight into this.

>1. If you’re installing something that you want to use in your program, using require('whatever'), then install it locally, at the root of your project.
>2. If you’re installing something that you want to use in your shell, on the command line or something, install it globally, so that its binaries end up in your PATH environment variable.

In this case I think postcss-cli is providing a binary that is called out to by some part of the program, so it needs to exist in the path and is installed globally

PR #4810, opened 2020-09-21
Skip cluster integration tests if changes only in docs folder

### LINKED ISSUE #4808: Travis CI can skip integration tests if only changes are in `docs` directory
Travis CI can skip the kind and k3d integration tests if only changes are in the `docs` directory. It should continue running the other tests.

gsquared94:
Fixes #4808 
Skips integration tests in kind and k3d clusters when the only change in a PR is in the `docs` folder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ _integration_test: &integration_test
   os: linux
   language: minimal
   before_install:
-    #Skip integration tests if changes are only in the `docs` directory. This only works for a pull-request based development workflow like Skaffold's. 
+    # Skip integration tests for `docs`-only changes (only works for PR-based dev workflows like Skaffold's).
     - git diff --name-only "$TRAVIS_BRANCH"... | grep -v '^docs/' || travis_terminate 0
   cache:
     directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,8 @@ _integration_test: &integration_test
   os: linux
   language: minimal
   before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,6 +14,8 @@ branches:
 _integration_test: &integration_test
   os: linux
   language: minimal
+  before_install:
+    - git diff --name-only "$TRAVIS_BRANCH"...HEAD | grep -v '^docs/' || travis_terminate 0
```
briandealwis, line 18: Nit: you can drop the ending `HEAD`: it's implied if not included.  
It's worth a comment that this only works for pull-request branches which is the normal development case for Skaffold since we all work from forks.

briandealwis (commented):
Great idea!

.travis.yml
gsquared94: done

.travis.yml
```diff
@@ -14,6 +14,9 @@ branches:
 _integration_test: &integration_test
   os: linux
   language: minimal
+  before_install:
+    #Skip integration tests if changes are only in the `docs` directory. This only works for a pull-request based development workflow like Skaffold's.
```
briandealwis, line 18: Nit: insert space after the comment.  WDYT of this shorter alternative:
```suggestion
    # Skip integration tests for `docs`-only changes (only works for PR-based dev workflows like Skaffold's).
```

briandealwis (approved):
Is it possible to do this for the unit test too?

Issue #4826, opened 2020-09-28
skaffold debug when using helm does not work when helm chart renders more than one manifest


doriandekoning:
### Expected behavior
When running `skaffold debug` with a helm deployment (supported by https://github.com/GoogleContainerTools/skaffold/pull/4732)  which renders multiple manifests the deployment is modified and a debugging port is forwarded to my machine.
### Actual behavior
Debugging does not work, without any error or warning.
### Information

- Skaffold version: version: `v1.14.0-40-gf4bd6387c`
- Operating system: Darwin
- Contents of skaffold.yaml: 

```yaml
apiVersion: skaffold/v2beta5
kind: Config
build:
  tagPolicy:
    sha256: {}
  artifacts:
    - image: "[redacted]"
      custom:
        buildCommand: "make docker_build_service SERVICE={{.SERVICE}}"
  local: {}
deploy:
  helm:
    releases:
      - name: "{{.SERVICE}}"
        chartPath: "weave/service"
        remote: true
        namespace: [redacted]
        upgradeOnChange: true
        recreatePods: true
        valuesFiles:
          - "../[recacted]/default_svc_values.yaml"
          - "../[redacted[/services/{{.SERVICE}}.yaml"
        setValues:
          skaffold: "true"
        artifactOverrides:
          image: "[redacted]"
        setValueTemplates:
          name: "{{.SERVICE}}"
          fullnameOverride: "{{.SERVICE}}"
          hosts[0]: "{{.SERVICE}}.{{.CLUSTER_HOST}}"
          grpctranscoding.apidescriptor_b64: "{{.API_DESCRIPTOR}}"


```

### Steps to reproduce the behavior

1. a clonable repository with the sample skaffold project
2. `skaffold debug  -f skaffold.yaml --cache-artifacts=false --verbosity=info --port-forward`

I've fixed this issue and will open an MR with a (I believe trival) fix.

PR #4827, opened 2020-09-28
fix: skaffold debug does not work for helm deployments with multiple …

doriandekoning:
**Fixes:** #4826 
**Related**:https://github.com/GoogleContainerTools/skaffold/pull/4732 
**Merge before/after**: _None_
**Description**
This mr splits the rendered helm resources in the `skaffold filter` command on `---` which separates resources. Currently only the first resource is modified, this causes the debug command not to work.  In this mr the input on `filter` (the rendered helm resources) is split on `---` and to each resource the debugging transformations are applied.

I'd like to add some tests but there are currently no meaning full tests for the filter command (as far as I can see) which I can extend.

### DIFF
--- a/cmd/skaffold/app/cmd/filter.go
+++ b/cmd/skaffold/app/cmd/filter.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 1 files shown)


doriandekoning:
@googlebot I signed it!

cmd/skaffold/app/cmd/filter.go
```diff
@@ -64,7 +65,12 @@ func runFilter(ctx context.Context, out io.Writer, debuggingFilters bool, buildA
 		if err != nil {
 			return err
 		}
-		manifestList := manifest.ManifestList([][]byte{bytes})
+
+		var manifestList manifest.ManifestList
+		for _, manifestStr := range strings.Split(string(bytes), "---") {
```
briandealwis, line 70: This will catch `---` where it occurs within a string.  It needs to be preceded and followed by newlines.

briandealwis (commented):
Thanks for catching this — I'll fix this up and add some further tests.

cmd/skaffold/app/cmd/filter.go
doriandekoning: Good catch

PR #4828, opened 2020-09-28
Support dependencies between build artifacts - 1

gsquared94:
Related: #4713 
Part 1 of several PRs.

Implements sections of the [design](https://github.com/GoogleContainerTools/skaffold/blob/master/docs/design_proposals/artifact-dependencies.md):
- Config schema
- Config validation
- Build controller(only parity with existing code)
- [Build logs reporting](https://github.com/GoogleContainerTools/skaffold/blob/master/docs/design_proposals/artifact-dependencies.md#build-logs-reporting)

The current set of changes do not modify the output behavior.

Couldn't reopen the previous [PR](https://github.com/GoogleContainerTools/skaffold/pull/4752) as I force pushed the branch.


pkg/skaffold/build/scheduler.go
```diff
@@ -87,19 +74,29 @@ func InParallel(ctx context.Context, out io.Writer, tags tag.ImageTags, artifact
...
 
-func runBuild(ctx context.Context, cw io.WriteCloser, tags tag.ImageTags, artifact *latest.Artifact, results *sync.Map, build ArtifactBuilder) {
-	event.BuildInProgress(artifact.ImageName)
+func runBuild(ctx context.Context, cw io.WriteCloser, tags tag.ImageTags, a *artifactChanModel, results *sync.Map, build ArtifactBuilder) {
+	defer cw.Close()
+	err := a.waitForDependencies(ctx)
+	event.BuildInProgress(a.artifact.ImageName)
+	if err != nil {
+		event.BuildFailed(a.artifact.ImageName, err)
+		results.Store(a.artifact.ImageName, err)
```
briandealwis, line 83: These should use `errors.Errorf("failure upstream: %w", err)` ?  Otherwise this artifact is reporting the dependency's error as its own?

pkg/skaffold/build/model.go
```diff
@@ -0,0 +1,97 @@
...
+// artifactChanModel models the artifact dependency graph using a set of channels.
+// Each artifact has a status struct that has success and a failure channel which it closes once it completes building by calling either markSuccess or markFailure respectively.
+// This notifies all listeners waiting for this artifact of a successful or failed build.
+// Additionally it has a reference to the channels for each of its dependencies.
+// Calling `waitForDependencies` ensures that all required artifacts' channels have already been closed and as such have finished building before the current artifact build starts.
+type artifactChanModel struct {
+	artifact                 *latest.Artifact
+	artifactStatus           status
+	requiredArtifactStatuses []status
+	concurrencySem           chan bool
```
briandealwis, line 35: This describe what it is, not so much what it's for.  `buildToken`?

pkg/skaffold/build/model.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+
+	if c == 0 {
+		c = len(artifacts)
+	}
+	// sem is a channel that will allow up to `c` concurrent operations.
+	sem := make(chan bool, c)
+
+	var acmSlice []*artifactChanModel
```
briandealwis, line 88: Google style recommends [dropping type names](http://goto.google.com/go-style/decisions#repetitive-with-type).

gsquared94: done

pkg/skaffold/build/scheduler.go
gsquared94: the error is already formatted as `fmt.Errorf("failed to build required artifact: %q", depStatus.imageName)`. It's indexed on the artifact name so the build failure logged is `couldn't build "artifact1": failed to build required artifact: "artifact2"`

gsquared94: The event is
```json
{
  "artifact": "artifact1",
  "status": "Failed",
  "err": "failed to build required artifact: \"artifact2\""
}
```

gsquared94: I think this is sufficient.

pkg/skaffold/build/model.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (a *artifactChanModel) markFailure() {
+	// closing channel notifies all listeners waiting for this build that it failed
+	close(a.artifactStatus.failure)
+	<-a.concurrencySem
+}
+func (a *artifactChanModel) waitForDependencies(ctx context.Context) error {
+	defer func() {
+		a.concurrencySem <- true
```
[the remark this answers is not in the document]
gsquared94: modeled better around a `scheduler` object. PTAL.

pkg/skaffold/build/model.go
gsquared94: renamed

pkg/skaffold/build/scheduler.go
briandealwis: 👍  I had meant to delete this comment.

pkg/skaffold/build/ordered.go
```diff
@@ -35,49 +35,51 @@ type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.A
...
-		go func(i int) {
-			sem <- true
-			runBuild(ctx, w, tags, artifacts[i], results, buildArtifact)
-			<-sem
-
+		// Create a goroutine for each element in acmSlice. Each goroutine waits on its dependencies to finish building.
+		// Because our artifacts form a DAG, at least one of the goroutines should be able to start building.
+		go func(a *latest.Artifact) {
+			// Run build and write output/logs to piped writer and store build result in sync.Map
+			scheduler.run(ctx, a,
```
briandealwis, line 65: I don't think separating this out with these helper functions improves the comprehensibility.  Could you try moving most of this function and `readOutputAndWriteToChannel()` onto the `scheduler`?

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,63 @@
...
+
+// scheduler handles build ordering and concurrency.
+type scheduler struct {
+	dagMap map[string]artifactDAG
+	// sem is a buffered channel that will allow up to `c` concurrent operations when its buffer size is `c`.
+	sem chan bool
+}
+
+func newScheduler(artifacts []*latest.Artifact, concurrency int) scheduler {
+	if concurrency == 0 {
```
tejal29, line 33: For current version, i thought we would always have non-parallel builds?

pkg/skaffold/build/ordered.go
```diff
@@ -35,49 +35,51 @@ type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.A
...
+					event.BuildComplete(a.ImageName)
+					ar := Artifact{ImageName: a.ImageName, Tag: tag}
+					results.Store(ar.ImageName, ar)
+					w.Close()
+				},
+				func(err error) {
+					event.BuildFailed(a.ImageName, err)
+					results.Store(a.ImageName, err)
+					w.Close()
+				})
```
tejal29, line 80: All of this could happen in the scheduler.run function. Is there a reason why you are passing this as functions? The same function is executed for every artifact.

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,63 @@
...
+		concurrency = len(artifacts)
+	}
+	return scheduler{
+		dagMap: makeArtifactDAG(artifacts),
+		sem:    make(chan bool, concurrency),
+	}
+}
+
+func (c *scheduler) run(ctx context.Context, a *latest.Artifact, build func() (string, error), onSuccess func(tag string), onError func(err error)) {
+	dag := c.dagMap[a.ImageName]
```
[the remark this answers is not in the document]
tejal29: or artifact

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,63 @@
...
+	err := dag.waitForDependencies(ctx)
```
tejal29, line 44: This should happen in a forever loop right until all dependent artifacts are build?

tejal29 (commented):
This is great. Few thoughts, 
1. Add a test case for requires to make sure this works or instructions for reviewers to check it out. 
2. I feel, the `waitForDependencies` should be called in a forever loop. Not sure how it is monitoring if all dependent artifacts are build.

pkg/skaffold/build/model.go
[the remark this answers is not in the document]
gsquared94: I undid the `scheduler` and only addressed you initial suggestion of keeping everything in `InOrder`. I'll do a bigger refactor of how we handle collecting the logs and build result in a subsequent PR.

pkg/skaffold/build/ordered.go
gsquared94: I removed this.

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"context"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+)
+
+// scheduler handles build ordering and concurrency.
+type scheduler struct {
+	dagMap map[string]artifactDAG
+	// sem is a buffered channel that will allow up to `c` concurrent operations when its buffer size is `c`.
```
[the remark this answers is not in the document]
gsquared94: no op

pkg/skaffold/build/scheduler.go
gsquared94: we'll support concurrent builds as we do currently- the build logs are printed sequentially, though independent builds can run in parallel.

pkg/skaffold/build/ordered.go
gsquared94: > 1. Add a test case for requires to make sure this works or instructions for reviewers to check it out.

there are success and failure test cases in `scheduler_test.go`

> 2. I feel, the `waitForDependencies` should be called in a forever loop. Not sure how it is monitoring if all dependent artifacts are build.

it tries reading from every channel that represents each of its dependencies. the single for loop over the dependencies slice will complete only when all of these channels are closed. So it doesn't need a forever loop.

gsquared94:
> 1. Add a test case for requires to make sure this works or instructions for reviewers to check it out.

there are success and failure test cases in `scheduler_test.go`

> 2. I feel, the `waitForDependencies` should be called in a forever loop. Not sure how it is monitoring if all dependent artifacts are build.

it tries reading from every channel that represents each of its dependencies. the single for loop over the dependencies slice will complete only when all of these channels are closed. So it doesn't need a forever loop.

pkg/skaffold/build/scheduler.go
gsquared94: ptal at `InOrder` function as I removed this class. Also my explanation below

pkg/skaffold/build/ordered.go
```diff
@@ -35,49 +35,51 @@ type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.A
...
-		return runInSequence(ctx, out, tags, artifacts, buildArtifact)
-	}
-
-	var wg sync.WaitGroup
-	defer wg.Wait()
-
+// InOrder builds a list of artifacts in dependency order.
+// `concurrency` specifies the max number of builds that can run at any one time. If concurrency is 0, then it's set to the length of the `artifacts` slice.
+// Each artifact build runs in its own goroutine. It limits the max concurrency using a buffered channel like a semaphore.
+// At the same time, it uses the `artifactDAG` to model the artifacts dependency graph and to make each artifact build wait for its required artifacts' builds.
```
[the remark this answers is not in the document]
gsquared94: done

pkg/skaffold/build/ordered.go
```diff
@@ -35,49 +35,51 @@ type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.A
...
-
-	if len(artifacts) == 1 || concurrency == 1 {
-		return runInSequence(ctx, out, tags, artifacts, buildArtifact)
-	}
-
-	var wg sync.WaitGroup
-	defer wg.Wait()
-
+// InOrder builds a list of artifacts in dependency order.
+// `concurrency` specifies the max number of builds that can run at any one time. If concurrency is 0, then it's set to the length of the `artifacts` slice.
```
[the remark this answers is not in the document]
gsquared94: done

pkg/skaffold/build/model.go
```diff
@@ -0,0 +1,100 @@
...
+
+// waitForDependencies returns an error if any dependency build fails
+func (a *buildNode) waitForDependencies(ctx context.Context) error {
+	for _, dep := range a.dependencies {
+		// wait for required builds to complete
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-dep.failure:
+			return fmt.Errorf("failed to build required artifact: %q", dep.imageName)
```
nkubala, line 58: how do the errors look when this happens? will this surface the actual error encountered from the build?

pkg/skaffold/build/scheduler.go
```diff
@@ -35,88 +35,85 @@ type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.A
...
 	// Print logs and collect results in order.
-	return collectResults(out, artifacts, results, outputs)
+	return collectResults(out, s.artifacts, &s.results, s.outputs)
 }
 
-func runBuild(ctx context.Context, cw io.WriteCloser, tags tag.ImageTags, artifact *latest.Artifact, results *sync.Map, build ArtifactBuilder) {
-	event.BuildInProgress(artifact.ImageName)
+func (s *scheduler) build(ctx context.Context, cw io.WriteCloser, tags tag.ImageTags, i int) {
+	defer cw.Close()
+	defer s.wg.Done()
```
nkubala, line 88: `defer` calls go onto a function-specific stack, so this will call `s.wg.Done()` before `cw.Close()`. I think they should be reversed no?

gsquared94: yes, you're right. Fixed.

pkg/skaffold/build/model.go
gsquared94: The existing code returns the error from the first element in the input `[]*latest.Artifact` that fails which necessarily isn't in build order. So you might get `couldn't build "artifact1": failed to build required artifact: "artifact2"` as the only error without getting the error for `artifact2` failure.

I have another PR in the works which will change this behavior to instead return all build failures so then you'd get:
```
couldn't build "artifact2": some error message
couldn't build "artifact1": failed to build required artifact: "artifact2"
```

pkg/skaffold/schema/latest/config.go
```diff
@@ -884,6 +887,16 @@ type ArtifactType struct {
 	CustomArtifact *CustomArtifact `yaml:"custom,omitempty" yamltags:"oneOf=artifact"`
 }
 
+// ArtifactDependency describes a specific build dependency for an artifact.
+type ArtifactDependency struct {
+	// ImageName is the artifact image name.
+	ImageName string `yaml:"image" yamltags:"required"`
+	// Alias is a token that will be replaced with the image reference in the builder definition files.
```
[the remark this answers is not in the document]
gsquared94: done

pkg/skaffold/schema/latest/config.go
```diff
@@ -884,6 +887,16 @@ type ArtifactType struct {
 	CustomArtifact *CustomArtifact `yaml:"custom,omitempty" yamltags:"oneOf=artifact"`
 }
 
+// ArtifactDependency describes a specific build dependency for an artifact.
+type ArtifactDependency struct {
+	// ImageName is the artifact image name.
```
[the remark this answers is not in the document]
gsquared94: done

pkg/skaffold/build/scheduler.go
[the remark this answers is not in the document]
gsquared94: done

pkg/skaffold/build/model.go
```diff
@@ -0,0 +1,84 @@
...
+func (a *artifactDAG) markSuccess() {
+	// closing channel notifies all listeners waiting for this build that it succeeded
+	close(a.status.success)
+}
+
+func (a *artifactDAG) markFailure() {
+	// closing channel notifies all listeners waiting for this build that it failed
+	close(a.status.failure)
+}
+func (a *artifactDAG) waitForDependencies(ctx context.Context) error {
```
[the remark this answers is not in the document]
gsquared94: done

pkg/skaffold/build/scheduler.go
gsquared94: to give more detail: we set the default concurrency to 1 in `defaults.go`. So concurrency `0` is the user's way of explicitly saying `max concurrency`. This is the existing code behavior and it works for the new scenario of build dependencies too. 
Currently in master, the output logs are displayed in the order specified in the input `latest.Artifact` slice. I think that was mostly a bug since if the concurrency is 1 and order of execution is artifact4, artifact2, artifact3 and artifact1 then we won't see any logs until artifact1 starts building and then they are all shown together.
I fix this in [another](https://github.com/GoogleContainerTools/skaffold/pull/4843) PR where the logs are printed in the actual build order. But I keep the current behavior of displaying concurrently run and concluded build logs together one after the other immediately. This behavior can later be substituted with a progress bar or multi drop-down writers to show all concurrent builds together but that is left as a future work item.
These are described in the design doc [here](https://github.com/GoogleContainerTools/skaffold/blob/master/docs/design_proposals/artifact-dependencies.md#build-logs-reporting)

tejal29:
> there are success and failure test cases in `scheduler_test.go`

 I meant an actual example so we can try out running `skaffold dev` and verify 
- how the output looks, 
- see artifacts built in order, 
- hit cntrl C in the middle and make sure there no go subroutines leaked.
- also try devloop
All things a user can do :)

tejal29:
> it tries reading from every channel that represents each of its dependencies. the single for loop over the dependencies slice will complete only when all of these channels are closed. So it doesn't need a forever loop.

ah got it. the for loop is blocked until the channel is closed. I am wondering how it will look when you are 2 or more dependencies for an artifact. If the second build fails before first completes, would the `waitForDependencies` get notified right away?

e.g.
```
  - requires: ["first-long-build", "second-short-build"]
```

gsquared94:
> > it tries reading from every channel that represents each of its dependencies. the single for loop over the dependencies slice will complete only when all of these channels are closed. So it doesn't need a forever loop.
> 
> ah got it. the for loop is blocked until the channel is closed. I am wondering how it will look when you are 2 or more dependencies for an artifact. If the second build fails before first completes, would the `waitForDependencies` get notified right away?
> 
> e.g.
> 
> ```
>   - requires: ["first-long-build", "second-short-build"]
> ```

It won't be notified immediately. It is blocked on each build sequentially.

tejal29:
> 
> It won't be notified immediately. It is blocked on each build sequentially.

I could be easy to poll for every artifact and cancel polling when we see first failure. WDYT?

```
resultMap := sync.Map{}
pollCtx, cancelPolling := context.WithCancel(ctx)
wg.Add(len( a.dependencies))
// Keep polling for all artifacts via separate go routines.
for _, dep := range a.dependencies {
     go func(ctx context.Context) {
        select {
               case <- ctx.Done()
                     wg.Done()
                     resultMap.Put(dep.ImageName, fmt.Errorf("build cancelled artifact: %q ", dep.ImageName))
                case <-dep.failure:
                       wg.Done()
		       resultMap.Put(dep.ImageName, fmt.Errorf("failed to build required artifact: %q", dep.ImageName))
                       cancelPolling() // Cancel all polling for all other artifacts
                case <-dep.success:
                       wg.Done()
         }
     }(pollCtx) 
}
// listen for user hitting control C.
for {
    select {
        // Keep listening for user hit cancel
	case <-ctx.Done():
               // if user hits cancel, stop all the polling goroutines.
                cancelPolling()
                return ctx.Err()
         case <- wg.Wait()
             // go through all syncmap results and then return all artifacts that failed which were not cancelled.
        }
   }
}()

}
```

gsquared94:
> > It won't be notified immediately. It is blocked on each build sequentially.
> 
> I could be easy to poll for every artifact and cancel polling when we see first failure. WDYT?
> 
> ```
> resultMap := sync.Map{}
> pollCtx, cancelPolling := context.WithCancel(ctx)
> wg.Add(len( a.dependencies))
> // Keep polling for all artifacts via separate go routines.
> for _, dep := range a.dependencies {
>      go func(ctx context.Context) {
>         select {
>                case <- ctx.Done()
>                      wg.Done()
>                      resultMap.Put(dep.ImageName, fmt.Errorf("build cancelled artifact: %q ", dep.ImageName))
>                 case <-dep.failure:
>                        wg.Done()
> 		       resultMap.Put(dep.ImageName, fmt.Errorf("failed to build required artifact: %q", dep.ImageName))
>                        cancelPolling() // Cancel all polling for all other artifacts
>                 case <-dep.success:
>                        wg.Done()
>          }
>      }(pollCtx) 
> }
> // listen for user hitting control C.
> for {
>     select {
>         // Keep listening for user hit cancel
> 	case <-ctx.Done():
>                // if user hits cancel, stop all the polling goroutines.
>                 cancelPolling()
>                 return ctx.Err()
>          case <- wg.Wait()
>              // go through all syncmap results and then return all artifacts that failed which were not cancelled.
>         }
>    }
> }
> 
> }
> ```

This makes me think that I should cancel **all** builds if one fails(I was not doing that, rather waiting for concurrent ones to complete and as such there could be multiple failures). I've reworked this to use an error group, so now any build failure cancels all running (concurrent) and pending(maybe dependent) ones as well.

PTAL.

pkg/skaffold/build/result.go
```diff
@@ -0,0 +1,128 @@
...
+		lines <- scanner.Text()
+	}
+	close(lines)
+}
+
+func newLogWriter(capacity int) logWriter {
+	return &logWriterImpl{capacity: capacity, messages: make(chan chan string, capacity)}
+}
+
+// resultStore stores the results of each artifact build.
```
briandealwis, line 99: I'd expect a _result-store_ to store all aspects of a build result.  `builtArtifacts`?

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/event"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+)
+
+type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.Artifact, tag string) (string, error)
+
+type scheduler struct {
+	artifacts       []*latest.Artifact
+	artifactBuilder ArtifactBuilder
```
briandealwis, line 36: Swap these so artifacts is next to nodes?
```suggestion
	artifactBuilder ArtifactBuilder
	artifacts       []*latest.Artifact
```

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	// print output for all artifact builds in order
+	s.logger.PrintInOrder(out)
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+	return formatResults(s.artifacts, s.results)
+}
+
+// formatResults returns the build results in the order of `latest.Artifacts` in the input slice.
```
briandealwis, line 76: Would t make sense to have this as a method on `resultsStore`?

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,144 @@
...
+	err := n.waitForDependencies(ctx)
+	release := s.sem.acquire()
+	defer release()
+	w, _ := s.logger.GetWriter()
+	defer w.Close()
+
+	if err == context.Canceled {
+		event.BuildCanceled(a.ImageName)
+		return err
+	}
```
briandealwis, line 102: This should happen immediately after `waitForDependencies()` (move in front of line 94)

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (s *scheduler) build(ctx context.Context, tags tag.ImageTags, i int) error {
+	n := s.nodes[i]
+	a := s.artifacts[i]
+
+	err := n.waitForDependencies(ctx)
+	release := s.sem.acquire()
+	defer release()
+	w, _ := s.logger.GetWriter()
```
briandealwis, line 96: `GetWriter()` returns a `nil` writer.  Is that ok in `getBuildResults()`?

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,144 @@
...
+	return builds, nil
+}
+
+func (s *scheduler) build(ctx context.Context, tags tag.ImageTags, i int) error {
+	n := s.nodes[i]
+	a := s.artifacts[i]
+
+	err := n.waitForDependencies(ctx)
+	release := s.sem.acquire()
+	defer release()
```
briandealwis, line 95: My concern in this function is that we're holding the build semaphore when issuing events, which is generally not recommended (since we could be triggering arbitrary code — including something blocking — while holding onto the semaphore).

This is a case for managing the release explicitly.

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,144 @@
...
+	if err == context.Canceled {
+		event.BuildCanceled(a.ImageName)
+		return err
+	}
+
+	event.BuildInProgress(a.ImageName)
+	if err != nil {
+		event.BuildFailed(a.ImageName, err)
+		return err
+	}
```
briandealwis, line 108: This err case should also be in front of line 94; it makes sense to pair BuildInProgress with BuildFailed, I guess, so you could just issue both within this if clause.

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,144 @@
...
+
+	s := newScheduler(artifacts, artifactBuilder, concurrency)
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	return s.run(ctx, out, tags)
+}
+
+func getBuildResult(ctx context.Context, cw io.Writer, tags tag.ImageTags, artifact *latest.Artifact, build ArtifactBuilder) (string, error) {
```
briandealwis, line 137: Rename as this is not a getter but triggers computation.  `performBuild`?

But really this could be inlined.

briandealwis (commented):
This is looking really good.  I think it would also make sense to move your example into here and incorporate it as an integration test.

pkg/skaffold/build/scheduler.go
briandealwis: I guess we're safe since the events are all dispatched from a separate go function.

pkg/skaffold/build/scheduler.go
gsquared94: done

pkg/skaffold/build/scheduler.go
gsquared94: I couldn't do this before as I wasn't cancelling log aggregator. But I've made that change so fixed this too.

pkg/skaffold/build/scheduler.go
gsquared94: fixed

pkg/skaffold/build/scheduler.go
gsquared94: I couldn't do this before as I wasn't cancelling log aggregator. But I've made that change so fixed this too.

pkg/skaffold/build/result.go
```diff
@@ -0,0 +1,128 @@
...
+
+const bufferedLinesPerArtifact = 10000
+
+// For testing
+var (
+	buffSize = bufferedLinesPerArtifact
+)
+
+// logWriter provides an interface to create an output writer for each artifact build and later aggregate the logs in build order.
+// The order of output is not guaranteed between multiple builds running concurrently.
```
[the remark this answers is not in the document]
gsquared94: renamed. 
> (And how does this interact with the builder logging in logfile.go?)

```
Building [app2]...
 - writing logs to /var/folders/8z/1cllh9c94nddcym3gwvtp8mw00qzkt/T/skaffold/build/app2.log
Building [app3]...
 - writing logs to /var/folders/8z/1cllh9c94nddcym3gwvtp8mw00qzkt/T/skaffold/build/app3.log
Building [app1]...
 - writing logs to /var/folders/8z/1cllh9c94nddcym3gwvtp8mw00qzkt/T/skaffold/build/app1.log
```

It works correctly IMO. `logfile.go` wraps the `ArtifactBuilder`, so for concurrently running builds we get concurrently written log files. The prompt above happens in build order.

gsquared94:
> This is looking really good. I think it would also make sense to move your example into here and incorporate it as an integration test.

I wanted to add the integration test after making the dependency reference working. WDYT?

briandealwis:
Having the IT inside this PR helps ensure the PR is working.

pkg/skaffold/build/result.go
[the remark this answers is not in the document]
briandealwis: But why does logAggregator need to buffer lines in memory when they're being written out to disk?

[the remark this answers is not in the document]
gsquared94: it buffers these lines:
```
Building [app2]...
 - writing logs to /var/folders/8z/1cllh9c94nddcym3gwvtp8mw00qzkt/T/skaffold/build/app2.log
Building [app3]...
 - writing logs to /var/folders/8z/1cllh9c94nddcym3gwvtp8mw00qzkt/T/skaffold/build/app3.log
Building [app1]...
 - writing logs to /var/folders/8z/1cllh9c94nddcym3gwvtp8mw00qzkt/T/skaffold/build/app1.log
```
and not the actual logs which are written directly. We obviously need to buffer these.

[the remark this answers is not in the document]
gsquared94: `logfile.go` when active [swaps out](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/logfile.go#L46) the buffered out writer with the file writer directly prior to build.

briandealwis (commented):
If we can bring in an integration test, I think this is ready to go.

pkg/skaffold/build/result.go
```diff
@@ -0,0 +1,152 @@
...
+func (ba *builtArtifactsImpl) GetTag(a *latest.Artifact) (string, error) {
+	v, ok := ba.m.Load(a.ImageName)
+	if !ok {
+		return "", fmt.Errorf("could not find build result for image %s", a.ImageName)
+	}
+	switch t := v.(type) {
+	case string:
+		return t, nil
+	default:
+		return "", fmt.Errorf("could not find build result for image %s", a.ImageName)
```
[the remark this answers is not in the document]
gsquared94: it'll never happen. I just added for completion.

gsquared94:
IT added. PTAL @briandealwis, @tejal29

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,131 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	return s.run(ctx, out, tags)
+}
+
+func performBuild(ctx context.Context, cw io.Writer, tags tag.ImageTags, artifact *latest.Artifact, build ArtifactBuilder) (string, error) {
+	color.Default.Fprintf(cw, "Building [%s]...\n", artifact.ImageName)
+	tag, present := tags[artifact.ImageName]
+	if !present {
```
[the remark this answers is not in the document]
gsquared94: this case should never happen; but it - checking for cases that'll never happen - seems to be done in a lot of places so I'll prefer to not change it.

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,131 @@
...
+
+type ArtifactBuilder func(ctx context.Context, out io.Writer, artifact *latest.Artifact, tag string) (string, error)
+
+type scheduler struct {
+	artifacts       []*latest.Artifact
+	nodes           []buildNode // size len(artifacts)
+	artifactBuilder ArtifactBuilder
+	logger          logAggregator
+	results         builtArtifacts
+	sem             countingSemaphore
```
tejal29, line 40: Optional rename `sem` to `concurrencyCount` to describe what this semaphore tracks or represents.

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,131 @@
...
+	event.BuildInProgress(a.ImageName)
+
+	w, err := s.logger.GetWriter()
+	if err != nil {
+		event.BuildFailed(a.ImageName, err)
+		return err
+	}
+	defer w.Close()
+
+	finalTag, err := performBuild(ctx, w, tags, s.artifacts[i], s.artifactBuilder)
```
tejal29, line 97: s.artifacts[i] can be replaced by a

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,131 @@
...
+
+	for i := range s.artifacts {
+		i := i
+
+		// Create a goroutine for each element in dag. Each goroutine waits on its dependencies to finish building.
+		// Because our artifacts form a DAG, at least one of the goroutines should be able to start building.
+		// Wrap in an error group so that all other builds are cancelled as soon as any one fails.
+		g.Go(func() error {
+			return s.build(gCtx, tags, i)
+		})
```
tejal29, line 66: instead of passing around tags and then performing the checks later, would it make sense to change this build to just pass the artifact and node. 
```
func (s *scheduler) build(ctx context.Context, actifact, node, tag tag.ImageTag) error {

}
```
the check for `tag` exists can be moved from `performBuild` to this go subroutine.
```
for i := range s.artifacts {
   a := s.artifacts[i]
   n := nodes[i]
   g.Go(func() error {
     tag, present := tags[a.ImageName]
     if !present {
       return fmt.Errorf(...)
     }
     return s.build(gCtx, a, n, tag)
    })

}
```

pkg/skaffold/build/result.go
```diff
@@ -0,0 +1,152 @@
...
+
+type logAggregatorImpl struct {
+	messages   chan chan string
+	size       int
+	capacity   int
+	countMutex sync.Mutex
+}
+
+func (l *logAggregatorImpl) GetWriter() (io.WriteCloser, error) {
+	if err := l.checkCapacity(); err != nil {
```
tejal29, line 54: this condition would never happen as channel capacity is size of artifacts. Can we drop this condition to make things simpler.

pkg/skaffold/build/result.go
```diff
@@ -0,0 +1,152 @@
...
+	Record(a *latest.Artifact, tag string)
+	GetTag(a *latest.Artifact) (string, error)
+	GetArtifacts(s []*latest.Artifact) ([]Artifact, error)
+}
+
+func newArtifactsStore() builtArtifacts {
+	return &builtArtifactsImpl{m: new(sync.Map)}
+}
+
+type builtArtifactsImpl struct {
```
tejal29, line 121: is this just a result store? i dont understand wht "buildArtifactsImpl" stands for

tejal29:
> IT added. PTAL @briandealwis, @tejal29

Looks really great. Some nit changes to get rid of if-else branches. 

Looking forward to try this out now

pkg/skaffold/build/result.go
briandealwis: Poor Gaurav, I suggested changing it from resultsStore as it didn't store any results :-)

pkg/skaffold/build/result.go
```diff
@@ -0,0 +1,152 @@
...
+type logAggregator interface {
+	// GetWriter returns an output writer tracked by the logAggregator
+	GetWriter() (io.WriteCloser, error)
+	// PrintInOrder prints the output from each allotted writer in build order.
+	// It blocks until the instantiated capacity of io writers have been all allotted and closed, or the context is cancelled.
+	PrintInOrder(ctx context.Context, out io.Writer)
+}
+
+type logAggregatorImpl struct {
+	messages   chan chan string
```
tejal29, line 47: Intresting use of chain of channels.  saw this video @Sajmani https://youtu.be/QDDwwePbDtw?t=999

pkg/skaffold/build/result.go
tejal29: sorry. :( upto Gaurav.

tejal29:
testing notes:
Example dependency graph
``` 
 image 3 -> image2  -> image1
  image 4 
 ```
1. checkout & make
```
hub pr checkout 4828
make
cd testdata/integration/build-dependencies
```

2. run normal case
```
time ../../../out/skaffold build -d gcr.io/tejal-test

Generating tags...
 - image1 -> gcr.io/tejal-test/image1:latest
 - image2 -> gcr.io/tejal-test/image2:latest
 - image3 -> gcr.io/tejal-test/image3:latest
 - image4 -> gcr.io/tejal-test/image4:latest
Checking cache...
 - image1: Not found. Building
 - image2: Not found. Building
 - image3: Not found. Building
 - image4: Not found. Building
<Followed by logs for docker build.>
12.74s user 7.35s system 30% cpu 1:06.72 total
```
Observations: 
a. No information on if build dependency graph is being used and builds are running in parallel. 
b. No indication if they are running concurrently.
c. I can infer from the logs, build3 got build first. However i did have to scroll up.
d. sequence is correct: image 3, image 4, image 2, image 1

2. Run with concurrency 3.
```
time ../../../out/skaffold build -d gcr.io/tejal-test -p concurrency-3 --cache-artifacts=false

13.07s user 7.83s system 43% cpu 47.798 total
```
Observations: 
* Definitely saw the build completed faster.
* Sequence is correct: image 3, image 4, image 2, image 1


3. Run with a build failure in image 4 which should cancel dependent artifacts image2 and image 3 build as well as already scheduled image3.
```
time ../../../out/skaffold build -d gcr.io/tejal-test --cache-artifacts=false -p concurrency-3
Generating tags...
 - image1 -> gcr.io/tejal-test/image1:latest
 - image2 -> gcr.io/tejal-test/image2:latest
 - image3 -> gcr.io/tejal-test/image3:latest
 - image4 -> gcr.io/tejal-test/image4:latest
Building [image3]...

Building [image4]...
Sending build context to Docker daemon 

docker build: error during connect: Post "https://127.0.0.1:32774/v1.40/build?buildargs=%7B%22FAIL%22%3A%220%22%2C%22SLEEP%22%3A%229%22%7D&cachefrom=null&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile&forcerm=1&labels=null&memory=0&memswap=0&networkmode=&nocache=1&rm=0&shmsize=0&t=gcr.io%2Ftejal-test%2Fimage4%3Alatest&target=&ulimits=null&version=": creating docker context: getting relative tar paths: file pattern [foo1] must match at least one file

6.09s user 3.75s system 152% cpu 6.443 total
```
Observations:
1. All builds were stopped and the failure for corresponding artifact is seen.
2. There is a line regarding image 3 artifact.  instead of closing the channel, should we first write something similar to 
" cancelling due to one or more builds failed"
```
Building [image3]...
Build cancelled due to other failures seen in other artifacts
Building [image4]...
Sending build context to Docker daemon 

docker build: error during connect: Post "https://127.0.0.1:32774/v1.40/build?buildargs=%7B%22FAIL%22%3A%220%22%2C%22SLEEP%22%3A%229%22%7D&cachefrom=null&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile&forcerm=1&labels=null&memory=0&memswap=0&networkmode=&nocache=1&rm=0&shmsize=0&t=gcr.io%2Ftejal-test%2Fimage4%3Alatest&target=&ulimits=null&version=": creating docker context: getting relative tar paths: file pattern [foo1] must match at least one file
```

4. Run with a build failure in image 1.
```
time ../../../out/skaffold build -d gcr.io/tejal-test --cache-artifacts=false -p concurrency-3
Generating tags...
 - image1 -> gcr.io/tejal-test/image1:latest
 - image2 -> gcr.io/tejal-test/image2:latest
 - image3 -> gcr.io/tejal-test/image3:latest
 - image4 -> gcr.io/tejal-test/image4:latest
Building [image4]...
Sending build context to Docker daemon   2.56kB
...
Building [image3]...
Sending build context to Docker daemon   2.56kB
....
Successfully tagged gcr.io/tejal-test/image3:latest

Building [image2]...
...
Building [image1]...
Sending build context to Docker daemon 

docker build: error during connect: Post "https://127.0.0.1:32774/v1.40/build?buildargs=%7B%22FAIL%22%3A%220%22%2C%22SLEEP%22%3A%221%22%7D&cachefrom=null&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile&forcerm=1&labels=null&memory=0&memswap=0&networkmode=&nocache=1&rm=0&shmsize=0&t=gcr.io%2Ftejal-test%2Fimage1%3Alatest&target=&ulimits=null&version=": creating docker context: getting relative tar paths: file pattern [foo1] must match at least one file
../../../out/skaffold build -d gcr.io/tejal-test --cache-artifacts=false -p   16.34s user 4.53s system 46% cpu 44.447 total
```

Observations:
1.  Looks good.

5. Hitting control C:
```
../../../out/skaffold build -d gcr.io/tejal-test --cache-artifacts=false -p concurrency-3
Generating tags...
 - image1 -> gcr.io/tejal-test/image1:latest
 - image2 -> gcr.io/tejal-test/image2:latest
 - image3 -> gcr.io/tejal-test/image3:latest
 - image4 -> gcr.io/tejal-test/image4:latest
Building [image3]...
^CTraceback (most recent call last):
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 104, in <module>
Traceback (most recent call last):
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 104, in <module>
    main()
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 76, in main
    main()
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 76, in main
    gcloud_main = _import_gcloud_main()
    gcloud_main = _import_gcloud_main()
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 56, in _import_gcloud_main
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 56, in _import_gcloud_main
    import googlecloudsdk.gcloud_main
    import googlecloudsdk.gcloud_main
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/gcloud_main.py", line 33, in <module>
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/gcloud_main.py", line 33, in <module>
    from googlecloudsdk.api_lib.iamcredentials import util as iamcred_util
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/api_lib/iamcredentials/util.py", line 25, in <module>
    from googlecloudsdk.api_lib.iamcredentials import util as iamcred_util
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/api_lib/iamcredentials/util.py", line 25, in <module>
    from googlecloudsdk.api_lib.util import apis_internal
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/api_lib/util/apis_internal.py", line 28, in <module>

Building [image4]...
    from googlecloudsdk.api_lib.util import apis_internal
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/api_lib/util/apis_internal.py", line 28, in <module>
    from googlecloudsdk.core import properties
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/core/properties.py", line 28, in <module>

    from googlecloudsdk.core import config
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/core/config.py", line 29, in <module>
WARN[0001] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image1%22%3Atrue%7D%7D": context canceled 
    from googlecloudsdk.core.util import files as file_utils
    from googlecloudsdk.core import properties
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/core/util/files.py", line 28, in <module>
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/core/properties.py", line 28, in <module>
WARN[0001] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image2%22%3Atrue%7D%7D": context canceled 
    import shutil
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/shutil.py", line 10, in <module>
WARN[0001] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image3%22%3Atrue%7D%7D": context canceled 
    from googlecloudsdk.core import config
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/core/config.py", line 29, in <module>
WARN[0001] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image4%22%3Atrue%7D%7D": context canceled 
    import fnmatch
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
    from googlecloudsdk.core.util import files as file_utils
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/core/util/files.py", line 28, in <module>
    import shutil
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/shutil.py", line 22, in <module>
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load_unlocked
    import bz2
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 914, in _find_spec
  File "<frozen importlib._bootstrap_external>", line 779, in exec_module
  File "<frozen importlib._bootstrap_external>", line 1342, in find_spec
  File "<frozen importlib._bootstrap_external>", line 911, in get_code
  File "<frozen importlib._bootstrap_external>", line 580, in _compile_bytecode
  File "<frozen importlib._bootstrap_external>", line 1314, in _get_spec
KeyboardInterrupt
  File "<frozen importlib._bootstrap_external>", line 1439, in find_spec
  File "<frozen importlib._bootstrap_external>", line 87, in _path_stat
KeyboardInterrupt
```

Observations:
1.Same error trace for multiple builds. In this case, both image3 and image4 show the same stack trace. Would it be a bing change to just print "Build interrupted / cancelled" 
2.  This error log is confusing. These warning messages indicate, 
```
WARN[0001] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image3%22%3Atrue%7D%7D": context canceled 
    from googlecloudsdk.core import config
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/core/config.py", line 29, in <module>
WARN[0001] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image4%22%3Atrue%7D%7D": context canceled 
``` 
We are still trying to list images from docker daemon.  This should be been aborted.


On master, i dont see these (ofcourse with sequential builds)
```
➜  microservices git:(fix-4713) ✗ skaffold build -d gcr.io/tejal-test --cache-artifacts=false
Generating tags...
 - leeroy-web -> gcr.io/tejal-test/leeroy-web:v1.15.0-26-gb1f9e2de2
 - leeroy-app -> gcr.io/tejal-test/leeroy-app:v1.15.0-26-gb1f9e2de2
Building [leeroy-web]...
^CTraceback (most recent call last):
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 104, in <module>
    main()
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 76, in main
    gcloud_main = _import_gcloud_main()
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 56, in _import_gcloud_main
    import googlecloudsdk.gcloud_main
There is a new version (1.15.0) of Skaffold available. Download it from:
  https://github.com/GoogleContainerTools/skaffold/releases/tag/v1.15.0

  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/gcloud_main.py", line 33, in <module>
    from googlecloudsdk.api_lib.iamcredentials import util as iamcred_util
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/api_lib/iamcredentials/util.py", line 31, in <module>
    from googlecloudsdk.core.credentials import transports
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/core/credentials/transports.py", line 25, in <module>
    from googlecloudsdk.core.credentials import requests
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/core/credentials/requests.py", line 24, in <module>
    from googlecloudsdk.core import requests
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/core/requests.py", line 31, in <module>
    import requests
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/third_party/requests/__init__.py", line 114, in <module>
    from .models import Request, Response, PreparedRequest
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/third_party/requests/models.py", line 16, in <module>
    import encodings.idna
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/encodings/idna.py", line 3, in <module>
    import stringprep, re, codecs
  File "<frozen importlib._bootstrap>", line 988, in _find_and_load
  File "<frozen importlib._bootstrap>", line 148, in __enter__
  File "<frozen importlib._bootstrap>", line 176, in _get_module_lock
KeyboardInterrupt
➜  microservices git:(fix-4713) ✗ 
```

File: pkg/skaffold/build/result.go (hunk 32 above)
Comment (gsquared94) [reply pkg/skaffold/build/result.go:54]: I'd like to keep it. I think that `logAggregator` can be reused somewhere else if we have a need for concurrent runs for something. So I wanted to keep it safe, otherwise there won't be any error messages if writer requests exceed the buffer capacity and the code would deadlock.

File: pkg/skaffold/build/scheduler.go (hunk 31 above)
Comment (gsquared94) [reply pkg/skaffold/build/scheduler.go:66]: I'd like to keep it this way for 2 reasons:
- The build failure events and logs are already setup if the error is raised in `performBuild` and I don't want to repeat them elsewhere
- I've tried to reuse existing code functions as much as possible and would like to avoid changes there.

File: pkg/skaffold/build/model.go [hunk 35]
```diff
@@ -0,0 +1,94 @@
...
+			wait:      make(chan interface{}),
+		}
+	}
+
+	var nodes []buildNode
+	for _, a := range artifacts {
+		ar := nodeMap[a.ImageName]
+		for _, d := range a.Dependencies {
+			ch, found := nodeMap[d.ImageName]
+			if !found {
```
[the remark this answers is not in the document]
gsquared94: actually we have a validation on config parse to check for missing dependencies. If the artifact isn't found here it'd mean that it's not required to be built (already found in cache or no updates during a new devloop). So we should ignore them.

pkg/skaffold/build/result.go
gsquared94: I'll leave it as is. 😆

pkg/skaffold/build/scheduler.go
gsquared94: fixed

pkg/skaffold/build/scheduler.go
gsquared94: fixed

pkg/skaffold/build/model.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+)
+
+// buildNode models the artifact dependency graph using a set of channels.
+// Each build node has a wait channel which it closes once it completes building by calling markComplete.
+// This notifies all listeners waiting for this node's build to complete.
+// Additionally it has a reference to the channels for each of its dependencies.
+// Calling `waitForDependencies` ensures that all required nodes' channels have already been closed and as such have finished building before the current artifact build starts.
+type buildNode struct {
```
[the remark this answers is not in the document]
gsquared94: done

gsquared94:
> run normal case
> * No indication if they are running concurrently.
>   Can we have something like `Building 4 artifacts in parallel`

Added message:
```go
	if concurrency > 1 {
		color.Default.Fprintf(out, "Building %d artifacts in parallel\n", concurrency)
	}
```

> Run with a build failure in image 4 which should cancel dependent artifacts image2 and image 3 build as well as already scheduled image3
> * There is a line "Building [image 3]" artifact which did not complete due to error in image 4.  Instead of closing the channel for in progress builds, should we first write something similar to
>   " cancelling due to one or more builds failed"

It shouldn't output partial logs for any build actually. If it starts outputting logs for image 3 it'll wait until all logs are printed. So the actual builder needs to handle context cancellation to output such a message.

> Hitting control C:
> Observations:
> 1.Same error trace for multiple builds. In this case, both image3 and image4 show the same stack trace. Would it be a big change to just print "Build interrupted / cancelled"

We do allow redirecting build logs into files so I think it's better to show the full stack trace for both failures. The actual stack trace seems to be a failure of the individual builders to handle context cancellation properly, I can add a bug for that but don't want to complicate this PR with these fixes.

> 2. This error log is confusing. These warning messages indicate,
> 
> ```
> WARN[0001] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image3%22%3Atrue%7D%7D": context canceled 
>     from googlecloudsdk.core import config
>   File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/core/config.py", line 29, in <module>
> WARN[0001] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image4%22%3Atrue%7D%7D": context canceled 
> ```
> 
> We are still trying to list images from docker daemon. This should have been aborted.
> 
> On master, i dont see these (ofcourse with sequential builds)

This was because image pruner wasn't handling context cancellation correctly. I've made this one fix. 🙂

gsquared94:
thanks @tejal29 for such detailed test notes. 👏🏽

pkg/skaffold/build/local/prune.go
```diff
@@ -158,6 +158,10 @@ func (p *pruner) collectImagesToPrune(ctx context.Context, artifacts []*latest.A
 
 		imgs, err := p.listImages(ctx, a.ImageName)
 		if err != nil {
+			switch err {
```
tejal29, line 161: nice catch!

integration/build_dependencies_test.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/GoogleContainerTools/skaffold/integration/skaffold"
+)
+
+func TestBuild_WithDependencies(t *testing.T) {
```
tejal29, line 28: Add an integration tests where `skaffold dev` fails on this example.

pkg/skaffold/build/scheduler.go
tejal29: sounds good.

tejal29:
> Added message:
> 
> ```go
> 	if concurrency > 1 {
> 		color.Default.Fprintf(out, "Building %d artifacts in parallel\n", concurrency)
> 	}
> ```
> 

sounds good. 


>  > Hitting control C:
> > Observations:
> > 1.Same error trace for multiple builds. In this case, both image3 and image4 show the same stack trace. Would it be a big change to just print "Build interrupted / cancelled"
> 
> We do allow redirecting build logs into files so I think it's better to show the full stack trace for both failures. The actual stack trace seems to be a failure of the individual builders to handle context cancellation properly, I can add a bug for that but don't want to complicate this PR with these fixes.
> 


> 
> This was because image pruner wasn't handling context cancellation correctly. I've made this one fix. 🙂

+1

pkg/skaffold/build/scheduler.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	s.results.Record(a, finalTag)
+	n.markComplete()
+	event.BuildComplete(a.ImageName)
+	return nil
+}
+
+// InOrder builds a list of artifacts in dependency order.
+func InOrder(ctx context.Context, out io.Writer, tags tag.ImageTags, artifacts []*latest.Artifact, artifactBuilder ArtifactBuilder, concurrency int) ([]Artifact, error) {
```
briandealwis, line 110: Re: colour output: `out` is normally `os.Stdout` here.

briandealwis (approved):
This is now always running builders with their output going to a pipe, so we loose colour in the build logs (since it's not `os.Stdout`, a TTY).  If we're in concurrency=1, it would be better to send the provider writer down and skip the logAggregator.

When running with concurrency=0, it seems an image is chosen at random to follow live.  Might be better to just hide the build logs and instead provide feedback on the builds as they happen and finish.  This can be done as a separate PR.

integration/build_dependencies_test.go
```diff
@@ -0,0 +1,98 @@
...
+				skaffold.Build(append(test.args, "--cache-artifacts=false")...).InDir("testdata/build-dependencies").RunOrFail(t)
+
+				// Run with artifact caching
+				skaffold.Build(append(test.args, "--cache-artifacts=true")...).InDir("testdata/build-dependencies").RunOrFail(t)
+
+				// Run a second time with artifact caching
+				out := skaffold.Build(append(test.args, "--cache-artifacts=true")...).InDir("testdata/build-dependencies").RunOrFailOutput(t)
+				if strings.Contains(string(out), "Not found. Building") {
+					t.Errorf("images were expected to be found in cache: %s", out)
+				}
```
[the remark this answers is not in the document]
gsquared94: It's the same integration test as `TestBuild`. I guess the idea to run without cacheing first is to make sure it doesn't use previous run cached images. The second run is for with cache but nothing is found in cache. The third is for with cache and everything is found in cache. 🤷🏾‍♂️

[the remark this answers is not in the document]
briandealwis: It would make more sense here to make a change to image2 and verify that image1 is rebuilt but image3 is not.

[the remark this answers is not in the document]
gsquared94: hashcode validation isn't implemented in this PR for build dependencies 🙂 

It only implements:
Config schema
Config validation
Build controller(only parity with existing code)
Build logs reporting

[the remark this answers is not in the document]
tejal29: I would say similar to `dev`, if `--cahce-artifacts=true` error out saying caching is not supported `build` with `requires`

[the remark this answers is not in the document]
gsquared94: ok, I've added checks that fail `skaffold build --cache-artifacts=true` and `skaffold dev` and `skaffold debug`. Also added integration tests.

pkg/skaffold/build/local/prune.go
gsquared94: the problem is that all docker client calls will fail immediately with a cancelled context so even if I don't skip here for `DeadlineExceeded` cleanup will still fail with these messages. We would need another context for the pruner and not the common context. I'll add a bug for this.

cmd/skaffold/app/cmd/build.go
```diff
@@ -67,7 +68,13 @@ func doBuild(ctx context.Context, out io.Writer) error {
 	}
 
 	return withRunner(ctx, func(r runner.Runner, config *latest.SkaffoldConfig) error {
-		bRes, err := r.BuildAndTest(ctx, buildOut, targetArtifacts(opts, config))
+		ar := targetArtifacts(opts, config)
+
+		// TODO: Remove this block after fixing artifact cache logic for artifacts with dependencies
```
tejal29, line 73: Please create an issue and add it.

integration/build_dependencies_test.go
```diff
@@ -96,3 +95,20 @@ func TestBuild_WithDependencies(t *testing.T) {
 		})
 	}
 }
+
+func TestDev_WithDependencies(t *testing.T) {
```
tejal29, line 99: Perfect!!

cmd/skaffold/app/cmd/dev.go
```diff
@@ -61,6 +61,10 @@ func runDev(ctx context.Context, out io.Writer) error {
 			return nil
 		default:
 			err := withRunner(ctx, func(r runner.Runner, config *latest.SkaffoldConfig) error {
+				// TODO: Remove this block after fixing devloop rebuild logic for artifacts with dependencies
```
tejal29, line 64: Add an issue link.

tejal29:
Another round. 

1. skaffold dev
```
../../../out/skaffold dev -d gcr.io/tejal-test
defining dependencies between artifacts is not yet supported for `skaffold dev` and `skaffold debug`
```

2. skaffold build with cache on
```
../../../out/skaffold build -d gcr.io/tejal-test
defining dependencies between artifacts is not yet supported for `skaffold build` with cache enabled. Run with `--cache-artifacts=false` flag
➜  build-dependencies git:(fix-4713) ✗ 
```


3. skaffold build with cache off and concurrency on.
```
../../../out/skaffold build -d gcr.io/tejal-test --cache-artifacts=false -p=concurrency-3
Generating tags...
 - image1 -> gcr.io/tejal-test/image1:latest
 - image2 -> gcr.io/tejal-test/image2:latest
 - image3 -> gcr.io/tejal-test/image3:latest
 - image4 -> gcr.io/tejal-test/image4:latest
Building 3 artifacts in parallel
Building [image4]...

````

4. Hit control-C
```
➜  build-dependencies git:(fix-4713) ✗ ../../../out/skaffold build -d gcr.io/tejal-test --cache-artifacts=false -p=concurrency-3
Generating tags...
 - image1 -> gcr.io/tejal-test/image1:latest
 - image2 -> gcr.io/tejal-test/image2:latest
 - image3 -> gcr.io/tejal-test/image3:latest
 - image4 -> gcr.io/tejal-test/image4:latest
Building 3 artifacts in parallel
Building [image4]...
^CTraceback (most recent call last):
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 104, in <module>
Traceback (most recent call last):
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 104, in <module>
    main()
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 76, in main
    main()
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 76, in main
    gcloud_main = _import_gcloud_main()
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 56, in _import_gcloud_main
    gcloud_main = _import_gcloud_main()
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/gcloud.py", line 56, in _import_gcloud_main
    import googlecloudsdk.gcloud_main
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/gcloud_main.py", line 33, in <module>
    import googlecloudsdk.gcloud_main
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/gcloud_main.py", line 33, in <module>

Building [image3]...

WARN[0003] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image1%22%3Atrue%7D%7D": context canceled 
WARN[0003] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image2%22%3Atrue%7D%7D": context canceled 
    from googlecloudsdk.api_lib.iamcredentials import util as iamcred_util
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/api_lib/iamcredentials/util.py", line 31, in <module>
    from googlecloudsdk.api_lib.iamcredentials import util as iamcred_util
.....
WARN[0003] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image3%22%3Atrue%7D%7D": context canceled 
  File "/Users/tejaldesai/Downloads/google-cloud-sdk2/lib/googlecloudsdk/api_lib/iamcredentials/util.py", line 31, in <module>
WARN[0003] failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image4%22%3Atrue%7D%7D": context canceled
➜  build-dependencies git:(fix-4713) ✗ 
````

I still see the warning when pruning.  I tried using `errors.Is(err, context.Canceled` to make sure all errors in the get unwrapped and compared. However it is not working. 
Using this diff, 
```
+++ b/pkg/skaffold/build/local/prune.go
@@ -19,6 +19,7 @@ package local
 import (
        "context"
        "sort"
+        "strings"
        "sync"
        "time"
 
@@ -158,8 +159,7 @@ func (p *pruner) collectImagesToPrune(ctx context.Context, artifacts []*latest.A
 
                imgs, err := p.listImages(ctx, a.ImageName)
                if err != nil {
-                       switch err {
-                       case context.Canceled, context.DeadlineExceeded:
+                       if strings.Contains(err.Error(), "context canceled") {
                                return nil
                        }
                        logrus.Warnf("failed to list images: %v", err)
```

I see the following:
```
➜  build-dependencies git:(fix-4713) ✗ ../../../out/skaffold build -d gcr.io/tejal-test --cache-artifacts=false -p=concurrency-3
Generating tags...
 - image1 -> gcr.io/tejal-test/image1:latest
 - image2 -> gcr.io/tejal-test/image2:latest
 - image3 -> gcr.io/tejal-test/image3:latest
 - image4 -> gcr.io/tejal-test/image4:latest
Building 3 artifacts in parallel
Building [image3]...
^C
Building [image4]...

➜  build-dependencies git:(fix-4713) ✗ 
```

tejal29 (commented):
Just one round of fixing the `Warnf` message on pruner.

tejal29 (approved):
Merge after creating issues for todos and adding them in the TODO comment.

cmd/skaffold/app/cmd/build.go
gsquared94: done!

cmd/skaffold/app/cmd/dev.go
gsquared94: done

PR #4829, opened 2020-09-28
Fix `skaffold filter` to handle multiple yaml documents

briandealwis:
Fixes: #4826, 
**Related**: #4827, #4732

**Description**
As discovered by @doriandekoning in #4826, the new Helm debug support does correctly handle the situation where a Helm chart may have multiple resources.

This PR does the following:
- it introduces a `manifest.Load()` function to create a `ManifestList` from a YAML input leveraging the `k8s.io/apimachinery`'s `YAMLReader`.
- `filter` now uses `manifest.Load()` to parse the manifests proferred on stdin
- adds a test for `debug` to verify that it applies to multiple manifests

### DIFF
--- a/pkg/skaffold/kubernetes/manifest/manifests.go
+++ b/pkg/skaffold/kubernetes/manifest/manifests.go
@@ -36,14 +36,15 @@ func Load(in io.Reader) (ManifestList, error) {
 	var docs [][]byte
 	for {
 		doc, err := r.Read()
-		if err == io.EOF {
-			break
-		} else {
+		switch {
+		case err == io.EOF:
+			return ManifestList(docs), nil
+		case err != nil:
 			return nil, err
+		default:
+			docs = append(docs, doc)
 		}
-		docs = append(docs, doc)
 	}
-	return ManifestList(docs), nil
 }
 
 func (l *ManifestList) String() string {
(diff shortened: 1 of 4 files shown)


pkg/skaffold/kubernetes/manifest/manifests.go
```diff
@@ -17,16 +17,33 @@ limitations under the License.
...
+
+	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 )
 
 // ManifestList is a list of yaml manifests.
 //nolint:golint
 type ManifestList [][]byte
 
+// Load uses the Kubernetes `apimachinery` to split YAML content into a set of YAML documents.
+func Load(in io.Reader) ManifestList {
```
gsquared94, line 34: how about `func Load(in io.Reader) (ManifestList, error)` and we can handle the error?

pkg/skaffold/kubernetes/manifest/manifests.go
```diff
@@ -17,16 +17,33 @@ limitations under the License.
...
+	r := k8syaml.NewYAMLReader(bufio.NewReader(in))
+	var docs [][]byte
+	for i := 0; ; i++ {
```
gsquared94, line 37:
```suggestion
	for {
```

pkg/skaffold/kubernetes/manifest/manifests.go
briandealwis: 👍

gsquared94 (approved):

Issue #4813, opened 2020-09-22
Update Quickstart with option to use Cloud Shell


briandealwis:
#4780 was opened as the user attempted to follow our Quickstart using Cloud Shell, which already has Skaffold, Minikube, and Kubectl installed.  The user installed Skaffold, Minikube, and Kubectl manually following the instructions, but something didn't quite work.

We should:
  1. Call out that some environments have these utilities pre-instaled
  2. Use Cloud Shell's [Open in Cloud Shell](https://cloud.google.com/shell/docs/open-in-cloud-shell) functionality to further simplify the quickstart by pre-populating the workspace.

PR #4830, opened 2020-09-28
Add Cloud Shell for simpler Quickstart

briandealwis:
Fixes: #4813 

**Description**

Adds an <kbd>Open in Cloud Shell</kbd> option to the Skaffold Quickstart.  Required a minor reorganization to the page to flow better (IMHO).


briandealwis:
I want to see if there's some better language for non-GCP customers to tell them how to create an account to use the Cloud Shell option.

MarlonGamez:
> I want to see if there's some better language for non-GCP customers to tell them how to create an account to use the Cloud Shell option.

What kind of information do you think we should give to the user about this? Does the user have to do any setup on their account besides creating a google account?

briandealwis:
PTAL @MarlonGamez — added a note that it requires a Google Account and links to the signup page.

MarlonGamez (approved):
Looks good!

Issue #4848, opened 2020-10-02
Skaffold will always return error adding artifacts to cache when builds are not local


ricardo-larosa:
### Expected behavior
Cache is populated with built artifacts avoiding rebuilding images every time
Related to https://github.com/GoogleContainerTools/skaffold/issues/4842 and https://github.com/GoogleContainerTools/skaffold/issues/4753

### Actual behavior
```
WARN[0087] error adding artifacts to cache; caching may not work as expected: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running? 
```
### Information

- Skaffold version: v1.14/v1.15
- Operating system: linux/macos
- Contents of skaffold.yaml: Any configuration with no local build
Example:
```yaml
profiles:
- name: oncluster
  build:
    artifacts:
    - image: registry.whatever/image-name
      kaniko:
        image: gcr.io/kaniko-project/executor:v1.0.0
        flags: 
          - --snapshotMode=redo
          - --verbosity=info
    cluster:
      annotations:
        sidecar.istio.io/inject: "false" # if you are not running istio comment it
      dockerConfig: 
        secretName: my-dockerconfigjson # is Opaque
      namespace: default
```

### Steps to reproduce the behavior

1. Use a profile oncluster that builds images remotely (kaniko, etc..)
2. Run `build -p oncluster --cache-file /a/local/path --verbosity trace`
3. Everything runs ok
4. Cache file is populated with latest build image
Now repeat, but before step 2 stop your local docker deamon
you will get this error:
```
WARN[0087] error adding artifacts to cache; caching may not work as expected: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running? 
```
and the cache file is not populated

PR #4850, opened 2020-10-02
Fix return of error adding artifacts to cache when images are built remotely

ricardo-larosa:
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/4848

**Description**
Fix return error adding artifacts to cache when builds are not local

### DIFF
--- a/pkg/skaffold/build/cache/retrieve.go
+++ b/pkg/skaffold/build/cache/retrieve.go
@@ -147,7 +147,6 @@ func maintainArtifactOrder(built []build.Artifact, artifacts []*latest.Artifact)
 }
 
 func (c *cache) addArtifacts(ctx context.Context, bRes []build.Artifact, hashByName map[string]string) error {
-
 	var err error
 
 	for _, a := range bRes {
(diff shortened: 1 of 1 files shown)


pkg/skaffold/build/cache/retrieve.go
```diff
@@ -147,8 +147,11 @@ func maintainArtifactOrder(built []build.Artifact, artifacts []*latest.Artifact)
 }
 
 func (c *cache) addArtifacts(ctx context.Context, bRes []build.Artifact, hashByName map[string]string) error {
```
gsquared94, line 149: you can simplify it a bit
```go
func (c *cache) addArtifacts(ctx context.Context, bRes []build.Artifact, hashByName map[string]string) error {
	for _, a := range bRes {
		entry := ImageDetails{}
		if c.imagesAreLocal {
			imageID, err := c.client.ImageID(ctx, a.Tag)
			if err != nil {
				return err
			}

			if imageID != "" {
				entry.ID = imageID
			}
		} else {
			ref, err := docker.ParseReference(a.Tag)
			if err != nil {
				return fmt.Errorf("parsing reference %q: %w", a.Tag, err)
			}
			entry.Digest = ref.Digest
		}
		c.cacheMutex.Lock()
		c.artifactCache[hashByName[a.ImageName]] = entry
		c.cacheMutex.Unlock()
	}
	return nil
}
```

ricardo-larosa: ok

ricardo-larosa: Thanks @gsquared94

tejal29 (approved):
Thanks for catching this  @ricardo-larosa 
The change looks good to me

PR #4871, opened 2020-10-07
Skaffold init support for polyglot-maven projects

MarlonGamez:
Fixes https://github.com/GoogleContainerTools/skaffold/issues/4335

Adds support for `skaffold init` to detect jib as a builder in different pom file extensions that are supported by https://github.com/takari/polyglot-maven.

I've added tests for a couple of extensions, but wasn't sure if I should for each case because I didn't want to clutter the tests with what is essentially just testing string matching

### DIFF
--- a/pkg/skaffold/build/jib/init.go
+++ b/pkg/skaffold/build/jib/init.go
@@ -89,13 +89,30 @@ func validate(path string, enableGradleAnalysis bool) []ArtifactConfig {
 	var builderType PluginType
 	var executable, wrapper, taskName, searchString, consoleFlag string
 	switch {
-	case strings.HasSuffix(path, "pom.xml"):
+	case isPomFile(path):
 		builderType = JibMaven
 		executable = "mvn"
 		wrapper = "mvnw"
 		searchString = "<artifactId>jib-maven-plugin</artifactId>"
 		taskName = "jib:_skaffold-init"
 		consoleFlag = "--batch-mode"
+
+		// Polyglot maven support: support pom files of different extension types
+		switch filepath.Ext(path) {
+		case ".atom", ".java", ".rb", ".kt", ".kts", ".ktm": // all define the plugin with similar syntax
+			searchString = ":jib-maven-plugin"
+		case ".clj":
+			searchString = "/jib-maven-plugin"
+		case ".groovy":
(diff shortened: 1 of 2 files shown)


pkg/skaffold/build/jib/init.go
```diff
@@ -89,13 +89,30 @@ func validate(path string, enableGradleAnalysis bool) []ArtifactConfig {
...
-	case strings.HasSuffix(path, "pom.xml"):
+	case isPomFile(path):
 		builderType = JibMaven
 		executable = "mvn"
 		wrapper = "mvnw"
 		searchString = "<artifactId>jib-maven-plugin</artifactId>"
 		taskName = "jib:_skaffold-init"
 		consoleFlag = "--batch-mode"
+
+		// Polyglot maven support: support pom files of different extension types
```
briandealwis, line 100: There's an easier and more robust way: polyglot maven [requires a `.mvn/extensions.xml` file](https://github.com/takari/polyglot-maven#usage).  Polyglot supports a bunch of other languages, including yaml 🙀

pkg/skaffold/build/jib/init.go
```diff
@@ -89,13 +89,30 @@ func validate(path string, enableGradleAnalysis bool) []ArtifactConfig {
 	var builderType PluginType
 	var executable, wrapper, taskName, searchString, consoleFlag string
 	switch {
-	case strings.HasSuffix(path, "pom.xml"):
+	case isPomFile(path):
 		builderType = JibMaven
 		executable = "mvn"
 		wrapper = "mvnw"
 		searchString = "<artifactId>jib-maven-plugin</artifactId>"
```
briandealwis, line 96: TBH I think just searching for `jib-maven-plugin` is enough.  After all, this string would be fooled if someone had `jib-maven-plugin` as a `<dependency>`.

MarlonGamez:
@briandealwis So do you think anything needs to be done beyond checking if something's a pom.* file and then searching for "jib-maven-plugin"?

briandealwis (approved):
I was going to say that we should check for the existence of `.mvn/extensions.xml`, but it could exist further up in the tree.  The chance of a pom.* that just happens to have the string `jib-maven-plugin` without being a Jib project is pretty slim.

Issue #4889, opened 2020-10-09
Update docs on behaviour of tryImportMissing


StarpTech:
### Expected behavior

A code change in the docker image should produce a different hash and cause a rebuild of the image.

### Actual behavior

Skaffold reports that the image **could be found** remotely. I double-checked it.

### Information

- Skaffold version: v1.15.0
- Operating system: Windows

```yaml
apiVersion: skaffold/v2beta8
kind: Config
build:
  tagPolicy:
    sha256: {}
  local:
    concurrency: 0
    tryImportMissing: true

profiles:
  - name: web
    build:
      artifacts:
        - image: eu.gcr.io/x/webapp
          context: frontpage
          docker:
            buildArgs:
              API_HOST: https://x.de
```

Github workflow

```yaml
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          submodules: true

      - name: Cache skaffold image builds & config
        uses: actions/cache@v2
        with:
          path: ~/.skaffold/
          key: fixed
```

StarpTech:
The reason is `tryImportMissing: true` in combination with the skaffold cache. Is it possible that the digest isn't respected here?

StarpTech:
Oh, `sha256` is the `latest` tag. That means skaffold tries to check if an image with the tag `latest` already exists. Due to the fact we have always a latest tag skaffold won't rebuild the image. This means `tryImportMissing` is not predictable with the `sha256` tagging strategy and should produce an error.

https://github.com/GoogleContainerTools/skaffold/blob/d9cf47da6aa179cc9b47e433b386491120f49578/pkg/skaffold/build/cache/lookup.go#L121-L127

briandealwis:
Do you have a `.dockerignore` file?

Can you include your build logs, ideally with `-v debug`?

StarpTech:
> Do you have a .dockerignore file?

Yes, but there are no changes between the runs.

> Can you include your build logs, ideally with -v debug?

Not yet.

briandealwis:
Yes, `tryImportMissing` won't be useful with the sha256 tagging strategy.

StarpTech:
> Yes, tryImportMissing won't be useful with the sha256 tagging strategy.

You mean incompatible?

briandealwis:
No, just what you've observed above.

[...]

PR #4890, opened 2020-10-09
validate tag policy constrain

StarpTech:
Fixes: #4889


pkg/skaffold/schema/validation/validation.go
```diff
@@ -56,6 +57,16 @@ func Process(config *latest.SkaffoldConfig) error {
 	return fmt.Errorf(strings.Join(messages, " | "))
 }
 
+// validateTaggingPolicy checks that the tagging policy is valid in combination with other options.
+func validateTaggingPolicy(bc latest.BuildConfig) (errs []error) {
+	if bc.LocalBuild != nil {
+		if bc.LocalBuild.TryImportMissing && bc.TagPolicy.ShaTagger != nil {
+			errs = append(errs, fmt.Errorf("tagging policy 'sha256' can not be used when 'tryImportMissing' is enabled"))
```
briandealwis, line 64: We should document why — we may change the sha256:
```suggestion
			// sha256 just uses `latest` tag, so tryImportMissing will virtually always succeed (#4889)
			errs = append(errs, fmt.Errorf("tagging policy 'sha256' can not be used when 'tryImportMissing' is enabled"))
```

pkg/skaffold/schema/validation/validation.go
```diff
@@ -56,6 +57,16 @@ func Process(config *latest.SkaffoldConfig) error {
...
+		}
+	}
+	return
```
briandealwis, line 67: Nit: this isn't needed

briandealwis (commented):
Thanks for the PR!  Could you please merge with HEAD and address these two issues, and we can get this in?

pkg/skaffold/schema/validation/validation.go
StarpTech: This is required because we have a named return value.

StarpTech:
23min for CI is a lot 😄

pkg/skaffold/schema/validation/validation.go
briandealwis: Oops! 🤦

briandealwis (approved):

Issue #4891, opened 2020-10-09
[Build Artifact Dependency] Handle proper image cache invalidation.


gsquared94:
PR https://github.com/GoogleContainerTools/skaffold/pull/4828 introduces the notion of defining dependencies between artifacts. However it doesn't handle invalidating the cache properly. 

For example: If user defines `artifact1` and `artifact2` where `artifact2` is a required artifact for `artifact1` then if there's a file change in `artifact2`'s workspace that invalidates the cache of `artifact2` then it should also invalidate the cache of `artifact1` by virtue of being a dependency.

PR #4893, opened 2020-10-09
Implement cache invalidation for dependent artifacts.

gsquared94:
Fixes: #4891 
**Related**: #4713 
**Merge after**: #4828 

**Description** (Part 2 of several PRs.)

This PR implements proper cache handling for artifacts with dependencies as described in the design for [supporting dependencies between build artifacts](https://github.com/GoogleContainerTools/skaffold/blob/master/docs/design_proposals/artifact-dependencies.md#image-cache).
**User facing changes**

If user defines `artifact1` and `artifact2` where `artifact2` is a required artifact for `artifact1` then if there's a file change in `artifact2`'s workspace that invalidates the cache of `artifact2` then it automatically also invalidates the cache of `artifact1` by virtue of being a dependency.

### DIFF
--- a/integration/build_dependencies_test.go
+++ b/integration/build_dependencies_test.go
@@ -99,6 +99,7 @@ func TestBuildDependenciesCache(t *testing.T) {
 	// These tests build 4 images and then make a file change to the images in `change`.
 	// The test then triggers another build and verifies that the images in `rebuilt` were built
 	// (e.g., the changed images and their dependents), and that the other images were found in the artifact cache.
+	// It runs the profile `concurrency-0` which builds with maximum concurrency.
 	tests := []struct {
 		description string
 		change      []int
@@ -136,14 +137,14 @@ func TestBuildDependenciesCache(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.description, func(t *testing.T) {
-			skaffold.Build("--cache-artifacts=true").InDir("testdata/build-dependencies").RunOrFail(t)
+			skaffold.Build("--cache-artifacts=true", "-p", "concurrency-0").InDir("testdata/build-dependencies").RunOrFail(t)
 			checkImagesExist(t)
 
 			// modify file `foo` to invalidate cache for target artifacts
 			for _, i := range test.change {
 				Run(t, fmt.Sprintf("testdata/build-dependencies/app%d", i), "sh", "-c", fmt.Sprintf("echo %s > foo", uuid.New().String()))
 			}
-			out, err := skaffold.Build("--cache-artifacts=true").InDir("testdata/build-dependencies").RunWithCombinedOutput(t)
+			out, err := skaffold.Build("--cache-artifacts=true", "-p", "concurrency-0").InDir("testdata/build-dependencies").RunWithCombinedOutput(t)
 			if err != nil {
 				t.Fatal("expected build to succeed")
 			}
--- a/integration/build_dependencies_test.go
+++ b/integration/build_dependencies_test.go
@@ -135,11 +135,11 @@ func TestBuildDependenciesCache(t *testing.T) {
 		},
 	}
 
+	skaffold.Build("--cache-artifacts=true", "-p", "concurrency-0").InDir("testdata/build-dependencies").RunOrFail(t)
+	checkImagesExist(t)
+
 	for _, test := range tests {
 		t.Run(test.description, func(t *testing.T) {
-			skaffold.Build("--cache-artifacts=true", "-p", "concurrency-0").InDir("testdata/build-dependencies").RunOrFail(t)
-			checkImagesExist(t)
-
 			// modify file `foo` to invalidate cache for target artifacts
 			for _, i := range test.change {
 				Run(t, fmt.Sprintf("testdata/build-dependencies/app%d", i), "sh", "-c", fmt.Sprintf("echo %s > foo", uuid.New().String()))
@@ -163,11 +163,11 @@ func TestBuildDependenciesCache(t *testing.T) {
 			}
 			checkImagesExist(t)
 		})
+	}
 
-		// revert file changes
-		for _, i := range test.change {
-			Run(t, fmt.Sprintf("testdata/build-dependencies/app%d", i), "sh", "-c", "> foo")
-		}
+	// revert file changes
+	for i := 1; i <= 4; i++ {
+		Run(t, fmt.Sprintf("testdata/build-dependencies/app%d", i), "sh", "-c", "> foo")
 	}
 }
 
--- a/integration/build_dependencies_test.go
+++ b/integration/build_dependencies_test.go
@@ -150,6 +150,7 @@ func TestBuildDependenciesCache(t *testing.T) {
 			}
 			log := string(out)
 
+			// TODO: [#4927] Instead of using the log messages we should use the event api to validate cache hits and misses.
 			for i := 1; i <= 4; i++ {
 				if !contains(test.rebuilt, i) && !strings.Contains(log, fmt.Sprintf("image%d: Found Locally", i)) {
 					logrus.Info("build output: ", string(out))
--- a/integration/build_dependencies_test.go
+++ b/integration/build_dependencies_test.go
@@ -17,15 +17,17 @@ limitations under the License.
 package integration
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 
+	"github.com/google/uuid"
 	"github.com/sirupsen/logrus"
 
 	"github.com/GoogleContainerTools/skaffold/integration/skaffold"
 )
 
-func TestBuild_WithDependencies(t *testing.T) {
+func TestBuildDependenciesOrder(t *testing.T) {
 	MarkIntegrationTest(t, CanRunWithoutGcp)
 
 	tests := []struct {
@@ -65,11 +67,6 @@ func TestBuild_WithDependencies(t *testing.T) {
 			args:        []string{"-p", "failed-dependency", "-p", "concurrency-0"},
 			failure:     `unable to stream build output: The command '/bin/sh -c [ "${FAIL}" == "0" ] || false' returned a non-zero code: 1`,
 		},
-		{
-			description:  "build failure with cache-artifacts=true",
-			cacheEnabled: true,
-			failure:      "defining dependencies between artifacts is not yet supported for `skaffold build` with cache enabled. Run with `--cache-artifacts=false` flag",
-		},
 	}
 
 	for _, test := range tests {
@@ -96,9 +93,95 @@ func TestBuild_WithDependencies(t *testing.T) {
 	}
 }
 
+func TestBuildDependenciesCache(t *testing.T) {
+	MarkIntegrationTest(t, CanRunWithoutGcp)
+
+	// These tests build 4 images and then make a file change to the images in `change`.
+	// The test then triggers another build and verifies that the images in `rebuilt` were built
+	// (e.g., the changed images and their dependents), and that the other images were found in the artifact cache.
+	tests := []struct {
+		description string
+		change      []int
+		rebuilt     []int
+	}{
+		{
+			description: "no change",
+		},
+		{
+			description: "change 1",
+			change:      []int{1},
+			rebuilt:     []int{1},
+		},
+		{
+			description: "change 2",
+			change:      []int{2},
+			rebuilt:     []int{1, 2},
+		},
+		{
+			description: "change 3",
+			change:      []int{3},
+			rebuilt:     []int{1, 2, 3},
+		},
+		{
+			description: "change 4",
+			change:      []int{4},
+			rebuilt:     []int{4},
+		},
+		{
+			description: "change all",
+			change:      []int{1, 2, 3, 4},
+			rebuilt:     []int{1, 2, 3, 4},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			skaffold.Build("--cache-artifacts=true").InDir("testdata/build-dependencies").RunOrFail(t)
+			checkImagesExist(t)
+
+			// modify file `foo` to invalidate cache for target artifacts
+			for _, i := range test.change {
+				Run(t, fmt.Sprintf("testdata/build-dependencies/app%d", i), "sh", "-c", fmt.Sprintf("echo %s > foo", uuid.New().String()))
+			}
+			out, err := skaffold.Build("--cache-artifacts=true").InDir("testdata/build-dependencies").RunWithCombinedOutput(t)
+			if err != nil {
+				t.Fatal("expected build to succeed")
+			}
+			log := string(out)
+
+			for i := 1; i <= 4; i++ {
+				if !contains(test.rebuilt, i) && !strings.Contains(log, fmt.Sprintf("image%d: Found Locally", i)) {
+					logrus.Info("build output: ", string(out))
+					t.Fatalf("expected image%d to be cached", i)
+				}
+
+				if contains(test.rebuilt, i) && !strings.Contains(log, fmt.Sprintf("image%d: Not found. Building", i)) {
+					logrus.Info("build output: ", string(out))
+					t.Fatalf("expected image%d to be rebuilt", i)
+				}
+			}
+			checkImagesExist(t)
+		})
+
+		// revert file changes
+		for _, i := range test.change {
+			Run(t, fmt.Sprintf("testdata/build-dependencies/app%d", i), "sh", "-c", "> foo")
+		}
+	}
+}
+
 func checkImagesExist(t *testing.T) {
 	checkImageExists(t, "gcr.io/k8s-skaffold/image1:latest")
 	checkImageExists(t, "gcr.io/k8s-skaffold/image2:latest")
 	checkImageExists(t, "gcr.io/k8s-skaffold/image3:latest")
 	checkImageExists(t, "gcr.io/k8s-skaffold/image4:latest")
+}
+
+func contains(sl []int, t int) bool {
+	for _, i := range sl {
+		if i == t {
+			return true
+		}
+	}
+	return false
 }
--- a/integration/build_dependencies_test.go
+++ b/integration/build_dependencies_test.go
@@ -150,7 +150,6 @@ func TestBuildDependenciesCache(t *testing.T) {
 			}
 			log := string(out)
 
-			// TODO: [#4927] Instead of using the log messages we should use the event api to validate cache hits and misses.
 			for i := 1; i <= 4; i++ {
 				if !contains(test.rebuilt, i) && !strings.Contains(log, fmt.Sprintf("image%d: Found Locally", i)) {
 					logrus.Info("build output: ", string(out))
--- a/pkg/skaffold/build/cache/cache.go
+++ b/pkg/skaffold/build/cache/cache.go
@@ -48,11 +48,11 @@ type cache struct {
 	artifactCache    ArtifactCache
 	cacheMutex       sync.RWMutex
 	client           docker.LocalDaemon
-	cfg              docker.Config
+	cfg              Config
 	cacheFile        string
 	imagesAreLocal   bool
 	tryImportMissing bool
-	hashForArtifact  func(ctx context.Context, a *latest.Artifact) (string, error)
+	lister           DependencyLister
 }
 
 // DependencyLister fetches a list of dependencies for an artifact
@@ -96,9 +96,7 @@ func NewCache(cfg Config, imagesAreLocal bool, tryImportMissing bool, dependenci
 		cacheFile:        cacheFile,
 		imagesAreLocal:   imagesAreLocal,
 		tryImportMissing: tryImportMissing,
-		hashForArtifact: func(ctx context.Context, a *latest.Artifact) (string, error) {
-			return getHashForArtifact(ctx, dependencies, a, cfg.Mode())
-		},
+		lister:           dependencies,
 	}, nil
 }
 
--- a/pkg/skaffold/build/cache/cache.go
+++ b/pkg/skaffold/build/cache/cache.go
@@ -26,6 +26,7 @@ import (
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/sirupsen/logrus"
 
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker"
@@ -46,6 +47,7 @@ type ArtifactCache map[string]ImageDetails
 // cache holds any data necessary for accessing the cache
 type cache struct {
 	artifactCache    ArtifactCache
+	artifactGraph    build.ArtifactGraph
 	cacheMutex       sync.RWMutex
 	client           docker.LocalDaemon
 	cfg              Config
@@ -67,7 +69,7 @@ type Config interface {
 }
 
 // NewCache returns the current state of the cache
-func NewCache(cfg Config, imagesAreLocal bool, tryImportMissing bool, dependencies DependencyLister) (Cache, error) {
+func NewCache(cfg Config, imagesAreLocal bool, tryImportMissing bool, dependencies DependencyLister, graph build.ArtifactGraph) (Cache, error) {
 	if !cfg.CacheArtifacts() {
 		return &noCache{}, nil
 	}
@@ -91,6 +93,7 @@ func NewCache(cfg Config, imagesAreLocal bool, tryImportMissing bool, dependenci
 
 	return &cache{
 		artifactCache:    artifactCache,
+		artifactGraph:    graph,
 		client:           client,
 		cfg:              cfg,
 		cacheFile:        cacheFile,
--- a/pkg/skaffold/build/cache/hash.go
+++ b/pkg/skaffold/build/cache/hash.go
@@ -52,7 +52,7 @@ type artifactHasherImpl struct {
 	artifacts build.ArtifactGraph
 	lister    DependencyLister
 	mode      config.RunMode
-	once      *util.Once
+	syncStore *util.SyncStore
 }
 
 // newArtifactHasher returns a new instance of an artifactHasher. Use newArtifactHasherFunc instead of calling this function directly.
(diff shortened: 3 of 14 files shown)


tejal29:
Please rebase! :)

integration/build_dependencies_test.go
```diff
@@ -17,14 +17,97 @@ limitations under the License.
...
+	"github.com/google/uuid"
 	"github.com/sirupsen/logrus"
 
 	"github.com/GoogleContainerTools/skaffold/integration/skaffold"
 )
 
+func TestBuild_WithCache(t *testing.T) {
+	MarkIntegrationTest(t, CanRunWithoutGcp)
+
+	tests := []struct {
```
briandealwis, line 33: It would be useful to have some comments here as to what these fields mean.
```suggestion
	// These tests builds 4 images and then makes a file change to the images in `change`.
	// The test then triggers another build and verifies that the images in `rebuilt` were built
	// (e.g., the changed images and their dependencies), and that the other images were not.
	tests := []struct {
```

integration/build_dependencies_test.go
```diff
@@ -17,14 +17,97 @@ limitations under the License.
...
+			skaffold.Build("--cache-artifacts=true").InDir("testdata/build-dependencies").RunOrFail(t)
+			checkImagesExist(t)
+
+			// modify file `foo` to invalidate cache for target artifacts
+			for _, i := range test.change {
+				Run(t, fmt.Sprintf("testdata/build-dependencies/app%d", i), "sh", "-c", fmt.Sprintf("echo %s > foo", uuid.New().String()))
+			}
+			out, err := skaffold.Build("--cache-artifacts=true").InDir("testdata/build-dependencies").RunWithCombinedOutput(t)
+
+			if err != nil {
```
briandealwis, line 85: Drop the missing line?
```suggestion
			out, err := skaffold.Build("--cache-artifacts=true").InDir("testdata/build-dependencies").RunWithCombinedOutput(t)
			if err != nil {
```

integration/build_dependencies_test.go
```diff
@@ -17,14 +17,97 @@ limitations under the License.
...
 )
 
+func TestBuild_WithCache(t *testing.T) {
+	MarkIntegrationTest(t, CanRunWithoutGcp)
+
+	tests := []struct {
+		description string
+		rebuilt     []int
+		cached      []int
+		change      []int
```
briandealwis, line 37: Reorder these.  Isn't `cache = images - rebuilt`?
```suggestion
		change      []int
		rebuilt     []int
		cached      []int
```

integration/build_dependencies_test.go
```diff
@@ -17,14 +17,97 @@ limitations under the License.
...
+	}{
+		{
+			description: "no change",
+			cached:      []int{1, 2, 3, 4},
+		},
+		{
+			description: "change 1",
+			rebuilt:     []int{1},
+			change:      []int{1},
+			cached:      []int{2, 3, 4},
```
briandealwis, line 47:
```suggestion
			change:      []int{1},
			rebuilt:     []int{1},
			cached:      []int{2, 3, 4},
```

pkg/skaffold/build/cache/once.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+)
+
+// hashOnce ensures that each artifact hash is computed at most once.
+// This is useful when several artifacts request the hash for a common required artifact.
+type hashOnce interface {
+	get(a *latest.Artifact, hashFunc func(*latest.Artifact) (string, error)) (string, error)
+}
```
briandealwis, line 31: Is the single-hash really part of the interface?  Or is it an implementation detail?  Do callers need to get the hash of an artifact with multiple hash-functions?  Could this be an `artifactHasher` that is instantiated with the hash-function?

pkg/skaffold/build/cache/hash.go
```diff
@@ -80,14 +105,19 @@ func getHashForArtifact(ctx context.Context, depLister DependencyLister, a *late
 	if args != nil {
 		inputs = append(inputs, args...)
 	}
+	return encode(inputs)
+}
 
+func encode(inputs []string) (string, error) {
+	if len(inputs) == 0 {
+		return "", nil
```
briandealwis, line 113: Is this what was produced previously if there were no inputs?  Or would the Encode() error?
(Would we receive `0000000000000000000` or something similar, and is this something we don't want to support?)

briandealwis (commented):
Some initial comments

pkg/skaffold/build/cache/hash.go
```diff
@@ -38,15 +38,40 @@ import (
...
-	artifactConfigFunction = artifactConfig
+	fileHasherFunc         = fileHasher
+	artifactConfigFunc     = artifactConfig
+	singleArtifactHashFunc = singleArtifactHash
 )
 
-func getHashForArtifact(ctx context.Context, depLister DependencyLister, a *latest.Artifact, mode config.RunMode) (string, error) {
+func getHashForArtifact(ctx context.Context, depLister DependencyLister, a *latest.Artifact, mode config.RunMode, h hashOnce, artifacts Artifacts) (string, error) {
+	hash, err := h.get(a, func(artifact *latest.Artifact) (string, error) {
+		return singleArtifactHashFunc(ctx, depLister, artifact, mode)
```
tejal29, line 48: This can return a empty string  has code "" if no dependencies are present.

pkg/skaffold/build/cache/hash_test.go
```diff
@@ -477,3 +477,9 @@ func TestHashBuildArgs(t *testing.T) {
 		})
 	}
 }
+
+type fakeHashOnceImpl struct{}
```
tejal29, line 481: Can we add some unit tests for `getHashForArtifact` with dependencies?
e.g. 
1. For  an artifact with same files, same artifact dependencies, in different order produce the same hash
2. For an artifact with same files, different artifact dependencies produce different hash.
3. For  an artifact with same files, same artifact dependency, when an artifact dependency hash changes, the  artifact hash should change

pkg/skaffold/build/cache/once.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+}
+
+type hashOnceImpl struct {
+	oncePerArtifact *sync.Map
+	results         *sync.Map
+}
+
+func (h *hashOnceImpl) get(a *latest.Artifact, hashFunc func(a *latest.Artifact) (string, error)) (string, error) {
+	once, _ := h.oncePerArtifact.LoadOrStore(a.ImageName, new(sync.Once))
```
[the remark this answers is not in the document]
gsquared94: I don't think we should be trying to micro optimize at our scale. But ya we do not benefit from using the `sync.Map` over `map` with `RWMutex` other than a cleaner implementation.

integration/build_dependencies_test.go
gsquared94: fixed

pkg/skaffold/build/cache/hash.go
gsquared94: Previously these lines were embedded in another function and it wasn't possible to get nil inputs since at least the `Artifact` struct was marshalled and added to it. But I agree that it should probably error. Fixed.

pkg/skaffold/build/cache/hash.go
gsquared94: if you're referring to the `encode` function, then I fixed that.

integration/build_dependencies_test.go
gsquared94: fixed

integration/build_dependencies_test.go
gsquared94: fixed

pkg/skaffold/build/cache/once.go
gsquared94: The `artifactHasher` is already embedded in `cache` as the function `getHashForArtifact`. I don't want to refactor too much of the code so I think it makes sense for `hashOnce` as a wrapper. 

It doesn't have to have an interface tied to hash really and I can move it to `utils` package and change it to:
```go
type Once {}
// Do calls the function f iff it's being called the first time for a specific key (evaluated from the key function).
// If it's called multiple times for the same key only the first call will execute and store the result of f.
// All other calls will be blocked until the running instance of f returns and all of them receive the same result.
func(*Once) Do(key func() interface{}, f func() interface{}) interface{}
```

WDYT?

gsquared94:
PTAL @nkubala @briandealwis @tejal29. I'll add more UT after everyone's ok with the code. 🙂

pkg/skaffold/build/cache/hash_test.go
gsquared94: Done

pkg/skaffold/build/cache/once.go
gsquared94: discussed over DM.

pkg/skaffold/build/cache/once.go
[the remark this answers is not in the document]
tejal29: its ok for now. i just wanted to make sure we are aware of this.

tejal29:
> PTAL @nkubala @briandealwis @tejal29. I'll add more UT after everyone's ok with the code. 🙂

+1 for me. Please write unit tests where we can make sure we don't see any issue when artifacts are build in parallel.

gsquared94:
> +1 for me. Please write unit tests where we can make sure we don't see any issue when artifacts are build in parallel.

I've added the common scenario UTs and changed the `TestBuildDependenciesCache` IT to build concurrently to cover this scenario. Also all cache checks conclude before the builds start so I don't think the behavior changes between sequential and concurrent builds.

pkg/skaffold/build/cache/hash_test.go
```diff
@@ -146,17 +148,102 @@ func TestGetHashForArtifact(t *testing.T) {
...
+			t.CheckDeepEqual(test.expected, actual)
+		})
+	}
+}
+
+func TestGetHashForArtifactWithDependencies(t *testing.T) {
+	tests := []struct {
+		description  string
+		dependencies map[string][]string
+		artifacts    []*latest.Artifact
```
briandealwis, line 173: This ordering is a bit odd

pkg/skaffold/build/cache/hash_test.go
```diff
@@ -146,17 +148,102 @@ func TestGetHashForArtifact(t *testing.T) {
...
+		mode         config.RunMode
+		expected     string
+	}{
+		{
+			description:  "hash for artifact with two dependencies",
+			dependencies: map[string][]string{"img1": {"a"}, "img2": {"b"}, "img3": {"c"}, "img4": {"d"}},
```
briandealwis, line 179: This is hard to understand: there is no "a", "b", "c", or "d" visible.  There should be a comment above to describe what these mean.

pkg/skaffold/build/cache/hash_test.go
```diff
@@ -47,6 +47,8 @@ var fakeArtifactConfig = func(a *latest.Artifact) (string, error) {
 	return "", nil
 }
 
+const Dockerfile = "Dockerfile"
```
briandealwis, line 50: This doesn't seem to save much :-)

pkg/skaffold/build/cache/hash.go
```diff
@@ -38,15 +38,72 @@ import (
...
+	fileHasherFunc        = fileHasher
+	artifactConfigFunc    = artifactConfig
 )
 
-func getHashForArtifact(ctx context.Context, depLister DependencyLister, a *latest.Artifact, mode config.RunMode) (string, error) {
+type artifactHasher interface {
+	hash(ctx context.Context, a *latest.Artifact) (string, error)
+}
+
+type Artifacts map[string]*latest.Artifact
```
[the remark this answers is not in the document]
briandealwis: You have a few places where you're `make(map[string]*latest.Artifact)` where this could be used instead.  Could be useful to have a creator function that takes an `[]latest.Artifact`?

pkg/skaffold/build/cache/hash.go
```diff
@@ -38,15 +38,72 @@ import (
...
+
+type artifactHasherImpl struct {
```
briandealwis, line 52: Could use this compile-time trick to ensure this type implements the function:
```
var _ artifactHasher = (*artifactHasherImpl)(nil)

File: pkg/skaffold/build/cache/lookup.go [hunk 15]
```diff
@@ -47,8 +51,8 @@ func (c *cache) lookupArtifacts(ctx context.Context, tags tag.ImageTags, artifac
 	return details
 }
 
-func (c *cache) lookup(ctx context.Context, a *latest.Artifact, tag string) cacheDetails {
-	hash, err := c.hashForArtifact(ctx, a)
+func (c *cache) lookup(ctx context.Context, a *latest.Artifact, tag string, h artifactHasher) cacheDetails {
```
briandealwis, line 54: What if you moved the artifactHasher into the `cache` instance?

pkg/skaffold/build/cache/lookup_test.go
```diff
@@ -108,11 +109,13 @@ func TestLookupLocal(t *testing.T) {
...
-				artifactCache:   test.cache,
-				client:          fakeLocalDaemon(test.api),
-				hashForArtifact: test.hasher,
+				imagesAreLocal: true,
+				artifactCache:  test.cache,
+				client:         fakeLocalDaemon(test.api),
+				cfg:            &mockConfig{mode: config.RunModes.Build},
 			}
+
+			t.Override(&newArtifactHasherFunc, func(_ Artifacts, _ DependencyLister, _ config.RunMode) artifactHasher { return test.hasher })
```
briandealwis, line 118: If you moved the artifactCacher into the `cache`, then you'd simplify your tests.

pkg/skaffold/build/cache/cache.go
```diff
@@ -48,11 +48,11 @@ type cache struct {
 	artifactCache    ArtifactCache
 	cacheMutex       sync.RWMutex
 	client           docker.LocalDaemon
-	cfg              docker.Config
+	cfg              Config
 	cacheFile        string
 	imagesAreLocal   bool
 	tryImportMissing bool
-	hashForArtifact  func(ctx context.Context, a *latest.Artifact) (string, error)
```
briandealwis, line 55: Why not just replace this with the `artifactHasher`?

gsquared94: I can do this but then `cache` will need a `Reset` method that flushes the `artifactHasher` since I want to recompute the values _on every dev loop_ but don't want to recompute any value _within a dev loop_. I think the current implementation of instantiating `artifactHasher` on every `lookup` is probably better.

pkg/skaffold/build/cache/lookup_test.go
gsquared94: same reason as another question
>I can do this but then cache will need a Reset method that flushes the artifactHasher since I want to recompute the values on every dev loop but don't want to recompute any value within a dev loop. I think the current implementation of instantiating artifactHasher on every lookup is probably better.

pkg/skaffold/build/cache/lookup.go
gsquared94: same reason as another question
>I can do this but then cache will need a Reset method that flushes the artifactHasher since I want to recompute the values on every dev loop but don't want to recompute any value within a dev loop. I think the current implementation of instantiating artifactHasher on every lookup is probably better.

pkg/skaffold/build/cache/hash_test.go
gsquared94: we are at the mercy of linters 🤷🏾‍♂️

pkg/skaffold/build/cache/hash.go
```diff
@@ -38,15 +38,72 @@ import (
...
+	val := h.once.Do(func() interface{} {
+		return a.ImageName
+	},
+		func() interface{} {
+			hash, err := singleArtifactHash(ctx, h.lister, a, h.mode)
+			if err != nil {
+				return err
+			}
+			return hash
+		})
```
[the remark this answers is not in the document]
gsquared94: so the scenario is that `singleArtifactHash` is getting called from multiple goroutines concurrently but I want to only run it once(any one goroutine) and have all the others wait until it completes and then all of them get the same result from the single processing. I think this is the simplest way of achieving this. 

Using a single `sync.Map` I can't block all goroutines from running, otherwise can you elaborate how it'd work with a single map?

pkg/skaffold/build/cache/hash.go
gsquared94: neat, but I don't think I need it here since I have the `newArtifactHasher` function which will also fail at compile-time.

integration/build_dependencies_test.go
```diff
@@ -96,9 +93,96 @@ func TestBuild_WithDependencies(t *testing.T) {
 	}
 }
 
+func TestBuildDependenciesCache(t *testing.T) {
```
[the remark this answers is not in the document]
gsquared94: Great point. I've moved the startup and teardown outside the TDD loop so it's effectively what you suggested and runs in ~30 secs.

pkg/skaffold/build/cache/hash_test.go
gsquared94: fixed

integration/build_dependencies_test.go
```diff
@@ -96,9 +93,96 @@ func TestBuild_WithDependencies(t *testing.T) {
...
+				Run(t, fmt.Sprintf("testdata/build-dependencies/app%d", i), "sh", "-c", fmt.Sprintf("echo %s > foo", uuid.New().String()))
+			}
+			out, err := skaffold.Build("--cache-artifacts=true", "-p", "concurrency-0").InDir("testdata/build-dependencies").RunWithCombinedOutput(t)
+			if err != nil {
+				t.Fatal("expected build to succeed")
+			}
+			log := string(out)
+
+			for i := 1; i <= 4; i++ {
+				if !contains(test.rebuilt, i) && !strings.Contains(log, fmt.Sprintf("image%d: Found Locally", i)) {
```
tejal29, line 154: +1 to not depend on output and consume the event api to make test if not a huge re-work.

pkg/skaffold/build/cache/hash.go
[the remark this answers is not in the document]
tejal29: a little bit confused here. 
Is the `hash` calculated after the build seq is defined? ie. when is artifact cache look up done?

If its done after determining the build seq, then there would be only be 1 goroutine writing/computing the hash for a base artifact. Once it is computed and stored if say in a `sync.Map, multiple goroutines can read it at the same time. There is no need to block reads.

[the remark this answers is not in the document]
gsquared94: It's in `BuildAndTest` function in `build_deploy.go`. We explicitly call `cache.Build` which computes the cache for all artifacts in parallel. For the filtered list of artifacts that need to be rebuilt we then call `InOrder` builder which orchestrates the builds in the correct order.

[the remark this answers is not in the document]
gsquared94: So the cache computation isn't happening in the build pipeline per artifact but rather as a pre step.

tejal29 (commented):
LGTM except `util.Once` if its really required.

integration/build_dependencies_test.go
gsquared94: Any specific reason to not depend on log output? There a few tests that do that already.

tejal29:
> LGTM except `util.Once` if its really required.


sorry i hit https://github.com/GoogleContainerTools/skaffold/pull/4893#discussion_r508236793 this as a single comment instead of a whole review. The need of `util.Once` makes more sense along with the above comment.

pkg/skaffold/build/cache/hash_test.go
briandealwis: _"We shape our tools, and then our tools shape us"_

integration/build_dependencies_test.go
briandealwis: The tests become brittle.  Our log output isn't API.  Those other tests likely predate the event API, and should be changed.

gsquared94: I was gonna start working with this idea but there's no event api for cache hits and misses.

gsquared94: we'd have to code against `BuildStarted` present and absent(polling over a time window). I think that's a bit brittle too.

gsquared94: I created issue https://github.com/GoogleContainerTools/skaffold/issues/4927 to add an event for cache hits. I'll add a TODO to change this test after that gets done.

pkg/skaffold/build/cache/cache.go
gsquared94: Done

pkg/skaffold/build/cache/hash.go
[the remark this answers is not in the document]
briandealwis: Pending
👍 This is great. Can I suggest a name change — a SyncStore? Why can't we pass in a key directly instead of a function?

integration/build_dependencies_test.go
briandealwis: You *won't* get buildEvents for cached images.  So accumulate the buildEvents seen and then check that only the images expected were rebuilt.  And the order should be safe too when they have dependencies.

nkubala: +1, that's similar to what's happening in the `rpc_test.go` integration test as well. we shouldn't need to add an event for cache hits/misses unless we think we'll use it outside of tests.

pkg/skaffold/build/cache/hash.go
```diff
@@ -38,15 +38,82 @@ import (
...
+	artifactConfigFunc    = artifactConfig
 )
 
-func getHashForArtifact(ctx context.Context, depLister DependencyLister, a *latest.Artifact, mode config.RunMode) (string, error) {
+type artifactHasher interface {
+	hash(ctx context.Context, a *latest.Artifact) (string, error)
+}
+
+// Artifacts is a map of [artifact image : artifact definition]
+type Artifacts map[string]*latest.Artifact
```
nkubala, line 51: `ArtifactMap`?

pkg/skaffold/build/cache/hash.go
```diff
@@ -38,15 +38,82 @@ import (
...
+
+// ToArtifacts creates an instance of `Artifacts` from `[]*latest.Artifact`
+func ToArtifacts(artifacts []*latest.Artifact) Artifacts {
```
nkubala, line 54: this is the most overloaded term in skaffold I think :) `ToArtifactMap`?

pkg/skaffold/build/cache/hash.go
```diff
@@ -130,8 +199,8 @@ func hashBuildArgs(artifact *latest.Artifact, mode config.RunMode) ([]string, er
 	return sl, nil
 }
 
-// cacheHasher takes hashes the contents and name of a file
-func cacheHasher(p string) (string, error) {
+// fileHasher takes hashes the contents and name of a file
```
nkubala, line 202:
```suggestion
// fileHasher hashes the contents and name of a file
```

pkg/skaffold/build/cache/hash.go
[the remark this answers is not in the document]
gsquared94: I'll keep `Once`. IMO from `Once.Do` it's clearer that it runs the function once. It's equivalent to `sync.Once.Do`. Also the `Do` method explains the usage so there shouldn't be any confusion.

integration/build_dependencies_test.go
gsquared94: I get the sentiment of this request but hear me out:
I'm testing 6 scenarios. It's necessary to test all of them. If I read from the log output then I can immediately determine if the cache was hit or miss and return immediately. The overall test takes ~30 seconds. 
If I swap this out with an rpc client then:
- I cannot test the `all artifacts found in cache` scenario, since that build completes so quickly then we can't even establish an rpc client connection in time before the build is complete.
- I need to artificially run every sub test for a fixed amount of time with a sufficient buffer so as to ***not make it flaky*** (which is the point of your argument against the log approach). With 5 scenarios and a 30 sec window per build thats 2.5 minutes.
- "Log output isn't an API contract". Agreed but that doesn't make the test flaky. If someone changes the log output then they'll also have to change the test accordingly. There is no scenario where the test might intermittently fail under the current approach.

Some of these points could be alleviated with an explicit event for build cache hits but then again point 1 can't be fixed and I don't want to skip a test case to adhere to _best practice_.

cc @tstromberg

pkg/skaffold/build/cache/hash.go
gsquared94: named this `ArtifactGraph`

pkg/skaffold/build/cache/hash.go
gsquared94: fixed

pkg/skaffold/build/cache/hash.go
gsquared94: renamed to `ArtifactGraph`

pkg/skaffold/build/cache/lookup.go
```diff
@@ -31,14 +31,16 @@ import (
 
 func (c *cache) lookupArtifacts(ctx context.Context, tags tag.ImageTags, artifacts []*latest.Artifact) []cacheDetails {
 	details := make([]cacheDetails, len(artifacts))
-
+	// create a new `artifactHasher` on every new dev loop.
```
briandealwis, line 34:
```suggestion
	// Create a new `artifactHasher` on every new dev loop.
```

pkg/skaffold/util/once.go
```diff
@@ -0,0 +1,48 @@
...
+limitations under the License.
+*/
+
+package util
+
+import (
+	"sync"
+)
+
+type Once struct {
```
briandealwis, line 23: Doc comment.  I really don't like this name as it is confusing given sync.Once.

briandealwis (approved):
One nit and a plea for a rename of util.Once (though I won't block it)

pkg/skaffold/util/once.go
gsquared94: I read through the Go style guide. Package names are supposed to be considered part of an exported type. So here `util.Once` is identifiable as different from `sync.Once`. Also the point about surrounding context help disambiguate types instead of packing information in the type names. I made the change since you insisted on it 🙂

Issue #4888, opened 2020-10-09
Image pruner operations fail if the context was cancelled or timed out during build or deploy.


gsquared94:
The type `pruner` in `prune.go` uses the same `context.Context` that are used by build or deploy. As such if the user hits `Ctrl+C` to cancel the build or deploy sometimes they get the error message:
```
failed to list images: error during connect: Get "https://127.0.0.1:32774/v1.40/images/json?filters=%7B%22reference%22%3A%7B%22image3%22%3Atrue%7D%7D": context canceled 
    from googlecloudsdk.core import config
```
which is returned from the docker api client that `pruner` calls with a cancelled context.

gsquared94:
@ilya-zuyev I think it's from https://github.com/GoogleContainerTools/skaffold/pull/4792
Specifically in `pkg/skaffold/build/local/local.go`:
```
	rt, err := build.InParallel(ctx, out, tags, artifacts, builder, *b.local.Concurrency)

	if b.prune {
		if b.mode == config.RunModes.Build {
			b.localPruner.synchronousCleanupOldImages(ctx, artifacts)
		} else {
			b.localPruner.asynchronousCleanupOldImages(ctx, artifacts)
		}
	}
```

ilya-zuyev:
Yep, that's me.
/assign

ilya-zuyev:
Yep, that's me. Thanks!

PR #4894, opened 2020-10-09
Do not log pruner context errors when Skaffold process is interrupted

ilya-zuyev:
Fixes https://github.com/GoogleContainerTools/skaffold/issues/4888

This PR improves logging of errors received from docker API in Pruner.
With this PR, the pruner returns immediately and doesn't log the error If a Docker API call was failed because of cancelled or timed out context.


pkg/skaffold/build/local/prune.go
```diff
@@ -115,7 +115,7 @@ func (p *pruner) runPrune(ctx context.Context, ids []string) error {
 
 	beforeDu, err := p.diskUsage(ctx)
 	if err != nil {
-		logrus.Warnf("Failed to get docker usage info: %v", err)
+		logrus.Debugf("Failed to get docker usage info: %v", err)
```
briandealwis, line 118: Should we continue with the pruning if the usage failed?  In the case in question, the context was cancelled so we should be returning ASAP.

ilya-zuyev: The usage may fail, e.g. if another docker request is in progress. It's ok for us, we need the usage data only for logging and just skip the log message if usage request failed. 

Good point about the early return. Fixed.

pkg/skaffold/build/local/prune.go
```diff
@@ -160,7 +166,11 @@ func (p *pruner) collectImagesToPrune(ctx context.Context, artifacts []*latest.A
 		if err != nil {
 			switch err {
 			case context.Canceled, context.DeadlineExceeded:
-				return nil
+				// ctx is done or cancelled or timed out
+				// Usually means the process is interrupted
+				// Skip logging and return immediately
+				// https://github.com/GoogleContainerTools/skaffold/issues/4888
```
briandealwis, line 172: I'm normally all for adding comments, but this is self-evident given the err values.  I'd suggest removing it.

briandealwis (approved):

pkg/skaffold/build/local/prune.go
ilya-zuyev: Removed the comment.

It could be a bit unclear how we ended up with a cancelled context. Having a comment about Ctl-C pressed could help to understand why we're here and what's the best way to handle the situation. But, on the other hand, checks like this are pretty common in our code, and there's no need to comment each one.

briandealwis: There are a number of ways that a context may be cancelled — we may even cancel it programmatically (e.g., file change and so we’re going to trigger another cycle of the dev loop). Such is life :-)

PR #4919, opened 2020-10-16
stop the kind cluster if it has created for integration tests

anshlykov:
Fixes: #4729

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,11 +201,12 @@ integration-in-kind: skaffold-builder
 		--network kind \
 		gcr.io/$(GCP_PROJECT)/skaffold-builder \
 		sh -eu -c ' \
-			NEED_STOP_KIND=1; \
-			kind get clusters | grep -q kind && NEED_STOP_KIND=0 || TERM=dumb kind create cluster --image=$(KIND_NODE); \
+			if ! kind get clusters | grep -q kind; then \
+			  trap "kind delete cluster" 0 1 2 15; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -201,9 +201,11 @@ integration-in-kind: skaffold-builder
 		--network kind \
 		gcr.io/$(GCP_PROJECT)/skaffold-builder \
 		sh -eu -c ' \
-			kind get clusters | grep -q kind || TERM=dumb kind create cluster --image=$(KIND_NODE); \
+			NEED_STOP_KIND=1; \
+			kind get clusters | grep -q kind && NEED_STOP_KIND=0 || TERM=dumb kind create cluster --image=$(KIND_NODE); \
 			kind get kubeconfig --internal > /tmp/kind-config; \
-			make integration \
+			make integration; \
+			[ "$$NEED_STOP_KIND" = 0 ] || kind delete cluster; \
```
briandealwis, line 208: I think it might be easier to use `trap`:
```
if ! kind get clusters | grep -q kind; then \
  trap "kind delete cluster" 0 1 2 15; \
  TERM=dumb kind create cluster --image=$(KIND_NODE); \
fi; \
kind get kubeconfig --internal > /tmp/kind-config; \
make integration \
'
```

briandealwis (commented):
Thanks for your PR!

Makefile
anshlykov: I think you're right, thanks for the advice

briandealwis (approved):
Thanks for the fix!

Issue #4806, opened 2020-09-18
Auto-sync should be able to be disabled explicitly


briandealwis:
Auto-sync is now enabled by default for buildpacks (#4656).  Although auto-sync can be disabled by setting a manual sync rule (#4805), we should be able to explicitly disable auto-sync in the `skaffold.yaml`; this will be is important to enable #4769.  

My suggestion is to turn the [`Sync.Auto`](https://github.com/GoogleContainerTools/skaffold/blob/77244b857eb8f7c9f16ff4fc4aa72430f8a29911/pkg/skaffold/schema/latest/config.go#L781) to be a boolean.

PR #4923, opened 2020-10-18
Make it possible to disable auto-sync 

briandealwis:
Fixes: #4806 

**Description**
Turn the `sync.auto` to a boolean.

- auto-sync can now be disabled with `sync.auto = false` on the respective artifacts

### DIFF
--- a/docs/content/en/docs/pipeline-stages/filesync.md
+++ b/docs/content/en/docs/pipeline-stages/filesync.md
@@ -98,20 +98,15 @@ Changes to other file types trigger an image rebuild.
 
 ##### Disable Auto Sync for Buildpacks
 
-To disable auto sync, simply provide a manual sync rule.  It does
-not matter if the sync rule does not match any actual files.
-For example:
+To disable auto sync, set `sync.auto = false`:
 
 ```
 artifacts:
 - image: xxx
   buildpacks:
     builder: gcr.io/buildpacks/builder:v1
-  # disable buildpacks auto-sync
   sync: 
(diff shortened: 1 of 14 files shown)


loosebazooka:
There's some docs and code in examples that use auto:{}
for jib specifically: (/integration)/examples/jib-syc

briandealwis:
@loosebazooka 5f29e27 fixed the `auto:{}` in `integration/examples/jib-sync`, so they'll come through on the next release.

loosebazooka:
oh weird, I must have clicked on the commit or something

loosebazooka:
Looks fine to me, what happens when we need to configure auto in the future? should buildpacks not automatically enable sync?

loosebazooka (approved):

briandealwis:
Auto sync is now enabled by default for buildpacks (#4656). Disabling requiring specifying a manual sync rule. Oops, I didn't update the docs!

loosebazooka:
How does that interact with the "one of" parsing of yaml? I'm not sure I remember what happens if the user injects some manual sync into the buildpacks artifact configuration?

briandealwis:
The `oneOf` is [only checked at parse time](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/yamltags/tags.go#L167): it has no effect on the underlying Go structs.

The auto-sync enablement happens as port-processing step, and only happens when there are no manual or inferred rules.

gsquared94 (approved):

Issue #4922, opened 2020-10-16
[Build Artifact Dependency] Implement required artifact resolution in individual builders


gsquared94:
Related: https://github.com/GoogleContainerTools/skaffold/issues/4713

This is a tracking issue to implement artifact dependency resolution in the individual builders:

- [x] Local Docker
- [x] Local Buildpacks
- [x] Local Jib
- [x] Custom
- [x] Kaniko
- [x] GCB Docker
- [x] GCB Buildpacks
- [x] GCB Jib
- [x] GCB Kaniko

gsquared94:
Closing via #5003

PR #4924, opened 2020-10-19
Implement artifact resolution in Docker and Buildpack builders

gsquared94:
**Related**: #4713, #4922
**Merge after**: #4893 

This PR implements resolving required artifacts as build args in docker builder; and as builder or run image in buildpacks, as described in the design for [supporting dependencies between build artifacts](https://github.com/GoogleContainerTools/skaffold/blob/master/docs/design_proposals/artifact-dependencies.md#docker-builder).
**User facing changes**

Modified example `integration/examples/microservices` app to test the required artifacts resolution as build args. Commands working:
```
skaffold dev --cache-artifacts=false --port-forward
```
```
skaffold build --cache-artifacts=false
```

#4893 will make this work with cache so it's a prereq PR. Will rebase and add tests after that gets in but general code review welcome!

### DIFF
--- a/integration/examples/microservices/base/Dockerfile
+++ b/integration/examples/microservices/base/Dockerfile
@@ -1,0 +1,5 @@
+FROM alpine:3.10
+# Define GOTRACEBACK to mark this container as using the Go language runtime
+# for `skaffold debug` (https://skaffold.dev/docs/workflows/debug/).
+ENV GOTRACEBACK=single
+CMD ["./app"]
--- a/integration/examples/microservices/leeroy-app/Dockerfile
+++ b/integration/examples/microservices/leeroy-app/Dockerfile
@@ -1,12 +1,9 @@
+ARG BASE
 FROM golang:1.12.9-alpine3.10 as builder
 COPY app.go .
 # `skaffold debug` sets SKAFFOLD_GO_GCFLAGS to disable compiler optimizations
 ARG SKAFFOLD_GO_GCFLAGS
 RUN go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -o /app .
 
-FROM alpine:3.10
-# Define GOTRACEBACK to mark this container as using the Go language runtime
-# for `skaffold debug` (https://skaffold.dev/docs/workflows/debug/).
-ENV GOTRACEBACK=single
-CMD ["./app"]
+FROM $BASE
 COPY --from=builder /app .
--- a/integration/examples/microservices/leeroy-web/Dockerfile
+++ b/integration/examples/microservices/leeroy-web/Dockerfile
@@ -1,12 +1,9 @@
+ARG BASE
 FROM golang:1.12.9-alpine3.10 as builder
 COPY web.go .
 # `skaffold debug` sets SKAFFOLD_GO_GCFLAGS to disable compiler optimizations
 ARG SKAFFOLD_GO_GCFLAGS
-RUN go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -o /web .
+RUN go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -o /app .
 
-FROM alpine:3.10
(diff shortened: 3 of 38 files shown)


tejal29 (commented):
One qq i had was adding `store` to the build signature. 
We have done this in past and there was a stage where the function signatures became crazy long. Every change affected multiple test files to add nil object to the test methods.
That is when we decided to introduce and add `runContext`. 
e.g. https://github.com/GoogleContainerTools/skaffold/pull/1824#pullrequestreview-215956698 and 
https://github.com/GoogleContainerTools/skaffold/pull/1885

I was wondering, if we could avoid the same here by  
e.g. have a method `WithArtifactStore` for builders that use the store.
I would also recommend breaking this PR in 2 stages
- Adding artifact resolution just to docker builder
- adding artifact resolution to buildpacks. 

Right now this PR is big.

I am not sure if `WithArtifactStore` would accomplish what you want to do. But i want to make sure we have investigated this direction where we avoid making changes to the interface.

gsquared94:
> One qq i had was adding `store` to the build signature.
> We have done this in past and there was a stage where the function signatures became crazy long. Every change affected multiple test files to add nil object to the test methods.
> That is when we decided to introduce and add `runContext`.
> e.g. [#1824 (review)](https://github.com/GoogleContainerTools/skaffold/pull/1824#pullrequestreview-215956698) and
> #1885
> 
> I was wondering, if we could avoid the same here by
> e.g. have a method `WithArtifactStore` for builders that use the store.
> I would also recommend breaking this PR in 2 stages
> 
> * Adding artifact resolution just to docker builder
> * adding artifact resolution to buildpacks.
> 
> Right now this PR is big.
> 
> I am not sure if `WithArtifactStore` would accomplish what you want to do. But i want to make sure we have investigated this direction where we avoid making changes to the interface.

I also felt the parameter explosion happening to so many `Builder` interfaces. Do you think this interface with an added method is better than the one with an added parameter?
```go
type Builder interface {
	// Store returns an interface to update or get the build results
	Store() BuiltArtifacts
	// Prune removes images built with Skaffold
	Prune(context.Context, io.Writer) error
}
```

tejal29:
> ```go
> Store() BuiltArtifacts
> ```

Thanks @gsquared94.  
The `Builder` should know how to
1. build artifacts
2. prune 

Would it be a builder's responsibility to update the built artifacts to a store? Should it know about a store?
I was reading on [Responsibility Driven Design](http://www.wirfs-brock.com/PDFs/A_Brief-Tour-of-RDD.pdf) and [0](https://www2.cs.arizona.edu/~mercer/Presentations/OOPD/12-RDD-Jukebox.pdf) and i feel like the builder should know just one think to build and clean up what it build. 

Embedding the `store` might mean we are giving too many responsibilities to it.
What do you think?

gsquared94:
> > ```go
> > Store() BuiltArtifacts
> > ```
> 
> Thanks @gsquared94.
> The `Builder` should know how to
> 
> 1. build artifacts
> 2. prune
> 
> Would it be a builder's responsibility to update the built artifacts to a store? Should it know about a store?
> I was reading on [Responsibility Driven Design](http://www.wirfs-brock.com/PDFs/A_Brief-Tour-of-RDD.pdf), [0](https://www2.cs.arizona.edu/~mercer/Presentations/OOPD/12-RDD-Jukebox.pdf)
> I feel like the builder should know just one thing to build and clean up what it build.
> 
> Embedding the `store` might mean we are giving too many responsibilities to it.
> What do you think?

Good point. `store` is an implementation detail and probably the `Builder` implementation can encapsulate that. It's doesn't need to be in the interface method parameter list either.

gsquared94:
closing this in favor of https://github.com/GoogleContainerTools/skaffold/pull/4932 where I've separated out the changes to only `docker builder`, rebased on master and changed the interface definitions according to @tejal29 's ideas.

PR #4937, opened 2020-10-24
(1/2) Use kpt deployer as a pipeline.

yuwenma:
Fixes: #4880 #3904

**Description**
1. Change the kpt core logic to work as a pipeline and run in the following order:
 read stdin (kpt fn source) -> hydrate (kustomize build) -> run functions (kpt fn run)
  -> store stdout (kpt fn sink)

2. Avoid storing intermediate data to keep the customer's directory is clean even if the kpt deployer operation fails.

3. Remove the  unexpected prefix from the file names.


pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -178,38 +177,79 @@ func (k *Deployer) Render(ctx context.Context, out io.Writer, builds []build.Art
...
-		return nil, fmt.Errorf("kustomize build: %w", err)
+	// Hydrate the manifests source.
+	_, err = kustomize.FindKustomizationConfig(k.Dir)
+	// Only run kustomize if kustomization.yaml is found.
+	if err == nil {
+		// Note: A tmp dir is used to provide kustomize the manifest directory.
+		// Once the unified kpt/kustomize is done, kustomize can be run as a kpt fn step and
+		// this additional directory creation/deletion will no longer be needed.
+
+		err := os.MkdirAll(tmpKustomizeDir, os.ModePerm)
```
tejal29, line 204: is it ok to have an  existing `.kustomize` with some files in it?
Shd the code also check if there is a `.kustomize` and clean it up?

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -178,38 +177,79 @@ func (k *Deployer) Render(ctx context.Context, out io.Writer, builds []build.Art
 // adding image digests, and adding run-id labels.
 func (k *Deployer) renderManifests(ctx context.Context, _ io.Writer, builds []build.Artifact,
 	flags []string) (manifest.ManifestList, error) {
+	var err error
 	debugHelpersRegistry, err := config.GetDebugHelpersRegistry(k.globalConfig)
 	if err != nil {
 		return nil, fmt.Errorf("retrieving debug helpers registry: %w", err)
 	}
-	if k.Fn.SinkDir == "" {
```
tejal29, line 185: If `k.Fn.SinkDir` is not set, what would the directory the rendered manifests would be sinked to?

tejal29 (commented):
Minor link failure.

pkg/skaffold/deploy/kpt/kpt.go
yuwenma: It won't sink to any directory. Users only want to specify the sink dir when they plan to do further git operations on the hydrated manifests (e.g. store it in another repo).

pkg/skaffold/deploy/kpt/kpt.go
yuwenma: sure.

yuwenma:
@tejal29 Thanks for the review Tejal. PTAL.

pkg/skaffold/deploy/kpt/kpt.go
tejal29: ah got it. does it make sense to remove `tmpSinkDir` from elsewhere?
Looks like it not required now.

tejal29 (changes requested):
Please fix the minor lint issue and unused `tmpSinkDir`. 
Everything else looks good.

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -178,38 +172,80 @@ func (k *Deployer) Render(ctx context.Context, out io.Writer, builds []build.Art
...
+			nil, nil)...)
+	buf, err = util.RunCmdOut(cmd)
+	if err != nil {
 		return nil, fmt.Errorf("reading config manifests: %w", err)
 	}
 
-	if err := k.kustomizeBuild(ctx); err != nil {
-		return nil, fmt.Errorf("kustomize build: %w", err)
+	// Hydrate the manifests source.
+	_, err = kustomize.FindKustomizationConfig(k.Dir)
```
tejal29, line 192: you could short cut this 

```suggestion
	if _, err := kustomize.FindKustomizationConfig(k.Dir); err == nil {
	         // Note: A tmp dir is used to provide kustomize the manifest directory.
		// Once the unified kpt/kustomize is done, kustomize can be run as a kpt fn step and
		// this additional directory creation/deletion will no longer be needed.
		if err := os.RemoveAll(tmpKustomizeDir); err != nil {
			return nil, fmt.Errorf("removing %v:%w", tmpKustomizeDir, err)
		}
		if err := os.MkdirAll(tmpKustomizeDir, os.ModePerm); err != nil {
			return nil, err
		}
		defer func() {
			os.RemoveAll(tmpKustomizeDir)
		}()
		err = k.sink(ctx, buf, tmpKustomizeDir)
		if err != nil {
			return nil, err
		}

		cmd := exec.CommandContext(ctx, "kustomize", append([]string{"build"}, tmpKustomizeDir)...)
		buf, err = util.RunCmdOut(cmd)
		if err != nil {
			return nil, fmt.Errorf("kustomize build: %w", err)
		}
	}
```

tejal29 (approved):
LGTM

PR #4940, opened 2020-10-25
(2/2)  Support `fn-path`  in kpt pipeline.

yuwenma:
Fixes: #4880 #3904 
**Related**: https://github.com/GoogleContainerTools/kpt/issues/1149
**Merge after**: #4937 

**Description**

Run kpt functions in a kpt pipeline.

Current kpt pipeline can not support fn-path nor using kpt functions along with kustomize (kustomzie exclude kpt functions if not adding to kustomization.yaml). Related issue https://github.com/GoogleContainerTools/kpt/issues/1149
    
This solution pulls kpt functions specifically from the kpt source inputs and adds it back to the pipeline after kustomize build finishes. It does this by using a kustomize builtin function to manipulate the Rnode and pulls out the "Items" from ResourceList (Kustomize terms).


tejal29:
@yuwenma can you please rebase so i can include this in tomorrow's release

yuwenma:
> @yuwenma can you please rebase so i can include this in tomorrow's release

Thanks! Rebased.

MarlonGamez:
Will we want to revert these changes once the issue is fixed on the kpt side?

yuwenma:
> Will we want to revert these changes once the issue is fixed on the kpt side?
> 
> Also, looks like there are some more linter errors from gci, but those should be quick

Linters Fixed.

I don't think it could be a quick fix on kpt side. We would want to revisit this approach once the unified kpt/kustomize is done (kustomize is read in as a kpt resource).

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -188,6 +189,25 @@ func (k *Deployer) renderManifests(ctx context.Context, _ io.Writer, builds []bu
 		return nil, fmt.Errorf("reading config manifests: %w", err)
 	}
 
+	// A workaround for issue https://github.com/GoogleContainerTools/kpt/issues/1149
+	// fn-path cannot be recognized in kpt pipeline mode, and it results in that the kpt functions
+	// in are ignored.
```
MarlonGamez, line 194: nit: Maybe expand this to explain what we're doing here for the workaround. I know you explained in the PR writeup but it might be nice to have it in the code as well

MarlonGamez (approved):
this LGTM. I left some a comment with just a nit since some of the stuff might be harder to understand without knowing what each of the kpt commands does

PR #4958, opened 2020-10-28
Add kpt function to file watcher.

yuwenma:
Fixes: #4879  #3904
**Merge after**: https://github.com/GoogleContainerTools/skaffold/pull/4940

**Description**
Add kpt functions to file watcher (trigger auto deploy when kpt function changes), even if the kpt function files are out of the deploy.kpt.dir


yuwenma:
@MarlonGamez

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -414,9 +479,8 @@ func (k *Deployer) getKptFnRunArgs() ([]string, error) {
 	}
 
 	count := 0
-
+	// fn-path is not supported due to kpt issue https://github.com/GoogleContainerTools/kpt/issues/1149
 	if len(k.Fn.FnPath) > 0 {
-		flags = append(flags, "--fn-path", k.Fn.FnPath)
```
tejal29, line 419: Its looks a little weird to not error out here. Can we error out with message "Config FnPath is not supported" ?
This could happen early on to fail fast when we validate schema

pkg/skaffold/deploy/kpt/kpt_test.go
```diff
@@ -115,9 +115,10 @@ func TestKpt_Deploy(t *testing.T) {
 				- foo.yaml`},
 			commands: testutil.
 				CmdRunOut("kpt fn source .", ``).
+				AndRunOut("kpt fn source kpt-func.yaml", ``).
 				AndRunOut(fmt.Sprintf("kpt fn sink %v", tmpKustomizeDir), ``).
 				AndRunOut(fmt.Sprintf("kustomize build %v", tmpKustomizeDir), ``).
-				AndRunOut("kpt fn run --dry-run --fn-path kpt-func.yaml", testPod).
+				AndRunOut("kpt fn run --dry-run", testPod).
```
tejal29, line 121: so, it ok to run `kpt fn run` ignoring the user defined `kpt-func.yaml` ?

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -188,6 +200,27 @@ func (k *Deployer) renderManifests(ctx context.Context, _ io.Writer, builds []bu
 		return nil, fmt.Errorf("reading config manifests: %w", err)
 	}
 
+	// A workaround for issue https://github.com/GoogleContainerTools/kpt/issues/1149
```
tejal29, line 203: Thanks for adding this comment. 
I am a bit hesitant on adding a workaround for this bug, if #1149 is being actively worked on. 

I am assuming, this code will be removed once #1149 is fixed is that right?

tejal29 (changes requested):

pkg/skaffold/deploy/kpt/kpt.go
yuwenma: Skaffold and kpt have different release cycle. We can revert it once the kpt fix is in. But it won't be done easily (IMO). For this release, the workaround is necessary to avoid kpt deployer from broken.

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -264,6 +297,38 @@ func (k *Deployer) renderManifests(ctx context.Context, _ io.Writer, builds []bu
...
+	var kptFn []byte
+	for i := range rl.Items {
+		item, err := rl.Items[i].String()
+		if err != nil {
+			return nil, fmt.Errorf("reading Item %w", err)
+		}
+		var obj unstructured.Unstructured
+		jByte, err := k8syaml.YAMLToJSON([]byte(item))
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
yuwenma: oh, it is fine if there's an error (actually it can be common if some resource is not yaml/json format). Added some comments.

pkg/skaffold/deploy/kpt/kpt.go
yuwenma: I changed the comment. Skaffold should support fn-path. Previous release works because fn-path is used in a wrong way (it supposes to be refers to a directory, but used as specifying a certain file).

pkg/skaffold/deploy/kpt/kpt_test.go
yuwenma: haha, it's a wrong usage in previous code. fn-path refers to directory, not a single file.

tejal29 (approved):

tejal29 (approved):

MarlonGamez (approved):
Looks good, thanks for addressing Tejal's comments

PR #4962, opened 2020-10-28
Implement required artifact resolution in buildpacks builder

gsquared94:
**Related**: #4713, #4922

**Description** (Part 4 of several PRs.)

This PR implements resolving required artifacts as either the builder image or the run image in buildpacks builder as described in the design for [supporting dependencies between build artifacts](https://github.com/GoogleContainerTools/skaffold/blob/master/docs/design_proposals/artifact-dependencies.md#buildpacks-builder).

### DIFF
--- a/pkg/skaffold/build/buildpacks/lifecycle.go
+++ b/pkg/skaffold/build/buildpacks/lifecycle.go
@@ -86,7 +86,7 @@ func (b *Builder) build(ctx context.Context, out io.Writer, a *latest.Artifact, 
 			}
 		}
 	}
-	builderImage, runImage, noPull := fromRequiredArtifacts(artifact, b.artifacts, a.Dependencies, b.pushImages)
+	builderImage, runImage, pull := fromRequiredArtifacts(artifact, b.artifacts, a.Dependencies, b.pushImages)
 	if err := runPackBuildFunc(ctx, out, b.localDocker, pack.BuildOptions{
 		AppPath:      workspace,
 		Builder:      builderImage,
@@ -94,7 +94,7 @@ func (b *Builder) build(ctx context.Context, out io.Writer, a *latest.Artifact, 
 		Buildpacks:   buildpacks,
 		Env:          env,
 		Image:        latest,
-		NoPull:       noPull,
+		NoPull:       !pull,
 		TrustBuilder: artifact.TrustBuilder,
 		// TODO(dgageot): Support project.toml include/exclude.
 		// FileFilter: func(string) bool { return true },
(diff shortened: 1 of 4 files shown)


pkg/skaffold/build/buildpacks/lifecycle.go
```diff
@@ -166,3 +164,49 @@ func envMap(env []string) map[string]string {
 
 	return kv
 }
+
+// fromRequiredArtifacts replaces the provided builder and run images with built images from the required artifacts if specified.
+func fromRequiredArtifacts(artifact *latest.BuildpackArtifact, r ArtifactResolver, deps []*latest.ArtifactDependency, pushImages bool) (string, string, bool) {
+	builderImage, runImage := artifact.Builder, artifact.RunImage
+	builderImageLocal, runImageLocal, noPull := false, false, false
```
MarlonGamez, line 171: nit: having this be called noPull makes readability a bit worse imo. Could be named pull and the logic flipped so that !pull is the same as noPull, and makes things easier to read in cases like !noPull

pkg/skaffold/build/buildpacks/lifecycle.go
```diff
@@ -166,3 +164,49 @@ func envMap(env []string) map[string]string {
...
+			}
+			builderImageLocal = true
+		}
+		if runImage == d.Alias {
+			runImage, found = r.GetImageTag(d.ImageName)
+			if !found {
+				logrus.Fatalf("failed to resolve build result for required artifact %q", d.ImageName)
+			}
+			runImageLocal = true
+		}
```
MarlonGamez, line 187: can these two code paths be combined into one? And then do the check at the end to either set `builderImageLocal = true` or `runImageLocal = true`

MarlonGamez (commented):
Looks good :) just have a couple nits/q's

pkg/skaffold/build/buildpacks/lifecycle.go
gsquared94: not quite sure how though, can you suggest the code change you're thinking?

pkg/skaffold/build/buildpacks/lifecycle.go
gsquared94: done

pkg/skaffold/build/buildpacks/lifecycle.go
MarlonGamez: On second look, maybe they can't, my bad 😅

Do you think the log messages should be differentiated to say either builder image or run image? Or do you think the user will have enough info from the ImageName?

gsquared94: this is actually an impossible scenario. It'll only happen if we ever introduce a bug into the code.

MarlonGamez (approved):
Looks good 👍🏼 thanks for addressing the readability change :)

PR #4972, opened 2020-10-29
Implement required artifact resolution for custom builder

gsquared94:
**Related**: #4713, #4922

**Description** (Part 5 of several PRs.)

This PR implements resolving required artifacts as environment variables in the custom builder as described in the design for [supporting dependencies between build artifacts](https://github.com/GoogleContainerTools/skaffold/blob/master/docs/design_proposals/artifact-dependencies.md#custom-builder).


MarlonGamez:
Looks like travis didn't auto run, might need to push an empty commit or something to try to get it to :/ but I'd say these changes look good 👍🏼 Will wait for travis

integration/testdata/build-dependencies/app1/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -e
+# image2 is set by Skaffold to be the image built for app2
+docker build -t $IMAGE --build-arg image2 .
+if $PUSH_IMAGE; then
```
briandealwis, line 5: This is a nifty construct that I haven't seen used before.  But you're executing data that is passed in, which we generally try to avoid.

integration/testdata/build-dependencies/app1/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -e
+# image2 is set by Skaffold to be the image built for app2
+docker build -t $IMAGE --build-arg image2 .
```
briandealwis, line 4: `"$IMAGE"` should be quoted here and further below

pkg/skaffold/build/local/local.go
```diff
@@ -89,36 +90,50 @@ func (b *Builder) buildArtifact(ctx context.Context, out io.Writer, a *latest.Ar
...
+// artifactBuilder represents a per artifact builder interface
+type artifactBuilder interface {
+	Build(ctx context.Context, out io.Writer, a *latest.Artifact, tag string) (string, error)
+}
 
+func getArtifactBuilder(b *Builder, a *latest.Artifact) (artifactBuilder, error) {
 	switch {
 	case a.DockerArtifact != nil:
-		return b.buildDocker(ctx, out, a, tag, b.mode)
+		return newDockerBuilder(b), nil
```
briandealwis, line 101: It there a reason to not follow the other builders and (1) make this a public function and (2) pass in the explicit builder parameters?

briandealwis (commented):
LGTM, though I'd prefer the docker builder to follow the same pattern as the other builders rather than embed the `Builder` instance.  WDYT?

pkg/skaffold/build/local/local.go
gsquared94: The problem is that `local` package takes on too much responsibility. It implements both the builder platform (i.e local vs gcb vs cluster) and the per artifact builder (docker, jib, etc) methods. We could extract out a docker package for the single artifact builder and make it consistent with the other single artifact builder packages. But that's a refactor for another time. 🙃

I mainly made this change to be able to write UT for this switch block.

integration/testdata/build-dependencies/app1/build.sh
gsquared94: makes sense, I pirated this code block from another existing example. I'll fix it in both places.

pkg/skaffold/build/local/local.go
gsquared94: the interface is unexported due to the same reason.

gsquared94: >  But that's a refactor for another time. 🙃

On second thoughts: https://github.com/GoogleContainerTools/skaffold/pull/4986

MarlonGamez (approved):

Issue #4970, opened 2020-10-29
Documentation reference of skaffold.yaml is empty


hermanbanken:
Probably a mistake in the rendering, but https://skaffold.dev/docs/references/yaml/ is now completely unusable. It used to contain an annotated YAML structure of all configuration options, right?

![Screenshot 2020-10-29 at 10 31 01](https://user-images.githubusercontent.com/791189/97550394-ed002100-19d1-11eb-8e94-f82268131929.png)

hermanbanken:
area/docs

hermanbanken:
I was taking a look and noticed this in the source. Is that supposed to be autofilled using JavaScript?
```
<table id="table" data-version="{{% skaffold-version %}}"></table>
```

tejal29:
Sorry about this. 
We will work on fixing this ASAP

hermanbanken:
In order to prevent such issues in the future, is it an idea to add something new [1]/[2]? Now the whole spec was down, while only a few examples were broken.

1. test that checkes that the table can be filled from that spec
2. include a try-catch at least a little bit higher in the logic, to partially mitigate broken examples

hermanbanken:
I don't know where to put that. I see that no testing is in place for Hugo yet.

Maybe either:
- Add a 'check' step to https://github.com/GoogleContainerTools/skaffold/blob/master/deploy/docs/build.sh#L25
- Add a firebase predeploy check step
- Add a GitHub Action workflow that tests https://github.com/GoogleContainerTools/skaffold/blob/master/docs/content/en/docs/references/yaml/main.js

tejal29:
skaffold.dev is now fixed!

PR #4974, opened 2020-10-30
fix examples to appropriate type

IsaacPD:
Fixes: #4970 

**Description**

Change examples to be wrapped in an array and have the names be wrapped in quotations.


MarlonGamez (approved):
Thanks for pushing this through :)

tejal29:
Thanks. I will regenerate the docs!

Issue #4981, opened 2020-10-31
Deploy - KPT - kustomizeVersionRegexP does not support non-Zulu BuildDate


DennisVis:
### Expected behavior
Kustomize versions with a non-Zulu BuildDate and a supported version number should be accepted by the KPT deployer.

### Actual behavior
All `kustomize version` outputs that do not contain a Zulu time for the BuildDate field are treated as invalid by the KPT deployer:
```
unknown kustomize version {Version:kustomize/v3.8.5 GitCommit:4052cd4fd8c76a17b5f64e32509f3fba9713fe75 BuildDate:2020-10-08T07:05:45+01:00 GoOs:darwin GoArch:amd64}
```

### Information
- Skaffold version: v1.16.0
- Operating system: MacOS
- Kustomize version: v3.8.5

### Steps to reproduce the behavior
1. Install a version of Kustomize that has a non-Zulu BuildDate (kustomize/v3.8.5)
2. Run Skaffold with a kpt deploy

yuwenma:
Thanks for catching the issue and the quick fix, Dennis! 

(Do you mind letting me know how you get the kustomize version? I'm wondering why the kustomize version info is different 🤣.  

Your catch really reminds me that in case the users may want to build the kustomize themselves, we actually can further loosen the kustomize version regex. I'll upload a new PR soon.

DennisVis:
I think a warning would be a solid approach here. 

I discovered another discrepancy between the version expected by the KPT deployer and the output of my Kustomize instance; the `kustomize/` prefix. I hadn't even considered the case of users building their own version. So there's already quite a few edge cases to consider.

Doing a sanity check resulting in a warning, stating the minimum version required, would give users the needed info when the deploy breaks due to an unsupported Kustomize version.

PR #4982, opened 2020-10-31
support non-zulu kustomize build date

DennisVis:
Fixes: #4981 

**Description**
This PR changes the regex, which checks for Kustomization versions within the KPT deployer, to support non-Zulu values for the BuildDate field of the Kustomize version output.

### DIFF
--- a/pkg/skaffold/deploy/kpt/kpt.go
+++ b/pkg/skaffold/deploy/kpt/kpt.go
@@ -56,7 +56,7 @@ const (
 
 	kustomizeDownloadLink  = "https://kubernetes-sigs.github.io/kustomize/installation/"
 	kustomizeMinVersion    = "v3.2.3"
-	kustomizeVersionRegexP = `{Version:(\S+) GitCommit:\S+ BuildDate:\d{4}-\d{2}-\d{2}T\d\d:\d\d:\d\dZ GoOs:\S+ GoArch:\S+}`
+	kustomizeVersionRegexP = `{Version:(\S+) GitCommit:(\S+) BuildDate:\S+ GoOs:\S+ GoArch:\S+}`
 )
 
 // Deployer deploys workflows with kpt CLI
@@ -80,7 +80,7 @@ func NewDeployer(cfg types.Config, labels map[string]string) *Deployer {
 
 var sanityCheck = versionCheck
 
-// sanityCheck guarantees the kpt and kustomize versions are compatible with skaffold.
+// versionCheck guarantees the kpt is compatible with skaffold and warns if the kustomize version is not.
 func versionCheck(dir string) error {
 	kptCmd := exec.Command("kpt", "version")
 	out, err := util.RunCmdOut(kptCmd)
@@ -88,6 +88,7 @@ func versionCheck(dir string) error {
 		return fmt.Errorf("kpt is not installed yet\nSee kpt installation: %v",
 			kptDownloadLink)
 	}
+
 	version := strings.TrimSuffix(string(out), "\n")
 	// kpt follows semver but does not have "v" prefix.
 	if !semver.IsValid("v" + version) {
@@ -103,30 +104,43 @@ func versionCheck(dir string) error {
 	// Users can choose not to use kustomize in kpt deployer mode. We only check the kustomize
 	// version when kustomization.yaml config is directed under .deploy.kpt.dir path.
 	_, err = kustomize.FindKustomizationConfig(dir)
-	if err == nil {
-		kustomizeCmd := exec.Command("kustomize", "version")
-		out, err := util.RunCmdOut(kustomizeCmd)
-		if err != nil {
-			return fmt.Errorf("kustomize is not installed yet\nSee kpt installation: %v",
-				kustomizeDownloadLink)
-		}
-		versionInfo := strings.TrimSuffix(string(out), "\n")
-		// Kustomize version information is in the form of
-		// {Version:$VERSION GitCommit:$COMMIT BuildDate:1970-01-01T00:00:00Z GoOs:darwin GoArch:amd64}
-		re := regexp.MustCompile(kustomizeVersionRegexP)
-		match := re.FindStringSubmatch(versionInfo)
-		if len(match) != 2 {
-			return fmt.Errorf("unknown kustomize version %v\nPlease upgrade your "+
+	if err != nil {
+		return nil
+	}
+
+	kustomizeCmd := exec.Command("kustomize", "version")
+	out, err = util.RunCmdOut(kustomizeCmd)
+	if err != nil {
+		return fmt.Errorf(
+			"kustomize is not installed yet\nSee kpt installation: %v",
+			kustomizeDownloadLink)
+	}
+
+	versionInfo := strings.TrimSuffix(string(out), "\n")
+	// Kustomize version information is in the form of
+	// {Version:$VERSION GitCommit:$COMMIT BuildDate:1970-01-01T00:00:00Z GoOs:darwin GoArch:amd64}
+	re := regexp.MustCompile(kustomizeVersionRegexP)
+	match := re.FindStringSubmatch(versionInfo)
+	if len(match) != 3 {
+		fmt.Printf(
+			"unknown kustomize version. \nPlease upgrade your "+
 				"local kustomize CLI to a version >= %v\nSee kustomize installation: %v",
-				string(out), kustomizeMinVersion, kustomizeDownloadLink)
-		}
-		if !semver.IsValid(match[1]) || semver.Compare(match[1], kustomizeMinVersion) < 0 {
-			return fmt.Errorf("you are using kustomize %q\n"+
+			kustomizeMinVersion, kustomizeDownloadLink)
+	} else {
+		version := strings.TrimPrefix(match[1], "kustomize/")
+		if !semver.IsValid(version) {
+			fmt.Printf(
+				"unknown kustomize version with commit %v\nPlease upgrade your "+
+					"local kustomize CLI to a version >= %v\nSee kustomize installation: %v",
+				match[2], kustomizeMinVersion, kustomizeDownloadLink)
+		} else if semver.Compare(version, kustomizeMinVersion) < 0 {
+			fmt.Printf("you are using kustomize %q\n"+
 				"Please update your kustomize version to >= %v\n"+
-				"See kustomize installation: %v", match[1], kustomizeMinVersion,
-				kustomizeDownloadLink)
+				"See kustomize installation: %v",
+				version, kustomizeMinVersion, kustomizeDownloadLink)
 		}
 	}
+
 	return nil
 }
 
--- a/pkg/skaffold/deploy/kpt/kpt_test.go
+++ b/pkg/skaffold/deploy/kpt/kpt_test.go
@@ -1087,22 +1087,58 @@ func TestVersionCheck(t *testing.T) {
 				AndRunOut("kustomize version", `{Version:v0.0.1 GitCommit:a0072a2cf92bf5399565e84c621e1e7c5c1f1094 BuildDate:2020-06-15T20:19:07Z GoOs:darwin GoArch:amd64}`),
 			kustomizations: map[string]string{"Kustomization": `resources:
 				- foo.yaml`},
-			shouldErr: true,
-			error: fmt.Errorf("you are using kustomize \"v0.0.1\"\n"+
-				"Please update your kustomize version to >= %v\n"+
-				"See kustomize installation: %v", kustomizeMinVersion, kustomizeDownloadLink),
+			shouldErr: false,
+			error:     nil,
 		},
 		{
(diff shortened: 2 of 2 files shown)


pkg/skaffold/deploy/kpt/kpt_test.go
```diff
@@ -1104,6 +1104,16 @@ func TestVersionCheck(t *testing.T) {
 				"local kustomize CLI to a version >= %v\nSee kustomize installation: %v",
 				kustomizeMinVersion, kustomizeDownloadLink),
 		},
+		{
```
yuwenma, line 1107: To fix the codedev blocker, suggest adding a sad path using the kustomize version value 
`{Version:unknown GitCommit:a0072a2cf92bf5399565e84c621e1e7c5c1f1094 BuildDate:2020-06-15T20:19:07+01:00 GoOs:darwin GoArch:amd64}`

yuwenma:
PR lgtm. You can add one more unittest to fix the codecov failure.

yuwenma (approved):
PR lgtm. You can add a sad path to fix the codecov failure, or just go ahead and merge it. I'll change the regex in the next PR so it does not care the build time.

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -56,7 +56,7 @@ const (
 
 	kustomizeDownloadLink  = "https://kubernetes-sigs.github.io/kustomize/installation/"
 	kustomizeMinVersion    = "v3.2.3"
-	kustomizeVersionRegexP = `{Version:(\S+) GitCommit:\S+ BuildDate:\d{4}-\d{2}-\d{2}T\d\d:\d\d:\d\dZ GoOs:\S+ GoArch:\S+}`
+	kustomizeVersionRegexP = `{Version:(\S+) GitCommit:\S+ BuildDate:\d{4}-\d{2}-\d{2}T\d\d:\d\d:\d\d(Z|[+-]\d\d:\d\d) GoOs:\S+ GoArch:\S+}`
```
yuwenma, line 59: I see the problem. You may not want to add a "group" to the BuildDate tail. The new logic (group > 3) will never be touched. 

For the previous logic, the group count can be either 0 or 2 meaning it either matches nothing (0) or matches exactly (group[0] is the whole string, and group[1] is the Version part). 😸

yuwenma:
Somehow I can't find my previous comment. @DennisVis do you mind making the following changes:

1. Change the regex to be "`{Version:(\S+) GitCommit:(\S+) BuildDate:\S+ GoOs:\S+ GoArch:\S+}`" and the group count should be matching 3?

2. unknown kustomize should not return error but warn users saying this is unknown version (showing group[2] which is the GitCommit sha1). Requesting them to make sure this version is >= v3.4.3 

3. older kustomize should not return error but warn the users that this is an old version and the feature may not be compatible with the latest skaffold/kpt functions.

yuwenma (changes requested):
Somehow I can't find my previous comment. @DennisVis do you mind making the following changes:

Change the regex to be "{Version:(\S+) GitCommit:(\S+) BuildDate:\S+ GoOs:\S+ GoArch:\S+}" and the group count should be matching 3?

unknown kustomize should not return error but warn users saying this is unknown version (showing group[2] which is the GitCommit sha1). Requesting them to make sure this version is >= v3.4.3

older kustomize should not return error but warn the users that this is an old version and the feature may not be compatible with the latest skaffold/kpt functions.

DennisVis:
I made sure the kustomize version errors are treated as warnings, except for when it's completely missing. Using `fmt.Printf()` to output those warnings, please let me know if there's a better way.

Also took a different approach to the parsing by replacing the regex with a little parse function, which also takes the `kustomize/` prefix into account. If you have any reservations about that I can put the regex back and use that.

yuwenma:
Hi @DennisVis, we are doing some extended work related to kpt and kustomize, so the structure of this kpt.go should not be changed. Do you mind only changing the code inside the versionCheck function and not define new functions? Otherwise, it will make the future work much harder and I guess we would eventually need to revert your change.

If this takes you too much time, please feel free to re-assign the issue 4981 to me. If this blocks you usage, let me know and I'll prioritize the hot-fix.

DennisVis:
No problem. Reverted back to using the regex and made no changes outside of the versionCheck function. The tests make a lot less sense this way, but they should cover all paths.

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -103,30 +104,43 @@ func versionCheck(dir string) error {
...
-			return fmt.Errorf("unknown kustomize version %v\nPlease upgrade your "+
+	if err != nil {
+		return nil
+	}
+
+	kustomizeCmd := exec.Command("kustomize", "version")
+	out, err = util.RunCmdOut(kustomizeCmd)
+	if err != nil {
+		return fmt.Errorf(
+			"kustomize is not installed yet\nSee kpt installation: %v",
```
yuwenma, line 115: typo: See kustomize installation.

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -103,30 +104,43 @@ func versionCheck(dir string) error {
...
+			kustomizeDownloadLink)
+	}
+
+	versionInfo := strings.TrimSuffix(string(out), "\n")
+	// Kustomize version information is in the form of
+	// {Version:$VERSION GitCommit:$COMMIT BuildDate:1970-01-01T00:00:00Z GoOs:darwin GoArch:amd64}
```
yuwenma, line 121: Can you update this comment as well?

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -103,30 +104,43 @@ func versionCheck(dir string) error {
...
+	if len(match) != 3 {
+		fmt.Printf(
+			"unknown kustomize version. \nPlease upgrade your "+
 				"local kustomize CLI to a version >= %v\nSee kustomize installation: %v",
-				string(out), kustomizeMinVersion, kustomizeDownloadLink)
-		}
-		if !semver.IsValid(match[1]) || semver.Compare(match[1], kustomizeMinVersion) < 0 {
-			return fmt.Errorf("you are using kustomize %q\n"+
+			kustomizeMinVersion, kustomizeDownloadLink)
+	} else {
```
yuwenma, line 129: nit: this else is not necessaroy.

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -103,30 +104,43 @@ func versionCheck(dir string) error {
...
+		version := strings.TrimPrefix(match[1], "kustomize/")
```
yuwenma, line 130: May I ask where you get your kustomize CLI and which os you use? I don't feel like the "kustomize/" prefix should be handled specifically, unless it's an official version.

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -103,30 +104,43 @@ func versionCheck(dir string) error {
...
-				kustomizeDownloadLink)
-		}
-		versionInfo := strings.TrimSuffix(string(out), "\n")
-		// Kustomize version information is in the form of
-		// {Version:$VERSION GitCommit:$COMMIT BuildDate:1970-01-01T00:00:00Z GoOs:darwin GoArch:amd64}
-		re := regexp.MustCompile(kustomizeVersionRegexP)
-		match := re.FindStringSubmatch(versionInfo)
-		if len(match) != 2 {
-			return fmt.Errorf("unknown kustomize version %v\nPlease upgrade your "+
+	if err != nil {
```
yuwenma, line 107: super nit: normally we don't change the code structure unless it's really bad, so as the code diff only reflects the real logic change. And this makes the code easier to review and have less potential bugs. Is there by any chance you can keep the (if err == nil) format?

pkg/skaffold/deploy/kpt/kpt_test.go
```diff
@@ -1087,22 +1087,58 @@ func TestVersionCheck(t *testing.T) {
 				AndRunOut("kustomize version", `{Version:v0.0.1 GitCommit:a0072a2cf92bf5399565e84c621e1e7c5c1f1094 BuildDate:2020-06-15T20:19:07Z GoOs:darwin GoArch:amd64}`),
 			kustomizations: map[string]string{"Kustomization": `resources:
 				- foo.yaml`},
-			shouldErr: true,
-			error: fmt.Errorf("you are using kustomize \"v0.0.1\"\n"+
-				"Please update your kustomize version to >= %v\n"+
-				"See kustomize installation: %v", kustomizeMinVersion, kustomizeDownloadLink),
+			shouldErr: false,
+			error:     nil,
```
yuwenma, line 1091: can you find a way to check if the warn message is correct?

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -103,30 +104,43 @@ func versionCheck(dir string) error {
...
+			"unknown kustomize version. \nPlease upgrade your "+
 				"local kustomize CLI to a version >= %v\nSee kustomize installation: %v",
-				string(out), kustomizeMinVersion, kustomizeDownloadLink)
-		}
-		if !semver.IsValid(match[1]) || semver.Compare(match[1], kustomizeMinVersion) < 0 {
-			return fmt.Errorf("you are using kustomize %q\n"+
+			kustomizeMinVersion, kustomizeDownloadLink)
+	} else {
+		version := strings.TrimPrefix(match[1], "kustomize/")
+		if !semver.IsValid(version) {
```
yuwenma, line 131: This is skeptical. We shall not ask users to upgrade if the version is not SemVer2, but verifying the SHA1 is new enough (showing the sha1 here, kustomizeMinVersion should older than the sha1).

pkg/skaffold/deploy/kpt/kpt.go
DennisVis: That would cause multiple messages to be printed in the case of a version which doesn't provide us with 3 match groups. This else makes sure that only the 'unknown version' message is being output in that situation.

pkg/skaffold/deploy/kpt/kpt.go
DennisVis: I install it through brew which is an [official channel](https://kubernetes-sigs.github.io/kustomize/installation/homebrew/). Just upgraded to 3.8.6 and this is the version output:

```{Version:kustomize/v3.8.6 GitCommit:c1747439cd8bc956028ad483cdb30d9273c18b24 BuildDate:2020-10-31T17:13:37+00:00 GoOs:darwin GoArch:amd64}```

File: pkg/skaffold/deploy/kpt/kpt.go (hunk 9 above)
Comment (DennisVis) [reply pkg/skaffold/deploy/kpt/kpt.go:131]: I'm afraid I don't quite follow. Could you elaborate a bit more? Should I drop the semver check altogether and do the comparison on git commit alone?

File: pkg/skaffold/deploy/kpt/kpt_test.go (hunk 8 above)
Comment (DennisVis) [reply pkg/skaffold/deploy/kpt/kpt_test.go:1091]: The only way to do that without changing the function is to catch the stdout. I've implemented test utils for this locally and they work well. A `SetStdout()` function to catch the stdout in a temporary file and a set of `CheckLog*()` functions that mimic the `CheckError*()` functionalities.

These are obviously best placed in the 'testutil/util.go' but seeing your previous reservations about code changes I'm guessing that is not a desirable way to go. I could move them into `kpt_test.go` file but that will make things more messy imo.

What do you advise?

**Comment by yuwenma** [comment]
No worries @DennisVis I guess code itself can explain better. I uploaded the quick fix here https://github.com/GoogleContainerTools/skaffold/pull/4994 Basically you can either use fmt.Fprintf or the color lib (what skaffold uses) to output the user message. Do you mind answering my previous question, how did you get your kustomize CLI (I'm interested in how your `kustomize version` has the `kustomize/` prefix)?

**Comment by DennisVis** [comment]
Sure, I already posted an answer to that below your comment but I'll repost it here:

```
I install it through brew which is an [official channel](https://kubernetes-sigs.github.io/kustomize/installation/homebrew/). Just upgraded to 3.8.6 and this is the version output:

{Version:kustomize/v3.8.6 GitCommit:c1747439cd8bc956028ad483cdb30d9273c18b24 BuildDate:2020-10-31T17:13:37+00:00 GoOs:darwin GoArch:amd64}
```

yuwenma:
> Sure, I already posted an answer to that below your comment but I'll repost it here:
> 
> I install it through brew which is an [official channel](https://kubernetes-sigs.github.io/kustomize/installation/homebrew/). Just upgraded to 3.8.6 and this is the version output:
> 
> {Version:kustomize/v3.8.6 GitCommit:c1747439cd8bc956028ad483cdb30d9273c18b24 BuildDate:2020-10-31T17:13:37+00:00 GoOs:darwin GoArch:amd64}

I see. Thank you for letting me know.

IsaacPD:
Closed in favor of #4994

Issue #4913, opened 2020-10-14
Output logs from `--mute-logs` contain ANSI sequences


briandealwis:
Our [colour output formatter functions](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/color/formatter.go#L104-L122) don't check that the given output is the same as the [output used to determine if it's a terminal and supports colour](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/color/formatter.go#L39).  So when we log output using `--mute-logs=status-check`, for example, the files contain ANSI sequences:
```
^[34mWaiting for deployments to stabilize... ^[0m
 - deployment/java-hello-world: FailedMount: MountVolume.SetUp failed for volume "default-token-hrwdq" : failed to sync secret cache: timed out waiting for the condition
    - pod/java-hello-world-6676dc5b65-f6xl9: FailedMount: MountVolume.SetUp failed for volume "default-token-hrwdq" : failed to sync secret cache: timed out waiting for the condition
 - deployment/java-hello-world is ready.
^[34mDeployments stabilized in 13.161922025s ^[0m
```

`Fprintln()` and `Fprintf()` should check that the output is the original console output instance.

PR #4990, opened 2020-11-03
Do not color file output with `--muted-logs`

IsaacPD:
Fixes: #4913 

**Description**

Wrap os.File in filelog.MutedLogger type that is checked on Color.Fprint to determine whether colors should be printed.

### **Example of change**

Running ```skaffold run --mute-logs=all``` in skaffold/examples/ruby
**Before**: 
```
...
Starting deploy...
- writing logs to /var/.../skaffold/deploy/2020-11-02_18-49-06.log
...
```
In /var/.../skaffold/deploy/2020-11-02_18-49-06.log
```
^[[34mStarting deploy...^[[0m
 - service/ruby configured
 - deployment.apps/ruby configured
```

The logs contain ANSI escape codes which do not represent the actual content intended to be logged.

**After**:
```
...
Starting deploy...
- writing logs to /var/.../skaffold/deploy/2020-11-02_18-53-51.log
...
```
In /var/.../skaffold/deploy/2020-11-02_18-53-51.log
```
Starting deploy...
 - service/ruby configured
 - deployment.apps/ruby configured
```

The only thing logged is the intended output.

### DIFF
--- a/pkg/skaffold/color/formatter.go
+++ b/pkg/skaffold/color/formatter.go
@@ -18,6 +18,7 @@ package color
 
 import (
 	"fmt"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/logfile"
 	"io"
 	"strings"
 
@@ -34,8 +35,6 @@ const DefaultColorCode = 34
 func init() {
 	colors.Disable(true)
 }
-
-var colorfulWriter io.Writer
 
 // SetupColors enables/disables coloured output.
 func SetupColors(out io.Writer, defaultColor int, forceColors bool) {
@@ -64,7 +63,6 @@ func SetupColors(out io.Writer, defaultColor int, forceColors bool) {
 		37: White,
 		0:  None,
 	}[defaultColor]
-	colorfulWriter = out
 }
 
 // Color can be used to format text so it can be printed to the terminal in color.
@@ -106,7 +104,7 @@ var (
 
 // Fprintln outputs the result to out, followed by a newline.
 func (c Color) Fprintln(out io.Writer, a ...interface{}) {
-	if c.color == nil || out != colorfulWriter {
+	if c.color == nil || isMuted(out) {
 		fmt.Fprintln(out, a...)
 		return
 	}
@@ -116,10 +114,19 @@ func (c Color) Fprintln(out io.Writer, a ...interface{}) {
 
 // Fprintf outputs the result to out.
 func (c Color) Fprintf(out io.Writer, format string, a ...interface{}) {
-	if c.color == nil || out != colorfulWriter {
+	if c.color == nil || isMuted(out) {
 		fmt.Fprintf(out, format, a...)
 		return
 	}
 
 	fmt.Fprint(out, c.color.Sprintf(format, a...))
+}
+
+func isMuted(out io.Writer) bool {
+	switch out.(type) {
+	case *logfile.MutedLogger:
+		return true
+	default:
+		return false
+	}
 }
--- a/pkg/skaffold/color/formatter.go
+++ b/pkg/skaffold/color/formatter.go
@@ -35,6 +35,8 @@ func init() {
 	colors.Disable(true)
 }
 
+var colorfulWriter io.Writer
+
 // SetupColors enables/disables coloured output.
 func SetupColors(out io.Writer, defaultColor int, forceColors bool) {
 	_, isTerm := util.IsTerminal(out)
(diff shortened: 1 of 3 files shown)


MarlonGamez:
Would this cause issues in a situation where output is written to a temporary buffer which is then dumped to the terminal?

IsaacPD:
> @MarlonGamez Would this cause issues in a situation where output is written to a temporary buffer which is then dumped to the terminal?

As long as the write to the terminal is done with ```Color.Fprintf/Fprintln``` then there should be no issues with coloring to the terminal.

```go
var b bytes.Buffer
Blue.Fprintf(&b, "some blue text")
fmt.Fprintf(out, b)
``` 

In the case above the final output would not have any special coloring in "out" if that's what you're referring to.

tejal29 (approved):

MarlonGamez:
When performing parallel builds, we're printing in a way that will result in a loss of colored logs. I believe each artifact being built basically has it's own buffer which is written to with a mix of colored and non-colored lines, and each buffer is then dumped to the terminal

IsaacPD:
Yeah I would not like to lose colored output, so I'll see what approaches are open to preserve that.

tejal29:
Thanks @MarlonGamez for going through the build in parallel flow. You are right, with the way to log for parallel builds, color output will be broken.

Wondering if we should create a `MuteLogger` object and use typecheck instead of the check `out != colorfulWriter `

IsaacPD:
That approach seems alright with me as it tackles the underlying issue. I cannot really think of any other ways that ANSI output my showup but i'll rename this issue to reflect that it does not potentially cover every case.

pkg/skaffold/color/formatter.go
```diff
@@ -113,10 +114,19 @@ func (c Color) Fprintln(out io.Writer, a ...interface{}) {
...
-	if c.color == nil {
+	if c.color == nil || isMuted(out) {
 		fmt.Fprintf(out, format, a...)
 		return
 	}
 
 	fmt.Fprint(out, c.color.Sprintf(format, a...))
 }
+
+func isMuted(out io.Writer) bool {
```
briandealwis, line 125: Would it make sense to push this into ‘logfile‘?

briandealwis (commented):
So thinking about it some more, this approach will continue to add colour sequences to other file output. I think we need to instead record the actual terminal io.Writer instance and compare the provided writer.

briandealwis:
Ah, missed the other discussion on this PR. We may be hosed on parallel output since we're redirecting the output and so the underlying builders won't colour their output (e.g., maven and gradle don’t colour their output).

briandealwis:
I'm thinking that we should create a coloured output: rather than our current practice:
```
func Xxxx(out io.Writer, ...) {
   color.Blue.Fprintf(out, "some blue text")
}
```

and trying to guess whether the output supports color, we instead create a Writer wrapper that carries whether it supports colour or not.
```
func main() {
   out := color.NewWriter(os.Stdout)
   Xxx(out)

   var b bytes.Buffer
   if color.isTerm(os.Stdout) { // or something like this
      b = color.NewByteBuffer(b)
   }
   Xxx(&b)
   b.WriteTo(os.Stdout)
}

func Xxxx(out io.Writer, ...) {
   color.Blue.Fprintf(out, "some blue text") // Fprintf could then see if the writer supports colour
}
```

Looks like we can force [Maven (for Jib) to colour output](https://mihai-nita.net/2020/08/23/force-maven-colors-in-eclipse-console/) with `-Dstyle.color=always -Djansi.passthrough=true`.  I'm sure there are equivalents for Gradle. 😱

IsaacPD:
I think this approach makes sense. So in this instance, https://github.com/GoogleContainerTools/skaffold/blob/d667827c018c592abcfcdd5738f42eb5848adb48/pkg/skaffold/build/result.go#L56-L65

Something like this would be added to determine whether to color builds based on whether the final output (stdout or a log file) is also colorable.
```go
if color.isColorable(l.out) {
    w := color.NewWriter(w)
}
```

briandealwis:
I just came across [github.com/creack/pty](https://github.com/creack/pty) which runs commands under a PTY so that commands are run under a pseudo terminal.  This seems like the right solution for parallel-mode build output.

IsaacPD:
I was having some issues earlier trying to use these flags you gave @briandealwis "-Dstyle.color=always -Djansi.passthrough=true" since they weren't working, but it turns out the order of the flags is relevant and if these were at the end then they didn't do anything. It might not be necessary then to use https://github.com/creack/pty

IsaacPD:
Closing in favor of #5014

PR #4994, opened 2020-11-04
Loose kustomize version requirements

yuwenma:
Fixes: #4981 #3904 

**Description**
We warn users if kustomize version is non-official or too old. We don't block users using an old kustomize version since we believe kpt can better deal with the kustomize versioning once using a builtin kustomize .

### DIFF
--- a/pkg/skaffold/deploy/kpt/kpt.go
+++ b/pkg/skaffold/deploy/kpt/kpt.go
@@ -88,7 +88,7 @@ func NewDeployer(cfg types.Config, labels map[string]string) *Deployer {
 
 var sanityCheck = versionCheck
 
-// versionCheck guarantees the kpt and kustomize versions are compatible with skaffold.
+// versionCheck returns an error if the kpt and kustomize versions are not compatible with skaffold.
 func versionCheck(dir string, stdout io.Writer) error {
 	kptCmd := exec.Command("kpt", "version")
 	out, err := util.RunCmdOut(kptCmd)
--- a/pkg/skaffold/deploy/kpt/kpt.go
+++ b/pkg/skaffold/deploy/kpt/kpt.go
@@ -34,6 +34,7 @@ import (
 	k8syaml "sigs.k8s.io/yaml"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy/kustomize"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy/types"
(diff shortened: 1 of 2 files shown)


pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -80,8 +88,8 @@ func NewDeployer(cfg types.Config, labels map[string]string) *Deployer {
 
 var sanityCheck = versionCheck
 
-// sanityCheck guarantees the kpt and kustomize versions are compatible with skaffold.
-func versionCheck(dir string) error {
+// versionCheck guarantees the kpt and kustomize versions are compatible with skaffold.
```
briandealwis, line 91: We're not guaranteeing any more.
```suggestion
// versionCheck returns an error if the kpt and kustomize versions are not compatible with skaffold.
```

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -116,15 +124,12 @@ func versionCheck(dir string) error {
 		re := regexp.MustCompile(kustomizeVersionRegexP)
 		match := re.FindStringSubmatch(versionInfo)
 		if len(match) != 2 {
-			return fmt.Errorf("unknown kustomize version %v\nPlease upgrade your "+
```
briandealwis, line 119: WDYT about moving this kustomize version parsing logic into the `kustomize` package, similar to how we do `kubectl.CompareVersionTo()` checks?   And similarly extracting the `kpt` version checking too?  Then the version-checking can be tested in isolation and not involve string matching against the warning text.

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -56,7 +57,14 @@ const (
...
 	kustomizeMinVersion    = "v3.2.3"
-	kustomizeVersionRegexP = `{Version:(\S+) GitCommit:\S+ BuildDate:\d{4}-\d{2}-\d{2}T\d\d:\d\d:\d\dZ GoOs:\S+ GoArch:\S+}`
+	kustomizeVersionRegexP = `{Version:(\S+) GitCommit:\S+ BuildDate:\S+ GoOs:\S+ GoArch:\S+}`
+)
+
+var (
+	kustomizeFurtherGuidance = fmt.Sprintf("Please make sure your local "+
+		"kustomize version >= the official version %v, otherwise some features may not be "+
+		"well supported. You can download the official version "+
+		"from %v", kustomizeMinVersion, kustomizeDownloadLink)
```
briandealwis, line 67: I don't think it makes sense to have this text as a var since the arguments as positional instead of named.  I think you're assigning this as a var to be used from the tests.

@tejal29 @PriyaModali here's a possible contribution to build failures that we should report upwards in our metrics: incompatible dependency versions.  We have this for kubectl, helm, and now kpt and kustomize.

pkg/skaffold/deploy/kpt/kpt.go
yuwenma: Very reasonable suggestion. We can add similar logic to kustomize package, but we may want to separate the version check from kpt. This is because we will replace the kustomize CLI to builtin kustomize package in kpt. This is pending until the unified kpt/kustomize design is done in the kpt team.

pkg/skaffold/deploy/kpt/kpt.go
yuwenma: I don't have strong opinions on the var usage. Yeah, this is just to save some lines in the unitests.

yuwenma: Updated.

yuwenma: For the incompatible dependency versions, I guess it is fine to ignore the conflict if users use kustomize in kpt deployer (like this PR only warns users but doesn't have hard requirements on actions) . This is because we will replace the kustomize here to some kpt builtin functions, so this version check may likely be changed someday in Q1 2021.

MarlonGamez:
I think this is looking good now. +1 to Brian's idea about separating the kustomize version checking into the kustomize package, but that should probably be in a different PR. I'll run kokoro and wait on the result

MarlonGamez (approved):

pkg/skaffold/deploy/kpt/kpt.go
tejal29: Ack @briandealwis . 
This is something i discussed with @tstromberg internally. Skaffold needs to provide a framework or mechanism to error out with an error code instead of playing this catch up game for new error messages.
Minikube has this implemented [here](https://github.com/kubernetes/minikube/tree/c459fea74df093f916e644fac55646755ea7f4ec/pkg/minikube/exit)

I am working on this idea. Happy to hear any thoughts.

Issue #4786, opened 2020-09-15
Deploying to EKS gives an error


devi-sandireddy:
```
Starting deploy...
Helm release myapp not installed. Installing...
NAME: myapp
LAST DEPLOYED: Tue Sep 15 12:55:30 2020
NAMESPACE: test-app
STATUS: deployed
REVISION: 1
TEST SUITE: None
panic: interface conversion: interface {} is nil, not string

goroutine 1 [running]:
github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy/kubectl.(*namespaceCollector).Visit(0xc0000100b0, 0xc00099c930, 0xc00064a578, 0x8, 0x57684c0, 0xc00099ca20, 0xc00064a501)
        /private/tmp/skaffold-20200903-47925-9prb4r/pkg/skaffold/deploy/kubectl/namespaces.go:61 +0x12c
github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy/kubectl.visitFields(0x57684c0, 0xc00099c930, 0x5d24220, 0xc0000100b0)
        /private/tmp/skaffold-20200903-47925-9prb4r/pkg/skaffold/deploy/kubectl/visitor.go:133 +0x103
github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy/kubectl.traverseManifestFields(0xc00099c930, 0x5d24220, 0xc0000100b0)
        /private/tmp/skaffold-20200903-47925-9prb4r/pkg/skaffold/deploy/kubectl/visitor.go:83 +0x65
github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy/kubectl.(*ManifestList).Visit(0xc000d5a768, 0x5d24220, 0xc0000100b0, 0x40fbc59, 0xc00099f2c0, 0xc000941444, 0x3, 0x3)
        /private/tmp/skaffold-20200903-47925-9prb4r/pkg/skaffold/deploy/kubectl/visitor.go:65 +0x1e5
github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy/kubectl.(*ManifestList).CollectNamespaces(0xc000d5a768, 0x104, 0x12d, 0xc000d5a75e, 0xa, 0xa)
        /private/tmp/skaffold-20200903-47925-9prb4r/pkg/skaffold/deploy/kubectl/namespaces.go:29 +0x90
[...]

PR #5001, opened 2020-11-05
Added logic to handle nil during interface conversion of namespace.

PriyaModali:
Fixes: #4786 

**Description**
Added logic to handle nil during interface conversion of namespace

User will not receive the below error when namespace is nil.
`panic: interface conversion: interface {} is nil, not string`

Added unit test.

### DIFF
--- a/pkg/skaffold/kubernetes/manifest/namespaces.go
+++ b/pkg/skaffold/kubernetes/manifest/namespaces.go
@@ -58,7 +58,11 @@ func (r *namespaceCollector) Visit(o map[string]interface{}, k string, v interfa
 		return true
 	}
 	if nsValue, present := metadata["namespace"]; present {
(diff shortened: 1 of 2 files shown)


pkg/skaffold/kubernetes/manifest/namespaces.go
```diff
@@ -58,7 +58,11 @@ func (r *namespaceCollector) Visit(o map[string]interface{}, k string, v interfa
 		return true
 	}
 	if nsValue, present := metadata["namespace"]; present {
-		if ns := strings.TrimSpace(nsValue.(string)); ns != "" {
+		nsString, isOk := nsValue.(string)
+		if !isOk || nsString == "" {
```
gsquared94, line 62:
```suggestion
		nsString, ok := nsValue.(string)
		if !ok || nsString == "" {
```

gsquared94 (approved):
lgtm!

PR #5002, opened 2020-11-06
Fix Kaniko build args eval from config `env`.

gsquared94:
Kaniko builder config has an `env` section that wasn't getting evaluated correctly for the build args after my recent changes to introduce required artifacts as build args.

For a `kaniko build` step like:
```yaml
build:
  artifacts:
    - image: skaffold-example
      kaniko:
        env:
          - name: foo
            value: bar
        buildArgs:
          - arg1: {{ .foo }}
```
the value of `arg1` is evaluated twice, once with only `os.Environ` without the `env` in `cluster.go`:
```go
buildArgs, err := docker.EvalBuildArgs(b.mode, a.Workspace, a.KanikoArtifact.DockerfilePath, a.KanikoArtifact.BuildArgs, requiredImages)
```

later again with `env` in `pod.go`:
```go
buildArgs, err := docker.EvaluateBuildArgs(artifact.BuildArgs, envMapFromVars(artifact.Env))
```

But the first evaluation will resolve all `Go templates` and the later evaluation will not work to resolve key value pairs defined in `env` above. So the result is that `arg1` is applied to the builder with the value `""`.

The fix is to remove the duplicate code `docker. EvaluateBuildArgs` which attempts to do the same things as `docker.EvalBuildArgs`. So we define `EvalBuildArgs` and `EvalBuildArgsWithEnv` as an overload of that method.

### DIFF
--- a/pkg/skaffold/build/cluster/cluster.go
+++ b/pkg/skaffold/build/cluster/cluster.go
@@ -66,12 +66,7 @@ func (b *Builder) runBuildForArtifact(ctx context.Context, out io.Writer, a *lat
 	requiredImages := docker.ResolveDependencyImages(a.Dependencies, b.artifactStore, true)
 	switch {
 	case a.KanikoArtifact != nil:
-		buildArgs, err := docker.EvalBuildArgs(b.mode, a.Workspace, a.KanikoArtifact.DockerfilePath, a.KanikoArtifact.BuildArgs, requiredImages)
-		if err != nil {
-			return "", fmt.Errorf("unable to evaluate build args: %w", err)
-		}
-		a.KanikoArtifact.BuildArgs = buildArgs
-		return b.buildWithKaniko(ctx, out, a.Workspace, a.KanikoArtifact, tag)
+		return b.buildWithKaniko(ctx, out, a.Workspace, a.KanikoArtifact, tag, requiredImages)
 
 	case a.CustomArtifact != nil:
 		return custom.NewArtifactBuilder(nil, b.cfg, true, append(b.retrieveExtraEnv(), util.EnvPtrMapToSlice(requiredImages, "=")...)).Build(ctx, out, a, tag)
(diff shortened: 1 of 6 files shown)


tejal29:
> Kaniko builder config has an `env` section that wasn't getting evaluated correctly for the build args after my recent changes to introduce required artifacts as build args. This PR fixes that behavior.

Can you please describe what the bug was in more detail and what did you fix?

Its very hard to identify the problem from the code change since you also ended up moving code.

gsquared94:
> > Kaniko builder config has an `env` section that wasn't getting evaluated correctly for the build args after my recent changes to introduce required artifacts as build args. This PR fixes that behavior.
> 
> Can you please describe what the bug was in more detail and what did you fix?
> 
> Its very hard to identify the problem from the code change since you also ended up moving code.

For the a `kaniko build` step like:
```yaml
build:
  artifacts:
    - image: skaffold-example
      kaniko:
        env:
          - name: foo
             value: bar
        buildArgs:
          - arg1: {{ .foo }}
```
the value of `arg1` is evaluated twice, once with only `os.Environ` without the `env` in `cluster.go`:
```go
buildArgs, err := docker.EvalBuildArgs(b.mode, a.Workspace, a.KanikoArtifact.DockerfilePath, a.KanikoArtifact.BuildArgs, requiredImages)
```

later again with `env` in `pod.go`:
```go
buildArgs, err := docker.EvaluateBuildArgs(artifact.BuildArgs, envMapFromVars(artifact.Env))
```

But the first evaluation will resolve all `Go templates` and the later evaluation will not work to resolve key value pairs defined in `env` above. So the result is that `arg1` is applied to the builder with the value `""`.

The fix is to remove the duplicate code `docker. EvaluateBuildArgs` which attempts to do the same things as `docker.EvalBuildArgs`. So we define `EvalBuildArgs` and `EvalBuildArgsWithEnv` as an overload of that method.

Let me know if I should add any more detail.

tejal29 (approved):

Issue #4464, opened 2020-07-13
`port-forward` should be able to select ports by service name


briandealwis:
Currently port-forwards are established by specifying the remote port as an integer.  Since `containerPorts` can be defined with a [`name`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#containerport-v1-core), and referenced by [`Service` definitions by name (`targetPort`)](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#serviceport-v1-core), we should allow referencing ports by service name too.

Motivated by discsussion in #4419.

PR #5009, opened 2020-11-08
`port-forward` should be able to select ports by service name

anshlykov:
Fixes: #4464 
**Merge after:** #5010

**Description**

This PR introduced the new type `IntOrString` and changed the type of `PortForwardResource.Port` to `IntOrString`. This change allows a named port or an int port to be used in the `kubectl port-forward ...` command.

Also, there are some breaking changes in the proto schema and changes in logging. I think this can be simply resolved but I have not done it yet because I want to validate my approach first (or maybe it is already OK)


proto/skaffold.proto
```diff
@@ -190,7 +190,7 @@ message ResourceStatusCheckEvent {
 // PortEvent Event describes each port forwarding event.
 message PortEvent {
     int32 localPort = 1; // local port for forwarded resource
-    int32 remotePort = 2; // remote port is the resource port that will be forwarded.
+    IntOrString remotePort = 2; // remote port is the resource port that will be forwarded.
```
tejal29, line 193: unfortunately, this will be a backward incompatible change.  Can we instead add a another field and mark this as deprecated.

pkg/skaffold/event/event.go
```diff
@@ -382,12 +383,16 @@ func FileSyncSucceeded(fileCount int, image string) {
 }
 
 // PortForwarded notifies that a remote port has been forwarded locally.
-func PortForwarded(localPort, remotePort int32, podName, containerName, namespace string, portName string, resourceType, resourceName, address string) {
+func PortForwarded(localPort int32, remotePort util.IntOrString, podName, containerName, namespace string, portName string, resourceType, resourceName, address string) {
 	handler.handle(&proto.Event{
 		EventType: &proto.Event_PortEvent{
 			PortEvent: &proto.PortEvent{
+				RemotePort: &proto.IntOrString{
```
tejal29, line 390: Can we keep remote port as is if its of type int and use new field to populate the `intOrString` port value.

tejal29 (changes requested):
This change looks good to me except for the change in the skaffold.proto.
we have existing clients who use this event and changing `int` to `StringOrInt` may be break integration.

I will follow up with manually testing the new skaffold event on the CC IDEs side.

Meanwhile, can you instead use a new field and deprecate the existing field.

tejal29:
@anshlykov Can you please rebase ? Would love to get this in before next release since we already merged the PR to upgrade skaffold config.

anshlykov:
I have rebased and added changes to the proto schema.

tejal29 (commented):
LGTM. 

Pending Manual verification.

tejal29 (approved):
Manually verified this 
1. Integration test sample

PR #5014, opened 2020-11-10
Output logs in color for parallel builds

IsaacPD:
When building in parallel, logs would not show up with colors for builders that had colorful logs. 

Also Fixes: #4913 
**Related**: #4990 

**Description**
Wrap the output writer that is passed through skaffold with a "colorful writer" struct. This allows detecting when we should enable printing with colors at any point by checking in color.Fprint(f/ln) if it is a colorful writer.

Because the writer is now wrapped in a struct we are no longer passing os.Stdout raw and so when running commands it will be necessary to force enable/disable coloring with flags when appropriate.

**User facing changes**
Logs that normally show up in color in a non-parallel build are also in color for builds done in parallel.

**Before (master)**: 
Parallel build containing a jib maven artifact
Screen Shot 2020-11-10 at 1 10 47 PM (2)

Running ```skaffold run --mute-logs=all``` in skaffold/examples/ruby
```
...
Starting deploy...
- writing logs to /var/.../skaffold/deploy/2020-11-02_18-49-06.log
...
```
In /var/.../skaffold/deploy/2020-11-02_18-49-06.log
```
^[[34mStarting deploy...^[[0m
 - service/ruby configured
 - deployment.apps/ruby configured
```

The logs contain ANSI escape codes which do not represent the actual content intended to be logged.

**After**:
Parallel build containing a jib maven artifact
Screen Shot 2020-11-10 at 1 12 28 PM (2)

Running ```skaffold run --mute-logs=all``` in skaffold/examples/ruby
```
...
Starting deploy...
- writing logs to /var/.../skaffold/deploy/2020-11-02_18-53-51.log
...
```
In /var/.../skaffold/deploy/2020-11-02_18-53-51.log
```
Starting deploy...
 - service/ruby configured
 - deployment.apps/ruby configured
```

The only thing logged is the intended output.


pkg/skaffold/color/formatter.go
```diff
@@ -62,13 +62,22 @@ func SetupColors(out io.Writer, defaultColor int, forceColors bool) {
...
+	}
+	return out
 }
 
 // Color can be used to format text so it can be printed to the terminal in color.
 type Color struct {
 	color *colors.Color
 }
 
+type colorableWriter struct {
```
gsquared94, line 77: nit: you can do an embedded struct like:
```go
type colorableWriter struct {
	io.Writer
}
```
then you don't need to redefine `Write` method.

IsaacPD: Neat, golang always impresses me.

tejal29:
Verified this locally 
1. `--mute-logs=false`
2. building jib artifacts in parallel (master VS on branch)

pkg/skaffold/build/gcb/jib.go
```diff
@@ -41,15 +41,15 @@ func (b *Builder) jibBuildSpec(artifact *latest.Artifact, tag string) (cloudbuil
 			Steps: []*cloudbuild.BuildStep{{
 				Name:       b.MavenImage,
 				Entrypoint: "sh",
-				Args:       fixHome("mvn", jib.GenerateMavenBuildArgs("build", tag, artifact.JibArtifact, b.skipTests, true, artifact.Dependencies, b.artifactStore, b.cfg.GetInsecureRegistries())),
+				Args:       fixHome("mvn", jib.GenerateMavenBuildArgs(nil, "build", tag, artifact.JibArtifact, b.skipTests, true, artifact.Dependencies, b.artifactStore, b.cfg.GetInsecureRegistries())),
```
tejal29, line 44: This is possible source of nil pointer exception in future. 
would it make sense to pass `boolean` true or false instead of `out` to all the `GenerateXXXArgs` since they rely on result of `color.IsColorable` and never use out.

tejal29 (changes requested):

pkg/skaffold/build/gcb/jib.go
IsaacPD: Done!

pkg/skaffold/build/jib/gradle.go
```diff
@@ -97,8 +98,8 @@ func getSyncMapCommandGradle(ctx context.Context, workspace string, a *latest.Ji
 }
 
 // GenerateGradleBuildArgs generates the arguments to Gradle for building the project as an image.
-func GenerateGradleBuildArgs(task string, imageName string, a *latest.JibArtifact, skipTests, pushImages bool, deps []*latest.ArtifactDependency, r ArtifactResolver, insecureRegistries map[string]bool) []string {
-	args := gradleBuildArgsFunc(task, a, skipTests, true, MinimumJibGradleVersion)
+func GenerateGradleBuildArgs(task string, imageName string, a *latest.JibArtifact, skipTests, pushImages bool, deps []*latest.ArtifactDependency, r ArtifactResolver, insecureRegistries map[string]bool, showColors bool) []string {
+	args := gradleBuildArgsFunc(task, a, skipTests, showColors, MinimumJibGradleVersion)
```
gsquared94, line 102: super nit: `colorOutput` instead of `showColors` ?

pkg/skaffold/color/formatter.go
```diff
@@ -36,7 +36,7 @@ func init() {
 }
 
 // SetupColors enables/disables coloured output.
```
gsquared94, line 38:
```suggestion
// SetupColors conditionally wraps the input `Writer` with a color enabled `Writer`.
```

gsquared94 (approved):
lgtm!

pkg/skaffold/build/jib/gradle.go
IsaacPD: https://github.com/GoogleContainerTools/skaffold/blob/8caebb9955417b706a092bb7f896e0b544249797/pkg/skaffold/build/jib/gradle.go#L114
Many of the tests and this line use ```showColors```, so i'll leave it be to stay consistent. I think it is clearer with ```showColors``` as well since in ```colorOutput```  "color" can either be an adjective or a noun.

pkg/skaffold/color/formatter.go
IsaacPD: Done!

tejal29 (approved):

PR #5016, opened 2020-11-10
Replace util.SyncStore implementation to use singleflight and sync.Map.

tejal29:
This is follow up of #4896

In #4896, I used a library `sync/singleflight` to make sure `getDependencies` results were cached and only computed once for dockerfile builder.
`sync/singleflight` ensure same function for given input key executes **only once** across multiple goroutines at a given time.  Along with `sync.Map`, we can make sure at any given time, the function for given input key only executes once across multiple subroutines. 
This is ensured by, wrapping the result of the function we want to execute and storing it  in a thread safe `sync.Map` 
For cache miss, execute the func and store the result. 

During review @gsquared94 mentioned, he added `util.Store` which uses `sync.Once` and `sync.Map` to achieve th same thing. 

`sync/singleflight` is safer for a corner case when same key is looked up exactly at the same point in different subroutines.

Hence replacing the `sync.Store` implementation with `sync/singleflight`

**No User facing changes**

### DIFF
--- a/pkg/skaffold/build/cache/hash.go
+++ b/pkg/skaffold/build/cache/hash.go
@@ -37,10 +37,6 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 )
 
-const (
-	storePrefix = "dependency_hash"
-)
-
 // For testing
 var (
 	newArtifactHasherFunc = newArtifactHasher
@@ -65,7 +61,7 @@ func newArtifactHasher(artifacts build.ArtifactGraph, lister DependencyLister, m
 		artifacts: artifacts,
 		lister:    lister,
 		mode:      mode,
-		syncStore: util.NewSyncStore(storePrefix),
+		syncStore: util.NewSyncStore(),
 	}
 }
 
--- a/pkg/skaffold/build/cache/hash.go
+++ b/pkg/skaffold/build/cache/hash.go
@@ -37,6 +37,10 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 )
 
+const (
+	storePrefix = "dependency_hash"
+)
+
 // For testing
 var (
 	newArtifactHasherFunc = newArtifactHasher
@@ -61,7 +65,7 @@ func newArtifactHasher(artifacts build.ArtifactGraph, lister DependencyLister, m
 		artifacts: artifacts,
 		lister:    lister,
 		mode:      mode,
-		syncStore: util.NewSyncStore(),
+		syncStore: util.NewSyncStore(storePrefix),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/skaffold/build/cache/hash.go
```diff
@@ -37,6 +37,10 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 )
 
+const (
+	storePrefix = "dependency_hash"
```
gsquared94, line 41: why is the `storePrefix` needed?

pkg/skaffold/util/store.go
```diff
@@ -17,33 +17,67 @@ limitations under the License.
...
 )
 
-// SyncStore exports a single method `Exec` to ensure single execution of a function and share the result between all callers of the function.
+// SyncStore exports a single method `Exec` to ensure single execution of a function
+// and share the result between all callers of the function.
 type SyncStore struct {
-	oncePerKey *sync.Map
-	results    *sync.Map
+	prefix  string
+	sf      singleflight.Group
```
gsquared94, line 30: shouldn't `sf` be `*singleflight.Group`? Otherwise won't the underlying `Mutex` get copied on every reference of the instantiated struct?

pkg/skaffold/util/store.go
```diff
@@ -17,33 +17,67 @@ limitations under the License.
...
+		}
+		return v, nil
 	})
-
-	val, _ := o.results.Load(key)
+	if err != nil {
+		return err
+	}
 	return val
 }
```
gsquared94, line 59: nit: can avoid explicit closure with named return type
```go
func (o *SyncStore) Exec(key string, f func() interface{}) interface{} {
	fullKey := o.prefix + key
	val, err := o.sf.Do(fullKey, func() (_ interface{}, err error) {
		// trap any runtime error due to synchronization issues.
		defer func() {
			if rErr := recover(); rErr != nil {
				err = retrieveError(key, rErr)
			}
		}()
		v, ok := o.results.Load(key)
		if !ok {
			v = f()
			o.results.Store(key, v)
		}
		return v, nil
	})
	if err != nil {
		return err
	}
	return val
}
```

pkg/skaffold/build/cache/hash.go
tejal29: The reason, i added storePrefix is to gaurantee there is no collision of keys if the `util.SyncStore` was used by different modules.

gsquared94: I think between different instances of `*SyncStore` getting same keys wouldn't matter right since they'd be using different instances of both the `singleflight.Group` and `sync.Map`?

pkg/skaffold/util/store.go
tejal29: Not really, i usually refer the tests as a documentation for using code.

https://github.com/golang/sync/blob/master/singleflight/singleflight_test.go#L23

Here i also realized we don't need sync.Map as pointer. 
https://play.golang.org/p/iramynV7eJV

pkg/skaffold/build/cache/hash.go
tejal29: you are right! Thanks. i will drop it :)

gsquared94 (approved):

PR #5023, opened 2020-11-13
[KPT CODELAB] (1/3) New codelab dir + the sample application resources 

yuwenma:
This PR adds the sample application for the kpt codelab.  

Fixes: #4881 #3904


MarlonGamez:
I think this should be fine to go in, since it's just the sample app files for the codelab. Took a look over everything and it seems good :) I'll run kokoro and merge assuming that goes well 👍🏼

yuwenma:
> I think this should be fine to go in, since it's just the sample app files for the codelab. Took a look over everything and it seems good :) I'll run kokoro and merge assuming that goes well 👍🏼

Thank you!

MarlonGamez (approved):

PR #5024, opened 2020-11-13
 remove const var kustomizeFurtherGuidance

yuwenma:
Fixes: #4981 #3904

Description
Improve the kpt version check. We don't block users using an old kustomize version since we believe kpt can better deal with the kustomize versioning once using a builtin kustomize .

### DIFF
--- a/pkg/skaffold/deploy/kpt/kpt.go
+++ b/pkg/skaffold/deploy/kpt/kpt.go
@@ -88,7 +88,7 @@ func NewDeployer(cfg types.Config, labels map[string]string) *Deployer {
 
 var sanityCheck = versionCheck
 
-// versionCheck returns an error if the kpt and kustomize versions are not compatible with skaffold.
+// versionCheck guarantees the kpt and kustomize versions are compatible with skaffold.
 func versionCheck(dir string, stdout io.Writer) error {
 	kptCmd := exec.Command("kpt", "version")
 	out, err := util.RunCmdOut(kptCmd)
--- a/pkg/skaffold/deploy/kpt/kpt.go
+++ b/pkg/skaffold/deploy/kpt/kpt.go
@@ -57,7 +57,7 @@ const (
 
 	kustomizeDownloadLink  = "https://kubernetes-sigs.github.io/kustomize/installation/"
 	kustomizeMinVersion    = "v3.2.3"
-	kustomizeVersionRegexP = `{Version:(\S+) GitCommit:\S+ BuildDate:\S+ GoOs:\S+ GoArch:\S+}`
+	kustomizeVersionRegexP = `{Version:(kustomize/)?(\S+) GitCommit:\S+ BuildDate:\S+ GoOs:\S+ GoArch:\S+}`
 )
 
 // Deployer deploys workflows with kpt CLI
@@ -81,7 +81,7 @@ func NewDeployer(cfg types.Config, labels map[string]string) *Deployer {
 
 var sanityCheck = versionCheck
 
-// versionCheck guarantees the kpt and kustomize versions are compatible with skaffold.
+// versionCheck checks if the kpt and kustomize versions meet the minimum requirements.
 func versionCheck(dir string, stdout io.Writer) error {
 	kptCmd := exec.Command("kpt", "version")
 	out, err := util.RunCmdOut(kptCmd)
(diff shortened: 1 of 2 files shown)


yuwenma:
This PR is to resolve @briandealwis comment from #4994

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -125,11 +118,17 @@ func versionCheck(dir string, stdout io.Writer) error {
 		match := re.FindStringSubmatch(versionInfo)
 		if len(match) != 2 {
 			color.Yellow.Fprintf(stdout, "unknown kustomize version %q\n"+
-				"Your kustomize may be not from the official release\n%v\n", string(out),
-				kustomizeFurtherGuidance)
+				"Your kustomize may be not from the official release\n"+
+				"Please make sure your local kustomize version >= the official version %v, "+
+				"otherwise some features may not be well supported. "+
```
briandealwis, line 123: Can we shorten and simplify this to:
```suggestion
				"Skaffold/Kpt requires kustomize %v or later.\n"+
```

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -125,11 +118,17 @@ func versionCheck(dir string, stdout io.Writer) error {
...
+				"You can download the official version from %v\n", string(out),
+				kustomizeMinVersion, kustomizeDownloadLink)
 		} else if !semver.IsValid(match[1]) || semver.Compare(match[1], kustomizeMinVersion) < 0 {
-			color.Yellow.Fprintf(stdout, "you are using kustomize version %q\n%v\n",
-				match[1], kustomizeFurtherGuidance)
+			color.Yellow.Fprintf(stdout, "you are using kustomize version %q\n"+
+				"Please make sure your local kustomize version >= the official version %v, "+
+				"otherwise some features may not be well supported. "+
```
briandealwis, line 129:
```suggestion
				"Skaffold/Kpt requires kustomize %v or later.\n"+
```

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -88,7 +81,7 @@ func NewDeployer(cfg types.Config, labels map[string]string) *Deployer {
 
 var sanityCheck = versionCheck
 
-// versionCheck returns an error if the kpt and kustomize versions are not compatible with skaffold.
+// versionCheck guarantees the kpt and kustomize versions are compatible with skaffold.
```
briandealwis, line 84: I don't think this guarantees since it returns an error?

pkg/skaffold/deploy/kpt/kpt.go
```diff
@@ -125,11 +118,17 @@ func versionCheck(dir string, stdout io.Writer) error {
 		match := re.FindStringSubmatch(versionInfo)
 		if len(match) != 2 {
 			color.Yellow.Fprintf(stdout, "unknown kustomize version %q\n"+
```
briandealwis, line 120: WDYT?
```suggestion
			color.Yellow.Fprintf(stdout, "unable to determine kustomize version from %q\n"+
```

briandealwis (approved):
Some suggestions to reword, to shorten the messages.

pkg/skaffold/deploy/kpt/kpt.go
yuwenma: skaffold/kpt does not *require* the kustomize to be >= a certain version. It is just that if users encounter some issues, we remind them that the old kustomize version may be the cause. This warning is especially helpful here because there's no kpt/kustomize version skew support yet.

pkg/skaffold/deploy/kpt/kpt.go
yuwenma: sure.

pkg/skaffold/deploy/kpt/kpt.go
briandealwis: That sounds like there’s a minimum version! Would _recommended_ work?

(I don’t think we’re doing anybody a favour if they’re using an older version that we suspect may lead to problems.)

yuwenma: good idea. 😄 Updated.

briandealwis (approved):

PR #5027, opened 2020-11-13
Fix Bug that prevents showing survey prompt

IsaacPD:
Fix bug I introduced where the survey prompt will only show if os.Stdout does not support colors i.e. when ```color.SetupColors(out, forceColors)``` does not wrap ```out```. 

**Description**

Fixed by actually checking if the passed io.Writer is a wrapped colorableWriter. Added a test case to ensure that ```color.IsStdout(color.NewWriter(os.Stdout))``` returns true

**Before (master)**:
With an empty global config and running ```skaffold build``` in examples/getting-started, the survey prompt does not show up. This is because  ```survey.isStdout(out)``` returns false when out = colorableWriter{os.Stdout}.

**After**:
Survey prompt shows up given the same conditions above.

### DIFF
--- a/pkg/skaffold/color/formatter.go
+++ b/pkg/skaffold/color/formatter.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Skaffold Authors
+Copyright 2020 The Skaffold Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


MarlonGamez (approved):
This looks good to me. Nice catch on that bug 👍🏼

PR #5037, opened 2020-11-18
[refactor] move DoInit phases into functions

MarlonGamez:
**Related**: #4915

**Description**
No funcional changes, just organizing some `skaffold init` related code into functions, which will be useful for later when implementing necessary changes for #4915

Implement flow for transparently running `skaffold init` when the user doesn't have a config

### DIFF
--- a/pkg/skaffold/initializer/init.go
+++ b/pkg/skaffold/initializer/init.go
@@ -47,7 +47,7 @@ func DoInit(ctx context.Context, out io.Writer, c config.Config) error {
 	}
 
 	newConfig, newManifests, err := Initialize(out, c, a)
-	if err != nil {
+	if err != nil || c.Analyze {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/skaffold/initializer/init.go
```diff
@@ -40,64 +41,99 @@ func DoInit(ctx context.Context, out io.Writer, c config.Config) error {
...
-		return errors.New(`Projects set up to deploy with helm must be manually configured.
+		return nil, errors.New(`Projects set up to deploy with helm must be manually configured.
 
 See https://skaffold.dev/docs/pipeline-stages/deployers/helm/ for a detailed guide on setting your project up with skaffold.`)
 	}
 
+	return a, nil
+}
+
+// Initialize uses the information gathered by the analyzer to create a skaffold config and generate kubernetes manifests.
```
tejal29, line 75: I see 3 return type. 
Can you add more description on what the `map` represents and when it return an error ?

pkg/skaffold/initializer/init.go
```diff
@@ -40,64 +41,99 @@ func DoInit(ctx context.Context, out io.Writer, c config.Config) error {
 		}
 	}
 
+	a, err := AnalyzeProject(c)
+	if err != nil {
+		return err
+	}
+
+	newConfig, newManifests, err := Initialize(out, c, a)
+	if err != nil || c.Analyze {
```
tejal29, line 50: this is a little bit confusing that you return error when `c.Analyze` is set. 
Note: you are checking `c.Analyze` is set in two places now.  
It would be great if we only check it once and decide the course of action. WDYT

Would it make sense to move the WriteData in the `Initialize` ?
It would align with the fact that if `c.Analyze` is true, `Initialize` write results to `out`

tejal29 (commented):
couple of nits

pkg/skaffold/initializer/init.go
MarlonGamez: I agree, it is clunky :/ I'd like to get rid of the second check if I can. I'm a bit hesitant to move the `WriteData` logic into `Initialize`, as we want to be able to call them separately in the future for the transparent init functionality

tejal29: sounds good.

pkg/skaffold/initializer/init.go
MarlonGamez: I added a comment to explain the the map that returns. Would you like me to explain each error scenario?

tejal29 (approved):

Issue #5044, opened 2020-11-20
Allow referencing previous skaffold.yaml versions in the docs


IsaacPD:
Currently on each release http://skaffold.dev/docs/references/yaml/ is updated to include the newest api version for skaffold.yaml removing any way to see configurations for previous versions. It would be nice to see the previous versions from outside of the development environment.

PR #5048, opened 2020-11-20
Multi version docs

IsaacPD:
Fixes: #5044 

**Description**

Access different skaffold.yaml versions by passing in the version into the url parameter "version". Support previous yaml configs by updating the description to the reference description when possible. Add list of skaffold versions to the side of https://skaffold.dev/docs/references/yaml/ page. It still defaults to the latest skaffold api version.

**User facing changes**

Previous skaffold configs are accessible in the skaffold.yaml reference page as links on the side panel or via the url parameter "version"

Screen Shot 2020-11-20 at 2 33 14 PM


tejal29:
This looks great!

My only request would be to add a tag/annotation (latest) next to v2beta10 on the right hand side,
image

tejal29 (approved):
Approved with 1 small request.

Issue #5049, opened 2020-11-20
Regression causing --mount option to break skaffold build


shaxbee:
### Expected behavior

Dockerfile with `# syntax = docker/dockerfile:1-experimental` and `RUN --mount=<value> <cmd>` no longer works on master.

### Actual behavior

Dockerfile with `dockerfile:1-experimental` syntax should work.

### Information

- Skaffold version: master
- Operating system: macOS

### Steps to reproduce the behavior

Added test case to [pkg/skaffold/docker/dependencies_test.go](https://github.com/GoogleContainerTools/skaffold/compare/master...shaxbee:bug-buildkit-dependency-resolve#diff-60da1d62ea286ae7cdadb04f16157c056600acd3293e174664ee5f6b2dee72ff).

This is ultimately failing in [pkg/skaffold/docker/parse.go#L81](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/docker/parse.go#L81).

shaxbee:
After commenting out the code in pkg/skaffold/docker/parse.go it passes the dependency check but fails at build.
Running latest stable release that does not happen though, wonder if it could be related to [pkg/skaffold/build/docker/docker.go#L49](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/docker/docker.go#L49), there used to be check for buildkit there.

MarlonGamez:
@shaxbee thanks for opening this! I'll take a look at your PR

PR #5052, opened 2020-11-21
fix: buildkit build

shaxbee:
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/5049, #5093 

**Description**
Buildkit builds are broken on master due to change of logic around CLI build.
This fix restores original behaviour by using docker CLI for buildkit builds, it also removes instructions.Parse which is not compatible with buildkit Dockerfiles with version of `moby/buildkit` currently being used.
Added testcase with buildkit Dockerfile to `pkg/skaffold/docker/dependencies_test.go`.

**User facing changes**
Buildkit builds fail.

**Follow-up Work**
Revert this fix after `moby/buildkit` has been upgraded and tests pass.


MarlonGamez:
Hey @shaxbee, thanks for catching this and opening creating a PR for it :) Do you know if the latest version of the moby/buildkit library supports the syntax you mentioned?

pkg/skaffold/docker/parse.go
```diff
@@ -77,10 +76,12 @@ func readCopyCmdsFromDockerfile(onlyLastImage bool, absDockerfilePath, workspace
 		return nil, fmt.Errorf("parsing dockerfile %q: %w", absDockerfilePath, err)
 	}
 
+	// TODO: re-enable this after updating moby/buildkit dependency
+	// https://github.com/GoogleContainerTools/skaffold/issues/5049
 	// instructions.Parse will check for malformed Dockerfile
-	if _, _, err := instructions.Parse(res.AST); err != nil {
-		return nil, fmt.Errorf("parsing dockerfile %q: %w", absDockerfilePath, err)
-	}
+	// if _, _, err := instructions.Parse(res.AST); err != nil {
```
tejal29, line 82: @shaxbee Unfortunately, skaffold needs to parse the dockerfile to get 
1. dependencies to watch
2. to create a docker context etc.

It would make sense to update the docker dependency.

tejal29 (changes requested):

shaxbee:
It seems that [moby/buildkit v0.8.0-rc1](https://github.com/moby/buildkit/releases/tag/v0.8.0-rc1) supports the --mount syntax, there is no official v0.8.0 release yet though, therefore current parsing logic in master will break the build for many users if it is released.

pkg/skaffold/docker/parse.go
MarlonGamez: @tejal29 I think there are actually two call to different `Parse` functions. The one that is modified here is the one that just checks for malformed Dockerfile commands. The one that actually gets the required information from the dockerfile is still intact

shaxbee: Exactly.

shaxbee:
Seems like release v1.17.0 went out with the broken buildkit build, I'm unable to run the sample stack anymore on latest skaffold: https://github.com/shaxbee/todo-app-skaffold

shaxbee:
@IsaacPD @tejal29 I can't really find where missing '--force-rm' flag is coming from, should I update tests instead?

IsaacPD:
> @IsaacPD @tejal29 I can't really find where error for missing `--force-rm` flag is coming from, should I update tests instead?

Yes please do, I was taking a look at those tests as well and it seems that there is some errors in the logic of them. I think I have something that fixes them on http://github.com/IsaacPD/skaffold/tree/re-enable-buildkit-with-flag if you need a place to start.

pkg/skaffold/docker/parse.go
shaxbee: Actually instructions are used to resolve ONBUILD so this cannot be disabled.

shaxbee:
Since moby/buildkit v0.8.0 with docker frontend support for --mount flag was released I bumped the dependency and `instructions.Parse` is working now for those Dockerfiles.

MarlonGamez:
Thanks for sticking with this and keeping us updated @shaxbee! I'll run kokoro for this and if all goes well it should be good to merge

pkg/skaffold/docker/parse.go
MarlonGamez: resolving this as it's no longer relevant to the PR

PR #5058, opened 2020-11-23
Add skaffold meter struct

IsaacPD:
Create and initialize struct to persist relevant metrics about the execution of the program.

**Description**

Initialize global variable with values that can be collected at the start of execution. Use InitMeter function to set values that need to be retrieved from the skaffold config and run context. Parse values from the Skaffold Config by using the raw yaml string as the label. Descriptions of the values in the struct:

```go
// ExitCode stores what exit code skaffold throws
ExitCode       int
// BuildArtifacts contains how many build artifacts in configuration
BuildArtifacts int
// Command is the command used for the iteration
Command        string
// Version of skaffold being used i.e. v1.17
Version        string
// OS that skaffold is running on (darwin, win, linux, etc.)
OS             string
// Arch that skaffold is running on (amd64, arm64, etc.)
Arch           string
// PlatformType being deployed to (local, cluster, googleCloudBuild)
PlatformType   string
// Deployers used for deployment
Deployers      []string
// EnumFlags flags with predefined values used in invocation
EnumFlags      map[string]interface{}
// Builders used to build artifacts
Builders       map[string]bool
// SyncType tracks whether infer, auto, and/or manual are being used
SyncType       map[string]bool
// DevIterations counts the triggers for each dev iteration (sync, build, or deploy)
DevIterations  map[string]int
// StartTime the start time of the skaffold program, used to track how long skaffold took to finish executing
StartTime      time.Time
// ErrorCode returned when skaffold throws an error
ErrorCode      proto.StatusCode
```

**Follow-up Work**

Populate the values for EnumFlags, Deployers, ErrorCode, and DevIterations, write metrics, and upload them, create relevant documentation, and prompt users to opt-out.

### DIFF
--- a/pkg/skaffold/instrumentation/meter.go
+++ b/pkg/skaffold/instrumentation/meter.go
@@ -17,8 +17,13 @@ limitations under the License.
 package instrumentation
 
 import (
+	"runtime"
 	"time"
 
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner/runcontext"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/version"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/yamltags"
 	"github.com/GoogleContainerTools/skaffold/proto"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/skaffold/instrumentation/meter.go
```diff
@@ -0,0 +1,87 @@
...
+		ExitCode:  0,
+		ErrorCode: proto.StatusCode_OK,
+	}
+)
+
+func InitMeter(runCtx *runcontext.RunContext, config *latest.SkaffoldConfig) {
+	meter.Command = runCtx.Opts.Command
+	meter.PlatformType = getYamlTag(config.Build.BuildType)
+	for _, artifact := range config.Pipeline.Build.Artifacts {
+		meter.Builders[getYamlTag(artifact.ArtifactType)] = true
```
tejal29, line 67: nice!

pkg/skaffold/instrumentation/meter.go
```diff
@@ -0,0 +1,87 @@
...
+		if artifact.Sync != nil {
+			meter.SyncType[getYamlTag(artifact.Sync)] = true
+		}
+	}
+	meter.BuildArtifacts = len(config.Pipeline.Build.Artifacts)
+}
+
+func getYamlTag(value interface{}) string {
```
tejal29, line 75: shd we move this to pkg/skaffold/yamltags?

There are some helper functions already defined there.

pkg/skaffold/instrumentation/meter.go
```diff
@@ -0,0 +1,87 @@
...
+	EnumFlags      map[string]interface{}
+	Builders       map[string]bool
+	SyncType       map[string]bool
+	DevIterations  map[string]int
+	StartTime      time.Time
+	ErrorCode      proto.StatusCode
+}
+
+var (
+	meter = skaffoldMeter{
```
tejal29, line 51: Is there a reason why this is a `var` 
I am thinking, if we need to make this struct thread safe. e.g. 
- would you accept this struct to be updated in different go subroutines?
e.g. - in case a build fails, does the struct count get updated then and there ?
Look at pkg/skaffold/event/State struct and how its protected with a mutex.

IsaacPD: At the moment I believe everything that would be captured in this struct can be done outside of go subroutines. But I do not think it should exclude those, given that I think the const label is more appropriate. I did not really have any intent with picking var 😅

tejal29:
please rebase from master to pick up the flaky test fix.

IsaacPD:
Rebased onto master.

tejal29 (approved):

Issue #4692, opened 2020-08-18
Too many UNKNOWN/ERROR type failures for Skaffold events


go-nelango:
Large number of skaffold event errors are being reported as UKNOWN/ERROR. We need more descriptive error codes for Cloud Code metrics tracking and investigations.

Based on this dashboard https://dashboards.corp.google.com/edit/_dcb794d1_3a93_4405_8560_8230e2e5325d, we see:
![image](https://user-images.githubusercontent.com/68573204/90555858-da08df00-e14c-11ea-9d83-00b27dd0b851.png)

On vscode side, we log skaffold version for cloudcode.skaffold.session.end event. So filtering based on versions:

68% error codes from v1.13.1 are UKNOWN/ERROR
![image](https://user-images.githubusercontent.com/68573204/90555636-88605480-e14c-11ea-985a-4d6349583c4b.png)

43% error codes from v1.13.0 are UKNOWN/ERROR
![image](https://user-images.githubusercontent.com/68573204/90555726-ac239a80-e14c-11ea-8189-3c090b19881b.png)

go-nelango:
@briandealwis @tejal29 @sivakku as FYI

tejal29:
Thanks @go-nelango for opening this issue.

tejal29:
We have an issue #4645 which @PriyaModali  is working on to reduce Deploy_UNKNOWN. 
@go-nelango, can you add some data on
1. BUILD_UNKNOWN by builders?
2. DEPLOY_UNKNOWN by deployers?

tejal29:
Handled BUILD_UNKONWN -> docker connectivity issues here #4914

go-nelango:
Please note charts below are based on IntelliJ metrics from 9/1/2020 to 9/30/2020 with skaffold version v1.13.2 and higher.
Vscode is currently not reporting Skaffold Metaevent and we are fixing it.

1. BUILD_UNKNOWN by builders
![image](https://user-images.githubusercontent.com/68573204/96059730-97f5d400-0e43-11eb-880d-664dc1efdc36.png)

2.DEPLOY_UNKNOWN by deployers
![image](https://user-images.githubusercontent.com/68573204/96059741-9debb500-0e43-11eb-997c-a08ff239818a.png)

tejal29:
We have now reduced unknowns to 23% of all failure that happen or 3% of total sessions

tejal29:
[...]

PR #5059, opened 2020-11-24
add docker build distinct error codes

tejal29:
should merge after #5060 
Related to #4692

In this PR, 
1. First add a pkg/skaffold/error.ErrDef helper error for builders, deployers etc to create a error with errorcode and suggestion code. 

Then, 

Add distinct error codes for docker build. 
Docker exposes two types of error
1. Internal system error 
All these error codes are from https://github.com/moby/moby/blob/master/errdefs/defs.go
These errors can happen for cases like, 
- docker context tar is empty, - probably issue with `getDependencies`
- docker build is accessing a volume mount which can't be created or is deleted.
- a conflict occurs when scheduling a build container etc.
- docker build/dockerfile  is using features which are not supported.

2. Docker build error due to user error or incorrect dockerfile.
In this case of errors executing the steps in dockerfile, a [`jsonmessage.JsonError`](https://github.com/moby/moby/blob/cf0ce96eb129ebcc7d07f0f47a8683c16f228c7d/builder/dockerfile/dispatchers.go#L387) is thrown.


pkg/skaffold/build/docker/docker.go
```diff
@@ -115,7 +112,7 @@ func (b *Builder) pullCacheFromImages(ctx context.Context, out io.Writer, a *lat
 		}
 
 		if err := b.localDocker.Pull(ctx, out, image); err != nil {
-			warnings.Printf("Cache-From image couldn't be pulled: %s\n", image)
+			warnings.Printf("cacheFrom image couldn't be pulled: %s\n", image)
```
briandealwis, line 115: This `err` seems like it might be useful to be reported too, doesn't it?

pkg/skaffold/build/docker/errors.go
```diff
@@ -0,0 +1,123 @@
...
+	"fmt"
+
+	"github.com/docker/docker/errdefs"
+	"github.com/docker/docker/pkg/jsonmessage"
+
+	sErrors "github.com/GoogleContainerTools/skaffold/pkg/skaffold/errors"
+	"github.com/GoogleContainerTools/skaffold/proto"
+)
+
+func newBuildError(err error) error {
```
briandealwis, line 30: This function needs some documentation!
```suggestion
// newBuildError turns Docker-specific errors into actionable errors.  The input errors are assumed to be
// from the Skaffold docker invocation.
func newBuildError(err error) error {
```

pkg/skaffold/build/docker/errors.go
```diff
@@ -0,0 +1,123 @@
...
+	errU := errors.Unwrap(err)
+	if errU == nil {
+		return err
```
briandealwis, line 33: What kind of errors are these?  Why do we return them instead of wrapping them with a BUILD_UNKNOWN error?  (is it because they'll be parsed and turned into a error elsewhere?)

proto/skaffold.proto
```diff
@@ -567,10 +602,19 @@ enum SuggestionCode {
...
     // Check if docker is running
     CHECK_DOCKER_RUNNING = 106;
+    // Fix User Build Error
+    FIX_USER_BUILD_ERR = 107;
+    // Docker build internal error, try again
+    DOCKER_BUILD_RETRY = 108;
+    // Fix `cacheFrom` config for given artifact and try again
+    FIX_CACHE_FROM_ARTIFACT_CONFIG = 109;
+    // Fix `dockerfile` config for a given artifact and try again.
+    FIX_DOCKERFILE_CONFIG = 110;
```
briandealwis, line 616: This name did make me think it was about the Dockerfile content rather than the `skaffold.yaml`'s docker configuration.  Would it make sense to have a class of fixes like `FIX_SKAFFOLD_CONFIG_DOCKERFILE`?

briandealwis (approved):

pkg/skaffold/build/docker/errors.go
tejal29: these are known docker errors. Docker wraps these for known types https://github.com/moby/moby/blob/master/errdefs/defs.go
e.g. here 
https://github.com/moby/moby/blob/master/errdefs/helpers.go#L138

proto/skaffold.proto
tejal29: done

pkg/skaffold/build/docker/errors.go
tejal29: added it now.

Issue #4796, opened 2020-09-16
Debug needs to work with CNB Platform v0.4 and the new launcher style


briandealwis:
Debugging a buildpacks image with pack v0.13 now uses lifecycle 0.9.1, which introduced a [new launcher-style (RFC 45)](https://github.com/buildpacks/rfcs/blob/main/text/0045-launcher-arguments.md).

Previously images were built with the `ENTRYPOINT=/cnb/lifecycle/launcher` and used `CNB_PROCESS_TYPE` to determine the process to launch.  But it meant running these images with arguments like `docker run <bpimage> arg ...` didn't have the expected effect as the args were interpreted by the _launcher_ rather than the actual command that would be run.  To pass arguments to the process required providing `--` to indicate end of args to the launcher:  `docker run <bpimage> -- arg ...`.

CNB Platform 0.4 (implemented by pack v0.13.1) now links the `/cnb/lifecycle/launcher` to `/cnb/process/<process-type>` for each process-type.  The launcher now uses its invoked name to determine the process type, and `CNB_PROCESS_TYPE` is now ignored.  The process-type name `launcher` is also forbidden.

Marking as a bug as this issue prevents updating to pack v0.13.1 (#4757).

tejal29:
It part of upcoming milestone. keeping the priority as is.

Issue #4733, opened 2020-08-28
`TestDevPortForwardDeletePod` failing integration test against Minikube on Darwin


gsquared94:
<details><summary>Error log</summary>

```
=== Failed Tests ===
/integration/TestDevPortForwardDeletePod
time="2020-08-28T10:21:47Z" level=info msg="Namespace: skaffoldqcht4"
time="2020-08-28T10:21:47Z" level=info msg="[skaffold dev --namespace skaffoldqcht4 --default-repo gcr.io/k8s-skaffold --port-forward --rpc-port 63274]"
    rpc_test.go:377: waiting for connection...
    rpc_test.go:377: waiting for connection...
    dev_test.go:267: event received timestamp:<seconds:1598610108 nanos:533826000 > event:<metaEvent:<entry:"Starting Skaffold: &{Version:67e81bc ConfigVersion:skaffold/v2beta7 GitVersion: GitCommit:67e81bce7eba453dcca2b57a92e8157dbca3ecc3 GitTreeState:clean BuildDate:2020-08-28T06:00:45Z GoVersion:go1.14.2 Compiler:gc Platform:darwin/amd64}" metadata:<build:<numberOfArtifacts:2 builders:<type:DOCKER count:2 > type:LOCAL > deploy:<deployers:<type:KUBECTL count:1 > cluster:MINIKUBE > > > > 
    dev_test.go:267: event received timestamp:<seconds:1598610108 nanos:533920000 > event:<devLoopEvent:<status:"In Progress" > > entry:"Update initiated due to file change" 
    dev_test.go:267: event received timestamp:<seconds:1598610110 nanos:375950000 > event:<buildEvent:<artifact:"leeroy-web" status:"In Progress" > > entry:"Build started for artifact leeroy-web" 
WARNING: Could not open the configuration file: [/Users/runner/.config/gcloud/configurations/config_default].
ERROR: (gcloud.auth.docker-helper) You do not currently have an active account selected.
Please run:

  $ gcloud auth login

to obtain new credentials, or if you have already logged in with a
different account:

  $ gcloud config set account ACCOUNT

to select an already authenticated account to use.
    dev_test.go:253: timed out waiting for port forwarding event
time="2020-08-28T10:22:49Z" level=info msg="Ran in 1m2.012523065s"
--- FAIL: TestDevPortForwardDeletePod (62.04s)
```
</details>

[...]

PR #5064, opened 2020-11-25
Update to pack 0.15 and add debug support for CNB Platform API 0.4

briandealwis:
Fixes: #4796 

**Description**
Update to pack 0.15.0, and update `skaffold debug` to support the Cloud Native Buildpacks _Platform API 0.4_, the default for the CNB `pack` ≥ 0.13.0 and lifecycle ≥ 0.9.  Platform API 0.4 changes how _processes_ are requested in an image:
  - prior to 0.4, the entrypoint points to a launcher binary in `/cnb/lifecycle/launcher`, and the process is set via a `CNB_PROCESS_TYPE` environment variable (defaulting to `web` if not present)
  - with 0.4, the processes have corresponding executables in `/cnb/process/xxx` which are symlinks to `/cnb/lifecycle/launcher`.  When executed, they cause the corresponding process to be run.  The default process is the entrypoint (e.g., `/cnb/process/web`).

The Platform API version is set via the new `CNB_PLATFORM_API` environment variable.

(This doesn't actually have any user-facing changes as to this point we still link with pack 0.12 and so create images with CNB Platform API 0.3.)


pkg/skaffold/debug/cnb.go
```diff
@@ -97,58 +148,51 @@ func updateForCNBImage(container *v1.Container, ic imageConfiguration, transform
...
+			result = append(result, clArgs...)
+			return result
+		}
 	}
 
 	if len(ic.arguments) == 0 {
-		// indicates an image mis-configuration as we should have resolved the the
-		// CNB_PROCESS_TYPE (if specified) or `web`.
-		logrus.Warnf("no CNB launch found for %s/%s", ic.artifact, processType)
+		logrus.Warnf("no CNB launch found for %s", ic.artifact)
```
tejal29, line 195: would it possible to debug a container if CNB launch is not found?

tejal29 (approved):
LGTM.

pkg/skaffold/debug/cnb.go
```diff
@@ -19,45 +19,89 @@ package debug
...
+//   3. _shell scripts_: the user can provide a shell script as the first argument and any
+//      remaining arguments are available as positional arguments like _script processes_.
 //
-// A key point is that the script-style launches allow support references to environment variables.
-// So we find the command line to be executed, whether that is a script or a direct, and turn it into
-// a normal command-line.  But we also return a rewriter to transform the the command-line back to
-// the original form.
+// Script-style launches support referencing environment variables since they are expanded by the shell.
+//
+// Configuring the launch depends on the CNB Platform API version being used, which is determined by
```
tejal29, line 83: Thanks for the details, it was a little bit easier to understand the code.

briandealwis: Going through this was painful :-)

pkg/skaffold/debug/cnb.go
briandealwis: No.  At this point we have an CNB launcher entrypoint (either `/cnb/lifecycle/launcher` or `/cnb/process/xxx`), but no corresponding process found.  So we don't have a command-line should be executed.

Basically we'll return, the debug transforms won't be able to figure anything out, and the container / pod will remain un-touched.  And the execution of the container will then fail as no process definition will be found.

briandealwis:
Marking TestDevPortForwardDeletePod to be skipped #4733.

nkubala (approved):
this LGTM, its essentially an improved version of similar changes I made in https://github.com/GoogleContainerTools/skaffold/pull/5038. thanks for the detailed comments @briandealwis

Issue #5005, opened 2020-11-06
humanize time in "stabilized in 5.23494327s"


drewpca:
https://github.com/GoogleContainerTools/skaffold/blob/331da68e249ff6db83038f64513da33fd944ecd5/pkg/skaffold/runner/deploy.go#L160

My preference would be for it to say "5.2 seconds" since 0.1 sec is enough precision for this step. I don't care if it switches to stuff like "821 milliseconds" for short times.

tejal29:
From the chat root, 

```
I've used https://github.com/dustin/go-humanize for this kind of thing in the past, not sure if Skaffold has some standardized library they work with for making time easier for humans to consume
```

gunadhya:
Hi, I'd like to work on this issue.

drewpca:
Thanks, gunadhya!

PR #5079, opened 2020-11-29
Fix for humanize time in #5005

gunadhya:
Fixes: #5005  

**Description**

Added [go-humanize](https://github.com/dustin/go-humanize) 
I couldn't figure out how to get milliseconds values in the go-humanize library so created a workaround for that.

Format for printing output [here](https://github.com/GoogleContainerTools/skaffold/blob/7608596722f98390b8e722f9880fbed94c7f56dc/pkg/skaffold/runner/deploy.go#L160)


tejal29:
Thanks a lot @gunadhya for your PR.

can you please show the change output and how the new line looks?

Also, can you please switch to human readable time elsewhere https://github.com/GoogleContainerTools/skaffold/search?q=time.Since%28start%29


Thanks
Tejal

tejal29 (changes requested):
Thanks a lot @gunadhya for your PR.

can you please show the change output and how the new line looks?

Also, can you please switch to human readable time elsewhere https://github.com/GoogleContainerTools/skaffold/search?q=time.Since%28start%29

Thanks
Tejal

gunadhya:
Hi @tejal29, So the output would look something like this : `Deployments stabilized 3 seconds ago`.  Here's the [doc example](https://pkg.go.dev/github.com/dustin/go-humanize#Time). Should I make the changes to other [files](https://github.com/GoogleContainerTools/skaffold/search?q=time.Since%28start%29) in this format too?

tejal29:
> Hi @tejal29, So the output would look something like this : `Deployments stabilized 3 seconds ago`. Here's the [doc example](https://pkg.go.dev/github.com/dustin/go-humanize#Time). 

Thanks!


> Should I make the changes to other [files](https://github.com/GoogleContainerTools/skaffold/search?q=time.Since%28start%29) in this format too?

yes please.

gunadhya:
I've made the changes as requested. Please do let me know if anything else seems broken or need fixing.

integration/skaffold/helper.go
```diff
@@ -256,8 +276,14 @@ func (b *RunBuilder) RunOrFailOutput(t *testing.T) []byte {
...
 	}
+	//Create human readable time string
+	showsTime := humanize.Time(start)
 
-	logrus.Infoln("Ran in", time.Since(start))
+	//Case for when it takes less than a second
+	if time.Since(start).Seconds() < 1 {
+		showsTime = time.Since(start).String() + " ago"
+	}
+	logrus.Infoln("Ran", showsTime)
```
tejal29, line 286: this would make an excellent utility function so you don't have to repeat the code. 
```suggestion
	logrus.Infoln("Ran", util.showHumanizeTime(start))
```
and

```
def showHumanizeTime(start time.Time) string {

showsTime := humanize.Time(start)
if time.Since(start).Seconds() < 1 {
	return  time.Since(start).String() + " ago"
}
return humanize.Time(start)
}
```

tejal29 (changes requested):

integration/skaffold/helper.go
gunadhya: I've added a function in util and reworked its usage in the files.

pkg/skaffold/util/util.go
```diff
@@ -341,3 +343,11 @@ func IsSubPath(basepath string, targetpath string) bool {
 func hasHiddenPrefix(s string) bool {
 	return strings.HasPrefix(s, hiddenPrefix)
 }
+
+// ShowHumanizeTime returns time in human readable format
+func ShowHumanizeTime(start time.Time) string {
+	if time.Since(start).Seconds() < 1 {
+		return time.Since(start).String() + " ago"
```
IsaacPD, line 350: I don't think it makes sense to use "ago" to indicate how long something took. If i said a deployment finished 3s ago, that means it finished in the past, but really we are printing basically when we know its done i.e. the present.

Further I think in order to accomplish just showing seconds without an excessive amount of decimals we can just use sprintf, as in this example https://play.golang.org/p/lAF0cEGMb27 and not rely on gohumanize in the first place.

IsaacPD (changes requested):

pkg/skaffold/util/util.go
tejal29: i think we can drop the `ago`. however i was under the impression humanize will round off time like 
70 seconds to something readble like 1 min 10 s

Thanks for bringing this to my attention.

IsaacPD: Yeah that is something nice that humanize does, it may be best instead to use [RelTime](https://pkg.go.dev/github.com/dustin/go-humanize#RelTime) instead as it will not append "ago" to the time.

gunadhya: @tejal29 @IsaacPD So [RelTime](https://pkg.go.dev/github.com/dustin/go-humanize#RelTime) requires labels to be defined for function call. I was thinking maybe remove the `" ago"` string from the `humanize.Time()` return value using [TrimSuffix](https://golang.org/pkg/strings/#TrimSuffix) from the standard library. What do you think?

IsaacPD: You can use it like RelTime(t, time.Now(), "", "") as in this example https://play.golang.org/p/_3_lAP3RauV and that will humanly print the difference from when it started to now without any extra text like "ago" at the end.

Although this will remove any precision less than the current magnitude (i.e 2.9 seconds and 2.0 seconds will both print "2 seconds" and similarly for "1hr59min and 1hr0min" as seen here https://play.golang.org/p/zEb8noqHPmF)

gunadhya: What do you think about using Round function from the time package ? ([link to doc](https://golang.org/pkg/time/#Duration.Round))

IsaacPD: I think it could work so long as we round down to the largest unit that is smallest than the duration, i.e.
```go
t := time.Since(start)
switch{
// i doubt rounding to days will be necessary
...
case t >= t.Hour:
      t := t.Round(t.Hour)
case t >= t.Minute:
     t := t.Round(t.Minute)
...
// etc to the smallest unit we want to show
}
```

gunadhya: I'm trying to understand how the round function works here https://play.golang.org/p/xcxGWUvDppv and I've noticed that for values greater than 1 second (in  `h`,`m`,`s`)  we can round using `t.Round(time.Millisecond)` for a three point precision. Can you tell me if this is the right way to go ?

IsaacPD: If you do not use ```humanize.Reltime``` it would be fine, although it's similar to just using ```[Duration.Truncate](https://pkg.go.dev/time?utm_source=gopls#Duration.Truncate)``` and so we'd lose the readable spacing and expansion of units that humanize provides

If you do use ```humanize.Reltime``` this would not be sufficient since it would still return 1 hour for "1hr59min" as input, and the approach i described in the comment above (which i just edited) would be more appropriate. This is clearer here https://play.golang.org/p/nppwgVaZCso

gunadhya:
I'm not sure why this test is failing? https://travis-ci.com/github/GoogleContainerTools/skaffold/jobs/464991267#L1013

pkg/skaffold/util/util_test.go
```diff
@@ -468,3 +469,31 @@ func TestIsURL(t *testing.T) {
...
+		{
+			description: "Case for 5.23494327s",
+			value:       currTime.Add(-duration2),
+			expected:    "5.2",
+		},
+	}
+	for _, test := range tests {
+		testutil.Run(t, test.description, func(t *testutil.T) {
+			humanizedValue := ShowHumanizeTime(test.value)
+			t.CheckMatches(test.expected, humanizedValue)
```
IsaacPD, line 496: In this case even if ```ShowHumanizeTime``` returned "5.23412341" (which is not "humanized") the test would still pass since ```t.CheckMatches``` does ```regexp.MatchString("5.2","5.23412341")```.

You should either make your pattern more strict or use ```t.CheckDeepEquals``` instead

IsaacPD (commented):
Everything looks good, just some changes to the test should be made.

pkg/skaffold/util/util.go
```diff
@@ -341,3 +342,28 @@ func IsSubPath(basepath string, targetpath string) bool {
 func hasHiddenPrefix(s string) bool {
 	return strings.HasPrefix(s, hiddenPrefix)
 }
+
+// ShowHumanizeTime returns time in human readable format
+func ShowHumanizeTime(start time.Time) string {
```
IsaacPD, line 347: Can we make this take ```time.Duration``` instead of ```time.Time``` so that the test don't have any variance?

gunadhya: I'll try that.

IsaacPD (approved):

PR #5084, opened 2020-11-30
Re-add colors to buildpacks and docker cli output

IsaacPD:
**Related**:  #5014 

**Description**
Because of #5014, buildpacks and docker buildkit were not receiving a terminal to write to and so they didnt color their output. Since buildpacks and docker don't have a flag to force output coloring from what i've seen, pass in the original writer that is wrapped by color.colorableWriter to buildpacks and docker.

**User facing changes**
Colors are back in buildpacks 

**Follow-up Work**

Buildpacks and docker buildkit will still not have any colored output when builds are done in parallel since the underlying writer in that instance would be a pipe. This can be remedied if an option is added to either to force colors, or if instead of pipes being directly passed a pseudo terminal could be used for building in parallel on skaffold's side.

### DIFF
--- a/pkg/skaffold/build/buildpacks/lifecycle.go
+++ b/pkg/skaffold/build/buildpacks/lifecycle.go
@@ -32,6 +32,7 @@ import (
 	"github.com/buildpacks/pack/project"
 	"github.com/sirupsen/logrus"
 
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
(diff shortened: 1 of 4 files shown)


gsquared94:
interesting idea about a pseudo terminal. Did you find any package that provides this?

IsaacPD:
> interesting idea about a pseudo terminal. Did you find any package that provides this?

Brian had previously found this library https://github.com/creack/pty to accomplish that.

gsquared94 (approved):

Issue #5093, opened 2020-12-02
useBuildKit: true for docker builds not working since 1.17.0


robertgates55:
With a simple `skaffold.yaml`:
```
apiVersion: skaffold/v2beta5
kind: Config

build:
  local:
    useBuildkit: true
  artifacts:
    - image: helloworld
      context: ./build/images/helper
      docker:
        dockerfile: Dockerfile
```

### Expected behavior
With v1.16.0, this invokes docker build with buildkit enabled:
```
➜  skaffold1160 build
Generating tags...
 - helloworld -> WARN[0000] Using "releases_95.2.2-1006-gdc512f74042" instead of "releases/95.2.2-1006-gdc512f74042" as an image tag 
helloworld:releases_95.2.2-1006-gdc512f74042
Checking cache...
 - helloworld: Not found. Building
Found [docker-desktop] context, using local docker daemon.
Building [helloworld]...
[+] Building 0.8s (10/10) FINISHED                                                                                                                                                                                                                                       
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                0.1s
 => => transferring dockerfile: 438B                                                                                                                                                                                                                                0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                                   0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                     0.0s
 => [internal] load metadata for docker.io/library/alpine:3.12.0                                                                                                                                                                                                    0.0s
 => [1/5] FROM docker.io/library/alpine:3.12.0                                                                                                                                                                                                                      0.0s
 => [internal] load build context                                                                                                                                                                                                                                   0.0s
 => => transferring context: 721B                                                                                                                                                                                                                                   0.0s
 => CACHED [2/5] RUN apk update &&   apk add --update libintl &&   apk add --virtual     build_deps     gettext     mysql-client     bash                                                                                                                           0.0s
 => [3/5] COPY --chown=1000:1000 scripts /scripts                                                                                                                                                                                                                   0.1s
 => [4/5] RUN chmod +x /scripts/*.sh                                                                                                                                                                                                                                0.4s
 => [5/5] WORKDIR /scripts                                                                                                                                                                                                                                          0.1s
 => exporting to image                                                                                                                                                                                                                                              0.1s
 => => exporting layers                                                                                                                                                                                                                                             0.1s
[...]

PR #5096, opened 2020-12-02
Use buildkit in docker cli if useBuildkit=true

IsaacPD:
Fixes: #5093 

**Description**
Currently if setting ```useBuildkit: true```, skaffold does not actually use buildKit unless ```useDockerCLI:true``` is also present. Fix this by using the dockerCLI if ```useBuildkit: true```.


IsaacPD:
Close since it is a duplicate of #5052

PR #5100, opened 2020-12-03
Capture deployers in skaffold meter struct

IsaacPD:
**Related**: #5058
**Merge Before**: #5105

**Description**
Return an array of yaml tags in order to get the deployers from the config.

**Follow-up Work**

write metrics, and upload them, create relevant documentation, and prompt users to opt-out.

### DIFF
--- a/pkg/skaffold/instrumentation/meter.go
+++ b/pkg/skaffold/instrumentation/meter.go
@@ -66,5 +66,6 @@ func InitMeter(runCtx *runcontext.RunContext, config *latest.SkaffoldConfig) {
 			meter.SyncType[yamltags.GetYamlTag(artifact.Sync)] = true
 		}
 	}
+	meter.Deployers = yamltags.GetYamlTags(config.Deploy.DeployType)
 	meter.BuildArtifacts = len(config.Pipeline.Build.Artifacts)
 }
--- a/pkg/skaffold/yamltags/tags.go
+++ b/pkg/skaffold/yamltags/tags.go
@@ -17,9 +17,12 @@ limitations under the License.
 package yamltags
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
 	"reflect"
 	"strings"
+	"unicode"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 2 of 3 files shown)


pkg/skaffold/errors/errors.go
```diff
@@ -79,6 +80,7 @@ func ShowAIError(err error) error {
 	var knownProblems = append(knownBuildProblems, knownDeployProblems...)
 	for _, v := range append(knownProblems, knownInitProblems...) {
 		if v.regexp.MatchString(err.Error()) {
+			instrumentation.SetErrorCode(v.errCode)
```
tejal29, line 83: Can you add same functionality at L76 when err is already an actionable error.
```
if e, ok := err.(ErrDef); ok {
  instrumentation.SetErrorCode(e.StatusCode())
  return err
}
```

pkg/skaffold/instrumentation/meter.go
```diff
@@ -66,5 +66,10 @@ func InitMeter(runCtx *runcontext.RunContext, config *latest.SkaffoldConfig) {
 			meter.SyncType[yamltags.GetYamlTag(artifact.Sync)] = true
 		}
 	}
+	meter.Deployers = yamltags.GetYamlTags(config.Deploy.DeployType)
```
tejal29, line 69: Are you capturing Dev iterations yet?
If yes,  you also need to add errorCode for current dev iteration
See https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/runner/dev.go#L176 for more information on how Skaffold EventAPI does this.

pkg/skaffold/yamltags/tags.go
```diff
@@ -74,6 +77,27 @@ func GetYamlTag(value interface{}) string {
...
+	buf, err := yaml.Marshal(value)
+	if err != nil {
+		logrus.Warnf("error marshaling %-v", value)
+		return tags
+	}
+
+	r := bufio.NewScanner(bytes.NewBuffer(buf))
+	for r.Scan() {
+		l := r.Text()
+		if !unicode.IsSpace(rune(l[0])) && l != "{}" {
```
tejal29, line 93: Can you explain what this does?
Probably would help if you add example on what the tag value looks like.
e.g. what is the json for input interface.
```
{yaml:"helm,omitempty" {yaml:"kubectl,omitempty"
```

tejal29 (changes requested):

pkg/skaffold/instrumentation/meter.go
IsaacPD: I am not capturing it yet, so that may be more appropriate for the next PR. Capturing errors could probably be included in that PR too instead of this one.

pkg/skaffold/yamltags/tags.go
IsaacPD: If ```value interface{}``` is 
```go
latest.DeployType{HelmDeploy: &HelmDeploy{...}, KustomizeDeploy: &KustomizeDeploy{...}}
```
then this parses that interface as it's raw yaml,
```yaml
kubectl:
  manifests:
  - k8s/*.yaml
kustomize:
  paths:
 - k8s/
```
and returns ["kubectl", "kustomize"], if latest.DeployType{} is passed or an interface in a similar manner than the raw yaml looks like ```{}```.

I tried using reflection to retrieve the tag name, but I could not get a method that generalized easily to values that were never nil.

pkg/skaffold/errors/errors.go
IsaacPD: Included this change in #5105

pkg/skaffold/yamltags/tags.go
tejal29: ah got it. Can you add this description in the code itself? just above the logic

IsaacPD: Done ✌

tejal29 (approved):

PR #5120, opened 2020-12-07
Capture enum flags

IsaacPD:
**Merge after**: #5100

**Description**
Mark which flags are enums by adding an ```IsEnum``` field to ```cmd.Flag``` and using the ```cmd.Flag``` type in all relevant instances that defined flags. 

write metrics, and upload them, create relevant documentation, and prompt users to opt-out.


tejal29:
FYI: Looked at your PR yesterday. Trying to think of a way where this change is minimal and we don't have to edit all flags.

tejal29 (approved):

Issue #5115, opened 2020-12-07
Skaffold v1.17.1 unable to stream build output when copying from directory of larger context


ahmad-musleh:
### Expected behavior

```bash
skaffold build 
Generating tags...
 - service-1 -> 1.2.3.4:32000/service-1:latest
 - service-2 -> 1.2.3.4:32000/service-2:latest
 - service-3 -> 1.2.3.4:32000/service-3:latest
Checking cache...
 - service-1: Not found. Building
 - service-2: Not found. Building
 - service-3: Not found. Building
Building [service-3]...
Sending build context to Docker daemon   5.12kB
Step 1/4 : FROM python:3
Step 2/4 : ARG service_dir
Step 3/4 : COPY $service_dir /general
Step 4/4 : CMD ["python3", "-c", "while True: continue"]
Successfully built 3abe1820e6ec
Successfully tagged 10.4.6.71:32000/service-3:latest
Building [service-1]...
Sending build context to Docker daemon   5.12kB
Step 1/4 : FROM python:3
Step 2/4 : ARG service_dir
Step 3/4 : COPY $service_dir /general
Step 4/4 : CMD ["python3", "-c", "while True: continue"]
Successfully built ca63b88113c0
Successfully tagged 10.4.6.71:32000/service-1:latest
Building [service-2]...
Sending build context to Docker daemon   5.12kB
Step 1/4 : FROM python:3
Step 2/4 : ARG service_dir
Step 3/4 : COPY $service_dir /general
Step 4/4 : CMD ["python3", "-c", "while True: continue"]
Successfully built 8a6e49f61653
```

This was using v1.16.0

### Actual behavior

```bash
skaffold build 
Generating tags...
 - service-1 -> 1.2.3.4:32000/service-1:latest
 - service-2 -> 1.2.3.4:32000/service-2:latest
 - service-3 -> 1.2.3.4:32000/service-3:latest
Checking cache...
 - service-1: Not found. Building
 - service-2: Not found. Building
 - service-3: Not found. Building
Building [service-3]...
Sending build context to Docker daemon   5.12kB
Step 1/4 : FROM python:3
Step 2/4 : ARG service_dir
Step 3/4 : COPY $service_dir /general
Building [service-1]...
Building [service-2]...
unable to stream build output: COPY failed: stat /var/lib/docker/tmp/docker-builder804227384/service_3: no such file or directory
```

### Information

- Skaffold version: version v1.17.0
[...]

PR #5121, opened 2020-12-07
Recompute docker dependencies across dev loops.

### LINKED ISSUE #5110: New files are not synced to running container.
When a new file which matches the glob in sync.manual is created, it is not synced to the running container. Changes to existing files are synced as expected.

This behaviour first appears in 1.17

### Expected behavior
New files which match sync.manual are synced to the running container.

### Actual behavior
The new file is not picked up

### Information

- Skaffold version: >=1.17
- Operating system: MacOS
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2alpha4
kind: Config
metadata:
  name: rally-app
build:
  local:
    useDockerCLI: true
    useBuildkit: true
profiles:
  - name: dev
    build:
      artifacts:
        - image: my-image
          sync:
            manual:
              - src: frontend/src/**/*
                dest: .
                strip: frontend/
              - src: frontend/public/**/*
                dest: .
                strip: frontend/
          docker:
            dockerfile: frontend/Dockerfile
    deploy:
      kustomize:
        paths:
          - kubernetes/environments/dev
      kubeContext: docker-desktop

```

tejal29:
fixes #5110 & #5115 

In #4896, we cache results for `docker.GetDependencies` since there were getting computed atleast 2 times for docker builds.
However, this PR ended up caching the results across dev loops. 

In this PR, 
- add `GetCachedDependencies` which will be used in subsequent calls. 
- change the key for caching dependencies result from "absolute dockerfile path" to "artifact" name
This fixes issue in #5115 where same dockerfile was used in different artifacts. 

### Alternate approach considered,
1. To create a  new [`dependencyCache`](https://github.com/GoogleContainerTools/skaffold/blob/e2a4106f6b2b328a7a3bc66f957361b625a43b9f/pkg/skaffold/docker/dependencies.go#L33) for each loop.

- this would have required a huge refactor as `docker.GetDependency` is a public method.

2. To invalidate the `dependencyCache` at end of each [dev loop](https://github.com/GoogleContainerTools/skaffold/blob/e2a4106f6b2b328a7a3bc66f957361b625a43b9f/pkg/skaffold/runner/dev.go#L122)
```
event.DevLoopComplete(r.devIteration)
// invalidate all cached GetDependencies.
docker.InvalidateDependencyCache()
logger.Unmute()

```
This would have required adding `pkg/skaffold/docker` package dependency in `pkg/skaffold/runner`.  If all dependencies were cached, we could have added `InvalidateDependencyCache()` to `Builder` interface. However only docker dependencies are cached, so went with the approach to add a new method `GetCachedDependencies` and use it when creating tar context

### Testing notes
1. make 
2. go to examples/hot-reload
3. `skaffold dev`
4. Add a new js file `touch node/src/new.js`
5. Verify a dev loop is started.

```
Watching for changes...
[node] Example app listening on port 3000!
INFO[0040] files added: [node/src/new.js] 
Syncing 1 files for node-example:682863e6ee15639b6fceb28c79308b206f6fb329188e18907fc0a766e77b2cc0
INFO[0040] Copying files: map[node/src/new.js:[/home/node/app/src/new.js]] to node-example:682863e6ee15639b6fceb28c79308b206f6fb329188e18907fc0a766e77b2cc0 

Watching for changes...
```

### DIFF
--- a/pkg/skaffold/build/custom/dependencies.go
+++ b/pkg/skaffold/build/custom/dependencies.go
@@ -30,11 +30,10 @@ import (
 )
 
 // GetDependencies returns dependencies listed for a custom artifact
-func GetDependencies(ctx context.Context, workspace string, a *latest.CustomArtifact, cfg docker.Config) ([]string, error) {
+func GetDependencies(ctx context.Context, workspace string, artifactName string, a *latest.CustomArtifact, cfg docker.Config) ([]string, error) {
 	switch {
 	case a.Dependencies.Dockerfile != nil:
-		dockerfile := a.Dependencies.Dockerfile
-		return docker.GetDependencies(ctx, workspace, dockerfile.Path, dockerfile.BuildArgs, cfg)
+		return docker.GetDependencies(ctx, getDockerBuildConfig(workspace, artifactName, a), cfg)
 
 	case a.Dependencies.Command != "":
 		split := strings.Split(a.Dependencies.Command, " ")
@@ -52,4 +51,9 @@ func GetDependencies(ctx context.Context, workspace string, a *latest.CustomArti
 	default:
 		return list.Files(workspace, a.Dependencies.Paths, a.Dependencies.Ignore)
 	}
+}
+
+func getDockerBuildConfig(ws string, artifact string, a *latest.CustomArtifact) docker.BuildConfig {
+	dockerfile := a.Dependencies.Dockerfile
+	return docker.NewBuildConfig(ws, artifact, dockerfile.Path, dockerfile.BuildArgs)
 }
--- a/pkg/skaffold/util/store.go
+++ b/pkg/skaffold/util/store.go
@@ -56,7 +56,7 @@ func (o *SyncStore) Exec(key string, f func() interface{}) interface{} {
 
 // Store will store the results for a key in a cache
 // This function is not safe to use if multiple subroutines store the
-// result for the same artifact.
+// result for the same key.
 func (o *SyncStore) Store(key string, r interface{}) {
 	o.results.Store(key, r)
 }
--- a/pkg/skaffold/util/store.go
+++ b/pkg/skaffold/util/store.go
@@ -54,6 +54,15 @@ func (o *SyncStore) Exec(key string, f func() interface{}) interface{} {
 	return val
 }
 
+// Store will store the results for a key in a cache
+// It makes sure only one execution is in-flight for a given key at any time.
+func (o *SyncStore) Store(key string, r interface{}) {
+	o.sf.Do(fmt.Sprintf("%s-put", key), func() (interface{}, error) {
+		o.results.Store(key, r)
+		return nil, nil
+	})
+}
+
 // NewSyncStore returns a new instance of `SyncStore`
 func NewSyncStore() *SyncStore {
 	return &SyncStore{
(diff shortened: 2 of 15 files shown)


pkg/skaffold/util/store.go
```diff
@@ -54,6 +54,15 @@ func (o *SyncStore) Exec(key string, f func() interface{}) interface{} {
 	return val
 }
 
+// Store will store the results for a key in a cache
+// It makes sure only one execution is in-flight for a given key at any time.
+func (o *SyncStore) Store(key string, r interface{}) {
+	o.sf.Do(fmt.Sprintf("%s-put", key), func() (interface{}, error) {
```
gsquared94, line 60: I think it's some sort of anti-pattern to call `sf.Do` with different function callbacks. Also maybe it can cause a bug since it's not guaranteed which function result will be returned if there are multiple goroutines calling different functions. WDYT?

tejal29: The `singleflight.Do` documentation does not say anything like that. Looking at the code, it based on the "key" string.

The reason, i have it here is because, `sync.Map` does not support reading/writing multiple keys across subroutines.

The chances of `Store` for an artifact key being called across multiple go subroutines are 0. i can drop the `sf.Do` here.

gsquared94: I totally missed the different key. You're right. We can do without it.

tejal29:
Hold, i am also trying to fix #5115 which is result of #4896

tejal29:
Performance stats

I ran `GetDependencies` on https://github.com/GoogleCloudPlatform/microservices-demo and did the time analysis on 
Number of Copy Command | Number of Files |  Build Time in Ms | Dev Loop Time spent in `GetDependencies`
-- | -- | -- | --
1 | 11 |  7.961 | 15.922
2 | 10 | 3.558 | 7.116
3 | 9 |  11.09 | 22.18
3 | 10 |  1.718 | 3.436
4 | 9 |  2.395 | 4.79
5 | 62 | 4.82 | 9.64
6 | 5 |  3.884 | 7.768
6 | 10 | 4.486 | 8.972
8 | 14 |  19.509 | 39.018
9 | 9 | 4.103 | 8.206
9 | 10 |  6.569 | 13.138


image

tejal29:
@gsquared94, @briandealwis and @nkubala please take a look. 
We could get into `gcloud` if released today.

pkg/skaffold/build/custom/dependencies.go
```diff
@@ -53,3 +52,8 @@ func GetDependencies(ctx context.Context, workspace string, a *latest.CustomArti
 		return list.Files(workspace, a.Dependencies.Paths, a.Dependencies.Ignore)
 	}
 }
+
+func getDockerBuildConfig(ws string, artifact string, a *latest.CustomArtifact) docker.BuildConfig {
+	dockerfile := a.Dependencies.Dockerfile
```
gsquared94, line 57: should we normalize the Dockerfile path here itself instead of later?

gsquared94 (approved):
lgtm

pkg/skaffold/build/custom/dependencies.go
tejal29: That is a good suggestion. So i `normalizeDockerpath` is used a lot many places. while going through the code, there is a huge scope of avoiding repetition across `docker.SyncMap` and `docker.GetDependencies`

I don't want to error out here in this stage of creating a `BuildConfig`

Issue #5054, opened 2020-11-23
WSL2 using --vm-driver=docker for minikube results in "container <name> is waiting to start... <image> can't be pulled"


cheslijones:
Testing once again in WSL2 to see if it is actually usable yet with `docker` + `kubectl` + `minikube` + `skaffold` which has been a bit of fools errand for over a year now. Using Ubuntu 20.04 in WSL2. This works perfectly fine in macOS with `hyperkit` and Linux with `kvm`.

# Expected behavior

```
minikube start --vm-driver=docker

git clone https://github.com/cheslijones/tls-dev-k8s.git
cd tls-dev-k8s

skaffold dev
```

It should build the `client` image and deploy to cluster in `minikube`.

### Actual behavior
```
Waiting for deployments to stabilize...
 - deployment/client-deployment-dev: container client is waiting to start: client:59211d1cc3475afb8860023d6b0785b4e03b12f77b8e3f44d87f75177eaf3bab can't be pulled
    - pod/client-deployment-dev-85dbc8587b-4w4rm: container client is waiting to start: client:59211d1cc3475afb8860023d6b0785b4e03b12f77b8e3f44d87f75177eaf3bab can't be pulled
 - deployment/client-deployment-dev failed. Error: container client is waiting to start: client:59211d1cc3475afb8860023d6b0785b4e03b12f77b8e3f44d87f75177eaf3bab can't be pulled.
Cleaning up...
```
### Information

- Skaffold version: v1.16.0
- Operating system: WSL2 in Windows 10
- Contents of skaffold.yaml: 

```
apiVersion: skaffold/v2beta1
kind: Config
build:
  artifacts:
  - image: client
    context: client
    sync:
      manual:
      - src: 'src/**/*.js'
        dest: .
      - src: 'src/**/*.jsx'
        dest: .
      - src: 'package.json'
        dest: .
      - src: 'public/**/*.html'
        dest: .
      - src: 'src/assets/sass/**/*.scss'
        dest: .
      - src: 'src/build/**/*.js'
        dest: .
    docker:
      dockerfile: Dockerfile.dev
  local:
    push: false
deploy:
  kubectl:
    manifests:
      - k8s/ingress-ip.yaml 
      # - k8s/ingress-domain.yaml 
      - client/deployment.yaml
```

### Steps to reproduce the behavior

```
minikube start --vm-driver=docker

git clone https://github.com/cheslijones/tls-dev-k8s.git
cd tls-dev-k8s

[...]

PR #5124, opened 2020-12-09
remove wsl detection logic

IsaacPD:
Fixes: #5085 
**Related**: #5054 

**Description**
The previous WSL logic is antiquated as WSL is its own fully functional linux system. There does not need to be any extra hacks or logic when running within WSL so long as the necessary tools as normal are present. Thus using Docker with the WSL2 backend is a must for anyone using WSL as it allows seamless integration with the tools (e.g. skaffold) that use it.

**User facing changes**
None hopefully, potentially a change for the better in how users use skaffold in WSL if they are not already working with all linux binaries from within WSL.


gsquared94 (approved):
lgtm!

PR #5125, opened 2020-12-09
Add flag to save events to a file

priyawadhwa:
Adds the --event-log-file flag, which when set to a filepath will save Skaffold events to a file when `--enable-rpc` is also enabled.

This will make it easier to collect data around performance metrics for the performance dashboards. We can just parse the events in the file after skaffold executes, and figure out how long each stage took.


MarlonGamez:
This looks good, It looks like jsonpb is deprecated in favor of [protojson](https://godoc.org/google.golang.org/protobuf/encoding/protojson) though

priyawadhwa:
Yah I noticed that too -- I briefly looking into upgrading in this PR, but I think upgrading would actually require a lot more work:
- Changing all of our files that require `jsonpb` (not too hard)
- Changing the [Dockerfile](https://github.com/GoogleContainerTools/skaffold/blob/master/hack/proto/Dockerfile) used for upgrading/testing the proto to use `protojson` and upgrade `protoc-gen-go` to match (feel like this is trickier than it sounds)
- Making sure the newly generated proto is backwards compatible with existing users/IDEs

I think that might have to be a separate effort.

MarlonGamez (approved):
LGTM! Thanks for clarifying the proto situation 👍🏼

Issue #5114, opened 2020-12-06
skaffold init: how to process after choosing the Dockerfiles?


brightzheng100:
### Expected behavior

Choose the right Dockerfile(s) and proceed to generate the initial `skaffold.yaml` file.

### Actual behavior

After choosing the Dockerfile(s), I don't have a way to proceed.

### Information

- Skaffold version: `v1.7.1`
- Operating system: Mac
- Contents of skaffold.yaml: N/A

### Steps to reproduce the behavior

1. In an existing `kubebuilder` project, issue `skaffold init`;

2. In the prompt, choose some Dockerfile(s);

```sh
$ skaffold init
? Choose the builder to build image controller  [Use arrows to move, enter to select, type to filter]
  Docker (Dockerfile)
> Docker (_output/release/manager.dockerbuild/Dockerfile)
  Docker (_output/release/syncer.dockerbuild/Dockerfile)
  Docker (_output/release/agent.dockerbuild/Dockerfile)
  None (image not built from these sources)
```

3. After choosing some Dockerfiles, don't know how to proceed:

```sh
$ skaffold init
? Choose the builder to build image controller Docker (_output/release/manager.dockerbuild/Dockerfile)
? Choose the builder to build image gcr.io/kubebuilder/kube-rbac-proxy Docker (_output/release/syncer.dockerbuild/Dockerfile)
? Choose the builder to build image image_url Docker (_output/release/agent.dockerbuild/Dockerfile)
? Choose the builder to build image xxxx/manager-amd64  [Use arrows to move, enter to select, type to filter]
  Docker (Dockerfile)
> None (image not built from these sources)
```

4. Keep typing Enter, the CLI will just clash:

```
? Choose the builder to build image virtualcluster/coredns None (image not built from these sources)
? Choose the builder to build image virtualcluster/etcd-v3.4.0 None (image not built from these sources)
? Choose the builder to build image virtualcluster/manager-amd64 None (image not built from these sources)
? Choose the builder to build image virtualcluster/sample-controller-amd64 None (image not built from these sources)
? Choose the builder to build image virtualcluster/syncer-amd64 None (image not built from these sources)
? Choose the builder to build image virtualcluster/agent-amd64 None (image not built from these sources)
using non standard workspace: _output/release/manager.dockerbuild
using non standard workspace: _output/release/syncer.dockerbuild
using non standard workspace: _output/release/agent.dockerbuild
panic: runtime error: invalid memory address or nil pointer dereference
[...]

PR #5129, opened 2020-12-09
update AlecAivazis/survey to v2

MarlonGamez:
Fixes https://github.com/GoogleContainerTools/skaffold/issues/5114 

**Description**
Bumps the version of https://github.com/AlecAivazis/survey to v2. The previously used version would allow for a situation where we dereference a nil pointer in `skaffold init`

The situation that would cause this is when the user is matching builders to images interactively, if they tried to filter by typing, but their search had no results and they hit enter anyway, it would effectively select nothing and return that value.

With the updated package, if the search returns nothing and they hit enter, nothing will happen. They will have to make a valid choice before moving on.


IsaacPD (approved):
Tried it locally, and it all works as described!

PR #5135, opened 2020-12-10
write metrics to file

IsaacPD:
**Description**
Get exit code from main function. Write skaffold meter struct to a file only if ```SKAFFOLD_METRICS``` environment variable is set and is ```true```. If metrics have been written before, append these metrics to the existing ones and write it out.

**Follow-up Work**

Embed secret into skaffold, and upload metrics to cloud monitoring. Update deploy configuration to write service account keys into skaffold. Create relevant documentation, and prompt users to opt-out.

### DIFF
--- a/pkg/skaffold/instrumentation/meter.go
+++ b/pkg/skaffold/instrumentation/meter.go
@@ -17,11 +17,18 @@ limitations under the License.
 package instrumentation
 
 import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"runtime"
 	"time"
 
+	"github.com/mitchellh/go-homedir"
 	flag "github.com/spf13/pflag"
 
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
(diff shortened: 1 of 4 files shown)


pkg/skaffold/instrumentation/meter.go
```diff
@@ -97,3 +105,29 @@ func AddDevIterationErr(i int, errorCode proto.StatusCode) {
 func AddFlag(flag *flag.Flag) {
 	meter.EnumFlags[flag.Name] = flag
 }
+
+func ExportMetrics(exitCode int) error {
+	home, _ := homedir.Dir()
```
MarlonGamez, line 110: I doubt this will return an error often (if ever), but should we still handle the error anyway?

MarlonGamez (commented):
This looks good, just one QQ

pkg/skaffold/instrumentation/meter.go
IsaacPD: Yeah makes sense to do so. Done!

MarlonGamez (approved):
LGTM!

Issue #5123, opened 2020-12-08
"Deploy Docs controller" build step failing


IsaacPD:
### Expected behavior
On each push to master all build triggers succeed
### Actual behavior
"Deploy docs controller on push to master" step is failing due to inconsistent vendoring when building the webhook [Dockerfile](https://github.com/GoogleContainerTools/skaffold/blob/master/deploy/webhook/Dockerfile).

### Information
<details>
<summary>Build Log</summary>

```
Already have image (with digest): gcr.io/k8s-skaffold/skaffold:v0.18.0
Starting build...
Building [gcr.io/k8s-skaffold/docs-controller]...
Sending build context to Docker daemon  538.1kB
Step 1/30 : FROM alpine:3.10 as download-docsy
3.10: Pulling from library/alpine
21c83c524219: Already exists
Digest: sha256:f0e9534a598e501320957059cb2a23774b4d4072e37c7b2cf7e95b241f019e35
Status: Downloaded newer image for alpine:3.10
 ---> be4e4bea2c2e
Step 2/30 : ENV DOCSY_VERSION a7141a2eac26cb598b707cab87d224f9105c315d
 ---> Running in aa27faa2a29c
 ---> f9e17a1192f0
Step 3/30 : ENV DOCSY_URL https://github.com/google/docsy.git
 ---> Running in d29794617f7a
 ---> 6175f4b21690
Step 4/30 : RUN apk add --no-cache git
 ---> Running in 0399777841c1
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/community/x86_64/APKINDEX.tar.gz
(1/6) Installing ca-certificates (20191127-r2)
(2/6) Installing nghttp2-libs (1.39.2-r1)
(3/6) Installing libcurl (7.66.0-r1)
(4/6) Installing expat (2.2.8-r0)
(5/6) Installing pcre2 (10.33-r0)
(6/6) Installing git (2.22.4-r0)
Executing busybox-1.30.1-r3.trigger
Executing ca-certificates-20191127-r2.trigger
OK: 21 MiB in 20 packages
 ---> 11693b5bb1cf
Step 5/30 : WORKDIR /docsy
 ---> Running in 43b793d8a783
 ---> 731133f77499
[...]

PR #5137, opened 2020-12-11
remove last traces of vendor

IsaacPD:
Fixes: #5123 
**Related**: #5090 

**Description**
The build was failing because we no longer have the vendor folder, so by removing ```-mod-vendor``` the build now completes successfully. To verify just run ```skaffold -f deploy/webhook/skaffold.yaml build``` and the build will succeed unlike on master.

image


MarlonGamez (approved):
Tried locally, seems good!

PR #5141, opened 2020-12-11
new init prompt function and helper functions

MarlonGamez:
**Related**: https://github.com/GoogleContainerTools/skaffold/pull/4915

**Description**
This PR adds the functions that will be used to implement the transparent init feature.
The main is the ConfirmInitOptions() prompt function.

There are also some new util functions defined in pkg/skaffold/initializer/util

**Follow-up Work**
Use these functions to implement transparent init

### DIFF
--- a/pkg/skaffold/initializer/util/util.go
+++ b/pkg/skaffold/initializer/util/util.go
@@ -20,6 +20,7 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/misc"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/yamltags"
 )
 
 // ListBuilders returns a list of builder names being used in the given build config.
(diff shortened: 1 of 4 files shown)


pkg/skaffold/initializer/util/util.go
```diff
@@ -0,0 +1,60 @@
...
+	results := util.NewStringSet()
+	for _, artifact := range build.Artifacts {
+		results.Insert(misc.ArtifactType(artifact))
+	}
+
+	return results.ToList()
+}
+
+// ListDeployers returns a list of deployer names being used in the given deploy config.
+func ListDeployers(deploy *latest.DeployConfig) []string {
```
IsaacPD, line 40: Looks good just thought this function might be useful since I used it for the same use case so I think it could help here too by using it as the underlying function. It just uses the yaml that deploy *latest.DeployConfig unmarshals to, to get the list of deployers.
https://github.com/GoogleContainerTools/skaffold/blob/d781368a5845d71c687e179456cc6ead58e971e8/pkg/skaffold/instrumentation/meter.go#L87
https://github.com/GoogleContainerTools/skaffold/blob/1f200b186c4a28858b39e9f8e689d5606946fca1/pkg/skaffold/yamltags/tags.go#L80-L92

IsaacPD: Also realizing there's tab right before ```kubectl``` in the comment that shouldn't be there 😅

IsaacPD (approved):

pkg/skaffold/initializer/util/util.go
MarlonGamez: thanks for sharing this, I think your function is probably what I should use here because it should be fine in the case of a new deployer being added to skaffold

PR #5142, opened 2020-12-11
replace our user input handling with survey package

MarlonGamez:
**Description**
This PR removes some inconsistency in our input handling by using https://github.com/AlecAivazis/survey wherever we can for `skaffold init`. This will provide a more consistent experience for this part of skaffold.

I've also added some unit tests for prompt functions that were missing them

There is now validation on the port forward information input, which will require the user to input a number or leave it empty.

### DIFF
--- a/pkg/skaffold/initializer/prompt/prompt.go
+++ b/pkg/skaffold/initializer/prompt/prompt.go
@@ -34,6 +34,7 @@ var (
 	BuildConfigFunc         = buildConfig
 	PortForwardResourceFunc = portForwardResource
 	askOne                  = survey.AskOne
+	ask                     = survey.Ask
 )
 
 func buildConfig(image string, choices []string) (string, error) {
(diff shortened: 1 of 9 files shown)


pkg/skaffold/initializer/prompt/prompt.go
```diff
@@ -64,36 +64,34 @@ func WriteSkaffoldConfig(out io.Writer, pipeline []byte, generatedManifests map[
...
+		Prompt: &survey.Input{Message: fmt.Sprintf("Select port to forward for %s (leave blank for none): ", imageName)},
+		Validate: func(val interface{}) error {
+			str := val.(string)
+			if _, err := strconv.Atoi(str); err != nil {
+				return errors.New("response must be a number, or empty")
+			}
+			return nil
+		},
+	}
+	err := ask([]*survey.Question{prompt}, &response)
```
IsaacPD, line 92: Could we use```askOne()``` here?

MarlonGamez: Unfortunately I don't think we can. To add a validation function to a question, you have to use the survey.Question type. survey.AskOne() takes survey.Prompt as a parameter, while survey.Ask() uses survey.Question.

IsaacPD: Gotcha, thanks for clarifying.

IsaacPD (approved):

PR #5146, opened 2020-12-14
fix usage of set-env in Github integration actions

IsaacPD:
**Description**
Update action due to https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

This fixes these failing actions
https://github.com/GoogleContainerTools/skaffold/actions/runs/420481722
https://github.com/GoogleContainerTools/skaffold/actions/runs/420472524

### DIFF
--- a/.github/workflows/integration-darwin.yml
+++ b/.github/workflows/integration-darwin.yml
@@ -32,7 +32,7 @@ jobs:
 
     - name: Skaffold binary version
       run: |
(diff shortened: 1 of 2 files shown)


MarlonGamez (approved):
Looks like we missed this when doing https://github.com/GoogleContainerTools/skaffold/pull/5057/
Thanks for the fix!

Issue #5140, opened 2020-12-11
Buildpack auto sync performs a full re-build on first change


etanshaul:
Tested using skaffold `1.17.2`

1) Clone the Cloud Code Go hello-world k8s sample: https://github.com/GoogleCloudPlatform/cloud-code-samples/tree/master/golang/go-hello-world
2) Configure your build to use buildpacks:
```
build:
  tagPolicy:
    sha256: {}
  artifacts:
  - image: go-hello-world
    context: .
    buildpack:
      builder: gcr.io/buildpacks/builder:v1
```
3) Run skaffold dev
4) Make a change in `main.go`

**Expected**: Skaffolds auto-sync for buildpacks would automtically sync the change without performing a full container rebuild.

**Actual**: The full build happens.

**Note**: This only seems to happen unpredictably; in my tests, some changes sync, others trigger a full rebuild (the changes are identical each time - e.g. a simple string modification).

[Example output](https://paste.googleplex.com/6448961318551552) (the first couple changes did a full rebuild, the last change synced).

abbycar:
+1 here's a [video](https://drive.google.com/file/d/1--BCPMmLOLYRn7LCS0HuhVJffNU6HcTg/view?usp=sharing&resourcekey=0-EjNvKzxhinhziZgMrIDicw) of the same behavior

tejal29:
@briandealwis i am going to look into this as a starter bug for buildpacks :)

tejal29:
Some notes from my debugging on command line

1. First error i see, is `.swp` file is being detected as modified which triggers the re-build. 
```
Deployments stabilized in 2.533034424s
Watching for changes...
DEBU[0148] Change detected notify.Create: "/Users/tejaldesai/workspace/cloud-code-samples/golang/go-hello-world/cmd/hello-world/4913" 
DEBU[0148] Change detected notify.Rename: "/Users/tejaldesai/workspace/cloud-code-samples/golang/go-hello-world/cmd/hello-world/main.go~" 
DEBU[0148] Change detected notify.Create: "/Users/tejaldesai/workspace/cloud-code-samples/golang/go-hello-world/cmd/hello-world/main.go" 
DEBU[0148] Change detected notify.Remove: "/Users/tejaldesai/workspace/cloud-code-samples/golang/go-hello-world/cmd/hello-world/main.go~" 
[...]

PR #5156, opened 2020-12-15
[Refactor] Move FSNotify trigger into a separate package

tejal29:
In preparation for #5140,
1. Move fsNotifyTrigger into its own package.
2. Add a function `Ignore` to configure the fsNotifyWatcher to ignore
   - files belonging to predefined dirs like .idea .vscode & .git [ follow up PR]
   - hidden file like .somefile.swp. [ follow up PR]

**User facing change**
None.

**Future Work**
1. Configure the `Ignore` to ignore files mentioned above

### DIFF
--- a/pkg/skaffold/trigger/triggers_test.go
+++ b/pkg/skaffold/trigger/triggers_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/rjeczalik/notify"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	fsNotify "github.com/GoogleContainerTools/skaffold/pkg/skaffold/trigger/fsnotify"
 	"github.com/GoogleContainerTools/skaffold/testutil"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/skaffold/trigger/triggers_test.go
```diff
@@ -222,33 +179,27 @@ func TestStartTrigger(t *testing.T) {
...
+			trigger:     fsNotify.New(nil, func() bool { return false }, 1),
+			mockWatch: func(string, chan<- notify.EventInfo, ...notify.Event) error {
+				return fmt.Errorf("failed to start watch trigger")
 			},
 		},
 	}
 
 	for _, test := range tests {
 		testutil.Run(t, test.description, func(t *testutil.T) {
+			t.Override(&fsNotify.Watch, test.mockWatch)
```
MarlonGamez, line 202: Is this working as intended? Wouldn't the trigger struct already have it's `watchFunc` set before this override happens? Since the `watchFunc` is grabbed from `fsNotify.Watch` upon calling `fsNotify.New`

tejal29: Thanks! fixed it.

MarlonGamez (approved):

Issue #5162, opened 2020-12-17
tryImportMissing always building new image, despite pull suceeding


wheelsandmetal:
I'm trying to use `tryImportMissing` to import images from a remote registry rather than building them locally

I'm seeing this in the `-vdebug` logs

```
time="2020-12-17T10:32:08Z" level=debug msg="Importing artifact IMAGE:TAG from docker registry"
time="2020-12-17T10:32:08Z" level=debug msg="Skipping credential configuration because docker-credential-gcloud is not on PATH."
time="2020-12-17T10:32:13Z" level=debug msg="Added digest for IMAGE:TAG to cache entry"
 - IMAGE: Not found. Building
```
after which it goes on and attempts to build the image. However if I stop it early I can see that it successfully pulled the image to my local docker.

I'm wondering if there is a missing `return found{hash: hash}` [here](https://github.com/GoogleContainerTools/skaffold/blob/d667827c018c592abcfcdd5738f42eb5848adb48/pkg/skaffold/build/cache/lookup.go#L65)?

---
edit:

I'm wrong about adding `return found{hash: hash}`

PR #5165, opened 2020-12-17
fix: lookup image id with tag rather than name during tryImport

wheelsandmetal:
Fixes: #5162 

**Description**

Lookup the imageID with the tag rather than the image name.

My guess at what is happening is that there are 2 images with the same name but different tags and thus it can't produce a singular image ID. In may case the 2 images where:
* The one created by `tagPolicy`
* The one with the immutable local-only skaffold tag (https://skaffold.dev/docs/pipeline-stages/taggers/#how-tagging-works)

In this case the `tag` also includes the image name so I don't think there is any chance that 2 images match again.


briandealwis (approved):
I can reproduce the failure and your change works for me.  Thanks for the fix!

PR #5168, opened 2020-12-17
Set meter.Command before an error could occur with createNewRunner

IsaacPD:
**Merge after**: #5157

**Description**

This change is necessary so we can correctly report the frequency of errors that occur before creating the runner and associate a command with them.

- Add contains function to util.StringSet and use that instead of map[string]bool to check if a command builds/deploys/should collect metrics.
- Set meter.Command in cobra PersistentPreFuncE instead of createNewRunner as the command should be available before any errors occur when creating the runner


tejal29:
Please rebase. Thanks!

tejal29 (approved):

Issue #5074, opened 2020-11-27
context directories with backslashes on Windows do not work


briandealwis:
From [StackOverflow](https://stackoverflow.com/q/65027759/600339) where a user was attempting to build a project from `skaffold init` on Windows

### Expected behavior
`skaffold init` generated a `skaffold.yaml` with paths containing backslashes.  Such paths should cause no problems as YAML does not interpret backslashes in an unquoted scalar, and Skaffold should accept backslash paths on Windows.

```yaml
build:
  artifacts:
  - image: testmicro
    context: src\Microservices
    docker:
      dockerfile: TestMicro\Dockerfile
```

### Actual behavior

> [`@penguinflip`](https://stackoverflow.com/questions/65027759/how-to-run-skaffold-with-a-bootstrapped-net-core-api-docker-project/65029009?noredirect=1#comment114966990_65029009): [...] Interestingly, the init script added the slashes the wrong way - `src\Microservices\TestMicro` (I guess because I'm on Windoze). Copying your code exactly worked, but it doesn't with the slashes reversed. This may be a bug with skaffold init...

### Information

`skaffold.yaml` was generated by `skaffold init`.

- Skaffold version: 1.15.0 (since generated `skaffold.yaml` has `v2beta8`)
- Operating system: Windows
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta8
kind: Config
metadata:
  name: microservices-investigation
build:
  artifacts:
  - image: testmicro
    context: src\Microservices
    docker:
      dockerfile: TestMicro\Dockerfile
deploy:
  kubectl:
    manifests:
    - k8s/TestMicro.yaml
```

PR #5177, opened 2020-12-21
Ensure init generates /-delimted paths

briandealwis:
Fixes: #5074

**Description**
Ensure paths are written using forward slashes during `init`.  Adds a test for Windows.

The choice of where to maintain and use the workspace is a bit confusing.
- Each builder provides an `ArtifactConfig` to capture a detected artifact (e.g., [Jib](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/build/jib/init.go#L40-L46)).
- Each `ArtifactConfig` implements an [`initializers/build.InitBuilder` interface](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/initializer/build/builders.go#L30-L48).  The `ArtifactType()` method generates a config schema's [`ArtifactType`](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/schema/latest/config.go#L873-L894) with the appropriate per- builder information.  This `ArtifactType` is contained with the artifact's workspace within the config schema's [`Artifact`](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/schema/latest/config.go#L873-L894) object.

For now I pass the workspace into the per-builder `ArtifactConfig.ArtifactType(workspace)` functions.  It would seem to make sense for the workspace to be passed around within the per-builder `ArtifactConfig` objects, except that they don't actually generate the `Artifact` object.  So we'd still need to swizzle the workspace when [creating the corresponding `Artifact` objects](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/initializer/build/util.go#L73-L79).  

But it might be worth having the `ArtifactConfig` objects be responsible for generating the `Artifact` object so that they could populate it with sync rules?

### DIFF
--- a/pkg/skaffold/build/buildpacks/init.go
+++ b/pkg/skaffold/build/buildpacks/init.go
@@ -49,7 +49,7 @@ func (c ArtifactConfig) Describe() string {
 }
 
 // ArtifactType returns the type of the artifact to be built.
-func (c ArtifactConfig) ArtifactType() latest.ArtifactType {
+func (c ArtifactConfig) ArtifactType(_ string) latest.ArtifactType {
 	return latest.ArtifactType{
 		BuildpackArtifact: &latest.BuildpackArtifact{
 			Builder: c.Builder,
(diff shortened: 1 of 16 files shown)


MarlonGamez:
this LGTM, I can approve and merge once we get our ci issues resolved

IsaacPD (approved):

PR #5181, opened 2020-12-23
Add flag that prints timestamps in skaffold logs.

jtzwu:
**Description**
Adds a flag that add timestamps to the skaffold logs. 

Before
```
DEBU[0000] found config for context "minikube"          
DEBU[0000] could not parse date ""                      
DEBU[0000] Defaulting build type to local build         
INFO[0000] Using kubectl context: minikube        
```
After
```
DEBU[2020-12-22T17:05:27-08:00] found config for context "minikube"           
DEBU[2020-12-22T17:05:27-08:00] could not parse date ""                       
DEBU[2020-12-22T17:05:27-08:00] Defaulting build type to local build  
```


MarlonGamez:
hey @jtzwu, thanks for opening this PR. Looks like there's a formatting issue here but that's a quick fix. You'll also need to run ./hack/generate-man.sh to update our docs to reflect the added flag. Thanks!

MarlonGamez (approved):
lgtm 👍🏼

PR #5184, opened 2020-12-23
proposal for user generated tagger.

piotr-szybicki:
Some more context for the design proposal. Because at the moment we are already using our own skaffold distro with the mention tagger. Imagine you have a java project with 5 modules:

project:

- commons 1
- commons 2
- executable 1
- executable 2
- executable 3

And following dependencies. 
```
         executable 3                                  executable 2
         /            \                                 /          \            
executable 1       commons 2                      executable 1      commons 1
```
And using multistage docker file that looks like this (a bit of a pseudo code):
```
FROM remoterepo/images/maven_cach:3.6 AS deps_maven_build
COPY /project /project
WORKDIR /project
RUN mvn clean install -Dmaven.javadoc.skip=true -Dmaven.test.skip=true

FROM gcr.io/distroless/java:11 AS executable_3
COPY --chown=1000:1000 --from=deps_maven_build /executable_2.jar /

FROM gcr.io/distroless/java:11 AS executable_2
COPY --chown=1000:1000 --from=deps_maven_build /executable_3.jar /
```
Now forget the existing taggers. `GitTreeSha`  will generate a tag base on a too narrow or too wide scope. And the rest is useless. Even the proposed `inputDigest` tagger will not do the job because it will hash every file that is touched by the `COPY` command in front of it. 

Any build tool (in this case maven ) has the full knowledge of the dependency graph that is part of a given executable. I think adding this tagger would create an interesting point of integration. I wrote a plugin for maven and it works extremely well. I think a similar thing can be written for gradle, bazel, npm, cmake, rake etc....

Looking for what the community has to say. 

I created the [PR](https://github.com/piotr-szybicki/skaffold/pull/1) in my own fork how could that work.

### DIFF
--- a/docs/design_proposals/digest-tagger.md
+++ b/docs/design_proposals/digest-tagger.md
@@ -1,4 +1,4 @@
-# Improve taggers
+1# Improve taggers
 
 * Author(s): David Gageot (@dgageot)
 * Date: 4 September 2019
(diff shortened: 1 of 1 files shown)


piotr-szybicki:
@googlebot I signed it!

MarlonGamez:
hey @piotr-szybicki, thanks for bringing up this proposal! I'd like to bring this up with the team to talk about after everyone is back from vacation. I'll update this if I get more information then :)

piotr-szybicki:
A friend of mine did some more research and it turns out that JIB is doing exactly what I described. I think this is still worth exploring because JIB can only work for maven and Gradle builds, but I guess it will cover a lot of cases. It definitely covers mine. That is as soon as the [PR](https://github.com/GoogleContainerTools/skaffold/pull/5192)  for input digest gets merged.

tejal29 (approved):

PR #5186, opened 2020-12-23
Add transparent init

### LINKED ISSUE #1273: Transparent `skaffold init`
When no skaffold configuration is found on disk, maybe we could run `skaffold init` in memory and use that.
I think it would be a very simple way of getting started with Skaffold and also a good way of supporting things like compose files transparently.
We’d, of course, print a warning to explain what’s going on and which configuration Skaffold is using.

MarlonGamez:
Fixes #1273 
**Related** #4915, #5037, #5141, #5155

**Description**
Adds the flow that handles users running commands without a skaffold config file being present. 

If users try to run a command without a skaffold config such as `skaffold dev`, skaffold will tell them that no config is found, and that it is trying to make one. If it succeeds, it will prompt the user, making sure that they want to continue.

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -461,20 +461,22 @@ var flagRegistry = []Flag{
 		DefinedOn:     []string{"test", "deploy"},
 	},
 	{
-		Name:          "try-transparent-init",
+		Name:          "auto-create-config",
 		Usage:         "If true, skaffold will try to create a config for the user's run if it doesn't find one",
-		Value:         &opts.TryTransparentInit,
+		Value:         &opts.AutoCreateConfig,
 		DefValue:      true,
 		FlagAddMethod: "BoolVar",
 		DefinedOn:     []string{"debug", "dev", "run"},
+		IsEnum:        true,
 	},
 	{
-		Name:          "skip-confirmation",
+		Name:          "assume-yes",
 		Usage:         "If true, skaffold will skip yes/no confirmation from the user and default to yes",
-		Value:         &opts.SkipConfirmation,
+		Value:         &opts.AssumeYes,
 		DefValue:      false,
 		FlagAddMethod: "BoolVar",
 		DefinedOn:     []string{"debug", "dev", "run"},
+		IsEnum:        true,
 	},
 }
 
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -460,6 +460,22 @@ var flagRegistry = []Flag{
 		FlagAddMethod: "Var",
 		DefinedOn:     []string{"test", "deploy"},
 	},
+	{
+		Name:          "try-transparent-init",
+		Usage:         "If true, skaffold will try to create a config for the user's run if it doesn't find one",
+		Value:         &opts.TryTransparentInit,
+		DefValue:      true,
+		FlagAddMethod: "BoolVar",
+		DefinedOn:     []string{"debug", "dev", "run"},
+	},
+	{
+		Name:          "skip-confirmation",
+		Usage:         "If true, skaffold will skip yes/no confirmation from the user and default to yes",
+		Value:         &opts.SkipConfirmation,
+		DefValue:      false,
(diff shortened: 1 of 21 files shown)


cmd/skaffold/app/cmd/runner.go
```diff
@@ -71,11 +75,58 @@ func createNewRunner(opts config.SkaffoldOptions) (runner.Runner, []*latest.Skaf
...
 	if err != nil {
 		if os.IsNotExist(errors.Unwrap(err)) {
-			return nil, nil, fmt.Errorf("skaffold config file %s not found - check your current working directory, or try running `skaffold init`", opts.ConfigurationFile)
+			color.Default.Fprintf(out, "Skaffold config file %s not found - Trying to create one for you...\n", opts.ConfigurationFile)
+			config, err := initializer.Transparent(context.Background(), out, initConfig.Config{Opts: opts})
+			if err != nil {
+				return nil, nil, fmt.Errorf("unable to generate skaffold config file automatically - try running `skaffold init`: %w", err)
+			}
+			if config == nil {
+				return nil, nil, fmt.Errorf("user not continuing")
```
gsquared94, line 124: should this be a bit more verbose?

gsquared94:
looks good. Should we add two flags:
- one to skip the confirmation `? Would you like to continue? Yes`
- one to skip this feature and fail as before

cmd/skaffold/app/cmd/runner.go
MarlonGamez: Is there anything specific you feel should be mentioned here? Maybe something like
> user not continuing with generated config

gsquared94: I was thinking something more fatal like:
```go
return nil, nil, errors.New("unable to generate skaffold config file automatically - try running `skaffold init`: action cancelled by user")
```

cmd/skaffold/app/cmd/flags.go
```diff
@@ -460,6 +460,22 @@ var flagRegistry = []Flag{
 		FlagAddMethod: "Var",
 		DefinedOn:     []string{"test", "deploy"},
 	},
+	{
+		Name:          "try-transparent-init",
```
gsquared94, line 464: `auto-create-config`?

cmd/skaffold/app/cmd/flags.go
```diff
@@ -460,6 +460,22 @@ var flagRegistry = []Flag{
...
+		Usage:         "If true, skaffold will try to create a config for the user's run if it doesn't find one",
+		Value:         &opts.TryTransparentInit,
+		DefValue:      true,
+		FlagAddMethod: "BoolVar",
+		DefinedOn:     []string{"debug", "dev", "run"},
+	},
+	{
+		Name:          "skip-confirmation",
```
gsquared94, line 472: `--assume-yes`? also will this be globally applied to skip prompt from all user interactions?

cmd/skaffold/app/cmd/runner.go
MarlonGamez: makes sense, I can change that

cmd/skaffold/app/cmd/flags.go
MarlonGamez: done

cmd/skaffold/app/cmd/flags.go
MarlonGamez: done

cmd/skaffold/app/cmd/flags.go
```diff
@@ -460,6 +460,22 @@ var flagRegistry = []Flag{
 		FlagAddMethod: "Var",
 		DefinedOn:     []string{"test", "deploy"},
 	},
+	{
+		Name:          "auto-create-config",
```
IsaacPD, line 464: Could you set the `IsEnum` field to `true` for this flag and `assume-yes`?

MarlonGamez: yep, done

IsaacPD (approved):

gsquared94 (approved):
lgtm!

PR #5205, opened 2021-01-06
Add go tag `filepath` that converts marked fields to absolute paths

gsquared94:
This PR introduces a new  tag `skaffold:"filepath"` for [`config.go`](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/schema/latest/config.go) which converts the marked field to an absolute path.
```go
type Artifact struct {
	// Workspace is the directory containing the artifact's sources.
	// Defaults to `.`.
	Workspace string `yaml:"context,omitempty" skaffold:"filepath"`
        ...
}

type KubectlDeploy struct {
	// Manifests lists the Kubernetes yaml or json manifests.
	// Defaults to `["k8s/*.yaml"]`.
	Manifests []string `yaml:"manifests,omitempty" skaffold:"filepath"`
        ...
}
```

 This is useful since the current code assumes all relative paths in the skaffold config are defined from the `os.Getwd()` directory. With https://github.com/GoogleContainerTools/skaffold/pull/5160 we'll open up support for multiple skaffold configs. Doing an upfront file path massaging avoids having to maintain the info for base paths for each config individually which can be error prone.

#### User facing changes:

There are no user facing changes. We retain the current behavior of relative path base being the CWD.

### DIFF
--- a/cmd/skaffold/app/cmd/runner.go
+++ b/cmd/skaffold/app/cmd/runner.go
@@ -37,8 +37,6 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/validation"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/update"
-	pkgutil "github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
-	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/yamltags"
 )
 
 // For tests
@@ -102,14 +100,6 @@ func runContext(opts config.SkaffoldOptions) (*runcontext.RunContext, []*latest.
 		}
 		if err := defaults.Set(config, setDefaultDeployer); err != nil {
 			return nil, nil, fmt.Errorf("setting default values: %w", err)
-		}
-		cwd, err := pkgutil.RealWorkDir()
-		if err != nil {
-			return nil, nil, err
-		}
-		// base for relative paths in the `skaffold.yaml` is the cwd and not the config dir
-		if err := yamltags.SetAbsFilePaths(config, cwd); err != nil {
-			return nil, nil, fmt.Errorf("setting absolute filepaths: %w", err)
 		}
 		pipelines = append(pipelines, config.Pipeline)
 		configs = append(configs, config)
--- a/cmd/skaffold/app/cmd/runner.go
+++ b/cmd/skaffold/app/cmd/runner.go
@@ -37,6 +37,7 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/validation"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/update"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/yamltags"
 )
 
 // For tests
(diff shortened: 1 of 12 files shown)


briandealwis:
`integration/TestGeneratePipeline` tests are failing as the context directories don't match.
```diff
@@ -12,7 +12,7 @@
           build:
             artifacts:
             - image: gcr.io/first-week-244218/getting-started-pipeline
-              context: /skaffold/integration/testdata/generate_pipeline/existing_other
+              context: .
               kaniko: {}
             tagPolicy:
               gitCommit: {}
```

cmd/skaffold/app/cmd/runner.go
```diff
@@ -101,6 +102,14 @@ func runContext(opts config.SkaffoldOptions) (*runcontext.RunContext, []*latest.
 		if err := defaults.Set(config, setDefaultDeployer); err != nil {
 			return nil, nil, fmt.Errorf("setting default values: %w", err)
 		}
+		cwd, err := os.Getwd()
```
briandealwis, line 105: Could you extract the following code to a utility function and use it here?

https://github.com/GoogleContainerTools/skaffold/blob/35214eb817bff5d541d101d36ad3046be81c183b/pkg/skaffold/trigger/fsnotify/workdir.go#L24-L40

pkg/skaffold/yamltags/paths.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+		}
+		return errs
+	default:
+		return nil
+	}
+}
+
+func filepathTagExists(f reflect.StructField) bool {
+	yamltags, ok := f.Tag.Lookup("yamltags")
```
briandealwis, line 117: `yamltags` just doesn't seem to be appropriate.  Is there some prior art that we could leverage here, like the [Go validator's `dir` and `file` types?](https://github.com/go-playground/validator#other)?

pkg/skaffold/yamltags/paths.go
```diff
@@ -0,0 +1,128 @@
...
+	"fmt"
+	"path/filepath"
+	"reflect"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+// SetAbsFilePaths recursively sets all fields marked with the yamltag `filepath` to absolute paths
+func SetAbsFilePaths(s interface{}, base string) error {
```
briandealwis, line 29: `MakeFilePathsAbsolute`?

pkg/skaffold/yamltags/paths.go
gsquared94: so do you mean like this?
```go
type Artifact struct {
	// Workspace is the directory containing the artifact's sources.
	// Defaults to `.`.
	Workspace string `yaml:"context,omitempty" validate:"dir"`
        ...
}
```

We are already using `yamltags` as the key for custom validation tags. Although I don't mind making this one different.

gsquared94: maybe `convert: "dir"`? I think validate isn't correct either since we aren't validating the actual directory exists.

gsquared94: I think that the [`validator`](https://godoc.org/gopkg.in/go-playground/validator.v9) package is validating the dir or file exists on the filesystem, which isn't what we're trying to do.

gsquared94:
After discussing with @briandealwis, removed the invocation of `MakeFilePathsAbsolute` from the parsed skaffold config. We'll use this function for configs that are resolved as dependency from other places, in subsequent PR.

MarlonGamez:
I think this looks good. As for Brian's point about using `yamltags`, I think that if we're already doing this, then we should be consistent here. If we want to change that name in the future, maybe we should have a discuss about it

pkg/skaffold/yamltags/paths.go
briandealwis: Although we are bashing these values to make them reference to files and directories on disk, these values may be relative to places other than the current directory (e.g., within an artifact workspace).

briandealwis:
My point is really that these has nothing to do with YAML.  Our `yamltags` describes structural information and constraints specific to YAML.  These fields are filepaths, even if we were pulling in a definition encoded in JSON or (gasp) XML.

Why don't we add a `skaffold` tag.  So we'd use `skaffold:filepath`?  That's clearly under our control and independent of any representation formats.

gsquared94:
Updated. PTAL @briandealwis, @MarlonGamez

briandealwis (approved):
Should `pkg/skaffold/yamltags/paths*.go` be moved to `pkg/skaffold/config`?

Issue #5106, opened 2020-12-03
Consider extending default status check timeout


fsladkey:
## Consider extending default status check timeout

After discussing some deployments that (due to automatic cluster scaling) often took longer than the two minute status-check timeout, @tejal29 mentioned that because status checks now fail deployments immediately if they encounter an irrecoverable error, it may be desirable to extend the default status check timeout. Creating this issue for further discussion.

tejal29:
Thanks @fsladkey for opening the issue. I have added this in next milestone.

tejal29:
Sorry this spilled my attention. 
I will try to see if we can make this happen this release. 
/cc @gsquared94 , @nkubala , @marlon-gamez @IsaacPD any takers?

MarlonGamez:
@tejal29 I can work on this 👍 do we have an idea of what the default timeout should be set to?

tejal29:
The default time should be 10 mins which is the default for deployments to stabilize.
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#failed-deployment

PR #5247, opened 2021-01-19
change default status check timeout to 10 minutes

MarlonGamez:
Fixes: #5106

**Description**
Changes the default timeout of status check to 10 minutes, according to conversation on #5106 

Users deploying with skaffold using status-check will have a longer period of time before their deployment times out


briandealwis (approved):

sampath:
Thanks for this fix :) @MarlonGamez

PR #5250, opened 2021-01-20
Fix statik directory structure

IsaacPD:
Fixes: #5253 

**Description**
For some reason the way that I had generated the statik file in #5157 for the metrics credentials conflicted with the existing method that creats the statik file structure. This resulted in calls to `skaffold schema get` and `skaffold credits` returning `file not found` if running skaffold built from master with all the licenses and etc. being built in.

In order to fix this I followed similarly to the previous statik generation and moved it into the temp directory that ultimately gets exported to `cmd/skaffold/app/cmd/statik/statik.go`

Fortunately this bug has not made it into a release so we can fix this and hopefully no one will notice 😅. If not fixed this would also ultimately affect the build release script since it relies on `skaffold credits` https://github.com/GoogleContainerTools/skaffold/blob/dafb4152be74c7cd7d363790f2099f2b6f057ac3/deploy/skaffold/Dockerfile#L23

**User facing changes**
`skaffold credits` and `skaffold schema get` work


MarlonGamez (approved):
tried locally, seems to be working as intended :)

Issue #5248, opened 2021-01-20
Maven builds sometimes fail with connection resets


briandealwis:
Seen on Travis surprisingly often:
```
        [builder] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.21.0:test (default-test) on project hello-spring-boot: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.21.0:test failed: Plugin org.apache.maven.plugins:maven-surefire-plugin:2.21.0 or one of its dependencies could not be resolved: The following artifacts could not be resolved: org.apache.maven.surefire:surefire-api:jar:2.21.0, org.apache.maven.surefire:surefire-logger-api:jar:2.21.0: Could not transfer artifact org.apache.maven.surefire:surefire-api:jar:2.21.0 from/to central (https://repo.maven.apache.org/maven2): Transfer failed for https://repo.maven.apache.org/maven2/org/apache/maven/surefire/surefire-api/2.21.0/surefire-api-2.21.0.jar: Connection reset -> [Help 1]
```

Similar complaints abound on the internet.  Suspicion is that Maven Central's CDN times out connections resulting in a connection reset.

Some relevant reading:
- [WAGON-545](https://issues.apache.org/jira/browse/WAGON-545) and [WAGON-603](https://issues.apache.org/jira/browse/WAGON-603)
- actions/virtual-environments/issues/1499

briandealwis:
Still [being seen](https://api.travis-ci.com/v3/job/473945864/log.txt):
```        [builder] [ERROR] Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:2.0.5.RELEASE:repackage (default) on project hello-spring-boot: Execution default of goal org.springframework.boot:spring-boot-maven-plugin:2.0.5.RELEASE:repackage failed: Plugin org.springframework.boot:spring-boot-maven-plugin:2.0.5.RELEASE or one of its dependencies could not be resolved: Could not transfer artifact org.apache.maven:maven-archiver:jar:2.6 from/to central (https://repo.maven.apache.org/maven2): Transfer failed for https://repo.maven.apache.org/maven2/org/apache/maven/maven-archiver/2.6/maven-archiver-2.6.jar: Connection reset -> [Help 1]

```

briandealwis:
[...]

PR #5251, opened 2021-01-20
Configure maven connection pool TTL to avoid connection resets from stale connections

briandealwis:
**Description**
Updates Maven wrapper properties to pull down Maven 3.6.3 to ensure we have latest bug fixes.
Adds JVM command-switches to change the Maven Wagon HTTP transport's TTL to 2 minutes.

Fixes: #5248 🤞


briandealwis:
Used commit-bit to fix tests — failing test is TestDebug/helm, being tracked in #5252.

PR #5256, opened 2021-01-20
Use gcr.io/google-appengine/openjdk:8 to avoid toomanyrequests

briandealwis:
Change `integration/testdata/debug/java/pom.xml` to use `gcr.io/google-appengine/openjdk:8` as the base image instead of `openjdk` from Docker Hub to avoid hitting Docker Hub rate limits.


briandealwis:
Committing to reduce number of spurious test fails.

PR #4926, opened 2020-10-19
Add an init phase to detect skaffold errors even before skaffold runner is created.

tejal29:
Relates to #4692

Add distinct error codes to reduce "UNKNOWN_ERROR" when skaffold was not able to create a runner. 
where phase= "UNKNOWN", and error code is "ERROR"
This happens when `skaffold.NewForConfig` fails when skaffold fails to create 
- a builder, 
- deployer, 
- init build dependencies
- init sync
- init watch trigger 

Any errors occurred here will now be part of "INIT" phase. 

This PR also identifies failure when deploy context is minikube and minikube cluster is stopped. 
On master run
1. `minikube stop`
2. `kubectl config use-context minikube`
3. `skaffold dev`

Error,
```
 skaffold dev     
creating runner: creating builder: getting docker client: getting minikube env: running [/Users/tejaldesai/Downloads/google-cloud-sdk2/bin/minikube docker-env --shell none -p minikube]
 - stdout: "* The control plane node must be running for this command\n  - To fix this, run: \"minikube start\"\n"
 - stderr: ""
 - cause: exit status 89
➜  microservices git:(minikube_errr) ✗ 
```

With this PR
```
 ../../out/skaffold dev
minikube is probably not running. Try running "minikube start".
```

Make sure the error is seen in Event API
Since this happens pretty fast, In another terminal run 
```
 watch --interval=0.1 curl localhost:50052/v1/events >> test.txt
```

in test.txt
```
{"result":{"timestamp":"2020-10-19T22:15:10.426813Z","event":{"metaEvent":{"entry":"Starting Skaffold: \u0026{Version:v1.15.0-23-gec73dafff ConfigVersi^[[10;1Hon:skaffold/v2beta9 GitVersion: GitCommit:ec73dafff8fd5714fabea148270d975de05b5449 GitTreeState:clean BuildDate:2020-10-19T14:58:30Z GoVersion:go1.14.3^[[11;2HCompiler:gc Platform:darwin/amd64}","metadata":{"build":{"numberOfArtifacts":2,"builders":[{"type":"DOCKER","count":2}],"type":"LOCAL"},"deploy":{"dep^[[12;1Hloyers":[{"type":"KUBECTL","count":1}],"cluster":"MINIKUBE"}}}}}}^M^[[13d{"result":{"timestamp":"2020-10-19T22:15:10.528741Z",

"event":{"sessionEndEvent":{
   "status":"Failed",
    "err":  {
[...]

Issue #5249, opened 2021-01-20
Flake: TestDebug/helm time out


briandealwis:
Seeing periodic failures of `TestDebug/helm`.  It looks like the actual `debug` deployment is in progress but 1min is not sufficient.  It seems surprising that the last log `Deployment` message is at 10:34 but the `kubectl get deploymens.app` is at 11:23 — why did nothing else happen in the time.

```
/integration/TestDebug/helm
time="2021-01-20T05:08:51Z" level=info msg="[skaffold build --default-repo gcr.io/k8s-skaffold]"
time="2021-01-20T05:08:51Z" level=info msg="Ran in 3 minutes 59.509 seconds"
time="2021-01-20T05:10:23Z" level=info msg="Ran in 1 minute 32.35 seconds"
time="2021-01-20T05:10:23Z" level=info msg="Namespace: skaffold6blwb"
time="2021-01-20T05:10:23Z" level=info msg="[skaffold debug --namespace skaffold6blwb --default-repo gcr.io/k8s-skaffold]"
time="2021-01-20T05:10:23Z" level=info msg="Waiting for deployments [skaffold-helm] to stabilize"
time="2021-01-20T05:10:33Z" level=info msg="Deployment skaffold-helm: Generation 0/1, Replicas 0/1, Available 0/1"
time="2021-01-20T05:10:33Z" level=info msg="Deployment skaffold-helm: Generation 0/1, Replicas 0/1, Available 0/1"
time="2021-01-20T05:10:33Z" level=info msg="Deployment skaffold-helm: Generation 1/1, Replicas 0/1, Available 0/1"
time="2021-01-20T05:10:33Z" level=info msg="Deployment skaffold-helm: Generation 1/1, Replicas 1/1, Available 0/1"
time="2021-01-20T05:10:34Z" level=info msg="Deployment skaffold-helm: Generation 1/1, Replicas 1/1, Available 0/1"
Filesystem      Size  Used Avail Use% Mounted on
overlay          68G   25G   44G  36% /
tmpfs            64M     0   64M   0% /dev
tmpfs           3.9G     0  3.9G   0% /sys/fs/cgroup
shm              64M     0   64M   0% /dev/shm
/dev/sda1        68G   25G   44G  36% /root/.cache
tmpfs           798M  8.8M  789M   2% /run/docker.sock
tmpfs           3.9G     0  3.9G   0% /proc/acpi
tmpfs           3.9G     0  3.9G   0% /proc/scsi
[...]

PR #5258, opened 2021-01-21
fix termination event not being sent

tejal29:
In #4926 I added a new event to send a termination event to detect failures in skaffold setup. 
However, during more rigorous testing, i saw some abnormalities where the event was received.
1. Send a nil event to indicate close of channel when skaffold terminates
2. Log all events.

Steps to Reproduce:
1. Please compile skaffold on this branch using `make`
2. Check if helm is installed and uninstall it or simple move.
3. if yes, 
  ```
   sudo mv /usr/local/bin/helm ~/workspace/helm1
  ```
4. Open a new terminal 
```
watch -n 0.5  "curl -H "Expect: any" localhost:50052/v1/events >> t.txt" 
```
5..  Checkout a skaffold helm example  run `skaffold dev`
```
cd examples/helm-deployment
../../out/skaffold dev
```
6. Open t.txt and verify if "TerminationEvent" is done
```
{"result":{"timestamp":"2021-01-20T23:56:59.856262Z","event":{"metaEvent":{"entry":"Starting Skaffold: \u0026{Version:v1.17.1-75-g1f519d095-dirty ConfigVersion:skaffold/v2beta11 GitVersion: GitCommit:1f519d095243f08f3b3c4db72f3c585d763b4454 GitTreeState:dirty BuildDate:2021-01-20T15:47:23Z GoVersion:go1.15.3 Compiler:gc Platform:darwin/amd64}","metadata":{"build":{"numberOfArtifacts":1,"builders":[{"type":"DOCKER","count":1}],"type":"LOCAL"},"deploy":{"deployers":[{"type":"HELM","count":1}],"cluster":"MINIKUBE"}}}}}}
{"result":{"timestamp":"2021-01-20T23:56:59.857213Z","event":{
   "terminationEvent":{"status":"Failed","err":{"errCode":"DEPLOY_HELM_VERSION_ERR","message":"creating deployer: Helm not found. Please install helm via https://helm.sh/docs/intro/install.","suggestions":[{"suggestionCode":"INSTALL_HELM","action":"Please install helm via https://helm.sh/docs/intro/install"}]}}}}}
{"result":{"timestamp":"2021-01-20T23:56:59.857219Z"}}

```

### DIFF
--- a/pkg/skaffold/errors/errors.go
+++ b/pkg/skaffold/errors/errors.go
@@ -108,7 +108,12 @@ func IsOldImageManifestProblem(err error) (string, bool) {
 }
 
 func getErrorCodeFromError(phase Phase, err error) (proto.StatusCode, []*proto.Suggestion) {
-	uErr := errors.Unwrap(err)
+	uErr := func(err error) error {
+		if e := errors.Unwrap(err); e != nil {
+			return e
+		}
+		return err
+	}(err)
 	if t, ok := uErr.(Error); ok {
 		return t.StatusCode(), t.Suggestions()
 	}
--- a/pkg/skaffold/errors/errors.go
+++ b/pkg/skaffold/errors/errors.go
@@ -108,18 +108,14 @@ func IsOldImageManifestProblem(err error) (string, bool) {
 }
 
 func getErrorCodeFromError(phase Phase, err error) (proto.StatusCode, []*proto.Suggestion) {
-	uErr := func(err error) error {
-		if e := errors.Unwrap(err); e != nil {
-			return e
-		}
-		return err
-	}(err)
-	if t, ok := uErr.(Error); ok {
-		return t.StatusCode(), t.Suggestions()
+	var sErr Error
+	if errors.As(err, &sErr) {
+		return sErr.StatusCode(), sErr.Suggestions()
 	}
+
 	if problems, ok := allErrors[phase]; ok {
 		for _, v := range problems {
-			if v.regexp.MatchString(uErr.Error()) {
+			if v.regexp.MatchString(err.Error()) {
 				return v.errCode, v.suggestion(skaffoldOpts)
 			}
 		}
--- a/pkg/skaffold/errors/errors.go
+++ b/pkg/skaffold/errors/errors.go
@@ -17,6 +17,7 @@ limitations under the License.
 package errors
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


briandealwis:
The TestDebug/helm is #5249 which should be fixed by #5262.

pkg/skaffold/errors/errors.go
```diff
@@ -107,12 +108,18 @@ func IsOldImageManifestProblem(err error) (string, bool) {
 }
 
 func getErrorCodeFromError(phase Phase, err error) (proto.StatusCode, []*proto.Suggestion) {
-	if t, ok := err.(Error); ok {
+	uErr := func(err error) error {
+		if e := errors.Unwrap(err); e != nil {
+			return e
+		}
+		return err
+	}(err)
```
briandealwis, line 116: This seems a bit complicated.  But I think this should use `errors.As()` which will try to unwrap recursively:
```
var t Error
if errors.As(err, &t) {
  return t.StatusCode(), t.Suggestions()
}
```

It seems a bit fraught to compare the regexps to the unwrapped error?

pkg/skaffold/event/event.go
```diff
@@ -739,6 +736,7 @@ func InititializationFailed(err error) {
 			},
 		},
 	})
+	handler.handle(nil)
```
briandealwis, line 739: A `nil` event seems really odd — if it indicates that the connection is being closed, why not a Closed event?  IIUC a nil event is not sent on normal termination?

Worth adding a comment as to why this is required.

pkg/skaffold/event/event.go
```diff
@@ -654,10 +654,7 @@ func (ev *eventHandler) handleExec(f firedEvent) {
 		case Failed:
 			logEntry.Entry = fmt.Sprintf("Update failed with error code %v", de.Err.ErrCode)
 		}
-	default:
-		return
 	}
-
 	ev.logEvent(*logEntry)
```
briandealwis, line 658: Aside: it doesn't seem that `logEntry` needs to be a pointer?

pkg/skaffold/event/event.go
tejal29: Can you take a look now.  I have added some logic to close the event channel there by indicating all receivers that no more events are coming.

pkg/skaffold/errors/errors.go
tejal29: Thanks. Done. I was looking for this function instead of `Unwrap`.

pkg/skaffold/event/event.go
```diff
@@ -56,8 +56,12 @@ func newHandler() *eventHandler {
...
 	go func() {
 		for {
-			ev := <-h.eventChan
-			h.handleExec(ev)
+			ev, isNotClosed := <-h.eventChan
+			if isNotClosed {
+				h.handleExec(ev)
+			} else {
+				break
+			}
```
gsquared94, line 64: nit:
```suggestion
			ev, open := <-h.eventChan
			if !open {
				break
			}
			h.handleExec(ev)
```

gsquared94 (approved):
lgtm!

PR #5267, opened 2021-01-21
Fix race condition when running tests for pkg/skaffold/instrumentation

MarlonGamez:
**Description**
Fixes a race condition that can occur when running tests for pkg/skaffold/instrumentation. To fix the issue of two threads accessing the `isOnline` variable, I've moved the responsibility of checking if the user is online to a new function and am calling that in skaffold's `main()`

### DIFF
--- a/cmd/skaffold/skaffold.go
+++ b/cmd/skaffold/skaffold.go
@@ -33,6 +33,7 @@ type ExitCoder interface {
 }
 
 func main() {
+	instrumentation.SetOnlineStatus()
 	var code int
 	if err := app.Run(os.Stdout, os.Stderr); err != nil {
 		if errors.Is(err, context.Canceled) {
--- a/pkg/skaffold/instrumentation/meter.go
+++ b/pkg/skaffold/instrumentation/meter.go
@@ -99,7 +99,9 @@ func init() {
 	doesDeploy.Insert("deploy", "dev", "debug", "run")
 }
 
-// SetOnlineStatus issues a GET request to see if the user is online
+// SetOnlineStatus issues a GET request to see if the user is online.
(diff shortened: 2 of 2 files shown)


cmd/skaffold/skaffold.go
```diff
@@ -33,6 +33,7 @@ type ExitCoder interface {
 }
 
 func main() {
+	instrumentation.SetOnlineStatus()
```
briandealwis, line 36: IMHO we should be pinging a server when the user has disable telemetry.

pkg/skaffold/instrumentation/meter.go
```diff
@@ -97,6 +97,10 @@ func init() {
 	meteredCommands.Insert("build", "delete", "deploy", "dev", "debug", "filter", "generate_pipeline", "render", "run", "test")
 	doesBuild.Insert("build", "render", "dev", "debug", "run")
 	doesDeploy.Insert("deploy", "dev", "debug", "run")
+}
+
+// SetOnlineStatus issues a GET request to see if the user is online
+func SetOnlineStatus() {
```
briandealwis, line 103: I suggest linking to https://www.chromium.org/chromium-os/chromiumos-design-docs/network-portal-detection to indicate that this is not some internal tracking server.

cmd/skaffold/skaffold.go
MarlonGamez: Good point. I'll add a conditional for this

pkg/skaffold/instrumentation/meter.go
MarlonGamez: done 👍🏼

IsaacPD (approved):
LGTM!

PR #5268, opened 2021-01-21
Configure Maven Wagon HTTP to retry on errors to successful connections

briandealwis:
It may be that our artifact requests are seeming to succeed, but the responses are failing (thus the connection resets).  This PR configured Maven's Wagon HTTP to retry even if the initial request seemed to have succeeded.

Towards #5248 🤞


MarlonGamez (approved):

briandealwis:
I’ve relaunched the Java-related builds three times with no issue.

PR #5270, opened 2021-01-21
substitute periods in sanitizeImageName()

MarlonGamez:
**Description**
When generating manifests during `skaffold init`, skaffold creates the kubernetes resource name based off of the image name is creates, which may contain periods. Since resources that have periods in their name can't applied properly, this PR modifies the `sanitizeImageName` function to replace periods.


IsaacPD (approved):
LGTM!

Issue #5266, opened 2021-01-21
`skaffold credits -d dir` writes to "dir/../skaffold-credits" and not "dir"


IsaacPD:
### Expected behavior
```bash
❯ skaffold credits -d ~/credits
2021/01/21 13:36:19 Successfully exported third party notices to /Users/isaacpd/credits
❯ ls ~/credits
4d63.com            github.com          go.opentelemetry.io gomodules.xyz       gopkg.in            knative.dev
cloud.google.com    go.opencensus.io    golang.org          google.golang.org   k8s.io              sigs.k8s.io
```
### Actual behavior
```bash
❯ skaffold credits -d ~/credits
2021/01/21 13:36:19 Successfully exported third party notices to /Users/isaacpd/credits
❯ ls ~/credits
ls: /Users/isaacpd/credits: No such file or directory
❯ ls ~/skaffold-credits
4d63.com            github.com          go.opentelemetry.io gomodules.xyz       gopkg.in            knative.dev
cloud.google.com    go.opencensus.io    golang.org          google.golang.org   k8s.io              sigs.k8s.io
```
### Information

- Skaffold version: version v1.17.2
- Operating system: macOS & Windows

### Steps to reproduce the behavior

1. skaffold credits -d ~/credits
2. ls ~/credits
3. ls ~/skaffold-credits

PR #5272, opened 2021-01-22
write credits to directory specified by -d

IsaacPD:
Fixes: #5266  

**Description**
The root statik fs file structure that is compiled in looks like
```bash
> ls .
skaffold-credits schemas secrets
```

In order to write the contents of skaffold-credits to the user specified dir, I just get rid of "skaffold-credits" from the beginning of the file path. This allows users to specify `skaffold credits -d skaffold-credits/skaffold-credits` if they want that for some reason.

**User facing changes**

`skaffold credits -d dir` now writes to dir as expected

### DIFF
--- a/cmd/skaffold/app/cmd/credits/export.go
+++ b/cmd/skaffold/app/cmd/credits/export.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


MarlonGamez (approved):
LGTM, I think we can ignore codecov since the integration test covers this flow

PR #5276, opened 2021-01-22
Improve `completion` missing-arg help text

briandealwis:
**Description**
Improves helper text for missing argument for `skaffold completion`.  This came up in the HaTS response data.

Before:
```console
$ skaffold completion
requires 1 arg, found 0
```

After:
```console
$ skaffold completion
missing shell: bash, zsh
```

(It's surprising to me that there is no helper function for this.)


IsaacPD (approved):
LGTM!

PR #5280, opened 2021-01-22
Change cloud build release script to pull metric credentials keys for release

IsaacPD:
**Description**
This changes asserts the existence of a google cloud storage bucket `gs://$PROJECT_ID-secrets` from which the necessary credentials for uploading metrics will exist and be pulled for the cloud build release context. 

This bucket has already been populated by myself, and the changes to the release process have been updated in the release doc.

**User facing changes**
If the environment variable `SKAFFOLD_EXPORT_METRICS` is set to `true` then metrics will be exported to the project specified by the credentials in `gs://$PROJECT_ID-secrets` (k8s-skaffold in our case)

### DIFF
--- a/deploy/cloudbuild-release.yaml
+++ b/deploy/cloudbuild-release.yaml
@@ -1,6 +1,15 @@
 # using default substitutions, provided by Google Cloud Build
 # see: https://cloud.google.com/container-builder/docs/configuring-builds/substitute-variable-values#using_default_substitutions
 steps:
+# Grab secret credentials from gcp bucket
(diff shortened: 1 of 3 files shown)


tejal29:
> If the environment variable `SKAFFOLD_EXPORT_METRICS` is set to `true` then metrics will be exported to the project specified by the credentials in `gs://$PROJECT_ID-secrets` (k8s-skaffold in our case)

This is not true any more right? If `collect-metrics` is set then the metrics will be uploaded. 
Can you change that?

tejal29 (approved):

Issue #5042, opened 2020-11-19
Publish macOS/arm64 binaries


tstromberg:
NOTE: No released version of Go supports macOS/arm64 as of this date, but v1.16 will. https://github.com/golang/go/issues/38485 has some details on how to get started with making builds from the go trunk source code.

We may have to do something about code-signing on macOS/arm64

briandealwis:
Found a Go program, [`makefat`](https://github.com/randall77/makefat), for producing multi-arch ("fat") binaries for macOS/Darwin.

briandealwis:
Support for darwin/arm64 requires using Go 1.16beta1. We currently build with _cgo_ so that our file-watcher can take advantage of [_rjeczalik/notify_](https://github.com/rjeczalik/notify)'s native capabilities (FSEvents, inotify, etc). We use a Docker-provided cross-building image (https://hub.docker.com/r/dockercore/golang-cross), which includes SDKs (toolchains, headers, and libraries) for Linux, Windows, and macOS.  The included macOS SDKs are for 10.x however, and don't support building for darwin/arm64 which requires an SDK for macOS 11.x.  The legality of including these SDKs is questionable, however.

#5286 is a stop-gap solution that disables using cgo for darwin/arm64 and uses _rjeczalik/notify_'s kqueue support instead.  The kqueue support has worse performance.  But this at least works.

The better (legal) solution requires that we build on macOS and use the right SDKs.  But we should also be signing our builds (#2154).  I think the right solution is to move our builds into Kokoro.  This is a more complex task.

briandealwis:
Error seen on start:
```
$ skaffold dev
ERRO[0000] failure getting variant                       error="getCPUInfo for OS darwin: not implemented"
...
```

This seems to have no effect.

PR #5286, opened 2021-01-25
Update release to build darwin/arm64 binary

briandealwis:
Towards #5042 

We currently build with cgo to enable the fast [notify](https://github.com/rjeczalik/notify) trigger.  But [`dockercore/golang-cross`](https://hub.docker.com/r/dockercore/golang-cross) has too old a toolchain to enable cgo for darwin/arm64.  So for darwin/arm64 we disable cgo and use notify's kqueue support instead.

This is a stop-gap measure since we need to produce signed binaries anyways (#2154).

This PR:
- Tweaks the Makefile variables to allow specifying the Go version, build tags, and CGO_ENABLED on a per os/arch basis.  Reorganize the variables on a per os/arch basis.
- Updates github.com/rjeczalik/notify to track master to avoid explicitly specifying `kqueue` (https://github.com/rjeczalik/notify/pull/182) and also to pull in bug fixes that are not yet in a release.
- Uses github.com/randall77/makefat to build a multi-arch ("fat") binary for Darwin for those that want it.


Makefile
```diff
@@ -42,31 +42,31 @@ ifeq "$(strip $(VERSION))" ""
...
+GO_BUILD_TAGS_darwin = release
 
 ifneq "$(strip $(LOCAL))" "true"
 	override STATIK_FILES =  cmd/skaffold/app/cmd/statik/statik.go
 endif
 
 # when build for local development (`LOCAL=true make install` can skip license check)
 $(BUILD_DIR)/$(PROJECT): $(STATIK_FILES) $(GO_FILES) $(BUILD_DIR)
-	GOOS=$(GOOS) GOARCH=$(GOARCH) CGO_ENABLED=1 go build -gcflags="all=-N -l" -tags $(GO_BUILD_TAGS_$(GOOS)) -ldflags $(GO_LDFLAGS_$(GOOS)) -o $@ $(BUILD_PACKAGE)
+	GOOS=$(GOOS) GOARCH=$(GOARCH) CGO_ENABLED=1 go build -gcflags="all=-N -l" -tags "$(GO_BUILD_TAGS_$(GOOS))" -ldflags "$(GO_LDFLAGS_$(GOOS))" -o $@ $(BUILD_PACKAGE)
```
IsaacPD, line 69: ```GO_LDFLAGS_$(GOOS)``` doesn't seem to be defined for windows, darwin, or linux after these changes, are they supposed to be defined somewhere here?

briandealwis: Hmm you're right — I've missed something here.

briandealwis:
Converting this to a draft as I'm not entirely sure if it's worth undertaking a stop-gap measure vs just doing the right thing in the first place.

briandealwis:
Converting this from draft: although using a CGO-less build is a stop-gap measure, it gets a native binary out.  A perfect solution can be brought in with #2154.

IsaacPD (approved):
I was able to build the darwin/arm64 binary, don't have a machine to test the actual binary on though, LGTM!

PR #5290, opened 2021-01-25
Fixes for generating new schema version

MarlonGamez:
Fixes: #5289
**Related**: https://github.com/GoogleContainerTools/skaffold/pull/5279

**Description**
Adds a new check that can be used in the walk package to see if a filename contains given substrings. `WhenNameContains` will be used instead of the current check for `WhenHasName`, as there have been example config files added recently that don't have the name `skaffold.yaml`. We will now check that the filename contains `"skaffold"` and `"yaml"`.

I've also updated tests in `cmd/skaffold/app/cmd/parse_config_test.go` as they were checking for a fixed version and would break upon creating a new version of the config

### DIFF
--- a/cmd/skaffold/app/cmd/parse_config_test.go
+++ b/cmd/skaffold/app/cmd/parse_config_test.go
@@ -55,7 +55,7 @@ profiles:
 
 func createCfg(name string, imageName string, workspace string, requires []latest.ConfigDependency) *latest.SkaffoldConfig {
 	return &latest.SkaffoldConfig{
-		APIVersion:   "skaffold/v2beta11",
+		APIVersion:   latest.Version,
 		Kind:         "Config",
 		Dependencies: requires,
 		Metadata:     latest.Metadata{Name: name},
--- a/hack/versions/cmd/new/version.go
+++ b/hack/versions/cmd/new/version.go
@@ -76,13 +76,13 @@ func main() {
 	hackschema.UpdateVersionComment(path("latest", "config.go"), false)
 
 	// Update skaffold.yaml in integration tests
-	walk.From("integration").WhenNameContains("skaffold").WhenHasExt("yaml").MustDo(func(path string, _ walk.Dirent) error {
+	walk.From("integration").WhenNameContains("skaffold").WhenHasExt(".yaml").MustDo(func(path string, _ walk.Dirent) error {
 		sed(path, current, next)
 		return nil
 	})
 
 	// Update skaffold.yaml in init tests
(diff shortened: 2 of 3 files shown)


cmd/skaffold/app/cmd/parse_config_test.go
```diff
@@ -55,7 +55,7 @@ profiles:
 
 func createCfg(name string, imageName string, workspace string, requires []latest.ConfigDependency) *latest.SkaffoldConfig {
 	return &latest.SkaffoldConfig{
-		APIVersion:   "skaffold/v2beta11",
+		APIVersion:   latest.Version,
```
gsquared94, line 58: my bad, thanks for fixing this

hack/versions/cmd/new/version.go
```diff
@@ -76,13 +76,13 @@ func main() {
 	hackschema.UpdateVersionComment(path("latest", "config.go"), false)
 
 	// Update skaffold.yaml in integration tests
-	walk.From("integration").WhenHasName("skaffold.yaml").MustDo(func(path string, _ walk.Dirent) error {
+	walk.From("integration").WhenNameContains("skaffold", "yaml").MustDo(func(path string, _ walk.Dirent) error {
```
gsquared94, line 79: just for the sake of argument, if we added a `generate_skaffold_yaml.sh` someday it'd match this filter. I wonder if we should add a `IsFileType(extension string)` filter and do 
```go
walk.From("integration").IsFileType("yaml").WhenNameContains("skaffold").MustDo(...)
```

MarlonGamez: good point, I can add this as well

briandealwis: Would it make sense to have `WhenNameMatches(glob)` and use [`path.Match()`](https://pkg.go.dev/path/#Match)?

MarlonGamez: @briandealwis I think that'd make sense too. It would definitely shorten things a bit as well. Do you think we should go with that? I just pushed changes if you want to take a look and let me know

briandealwis: I would just have keep it as simple as possible: since `WhenNameContains` and `IsFileType()` can be implemented with `WhenNameMatches()`, I would just keep `WhenNameMatches()`.

(My quibble with `IsFileType()` is that it could imply doing content-matching.)

MarlonGamez: Makes sense :) I'll make changes in favor of `WhenNameMatches()`

briandealwis (approved):

PR #5291, opened 2021-01-25
Add missing fields to global config doc

briandealwis:
Add doc for `debug-helpers-registry`, `k3d-disable-load`, `kind-disable-load`.


MarlonGamez (approved):
lgtm 👍

Issue #5292, opened 2021-01-25
site: enable offline search remove ads in results


medyagh:
Skaffold Search shows google ads in the results

![Screen Shot 2021-01-25 at 1 42 42 PM](https://user-images.githubusercontent.com/4564227/105769912-6052f980-5f13-11eb-98d3-4c52b2cdd41c.png)

PR #5294, opened 2021-01-25
enable local search for skaffold remove ads

medyagh:
removes ads from search result and make it faster

https://github.com/GoogleContainerTools/skaffold/issues/5292


IsaacPD (approved):
LGTM!

image

PR #5304, opened 2021-01-26
Minor doc updates to Telemetry

sanguinecloud:
Neaten up wording & formatting for Telemetry doc.


IsaacPD:
The json that is used to populate this page is generated from the struct here https://github.com/GoogleContainerTools/skaffold/blob/8848703c9da3b977e091d49465c7bbc29e64d662/pkg/skaffold/instrumentation/meter.go#L53. 
Are you able to update the comments there as well? The json is generated by doing `make generate-telemetry-json` if you want to verify.

Issue #5300, opened 2021-01-26
Go-based examples break with debug


briandealwis:
`container-debug-support` updated Delve to 1.5.1 which raised the [minimum Go version supported](https://github.com/go-delve/delve/pull/2133) to Go 1.13 (there was nothing noted in the release notes).  As a result, debugging programs using Go 1.12 now break.

### Expected behavior

`cd examples/getting-started && skaffold debug` should work

### Actual behavior

```
$ cd examples/getting-started && skaffold debug
...
Not watching for changes...
[install-go-debug-support] Installing runtime debugging support files in /dbg
[install-go-debug-support] Installation complete
[getting-started] API server listening at: [::]:56268
[getting-started] Version of Go is too old for this version of Delve (minimum supported version 1.13, suppress this error with --check-go-version=false)
[getting-started] API server listening at: [::]:56268
[getting-started] Version of Go is too old for this version of Delve (minimum supported version 1.13, suppress this error with --check-go-version=false) 
```

briandealwis:
TIL: the `golang:N-alpine` images don't include a C compiler (unlike `golang:N`), and using `go build -gcflags=all="-N -l"` calls out to the C compiler.  Thus our debug-related tests fail if they use the `-alpine` images.

PR #5307, opened 2021-01-27
Update language runtime image versions

briandealwis:
Fixes: #5300 

**Description**
This PR bumps the language runtime image versions for Go, Python, and NodeJS to latest supported versions.  It also changes `alpine:` image references to `alpine:3` — since we're not pinning to particular digests, `3` vs `3.13` doesn't much matter.

This change was motivated by #5300: Delve 1.5.1 silently bumped the minimum Go version supported to 1.13, which means `skaffold debug` breaks on our examples which were being compiled with Go 1.12.  It also turns out that `golang:N-alpine` images cannot be used for `skaffold debug`-able images as they do not include a C compiler which is required for `go build -gcflags=all="-N -l"` to disable optimizations and inlining.

Note: this doesn't change our internal tools (e.g., under `deploy/`).  My first attempt update those and things broke, so I'll do that separately.


tejal29:
is it going to a big ask to replace the tag `golang:1.15` with sha to make examples more secure like @medyagh pointed out ?

If not, we could create another issue.

briandealwis:
I think we should get more opinions on pinning image digests before we roll it out across our projects.

gsquared94 (approved):

PR #5308, opened 2021-01-27
Make integration.WaitForPodsReady use pod `Ready` condition

briandealwis:
Towards #5249
Merge after: #5307 because `debug` tests will fail due to dlv/go1.12 breakage #5300

**Description**
`WaitForPodsReady` uses `WaitForPodsInPhase` to check that the pods' phases are `Running`.  But this does not account for readiness probes.  It also turns out that our `waitForDebugEvents()` calls `WaitForPodsReady()` with no pod names, expecting it to wait for all pods to be ready.

This PR splits out `WaitForPodsReady()` into a utility `waitForPods()` that takes a predicate and waits for pods to satisfy that predicate.  `WaitForPodsReady()` calls into `waitForPods()` with a predicate that waits for the pod's `Ready` condition to become true.  `WaitForPodsInPhase()` calls into `waitForPods()` with predicate that checks the pod's phase.  `waitForPods()` takes a list of pod names to monitor, and if empty then it waits for all pods (in that namespace).


briandealwis:
Tests are failing as expected since #5307 isn't committed \o/

integration/util.go
```diff
@@ -200,30 +219,36 @@ func (k *NSKubernetesClient) WaitForPodsInPhase(expectedPhase v1.PodPhase, podNa
...
 			}
-			phases[pod.Name] = pod.Status.Phase
 
+			if _, found := podsReady[pod.Name]; !found && waitForAllPods {
+				podNames = append(podNames, pod.Name)
+			}
+			podsReady[pod.Name] = podReady(pod)
+
+			var waiting []string
 			for _, podName := range podNames {
```
IsaacPD, line 243: would this cause any issues when waitForAllPods is true? I'm just thinking that maybe a pod goes through all the necessary phases before any other and so it ends up being the only one in `podNames` resulting in `waiting` being empty despite there potentially being more pods who haven't reached a phase yet.

briandealwis: That is true: there is a race condition where pod creation events are pending in the watcher.  I guess I could check `len(w.ResultChan())` to see if there are other events pending.

briandealwis: I hope these pods will eventually reach a fixpoint :-/  I suppose we do have a timeout.  I print out how many pending messages there are.

briandealwis:
(note: tests should fail as still haven't merged with HEAD to pick up #5307)

IsaacPD (approved):

PR #5313, opened 2021-01-27
print otel errors with logrus instead of straight to stderr

IsaacPD:
**Description**
Currently if an error is thrown from open telemetry while uploading metrics it is just spit onto stderr.

By setting `global.SetErrorHandler` we can instead absorb those errors and print them into the debug logs so they are not shown unconditionally.

**User facing changes**
Errors from open telemetry are only shown when using `-v debug`


MarlonGamez (approved):
lgtm 👍

PR #5322, opened 2021-01-30
export metrics related to user enum flags

### LINKED ISSUE #5195: Determine which specific flags used to upload to metrics
This issue is to determine which flags (e.g. "--port-forward") we want to keep track of in our metrics when keeping track of errors and commands used.

There are 4 different error metrics to consider:
- "errors": the least specific error metric that tracks all errors
- "errors/unknown": For unknown skaffold errors
- "deploy/unknown": For unknown deploy errors
- "build/unknown": For unknown build errors

And the command metrics ("build", "delete", "deploy", "dev", "debug", "filter", "generate_pipeline", "render", "run", "test")

For each of these we want to consider which flags will give us the most information for questions we want to answer.

e.g. What launch scenarios result in the most "deploy/unknown" errors and how could they be related? What flags are the most/least used for a given command?

All this has to be with the consideration that there are only a max of 10 labels each for a given metric. But there could be some workaround with that by allowing multiple metrics of the same type (e.g. "dev-2", "build/unknown-2") since there is a significantly higher limit on custom metric descriptors in Cloud Monitoring.

https://cloud.google.com/monitoring/custom-metrics/creating-metrics#md-modify

**Adding Labels**
Adding labels is very easy so long as there are less than 10 labels for the given metrics already.
1. Just add to the code the label that should be tracked (e.g. `metric.Record(ctx, count, oldLabels, newLabel)`)
If there are already 10 labels then it is not exactly possible and so a new metrics must be created

**Removing Labels**
Not possible. Once associated with a metric it can not be disassociated even if it is not used. Similarly to over adding labels, a new metric must be created that would replace the previous one.

IsaacPD:
Fixes: #5195  

**Description**
This PR takes user enum flags run on commands that export metrics and exports them to GCP, prior they were just stored in the skaffoldMeter struct, but now they are exported from there.

To resolve #5195 , all appropriate flags will have their own metric descriptor and will be labeled with the command used, any error, and the value of the flag that was used.

### DIFF
--- a/pkg/skaffold/instrumentation/export.go
+++ b/pkg/skaffold/instrumentation/export.go
@@ -43,6 +43,10 @@ import (
 	"github.com/GoogleContainerTools/skaffold/proto"
 )
 
+const (
+	FlagsPrefix = "flags/"
+)
+
 func ExportMetrics(exitCode int) error {
 	if !shouldExportMetrics || meter.Command == "" {
 		return nil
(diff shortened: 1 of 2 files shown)


MarlonGamez (commented):
This LGTM 👍🏼 I guess let's just wait on integration? I'll approve and merge if we can get it green

MarlonGamez (approved):

PR #5323, opened 2021-01-30
Programatically create a dashboard for flag metric descriptors

IsaacPD:
**Related**: #5195 
**Merge after**: #5354

**Description**
Add make target `fags-dashboard` that when run creates/updates a dashboard named "Skaffold Flags" in Cloud Monitoring with graphs of all the enum flags that have metrics. The command does the following
1. creates the dashboard if it doesn't exist, otherwise it updates the existing one
2. creates a graph for each command based on the "skaffold/flags" metric descriptor that is filtered by the respective command, and grouped by the flag_value, and flag_name


MarlonGamez:
Will we basically want to run this each time we add new flags to skaffold?

IsaacPD:
> Will we basically want to run this each time we add new flags to skaffold?

Yeah that's the idea, alternatively the dashboard can be modified manually but its a bit more arduous

IsaacPD:
Drafting as a better implementation for flag metrics is under development that would not necessitate this change

tejal29 (approved):

PR #5324, opened 2021-01-30
refactor instrumentation package

IsaacPD:
**Description**

`meter.go` was getting very large, I mean just look at all those imports. Re-factoring sooner rather than later.


MarlonGamez:
Nice :D this looks good, I think there are just a couple linter issues that need to be fixed

gsquared94 (approved):

Issue #5342, opened 2021-02-03
Flake: docker hub requiring authentication


briandealwis:
Suddenly having a rash of failures on Travis trying to pull images from Docker Hub.  Assuming they're transient.

```
--- FAIL: TestRun (679.14s)
/integration/TestRun/buildpacks_Go
time="2021-02-03T17:49:24Z" level=info msg="Namespace: skaffoldk44vr"
time="2021-02-03T17:49:24Z" level=info msg="Running [skaffold run --namespace skaffoldk44vr --default-repo gcr.io/k8s-skaffold] in examples/buildpacks"
    helper.go:229: skaffold run: exit status 1, Generating tags...
         - skaffold-buildpacks -> gcr.io/k8s-skaffold/skaffold-buildpacks:c2040d5
        Checking cache...
         - skaffold-buildpacks: Not found. Building
        Found [k3d-k3s-default] context, using local docker daemon.
[...]
        v1: Pulling from buildpacks/gcp/run
        9f2825a1003a: Already exists
        7e79098fe9b4: Already exists
        3c2cba919283: Already exists
        261de18b35a6: Already exists
        c26187d600b1: Already exists
        Digest: sha256:578631b17d0db25f29d6157d0390e84e657446f908a26dc1fe70ee161eb8f7b5
        Status: Downloaded newer image for gcr.io/buildpacks/gcp/run:v1
        0.9.3: Pulling from buildpacksio/lifecycle
    panic.go:617: failed to build: fetching lifecycle image: unauthorized: authentication required
    --- FAIL: TestRun/buildpacks_Go (26.61s)
/integration/TestRun/kustomize
time="2021-02-03T17:52:58Z" level=info msg="Namespace: skaffoldnrdn7"
time="2021-02-03T17:52:58Z" level=info msg="Running [skaffold run --namespace skaffoldnrdn7 --default-repo gcr.io/k8s-skaffold] in examples/getting-started-kustomize"
    helper.go:229: skaffold run: exit status 1, Generating tags...
         - skaffold-kustomize -> gcr.io/k8s-skaffold/skaffold-kustomize:c2040d5
        Checking cache...
         - skaffold-kustomize: Error checking cache.
[...]

PR #5344, opened 2021-02-03
Configure k3d to use registry-mirrors

briandealwis:
Configure k3d to use the registry-mirrors configured in the Docker daemon as per `kind` in #5237.  This should further reduce Docker Hub problems such as #5342.

### DIFF
--- a/hack/generate-k3d-registries.sh
+++ b/hack/generate-k3d-registries.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+# Copyright 2019 The Skaffold Authors
+#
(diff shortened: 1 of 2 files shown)


hack/generate-k3d-registries.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+# Copyright 2019 The Skaffold Authors
```
gsquared94, line 3:
```suggestion
# Copyright 2020 The Skaffold Authors
```

Issue #5352, opened 2021-02-04
Export the platform type in skaffold metrics


IsaacPD:
### Expected behavior
Can categorize user launches by the platform type (gcb, local, cluster)
### Actual behavior
Data is not visible in Cloud Monitoring

PR #5353, opened 2021-02-04
track the platform type at the launch level

IsaacPD:
Fixes: #5352  

**Description**
Add label to launch metric that describes the platform type that is being used. With this there are only 3 more labels that can be added to the launches metric.


briandealwis (approved):
LGTM

Issue #5341, opened 2021-02-03
skaffold deploy does not honour --default-repo flag


dan-j:
### Expected behavior

I've upgraded to v1.19, I'm not 100% which version from and running into the following issue:

```
skaffold deploy -p profile --default-repo eu.gcr.io/my-registry --images my-image:v1.0.0
```

I expect the image to be deployed be `eu.gcr.io/my-registry/my-image:v1.0.0`

### Actual behavior

The image deployed is `my-image:v1.0.0`

### Information

I've verified this working as expected on v1.17.0, but not sure which exact release it was last functioning.

- Skaffold version: v1.19.0
- Operating system: tested on both MacOS and Ubuntu

### Steps to reproduce the behavior

Will attach a link to a repo asap, sorry!

dan-j:
I've never contributed to this project before but love it. If somebody can point me in the right direction I don't mind looking at fixing it

gsquared94:
Thanks for opening this issue. This seems to be a case where our documentation doesn't match the behavior so I'll take that as a bug. The `--images` flag is setting [this](https://github.com/GoogleContainerTools/skaffold/blob/d956d4465c2eacf6a75e85f964363d1533ecff30/cmd/skaffold/app/cmd/deploy.go#L33) variable. If you navigate to its usage it's being merged with the list of skaffold built artifacts [here](https://github.com/GoogleContainerTools/skaffold/blob/d956d4465c2eacf6a75e85f964363d1533ecff30/cmd/skaffold/app/cmd/util.go#L29). You can apply the transformation [`ApplyDefaultRepo`](https://github.com/GoogleContainerTools/skaffold/blob/d956d4465c2eacf6a75e85f964363d1533ecff30/pkg/skaffold/deploy/util/util.go#L27) here prior to this merge to fix this issue.

gsquared94:
I misdiagnosed this issue. Adding `needs-reproduction` back

dan-j:
https://github.com/dan-j/skaffold-issue-5341

PR #5397, opened 2021-02-13
Fix issue where default-repo wasn't being added to artifact tags (#5341)

dan-j:
Fixes: #5341, #5463

**Description**
Using the `--default-repo` option on `deploy` commands wasn't being honoured and the default tag was being deployed.

It's a simple bug... The for-loop in `getBuildArtifactsAndSetTags` is correctly prefixing the tag with the --default-repo option but was setting it on the for-loop variable (which is a pointer re-used in each loop), and not updating the actual `buildArtifacts` slice.

I haven't added any tests, this is my first contribution and not sure how to get a `runner.Runner` in a test environment. If you want tests adding for this change could somebody give me a point in the right direction 🙂

### DIFF
--- a/cmd/skaffold/app/cmd/util.go
+++ b/cmd/skaffold/app/cmd/util.go
@@ -21,25 +21,32 @@ import (
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/tag"
-	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 )
 
-func getBuildArtifactsAndSetTags(r runner.Runner, artifacts []*latest.Artifact) ([]build.Artifact, error) {
+// DefaultRepoFn takes an image tag and returns either a new tag with the default repo prefixed, or the original tag if
+// no default repo is specified.
+type DefaultRepoFn func(string) (string, error)
+
+func getBuildArtifactsAndSetTags(artifacts []*latest.Artifact, defaulterFn DefaultRepoFn) ([]build.Artifact, error) {
 	buildArtifacts, err := mergeBuildArtifacts(fromBuildOutputFile.BuildArtifacts(), preBuiltImages.Artifacts(), artifacts)
 	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/skaffold/app/cmd/util.go
```diff
@@ -31,12 +31,12 @@ func getBuildArtifactsAndSetTags(r runner.Runner, artifacts []*latest.Artifact)
 		return nil, err
 	}
 
-	for _, artifact := range buildArtifacts {
-		tag, err := r.ApplyDefaultRepo(artifact.Tag)
+	for i := range buildArtifacts {
+		updatedTag, err := r.ApplyDefaultRepo(buildArtifacts[i].Tag)
```
dan-j, line 35: renamed the variable because it was colliding with the `tag` package.

gsquared94:
Thanks for your contribution @dan-j 
> I haven't added any tests, this is my first contribution and not sure how to get a `runner.Runner` in a test environment. If you want tests adding for this change could somebody give me a point in the right direction 🙂

You can modify the function signature to:
```go
func getBuildArtifactsAndSetTags(artifacts []*latest.Artifact, applyDefaultRepo func(string)(string, error)) ([]build.Artifact, error) {
  ...
```
and all non-test references look like:
```go
buildArtifacts, err := getBuildArtifactsAndSetTags(artifacts, r.ApplyDefaultRepo)
```

That removes the `pkg/skaffold/runner` dependency from this file.

dan-j:
Oops, I had forgotten about this and then just ran into the problem this solves 😬

I got halfway through the tests but then sidetracked, will finish this off this week at some point

MarlonGamez:
hey @dan-j, have you had any time to work on this? Not meaning to pressure, just want to make sure that you have bandwith to finish this :) If not, I'm sure someone on the team could pick it up

dan-j:
Hey! Yeah I got completely sidetracked by other stuff and forgot, until I've just hit the exact same issue again when running a manual deployment 😅

I was already halfway through the tests, will try finish them tonight, if not it will be this week.

dan-j:
OK, I've added some tests.

The change @gsquared94 suggested wasn't completely straight forward because that function also uses the global flags for merging build artifacts. I've refactored the problematic code into a new function and have added tests for this new function.

MarlonGamez (approved):
looks good to me! Thanks @dan-j for this :)

PR #5402, opened 2021-02-16
run go mod tidy

MarlonGamez:
**Description**
Linter was warning about `go.mod` being out of sync with `go.sum`. This PR just runs `go mod tidy` and commits the results.


nkubala (approved):
can't believe this PR casually bumps `k8s.io/` deps and doesn't completely break the build 😆

PR #5405, opened 2021-02-17
Fixes #5404: Skaffold configs downloaded from a url can define remote config dependencies

### LINKED ISSUE #5404: Providing a url to `-f` flag doesn't work with remote configs
### Expected behavior
Run:
```bash
skaffold run -f https://raw.githubusercontent.com/gsquared94/skaffold-remote-configs-demo/main/skaffold.yaml
```
Skaffold uses the url provided as the main config and downloads any remote configs specified in that file.
### Actual behavior
Skaffold fails with the error:
```
parsing skaffold config: could not find skaffold config /Users/gaghosh/https:/raw.githubusercontent.com/gsquared94/skaffold-remote-configs-demo/main/skaffold.yaml that is referenced as a dependency in config /Users/gaghosh/https:/raw.githubusercontent.com/gsquared94/skaffold-remote-configs-demo/main/skaffold.yaml
```
### Information

- Skaffold version: v1.20.0
- Operating system: ...
- Contents of skaffold.yaml:
https://raw.githubusercontent.com/gsquared94/skaffold-remote-configs-demo/main/skaffold.yaml

gsquared94:
Fixes #5404 

This PR makes two changes:
- Fix the config parsing logic that was erroneously checking that a dependency config file exists on disk even if it's a URL. 
- Add missing check that multiple configs in the same file cannot have the same name.

### DIFF
--- a/cmd/skaffold/app/cmd/parse_config.go
+++ b/cmd/skaffold/app/cmd/parse_config.go
@@ -219,6 +219,9 @@ func processEachDependency(d latest.ConfigDependency, cfgOpts configOpts, opts c
 			path = filepath.Join(path, "skaffold.yaml")
 		}
 	}
+
+	// if the current and previous configuration files are the same, then current config should be treated as a dependency config if the previous config was also a dependency config.
+	// Otherwise the current config is always a dependency config if the file path is different than the previous.
 	cfgOpts.isDependency = cfgOpts.isDependency || path != cfgOpts.file
 	cfgOpts.file = path
 	cfgOpts.selection = d.Names
--- a/cmd/skaffold/app/cmd/parse_config.go
+++ b/cmd/skaffold/app/cmd/parse_config.go
@@ -92,6 +92,19 @@ func getConfigs(cfgOpts configOpts, opts config.SkaffoldOptions, r *record) ([]*
 	}
 	logrus.Debugf("parsed %d configs from configuration file %s", len(parsed), cfgOpts.file)
 
+	// validate that config names are unique if specified
+	seen := make(map[string]bool)
+	for _, cfg := range parsed {
+		cfgName := cfg.(*latest.SkaffoldConfig).Metadata.Name
+		if cfgName == "" {
+			continue
+		}
+		if seen[cfgName] {
+			return nil, fmt.Errorf("multiple skaffold configs named %q found in file %q", cfgName, cfgOpts.file)
+		}
+		seen[cfgName] = true
+	}
+
 	var configs []*latest.SkaffoldConfig
 	for i, cfg := range parsed {
 		config := cfg.(*latest.SkaffoldConfig)
@@ -104,22 +117,16 @@ func getConfigs(cfgOpts configOpts, opts config.SkaffoldOptions, r *record) ([]*
 	return configs, nil
 }
 
-// processEachConfig processes each parsed config by applying profiles and recursively processing it's dependencies
+// processEachConfig processes each parsed config by applying profiles and recursively processing its dependencies.
+// The `index` parameter specifies the index of the current config in its `skaffold.yaml` file. We use the `index` instead of the config `metadata.name` property to uniquely identify each config since not all configs define `name`.
 func processEachConfig(config *latest.SkaffoldConfig, cfgOpts configOpts, opts config.SkaffoldOptions, r *record, index int) ([]*latest.SkaffoldConfig, error) {
 	// check that the same config name isn't repeated in multiple files.
 	if config.Metadata.Name != "" {
-		prev, found := r.configNameToFile[config.Metadata.Name]
-		if found {
-			sl := strings.SplitN(prev, "@", 2) // map value is always formatted as `file_name@config_index`
-			prevConfig, prevIndex := sl[0], sl[1]
-			if prevConfig != cfgOpts.file {
-				return nil, fmt.Errorf("skaffold config named %q found in multiple files: %q and %q", config.Metadata.Name, prevConfig, cfgOpts.file)
-			}
-			if prevIndex != fmt.Sprint(index) {
-				return nil, fmt.Errorf("multiple skaffold configs named %q found in file %q", config.Metadata.Name, cfgOpts.file)
-			}
+		prevConfig, found := r.configNameToFile[config.Metadata.Name]
+		if found && prevConfig != cfgOpts.file {
+			return nil, fmt.Errorf("skaffold config named %q found in multiple files: %q and %q", config.Metadata.Name, prevConfig, cfgOpts.file)
 		}
-		r.configNameToFile[config.Metadata.Name] = fmt.Sprintf("%s@%d", cfgOpts.file, index)
+		r.configNameToFile[config.Metadata.Name] = cfgOpts.file
 	}
 
 	// configSelection specifies the exact required configs in this file. Empty configSelection means that all configs are required.
--- a/cmd/skaffold/app/cmd/parse_config.go
+++ b/cmd/skaffold/app/cmd/parse_config.go
@@ -210,7 +210,7 @@ func processEachDependency(d latest.ConfigDependency, cfgOpts configOpts, opts c
 	if !util.IsURL(path) {
 		fi, err := os.Stat(path)
 		if err != nil {
-			if os.IsNotExist(errors.Unwrap(err)) {
+			if errors.Is(err, os.ErrNotExist) {
 				return nil, fmt.Errorf("could not find skaffold config %s that is referenced as a dependency in config %s", path, cfgOpts.file)
 			}
 			return nil, fmt.Errorf("parsing dependencies for skaffold config %s: %w", cfgOpts.file, err)
--- a/cmd/skaffold/app/cmd/parse_config.go
+++ b/cmd/skaffold/app/cmd/parse_config.go
@@ -81,7 +81,7 @@ func getConfigs(cfgOpts configOpts, opts config.SkaffoldOptions, r *record) ([]*
 		return nil, err
 	}
 
-	if !filepath.IsAbs(cfgOpts.file) {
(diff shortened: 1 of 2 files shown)


cmd/skaffold/app/cmd/parse_config.go
```diff
@@ -108,11 +108,18 @@ func getConfigs(cfgOpts configOpts, opts config.SkaffoldOptions, r *record) ([]*
 func processEachConfig(config *latest.SkaffoldConfig, cfgOpts configOpts, opts config.SkaffoldOptions, r *record, index int) ([]*latest.SkaffoldConfig, error) {
 	// check that the same config name isn't repeated in multiple files.
 	if config.Metadata.Name != "" {
-		prevConfig, found := r.configNameToFile[config.Metadata.Name]
-		if found && prevConfig != cfgOpts.file {
-			return nil, fmt.Errorf("skaffold config named %q found in multiple files: %q and %q", config.Metadata.Name, prevConfig, cfgOpts.file)
+		prev, found := r.configNameToFile[config.Metadata.Name]
+		if found {
+			sl := strings.SplitN(prev, "@", 2) // map value is always formatted as `file_name@config_index`
```
briandealwis, line 113: This overloading of `configNameToFile` seems error-prone (`@` is a valid character in file names).  It also seems unnecessary since it's entirely our code: we could just have the map take `string` &rarr; `struct { name, index }`.  But more to the point, we don't seem to use this index anywhere else, so couldn't this search-for-duplicates just use a local variable to maintain its state?

And maybe this search-for-duplicates should be done separately?

cmd/skaffold/app/cmd/parse_config.go
```diff
@@ -108,11 +108,18 @@ func getConfigs(cfgOpts configOpts, opts config.SkaffoldOptions, r *record) ([]*
 func processEachConfig(config *latest.SkaffoldConfig, cfgOpts configOpts, opts config.SkaffoldOptions, r *record, index int) ([]*latest.SkaffoldConfig, error) {
```
briandealwis, line 108: Would be helpful to describe what `index` represents here.  (And `it's` &rarr; `its`)

cmd/skaffold/app/cmd/parse_config.go
```diff
@@ -108,11 +108,18 @@ func getConfigs(cfgOpts configOpts, opts config.SkaffoldOptions, r *record) ([]*
...
+		prev, found := r.configNameToFile[config.Metadata.Name]
+		if found {
+			sl := strings.SplitN(prev, "@", 2) // map value is always formatted as `file_name@config_index`
+			prevConfig, prevIndex := sl[0], sl[1]
+			if prevConfig != cfgOpts.file {
+				return nil, fmt.Errorf("skaffold config named %q found in multiple files: %q and %q", config.Metadata.Name, prevConfig, cfgOpts.file)
+			}
+			if prevIndex != fmt.Sprint(index) {
+				return nil, fmt.Errorf("multiple skaffold configs named %q found in file %q", config.Metadata.Name, cfgOpts.file)
+			}
```
briandealwis, line 120: Can duplicate references not be handled by `ParseConfigAndUpgrade()`?  So all we'd need is a set of seen config names?

cmd/skaffold/app/cmd/parse_config.go
```diff
@@ -193,15 +200,17 @@ func processEachDependency(d latest.ConfigDependency, cfgOpts configOpts, opts c
...
 		path = cfgOpts.file
 	}
-	fi, err := os.Stat(path)
-	if err != nil {
-		if os.IsNotExist(errors.Unwrap(err)) {
-			return nil, fmt.Errorf("could not find skaffold config %s that is referenced as a dependency in config %s", path, cfgOpts.file)
+	if !util.IsURL(path) {
+		fi, err := os.Stat(path)
+		if err != nil {
+			if os.IsNotExist(errors.Unwrap(err)) {
```
briandealwis, line 206: The unwrap isn't correct: it's possible the `ErrNoExist` is actually a wrapped error.  See the example at
https://golang.org/pkg/os/#IsNotExist

cmd/skaffold/app/cmd/parse_config.go
```diff
@@ -193,15 +200,17 @@ func processEachDependency(d latest.ConfigDependency, cfgOpts configOpts, opts c
...
+		}
+		if fi.IsDir() {
+			path = filepath.Join(path, "skaffold.yaml")
 		}
-		return nil, fmt.Errorf("parsing dependencies for skaffold config %s: %w", cfgOpts.file, err)
-	}
-	if fi.IsDir() {
-		path = filepath.Join(path, "skaffold.yaml")
 	}
 	cfgOpts.isDependency = cfgOpts.isDependency || path != cfgOpts.file
```
briandealwis, line 215: This line is a bit puzzling and warrants a comment as I'm not sure how this situation can arise.

cmd/skaffold/app/cmd/parse_config_test.go
```diff
@@ -419,8 +441,11 @@ requires:
 				Profiles:            test.profiles,
 			})
 
-			t.CheckDeepEqual(test.err, err, cmp.Comparer(errorsComparer))
-			t.CheckErrorAndDeepEqual(test.err != nil, err, test.expected, cfgs)
+			if test.err != "" {
+				t.CheckDeepEqual(errors.New(strings.ReplaceAll(test.err, "WORK_DIR", wd)), err, cmp.Comparer(errorsComparer))
+			} else {
+				t.CheckDeepEqual(test.expected, cfgs)
+			}
```
briandealwis, line 448: Might be easier to have an `duplicateError` type, and instead use a function to check the correctness of the error.

Better yet: create an error-code for this situation and test that error code.  This should be reported upwards.

cmd/skaffold/app/cmd/parse_config.go
gsquared94: It can't exactly be inside `ParseConfigAndUpgrade()` since that works with `VersionedConfig` which doesn't have any method like `GetName`. To make that change I'd have to modify every version of the skaffold config to satisfy that interface. But you're right in that I can do this check upfront. Moved.

cmd/skaffold/app/cmd/parse_config.go
gsquared94: fixed typo; added description

cmd/skaffold/app/cmd/parse_config.go
gsquared94: don't know what I was doing here 🤦🏽‍♂️. Removed.

gsquared94: We can't effectively maintain all info in a local variable since we also want to print out the file names that cause the conflict.

cmd/skaffold/app/cmd/parse_config.go
gsquared94: This is a weird one. It was existing code so I didn't mess around with it. It's calculating that the `FileNotFound` error from `ReadConfiguration` is wrapped exactly once in `ParseConfig`. So it unwraps exactly once to get the underlying error. 🤦🏽‍♂️ 

I'll convert this to a concrete error type.

briandealwis: The [code for `IsNotExist` notes](https://golang.org/src/os/error.go?s=3221:3252#L86):
> New code should use `errors.Is(err, os.ErrNotExist)`

`errors.Is()` will unwrap the error sequentially.

gsquared94: oh, got it. Fixed

cmd/skaffold/app/cmd/parse_config_test.go
gsquared94: I will make this change in fixing https://github.com/GoogleContainerTools/skaffold/issues/5412

cmd/skaffold/app/cmd/parse_config.go
gsquared94: added

cmd/skaffold/app/cmd/parse_config.go
```diff
@@ -92,6 +92,19 @@ func getConfigs(cfgOpts configOpts, opts config.SkaffoldOptions, r *record) ([]*
 	}
 	logrus.Debugf("parsed %d configs from configuration file %s", len(parsed), cfgOpts.file)
 
+	// validate that config names are unique if specified
+	seen := make(map[string]bool)
+	for _, cfg := range parsed {
+		cfgName := cfg.(*latest.SkaffoldConfig).Metadata.Name
```
briandealwis, line 98: Maybe we should have a `ConfigName()` method on `VersionedConfig`?

cmd/skaffold/app/cmd/parse_config.go
briandealwis: What's previous and current?  Oh!  Is configOpts from the config that required `d` as a dependency?  That was not clear.

cmd/skaffold/app/cmd/parse_config_test.go
briandealwis: The tests are failing as you're getting windows path separators.  Maybe you need to replace `/` &rarr; `\` too in the WORKDIR references.

cmd/skaffold/app/cmd/parse_config.go
```diff
@@ -92,6 +92,19 @@ func getConfigs(cfgOpts configOpts, opts config.SkaffoldOptions, r *record) ([]*
...
+		if cfgName == "" {
+			continue
+		}
+		if seen[cfgName] {
+			return nil, fmt.Errorf("multiple skaffold configs named %q found in file %q", cfgName, cfgOpts.file)
```
briandealwis, line 103: Maybe `seen` should be a map of `name` &rarr; file?  Then you could include at least two of the file nams.

briandealwis (approved):
LGTM once the tests pass on Windows.

PR #5410, opened 2021-02-17
Add repo field to helm release

bobby-richard:
Fixes: #1665

**Description**
Currently, to use a a remote helm chart with skaffold, you must manually add the repository to your local helm config using `helm repo add`. In order to keep skaffold self contained, a repo field has been added to helm releases. If this field is set, the --repo argument will be passed to helm commands, allowing you to deploy remote helm charts without any additional helm configuration.


bobby-richard-sed:
Any comments? Would love to get this merged.

MarlonGamez:
Hey @bobby-richard, thanks for opening this. I'll be discussing this with the team shortly, I'll try to keep this updated with what we talk about. Do you think that we should simply remove the remote flag with this change? Since the repo field essentially tells us that it's remote and gives a bit more info?

bobby-richard:
Thanks @marlon-gamez. I thought about removing the remote property, but wasn't sure how to handle a breaking change to the schema. Also, if you remove the remote property, you lose the ability to specify the chartPath using the repo alias (though I'm not sure why you'd want to do that, if the repo field was available). I'm happy to take a stab at removing the remote field if you think that's the way to go.

tejal29:
> Thanks @marlon-gamez. I thought about removing the remote property, but wasn't sure how to handle a breaking change to the schema. Also, if you remove the remote property, you lose the ability to specify the chartPath using the repo alias (though I'm not sure why you'd want to do that, if the repo field was available). I'm happy to take a stab at removing the remote field if you think that's the way to go.

Thanks @bobby-richard & @marlon-gamez for bring the `remote` config discussion.

I looked more into the code to understand what the implication would be if we remove  `remote`.
This config  is used multiple places to detect remote charts and ignore file changes for dependencies etc in dev loop. 

This change looks good to me and i will merge it when kokoro passes.

tejal29 (approved):

bobby-richard:
@tejal29 it looks like kokoro passed

MarlonGamez:
Hey @bobby-richard, sorry for the wait, we wanted to see if the author of #5193 had any ideas before we merge this. Thank you for the contribution! :)

PR #5415, opened 2021-02-18
Add v2 Proto definitions

MarlonGamez:
**Related**: #5368 , #5422, #5423

**Description**
This PR adds the proto file containing the definitions for message types that will be used as part of the v2 API. Most of this is copy-paste from proto/v1/skaffold.proto. The main changes are to the Event API related messages, as well as the service definition.

- Add Dockerfile and update scripts to generate proto files
- Metadata proto messages will be updates later based on design from VSC IDE team


proto/v2/skaffold.proto
```diff
@@ -0,0 +1,819 @@
...
+    StatusCode statusCode = 6;
+    ActionableErr actionableErr = 7; // actionable error message
+}
+
+// PortForwardEvent Event describes each port forwarding event.
+message PortForwardEvent {
+    string id = 1; // id of the subtask which will be used in SkaffoldLog
+    string task_id = 2; // id of the task of skaffold that this event came from
+    int32 localPort = 3; // local port for forwarded resource
+    int32 remotePort = 4; // Deprecated. Uses targetPort.intVal.
```
tejal29, line 239: remove deprecated event here and everywhere.

proto/v2/skaffold.proto
```diff
@@ -0,0 +1,819 @@
...
+        option (google.api.http) = {
+            post: "/v2/events/handle"
+            body: "*"
+        };
+    }
+
+}
+
+// Enum indicating builders used
+enum BuilderType {
```
tejal29, line 379: I was wondering if there is a way to keep enum for types and  error codes not version specific. That ways, as we add error codes, you dont have to add them in two places.

tejal29 (changes requested):
Over all LGTM
1) Remove deprecated fields
2) How can we make enums not version specific.

proto/v2/skaffold.proto
MarlonGamez: Thanks for catching this, didn't realize I missed some

proto/v2/skaffold.proto
MarlonGamez: I was also wondering about this. Maybe we should split enum types off into a separate file and import them in both?
https://developers.google.com/protocol-buffers/docs/proto3#importing_definitions

proto/v2/skaffold.proto
```diff
@@ -0,0 +1,816 @@
...
+
+// `StatusCheckState` describes the state of status check of current deployed resources.
+message StatusCheckState {
+    string status = 1;
+    // A map of `resource name -> status-check-state`. Where `resource-name` is the kubernetes resource name.
+    // The `status-check-state` can be <br>
+    // - `"Not started"`: indicates that `status-check` has just started. <br>
+    // - `"In progress"`: InProgress is sent after every resource check is complete. <br>
+    // - `"Succeeded"`:
+    // - `"Failed"`:
```
briandealwis, line 108: Since these are codes, my 2¢ is to remove the spaces and camel case them so that they do not look like English.  Especially given the comment uses `InProgress`.

proto/v2/skaffold.proto
briandealwis: I mentioned the UNRECOGNIZED issue we had with IntelliJ to @tstromberg and he mentioned that Google APIs generally uses a small set of high-level error codes and then a detail `reason` which is a string representations. See [AIP-193](https://google.aip.dev/193).  I wonder if we should just reuse those definitions.

proto/v2/skaffold.proto
MarlonGamez: Good catch, this is actually in inconsistency that exists in v1, as the state API uses "In Progress" as a status while the event API uses "InProgress" as a status

proto/v2/skaffold.proto
MarlonGamez: ok, I will make a follow up PR to address this 👍🏼

tejal29 (approved):

PR #5416, opened 2021-02-19
Handle nil PortForward item on setting defaults

Laterality:
Fixes: #5363 

**Description**

Handles default-setting with list of `PortForward`s containing `nil` item.

When user run with following `skaffold.yaml`:

```yaml
apiVersion: skaffold/v2beta10
kind: Config
metadata:
  name: bug
portForward:
 -
```

Then, will get looks like:

```bash
parsing skaffold config: setting default values: Config's portForward[0] is empty, Please check if it has valid values
```

### DIFF
--- a/pkg/skaffold/schema/defaults/defaults.go
+++ b/pkg/skaffold/schema/defaults/defaults.go
@@ -98,7 +98,7 @@ func Set(c *latest.SkaffoldConfig) error {
 
 	for i, pf := range c.PortForward {
 		if pf == nil {
-			return fmt.Errorf("portForward[%d] of config name with '%s' is empty, Please check if it has valid values", i, c.Metadata.Name)
+			return fmt.Errorf("portForward[%d] of config with name '%s' is empty, Please check if it has valid values", i, c.Metadata.Name)
 		}
 		setDefaultLocalPort(pf)
 		setDefaultAddress(pf)
--- a/pkg/skaffold/schema/defaults/defaults.go
+++ b/pkg/skaffold/schema/defaults/defaults.go
@@ -96,7 +96,10 @@ func Set(c *latest.SkaffoldConfig) error {
 		return err
 	}
 
-	for _, pf := range c.PortForward {
+	for i, pf := range c.PortForward {
+		if pf == nil {
+			return fmt.Errorf("Config's portForward[%d] is empty, Please check if it has valid values", i)
+		}
 		setDefaultLocalPort(pf)
 		setDefaultAddress(pf)
 	}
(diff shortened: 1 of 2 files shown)


pkg/skaffold/schema/defaults/defaults.go
```diff
@@ -96,7 +96,10 @@ func Set(c *latest.SkaffoldConfig) error {
 		return err
 	}
 
-	for _, pf := range c.PortForward {
+	for i, pf := range c.PortForward {
+		if pf == nil {
+			return fmt.Errorf("config's portForward[%d] is empty, Please check if it has valid values", i)
```
MarlonGamez, line 101: Since it's possible for multiple configs to be tied together now/have multiple configs in the same file, we might want to try grabbing `c.Metadata.Name` so we can give them a little more information to the user with the error message

pkg/skaffold/schema/defaults/defaults_test.go
```diff
@@ -341,6 +341,20 @@ func TestSetPortForwardLocalPort(t *testing.T) {
...
+	cfg := &latest.SkaffoldConfig{
+		Pipeline: latest.Pipeline{
+			Build: latest.BuildConfig{},
+			PortForward: []*latest.PortForwardResource{
+				nil,
+			},
+		},
+	}
+	err := Set(cfg)
+	SetDefaultDeployer(cfg)
```
MarlonGamez, line 354: does `SetDefaultDeployer()` have to be called for this test?

MarlonGamez (commented):
Thanks for opening this PR! Have a couple comments but this looks good :)

pkg/skaffold/schema/defaults/defaults.go
Laterality: Applied, Seems quite handy.

pkg/skaffold/schema/defaults/defaults_test.go
Laterality: Not at all. I removed this line.

pkg/skaffold/schema/defaults/defaults.go
```diff
@@ -96,7 +96,10 @@ func Set(c *latest.SkaffoldConfig) error {
 		return err
 	}
 
-	for _, pf := range c.PortForward {
+	for i, pf := range c.PortForward {
+		if pf == nil {
+			return fmt.Errorf("portForward[%d] of config name with '%s' is empty, Please check if it has valid values", i, c.Metadata.Name)
```
MarlonGamez, line 101:
```suggestion
			return fmt.Errorf("portForward[%d] of config with name '%s' is empty, Please check if it has valid values", i, c.Metadata.Name)
```
I think the position of this word should be flipped

MarlonGamez (changes requested):
This looks good :) Just a small grammar change request

pkg/skaffold/schema/defaults/defaults.go
Laterality: Right, This should be 😅

MarlonGamez:
Thanks for making the changes :) Once CI passes I'll approve and merge!

MarlonGamez (approved):

PR #5426, opened 2021-02-20
Add example to use docker buildx via the custom builder

briandealwis:
**Description**
This example demonstrates how to use `docker buildx` as a Skaffold custom builder.

The one downside is that it requires setting up a `docker buildx` builder instance as a separate step.  Something like:

    docker buildx create --name skaffold-builder --platform linux/arm64,linux/amd64

We could create a more sophisticated build script that would examine the existing builder instances, but that's a bit more involved.

### DIFF
--- a/examples/custom-buildx/README.md
+++ b/examples/custom-buildx/README.md
@@ -47,7 +47,7 @@ $ cd skaffold/examples/custom-buildx
 
 Take a look at the [skaffold.yaml](skaffold.yaml), which uses a
 _custom builder_ to invoke `docker buildx` to containerize source
-code. 
+code.
 For more information about custom builders, see the Skaffold custom
 builder [documentation](https://skaffold.dev/docs/how-tos/builders/#custom-build-script-run-locally).
 Note that Skaffold builders are different from Docker Buildx builders.
--- a/examples/custom-buildx/README.md
+++ b/examples/custom-buildx/README.md
@@ -115,7 +115,7 @@ depending on the `$PUSH_IMAGE` flag.
 
 But Skaffold is unaware that the build result differs based on `$PUSH_IMAGE`.
 So on a local-build, Skaffold will cache the local-platform artifact,
-and that artifact will be used that for subsequent deployments even if
+and that artifact will be used that for subsequent deployments even if pushed
 to a remote registry (assuming the source is unchanged).  To avoid
 this scenario, disable Skaffold's artifact caching when the result
 is to be pushed to a remote registry:
@@ -123,4 +123,3 @@ is to be pushed to a remote registry:
 ```
 skaffold build --cache-artifacts=false
 ```
-
--- a/examples/custom-buildx/README.md
+++ b/examples/custom-buildx/README.md
@@ -6,7 +6,7 @@ platforms.
 
 This example shows how `docker buildx` can be used as a
 Skaffold _custom builder_ to create container images for
-for two different platforms: linux/arm64 and linux/amd64.
+for two different platforms: `linux/arm64` and `linux/amd64`.
 
 #### Before you begin
 
@@ -47,10 +47,17 @@ $ cd skaffold/examples/custom-buildx
 
 Take a look at the [skaffold.yaml](skaffold.yaml), which uses a
 _custom builder_ to invoke `docker buildx` to containerize source
-code.
+code for two platforms.
 For more information about custom builders, see the Skaffold custom
 builder [documentation](https://skaffold.dev/docs/how-tos/builders/#custom-build-script-run-locally).
-Note that Skaffold builders are different from Docker Buildx builders.
+Note that Skaffold builders are different from `docker buildx` builders.
+
+Next look at the Kubernetes [pod descriptor](k8s/pod.yaml) and notice
+the use of _node affinities_ to instruct Kubernetes to scheduler the workload
+on nodes running either `linux/arm64` or `linux/amd64`.  It is important
+to realize that Kubernetes does not examine the referenced container images
+to determine the possible platforms.
+
 
 ##### Step 3: Build and deploy the example
 
--- a/examples/custom-buildx/README.md
+++ b/examples/custom-buildx/README.md
@@ -97,11 +97,11 @@ Using `buildx` to build for multiple platforms has some subtle
 interactions with Skaffold's artifact caching.
 
 Skaffold normally caches the artifact after a successful container
-image build using the principle that a build should produce the
-same container image given the same source.  Skaffold records an
-association of the resulting image digest with a hash of the artifact
-source.  This principle is the same as how Docker and other builders
-cache image layers to speed up builds.
+image build.  Using the principle that given the same source inputs,
+a build should produce the same container image, Skaffold records
+an association of the resulting image digest with a hash of the
+artifact source.  This is the same principle used followed Docker
+and other builders to cache image layers to speed up builds.
 
 In this example, the build script configures Buildx differently
 depending on the `$PUSH_IMAGE` flag. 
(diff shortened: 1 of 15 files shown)


integration/examples/custom-buildx/main.go
```diff
@@ -0,0 +1,14 @@
+package main
+
+import (
+	"fmt"
+	"time"
+)
+
+func main() {
+	for {
+		fmt.Println("Hello world!")
```
tejal29, line 10: we could change this line to run 
```suggestion
		fmt.Printf("hello multi platform, I am %s-%s\n", runtime.GOOS, runtime.GOARCH) 
```

integration/examples/custom-buildx/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.16-alpine as builder
+COPY main.go .
+# `skaffold debug` sets SKAFFOLD_GO_GCFLAGS to disable compiler optimizations
+ARG SKAFFOLD_GO_GCFLAGS
+RUN go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -o /app main.go
```
tejal29, line 5: should we use `TARGETARCH` and `TARGETOS`  args here and pass them to `go build`
```suggestion
FROM --platform=$BUILDPLATFORM golang:alpine AS build                                                                       
                                                                       
ARG TARGETOS                                                                                                                
ARG TARGETARCH                                                                                                                                                                                          
RUN GOOS=$TARGETOS GOARCH=$TARGETARCH  go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -o /app main.go 
```

integration/examples/custom-buildx/README.md
```diff
@@ -0,0 +1,82 @@
...
+$ skaffold run --tail --default-repo <your repo>
+```
+
+With this command, Skaffold will build the artifact with `docker buildx`
+and deploy the application to Kubernetes.  You should be able to
+see *Hello, World!* printed every second in the Skaffold logs.
+
+If Skaffold fails with a message like the following, then Skaffold is
+attempting to load the multi-platform images into the Docker Daemon,
+is pushing the images
```
tejal29, line 69: Not sure i understand what this means. 
Is there a way to recover from this error?

integration/examples/custom-buildx/skaffold.yaml
```diff
@@ -0,0 +1,29 @@
...
+  # Note that building for multiple platforms requires pushing to a registry,
+  # as the Docker Daemon cannot load multi-platform images.
+  local:
+    push: true
+  artifacts:
+  - image: skaffold-examples-buildx
+    custom:
+      # This buildCommand depends on a `docker buildx` builder to have been created
+      # with the name `skaffold-builder`:
+      #   $ docker buildx create --name skaffold-builder --platform linux/amd64,linux/arm64
```
tejal29, line 14: do we also need to add the `docker buildx  use skaffold-builder` command. 
or add the `--use`  flag to `docker buildx create --name skaffold-builder --platform linux/amd64,linux/arm64` command.

integration/examples/custom-buildx/README.md
briandealwis: Oops, it looks like a line was deleted.

integration/examples/custom-buildx/skaffold.yaml
briandealwis: I have a slight preference to explicitly specifying the builder with `--builder` as the message if the user *hasn't* created a builder and continues to use the default docker builder is cryptic:

    error: multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")


Whereas explicitly specifying `--builder xxx` where `xxx` doesn't exist fails with a slightly more coherent message:

    error: open /Users/bdealwis/.docker/buildx/instances/foo: no such file or directory

examples/custom-buildx/README.md
```diff
@@ -0,0 +1,83 @@
+### Example: use the custom builder with `docker buildx`
+
+[Docker Buildx](https://github.com/docker/buildx#buildx) is an
+experimental feature for building container images for multiple
+platforms.
+
+This example shows how `docker buildx` can be used as a
+Skaffold _custom builder_ to create container images for
+for two different platforms: `linux/arm64` and `linux/amd64`.
```
PriyaModali, line 9: Remove duplicate `for`

examples/custom-buildx/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+Take a look at the [skaffold.yaml](skaffold.yaml), which uses a
+_custom builder_ to invoke `docker buildx` to containerize source
+code for two platforms.
+For more information about custom builders, see the Skaffold custom
+builder [documentation](https://skaffold.dev/docs/how-tos/builders/#custom-build-script-run-locally).
+Note that Skaffold builders are different from `docker buildx` builders.
+
+Next look at the Kubernetes [pod descriptor](k8s/pod.yaml) and notice
+the use of _node affinities_ to instruct Kubernetes to scheduler the workload
```
PriyaModali, line 56: Replace `scheduler` with `schedule`.

PriyaModali (dismissed):

briandealwis:
PTAL @PriyaModali

examples/custom-buildx/README.md
```diff
@@ -0,0 +1,126 @@
...
+  - When `true`, the result is to be pushed to a registry, and all
+    registries support multi-platform images. 
+  - When `false`, the result is to be loaded to the Docker Daemon. 
+    But Buildx does not support loading multi-platform images to
+    the Docker Daemon, and so this example builds a single container
+    image for the local platform.
+
+But Skaffold is unaware that the build result differs based on `$PUSH_IMAGE`.
+So on a local-build, Skaffold will cache the local-platform artifact,
+and that artifact will be used that for subsequent deployments even if
```
tejal29, line 118:
```suggestion
and that artifact will be used that for subsequent deployments even if pushed
```

examples/custom-buildx/README.md
```diff
@@ -0,0 +1,126 @@
...
+to a remote registry (assuming the source is unchanged).  To avoid
+this scenario, disable Skaffold's artifact caching when the result
```
tejal29, line 120: This makes me wonder, if users used `mult-arch` images in dev loop but in `build` command for ci-cd purpose when they want to build a multi-arch images.

If you could document something along the lines at the beginning, 
```
Using `buildx` to build for multiple platforms has some subtle
interactions with Skaffold's artifact caching.
We don't recommend using `buildx` in skaffold dev mode. Using `buildx` is better suited for building and pushing multi-arch images in your CI release pipeline.

```

examples/custom-buildx/README.md
```diff
@@ -0,0 +1,126 @@
...
+depending on the `$PUSH_IMAGE` flag. 
+
+  - When `true`, the result is to be pushed to a registry, and all
+    registries support multi-platform images. 
+  - When `false`, the result is to be loaded to the Docker Daemon. 
+    But Buildx does not support loading multi-platform images to
+    the Docker Daemon, and so this example builds a single container
+    image for the local platform.
+
+But Skaffold is unaware that the build result differs based on `$PUSH_IMAGE`.
```
tejal29, line 116:
```suggestion 
Skaffold is unaware that the build result differs based on `$PUSH_IMAGE`.
```

tejal29 (changes requested):

examples/custom-buildx/README.md
briandealwis: The "But" is helpful here as it is contrary to expectations.

examples/custom-buildx/README.md
briandealwis: The dev+multi-platform is an excellent observation and one that we'll take it into account with multi-platform support in Skaffold. (DD to come RSN.)

I'll add a note about `dev` use and suggest to use command-activated profiles.  This example is hairy enough! :-)

examples/custom-buildx/README.md
tejal29: they were too many buts. There was a but in the sentence above too. I am not an expert on english **but* in my mind  all sentences after the first "But" inherit the "but" clause?

tejal29 (approved):

examples/custom-buildx/README.md
tejal29: yeah i didn't say explicitly *but* in my mind I meant we can keep the example as is.

PR #5432, opened 2021-02-22
Adding distinct error codes for custom test failures

PriyaModali:
Fixes: #5431

Description
Adding error codes for custom test failures

User facing changes
These changes will enable Skaffold dev to return distinct error codes when custom tests fail.

There are no visible user facing changes.


tejal29:
Let's combine this with code changes and not a one off change.

PR #5438, opened 2021-02-22
Moving structure test's dependency logic to NewTester()

PriyaModali:
Fixes: #

**Description**
Moving structure test's dependency file extraction logic to NewTester() to reduce code redundancy.

There are no user facing changes.

This is part of adding new custom tester to Skaffold pipeline tracked by #5333.


tejal29:
After thinking more, i dont feel we should refactor this anymore. 
Test class is instantiated in init phase, but re-used in subsequents dev loops. We would want `TestDependencies` to pick up any changes in subsequent dev loop e.g. new dependency file added.

PR #5447, opened 2021-02-24
Add support for no-option-value; surface per-option-defaults

briandealwis:
This PR brings in some small changes to our command-flag processing:

- add support for pflag's `NoOptDefVal` for options that transition from being a boolean to a flag-with-values
- create individual pflag instances for each command so that per-command defaults are surfaced in the help text
- make `--enable-rpc` default to true for `debug`, like it is for `dev`.
- remove now-unecessary `by default` usage text since the default is now shown in-line

The proposal in #4832, to revise Skaffold's port-forwarding behaviour, wants to promote `--port-forward` from being a boolean flag to a string-slice flag.  The `NoOptDefVal` allows `--port-forward` to continue to work even though.

The per-command defaults lets the help text for options like `--enable-rpc` (which has [a different default for `dev`](https://github.com/briandealwis/skaffold/blob/eb355d666c81d65a6d8afac0897cb19ed9cf5ecf/cmd/skaffold/app/cmd/flags.go#L140-L149)) show the actual default value:
```
$ skaffold help run | grep enable-rpc
      --enable-rpc=false: Enable gRPC for exposing Skaffold events
      --event-log-file='': Save Skaffold events to the provided file after skaffold has finished executing, requires --enable-rpc=true

$ skaffold help dev | grep enable-rpc
      --enable-rpc=true: Enable gRPC for exposing Skaffold events
      --event-log-file='': Save Skaffold events to the provided file after skaffold has finished executing, requires --enable-rpc=true
```

### DIFF
--- a/cmd/skaffold/app/cmd/commands.go
+++ b/cmd/skaffold/app/cmd/commands.go
@@ -89,7 +89,7 @@ func (b *builder) WithFlagAdder(adder func(*pflag.FlagSet)) Builder {
 
 func (b *builder) WithFlags(flags []*Flag) Builder {
 	for _, f := range flags {
-		fl := f.flag()
+		fl := f.flag(b.cmd.Use)
 		b.cmd.Flags().AddFlag(fl)
 	}
 	b.cmd.PreRun = func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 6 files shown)


MarlonGamez:
PR looks good! Think it just needs a  ./hack/generate-man.sh

MarlonGamez (approved):

PR #5461, opened 2021-02-25
Update grpc.md

grmoon:
**Description**
Update broken github link to `skaffold.proto`


gsquared94 (approved):

MarlonGamez (approved):
Thanks for this!

MarlonGamez (changes requested):
grpc.md is actually generated from ./hack/generate-proto.sh. To make this change we should edit proto/v1/markdown.tmpl line 26 (and line 10 while we're at it) then generate the file again.

MarlonGamez (approved):

PR #5472, opened 2021-03-02
Modify `TestRun` to rerun all tests as imported configs

gsquared94:
Modify `TestRun` integration test to rerun all target tests by importing them as dependencies in a temp config.
```
kind: Config
apiVersion: skaffold/v2beta13
requires:
  - path: <test.dir>
```

Needed to add explicit `deploy` stanza since the default `kubectl` deployer isn't added to imported configs.

---

The motivation to add this test is to ensure new example/feature added to the list of test targets for `TestRun` also test that they work as dependencies.

### DIFF
--- a/integration/examples/bazel/skaffold.yaml
+++ b/integration/examples/bazel/skaffold.yaml
@@ -5,3 +5,7 @@ build:
   - image: skaffold-bazel
     bazel:
       target: //:skaffold_example.tar
+deploy:
+  kubectl:
+    manifests:
+      - k8s/*
--- a/integration/examples/buildpacks-java/skaffold.yaml
+++ b/integration/examples/buildpacks-java/skaffold.yaml
@@ -7,6 +7,10 @@ build:
       builder: "gcr.io/buildpacks/builder:v1"
       env:
       - GOOGLE_RUNTIME_VERSION=8
+deploy:
+  kubectl:
+    manifests:
+      - k8s/*
 profiles:
 - name: gcb
   build:
--- a/integration/examples/buildpacks-node/skaffold.yaml
+++ b/integration/examples/buildpacks-node/skaffold.yaml
@@ -5,3 +5,7 @@ build:
   - image: skaffold-buildpacks-node
     buildpacks:
       builder: "gcr.io/buildpacks/builder:v1"
+deploy:
+  kubectl:
+    manifests:
+      - k8s/*
--- a/integration/examples/buildpacks-python/skaffold.yaml
+++ b/integration/examples/buildpacks-python/skaffold.yaml
@@ -5,6 +5,10 @@ build:
   - image: skaffold-buildpacks
     buildpacks:
       builder: "gcr.io/buildpacks/builder:v1"
+deploy:
+  kubectl:
+    manifests:
+      - k8s/*
 profiles:
 - name: gcb
   build:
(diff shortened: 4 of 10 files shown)


briandealwis (commented):
> The motivation to add this test is to ensure new example/feature added to the list of test targets for TestRun also test that they work as dependencies.

Why is this important?  Is this exercising new functionality that would otherwise be untested?  Is there functionality in our examples that others may want to reuse?

This change will add more time to our tests, and I'd like to ensure there's a benefit.

integration/examples/bazel/skaffold.yaml
```diff
@@ -5,3 +5,7 @@ build:
   - image: skaffold-bazel
     bazel:
       target: //:skaffold_example.tar
+deploy:
+  kubectl:
+    manifests:
+      - k8s/*
```
briandealwis, line 11: Since Skaffold adds this by default, should it be defaulted for the required-config case too?

gsquared94: > Why is this important? Is this exercising new functionality that would otherwise be untested? Is there functionality in our examples that others may want to reuse?
> 
> This change will add more time to our tests, and I'd like to ensure there's a benefit.

So the main point is to test that new features don't incorrectly assume `working directory`. For instance, https://github.com/GoogleContainerTools/skaffold/pull/5451 was adding new config fields that reference to files on the filesystem. It'd be a manual activity every time to make sure that `skaffold:"filepath"` tags are added or that there's no assumption about a global `workDir`.  WDYT?

gsquared94: > Since Skaffold adds this by default, should it be defaulted for the required-config case too?

In a single config it's plausible to add a default deploy stanza. For multiple configs, there might be build-only configs (see our [microservices](https://github.com/GoogleContainerTools/skaffold/blob/master/examples/multi-config-microservices/base/skaffold.yaml) example) which aren't deployed. WDYT?

briandealwis: Not all of our examples are included in this TestRun.  This seems an expensive way to catch fields missing the filepath tag.

I think you could have a test that walks all of the `skaffold.yaml` found under `integration/` and:
  1. loads the skaffold.yaml, 
  2. walks the fields looking for fields *without* the `filepath` tag
  3. checks to see if a file by the value exists under the path of the skaffold.yaml and if so, throws an error.

(We'll need special casing for file references within artifacts.)

gsquared94: We don't do any validations with our `filepath` tag so I'm not sure about your suggestion. To clarify about my intent, consider some feature implemented whose definition in `skaffold.yaml` looks like:
```yaml
fooFeature:
  path: ./bar.sh
```
 and the `config.go` equivalent struct is:
 ```go
 type FooFeature  struct {
  Path string `yaml:"path"`   // missing `filepath` tag
}
 ```

and in some `foo.go` file it's logic is:
```go
wd, _ := os.Getwd()
fooPath := filepath.Join(wd, config.FooFeature.Path)
```
This `skaffold.yaml` cannot be used as a dependency from other `skaffold.yaml` files in different directories since the value of `fooFeature.path` changes with `wd`.

If the above struct definition had the `filepath` tag included:

```go

 type FooFeature  struct {
  Path string `yaml:"path" skaffold:"filepath"` 
}
```

then the path would be always relative to the `skaffold.yaml` file it's defined in. This way the same file can be imported as a dependency from any other directory.

---

If it's too expensive to run it on the list of sample projects how about running it only on newly added projects (diff between `integration/examples` and `examples`) ? That'd add an occasional 1-2 tests.

Issue #5425, opened 2021-02-20
Debug docs should warn about re-entrancy with probes


briandealwis:
We encountered a situation where ~a breakpoint was placed on the calling path of~ a user was stopped in a breakpoint.  This pod had a liveness probe that called back into the app. Because the server was blocked due to the debug session in progress, and liveness probes have a timeout of 1 second, the probe was deemed to have failed and k8s restarted the container.
```
  containers:
  - name: example
    image: example-image
    ports:
    - containerPort: 8080
    # connect to the app port
    livenessProbe:
      httpGet:
        path: /
        port: 8080
      initialDelaySeconds: 2
      periodSeconds: 10
```

it makes sense that a developer might want to debug a liveness or readiness or startup probe. Debug could rewrite such probes to set/increase the timeout period of 5 minutes (ideally configurable in some way).

cc: @VeerMuchandi

PR #5474, opened 2021-03-02
Debug probes

briandealwis:
Fixes: #5425

**Description**
Cause `skaffold debug` to rewrite any readiness/liveness/startup probes using HTTP-Get to set the timeout value to 10 minutes (from the default of 1s).  This change allows debugging calls from such probes, and also avoids timeouts when a probe is made but the application is suspended because of an ongoing debug session.

- `skaffold debug` alters Kubernetes HTTP-Get probes to have a 10 minute timeout to allow debugging probe calls

Consider making this configurable somehow.

### DIFF
--- a/pkg/skaffold/debug/transform.go
+++ b/pkg/skaffold/debug/transform.go
@@ -55,6 +55,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	shell "github.com/kballard/go-shellquote"
 	"github.com/sirupsen/logrus"
@@ -116,7 +117,12 @@ const (
 	debuggingSupportFilesVolume = "debugging-support-files"
 
 	// DebugConfigAnnotation is the name of the podspec annotation that records debugging configuration information.
+	// The annotation should be a JSON-encoded map of container-name to a `ContainerDebugConfiguration` object.
 	DebugConfigAnnotation = "debug.cloud.google.com/config"
+
+	// DebugProbesAnnotation is the name of the podspec annotation that disables rewriting of probe timeouts.
+	// The annotation value should be `skip`.
+	DebugProbeTimeoutsAnnotation = "debug.cloud.google.com/probe/timeouts"
 )
 
 // containerTransforms are the set of configured transformers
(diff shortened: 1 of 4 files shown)


pkg/skaffold/debug/transform.go
```diff
@@ -199,6 +199,51 @@ func transformManifest(obj runtime.Object, retrieveImageConfiguration configurat
...
+		return false
+	}
+	var config map[string]ContainerDebugConfiguration
+	if err := json.Unmarshal([]byte(annotation), &config); err != nil {
+		logrus.Warn("error unmarshalling config", err)
+		return false
+	}
+
+	changed := false
+	var minTimeout int32 = 10 * 60 // make it configurable?
```
tejal29, line 221: Add seconds to clarify this is in seconds as default duration is Nanosecond
```suggestion
	var minTimeoutSeconds int32 = 10 * 60 // make it configurable?
```

pkg/skaffold/debug/transform.go
```diff
@@ -199,6 +199,51 @@ func transformManifest(obj runtime.Object, retrieveImageConfiguration configurat
...
+	containers := rewriteContainers(metadata, podSpec, retrieveImageConfiguration, debugHelpersRegistry)
+	timeouts := rewriteProbes(metadata, podSpec) // must rewriteProbes after
+	return containers || timeouts
+}
+
+// rewriteProbes rewrites k8s probes to expand timeouts to 10 minutes to allow debugging local probes.
+func rewriteProbes(metadata *metav1.ObjectMeta, podSpec *v1.PodSpec) bool {
+	annotation, found := metadata.Annotations[DebugConfigAnnotation]
+	if !found {
+		logrus.Warn("no debug config found")
```
tejal29, line 211: Debug annotation not present was not surfaced out as a warning before. Is there a reason why we want to show it now?

tejal29: Can we move checking the pod debug annotation at L202 and skip both if no annotation is found?

```
 func transformPodSpec(metadata *metav1.ObjectMeta, podSpec *v1.PodSpec, retrieveImageConfiguration configurationRetriever, debugHelpersRegistry string) bool {
	// skip annotated podspecs — allows users to customize their own image
	if _, found := metadata.Annotations[DebugConfigAnnotation]; found {
		return false
	}
        containers := rewriteContainers(metadata, podSpec, retrieveImageConfiguration, debugHelpersRegistry)
	timeouts := rewriteProbes(metadata, podSpec) // must rewriteProbes after
	return containers || timeouts
}
```

tejal29 (approved):
LGTM. 
few nits

pkg/skaffold/debug/transform.go
briandealwis: You're right: this shouldn't be a warning.  There's an assumption that the probes are only rewritten for containers that were transformed for debugging (I'll make that explicit).

pkg/skaffold/debug/transform.go
briandealwis: Will change to a duration.

briandealwis:
@tejal29 could you take a quick peek: I added support for disabling the probes and also overriding the probe timeout value too.

tejal29 (approved):

PR #5481, opened 2021-03-03
move enum protos into separate folder/package

MarlonGamez:
**Related**: #5368 

Apologies for large delta, most of this is generated files.

**Description**
This moves the enum definitions from proto/v1/skaffold.proto into proto/enums/enums.proto. This is to keep all of our enum definitions in one place with the addition of proto/v2, which will use the enums as well. To keep enum definitions on the docs page, I generate the enum docs, then the API docs and append the enum docs to the API ones. 

I've also updated some of the dependencies used to generate docs, as I think the updates to the dependencies make the docs a bit more readable by clarifying between enum types and standard proto message types.


MarlonGamez:
Looks like this is causing issues with the HTTP docs generation, I'm looking into how I can fix this

tejal29 (dismissed):

tejal29 (commented):
LGTM!

MarlonGamez:
@tejal29 I think the HTTP docs should be fixed now, PTAL when you get a chance :)

tejal29 (approved):

docs/content/en/docs/references/api/grpc.md
```diff
@@ -19,11 +19,11 @@ We also generate the [reference doc for the HTTP layer]({{<relref "/docs/referen
 
 
 
-<a name="skaffold.proto"></a>
```
briandealwis, line 22: The comments at the top of this file references `proto/markdown.tmpl`.  Maybe `hack/generate_proto.sh` needs to be fixed up?

proto/v1/skaffold.pb.go
```diff
@@ -1,11 +1,12 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
-// source: skaffold.proto
+// source: v1/skaffold.proto
 
 package proto
 
 import (
 	context "context"
 	fmt "fmt"
+	enums "github.com/GoogleContainerTools/skaffold/proto/enums"
```
briandealwis, line 9: Can this be sorted to preserve the Skaffold style of placing Skaffold imports at the end?

Issue #5459, opened 2021-02-25
helm remote deploy fails when "required" by another config


m1o1:
- Skaffold version: version: **v1.20.0**
- Operating system: **Windows 10**

Scenario 1:

If I skaffold run with this, it successfully installs the ingress-nginx helm chart:
`skaffold.yaml`:
```yaml
apiVersion: skaffold/v2beta12
kind: Config
metadata:
  name: myapp
deploy:
  helm:
    releases:
      - name: nginx
        namespace: nginx
        createNamespace: true
        chartPath: ingress-nginx/ingress-nginx
        remote: true
        setValues:
          controller.service.type: NodePort
```
---
Scenario 2:

If I want to put this nginx config in a different skaffold config like so:
`skaffold.yaml`:
```yaml
apiVersion: skaffold/v2beta12
kind: Config
metadata:
  name: myapp
requires:
- path: ./skaffold/nginx.yaml
```
`skaffold/nginx.yaml`:
```yaml
apiVersion: skaffold/v2beta12
kind: Config
metadata:
  name: nginx
deploy:
  helm:
    releases:
      - name: nginx
        namespace: nginx
        createNamespace: true
        chartPath: ingress-nginx/ingress-nginx
        remote: true
        setValues:
          controller.service.type: NodePort
```
then it fails with this error:
```
$ skaffold run
Generating tags...
Checking cache...
Tags used in deployment:
Starting deploy...
Helm release nginx not installed. Installing...
Error: path "D:\\dev\\myapp\\skaffold\\ingress-nginx\\ingress-nginx" not found
install: exit status 1
```
It seems like in this configuration, it's ignoring the `remote: true` in the helm deploy configuration.

m1o1:
I'm guessing this is related to the [Multiple working directories](https://docs.google.com/document/d/1dA1549Rj9xfAt5sHh-wHng1MQT1A9SOylLt4xEjctGk/edit#) of the design doc for multi-config skaffold. 

Also, in the `chartPath` and `valuesFiles` below, the path is somewhat unintuitive since it's relative to the top-level skaffold file (or maybe where I'm running skaffold from), not the config that it's parsing at that time. Maybe this is what the design doc is planning to fix though (I assumed it was merged in already though, since the `requires` field is in).

With the "top-level" skaffold in `$workspaceDir/.local-dev/skaffold.yaml`, and the following skaffold.yaml for myapp2:

`$workspaceDir/myapp2/skaffold.yaml`:
[...]

PR #5482, opened 2021-03-03
decouple `helm` deployer `chartPath` into `chartPath` and `remoteChart`

gsquared94:
Fixes: #5459  

---

**Description**

Currently the `helm` deployer's `chartPath` property can be used to specify the path to a local Helm chart (packaged or unpacked) or a Helm chart reference (like `stable/nginx`) or even a URL (like `https://example.com/charts/nginx-1.2.3.tgz`). It uses the additional `remote` property to hint skaffold about this.

This PR replaces `chartPath` and `remote` properties with explicit `chartPath` and `remoteChart` properties.

---

**Before**
```yaml
deploy:
  helm:
    releases:
    - name: foo1
      chartPath: ./path/to/chart
    - name: foo2
      chartPath: foo2/bar
      remote: true
```

**After**
```yaml
deploy:
  helm:
    releases:
    - name: foo1
      chartPath: ./path/to/chart
    - name: foo2
      remoteChart: foo2/bar
```

### DIFF
--- a/pkg/skaffold/deploy/helm/deploy.go
+++ b/pkg/skaffold/deploy/helm/deploy.go
@@ -173,7 +173,7 @@ func (h *Deployer) Dependencies() ([]string, error) {
 		r := release
 		deps = append(deps, r.ValuesFiles...)
 
-		if r.LocalChartPath == "" {
+		if r.ChartPath == "" {
 			// chart path is only a dependency if it exists on the local filesystem
 			continue
 		}
@@ -202,7 +202,7 @@ func (h *Deployer) Dependencies() ([]string, error) {
 			}
 
 			for _, v := range chartDepsDirs {
-				if strings.HasPrefix(path, filepath.Join(release.LocalChartPath, v)) {
+				if strings.HasPrefix(path, filepath.Join(release.ChartPath, v)) {
 					return false, nil
 				}
 			}
@@ -216,7 +216,7 @@ func (h *Deployer) Dependencies() ([]string, error) {
 			return true, nil
 		}
 
-		if err := walk.From(release.LocalChartPath).When(isDep).AppendPaths(&deps); err != nil {
+		if err := walk.From(release.ChartPath).When(isDep).AppendPaths(&deps); err != nil {
 			return deps, userErr("issue walking releases", err)
 		}
 	}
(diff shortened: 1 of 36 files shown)


pkg/skaffold/deploy/helm/deploy.go
```diff
@@ -253,7 +253,13 @@ func (h *Deployer) Render(ctx context.Context, out io.Writer, builds []build.Art
 	renderedManifests := new(bytes.Buffer)
 
 	for _, r := range h.Releases {
-		args := []string{"template", r.ChartPath}
+		var chartPath string
+		if r.LocalChartPath != "" {
+			chartPath = r.LocalChartPath
+		} else {
+			chartPath = r.RemoteChartPath
+		}
```
tejal29, line 261: How about reducing this or adding a utility function which 
```suggestion
		chartPath := r.RemoteChartPath
		if r.LocalChartPath != "" {
			chartPath = r.LocalChartPath
		}
```

e.g. 
```
func getChartPath(h latest.HelmRelease) string {
   if h.LocalChartPath != "" {
      return h.LocalChartPath
   } 
  return r.RemoteChartPath
}

```

tejal29 (commented):
just a nit

pkg/skaffold/deploy/helm/deploy.go
gsquared94: done

MarlonGamez:
I have a question about this since https://github.com/GoogleContainerTools/skaffold/pull/5410 just got merged. Would it be possible to keep just the `chartPath` field, remove the `remote` field, and basically toggle on/off the remote functionality based on whether or not the user sets the `repo` field (introduced in #5410)?

gsquared94:
> I have a question about this since #5410 just got merged. Would it be possible to keep just the `chartPath` field, remove the `remote` field, and basically toggle on/off the remote functionality based on whether or not the user sets the `repo` field (introduced in #5410)?

I think that approach won't work because the `chartPath` can be remote without specifying `--repo` if the user has already manually pulled it. Also it can be an absolute url in which case the `--repo` is not required as it is part of that url.

briandealwis (commented):
`remoteChartPath` isn't actually a path.  Could we try the following:
- keep `chartPath` for local charts
- `remoteChartPath` &rarr; `chartRef` or `remoteChart`?

MarlonGamez (approved):
LGTM. Do we have anything in `examples/` that uses remote helm charts? If not I think that would be a good follow up for this

gsquared94:
> LGTM. Do we have anything in `examples/` that uses remote helm charts? If not I think that would be a good follow up for this

Good idea. Added #5517

zestrells:
Hello @gsquared94 , It looks like when we I try and use the remoteChart variable, it does not work. 
`deploy:
  helm:
    releases:
      - name: mysql
        namespace: data
        remoteChart: https://gitlab.com/helm-charts/devops/simple-template
        valuesFiles: [ helm/mysql-values.yaml ]
`

briandealwis:
@zestrells try `skaffold fix`: it provides more detail on the issue.  I filed #5659 to improve this error message.

briandealwis:
Oh and more specifically, that error often indicates that you need to update the version of your skaffold.yaml

Issue #5485, opened 2021-03-03
internal server error on helm install should be system error


briandealwis:
@PriyaModali saw an _Internal server error_ when deploying a Helm chart in Kokoro, but Skaffold mis-attributes the error as a user error (DEPLOY_HELM_USER_ERR):

```
    dev_test.go:265: event received timestamp:<seconds:1614793438 nanos:168027477 > event:<deployEvent:<status:"Failed" err:"install: exit status 1" errCode:DEPLOY_HELM_USER_ERR actionableErr:<errCode:DEPLOY_HELM_USER_ERR message:"install: exit status 1" > > > entry:"Deploy failed"
    dev_test.go:265: event received timestamp:<seconds:1614793438 nanos:168871950 > event:<devLoopEvent:<status:"Failed" err:<errCode:DEPLOY_HELM_USER_ERR message:"install: exit status 1" > > > entry:"Update failed with error code DEPLOY_HELM_USER_ERR"
exiting dev mode because first deploy failed: install: exit status 1
time="2021-03-03T17:43:59Z" level=info msg="Ran [skaffold dev --namespace skaffoldq6hn5 --default-repo gcr.io/k8s-skaffold --port-forward --rpc-port 61893] in 23.358 seconds"
    dev_test.go:251: timed out waiting for port forwarding event
    helper.go:184: Skaffold log:
         Listing files to watch...
        >  - gke-loadbalancer
        > Generating tags...
        >  - gke-loadbalancer -> gcr.io/k8s-skaffold/gke-loadbalancer:v1.20.0-64-gd962ed65f
        > Checking cache...
        >  - gke-loadbalancer: Found Remotely
        > Tags used in deployment:
        >  - gke-loadbalancer -> gcr.io/k8s-skaffold/gke-loadbalancer:v1.20.0-64-gd962ed65f@sha256:e496ba5bd8562eeb30e09a14d24844aaee13598af5bf83ac5c4266efd6f23d90
        > Starting deploy...
        > Helm release skaffold-helm-skaffoldq6hn5 not installed. Installing...
        > Error: an error on the server ("Internal Server Error: \"/api/v1/namespaces/skaffoldq6hn5/services\": the server is currently unable to handle the request") has prevented the request from succeeding (post services)
```

### Information

- Skaffold version: 
- Operating system: 
- Installed via: 
- Contents of skaffold.yaml:

```yaml
<paste your skaffold.yaml here>
```

### Steps to reproduce the behavior

[...]

PR #4933, opened 2020-10-23
Adding distinct exit codes for cluster connection failures.

PriyaModali:
Fixes: #4645 
**Related**: Tracking issue #4921
**Description**
Adding distinct exit codes for cluster connection failures.

**User facing changes**
These changes will enable Skaffold dev, deploy, to return distinct error codes when kubernetes cluster could not be reached.

Ex. before error messages:
```
Unable to connect to the server: net/http: TLS handshake timeout

exiting dev mode because first deploy failed: unable to connect to Kubernetes: Get "https://192.168.64.3:8443/version?timeout=32s": net/http: TLS handshake timeout
```

Ex. after error messages:
```
Deploy Failed. Could not connect to <cluster name> cluster.

Deploy Failed. Could not connect to cluster test_cluster due to \"https://192.168.64.3:8443/version?timeout=32s\": net/http: TLS handshake timeout. Check your connection for the cluster.
```

PR #5490, opened 2021-03-04
Use runCtx instead of of config.Opts for suggesting suggestions. 

tejal29:
Relates to #5485

This is a refactor, to break change in smaller PR for #5485.
When adding deploy error codes #4933, i missed the part where `kubectx` gets called directly. We have that information in `runCtx`.  It easier to use the evaluated context there instead of re-writing logic to look up `kubeContext`

Pros
1) Remove dependency on `pkg/skaffold/config` and `pkg/skaffold/kubernetes/context` from errors. 
2) In upcoming PR, we need the logic to determine minikube vs remote cluster for "Internal System Error". Using runCtx is much easier than again evaluating kubecontext in `deploy/errors` package.

### DIFF
--- a/pkg/skaffold/errors/errors.go
+++ b/pkg/skaffold/errors/errors.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/instrumentation"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner/runcontext"
 	"github.com/GoogleContainerTools/skaffold/proto/v1"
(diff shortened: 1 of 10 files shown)


pkg/skaffold/errors/errors.go
```diff
@@ -36,16 +37,16 @@ const (
...
-
-	// Report issue text
-	reportIssueText = "If above error is unexpected, please open an issue https://github.com/GoogleContainerTools/skaffold/issues/new to report this error"
 )
 
 var (
-	setOptionsOnce sync.Once
-	skaffoldOpts   config.SkaffoldOptions
+	// Report issue text
+	reportIssueText = fmt.Sprintf("If above error is unexpected, please open an issue %s to report this error", constants.GithubIssueLink)
```
MarlonGamez, line 44: should we just move this to const as well?

MarlonGamez (commented):
refactor looks good, have a small question

pkg/skaffold/errors/errors.go
tejal29: I usually prefer `fmt.Sprintf` over concatenation but i can change it.

MarlonGamez: oh my bad, I didn't realize you can't use it to create const values. Up to you if you want to leave it or not

MarlonGamez (approved):
Looks good to me :)

PR #5527, opened 2021-03-10
Refactor and reuse util.StringSet

tejal29:
Remove code redundancy and reuse `util.StringSet` to merge namespaces to array with de-dups.

The change in test is due to
1) util.StringSet.ToList will return nil []string instead of empty array of string.
This should not result into any nil point exception as we usually iterate over this array using `range` which should work with nil arrays.

### DIFF
--- a/pkg/skaffold/runner/runcontext/context.go
+++ b/pkg/skaffold/runner/runcontext/context.go
@@ -19,7 +19,6 @@ package runcontext
 import (
 	"fmt"
 	"os"
-	"sort"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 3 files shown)


pkg/skaffold/runner/runcontext/context.go
```diff
@@ -234,20 +234,8 @@ func (rc *RunContext) UpdateNamespaces(ns []string) {
...
-	}
-
-	// Update RunContext Namespace
-	updated := make([]string, 0, len(nsMap))
-	for k := range nsMap {
-		updated = append(updated, k)
-	}
-	sort.Strings(updated)
-	rc.Namespaces = updated
+	set := util.NewStringSet()
```
MarlonGamez, line 237: would rename `set` to `namespaces` or something a bit more clear

MarlonGamez (commented):
Looks good! Just a tiny nit

MarlonGamez (approved):
LGTM!

PR #5531, opened 2021-03-11
fix: `skaffold diagnose` outputs incorrect `yaml` for multiconfig projects

gsquared94:
`skaffold diagnose --yaml-only` used on a multiconfig project outputs a merged list of all resolved configs.
For example:
_skaffold.yaml_:

```yaml
...
requires:
  - path: ./skaffold2.yaml
build:
  - image: foo1
...
```
_skaffold2.yaml_:
```yaml
...
build:
  - image: foo2
...
```
should produce:
```yaml
...
build:
  - image: foo1
...
---
...
build:
  - image: foo2
...
```
But it's actually producing:
```yaml
...
requires:
  - path: ./skaffold2.yaml
build:
  - image: foo1
...
---
...
build:
  - image: foo2
...
```
If this output yaml is used as the input for another skaffold run then the `skaffold2.yaml` artifacts are duplicated since it's reimported. 
This PR fixes this issue by removing the dependency list prior to marshalling.

### DIFF
--- a/cmd/skaffold/app/cmd/diagnose.go
+++ b/cmd/skaffold/app/cmd/diagnose.go
@@ -64,6 +64,10 @@ func doDiagnose(ctx context.Context, out io.Writer) error {
 			color.Blue.Fprintln(out, "\nConfiguration")
 		}
 	}
+	// remove the dependency config references since they have already been imported and will be marshalled together.
+	for i := range configs {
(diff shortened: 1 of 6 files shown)


nkubala (approved):
nice, thanks for jumping on this so quickly! tried it out and it works perfectly

PR #5548, opened 2021-03-14
Fix setting of defaults for flags

briandealwis:
**Related**: #5447

**Description**
With the `--port-forward` refactoring (#4832), I'm turning `port-forward` into a string-slice argument with different per-command defaults.  The default will generally be `user`,  but for `skaffold debug` it will be `user,debug`.  

#5447 fixed how per-command default values were surfaced in help text. But in processing the arguments, the port-forward values for `skaffold debug` were being set to `[]string{"[user debug]"}` rather than `[]string{"user","debug"}`.

It turns out the helper function that actually set the per-command defaults (`ParseFlags()`) doesn't handle string-slice arguments and would use [`String()` on a `[]string{"user","debug"}`](https://github.com/briandealwis/skaffold/blob/f6da11b12fa5018ce438dca7cdbe99bfddb01b2a/cmd/skaffold/app/cmd/flags.go#L558).

By way of background, the cobra+pflags packages define a set of flags for a set of commands.  Each flag has a pointer to a _value holder_ (e.g., pointer to `string` for a string value), and a default value.  The value holder is set to the default when the flag is defined.  So if multiple flags are defined with the same pointer (e.g., the `run`, `dev`, and `debug` commands all define `--port-forward`  but with different defaults, and the value holder is the global SkaffoldOptions.PortForward string slice) then the value holder will be the default for the last command processed.  To remedy this, [our command setup calls `ParseFlags()` to reset the flag defaults](https://github.com/briandealwis/skaffold/blob/f6da11b12fa5018ce438dca7cdbe99bfddb01b2a/cmd/skaffold/app/cmd/flags.go#L584-L585) before processing the command-line arguments.

This PR:

  1. Causes `ParseFlags()` to *always* reset the flag value, instead of just for per-command defaults.  This value is set  each time a command flag is created to the flag's default value, and so the value will be set to the last command flag's default.
  2. Causes `ParseFlags()` to handle string-slice values by using `SliceValue.Replace()`.
  1. Renames `ParseFlags` to `ResetFlagDefaults()`.
  2. Adds tests to verify the values are set to the expected values.

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -548,15 +548,21 @@ func reflectValueOf(values []interface{}) []reflect.Value {
 	return results
 }
 
-func ParseFlags(cmd *cobra.Command, flags []*Flag) {
+func ResetFlagDefaults(cmd *cobra.Command, flags []*Flag) {
 	// Update default values.
 	for _, fl := range flags {
 		flag := cmd.Flag(fl.Name)
-		if fl.DefValuePerCommand != nil {
-			if defValue, present := fl.DefValuePerCommand[cmd.Use]; present {
-				if !flag.Changed {
-					flag.Value.Set(fmt.Sprintf("%v", defValue))
+		if !flag.Changed {
+			defValue := fl.DefValue
+			if fl.DefValuePerCommand != nil {
+				if d, present := fl.DefValuePerCommand[cmd.Use]; present {
+					defValue = d
 				}
+			}
+			if sv, ok := flag.Value.(pflag.SliceValue); ok {
+				reflect.ValueOf(sv).MethodByName("Replace").Call(reflectValueOf([]interface{}{defValue}))
+			} else {
+				flag.Value.Set(fmt.Sprintf("%v", defValue))
 			}
 		}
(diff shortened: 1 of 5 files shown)


cmd/skaffold/app/cmd/flags.go
```diff
@@ -548,16 +548,22 @@ func reflectValueOf(values []interface{}) []reflect.Value {
...
-					flag.Value.Set(fmt.Sprintf("%v", defValue))
+		if !flag.Changed {
+			defValue := fl.DefValue
+			if fl.DefValuePerCommand != nil {
+				if d, present := fl.DefValuePerCommand[cmd.Use]; present {
+					defValue = d
 				}
 			}
+			if sv, ok := flag.Value.(pflag.SliceValue); ok {
+				reflect.ValueOf(sv).MethodByName("Replace").Call(reflectValueOf([]interface{}{defValue}))
```
tejal29, line 563: I like to stay away from reflections as it can easily lead to runtime errors. However, unit tests would cover it in and if we upgrade the cobra library and this method is no longer supportedl

tejal29 (approved):

PR #5552, opened 2021-03-15
add number of releases in helm config

tejal29:
when re-designing config for helm, i saw we have 
```
helm:
  releases:
      - name: "release-foo"
      - name: "release-bar"
```

I was wondering how many users did have multiple helm releases defined in one app. 
If this is unused, it will simply config redesign a lot.

Also note: When doing some competitive config research, i found out, they don't support multiple releases since the `helm` is not used to deploy but `helm template | kubectl apply` is used.


gsquared94 (approved):
lgtm!

PR #5555, opened 2021-03-16
Remove unwanted value recorded.

tejal29:
While doing Code walks, @IsaacPD mentioned,  both "flag_value" and "value" are duplicates and used for recording flag value 
```
			label.String("flag_value", v),
			label.String("command", meter.Command),
			label.String("value", v),
```

Removing "value" in favor of the other.


briandealwis (approved):

tejal29:
ignoring the checks as the branch was created from master using Github UI

Issue #5413, opened 2021-02-18
Add control API to get list of all modules reachable from a target `skaffold.yaml` file


gsquared94:
Provide a control API to get the list of valid module names that can be provided as input to `--module` flag.
Required for integration with Cloud Code IDEs.

gsquared94:
Solved with the command `skaffold inspect modules list` added in https://github.com/GoogleContainerTools/skaffold/pull/5765

PR #5564, opened 2021-03-17
Add design proposal for `skaffold inspect`

gsquared94:
Related: #5413 
Add design proposal for new skaffold command `inspect`.

### DIFF
--- a/docs/design_proposals/inspect-cmd.md
+++ b/docs/design_proposals/inspect-cmd.md
@@ -21,7 +21,7 @@ ___
 skaffold inspect --help
 Display all resolved skaffold configurations.
 
- You can use --output jsonpath={...} to extract specific values using a jsonpath expression.
+ You can use --output jsonpath={...} to extract specific values using a jsonpath expression. The configs defined in the main `skaffold.yaml` file can be filted using the special `#` index; if there are multiple matches then it'll return a slice. 
 
 Examples:
   # Show all resolved configurations with applied profiles.
@@ -31,10 +31,10 @@ Examples:
   skaffold inspect --raw
   
   # Get all named configs (modules) as a list
-  skaffold inspect -o jsonpath='{.metadata[?(@.name != "")].name}'
+  skaffold inspect -o jsonpath='{.configs[?(@.metadata.name != "")].metadata.name}'
 
-  # Get all profile names.
-  skaffold inspect -o jsonpath='{.profiles[*].name}'
+  # Get all profile names from the main `skaffold.yaml` configs.
+  skaffold inspect -o jsonpath='{.configs[#].profiles[*].name}'
 
 Options:
       --allow-missing-template-keys=true: If true, ignore any errors in templates when a field or map key is missing in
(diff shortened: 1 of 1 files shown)


gsquared94:
cc. @sujit-kamireddy @ivanporty

docs/design_proposals/inspect-cmd.md
```diff
@@ -0,0 +1,65 @@
...
+Usage:
+  skaffold inspect [flags] [options]
+
+Use "skaffold options" for a list of global command-line options (applies to all
+commands).
+```
+
+## Implementation plan
+
+1. Add command with flag defaults and hidden.
```
tejal29, line 57: Can add these features to `skaffold diagnose` command ? The intention behind skaffold diagnose is what you have here.

docs/design_proposals/inspect-cmd.md
```diff
@@ -0,0 +1,65 @@
...
+
+  # Get all profile names from the main `skaffold.yaml` configs.
+  skaffold inspect -o jsonpath='{.configs[#].profiles[*].name}'
+
+Options:
+      --allow-missing-template-keys=true: If true, ignore any errors in templates when a field or map key is missing in
+the template. Only applies to jsonpath output formats.
+  -o, --output='yaml': Output format. One of:
+json|yaml|jsonpath.
+      --raw=false: Raw output will not apply profiles and just display the contents of the parsed files.
```
briandealwis, line 46: This would mean that the `activeProfiles` is not evaluated?  `raw` doesn't quite capture that, to my mind.  Could we be explicit with `--apply-profiles=true/false`?

docs/design_proposals/inspect-cmd.md
briandealwis: The name `diagnose` implies _identify problems_, and doesn't match this intent though.  I personally found `diagnose --yaml-only` to be a confusing combination.

It would be nice to come up with a name that can comfortably hold patching-style functions like we discussed to support IDE use though.

docs/design_proposals/inspect-cmd.md
```diff
@@ -0,0 +1,65 @@
...
+  skaffold inspect
+  
+  # Show all resolved configurations with profiles not applied.
+  skaffold inspect --raw
+  
+  # Get all named configs (modules) as a list
+  skaffold inspect -o jsonpath='{.configs[?(@.metadata.name != "")].metadata.name}'
+
+  # Get all profile names from the main `skaffold.yaml` configs.
+  skaffold inspect -o jsonpath='{.configs[#].profiles[*].name}'
```
briandealwis, line 39: This seems powerful but complex — and is tied to the underlying representation; part of the motivation was to add some distance in having to know the details of the model and resolution.

Could we support the IDE cases explicitly with subcommands:
  - `skaffold inspect profiles`: list all profiles
  - `skaffold inspect artifacts`: list all artifacts
  -  `skaffold inspect configs`: list all config names

Should these list the underlying schema objects or an intermediate step?  (E.g., what should happen if we rework the skaffold model?)

Issue #5539, opened 2021-03-12
Add events for Test phase to support re-executing tests 


PriyaModali:
Add events for Test phase to support re-executing tests when test dependency changes

PR #5573, opened 2021-03-18
Adding events for Test phase

PriyaModali:
Fixes: #5539 

**Description**
Adding events for Test phase

Skaffold currently doesn't log events for test phase progress.
With this change the tests log test progress events .

Related:
#5333

### Example
The functionality can be tested using the example in the integration tests:
`skaffold/integration/examples/test-events-test`


integration/test_events_test.go
```diff
@@ -0,0 +1,121 @@
...
+			rpcAddr := randomPort()
+
+			skaffold.Dev("--rpc-port", rpcAddr).InDir(test.testDir).WithConfig(test.config).InNs(ns.Name).RunBackground(t)
+
+			client.WaitForPodsReady(test.podName)
+
+			// Ensure we see a test is triggered in the event log
+			_, entries := apiEvents(t, rpcAddr)
+
+			waitForTestEvent(t, entries, func(e *proto.LogEntry) bool {
```
gsquared94, line 80: is `waitForTestEvent` doing anything different than `verifyTestCompletedWithEvents`?

integration/testdata/test-events/custom/main.go
```diff
@@ -0,0 +1,14 @@
+package main
```
gsquared94, line 1: reduce code duplication by merging `test-events/custom` and `test-events/structure` in the same project and having different `profiles` for the `custom` and `structure` test cases in the `skaffold.yaml`.

integration/test_events_test.go
```diff
@@ -0,0 +1,121 @@
...
+			testType:     "custom",
+		},
+		{
+			description: "test events for structure test",
+			podName:     "structure-test-events",
+			testDir:     "testdata/test-events/structure",
+			testFile:    "testdata/test-events/test",
+			config:      "skaffold.yaml",
+			testType:    "structure",
+		},
```
gsquared94, line 59: have a case that runs both. See below comment about using `profiles`

pkg/skaffold/test/custom/custom.go
```diff
@@ -83,6 +92,7 @@ func (ct *Runner) Test(ctx context.Context, out io.Writer, artifacts []build.Art
 		return cmdRunRetrieveErr(command, ct.imageName, err)
 	}
 
+	event.TestInProgress()
```
gsquared94, line 95: shouldn't `TestInProgress` be raised at the start of the method? Otherwise consider it returns on a previous line error. Then you'd have `TestFailed` event without ever having the `TestInProgress` event.

gsquared94 (changes requested):
Please address the review comments. Otherwise changes look good.

pkg/skaffold/test/custom/custom.go
PriyaModali: Moved it to start of the method.

integration/test_events_test.go
PriyaModali: Removed it.
This is just an additional check to make sure that test hasn't started yet.

integration/testdata/test-events/custom/main.go
PriyaModali: Updated test case.

integration/test_events_test.go
PriyaModali: Updated test case.

PriyaModali:
> Please address the review comments. Otherwise changes look good.

Addresses the feedback.

pkg/skaffold/event/event.go
```diff
@@ -677,6 +730,14 @@ func ResetStateOnBuild() {
 	handler.setState(newState)
 }
 
+// ResetStateOnTest resets the deploy, sync and status check state
+func ResetStateOnTest() {
+	newState := handler.getState()
+	newState.DeployState.Status = NotStarted
```
gsquared94, line 736: shouldn't this reset only `TestState`?

gsquared94: also I couldn't find a reference for `ResetStateOnTest`. Should it be called in `doDev` function in `dev.go`?

PriyaModali: Updated. Thanks for catching.

pkg/skaffold/event/event.go
```diff
@@ -677,6 +730,15 @@ func ResetStateOnBuild() {
 	handler.setState(newState)
 }
 
+// ResetStateOnTest resets the test, deploy, sync and status check state
+func ResetStateOnTest() {
+	newState := handler.getState()
+	newState.TestState.Status = NotStarted
+	newState.DeployState.Status = NotStarted
+	newState.DeployState.StatusCode = proto.StatusCode_OK
```
gsquared94, line 738: do not reset `Deploy` relates state.
```suggestion
	newState.TestState.Status = NotStarted
```

PriyaModali: Removed deploy state updates in  ResetStateOnTest().

gsquared94 (approved):
lgtm!

Issue #5479, opened 2021-03-02
Syntay to handover multiple image in "--images=: A list of pre-built images to deploy"


Miles-Ahead-Digital:
Hi,
i want to use the parameter --images= 
in a skaffold shell command to hand over 2 images. 
What is the exact syntax? I can get it only working for one...
Best regards
Stefan

skaffold deploy -p gke -i 'image1 image2' -t latest

does not work...

briandealwis:
@Miles-Ahead-Digital specify the `-i` for each image:
```
skaffold deploy -p gke -i image1 -i image2 -t latest
```

The underlying problem is that the [flags.Images type](https://github.com/GoogleContainerTools/skaffold/blob/master/cmd/skaffold/app/flags/image.go#L39-L67) implements the pflag.Value interface and not the pflag.SliceValue interface.

sladyn98:
@briandealwis I would like to give this a shot. Could I get some help on what methods would I need to implement for `pflag.SliceValue` or just the starting point for the code and I could get started thanks.

briandealwis:
@sladyn98 I'm so sorry, I missed this notification.  I think it should be as easy as implementing pflag's [SliceValue](https://github.com/spf13/pflag/blob/85dd5c8bc61cfa382fecd072378089d4e856579d/flag.go#L193-L203) interface.

seventieskid:
👍

briandealwis:
> I think it should be as easy as implementing pflag's SliceValue interface.

I should stop writing things like this.

PR #5575, opened 2021-03-18
Implement pflag slice value interface for image types

sladyn98:
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/5479

**Description**

This one fixes https://github.com/GoogleContainerTools/skaffold/issues/5479

### DIFF
--- a/cmd/skaffold/app/flags/image.go
+++ b/cmd/skaffold/app/flags/image.go
@@ -36,23 +36,18 @@ type image struct {
 	artifact *build.Artifact
 }
 
-// String Implements String() method for pflag interface and
+// String Implements String() method for pflag Slice Value interface and
 // returns a comma separated list of images.
-func (i *Images) String() string {
+func (i *Images) GetSlice() []string {
 	names := make([]string, len(i.images))
 	for i, image := range i.images {
 		names[i] = image.name
 	}
-	return strings.Join(names, ",")
-}
-
-// Usage Implements Usage() method for pflag interface
-func (i *Images) Usage() string {
-	return i.usage
+	return names
 }
 
 // Set Implements Set() method for pflag interface
-func (i *Images) Set(value string) error {
+func (i *Images) Append(value string) error {
 	a, err := convertImageToArtifact(value)
 	if err != nil {
 		return err
@@ -61,10 +56,10 @@ func (i *Images) Set(value string) error {
 	return nil
 }
 
-// Type Implements Type() method for pflag interface
-func (i *Images) Type() string {
-	return fmt.Sprintf("%T", i)
+func(i *Images) Replace([] string) error {
+
 }
+
 
 // Artifacts returns an artifact representation for the corresponding image
 func (i *Images) Artifacts() []build.Artifact {


cmd/skaffold/app/flags/image.go
```diff
@@ -61,11 +56,11 @@ func (i *Images) Set(value string) error {
 	return nil
 }
 
-// Type Implements Type() method for pflag interface
-func (i *Images) Type() string {
-	return fmt.Sprintf("%T", i)
+func(i *Images) Replace([] string) error {
+
```
sladyn98, line 60: @briandealwis I am not too sure how would this interface be implemented, would we provide an array of strings and then replace thme in the images array ?

MarlonGamez: @sladyn98 I think so, Looks like you could basically loop over the passed in data and use the `convertImageToArtifact()` function to populate a new `[]image`

cmd/skaffold/app/flags/image.go
```diff
@@ -36,23 +36,18 @@ type image struct {
 	artifact *build.Artifact
 }
 
-// String Implements String() method for pflag interface and
+// String Implements String() method for pflag Slice Value interface and
 // returns a comma separated list of images.
```
MarlonGamez, line 40:
```suggestion
// GetSlice Implements GetSlice() method for pflag SliceValue interface and
// returns a slice image names.
```

cmd/skaffold/app/flags/image.go
```diff
@@ -36,23 +36,18 @@ type image struct {
...
-	return strings.Join(names, ",")
-}
-
-// Usage Implements Usage() method for pflag interface
-func (i *Images) Usage() string {
-	return i.usage
+	return names
 }
 
 // Set Implements Set() method for pflag interface
```
MarlonGamez, line 49:
```suggestion
// Append Implements Append() method for pflag SliceValue interface
```

cmd/skaffold/app/flags/image.go
```diff
@@ -61,11 +56,11 @@ func (i *Images) Set(value string) error {
 	return nil
 }
 
-// Type Implements Type() method for pflag interface
-func (i *Images) Type() string {
-	return fmt.Sprintf("%T", i)
+func(i *Images) Replace([] string) error {
```
MarlonGamez, line 59: I would say add another function comment here

MarlonGamez (changes requested):

cmd/skaffold/app/flags/image.go
```diff
@@ -61,9 +56,16 @@ func (i *Images) Set(value string) error {
...
+func (i *Images) Replace(images []string) error {
+	newImages := NewEmptyImages("new list")
+	for _, value := range images {
+		a, err := convertImageToArtifact(value)
+		if err != nil {
+			return err
+		}
+		newImages.images = append(newImages.images, image{name: value, artifact: a})
+	}
 }
```
MarlonGamez, line 69: I don't know that we need to use `NewEmptyImages()` here. I think we can just use `newImages := make([]image)`, then iterate over images like you've done and populate newImages, then do something like `i.images = newImages` to replace the underlying data

MarlonGamez (changes requested):
This is looking pretty good, I just have some comments about the `Replace()` function

cmd/skaffold/app/flags/image.go
sladyn98: Got it :+1

briandealwis (commented):
Skaffold has special defaults handling to workaround https://github.com/spf13/pflag/issues/257 and it currently doesn't support `SliceValue`, just `Value`.  I was banging on this recently but I haven't gotten a fix in yet.  Let me see if I can whip it into shape.

sladyn98:
> Skaffold has special defaults handling to workaround [spf13/pflag#257](https://github.com/spf13/pflag/issues/257) and it currently doesn't support `SliceValue`, just `Value`. I was banging on this recently but I haven't gotten a fix in yet. Let me see if I can whip it into shape.

Sure, let me know if you need any help, till then I can help out on some other tickets :+1:

cmd/skaffold/app/flags/image.go
```diff
@@ -61,9 +56,18 @@ func (i *Images) Set(value string) error {
...
+	for _, value := range images {
+		a, err := convertImageToArtifact(value)
+		if err != nil {
+			return err
+		}
+		newImages = append(newImages, image{name:value, artifact:a})
+		i.images = newImages
+	}
+	return nil
 }
```
MarlonGamez, line 71:
```suggestion
		}
		newImages = append(newImages, image{name:value, artifact:a})
	}
	i.images = newImages
	return nil
}
```
move this assignment to after the loop, as it only needs to happen once when we're done filling `newImages`

MarlonGamez (commented):
@sladyn98 I believe this should be unblocked now and can be finished up soon!

briandealwis:
@sladyn98 hope you don't mind but I merged this with HEAD (which now has the fixes for SliceValue) and fixed up your implementation: SliceValue objects must implement Value too.

cmd/skaffold/app/flags/image.go
```diff
@@ -37,22 +37,40 @@ type image struct {
...
+// SetNil Implements SetNil() method for our Nillable interface
+func (i *Images) SetNil() error {
+	i.images = []image{}
+	return nil
+}
+
+// Set Implements Set() method for pflag interface
+func (p *Images) Set(csv string) error {
+	split := strings.Split(csv, ",")
+	return p.Replace(split)
```
briandealwis, line 59: I realized that this should be `.Append()` to maintain compatibility with our previous behaviour where we required used to provide multiple `-i` flags, one for for each image

sladyn98:
@briandealwis Hey no worries, let me know if these needs any more action, if not I can pick some more issues to solve in the repo

MarlonGamez (approved):
lgtm 👍🏼

PR #5582, opened 2021-03-19
Metric labels need to be string type

gsquared94:
Cloud monitoring doesn't support non-string type values in `labels`.

Labels parsed [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/402033a575703385b81fee3bdf703b0dd61ff9f8/exporter/metric/metric.go#L490). Their [`AsString`](https://github.com/open-telemetry/opentelemetry-go/blob/ecf65d7968225482a4d50e4955d6bc826119b49c/attribute/value.go#L149) method returns `""` for all non-string type labels.

### DIFF
--- a/pkg/skaffold/instrumentation/export.go
+++ b/pkg/skaffold/instrumentation/export.go
@@ -139,6 +139,8 @@ func createMetrics(ctx context.Context, meter skaffoldMeter) {
 	randLabel := label.String("randomizer", strconv.Itoa(rand.Intn(75000)))
 
 	m := global.Meter("skaffold")
+
+	// cloud monitoring only supports string type labels
 	labels := []label.KeyValue{
 		label.String("version", meter.Version),
 		label.String("os", meter.OS),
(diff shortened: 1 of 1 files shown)


nkubala (approved):
😞

pkg/skaffold/instrumentation/export.go
```diff
@@ -139,14 +139,16 @@ func createMetrics(ctx context.Context, meter skaffoldMeter) {
...
+
+	// cloud monitoring only supports string type labels
 	labels := []label.KeyValue{
 		label.String("version", meter.Version),
 		label.String("os", meter.OS),
 		label.String("arch", meter.Arch),
 		label.String("command", meter.Command),
 		label.String("error", meter.ErrorCode.String()),
 		label.String("platform_type", meter.PlatformType),
-		label.Int("config_count", meter.ConfigCount),
```
tejal29, line 149: Ah makes sense. You are recording it as a label.
`helmCount` is recorded as a NewMetric.

PR #5088, opened 2020-12-01
enabling using another container's network stack on build process

zedfmario:
Fixes: #5087

**Description**
Modifying docker network mode validation to accept `container:<name|id>` network identifier, allowing to reuse the network stack from an **existing** container (defined by `name` or `id`).

**User facing changes (remove if N/A)**
#### Before
For a given `skaffold.yaml`: 
```
apiVersion: skaffold/v2beta9
kind: Config
build:
  artifacts:
  - image: skaffold-example
    docker:
      network: "container:alpine"
deploy:
  kubectl:
    manifests:
      - k8s-*
```
the output was (described in the issue):
```
invalid skaffold config: artifact skaffold-example has invalid networkMode 'container:alpine'
```

##### When there is NO running container with such name
Checking that no alpine container runs in minikube:

```
> minikube ssh
                         _             _            
            _         _ ( )           ( )           
  ___ ___  (_)  ___  (_)| |/')  _   _ | |_      __  
/' _ ` _ `\| |/' _ `\| || , <  ( ) ( )| '_`\  /'__`\
| ( ) ( ) || || ( ) || || |\`\ | (_) || |_) )(  ___/
(_) (_) (_)(_)(_) (_)(_)(_) (_)`\___/'(_,__/'`\____)

$ docker ps | grep alpine
$ exit
```
Running `skaffold` with the previous `yaml` file:
```
> skaffold -f skaffold.yaml build
Generating tags...
 - skaffold-example -> skaffold-example:v1.17.0-9-gdcae2d749-dirty
Found [minikube] context, using local docker daemon.
Building [skaffold-example]...
Sending build context to Docker daemon  3.072kB
Step 1/8 : FROM golang:1.12.9-alpine3.10 as builder
 ---> e0d646523991
Step 2/8 : COPY main.go .
 ---> 7d8855ef4f5e
Step 3/8 : ARG SKAFFOLD_GO_GCFLAGS
 ---> Running in d12006f12763
Removing intermediate container d12006f12763
 ---> 476ac0f33bfa
Step 4/8 : RUN go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -o /app main.go
 ---> Running in acffd1f7b525
Removing intermediate container acffd1f7b525
unable to stream build output: No such container: alpine. Please fix the Dockerfile and try again..

```

##### When there is a running container with such name
Starting an `alpine` container within minikube:
```
> minikube ssh 
                         _             _            
            _         _ ( )           ( )           
[...]

PR #5589, opened 2021-03-22
Adapting validation for docker container network mode to include ENV_VARS

zedfmario:
**Description**

In PR #5088 I enabled the possibility of reusing the network stack from another docker container. 
That mode required either a _name_ or a _docker id_ to check what container you want to reuse the network stack from.

The configuration at that point required the exact _name_ or _id_. It's quite common in the docker environment, when referencing by _id_, to use the very first 12 chars of the container name, as those are the ones shown when `docker ps` is executed.
Example:
```
> docker ps
CONTAINER ID        IMAGE               COMMAND               CREATED             STATUS              PORTS               NAMES
7765077b5d37        alpine              "tail -f /dev/null"   About an hour ago   Up About an hour                        alpine
```
This PR finds docker containers by ids using HasPrefix instead of equal. It also adapts the validation for docker container network mode to include ENV_VARS. It will automatically translate the ENV_VAR with the `os.Environment` variable.

from now on, it is possible to write the container name as follows:

```
build:
  docker:
    network: "container:{{.ENV_VAR}}"
```
where  `ENV_VAR` could be `7765077b5d37`, for example.


pkg/skaffold/schema/validation/validation.go
```diff
@@ -219,6 +219,72 @@ func validateUniqueDependencyAliases(artifacts []*latest.Artifact) (errs []error
...
+	if strings.HasPrefix(strings.ToLower(mode), "container:") {
+		// Up to this point, we know that we can strip until the colon symbol and keep the second part
+		// this is helpful in case someone sends container not in lowercase
+		maybeID := strings.SplitN(mode, ":", 2)[1]
+		id, err := util.ExpandEnvTemplate(maybeID, map[string]string{})
+		if err != nil {
+			return "", sErrors.NewError(err,
+				proto.ActionableErr{
+					Message: fmt.Sprintf("unable to parse container name %s: %s", mode, err),
+					ErrCode: proto.StatusCode_TEST_CUSTOM_CMD_PARSE_ERR,
```
tejal29, line 234: shd we reuse the  `StatusCode_INIT_DOCKER_NETWORK_INVALID_CONTAINER_NAME`  error code and the suggestion code should be in regards to fixing docker container name
I understand you want to define a new error code for the env template parse error. However the error code you are using right now is specific to "TEST" phase.

Either use the  `StatusCode_INIT_DOCKER_NETWORK_INVALID_CONTAINER_NAME`  or define a new one in `INIT` phase e.g. `StatusCode_INIT_DOCKER_NETWORK_PARSE_ERR`
You can reuse the suggestion code.

pkg/skaffold/schema/validation/validation.go
```diff
@@ -219,6 +219,72 @@ func validateUniqueDependencyAliases(artifacts []*latest.Artifact) (errs []error
...
+					},
+				})
+		}
+		return id, nil
+	}
+	errMsg := fmt.Sprintf("extracting container name from a non valid container network mode '%s'", mode)
+	return "", sErrors.NewError(fmt.Errorf(errMsg),
+		proto.ActionableErr{
+			Message: errMsg,
+			ErrCode: proto.StatusCode_TEST_CUSTOM_CMD_PARSE_ERR,
```
tejal29, line 249: Reuse the invalid container name error code and suggestion or similar to above comment.

pkg/skaffold/schema/validation/validation.go
```diff
@@ -229,23 +295,11 @@ func validateDockerNetworkMode(artifacts []*latest.Artifact) (errs []error) {
 		if mode == "none" || mode == "bridge" || mode == "host" {
 			continue
 		}
-		containerRegExp := regexp.MustCompile("^container:[a-zA-Z0-9][a-zA-Z0-9_.-]*$")
-		if containerRegExp.MatchString(mode) {
+		networkModeErr := validateDockerNetworkModeExpression(a.ImageName, a.DockerArtifact.NetworkMode)
```
tejal29, line 298: shd this be mode
```suggestion
		networkModeErr := validateDockerNetworkModeExpression(a.ImageName, mode)
```

tejal29 (commented):
few nits.

pkg/skaffold/schema/validation/validation.go
zedfmario: As far as I understand we allow passing `Container:<id|name>` (note the capital C) in the _skaffold.yaml_. I was trying to avoid executing `strings.ToLower` since it changes the container name (if it has the container with capital letters, what's allowed). That's why I'm using `a.DockerArtifact.NetworkMode`. In the current code, there's a bug because of this explanation...

pkg/skaffold/schema/validation/validation.go
zedfmario: I've already added a couple of new status and suggestion codes

pkg/skaffold/schema/validation/validation.go
zedfmario: Same here

tejal29 (approved):

Issue #5590, opened 2021-03-22
port-forwards should not allocate system ports


briandealwis:
Our port-forwarding handling for services requests the same local port as the service port:
https://github.com/GoogleContainerTools/skaffold/blob/66fe1c7e3da3fce3ec3d51b2c67fef766882fd2e/pkg/skaffold/kubernetes/portforward/resource_forwarder.go#L144

On Windows systems, where system ports can be allocated by ordinary processes, this means that [a service can allocate ports 80 or 443](https://github.com/GoogleContainerTools/skaffold/pull/5554#issuecomment-803270340) which is generally not desirable.  This issue is not normally seen on *nix systems as allocating system ports require running as root.

This can lead to puzzling behaviour as Skaffold's port-forwards listen only on localhost, which allows co-existing with another process that listens on IP_ADDRANY on the same port (such as the k8s cluster's ingress port!).

As a workaround, users can set up explicit port-forwards for the corresponding services.

### Information

- Operating system: Windows

### Steps to reproduce the behavior

1. Alter Skaffold's `example/typescript`'s `k8s/deployment.yaml` to use port 80
2. `skaffold dev --port-forward`
3. See Skaffold port-forward to port 80

alisters:
Thanks for raising this and the speedy change. This, along with the helm debugging support you've added over time will bring us close to "git pull, skaffold run" across a set of disparate enterprise services and charts. Much appreciated.

PR #5591, opened 2021-03-22
Modify util.GetAvailablePort to produce less confusing port allocations

briandealwis:
Fixes: #5590 

**Description**
This PR modifies our `GetAvailablePort()` function (and related functions) to:
- Never attempt to allocate a system ports (port 1-1023) to avoid confusing results such as allocating ports 80 or 443.
- Allocate ports using INADDR_ANY (0.0.0.0) as it is possible to listen on a port for a specific interface even when another process is listening on INADDR_ANY with the same port.

- port-forwarding no longer allocates ports in the 1-1023 range

### DIFF
--- a/pkg/skaffold/util/port.go
+++ b/pkg/skaffold/util/port.go
@@ -87,34 +87,37 @@ func (f *PortSet) List() []int {
 	return list
 }
 
+// GetAvailablePort returns an available port that is near the requested port when possible.
 // First, check if the provided port is available on the specified address. If so, use it.
 // If not, check if any of the next 10 subsequent ports are available.
 // If not, check if any of ports 4503-4533 are available.
 // If not, return a random port, which hopefully won't collide with any future containers
-
-// See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt,
-func GetAvailablePort(address string, port int, usedPorts *PortSet) int {
-	if getPortIfAvailable(address, port, usedPorts) {
-		return port
-	}
-
-	// try the next 10 ports after the provided one
-	for i := 0; i < 10; i++ {
-		port++
-		if getPortIfAvailable(address, port, usedPorts) {
-			logrus.Debugf("found open port: %d", port)
+//
+// See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt
+func GetAvailablePort(port int, usedPorts *PortSet) int {
+	if port > 0 {
+		if getPortIfAvailable(port, usedPorts) {
 			return port
+		}
+
+		// try the next 10 ports after the provided one
+		for i := 0; i < 10; i++ {
+			port++
+			if getPortIfAvailable(port, usedPorts) {
+				logrus.Debugf("found open port: %d", port)
+				return port
+			}
 		}
 	}
 
 	for port = 4503; port <= 4533; port++ {
-		if getPortIfAvailable(address, port, usedPorts) {
+		if getPortIfAvailable(port, usedPorts) {
 			logrus.Debugf("found open port: %d", port)
 			return port
 		}
 	}
 
-	l, err := net.Listen("tcp", fmt.Sprintf("%s:0", address))
+	l, err := net.Listen("tcp", ":0")
 	if err != nil {
 		return -1
 	}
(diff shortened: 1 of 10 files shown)


pkg/skaffold/util/port_test.go
```diff
@@ -75,3 +75,13 @@ func TestGetAvailablePort(t *testing.T) {
 		t.Errorf("available port (%d) couldn't be used: %w", port, err)
 	}
 }
+
+func TestSystemPorts(t *testing.T) {
+	ps := PortSet{}
+	ps.lock.Lock() // cause deadlock: getPortIfAvailable should never check the PortSet
```
gsquared94, line 81: not sure why this is required. There's already an error if `getPortIfAvailable` returns true. Also if I deliberately introduce a bug then this deadlock causes the test to hang till timeout.

briandealwis: This is a whitebox test that verifies that Skaffold should never check the set-ports-list for system ports.  The intent is that if this test hangs (e.g., someone violated this invariant), someone will quickly discover that this invariant has been violated.

Let me update this comment to be more clear.

gsquared94: but won't that get tested anyway in `t.Errorf(...)`? Also can't ps just be `nil` and you get the same thing?

briandealwis: I really do want to explicitly verify that `GetPortIfAvailable()` never considers ports 1–1023 separately from the provided `PortSet`.  The `PortSet` uses locks to protect the data structure, and there's no reason to check it for this port range.

pkg/skaffold/util/port.go
```diff
@@ -87,34 +87,35 @@ func (f *PortSet) List() []int {
...
+func GetAvailablePort(port int, usedPorts *PortSet) int {
+	if getPortIfAvailable(port, usedPorts) {
 		return port
 	}
 
 	// try the next 10 ports after the provided one
 	for i := 0; i < 10; i++ {
 		port++
-		if getPortIfAvailable(address, port, usedPorts) {
+		if getPortIfAvailable(port, usedPorts) {
```
tejal29, line 105: Shd we skip iterating over next 10 ports if port number is less than 1022?

briandealwis:
Need to revisit this as it can be necessary to map to ports 80 or 443 to enable testing of OAuth.  I think the way forward here is to allow explicit user-defined port-forwards to specific ports.

briandealwis:
Verified that explicit local ports on user-define ports in `skaffold.yaml` do not use `util.GetAvailablePort()`:
```diff
--- examples/microservices/skaffold.yaml
+++ examples/microservices/skaffold.yaml
@@ -23,7 +23,7 @@ portForward:
   - resourceType: deployment
     resourceName: leeroy-web
     port: 8080
-    localPort: 9000
+    localPort: 80
   - resourceType: deployment
     resourceName: leeroy-app
     port: http
```

And the result:
```
Deployments stabilized in 3.995 seconds
Press Ctrl+C to exit
Watching for changes...
[leeroy-app] 2021/04/12 20:54:05 leeroy app server ready
[leeroy-web] 2021/04/12 20:54:05 leeroy web server ready
port forwarding deployment-leeroy-web-default-8080 got terminated: output: Unable to listen on port 80: Listeners failed to create with the following errors: [unable to create listener: Error listen tcp4 127.0.0.1:80: bind: permission denied unable to create listener: Error listen tcp6 [::1]:80: bind: permission denied]
error: unable to listen on any of the requested ports: [{80 8080}]

Port forwarding deployment/leeroy-app in namespace default, remote port http -> 127.0.0.1:9001
```

But it would be nice if service ports did map to corresponding local ports if they were available:
```
$ skaffold dev --port-forward=services
...
Press Ctrl+C to exit
Watching for changes...
[leeroy-app] 2021/04/12 20:56:42 leeroy app server ready
[leeroy-web] 2021/04/12 20:56:42 leeroy web server ready
Port forwarding service/leeroy-app in namespace default, remote port 50051 -> 127.0.0.1:4503
Port forwarding service/leeroy-web in namespace default, remote port 8080 -> 127.0.0.1:4504
```

pkg/skaffold/util/port.go
```diff
@@ -87,34 +87,37 @@ func (f *PortSet) List() []int {
...
-
-	// try the next 10 ports after the provided one
-	for i := 0; i < 10; i++ {
-		port++
-		if getPortIfAvailable(address, port, usedPorts) {
-			logrus.Debugf("found open port: %d", port)
+//
+// See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt
+func GetAvailablePort(port int, usedPorts *PortSet) int {
+	if port > 0 {
```
MarlonGamez, line 98: Sorry if this is a dumb question: Where are we preventing trying to allocate 1-1023?

briandealwis:
Closing this: the system ports was an unnecessary diversion

pkg/skaffold/util/port.go
briandealwis: See https://github.com/GoogleContainerTools/skaffold/pull/5554#issuecomment-803270340 for the motivation.  We had another group ask as they wanted to control which service is bound to port 443 for https.

Issue #5612, opened 2021-03-29
State.ForwardedPorts may be nil


briandealwis:
I had a panic with Skaffold in Cloud Shell, running Skaffold version: v1.20.0:

```
Deployments stabilized in 27.762 seconds
Press Ctrl+C to exit
Streaming logs from pod: nodetodo-deployment-766cb4c7d8-tkw6k container: nodetodo
Streaming logs from pod: nodetodo-deployment-766cb4c7d8-tkw6k container: init-db-ready
files modified: [.git/index]
Changed file .git/index does not match any sync pattern. Skipping sync
Not watching for changes...
[nodetodo] 
[nodetodo] > nodejs-todo-app@1.0.0 dev /opt/backend
[nodetodo] > node app.js
[nodetodo] 
[nodetodo] 0: /usr/local/bin/node
[nodetodo] 1: /opt/backend/app.js
[nodetodo] service host and port 10.100.169.59 27017
[nodetodo] Username and password username password
[nodetodo] To connect at mongodb://username:password@10.100.169.59:27017/admin
[nodetodo] Your Listening to the port 3000
[init-db-ready] Waiting for mongodb to start...
[init-db-ready] Waiting for connection for 2 sec.
[init-db-ready] Waiting for connection for 2 sec.
Port forwarding service/mongodb-service in namespace default, remote port 27017 -> address 127.0.0.1 port 27017
panic: assignment to entry in nil map

goroutine 8 [running]:
github.com/GoogleContainerTools/skaffold/pkg/skaffold/event.(*eventHandler).handleExec(0xc00052e0c0, 0xc000b3c000, 0xc000d4e080)
/skaffold/pkg/skaffold/event/event.go:582 +0x10bb
github.com/GoogleContainerTools/skaffold/pkg/skaffold/event.newHandler.func1(0xc00052e0c0)
/skaffold/pkg/skaffold/event/event.go:63 +0x37
created by github.com/GoogleContainerTools/skaffold/pkg/skaffold/event.newHandler
/skaffold/pkg/skaffold/event/event.go:57
+0x90
Exited with code 2.
Cleaning up...
 - deployment.apps "mongodb-deployment" deleted
 - service "mongodb-service" deleted
 - secret "mongodb-secret" deleted
 - deployment.apps "nodetodo-deployment" deleted
 - service "nodetodo-external" deleted
 - configmap "todo-config" deleted
```

`event.go` line 582 in v1.20.0:
[...]

PR #5615, opened 2021-03-29
Handle State.ForwardedPorts == nil

briandealwis:
Fixes: #5612

**Description**
`pkg/skaffold/event.getState()` uses JSON marshalling and unmarshalling to create a deep copy of the event state object.  As `State.ForwardedPorts` is marked as `omitempty`:
https://github.com/GoogleContainerTools/skaffold/blob/834d5eaba317b24195f2cc09dc3a850890d1a1bc/proto/v1/skaffold.pb.go#L419
then `getState()` may return an object with a nil `ForwardPorts` map.

This change fixes the handling of `PortEvent`s to deal with a nil `State.ForwardPorts` map.  It also adds a test to trigger the panic seen in #5612.


gsquared94 (approved):
lgtm

PR #5622, opened 2021-04-01
Add event API v2 server handler

MarlonGamez:
**Related**: #5422 , #5368 

**Description**
Adds `pkg/skaffold/server/v2` package which contains a new struct that handles the endpoints for the v2 API. Users will now be able to hit both the v1 and v2 endpoints through the running gRPC server.

Most of the server definition is copied from `pkg/skaffold/server` files, and the functions added to `pkg/skaffold/event/v2/event.go` are straight from `pkg/skaffold/event/event.go`

Emit protos on this server

### DIFF
--- a/pkg/skaffold/event/v2/event.go
+++ b/pkg/skaffold/event/v2/event.go
@@ -167,6 +167,10 @@ func emptyStateWithArtifacts(builds map[string]string, metadata *proto.Metadata,
 			AutoTrigger: autoBuild,
 			StatusCode:  proto.StatusCode_OK,
 		},
+		TestState: &proto.TestState{
+			Status:     NotStarted,
+			StatusCode: proto.StatusCode_OK,
+		},
 		DeployState: &proto.DeployState{
 			Status:      NotStarted,
 			AutoTrigger: autoDeploy,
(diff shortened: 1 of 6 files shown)


MarlonGamez:
Still need to add tests, but putting this up here in case anyone wants to take a look :)

pkg/skaffold/event/v2/event.go
```diff
@@ -190,6 +241,19 @@ func emptyStatusCheckState() *proto.StatusCheckState {
 	}
 }
 
+func AutoTriggerDiff(name string, val bool) (bool, error) {
+	switch name {
+	case "build":
```
tejal29, line 246: we could reuse `pkg/errors.Phase` type and add `sync` to it?

tejal29 (approved):

pkg/skaffold/event/v2/event.go
MarlonGamez: Good point! I'll update this

MarlonGamez: https://github.com/GoogleContainerTools/skaffold/pull/5625

PR #5623, opened 2021-04-01
Add `BuilderContext` interface to encapsulate `RunContext` and provide additional builder constructor requirements.

gsquared94:
Refactor builder constructor to take a `BuilderContext` instead of only the `RunContext` to provide additional builder requirements.

This will simplify https://github.com/GoogleContainerTools/skaffold/pull/5614

This PR doesn't change any existing behavior.

### DIFF
--- a/pkg/skaffold/build/local/types.go
+++ b/pkg/skaffold/build/local/types.go
@@ -69,14 +69,19 @@ type Config interface {
 	Muted() config.Muted
 }
 
+type BuilderContext interface {
+	Config
+	ArtifactStore() build.ArtifactStore
+}
+
 // NewBuilder returns an new instance of a local Builder.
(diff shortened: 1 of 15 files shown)


pkg/skaffold/build/local/types.go
```diff
@@ -90,26 +95,23 @@ func NewBuilder(cfg Config, buildCfg *latest.LocalBuild) (*Builder, error) {
 
 	return &Builder{
 		local:              *buildCfg,
-		cfg:                cfg,
-		kubeContext:        cfg.GetKubeContext(),
+		cfg:                bCtx,
```
briandealwis, line 98: It seems odd to be pulling out the different values from the config/context (kubecontext, skiptests, mode, etc) when we could just retain the config/context object.  I suppose ultimately this config object is the RunContext, which has mutable fields.

briandealwis (approved):

PR #5625, opened 2021-04-02
rename errors.FileSync errors.Sync

MarlonGamez:
**Description**
Simple refactor, just renames and changes value of `errors.FileSync` to `errors.Sync`. Also modifies some `pkg/skaffold/event` and `pkg/skaffold/server` funcs to use `errors.Phase`

### DIFF
--- a/pkg/skaffold/event/v2/event.go
+++ b/pkg/skaffold/event/v2/event.go
@@ -241,16 +241,16 @@ func emptyStatusCheckState() *proto.StatusCheckState {
 	}
 }
 
-func AutoTriggerDiff(name string, val bool) (bool, error) {
-	switch name {
-	case "build":
+func AutoTriggerDiff(phase sErrors.Phase, val bool) (bool, error) {
+	switch phase {
+	case sErrors.Build:
 		return val != handler.getState().BuildState.AutoTrigger, nil
-	case "sync":
+	case sErrors.Sync:
 		return val != handler.getState().FileSyncState.AutoTrigger, nil
-	case "deploy":
+	case sErrors.Deploy:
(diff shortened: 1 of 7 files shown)


pkg/skaffold/event/v2/event.go
```diff
@@ -241,16 +241,16 @@ func emptyStatusCheckState() *proto.StatusCheckState {
 	}
 }
 
-func AutoTriggerDiff(name string, val bool) (bool, error) {
-	switch name {
-	case "build":
+func AutoTriggerDiff(phase sErrors.Phase, val bool) (bool, error) {
+	switch phase {
+	case sErrors.Build:
```
gsquared94, line 246: seems a bit counterintuitive that the the event `phases` are imported from the error package.

MarlonGamez: I agree and have been wondering where a better place to house these would be. Since phases are a broader concept of skaffold, maybe they should be in something more general?

gsquared94: would it be better to move the `Phase` definition into the `proto` package?

MarlonGamez: I think I'm not sure on that just because they aren't specific to anything proto related. I would say it almost seems like these should just be consts in the `constants` package? 

@tejal29 any reason for if these should stay in the `errors` package?

Either way, I think this could be handled in another follow up refactor if you're okay with that

gsquared94 (approved):

pkg/skaffold/event/v2/event.go
gsquared94: `Phase` could be an enum in `enums.proto`

PR #5631, opened 2021-04-06
Bring survey prompt back to 10 days and every 90 days.

tejal29:
Revert #4626 and bring back survey prompts to normal cadency 
1) Show one every 90 days
2) Keep prompting after every 10 days until a user fills.

### DIFF
--- a/pkg/skaffold/config/util.go
+++ b/pkg/skaffold/config/util.go
@@ -38,6 +38,8 @@ import (
 const (
 	defaultConfigDir  = ".skaffold"
 	defaultConfigFile = "config"
+	tenDays           = time.Hour * 24 * 10
+	threeMonths       = time.Hour * 24 * 90
 )
 
 var (
@@ -321,7 +323,7 @@ func recentlyPromptedOrTaken(cfg *ContextConfig) bool {
 	if cfg == nil || cfg.Survey == nil {
 		return false
 	}
-	return lessThan(cfg.Survey.LastTaken, 365*24*time.Hour) || lessThan(cfg.Survey.LastPrompted, 60*24*time.Hour)
+	return lessThan(cfg.Survey.LastTaken, threeMonths) || lessThan(cfg.Survey.LastPrompted, tenDays)
 }
 
 func lessThan(date string, duration time.Duration) bool {
--- a/pkg/skaffold/config/util.go
+++ b/pkg/skaffold/config/util.go
@@ -38,8 +38,6 @@ import (
 const (
 	defaultConfigDir  = ".skaffold"
 	defaultConfigFile = "config"
-	tenDays           = time.Hour * 24 * 10
-	threeMonths       = time.Hour * 24 * 90
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/skaffold/config/util.go
```diff
@@ -321,7 +321,7 @@ func recentlyPromptedOrTaken(cfg *ContextConfig) bool {
 	if cfg == nil || cfg.Survey == nil {
 		return false
 	}
-	return lessThan(cfg.Survey.LastTaken, 365*24*time.Hour) || lessThan(cfg.Survey.LastPrompted, 60*24*time.Hour)
+	return lessThan(cfg.Survey.LastTaken, 90*24*time.Hour) || lessThan(cfg.Survey.LastPrompted, 10*24*time.Hour)
```
gsquared94, line 324: Not that I think it's a problem but if most of our users download the latest release around the same day then these reminders will be in sync 😅 Should we have a ( +- random number of days) to introduce slight disorder?

briandealwis: I doubt our survey traffic will be noticed by Google Forms :-)

briandealwis (approved):

pkg/skaffold/config/util.go
gsquared94: I wasn't concerned about traffic, more about boosting some bias in the responses. Say the first release with this has some other bug, and a lot of feedback comes in on the same day around that bug. You can see this cycle repeating where a lot of feedback is clustered on specific dates. Though even that probably isn't significant at our scale.

briandealwis: That's a very good point.  Though we've never issued a dot release for bug fixes within a day or so.
- 1.13.0 -> 1.13.1: Jul 30 -> Aug 4
- 1.13.1 -> 1.13.2: Aug 4 -> Aug 20
- 1.17.0 -> 1.17.1: Nov 23 -> Dec 1 
- 1.17.1 -> 1.17.1: Dec 1 -> Dec 7

gsquared94: In that case consider the example that we have 10 diligent users who respond to the survey as soon as they see the prompt. They'll be reprompted on the exact date 3 months later and that cycle continues. Now we only have feedback for those specific skaffold versions and very few in between. 🤷🏾‍♂️ 

None of that might be statistically significant and I'm ok with the change though.

gsquared94: basically everyone will see this prompt as soon as they download the new binary (assuming most users don't submit feedback). Should there be a first prompt offset of random days between (0 - 15) so that it's more evened out?

tejal29: I did consider/mention this in  the Survey DD where we should show the prompt X days after download.
There is no clear way to distinguishing a new user Vs an old user upgrading to a version. (Note, feedbacks are not associated with version)
 
I feel a new user would probably ignore the survey prompt first tine and the next one should show  after 10 days at which he point they may choose to respond. 

If its a not small change in code, i would not mind adding it in a separate PR. However i feel we might over engineer when most times a new user will ignore.

PR #5632, opened 2021-04-06
Flag defaults should support pflag SliceValue

briandealwis:
Unblocks #5575 and #5554

**Description**
This PR brings in support to our [flag-defaults handling](http://github.com/GoogleContainerTools/skaffold/blob/master/cmd/skaffold/app/cmd/flags.go) add-on to [`spl13/pflag`](https://github.com/spf13/pflag) to handle value objects that implement [pflag' SliceValue](https://github.com/spf13/pflag/blob/master/flag.go#L193).  

One thing to note is that pflag's `Var*` methods, which are used to provide a value object, does not take a default value and instead assumes the value object is already set to its default value.  This assumption doesn't hold with our flags defaults, and so we must explicitly set the defaults.

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -21,6 +21,7 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


MarlonGamez (approved):
looks good to me 👍🏼

Issue #5530, opened 2021-03-11
Changes to test-specific dependencies should re-run the test they are associated with


PriyaModali:
Changes to the test dependencies should re-run just the individual test that the dependencies are defined on and not all the tests.

PriyaModali:
Implemented.

PriyaModali:
Accidentally closed.

PR #5633, opened 2021-04-06
Refactor `skaffold test` implementation

gsquared94:
This PR makes following changes to the `skaffold test` implementation:
- rename `runner` interface to `ImageTester` and change the signature from accepting a slice of artifacts to just taking the image tag that needs to be tested.
- Change `FullTester` from running all the known testers and later skip for images that are not rebuilt, to only iterate on the newly build slice of images.

This removes redundant tester invocations which are getting skipped anyways. Also will simplify fixing https://github.com/GoogleContainerTools/skaffold/issues/5530

### DIFF
--- a/pkg/skaffold/test/custom/custom.go
+++ b/pkg/skaffold/test/custom/custom.go
@@ -26,13 +26,10 @@ import (
 	"runtime"
 	"time"
 
-	"github.com/sirupsen/logrus"
-
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/list"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker"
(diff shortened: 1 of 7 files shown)


pkg/skaffold/test/custom/custom.go
```diff
@@ -60,17 +57,17 @@ func New(cfg docker.Config, imageName string, wd string, ct latest.CustomTest) (
 }
 
 // Test is the entrypoint for running custom tests
-func (ct *Runner) Test(ctx context.Context, out io.Writer, artifacts []build.Artifact) error {
+func (ct *Runner) Test(ctx context.Context, out io.Writer, imageTag string) error {
 	event.TestInProgress()
-	if err := ct.runCustomTest(ctx, out, artifacts); err != nil {
+	if err := ct.runCustomTest(ctx, out, imageTag); err != nil {
```
tejal29, line 62: Maybe needs a follow up later or /cc @marlon-gamez to fix this in V2.
Add the artifact name for the test in progress event.

tejal29 (approved):

PR #5635, opened 2021-04-06
only explicitly set flag defaults for flags that have command specific defaults

MarlonGamez:
Fixes: #5588  

**Description**
Changes our flag parsing logic to only explicitly update the defaults of flags who have different defaults defined for different commands, since the pflag library should already nicely handle setting defaults for those with a universal default.

Calling `flag.Value.Set()` for the `--default-repo` flag was causing the underlying value to be represented as an empty string, when we really want it to be a nil pointer. This fixes that issue, however if there ever were a case where we defined defaults for this flag based on the user's command, I believe this could arise again for this flag.

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -558,11 +558,12 @@ func ResetFlagDefaults(cmd *cobra.Command, flags []*Flag) {
 				if d, present := fl.DefValuePerCommand[cmd.Use]; present {
 					defValue = d
 				}
-			}
-			if sv, ok := flag.Value.(pflag.SliceValue); ok {
-				reflect.ValueOf(sv).MethodByName("Replace").Call(reflectValueOf([]interface{}{defValue}))
-			} else {
-				flag.Value.Set(fmt.Sprintf("%v", defValue))
+
+				if sv, ok := flag.Value.(pflag.SliceValue); ok {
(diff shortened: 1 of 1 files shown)


tejal29:
Great find!! Can we add an integration test for this?

briandealwis:
This isn’t enough unfortunately. The problem is that pflag bashes the pointed to value each time the pointer is provided, and so the last default  value set “wins”.   So we need to bash the value for the command to be processed.

I think the right answer here is to define another `pflag.Value`-like interface for these `nil`able values. If the flags’ default value is nil, then we’d call a `Nil()` instead of calling `Value.Set(string)`. It better yet, an `ObjectValue` with a `Set(interface{}) error` that takes an object to be set, which can include `nil`.

MarlonGamez:
@briandealwis I'm not completely sure how this would work, as we're still using pflag objects when iterating over the flags and have to comply with the `pflag.Value` interface right?

My idea is to just add something like `NilDefault` field to our own `Flag` definition, then when running through `ResetFlagDefaults()`, only call `flag.Value.Set()` if that field is `false`

briandealwis:
The short answer is: no we don’t have to use just Value or SliceValue. The critical case is where we process 

1. a command with a default, then 
2. a command without a command-specific default, then 
3. a command with a default.

They all need to set the value.

https://github.com/GoogleContainerTools/skaffold/compare/master...briandealwis:nillable

It’s best to be working against #5632

PR #5646, opened 2021-04-08
fix concurrency issue in multi-config

### LINKED ISSUE #5645: Build concurrency setting not applied correctly when one config uses only imported artifacts
Issue created from [Slack](https://kubernetes.slack.com/archives/CABQMSZA6/p1617799734103200) 

### Repro
In the [`multi-config-microservices`](https://github.com/GoogleContainerTools/skaffold/tree/master/examples/multi-config-microservices) example, if we modify the three component [`base/skaffold.yaml`](https://github.com/GoogleContainerTools/skaffold/blob/master/examples/multi-config-microservices/base/skaffold.yaml), [`leeroy-app/skaffold.yaml`](https://github.com/GoogleContainerTools/skaffold/blob/master/examples/multi-config-microservices/leeroy-app/skaffold.yaml) and [`leeroy-web/skaffold.yaml`](https://github.com/GoogleContainerTools/skaffold/blob/master/examples/multi-config-microservices/leeroy-web/skaffold.yaml) with the `concurrency` setting:
```
local:
  concurrency: 3
```
we expect that after `base` is built, `leeroy-web` and `leeroy-app` will be built parallely. However this doesn't happen and all of them are built sequentially.

This is because the root [`skaffold.yaml`](https://github.com/GoogleContainerTools/skaffold/blob/master/examples/multi-config-microservices/skaffold.yaml) doesn't define any artifacts or build envs. So Skaffold adds a default local build env with concurrency of `1` to it. 
Finally in the [`BuilderMux`](https://github.com/GoogleContainerTools/skaffold/blob/cc262b48867959858e592f76674f8ea0f37e0d8e/pkg/skaffold/build/builder_mux.go#L47-L63) we select the lowest concurrency of all the configs which comes to this `1`.

 cc. @briandealwis

gsquared94:
Fixes #5645 

- Do not set a default concurrency for build environments if there are no artifacts to build. 
- Can be verified on the repro of #5645

### DIFF
--- a/pkg/skaffold/build/builder_mux.go
+++ b/pkg/skaffold/build/builder_mux.go
@@ -20,6 +20,9 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"reflect"
+
+	"github.com/sirupsen/logrus"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/graph"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
@@ -45,7 +48,7 @@ func NewBuilderMux(cfg Config, store ArtifactStore, builder func(p latest.Pipeli
 	m := make(map[string]PipelineBuilder)
 	var pb []PipelineBuilder
 	minConcurrency := -1
-	for _, p := range pipelines {
+	for i, p := range pipelines {
 		b, err := builder(p)
 		if err != nil {
 			return nil, fmt.Errorf("creating builder: %w", err)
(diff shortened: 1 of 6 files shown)


pkg/skaffold/schema/versions_test.go
```diff
@@ -482,10 +482,13 @@ func format(t *testutil.T, configs []string, apiVersions []string) string {
...
 func withLocalBuild(ops ...func(*latest.BuildConfig)) func(*latest.SkaffoldConfig) {
 	return func(cfg *latest.SkaffoldConfig) {
-		b := latest.BuildConfig{BuildType: latest.BuildType{LocalBuild: &latest.LocalBuild{Concurrency: &constants.DefaultLocalConcurrency}}}
+		b := latest.BuildConfig{BuildType: latest.BuildType{LocalBuild: &latest.LocalBuild{}}}
 		for _, op := range ops {
 			op(&b)
 		}
+		if len(b.Artifacts) > 0 {
+			b.LocalBuild.Concurrency = &constants.DefaultLocalConcurrency
+		}
```
briandealwis, line 491: What if we instead pushed this to the local builder itself, so that it determined the default concurrency at build time?

(And isn't this duplicating what is being done in the call to `withLocalBuild()` in defaults.go?)

gsquared94: this is a function in the test file that sets up the _expected_ config

gsquared94: thought about pushing this to `local.Builder` but it doesn't maintain the `[]*latest.Artifact` slice and it seemed more convoluted to first set the default to 1 but later ignore it in the method.

pkg/skaffold/build/builder_mux.go
```diff
@@ -57,7 +57,7 @@ func NewBuilderMux(cfg Config, store ArtifactStore, builder func(p latest.Pipeli
 		concurrency := b.Concurrency()
 		if minConcurrency < 0 {
 			minConcurrency = concurrency
-		} else if concurrency > 0 && concurrency < minConcurrency {
+		} else if concurrency > 0 && (minConcurrency == 0 || concurrency < minConcurrency) {
```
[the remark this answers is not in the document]
briandealwis: That's awkward and would be nice to fix as it can be hard to track down errors when dealing with multiple configs.

pkg/skaffold/schema/versions_test.go
briandealwis: Couldn't we just change this method in the local builder to return 1?
https://github.com/GoogleContainerTools/skaffold/blob/89b83154e58d7982d7db510a70f273971206c27a/pkg/skaffold/build/local/local.go#L60-L65

gsquared94: If we don't set it in `default.Set` then we'll need a reference to the `[]*latest.Artifact` per config in `local.Builder` to decide if it should return 0 or 1 by default (0 if len(artifacts) == 0, otherwise 1). That plumbing seems more convoluted that this fix.

briandealwis: Oh I didn't clue in that this definition was in a test — I thought it was the implementation of `withLocalBuilder` in pkg/skaffold/schema/defaults/ 🤦

briandealwis (approved):

PR #5648, opened 2021-04-08
Fix setting helm `--setFiles` for Windows

### LINKED ISSUE #5595: [windows] Multiconfig skaffold bug with `helm.setFiles`
### Actual behavior
In the multi configuration repository with linked skaffold.yaml's + helm an error is currently thrown under windows that a file used in helm.setFiles is not known because the slashes are missing.

### Expected behavior
It should work like under macOS/linux or if you go directly into the skaffold.yaml directory and let skaffold build the environment i.e. without errors and with the correct file path links.

Is currently only a problem under windows in combination with the multiconfig feature and (helm.setFiles)functionality
Currently don't know if it occurs with the values files as well, you could check this when solving the issue

### Information

- Skaffold version: 1.21.0
- Operating system: Windows

- Contents of skaffold.yaml:

this is the path:
https://github.com/ReneWerner87/skaffoldHelmMultiRepo/blob/master/repositories/firstService/skaffold.yaml#L34
```yaml
            setFiles:
              envFile: ".env.development"
```

1. checkout my example repository
https://github.com/ReneWerner87/skaffoldHelmMultiRepo

2. run in powershell
`skaffold run -p dev`

Screens with the working and not working operation systems:
Mac:
https://raw.githubusercontent.com/ReneWerner87/skaffoldHelmMultiRepo/master/macosWorking.png

Windows:
https://raw.githubusercontent.com/ReneWerner87/skaffoldHelmMultiRepo/master/windowsIssue.png

Example of the ERROR Message:
```shell
Error: failed parsing --set-file data: open C:UsersdeveloperDesktopreposgco.clusterrepositoriesgco-serviceconfigproduct.env: The system cannot find the file specified.
exiting dev mode because first deploy failed: install: exit status 1
```

gsquared94:
Fixes #5595 

Can be tested against the repro provided [here](https://github.com/GoogleContainerTools/skaffold/issues/5595#issue-839866092)

### DIFF
--- a/pkg/skaffold/deploy/helm/args.go
+++ b/pkg/skaffold/deploy/helm/args.go
@@ -246,7 +246,9 @@ func sanitizeFilePath(s string, isWindowsOS bool) string {
 		}
 		b = append(b, s[i])
 	}
-
+	if slashes == 1 {
+		b = append(b, '\\')
+	}
 	if needsQuotes {
 		return fmt.Sprintf(`"%s"`, string(b))
 	}
--- a/pkg/skaffold/deploy/helm/args.go
+++ b/pkg/skaffold/deploy/helm/args.go
@@ -18,6 +18,7 @@ package helm
 
 import (
 	"fmt"
+	"runtime"
 	"strconv"
 
 	"github.com/blang/semver"
@@ -55,6 +56,8 @@ func constructOverrideArgs(r *latest_v1.HelmRelease, builds []graph.Artifact, ar
 		if err != nil {
 			return nil, fmt.Errorf("unable to expand %q: %w", r.SetFiles[k], err)
 		}
+		exp = sanitizeFilePath(exp, runtime.GOOS == "windows")
+
 		record(exp)
 		args = append(args, "--set-file", fmt.Sprintf("%s=%s", k, exp))
 	}
@@ -200,4 +203,52 @@ func (h *Deployer) installArgs(r latest_v1.HelmRelease, builds []graph.Artifact,
 	}
 
 	return args, nil
+}
+
+// sanitizeFilePath is used to sanitize filepaths that are provided to the `setFiles` flag
+// helm `setFiles` doesn't work with the unescaped filepath separator (\) for Windows or if there are unescaped tabs and spaces in the directory names.
+// So we escape all odd count occurrences of `\` for Windows, and wrap the entire string in quotes if it has spaces.
+// This is very specific to the way helm handles its flags.
+// See https://github.com/helm/helm/blob/d55c53df4e394fb62b0514a09c57bce235dd7877/pkg/cli/values/options.
+// Otherwise the windows `syscall` package implements its own sanitizing for command args that's used by `exec.Cmd`.
+// See https://github.com/golang/go/blob/6951da56b0ae2cd4250fc1b0350d090aed633ac1/src/syscall/exec_windows.go#L27
+func sanitizeFilePath(s string, isWindowsOS bool) string {
+	if len(s) == 0 {
+		return `""`
+	}
+	needsQuotes := false
+	for i := 0; i < len(s); i++ {
+		if s[i] == ' ' || s[i] == '\t' {
+			needsQuotes = true
+			break
+		}
+	}
+
+	if !isWindowsOS {
+		if needsQuotes {
+			return fmt.Sprintf(`"%s"`, s)
+		}
+		return s
+	}
+
+	var b []byte
+	slashes := 0
+	for i := 0; i < len(s); i++ {
+		switch s[i] {
+		case '\\':
+			slashes++
+		default:
+			// ensure a single slash is escaped
+			if slashes == 1 {
+				b = append(b, '\\')
+			}
+			slashes = 0
+		}
+		b = append(b, s[i])
+	}
+
+	if needsQuotes {
+		return fmt.Sprintf(`"%s"`, string(b))
+	}
+	return string(b)
 }
--- a/pkg/skaffold/deploy/helm/args_test.go
+++ b/pkg/skaffold/deploy/helm/args_test.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2021 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helm
+
+import (
+	"testing"
+
+	"github.com/GoogleContainerTools/skaffold/testutil"
+)
+
+func TestSanitizeFilePath(t *testing.T) {
+	tests := []struct {
+		description string
+		isWindowsOS bool
+		input       string
+		output      string
+	}{
+		{
+			description: "unescaped relative path on Windows",
+			isWindowsOS: true,
+			input:       `.\foo\not.escaped.relative.yaml`,
+			output:      `.\\foo\\not.escaped.relative.yaml`,
+		},
+		{
+			description: "unescaped absolute path on Windows",
+			isWindowsOS: true,
+			input:       `C:\Users\foo\not.escaped.abs.yaml`,
+			output:      `C:\\Users\\foo\\not.escaped.abs.yaml`,
+		},
+		{
+			description: "escaped relative path on Windows",
+			isWindowsOS: true,
+			input:       `.\\foo\\escaped.relative.yaml`,
+			output:      `.\\foo\\escaped.relative.yaml`,
+		},
+		{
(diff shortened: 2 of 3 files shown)


pkg/skaffold/deploy/helm/args.go
```diff
@@ -201,3 +204,54 @@ func (h *Deployer) installArgs(r latest.HelmRelease, builds []build.Artifact, va
...
+		switch s[i] {
+		case ' ', '\t':
+			needsQuotes = true
+			break
+		}
+	}
+
+	if runtime.GOOS != "windows" {
+		if needsQuotes {
+			return fmt.Sprintf(`"%s"`, s)
```
briandealwis, line 227: What about paths with embedded quotes? 

(It feels like there should be a library out there that does this!)

pkg/skaffold/deploy/helm/helm_windows_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/mitchellh/go-homedir"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
+	schemautil "github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/warnings"
+	"github.com/GoogleContainerTools/skaffold/testutil"
+)
+
+func TestSetFilesFlagInDeploy(t *testing.T) {
```
briandealwis, line 38: It would seem better to extract `sanitizeFilePath()` as a library function into `pkg/skaffold/util` and just have specific tests around that.

Maybe we can make the file-sanitizer pluggable so that we can remove Windows-isms from these tests?

gsquared94: I actually think this is very specific to the way helm [handles](https://github.com/helm/helm/blob/d55c53df4e394fb62b0514a09c57bce235dd7877/pkg/cli/values/options.go#L75) its flags. We otherwise use `exec.Cmd` everywhere else and the windows [`syscall`](https://github.com/golang/go/blob/6951da56b0ae2cd4250fc1b0350d090aed633ac1/src/syscall/exec_windows.go#L27) package implements its own sanitizing for command args. So I don't think we need to move this to a common util.

pkg/skaffold/deploy/helm/args.go
gsquared94: can a filepath have embedded quotes? Currently the only client of this is the `setFiles` parameter for the `helm` deployer and the yaml parser doesn't preserve wrapping quotes anyways.

ReneWerner87:
is there a progress in the pull request, under windows no work with multi configurations is possible without this fix

would be cool if you could push this a little bit

pkg/skaffold/deploy/helm/args.go
```diff
@@ -201,3 +204,54 @@ func (h *Deployer) installArgs(r latest.HelmRelease, builds []build.Artifact, va
 
 	return args, nil
 }
+
+// sanitizeFilePath is used to sanitize filepaths that are provided to the `setFiles` flag
+// helm `setFiles` doesn't work with the unescaped filepath separator (\) for Windows or if there are unescaped tabs and spaces in the directory names.
+// So we escape all odd count occurrences of `\` for Windows, and wrap the entire string in quotes if it has spaces.
```
briandealwis, line 210: Can you include the information you [mention below](https://github.com/GoogleContainerTools/skaffold/pull/5648/files#r610792583)? 

Does the oddness of backslashes matter?  Can I realistically use three backslashes?

briandealwis (commented):
I have more questions

pkg/skaffold/deploy/helm/args.go
gsquared94: I think erroring out with the `helm` propagated error message is enough in these scenarios?

pkg/skaffold/deploy/helm/args.go
gsquared94: I'll do away with this odd-even checks just escaping single `\` characters.

gsquared94:
> I have more questions

Can you take another look?

pkg/skaffold/deploy/helm/args.go
```diff
@@ -201,3 +204,53 @@ func (h *Deployer) installArgs(r latest_v1.HelmRelease, builds []graph.Artifact,
...
+// This is very specific to the way helm handles its flags.
+// See https://github.com/helm/helm/blob/d55c53df4e394fb62b0514a09c57bce235dd7877/pkg/cli/values/options.
+// Otherwise the windows `syscall`package implements its own sanitizing for command args that's used by `exec.Cmd`.
```
briandealwis, line 213:
```suggestion
// Otherwise the windows `syscall` package implements its own sanitizing for command args that's used by `exec.Cmd`.
```

pkg/skaffold/deploy/helm/args.go
```diff
@@ -201,3 +204,53 @@ func (h *Deployer) installArgs(r latest_v1.HelmRelease, builds []graph.Artifact,
...
+// See https://github.com/golang/go/blob/6951da56b0ae2cd4250fc1b0350d090aed633ac1/src/syscall/exec_windows.go#L27
+func sanitizeFilePath(s string) string {
+	needsQuotes := false
```
briandealwis, line 216: nit: Move this variable definition below, since it's not used by the `if`

pkg/skaffold/deploy/helm/args.go
briandealwis: Ok — I wrote the above thinking that we needed to do escaping for `CMD /C`, not Helm.

pkg/skaffold/deploy/helm/args.go
```diff
@@ -201,3 +204,53 @@ func (h *Deployer) installArgs(r latest_v1.HelmRelease, builds []graph.Artifact,
...
+			continue
+		}
+
+		if slashes == 1 {
+			b = append(b, '\\')
+		}
+		for ; slashes > 0; slashes-- {
+			b = append(b, '\\')
+		}
+		b = append(b, s[i])
```
briandealwis, line 249: I looked at the PR on the weekend and I found this puzzling enough to leave until this morning.  I had to read through this a few times to understand it (high cyclomatic complexity).

WDYT of this simplification:
```suggestion
		switch s[i] {
		case '\\':
			slashes++

		default:
			// ensure a single slash is escaped
			if slashes == 1 {
				b = append(b, '\\')
			}
			slashes = 0
		}
		b = append(b, s[i])
```

pkg/skaffold/deploy/helm/helm_windows_test.go
```diff
@@ -0,0 +1,108 @@
...
+				SetFiles: map[string]string{
+					"expanded":                    "~/file.yaml",
+					"not.escaped.relative":        `.\foo\not.escaped.relative.yaml`,
+					"not.escaped.abs":             `C:\Users\foo\not.escaped.abs.yaml`,
+					"escaped.relative":            `.\\foo\\escaped.relative.yaml`,
+					"escaped.abs":                 `C:\\Users\\foo\\escaped.abs.yaml`,
+					"escaped.spaces.relative":     `.\\foo bar\\escaped.spaces.relative.yaml`,
+					"escaped.spaces.abs":          `C:\\Users\\foo bar\\escaped.spaces.abs.yaml`,
+					"not.escaped.spaces.relative": `.\foo bar\not.escaped.spaces.relative.yaml`,
+					"not.escaped.spaces.abs":      `C:\Users\foo bar\not.escaped.spaces.abs.yaml`,
```
briandealwis, line 97: Honestly, I'd rather we just have a specific test for `sanitizeFilePath` with these different permutations — including quotation tests for non-Windows — with some of the tests marked as being Windows specific.  The rest of this test is re-testing functionality already checked in `TestHelmDeploy`.  Or am I missing something?

pkg/skaffold/deploy/helm/args.go
```diff
@@ -201,3 +204,51 @@ func (h *Deployer) installArgs(r latest_v1.HelmRelease, builds []graph.Artifact,
...
+
+	var b []byte
+	slashes := 0
+	for i := 0; i < len(s); i++ {
+		switch s[i] {
+		case '\\':
+			slashes++
+		default:
+			// ensure a single slash is escaped
+			if slashes == 1 {
```
briandealwis, line 242: Do we need to handle a trailing slash? Or two trailing slashes?

gsquared94: I don't think so. We're expecting these to end in filenames.

briandealwis: Can we validate that these values are files?  Then we should be good to go.

gsquared94: I need to use my Windows machine to test. Will try and do it tonight.

gsquared94: `helm` will fail with the right validation errors. I don't think we should be doing another validation.

briandealwis (approved):

ngraf:
A colleague of mine, who is working on Windows machine, installed latest Skaffold binary and confirmed it is fixed. 💚

Issue #5478, opened 2021-03-02
Add `--user=skaffold` flag to all minikube command invokations.


tejal29:
In minikube v1.8.0,  minikube team added a flag `--user` to audit commands.

We should use this flag, if user is on v1.8.0

tejal29:
Reopening as this was reverted in v1.22.0

tejal29:
Revert PR #5682

tejal29:
This went in as #5732

PR #5649, opened 2021-04-08
Add user flag to all minikube command invocations

PriyaModali:
Fixes: #5478

Add --user=skaffold flag to all minikube command invocations.

In v1.8.0, minikube team added a flag `--user` to audit commands.

We should use this flag, if user is on v1.8.0 or higher.


pkg/skaffold/cluster/minikube.go
```diff
@@ -93,9 +103,46 @@ func minikubeExec(arg ...string) (*exec.Cmd, error) {
...
+	cmd := exec.Command("minikube", "version")
+	out, err := util.RunCmdOut(cmd)
+	if err != nil {
+		return false, err
+	}
+
+	currentVersion, err := getCurrentVersion(string(out))
+	if err != nil {
+		return false, err
+	}
```
briandealwis, line 128: This code will cause `minikube version` to be executed each time we invoke minikube.  You need to use the cached currentVersion to determine whether to run `minikube version`.

Better yet, move this code into `minikubeBinary()`.

pkg/skaffold/cluster/minikube_test.go
```diff
@@ -59,28 +59,28 @@ func TestClientImpl_IsMinikube(t *testing.T) {
 		{
 			description:        "cluster cert outside minikube dir",
 			kubeContext:        "test-cluster",
-			minikubeProfileCmd: testutil.CmdRunOut("minikube profile list -o json", fmt.Sprintf(profileStr, "test-cluster", "docker", "172.17.0.3", 8443)),
+			minikubeProfileCmd: testutil.CmdRunOut("minikube profile list -o json --user=skaffold", fmt.Sprintf(profileStr, "test-cluster", "docker", "172.17.0.3", 8443)),
```
briandealwis, line 62: There should be a test for minikube < 1.18.0 and for invalid versions.

pkg/skaffold/cluster/minikube.go
```diff
@@ -24,21 +24,31 @@ import (
...
 	"k8s.io/client-go/util/homedir"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/context"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/version"
 )
 
+const (
+	MinikubeVersionWithUserFlag = "1.18.0"
```
tejal29, line 40: this can be made private

pkg/skaffold/cluster/minikube.go
```diff
@@ -93,9 +103,46 @@ func minikubeExec(arg ...string) (*exec.Cmd, error) {
 	if err != nil {
 		return nil, fmt.Errorf("getting minikube executable: %w", err)
 	}
+
+	minVer, err := supportsUserFlagFunc()
```
tejal29, line 107: It would make sense for `minikubeBinaryFunc` to return the binary path as well as version so you don't have to return err twice. 
1) minikube binary issue
2) minikube version parsing.

pkg/skaffold/cluster/minikube.go
```diff
@@ -93,9 +103,46 @@ func minikubeExec(arg ...string) (*exec.Cmd, error) {
...
+	if err != nil {
+		return nil, err
```
tejal29, line 109: Instead of returning err here, should we fall back to not adding minikube user flag.

pkg/skaffold/cluster/minikube.go
tejal29, line 107:
```suggestion
	isSupported, err := supportsUserFlagFunc()
```

pkg/skaffold/cluster/minikube.go
```diff
@@ -24,21 +24,36 @@ import (
...
-	getClusterInfo     = context.GetClusterInfo
+	minikubeBinaryFunc    = minikubeBinary
+	getClusterInfo        = context.GetClusterInfo
+	GetCurrentVersionFunc = getCurrentVersion
+
+	findOnce sync.Once
+	mk       = struct {
+		version semver.Version
+		path    string
+		err     error
```
briandealwis, line 55: I would move err to the top as it determines the validity of the remaining fields.  If `err != nil`, then path and version are unset.  If `err == nil` then path and version should be valid.

pkg/skaffold/cluster/minikube.go
```diff
@@ -24,21 +24,36 @@ import (
...
 	"k8s.io/client-go/util/homedir"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/context"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/version"
+)
+
+const (
+	minikubeVersionWithUserFlag = "1.18.0"
```
briandealwis, line 40: Since this is our string, we can just use `MustParse()` here
```suggestion
	minikubeVersionWithUserFlag = semver.MustParse("1.18.0")
```

and then turn `supportsUserFlag()` into a predicate:
```
func supportsUserFlag(ver semver.Version) bool {
	return ver.GE(minikubeVersionWithUserFlag)
}
```

with the cascading changes.

pkg/skaffold/cluster/minikube_test.go
```diff
@@ -31,13 +32,14 @@ import (
...
-		minikubeNotInPath  bool
-		expected           bool
+		description             string
+		kubeContext             string
+		certPath                string
+		serverURL               string
+		minikubeProfileCmd      util.Command
+		minikubeNotInPath       bool
+		expected                bool
+		minikuneWithoutUserFalg bool
```
briandealwis, line 42:
```suggestion
		minikubeWithoutUserFlag bool
```

pkg/skaffold/cluster/minikube_test.go
```diff
@@ -56,41 +58,60 @@ func TestClientImpl_IsMinikube(t *testing.T) {
...
 			if test.minikubeNotInPath {
-				t.Override(&minikubeBinaryFunc, func() (string, error) { return "", fmt.Errorf("minikube not in PATH") })
+				ver, _ := semver.Make("1.18.0")
+				t.Override(&minikubeBinaryFunc, func() (string, semver.Version, error) {
+					return "", ver, fmt.Errorf("minikube not in PATH")
+				})
 			} else {
-				t.Override(&minikubeBinaryFunc, func() (string, error) { return "minikube", nil })
+				if test.minikuneWithoutUserFalg {
+					ver, _ := semver.Make("1.17.0")
```
briandealwis, line 107: Use `MustParse()` or 
```suggestion
					ver := semver.Version{Major: 1, Minor: 17, Patch: 0}
```

briandealwis (approved):
Just a few small things to address.

pkg/skaffold/test/structure/structure_test.go
```diff
@@ -38,6 +41,8 @@ func TestNewRunner(t *testing.T) {
 
 	testutil.Run(t, "", func(t *testutil.T) {
 		tmpDir := t.NewTempDir().Touch("test.yaml")
+		t.Override(&cluster.GetCurrentVersionFunc, func() (semver.Version, error) { return semver.Version{}, nil })
```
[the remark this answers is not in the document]
briandealwis: And then we can avoid exporting GetCurrentVersionFunc.

pkg/skaffold/cluster/minikube.go
```diff
@@ -89,22 +104,60 @@ func (clientImpl) MinikubeExec(arg ...string) (*exec.Cmd, error) {
...
 func minikubeExec(arg ...string) (*exec.Cmd, error) {
-	b, err := minikubeBinaryFunc()
+	b, v, err := minikubeBinaryFunc()
 	if err != nil {
 		return nil, fmt.Errorf("getting minikube executable: %w", err)
 	}
+
+	isSupported, _ := supportsUserFlag(v)
+	if isSupported {
+		arg = append(arg, "--user=skaffold")
```
gsquared94, line 114: does `arg = append(arg, "--user", "skaffold")` not work? I think it's preferred to have each arg be a single slice element.

PriyaModali: That seem to add a space in-between `--user` & `skaffold`.

briandealwis:
Unfortunately this change broke the minikube support.  The output of `minikube version` is not a single semver version.
```
$ minikube version
minikube version: v1.18.1
commit: 09ee84d530de4a92f00f1c5dbc34cead092b95bc
```

This PR should:
- use `minikube version --short` — or perhaps `minikube version -o json`
- add explicit tests for different inputs
- treat an unparseable result from `minikube version` as an old version of minikube.

PR #5650, opened 2021-04-08
Upgrade Jib versions to 3.0

chanseokoh:

PR #5651, opened 2021-04-08
Update Jib to 3.0 and set base images

chanseokoh:
Put up a new PR. (Supersedes #5650. #5650 has a weird check failure, so I did a fresh clone. Hopefully this resolves the failure.)

Also explicitly sets base images to avoid Docker Hub rate limiting.


briandealwis (approved):
LGTM: the tests failure is because we usually create PRs from forks (#5656).

PR #5654, opened 2021-04-09
Fix `default-repo` by supporting nil as default value for flags

briandealwis:
Fixes: #5588
**Related**: #5632, #5548

**Description**
Adds support for "nillable" values like the use of `StringOrUndefined` used for `--default-repo`.  

Prior to #5632, we never initialized struct values (called `Var` in pflag).  But with the new `--port-forward` implementation (#5554), we now needed to initialize struct values with possibly different values depending on the command (#5548).

With #5548 we now reset `Var` type values to their default value, but it turns out that `--default-repo` had an incorrect default value of the empty string:
https://github.com/GoogleContainerTools/skaffold/blob/bcb2eaac8a642e71a8dcd2da1864b93295bd84aa/cmd/skaffold/app/cmd/flags.go#L99-L105

This default value should have been `nil`: the default repo support differentiates between `nil` and the empty string "".  When `nil`, Skaffold will look to see if there is a value set in the global config; whereas the empty string indicates that no default-repo should be used.

- the `default-repo` is able to be used from the global config again

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -34,6 +34,13 @@ var (
 	fromBuildOutputFile flags.BuildOutputFileFlag
 )
 
+// Nillable is used to reset objects that implement pflag's `Value` and `SliceValue`.
+// Some flags, like `--default-repo`, use nil to indicate that they are unset, which
+// is different from the empty string.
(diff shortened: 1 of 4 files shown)


MarlonGamez (approved):
This looks good to me. I've also tested locally to verify it works

Issue #5660, opened 2021-04-09
InputDigest is not supported by CustomTemplateTagger


mikekamornikov:
### Expected behavior

I expect image tag with both some custom text and input digest created with `customTemplate` `tagPolicy`

### Actual behavior

```sh
$ skaffold build -p test-custom-template
creating runner: creating tagger: creating tagger: creating components: unknown component for custom template: DIGEST {GitTagger:<nil> ShaTagger:<nil> EnvTemplateTagger:<nil> DateTimeTagger:<nil> CustomTemplateTagger:<nil> InputDigest:0x7a479b8}
. If above error is unexpected, please open an issue https://github.com/GoogleContainerTools/skaffold/issues/new to report this error.
```

### Information

- Skaffold version: `v1.21.0-29-g89b83154e-dirty`
- Operating system: `macOS 10.15.7`
- Installed via: `built from master`
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta14
kind: Config
profiles:
- name: test-custom-template
  build:
    tagPolicy:
      customTemplate:
        template: "release-aaa-{{.DIGEST}}"
        components:
        - name: DIGEST
          inputDigest: {}
    artifacts:
    - image: quay.io/my/project
      context: ../../..
      bazel:
        target: //project:docker.tar
```

### Steps to reproduce the behavior

1. clone this repo and build `skaffold` with `LOCAL=true make install`
2. `skaffold build -p test-custom-template`
3. ...

mikekamornikov:
@piotr-szybicki This one is related to #5192 .

tejal29:
will be fixed by https://github.com/GoogleContainerTools/skaffold/pull/5661

PR #5661, opened 2021-04-09
Add InputDigest support to CustomTemplateTagger

mikekamornikov:
Fixes: #5660  
**Related**: #5192 

**Description**
Allows `InputDigest` to be added to a map of taggers of `CustomTemplateTagger`.


mikekamornikov:
@piotr-szybicki I'm not sure if it's enough assuming `func (t *customTemplateTagger) EvaluateComponents` itself makes additions to `customMap` in the runtime. But this way it creates the tags I expect.

MarlonGamez (approved):
lgtm! Confirmed working locally. Thanks for taking the time to add this and update tests as well :)

PR #5663, opened 2021-04-10
Try reducing ttl to 30 seconds

briandealwis:
Re: #5248

**Description**
Try reducing the TTL for Maven Wagon pooled connections to 30s in the hopes that it may prevent connection resets during tests.


briandealwis:
Committing directly as this is a pretty trivial testing fix.

PR #5669, opened 2021-04-13
Allocate ports against INADDR_ANY

briandealwis:
Towards: #5590 

**Description**
This PR modifies our `GetAvailablePort()` function, and its uses, to allocate ports using INADDR_ANY (0.0.0.0) rather than `localhost` (127.0.0.1) as it is possible to listen on a port for a specific interface _even as another process is listening on INADDR_ANY with the same port_.


briandealwis:
Merged with #5670

PR #5670, opened 2021-04-13
Forwarding resources should not allocate system ports

briandealwis:
Fixes: #5590 
**Merge before/after**: #5669 

**Description**
This PR modifies our resource forwarder, used to forward ports defined by services and pods, to only request matching local ports when the remote port is > 1023.  So a service on port 8080 should still map to port 8080 (if available), but a service on port 80 will be mapped to some port ≥ 1024.

User-defined port-forwards in the `skaffold.yaml` are unaffected and can bind to system ports.

- automatic port-forwards no longer allocate ports 1–1023


nkubala (approved):
logic of this LGTM. we might want to call this out in https://skaffold.dev/docs/pipeline-stages/port-forwarding/ to avoid potential confusion, but I'll leave that up to you

PR #5674, opened 2021-04-13
[WIP] [V3] basic CMD struct: Add entrypoints to skaffold CMD

yuwenma:
**Related**: 5673

**Description**
1. Add a new flag "render-v2" so the render and deploy related commands can use the v3 path.
2. Add placeholders for later implementation.


tejal29:
Thanks @yuwenma Will take a look today afternoon!

yuwenma:
Need some refactoring before adding the render struct.

PR #5689, opened 2021-04-15
Emit TaskEvents Protos for PortForwarding

MarlonGamez:
Fixes: #5422
**Related**: #5368 #5422

**Description**
Emit taskevent protos on v2 endpoints

Users can now use the v2 event API to see the start, end, or failure of the portforward phase


pkg/skaffold/kubernetes/portforward/forwarder_manager.go
```diff
@@ -39,11 +41,12 @@ type Forwarder interface {
 
 // ForwarderManager manages all forwarders
 type ForwarderManager struct {
+	iteration  int
 	forwarders []Forwarder
 }
 
 // NewForwarderManager returns a new port manager which handles starting and stopping port forwarding
-func NewForwarderManager(out io.Writer, cli *kubectl.CLI, podSelector kubernetes.PodSelector, namespaces []string, label string, runMode config.RunMode, options config.PortForwardOptions, userDefined []*latest.PortForwardResource) *ForwarderManager {
+func NewForwarderManager(out io.Writer, cli *kubectl.CLI, podSelector kubernetes.PodSelector, iteration int, label string, runMode config.RunMode, options config.PortForwardOptions, userDefined []*latest.PortForwardResource) *ForwarderManager {
```
tejal29, line 49: I was under the impression, `NewForwarderManager`  was called only once at the start `devloop`. Can you confirm if you are seeing events for subsequent dev loops with correct dev iteration?

MarlonGamez: You're right, I meant to come back to this PR and update it as this approach doesn't work. In https://github.com/GoogleContainerTools/skaffold/pull/5710 I wanted to add the iteration field to the event handler and have a function to grab that value so that it doesn't have to be plumbed down into unrelated skaffold logic like I'm doing here

pkg/skaffold/kubernetes/portforward/forwarder_manager.go
```diff
@@ -107,11 +109,15 @@ func (p *ForwarderManager) Start(ctx context.Context, namespaces []string) error
 		return nil
 	}
 
+	eventV2.TaskInProgress(constants.PortForward, eventV2.GetIteration())
```
tejal29, line 112: This kind of looks a but odd. Shd we just change all the other uses of `eventV2.TaskInProgress` to use `eventV2.GetIteration()`?
```suggestion
	eventV2.TaskInProgress(constants.PortForward)
```

MarlonGamez: Good point, or I can just change all `Task*()` functions to use the underlying handler's iteration value and remove the extra parameter altogether. The handler's iteration field could then just increment when it sees `constants.DevLoop`

tejal29: yeah. Great idea. We don't even need the `devIteration` in runner.

tejal29 (approved):

Issue #5673, opened 2021-04-13
[Epic] Skaffold Render v2


yuwenma:
This is a tracking issue for skaffold v2 

 - [x] Legacy helm support: a user can continue to use `helm install` or `helm upgrade`
 - [ ] #7564
-  [x] Skaffold fix to translate v1 -> v2 configs
-  [x] Skaffold init support for v2 configs
-  [ ] Document new config schema and pipeline
-  [ ] #7565
-  [ ] #7566
-  [ ] Cloud Code Modules integration test - `skaffold debug` with modules makes all artifacts debuggable and port-forwarded

tejal29:
Moving this to july end milestone since we need to finish it by then. 
Can you add individual tasks for this EPIC?

tejal29:
Sync with @yuwenma to close this issue.

aaron-prindle:
Closing this in favor of individual bugs created for each bullet that are currently in the `alpha-fixup2` milestone:
https://github.com/GoogleContainerTools/skaffold/milestone/45

PR #5692, opened 2021-04-16
[V3] New V3 SkaffoldRunner

yuwenma:
**Related**:  #5673

**Description**
1. Refactored the runner directory.
2. Create a New SkaffoldRunner for V3. Both runner.SkaffoldRunner and v3.SkaffoldRunner implements the Runner interface.
2. Build, Test, Prune and other shared methods are moved from SkaffoldRunner to several individual structs (BuildRunner, TestRunner).
3. Both SkaffoldRunner and the V3 SkaffoldRunner have the shared structs as components.
4. Unittests are updated.
5. Add placeholders (not implemented error) in v3 SkaffoldRunner for later implementation.


pkg/skaffold/runner/build_deploy.go
```diff
@@ -99,57 +116,13 @@ func (r *SkaffoldRunner) Build(ctx context.Context, out io.Writer, artifacts []*
...
-func (r *SkaffoldRunner) Test(ctx context.Context, out io.Writer, artifacts []graph.Artifact) error {
-	if err := r.tester.Test(ctx, out, artifacts); err != nil {
-		return err
-	}
-
-	return nil
-}
-
-// DeployAndLog deploys a list of already built artifacts and optionally show the logs.
-func (r *SkaffoldRunner) DeployAndLog(ctx context.Context, out io.Writer, artifacts []graph.Artifact) error {
```
yuwenma, line 112: Moved to Deploy.go

pkg/skaffold/runner/v3/generate_pipeline.go
```diff
@@ -0,0 +1,28 @@
+/*
```
MarlonGamez, line 1: not completely sure we'll be supporting this feature in the future, might be best to just leave it out of newer versions?
wdyt @tejal29 @nkubala ?

MarlonGamez (commented):
This looks good to me, I think the thing I'm confused about is the v3 naming. I understand that it's named off the schema, but is kinda confusing since skaffold's major version number is still 1

pkg/skaffold/runner/v3/generate_pipeline.go
yuwenma: +1

I'd vote for removing the generate-pipeline from v3, or at most p2. 
For the v2 main stream, skaffold+ tekton seems to be a nice combination. the tekton also use skaffold in their examples see [here](https://github.com/tektoncd/pipeline/blob/main/docs/tutorial.md#running-your-task)

yuwenma:
> This looks good to me, I think the thing I'm confused about is the v3 package naming. I understand that it's named off the schema, but is kinda confusing since skaffold's major version number is still 1

yeah, when changing the schema from v1 to v2, I think we actually should bump the skaffold version to v2 at that time. Do you suggest using v2 here or re-align the schema and master version in this new skaffold render?

pkg/skaffold/runner/v3/generate_pipeline.go
tejal29: hmm i can see that generated yaml would change. We could re-implement what we have as P2.

We should also instrument its usage.

tejal29 (changes requested):

pkg/skaffold/runner/build_deploy.go
```diff
@@ -26,18 +26,35 @@ import (
...
 	eventV2 "github.com/GoogleContainerTools/skaffold/pkg/skaffold/event/v2"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/graph"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner/runcontext"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/tag"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 )
 
+type BuildRunner struct {
```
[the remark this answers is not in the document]
yuwenma: Sure. was trying to distinguish this builder with the build.Builder.

tejal29 (approved):

PR #5694, opened 2021-04-16
[V3] Add  flag as v3 entrypoint.

yuwenma:
**Related**:  #5673 

**Description**
Allow users to have access to V3 (skaffold render v2) via a "--experimental" flag in `skaffold dev|render|deploy|run` conmmands. 

This flag can be turned off once all V3 features are completed. Due to the long developing period as well as the kpt dependencies, we will support both v2 and V3 version streams developing and releasing together (A single skaffold binary) until V3 is ready for Beta.
 
Users can switch between V2 and V3 via the experimental flag. e.g. `skaffold dev --experimental ...` uses V3.

We also need a release.md doc (for developers) to talk about the new stream.
 
**Following Changes**
Schema refactoring for V3
Runner refactoring: #5692

PR #5695, opened 2021-04-16
[WIP][V3] Add the new schema v3alpha1.

yuwenma:
**Related**: #5673

**Description**
his schema contains the main fields to support the new skaffold render and deployment, powered by kpt. Some config names may change based on the the user survey results.

For reviewers:
 - `Render`: Top tier section. It contains the basic render functions `generate`, `validate` and `trasform`. It does not include the helmcharts (generate) and kpt package magement yet, which will be added on following CLs.
 - `Deploy`: Simplified. The deployers (helm, kubectl, kpt, kustomize) are removed. Some helm and kubectl features, and flags are not covered in the `deploy` section yet. Will be added in following CLs.
 -  v2 and v3 versions are not compatible. So the `upgrade` functions do not need to change. But we will add more checks on other codes to guarantee the v2 and v3 are not mixed (especially in multi-config mode).
 -  Generated doc via "go run hack/schemas/main.go".

### DIFF
--- a/pkg/skaffold/schema/v3alpha1/config.go
+++ b/pkg/skaffold/schema/v3alpha1/config.go
@@ -1,0 +1,1405 @@
+/*
+Copyright 2021 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v3alpha1
+
+import (
+	"encoding/json"
+
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
+)
+
+const Version string = "skaffold/v3alpha1"
+
+// NewSkaffoldConfig creates a SkaffoldConfig
+func NewSkaffoldConfig() util.VersionedConfig {
+	return new(SkaffoldConfig)
+}
+
+// SkaffoldConfig holds the fields parsed from the Skaffold configuration file (skaffold.yaml).
+type SkaffoldConfig struct {
+	// APIVersion is the version of the configuration.
+	APIVersion string `yaml:"apiVersion" yamltags:"required"`
+
+	// Kind is always `Config`. Defaults to `Config`.
+	Kind string `yaml:"kind" yamltags:"required"`
+
+	// Metadata holds additional information about the config.
+	Metadata Metadata `yaml:"metadata,omitempty"`
+
+	// Dependencies describes a list of other required configs for the current config.
+	Dependencies []ConfigDependency `yaml:"requires,omitempty"`
+
+	// Pipeline defines the Build/Test/Deploy phases.
+	Pipeline `yaml:",inline"`
+
+	// Profiles *beta* can override be used to `build`, `test` or `deploy` configuration.
+	Profiles []Profile `yaml:"profiles,omitempty"`
+}
+
+// Metadata holds an optional name of the project.
+type Metadata struct {
+	// Name is an identifier for the project.
+	Name string `yaml:"name,omitempty"`
+}
+
+// Pipeline describes a Skaffold pipeline.
+type Pipeline struct {
+	// Build describes how images are built.
+	Build BuildConfig `yaml:"build,omitempty"`
+
+	// Test describes how images are tested.
+	Test []*TestCase `yaml:"test,omitempty"`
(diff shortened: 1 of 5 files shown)


pkg/skaffold/schema/v3alpha1/config.go
```diff
@@ -0,0 +1,1405 @@
+/*
```
gsquared94, line 1: If we're not going to immediately release this config I think we should move it to `pkg/skaffold/schema/experimental` or something. I think that'll help running into issues with all the `make` targets that do releases, new configs, etc.

yuwenma: This new version will go public in the next skaffold release (May 13th). It will allow early v3 users to try it out and won't mess up the current user experience.

Here's a corresponding PR. https://github.com/GoogleContainerTools/skaffold/pull/5694
Basically, v3 users run `skaffold dev --experimental`, v2 users run `skaffold dev`. 
New v3 release will bump the v3 master version v3alpha1 v3alpha2, while we also bump v2 versions for each month release v2beta14, v2beta15. v2 and v3 are **incompatible** with each other.

Thank you for raising this concern @gsquared94. I'll double check the make rules. BTW, the make rules are pretty cool.(my CLs have broken quicktest, coverage, linters, so far. Really nice coverage! 😄)  
 
cc @viglesiasce @tejal29 @nkubala  @marlon-gamez @aaron-prindle

MarlonGamez (commented):
This is looking good, but we need to modify hack/check-samples here to not treat this config as the latest released one. Right now in `hack/versions/cmd/latest-released/version.go` it's grabbing the latest version by doing `schema.SchemaVersions[len(schema.SchemaVersions)-2].APIVersion`, which will hit `v3alpha1` since that slice now looks like 
```
...
        {v2beta13.Version, v2beta13.NewSkaffoldConfig, true},
        {v2beta14.Version, v2beta14.NewSkaffoldConfig, true},
	{v3alpha1.Version, v3alpha1.NewSkaffoldConfig, false},
	{latest.Version, latest.NewSkaffoldConfig, true},
}
```

pkg/skaffold/schema/v3alpha1/config.go
tejal29: Thanks @yuwenma for  bringing this up. 

We are now moving to 2 weeks release cycle and this might get published before May 13th.
That said, we are running behind on V3 release plan. 
I don't mind the version number v2alpha1, v3alpha2 as alpha signifies experimental phase and users can choose if they want to move to the next experimental config or else config management would be confusing for users.

Like @marlon-gamez pointed previously, we should not point `v3alpha*` as latest. Can you address that in a separate PR?
Since this is the first version, we could do any from below
1) Make the `v3` flag hidden and not publish any docs for `v3alpha*` config,.
2) Make the `v3` flag hidden, create a directory https://skaffold-latest.firebaseapp.com/docs/references/yaml/v3/

When we do announce skaffold kpt integration, May 13th potentially
1) Unhide `v3` flag and publish  first `v3alpha1` schema version in https://skaffold-latest.firebaseapp.com/docs/references/yaml/v3/
2) From there on, any config changes would incremental config version

tejal29: Freezing the v3 config after the announcement of kpt skaffold integration would help you reiterate on the v3 config until the release without worrying about versioning.

pkg/skaffold/schema/versions.go
```diff
@@ -79,53 +80,55 @@ type APIVersion struct {
...
+	{v2beta5.Version, v2beta5.NewSkaffoldConfig, true},
+	{v2beta6.Version, v2beta6.NewSkaffoldConfig, true},
+	{v2beta7.Version, v2beta7.NewSkaffoldConfig, true},
+	{v2beta8.Version, v2beta8.NewSkaffoldConfig, true},
+	{v2beta9.Version, v2beta9.NewSkaffoldConfig, true},
+	{v2beta10.Version, v2beta10.NewSkaffoldConfig, true},
+	{v2beta11.Version, v2beta11.NewSkaffoldConfig, true},
+	{v2beta12.Version, v2beta12.NewSkaffoldConfig, true},
+	{v2beta13.Version, v2beta13.NewSkaffoldConfig, true},
+  {v2beta14.Version, v2beta14.NewSkaffoldConfig, true},
```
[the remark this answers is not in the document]
yuwenma: done

pkg/skaffold/schema/v3alpha1/config.go
yuwenma: @tejal29 

a qq: "we should not point v3alpha* as latest. Can you address that in a separate PR?" The `latest/config.go` is pointed to v2beta*. Do you mean somewhere else?

FYI, there's [doc](https://docs.google.com/document/d/1GFy_5ya-ajjSVe1t6f7IWUlAQu35eTHAZZ1GkJBiqcA/edit?resourcekey=0-sKE-yXjuymQ0w2qqNXnTzw) talking more about the refactoring. Let me know if that makes sense and I can update this PR accordingly.

tejal29: reviewed the doc. This can now be resolved.

yuwenma:
Deprecated.

Issue #5647, opened 2021-04-08
Implement `--build-concurrency` flag to support overriding the builder concurrency setting


gsquared94:
Currently the builder [concurrency](https://skaffold.dev/docs/references/yaml/#build-local-concurrency) setting is specified in each `skaffold.yaml` file. In a multi-config project it's cumbersome to make this change in each and every file.  
It'd be useful to have a flag so users could run like:
```
skaffold dev --build-concurrency=5
```
to immediately make their builds parallel without having to make this change in every `skaffold.yaml` file.

The implementation would involve add the flag to [`flags.go`](https://github.com/GoogleContainerTools/skaffold/blob/master/cmd/skaffold/app/cmd/flags.go) and ultimately modifying the `concurrency` field in [`builder_mux.go`](https://github.com/GoogleContainerTools/skaffold/blob/cc262b48867959858e592f76674f8ea0f37e0d8e/pkg/skaffold/build/builder_mux.go#L34).

PR #5699, opened 2021-04-16
Add build-concurrency flag

sladyn98:
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/5647

**Description**

So this change adds the build concurrency flag to run builds in parallel, instead of having to replace it in every file we would like to provide it as a build option


pkg/skaffold/runner/runcontext/context.go
```diff
@@ -185,6 +185,7 @@ func (rc *RunContext) Tail() bool                                { return rc.Opt
 func (rc *RunContext) Trigger() string                           { return rc.Opts.Trigger }
 func (rc *RunContext) WaitForDeletions() config.WaitForDeletions { return rc.Opts.WaitForDeletions }
 func (rc *RunContext) WatchPollInterval() int                    { return rc.Opts.WatchPollInterval }
+func (rc *RunContext) BuildConcurrency() int                     { return rc.Opts.BuildConcurrency }
```
sladyn98, line 188: Would I need to call this from builder mux https://github.com/GoogleContainerTools/skaffold/blob/cc262b48867959858e592f76674f8ea0f37e0d8e/pkg/skaffold/build/builder_mux.go#L34

pkg/skaffold/config/options.go
```diff
@@ -85,6 +85,7 @@ type SkaffoldOptions struct {
 	Command            string
 	RPCPort            int
 	RPCHTTPPort        int
+	BuildConcurrency   int
```
sladyn98, line 88: I tried to follow the conventions of other flags hence added it here

cmd/skaffold/app/cmd/flags.go
```diff
@@ -509,6 +509,14 @@ var flagRegistry = []Flag{
 		DefinedOn:     []string{"debug", "dev", "run"},
 		IsEnum:        true,
 	},
+	{
+		Name:          "build-concurrency",
+		Usage:         "Number of concurrently running builds",
+		Value:         &opts.AssumeYes,
```
gsquared94, line 515:
```suggestion
		Value:         &opts.BuildConcurrency,
```

cmd/skaffold/app/cmd/flags.go
```diff
@@ -509,6 +509,14 @@ var flagRegistry = []Flag{
...
+		DefValue:      1,
```
gsquared94, line 516: we should only use the value in this flag if it was explicitly set by the user. Only way could be to have the default be `-1` instead, and have a check later if it was overridden by a non-negative integer.

pkg/skaffold/runner/runcontext/context.go
gsquared94: In `builder_mux.go` you can add this method to the `Config` interface:
```go

type Config interface {
	GetPipelines() []latest.Pipeline
	DefaultRepo() *string
	GlobalConfig() string
	BuildConcurrency() int  
}
```

In `NewBuilderMux` function, skip the `minConcurrency` [calculations](https://github.com/GoogleContainerTools/skaffold/blob/e947844e558b9181911906399cb5e3df7fdb28ef/pkg/skaffold/build/builder_mux.go#L62-L73) if the value from that method is non-negative (see above comment also).

pkg/skaffold/config/options.go
gsquared94: looks good

cmd/skaffold/app/cmd/flags.go
sladyn98: Yeah this sounds goo

aaron-prindle:
It might also make sense to update the docs w/ this flag:
docs/content/en/docs/references/cli/_index.md

See #4039 for an example

sladyn98:
Thanks @aaron-prindle I ran the script

gsquared94:
Please run `make linters` and fix whatever is offending it. Everything else looks good!

pkg/skaffold/build/builder_mux.go
```diff
@@ -59,17 +60,22 @@ func NewBuilderMux(cfg Config, store ArtifactStore, builder func(p latest.Pipeli
...
-		case minConcurrency < 0:
-			minConcurrency = concurrency
-			logrus.Infof("build concurrency first set to %d parsed from %s[%d]", minConcurrency, reflect.TypeOf(b).String(), i)
-		case concurrency > 0 && (minConcurrency == 0 || concurrency < minConcurrency):
-			minConcurrency = concurrency
-			logrus.Infof("build concurrency updated to %d parsed from %s[%d]", minConcurrency, reflect.TypeOf(b).String(), i)
-		default:
-			logrus.Infof("build concurrency value %d parsed from %s[%d] is ignored since it's not less than previously set value %d", concurrency, reflect.TypeOf(b).String(), i, minConcurrency)
+
+		if cfg.BuildConcurrency() > 0 {
```
gsquared94, line 64:
```suggestion
		if cfg.BuildConcurrency() >= 0 {
```

concurrency can be set to `0` which means run all in parallel.

cmd/skaffold/app/cmd/flags.go
```diff
@@ -509,6 +509,14 @@ var flagRegistry = []Flag{
 		DefinedOn:     []string{"debug", "dev", "run"},
 		IsEnum:        true,
 	},
+	{
+		Name:          "build-concurrency",
+		Usage:         "Number of concurrently running builds",
```
gsquared94, line 514: Can you be a bit more verbose:
 `Number of concurrently running builds. Set to 0 to run all builds in parallel. Doesn't violate build order among dependencies.`

gsquared94: Sorry for the bits-n-pieces review.

gsquared94: will have to run `./hack/generate-man.sh` script again to update the docs

gsquared94 (approved):
lgtm!

PR #5708, opened 2021-04-19
add skaffold build --push flag

### LINKED ISSUE #5667: `skaffold build` should support a `--push` flag to override build.local.push
The container-debug-support images uses `skaffold build` to build and test images.  These builds don't have a cluster available and so a plain `skaffold build` causes Skaffold to attempt to build and push images, which fails.  So we  explicitly configure `build.local.push: false` so that the images are only built to the local Docker daemon.

aaron-prindle:
Fixes #5667, adding --push flag to skaffold build.  This --push flag overrides any skaffold.yaml "build.local.push: false" configuration for a build allowing easier artifact pushing after local dev without having to modify yaml configuration.

This approach follows the style and convention for adding a flag that has been used prior for adding a flag used by a single command, similar to #4039. Additionally this approach does not change the skaffold build default push behaviour to be false, instead only taking precendence when the --push flag is set true which is the intended use case.

N/A

The addition of this single flag should not have side effects on skaffold functionally outside of --push, the default behavior for skaffold build will be preserved w/o --push specified.

N/A

### DIFF
--- a/pkg/skaffold/runner/new.go
+++ b/pkg/skaffold/runner/new.go
@@ -188,10 +188,15 @@ func isImageLocal(runCtx *runcontext.RunContext, imageName string) (bool, error)
 
 	cl := runCtx.GetCluster()
 	var pushImages bool
-	if pipeline.Build.LocalBuild.Push == nil {
+
+	switch {
+	case runCtx.Opts.PushImages.Value() != nil:
+		logrus.Debugf("push value set via skaffold build --push flag, --push=%t", *runCtx.Opts.PushImages.Value())
+		pushImages = *runCtx.Opts.PushImages.Value()
+	case pipeline.Build.LocalBuild.Push == nil:
 		pushImages = cl.PushImages
 		logrus.Debugf("push value not present, defaulting to %t because cluster.PushImages is %t", pushImages, cl.PushImages)
-	} else {
+	default:
 		pushImages = *pipeline.Build.LocalBuild.Push
 	}
 	return !pushImages, nil
--- a/cmd/skaffold/app/cmd/build.go
+++ b/cmd/skaffold/app/cmd/build.go
@@ -55,6 +55,7 @@ func NewCmdBuild() *cobra.Command {
 			{Value: buildFormatFlag, Name: "output", Shorthand: "o", DefValue: defaultBuildFormatTemplate, Usage: "Used in conjunction with --quiet flag. " + buildFormatFlag.Usage()},
 			{Value: &buildOutputFlag, Name: "file-output", DefValue: "", Usage: "Filename to write build images to"},
 			{Value: &opts.DryRun, Name: "dry-run", DefValue: false, Usage: "Don't build images, just compute the tag for each artifact.", IsEnum: true},
+			{Value: &opts.PushImages, Name: "push", DefValue: nil, Usage: "Push the built images to the specified image repository.", IsEnum: true, NoOptDefVal: "true"},
 		}).
 		WithHouseKeepingMessages().
 		NoArgs(doBuild)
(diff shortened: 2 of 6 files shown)


pkg/skaffold/runner/new.go
```diff
@@ -186,7 +186,10 @@ func isImageLocal(runCtx *runcontext.RunContext, imageName string) (bool, error)
 
 	cl := runCtx.GetCluster()
 	var pushImages bool
-	if pipeline.Build.LocalBuild.Push == nil {
+	if runCtx.Opts.PushImages {
```
gsquared94, line 189: This is unable to distinguish if the `false` flag value was set by default or provided by the user. As such running `skaffold build --push=false` will be the same as running `skaffold build`.  I don't think this works for the linked issue as is.

aaron-prindle: I definitely agree, the UX for the `--push=false` case is not great for users and should not be merged as is.  Any opinions on how to detect if a flag value was set by default or provided by the user?  From our slack discussion there were these options (thanks @MarlonGamez for these suggestions!):
- Implement `BoolOrUndefined` type (similar to `StringOrUndefined` type)
-  Look up flags after creation and check the flag.Changed field.  This might not be possible (or be difficutl) as you would need access to the original cobra.Cmd to lookup flags on it

tejal29: What was the resolution here?

gsquared94: I think we can go with `BoolOrUndefined`. That seems like the simplest approach and the precedence is already set with `StringOrUndefined`.

aaron-prindle: Updated the code to add a BoolOrUndefined type that is now used for the `--push` flag

aaron-prindle: Currently the syntax required here is:
--push=true
vs
--push

Not sure if this is an issue for merging, looking at how this might be fixed atm.

briandealwis: Check out `NoOptDefVal`

aaron-prindle: Great, did exactly what I needed!  Thanks!

pkg/skaffold/runner/new.go
```diff
@@ -188,10 +188,15 @@ func isImageLocal(runCtx *runcontext.RunContext, imageName string) (bool, error)
...
+	switch {
+	case runCtx.Opts.PushImages.Value() != nil:
+		logrus.Debugf("push value set via skaffold build --push flag, --push=%t", *runCtx.Opts.PushImages.Value())
+		pushImages = *runCtx.Opts.PushImages.Value()
+	case pipeline.Build.LocalBuild.Push == nil:
 		pushImages = cl.PushImages
 		logrus.Debugf("push value not present, defaulting to %t because cluster.PushImages is %t", pushImages, cl.PushImages)
-	} else {
+	default:
 		pushImages = *pipeline.Build.LocalBuild.Push
```
tejal29, line 200: Can we add tests here so we never unintentionally break the precedence?
Test to make sure if `runCtx.Opts.PushImages` is not set a and pipeline.Build.LocalBuild.Push is set, then `pushImages` is true. 

If ``runCtx.Opts.PushImages` is set to false and `pipeline.Build.LocalBuild.Push` is true, pushImages is set to false?

tejal29 (commented):
LGTM. 
Please rebase and add tests.

aaron-prindle:
Rebased using the BooleanOrUndefined type from:
https://github.com/GoogleContainerTools/skaffold/pull/5712

pkg/skaffold/runner/new.go
aaron-prindle: I'll add this, thanks!  Do you have any code examples of how to mock/set skaffold.yaml values or directly set the pipelines values?  Still trying to understand how to do this from poking around.

gsquared94: I think you can get away with just testing `isImageLocal` function with manually set values for `runcontext.RunContext`?

aaron-prindle: Ah makes sense, added `TestIsImageLocal` to new_test.go testing all combinations of --push=[nil|false|true] and pipeline.Build.LocalBuild.Push=[nil|false|true] w/ desired functionality

aaron-prindle:
Rebased and added tests.

Issue #5423, opened 2021-02-19
Have skaffold emit various SubtaskEvent protos


MarlonGamez:
Skaffold should emit the new SubtaskEvent protos for API v2

- [x] [build](https://github.com/GoogleContainerTools/skaffold/issues/5726) 
- [x] [test](https://github.com/GoogleContainerTools/skaffold/issues/5784)
- [x] [deploy](https://github.com/GoogleContainerTools/skaffold/issues/5782)
- [x] [statuscheck](https://github.com/GoogleContainerTools/skaffold/issues/5785)

aaron-prindle:
Friendly ping

tejal29:
Fixed by #5785

PR #5710, opened 2021-04-20
Add BuildSubtask emission for v2 API

MarlonGamez:
**Related**: #5368 #5423 

**Description**
Adds emission of BuildSubtask type message for skaffold API v2

Users can now hit the gRPC or HTTP endpoint to see BuildSubtask event types being sent out

### DIFF
--- a/pkg/skaffold/event/v2/event.go
+++ b/pkg/skaffold/event/v2/event.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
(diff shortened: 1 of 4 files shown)


pkg/skaffold/event/v2/event.go
```diff
@@ -81,6 +83,10 @@ type listener struct {
 	closed   bool
 }
 
+func GetIteration() int {
```
tejal29, line 86: Nice!!

tejal29 (approved):

PR #5712, opened 2021-04-21
Make status-check flag nillable

maggieneterval:
Fixes: N/A
**Related**: https://github.com/GoogleContainerTools/skaffold/issues/4904
**Merge before/after**: Merge before https://github.com/GoogleContainerTools/skaffold/pull/5706

**Description**
As [suggested](https://github.com/GoogleContainerTools/skaffold/pull/5706#issuecomment-822825728) by @gsquared94 and @MarlonGamez, make the `status-check` flag nullable in order to distinguish among true, false, and unset values. There is no user-facing impact of this change, but it will enable us to add a new `StatusCheck` field in the skaffold.yaml (https://github.com/GoogleContainerTools/skaffold/pull/5706) that, when set to false, will disable the healthcheck when the `status-check` flag is false or unset.

Update https://github.com/GoogleContainerTools/skaffold/pull/5706.

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -280,6 +280,7 @@ var flagRegistry = []Flag{
 		FlagAddMethod: "Var",
 		DefinedOn:     []string{"dev", "debug", "deploy", "run", "apply"},
 		IsEnum:        true,
+		NoOptDefVal:   "true",
 	},
 	{
 		Name:          "render-only",
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -277,7 +277,7 @@ var flagRegistry = []Flag{
 		Usage:         "Wait for deployed resources to stabilize",
 		Value:         &opts.StatusCheck,
 		DefValue:      true,
-		FlagAddMethod: "BoolVar",
+		FlagAddMethod: "Var",
 		DefinedOn:     []string{"dev", "debug", "deploy", "run", "apply"},
 		IsEnum:        true,
 	},
--- a/pkg/skaffold/config/types.go
+++ b/pkg/skaffold/config/types.go
@@ -53,6 +53,57 @@ func (s *StringOrUndefined) String() string {
 	return *s.value
 }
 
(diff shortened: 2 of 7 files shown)


cmd/skaffold/app/cmd/flags.go
```diff
@@ -277,7 +277,7 @@ var flagRegistry = []Flag{
 		Usage:         "Wait for deployed resources to stabilize",
 		Value:         &opts.StatusCheck,
 		DefValue:      true,
-		FlagAddMethod: "BoolVar",
+		FlagAddMethod: "Var",
```
gsquared94, line 280: we should also add `NoOptDefVal:   "true"`. This lets us invoke `skaffold foo --status-check` without the `=true` argument.

pkg/skaffold/config/types.go
```diff
@@ -53,6 +53,57 @@ func (s *StringOrUndefined) String() string {
...
+	if b == nil {
+		return ""
+	}
+	if *b {
+		return "true"
+	}
+	if !*b {
+		return "false"
+	}
+	return ""
```
gsquared94, line 100:
```suggestion
	if b == nil {
		return ""
	}
	if *b {
		return "true"
	} else {
		return "false"
	}
```

pkg/skaffold/runner/runcontext/context.go
```diff
@@ -141,6 +141,14 @@ func (rc *RunContext) Deployers() []latest.DeployType { return rc.Pipelines.Depl
 
 func (rc *RunContext) TestCases() []*latest.TestCase { return rc.Pipelines.TestCases() }
 
+func (rc *RunContext) StatusCheck() bool {
```
gsquared94, line 144: should this return `*bool` since you want to handle the `nil` case against the `skaffold.yaml` property elsewhere?

cmd/skaffold/app/cmd/flags.go
maggieneterval: Done, thanks!

pkg/skaffold/config/types.go
maggieneterval: 👍

gsquared94 (approved):
lgtm!

PR #5714, opened 2021-04-21
Improve multi-config documentation

gsquared94:


gsquared94:
http://35.236.46.78:1313/docs/design/config/#multiple-configuration-support

docs/content/en/docs/design/config.md
```diff
@@ -74,15 +74,12 @@ relative to the current directory in `helm/project`.
...
 
-## Configuration dependencies
-
-In addition to authoring pipelines in a Skaffold configuration file, we can also import pipelines from other existing configurations as dependencies. Skaffold manages all imported and defined pipelines in the same session. It also ensures all artifacts in a required configs are built prior to those in current config (provided the artifacts have dependencies defined); and all deploys in required configs are applied prior to those in current config.
+## Multiple configuration support
 
-Note that in imported configurations, files are resolved relative to the location of imported Skaffold configuration file.
-
-### Local config dependency
+A single `skaffold.yaml` file can define multiple skaffold configurations in the schema described above using the separator `---`. If these configuration objects have the `metadata.name` property then you can individually invoke a selection of configurations to run in any skaffold session.
```
viglesiasce, line 79: we should introduce this as a "module" since the flag is called that... namely a module is when you have a named skaffold config

docs/content/en/docs/design/config.md
```diff
@@ -105,7 +102,19 @@ deploy:
   # deploy definition
 ```
 
-Configurations `cfg1` and `cfg2` from the above file can be imported as dependencies in your current config, via:
+Here `cfg1` and `cfg2` are independent skaffold configurations. Running `skaffold dev` for instance will execute actions from both these configurations. You could also run `skaffold dev -m cfg1` to only activate the `cfg1` configuration and skip `cfg2`.
```
viglesiasce, line 105:
```suggestion
Here `cfg1` and `cfg2` are independent skaffold modules. Running `skaffold dev` for instance will execute actions from both these modules. You could also run `skaffold dev --module cfg1` to only activate the `cfg1` module and skip `cfg2`.
```

docs/content/en/docs/design/config.md
```diff
@@ -105,7 +102,19 @@ deploy:
...
+
+## Configuration dependencies
+
+In addition to authoring configurations in a `skaffold.yaml` file, we can also import other existing configurations as dependencies. Skaffold manages all imported and defined configurations in the same session. It also ensures all artifacts in a required config are built prior to those in current config (provided the artifacts have dependencies defined); and all deploys in required configs are applied prior to those in current config.
```
viglesiasce, line 109:
```suggestion
In addition to authoring configurations in a `skaffold.yaml` file, we can also import other existing modules as dependencies. Skaffold manages all imported and defined modules in the same session. It also ensures all artifacts in a required config are built prior to those in current config (provided the artifacts have dependencies defined); and all deploys in required modules are applied prior to those in current config.
```

docs/content/en/docs/design/config.md
```diff
@@ -105,7 +102,19 @@ deploy:
...
+
+{{< alert title="Note:" >}}
+Running `skaffold <command> -m <config-name>` will filter to the specified target configuration, but also include the transitive closure of all other configurations in its dependency graph. For instance, if a config `cfg1` imported `cfg2` as a dependency while `cfg2` imported `cfg3` and `cfg4`, then running `skaffold dev -m cfg1` would activate all of `cfg1`, `cfg2`, `cfg3` and `cfg4` and execute them in dependency order.
```
viglesiasce, line 112:
```suggestion
Running `skaffold <command> --module <config-name>` will filter to the specified target module, but also include the transitive closure of all other configurations in its dependency graph. For instance, if a config `cfg1` imported `cfg2` as a dependency while `cfg2` imported `cfg3` and `cfg4`, then running `skaffold dev --module cfg1` would activate all of `cfg1`, `cfg2`, `cfg3` and `cfg4` and execute them in dependency order.
```

docs/content/en/docs/design/config.md
```diff
@@ -105,7 +102,19 @@ deploy:
...
+{{< /alert >}}
+
+### Local config dependency
+
+Consider the same `skaffold.yaml` defined above. Configurations `cfg1` and `cfg2` from the above file can be imported as dependencies in your current config, via:
```
viglesiasce, line 117:
```suggestion
Consider the same `skaffold.yaml` defined above. Modules `cfg1` and `cfg2` from the above file can be imported as dependencies in your current config, via:
```

viglesiasce (changes requested):
Overall the info is good but i think at this point we need to bite off explicitly calling things modules. That is what the flag is and it will be easier to change the docs than the flag i believe.

docs/content/en/docs/design/config.md
gsquared94: done

docs/content/en/docs/design/config.md
gsquared94: done

docs/content/en/docs/design/config.md
gsquared94: done

docs/content/en/docs/design/config.md
gsquared94: done

docs/content/en/docs/design/config.md
gsquared94: I've changed all other occurrences of `modules` vs `configurations`. The distinction is only important for the `--module` flag which requires named-configurations. So a module is then a named configuration.

tejal29 (approved):

Issue #5713, opened 2021-04-21
RPC free port detection is broken


dat-boris:
In https://github.com/GoogleContainerTools/skaffold/commit/32a04163c3e2dc3a535e1169749ac4546509aeb2 we updated our `IsPortFree` to not look at the address.

This turn out that will cause an issue (at least on Mac).

GRPC by default listen to IPv4 (due to specifying loopback address of `127.0.0.1` in `util.Loopback`).  But the detection code, when we don't specify a loopback address, by default it listens to IPv6.

### Expected behavior

The port detection behaviour implemented in https://github.com/GoogleContainerTools/skaffold/pull/4407 should work.

```
WARN[0000] starting gRPC server on port 50060. (50051 is already in use) 
WARN[0000] starting gRPC HTTP server on port 50061. (50052 is already in use) 
```

### Actual behavior

This is now broken that we get message like:

```
initializing api server: starting gRPC server: creating listener: listen tcp 127.0.0.1:50060: bind: address already in use
```

### Information

- Skaffold version: 1.22
- Operating system: MacOS (10.14.6)
- Installed via: skaffold.dev
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta12
kind: Config
metadata:
  name: k-s
build:
  artifacts:
    - image: us.gcr.io/khan-academy/services/admin-image
      docker:
        dockerfile: Dockerfile
      # Set context folder
      context: ../..
      sync:
  local:
    push: false
    useBuildkit: true
deploy:
  kubectl:
    manifests:
      - k8s.dev.*.yaml
```

### Steps to reproduce the behavior

1. Start a skaffold server (e.g. `skaffold dev  --filename='skaffold.xxx.yaml'`)
2. Run above command in different window
3. It will run into port collision.

For example, below I can use the [script here](https://gist.github.com/dat-boris/51f42ed243f4af69e572ed10d2370b69) we can try the different `net.Listen` detection nad see the output:
```
$ go run test_port.go 50051
With address: <nil>  # <-- detect with loopback address that port is occupied
Without address: &{0xc000116080 {<nil> 0}}  # <-- incorrectly think port is not occupied - test_port.go can listen on IpV6
```
And if we looks at the process listening on 50051
```
$ lsof -i :50051
COMMAND     PID     USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
[...]

PR #5715, opened 2021-04-21
Added fix for RPC port detection

dat-boris:
Fixes: #5713
**Related**: #5590
**Merge before/after**: None

**Description**

In #5713 we discover that the new port detection logic does not apply
to GRPC startup, which default starts on Mac on IPv4 loopback.

Test plan:

Locally run multiple version of skaffold, ensure that none
of them have issues with starting up their own RPC server.

Instead of crashing, it now correctly skip previous ports used by RPC

```
WARN[0000] starting gRPC server on port 50053. (50051 is already in use) 
WARN[0000] starting gRPC HTTP server on port 50054. (50052 is already in use) 
```


dat-boris:
This is my first PR for skaffold so any review comments are much appreciated.  I know also #5590 so would be nice if we can confirm that on Windows machine this will not cause any regression (hopefully shouldn't do - it was a revert to previous behaviour for GRPC).

pkg/skaffold/util/port.go
```diff
@@ -87,37 +90,39 @@ func (f *PortSet) List() []int {
...
 func GetAvailablePort(port int, usedPorts *PortSet) int {
-	if port > 0 {
-		if getPortIfAvailable(port, usedPorts) {
-			return port
-		}
+	// We map this to allow
+	return GetAvailablePortWithAddreess(Any, port, usedPorts)
+}
+
+func GetAvailablePortWithAddreess(address string, port int, usedPorts *PortSet) int {
```
MarlonGamez, line 104:
```suggestion
func GetAvailablePortWithAddress(address string, port int, usedPorts *PortSet) int {
```

pkg/skaffold/util/port.go
```diff
@@ -87,37 +90,39 @@ func (f *PortSet) List() []int {
...
 // First, check if the provided port is available on the specified address. If so, use it.
 // If not, check if any of the next 10 subsequent ports are available.
 // If not, check if any of ports 4503-4533 are available.
 // If not, return a random port, which hopefully won't collide with any future containers
-//
-// See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt
+
+// See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt,
 func GetAvailablePort(port int, usedPorts *PortSet) int {
-	if port > 0 {
```
MarlonGamez, line 98: do we want to get rid of this check?

pkg/skaffold/util/port.go
```diff
@@ -87,37 +90,39 @@ func (f *PortSet) List() []int {
...
 
-// GetAvailablePort returns an available port that is near the requested port when possible.
 // First, check if the provided port is available on the specified address. If so, use it.
 // If not, check if any of the next 10 subsequent ports are available.
 // If not, check if any of ports 4503-4533 are available.
 // If not, return a random port, which hopefully won't collide with any future containers
-//
-// See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt
+
+// See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt,
```
MarlonGamez, line 98:
```suggestion
// See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt
```

MarlonGamez (commented):
Thanks for opening this @dat-boris ! Have a few tiny nit comments :)

pkg/skaffold/util/port.go
dat-boris: Good catch - thank you!

pkg/skaffold/util/port.go
```diff
@@ -129,16 +138,20 @@ func GetAvailablePort(port int, usedPorts *PortSet) int {
...
 func IsPortFree(p int) bool {
-	l, err := net.Listen("tcp", fmt.Sprintf(":%d", p))
+	return IsPortFreeWithAddress(Any, p)
+}
+
+func IsPortFreeWithAddress(address string, p int) bool {
+	l, err := net.Listen("tcp", fmt.Sprintf("%s:%d", address, p))
 	if err != nil {
 		return false
 	}
```
briandealwis, line 157: So this fix isn't quite enough as it still leaves the previous situation, where a local port may be available but the INADDR_ANY:port isn't available.  We should test both conditions.   Here's my suggestion:
```suggestion
	// Ensure the port is available across all interfaces
	if l, err := net.Listen("tcp", fmt.Sprintf(":%d", p));  err != nil {
		return false
	} else {
		l.Close()
	}
	if address != Any {
		// Ensure the port is available on the specific interface too
		if l, err := net.Listen("tcp", fmt.Sprintf("%s:%d", address, p));  err != nil {
			return false
		} else {
			l.Close()
		}
	}
```

briandealwis:
Sorry @dat-boris for the delay — I held off as I wanted to check how we were using `GetAvailablePort()`. Prior to #5670 we were explicitly passing the interface.  I think we should restore that aspect and I'll push that change here.

MarlonGamez (approved):
lgtm, thanks @dat-boris for the contribution :)

dat-boris:
Thank you @briandealwis @MarlonGamez for the reviews! Yeah it make sense to revert to the old `GetAvailablePort` and improve the check!

PR #5718, opened 2021-04-22
bring back error coloring

MarlonGamez:
Fixes: #5643 
**Description**
Coloring on our error output has been broken since we made changes to validating color output a while ago. This change reintroduces red coloring on errors from skaffold. I've also added more validation for color, as some terminals don't support colors.

Users should now see errors in red when running skaffold


cmd/skaffold/skaffold.go
```diff
@@ -38,7 +38,8 @@ func main() {
 		if errors.Is(err, context.Canceled) {
 			logrus.Debugln("ignore error since context is cancelled:", err)
 		} else {
-			color.Red.Fprintln(os.Stderr, err)
+			errOut := color.SetupColors(os.Stderr, color.DefaultColorCode, false)
```
briandealwis, line 41: Could we not just do the color setup entirely here in `main()`?  It seems odd to have a special case here for the error condition.

pkg/skaffold/util/term.go
```diff
@@ -35,3 +41,21 @@ func IsTerminal(w io.Writer) (uintptr, bool) {
...
+	numColors, err := strconv.Atoi(strings.TrimSpace(string(res)))
+	if err != nil {
+		return false, err
+	}
+
+	return numColors != 0, nil
+}
+
+func tputColors() ([]byte, error) {
+	return exec.Command("tput", "colors").Output()
```
briandealwis, line 60: This won't work for Windows, which normally does support ANSI sequences.

cmd/skaffold/skaffold.go
MarlonGamez: that's what I was hoping to do, but I realized that we allow some color setup using cli flags, so I'm not sure if that would be possible

pkg/skaffold/util/term.go
MarlonGamez: realized this when testing locally 😅 I can add a case to this function to check OS

cmd/skaffold/skaffold.go
briandealwis: Oh!  That's worth commenting either here or in main().

pkg/skaffold/util/term.go
```diff
@@ -35,3 +44,25 @@ func IsTerminal(w io.Writer) (uintptr, bool) {
...
+	if err != nil {
+		return false, fmt.Errorf("checking terminal colors: %w", err)
+	}
+
+	numColors, err := strconv.Atoi(strings.TrimSpace(string(res)))
+	if err != nil {
+		return false, err
+	}
+
+	return numColors != 0, nil
```
briandealwis, line 63:
```suggestion
	return numColors > 0, nil
```

On macOS:
```console
$ TERM=dumb tput colors
-1
```

pkg/skaffold/util/term.go
```diff
@@ -35,3 +44,25 @@ func IsTerminal(w io.Writer) (uintptr, bool) {
...
+
+	res, err := colors()
+	if err != nil {
+		return false, fmt.Errorf("checking terminal colors: %w", err)
+	}
+
+	numColors, err := strconv.Atoi(strings.TrimSpace(string(res)))
+	if err != nil {
+		return false, err
+	}
```
briandealwis, line 61: This code is a bit hard to understand without looking up to see that `colors()` is an alias to `tputColors()`.  I think you can just inline `tputColors()` here.  And instead use `util.RunCmdOut()` to execute so that the test can use `testutil.FakeCmd`.

MarlonGamez: done 👍🏼

pkg/skaffold/util/term.go
MarlonGamez: Thanks!

cmd/skaffold/skaffold.go
```diff
@@ -38,7 +38,10 @@ func main() {
 		if errors.Is(err, context.Canceled) {
 			logrus.Debugln("ignore error since context is cancelled:", err)
 		} else {
-			color.Red.Fprintln(os.Stderr, err)
+			// Because we use cobra flags to setup colors for the main run, we can't run SetupColors()
+			// for the entire skaffold run here, and have to call it again before we print an error
```
briandealwis, line 42:
```suggestion
			// As we allow some color setup using CLI flags for the main run, we can't run SetupColors()
			// for the entire skaffold run here. It's possible SetupColors() was never called, so call it again
			// before we print an error to get the right coloring.
```

briandealwis (approved):

Issue #5727, opened 2021-04-27
[v3] Code Refactor for skaffold v3


yuwenma:
See doc [here](https://docs.google.com/document/d/1GFy_5ya-ajjSVe1t6f7IWUlAQu35eTHAZZ1GkJBiqcA/edit?resourcekey=0-sKE-yXjuymQ0w2qqNXnTzw#heading=h.bqhpluvwrmds)

PR #5729, opened 2021-04-27
[V3] (Part 2) Add new schema to latest/v2

yuwenma:
**Related**: #5727
 
**Description**
- Define a new schema `v3alpha1` for the skaffold v2 feature.
- Create `latest/v2` dir to track `v3alpha*` schemas.

**Follow up things**
- Generate new document page for v3
- Update pkg/skaffold/schemas/versions.go to sort v1 and v2 independently.

### DIFF
--- a/pkg/skaffold/schema/latest/v2/config.go
+++ b/pkg/skaffold/schema/latest/v2/config.go
@@ -1,0 +1,177 @@
+/*
+Copyright 2021 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v2
+
+import (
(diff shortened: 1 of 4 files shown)


pkg/skaffold/schema/latest/v2/config.go
```diff
@@ -0,0 +1,174 @@
...
+// Pipeline describes a Skaffold pipeline.
+type Pipeline struct {
+	// Build describes how images are built.
+	Build latest_v1.BuildConfig `yaml:"build,omitempty"`
+
+	// Test describes how images are tested.
+	Test []*latest_v1.TestCase `yaml:"test,omitempty"`
+
+	// Render describes how the original manifests are hydrated, validated and transformed.
+	Render RenderConfig `yaml:"render,omitempty"`
```
tejal29, line 68: Sorry, just saw the config in detail now. 
Shd we instead use ?
```
manifests:
   
    validate:

    transform:

```

yuwenma: Yeah, I used the field name from the design doc. Do we get a conclusion about the render UX? I can renaming these fields once we determin the final naming (and update the design doc). For now, this is just to align with the design doc. Whom do you suggest determining (~~documenting~~) the render field names from the survey?

@tejal29 @viglesiasce

yuwenma: I updated the UX from "render" to "manifests" and keep the struct name as "render" (manifests seem to be over-used). I can rename the fields if there're some better choices coming up.

tejal29 (approved):

PR #5730, opened 2021-04-27
Default to empty secret path for Kaniko to use Workload Identity credentials

chtcvl:
Fixes: #3468

**Description**
If `GOOGLE_APPLICATION_CREDENTIALS` is set, Kaniko doesn't opt for using Workload Identity. Current setup always forces non-empty `GOOGLE_APPLICATION_CREDENTIALS` value. Proposed change make empty value default for `pullSecretPath`, unless MountPath is specified in the config. This in turn prevents `GOOGLE_APPLICATION_CREDENTIALS` from being set.

### DIFF
--- a/pkg/skaffold/build/cluster/pod.go
+++ b/pkg/skaffold/build/cluster/pod.go
@@ -122,14 +122,7 @@ func (b *Builder) kanikoPodSpec(artifact *latest_v1.KanikoArtifact, tag string) 
 }
 
 func (b *Builder) env(artifact *latest_v1.KanikoArtifact, httpProxy, httpsProxy string) []v1.EnvVar {
-	pullSecretPath := strings.Join(
-		[]string{b.ClusterDetails.PullSecretMountPath, b.ClusterDetails.PullSecretPath},
-		"/", // linux filepath separator.
-	)
 	env := []v1.EnvVar{{
-		Name:  "GOOGLE_APPLICATION_CREDENTIALS",
-		Value: pullSecretPath,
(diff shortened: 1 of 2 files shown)


chtcvl:
@googlebot I signed it!

chtcvl:
@googlebot I signed it!

pkg/skaffold/build/cluster/pod.go
```diff
@@ -155,6 +148,18 @@ func (b *Builder) env(artifact *latest.KanikoArtifact, httpProxy, httpsProxy str
 		})
 	}
 
+	// if cluster.PullSecretName  is non-empty populate secret path and use as GOOGLE_APPLICATION_CREDENTIALS
```
tejal29, line 151: @chtcvl  Shouldn't both `PullSecretName` and `PullSecretPath` be empty when using workload identity?
In case if `pullSecretPath` is specified, skaffold will create a secret in the k8 cluster.

chtcvl: @tejal29 they should be empty, but currently `pullSecretPath` is always non-empty - line 125 before change. This merge request addresses this issue - it populates `pullSecretPath` with ` strings.Join([]string{b.ClusterDetails.PullSecretMountPath, b.ClusterDetails.PullSecretPath}, "/", // linux filepath separator.` only if `PullSecretName` (which is a configuration setting) is not empty.
 		)`

tejal29 (approved):
Please rebase!

PR #5731, opened 2021-04-27
Add user flag with allowed user list to upload metrics

tejal29:
Relates to #5711

In this PR, add a hidden `--user` flag to skaffold which IDEs and client can use to export metrics. 

This user label is added to all exported metrics. 

1) For now, only few values of `user` are recorded. `gcloud`, `vsc` and `intellij`
2) An easier way to see metrics exported while developing via `EXPORT_TO_LOCAL` env variable if the skaffold binary does not contain the secret to export metrics to cloud monitoring.

Testing Notes
```
LOCAL=true make

EXPORT_TO_STDOUT=true ../../out/skaffold run --user intellij

[
	{
		"Name": "artifact-dependencies{instrumentation.name=skaffold,builder=docker,randomizer=11151,user=intellij}",
		...
	},
	{
		"Name": "deployer{instrumentation.name=skaffold,deployer=kubectl,randomizer=11151,user=intellij}",
		...
	},
	{
		"Name": "launches{instrumentation.name=skaffold,arch=amd64,command=run,config_count=1,error=OK,os=darwin,platform_type=local,randomizer=11151,user=intellij,version=v1.21.0-68-gb96f66762-dirty}",
		...
	},
	{
		"Name": "launch/duration{instrumentation.name=skaffold,arch=amd64,command=run,config_count=1,error=OK,os=darwin,platform_type=local,randomizer=11151,user=intellij,version=v1.21.0-68-gb96f66762-dirty}",
		..
	},
	{
		"Name": "run{instrumentation.name=skaffold,error=OK,randomizer=11151,user=intellij}",
		...
	},
	{
		"Name": "builders{instrumentation.name=skaffold,builder=docker,randomizer=11151,user=intellij}",
		...
	},
	{
		"Name": "artifacts{instrumentation.name=skaffold,builder=docker,randomizer=11151,user=intellij}",
		...
	}
]

```

PR #5734, opened 2021-04-27
use os.Lookup instead of os.Get

tejal29:
In #5731, I used the `os.Get` function to see if a `SKAFFOLD_EXPORT_TO_STDOUT` is defined. 

`os.Get` returns an empty string, if environment vairable is not present.  However, i just wanted to check for the presence.
 `os.Lookup` does that.


nkubala (approved):
out of curiosity, is this any faster? :)

tejal29:
> out of curiosity, is this any faster? :)

I don't think so it's faster.

Issue #5707, opened 2021-04-19
Config Imports Should be default pass through the current profile(s)


dhodun:
By default Importing a config (I think) is just running the default profile unless otherwise specified. We have multiple nestings of skaffold.yaml configs (for microservices and groups of microservices) and were expecting that profiles were inherited on the way down.

We would much prefer and expect that skaffold applies the current profile(s) unless otherwise specified.

Our root skaffold.yaml has gotten pretty messy, and is 50% profile activation code. We ultimately fixed with yaml anchors. But when we add a profile, we basically have to track down every single 'requires' section in all our skaffold yamls (not just root) to make sure we add the appropriate profiles.

This would be breaking change for anyone importing a config and just using the default profile (not many use cases for this I'm thinking?)

```yaml
# Master Skaffold
apiVersion: skaffold/v2beta12
kind: Config
metadata:
  name: backends
requires:
- path: ./microservices/authentication
  configs: [authentication]
  activeProfiles:
  - name: custom
    activatedBy: [custom]
  - name: dev
    activatedBy: [dev]

- path: ./common
  configs: [common]
  activeProfiles: 
  - name: custom
    activatedBy: [custom] 
  - name: dev
    activatedBy: [dev] 

- path: ./microservices/dashboard
  configs: [dashboard]
  activeProfiles:
  - name: custom
    activatedBy: [custom]
  - name: dev
    activatedBy: [dev]

- path: ./microservices/messages
  configs: [messages]
  activeProfiles:
  - name: custom
    activatedBy: [custom]
  - name: dev
    activatedBy: [dev]
```

briandealwis:
I do the same in [container-debug-support](https://github.com/GoogleContainerTools/container-debug-support/blob/duct-tape/skaffold.yaml#L6).

gsquared94:
I can agree with inverting the behavior: making all profiles apply to all configs recursively, and providing a mechanism to avoid selected profiles. 

We could have a flag `--recursively-applied-profiles` with values like `all` (default), `none`, `only-for=p1,p2,p3`, `not-for=p1,p2,p3`

PR #5755, opened 2021-04-29
add flag `apply-profiles-recursively`; change default behavior for multi-config projects

gsquared94:
Fixes: #5707 

**Description**

Change the bahavior of `--profile` flag for multi-config projects:
 - Profiles specified by `--profile` are propagated across all imported configs recursively by default
 - This can be disabled by setting `apply-profiles-recursively` to `false` 

  Before:
Users needed to defined this `activeProfiles` stanza for each and every config dependency in order to activate `profiles` with the same name (and behavior) across files.
```
    activeProfiles:                                     
     - name: p1                               
       activatedBy: [p1] 
```
  After:
Users no longer have to specify `activeProfiles` stanza if they want to activate the same named profile across configs.

`apply-profiles-recursively` can be changed from a bool to accept an allow and deny-list of profile names. This can be revisited if there is enough ask for that.

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -366,6 +366,15 @@ var flagRegistry = []Flag{
 		IsEnum:        true,
 	},
 	{
+		Name:          "apply-profiles-recursively",
+		Usage:         "Set to false to disable propagating profiles set by the '--profile' flag across config dependencies",
+		Value:         &opts.ApplyProfilesRecursively,
+		DefValue:      true,
+		FlagAddMethod: "BoolVar",
+		DefinedOn:     []string{"dev", "run", "debug", "deploy", "render", "build", "delete", "diagnose"},
+		IsEnum:        true,
+	},
+	{
 		Name:          "trigger",
 		Usage:         "How is change detection triggered? (polling, notify, or manual)",
 		Value:         &opts.Trigger,
--- a/docs/content/en/docs/design/config.md
+++ b/docs/content/en/docs/design/config.md
@@ -156,7 +156,9 @@ The remote config gets treated like a local config after substituting the path w
 
 ### Profile Activation in required configs
 
-Additionally the `activeProfiles` stanza can define the profiles to be activated in the required configs, via:
+Skaffold will try to apply `profiles` specified by the `--profile` flag, recursively to all configurations imported as dependencies. This behavior can be disabled by setting the `--apply-profiles-recursively` flag to `false`.
+
+You can additionally set up more granular and conditional profile activations across dependencies through the `activeProfiles` stanza:
 
 ```yaml
 apiVersion: skaffold/v2beta11
@@ -172,6 +174,8 @@ requires:
 ```
 
 Here, `profile1` is a profile that needs to exist in both configs `cfg1` and `cfg2`; while `profile2` and `profile3` are profiles defined in the current config `cfg`. If the current config is activated with either `profile2` or `profile3` then the required configs `cfg1` and `cfg2` are imported with `profile1` applied. If the `activatedBy` clause is omitted then that `profile1` always gets applied for the imported configs.
+
+
 
 {{< alert title="Follow up" >}}
 Take a look at the [tutorial]({{< relref "/docs/tutorials/config-dependencies" >}}) section to see this in action.
--- a/pkg/skaffold/parser/config.go
+++ b/pkg/skaffold/parser/config.go
@@ -164,7 +164,7 @@ func processEachConfig(config *latest_v1.SkaffoldConfig, cfgOpts configOpts, opt
 
 	var configs []*latest_v1.SkaffoldConfig
 	for _, d := range config.Dependencies {
-		newOpts := configOpts{file: cfgOpts.file, profiles: filterActiveProfiles(d, profiles), isRequired: required, isDependency: cfgOpts.isDependency}
+		newOpts := configOpts{file: cfgOpts.file, profiles: filterActiveProfiles(d, profiles, opts), isRequired: required, isDependency: cfgOpts.isDependency}
 		depConfigs, err := processEachDependency(d, newOpts, opts, r)
 		if err != nil {
 			return nil, err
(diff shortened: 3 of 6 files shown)


docs/content/en/docs/design/config.md
```diff
@@ -156,7 +156,9 @@ The remote config gets treated like a local config after substituting the path w
 
 ### Profile Activation in required configs
 
-Additionally the `activeProfiles` stanza can define the profiles to be activated in the required configs, via:
+Skaffold will try to apply `profiles` specified by the `--profile` flag, recursively to all configurations imported as dependencies. This behavior can be disabled by setting the `--apply-profiles-recursively` flag to `false`.
```
briandealwis, line 159: Maybe "propagates" instead of "recursively"?
```suggestion
Skaffold recursively applies profiles specified by the `--profile` flag to all configurations imported as dependencies. This behavior can be disabled by setting the `--apply-profiles-recursively` flag to `false`.
```

What happens if a dependent config doesn't specify that profile?  Is it silently dropped?

cmd/skaffold/app/cmd/flags.go
```diff
@@ -365,6 +365,15 @@ var flagRegistry = []Flag{
 		DefinedOn:     []string{"dev", "run", "debug", "deploy", "render", "build", "delete", "diagnose"},
 		IsEnum:        true,
 	},
+	{
+		Name:          "apply-profiles-recursively",
```
briandealwis, line 369: WDYT?

```suggestion
		Name:          "propagate-profiles",
```

pkg/skaffold/parser/config.go
```diff
@@ -179,7 +179,7 @@ func processEachConfig(config *latest_v1.SkaffoldConfig, cfgOpts configOpts, opt
 }
 
 // filterActiveProfiles selects the set of profiles to activate in the dependency config based on the current set of active profiles.
-func filterActiveProfiles(d latest_v1.ConfigDependency, profiles []string) []string {
+func filterActiveProfiles(d latest_v1.ConfigDependency, profiles []string, opts config.SkaffoldOptions) []string {
 	var depProfiles []string
 	for _, ap := range d.ActiveProfiles {
```
briandealwis, line 184: We should be including the propagated profiles here too, right?  (My suggestion above will do this.)

pkg/skaffold/parser/config.go
```diff
@@ -164,7 +164,7 @@ func processEachConfig(config *latest_v1.SkaffoldConfig, cfgOpts configOpts, opt
 
 	var configs []*latest_v1.SkaffoldConfig
 	for _, d := range config.Dependencies {
-		newOpts := configOpts{file: cfgOpts.file, profiles: filterActiveProfiles(d, profiles), isRequired: required, isDependency: cfgOpts.isDependency}
+		newOpts := configOpts{file: cfgOpts.file, profiles: filterActiveProfiles(d, profiles, opts), isRequired: required, isDependency: cfgOpts.isDependency}
```
briandealwis, line 167: `filterActiveProfiles()` should also consider the propagated profiles too, right?  So maybe it makes more sense to build up the list of active profiles here (`profiles` + `opts.Profiles` if propagating), and then add the results of `filterActiveProfiles()` which is really just looking for activated profiles?  

Something like
```suggestion
		active := profiles
		if opts.ApplyProfilesRecursively {
			for _, p := range opts.Profiles {
				if !util.StrSliceContains(active, p) {
					active = append(active, p)
				}
			}
		}
		for _, p := range filterActiveProfiles(d, active) {
			if !util.StrSliceContains(active, p) {
				active = append(active, p)
			}
		}
		newOpts := configOpts{file: cfgOpts.file, profiles: active, isRequired: required, isDependency: cfgOpts.isDependency}
```

pkg/skaffold/parser/config.go
```diff
@@ -179,7 +179,7 @@ func processEachConfig(config *latest_v1.SkaffoldConfig, cfgOpts configOpts, opt
 }
 
 // filterActiveProfiles selects the set of profiles to activate in the dependency config based on the current set of active profiles.
-func filterActiveProfiles(d latest_v1.ConfigDependency, profiles []string) []string {
+func filterActiveProfiles(d latest_v1.ConfigDependency, profiles []string, opts config.SkaffoldOptions) []string {
```
briandealwis, line 182: WDYT?

```suggestion
func filterActivatedProfiles(d latest_v1.ConfigDependency, profiles []string, opts config.SkaffoldOptions) []string {
```

pkg/skaffold/parser/config_test.go
```diff
@@ -402,6 +403,30 @@ requires:
...
+requires:
+  - path: doc1
+    configs: [cfg10]
+`}, {name: "cfg01", requiresStanza: ""}}},
+				{path: "doc1/skaffold.yaml", configs: []mockCfg{{name: "cfg10", requiresStanza: `
+requires:
+  - path: ../doc2
+    configs: [cfg21]
+`}, {name: "cfg11", requiresStanza: ""}}},
+				{path: "doc2/skaffold.yaml", configs: []mockCfg{{name: "cfg20", requiresStanza: ""}, {name: "cfg21", requiresStanza: ""}}},
```
briandealwis, line 421: Can you put this last one on a separate line?
```suggestion
				{path: "doc2/skaffold.yaml", configs: []mockCfg{{name: "cfg20", requiresStanza: ""},
				{name: "cfg21", requiresStanza: ""}}},
```

pkg/skaffold/parser/config_test.go
```diff
@@ -402,6 +403,30 @@ requires:
 				createCfg("cfg01", "image01", ".", nil),
 			},
 		},
+		{
+			description:              "recursively applied profiles",
```
briandealwis, line 407: Could you add an `activatedWhen`?

gsquared94:
I need to rethink this approach. Currently skaffold errors out if a profile name doesn't match the list of valid profiles. If we start propagating profiles, do we need to relax that assumption? Since now the profile search space is not a single config but the graph of all imported configs.

briandealwis:
Or we move profile verification after we've built up our list of active profiles.

Issue #5756, opened 2021-04-29
Feature Request: add skaffold trace information (design + plumbing)


aaron-prindle:
For getting more information about skaffold's performance and ensuring that skaffold does not have performance degradation over time, skaffold should adding trace information to skaffold commands.  Currently skaffold metrics can track the length of a `skaffold dev` session but there is no "child" trace information showing how specific actions took -  yaml parsing, builds, deploys, etc.

The steps that I have outlined for this work include:
- [x] update opentelemetry (`otel`) libs (v0.13.0 -> v0.20.0) for skaffold (to use updated APIs & trace implementation) - #5757
- [x] fully plumb trace exporters, etc. (similar to metric exporter) & plumb tracing through skaffold (usable by env var) - #5854 
- [x] add functionality/example where users can get generate & view their own trace information w/ jaeger - #5854

MarlonGamez:
trying to make triage-party happy :)

aaron-prindle:
Fixed with https://github.com/GoogleContainerTools/skaffold/pull/5854

PR #5757, opened 2021-04-29
update otel libs from v0.13.0 -> v0.20.0

aaron-prindle:
Part of #5756, adding opentelemetry trace information to skaffold commands. Updating out libs to the latest otel version adds additional useful functionality for tracing.

This approach uses go mod (updated via "go get <pkg>") and minor API changes to our otel API usage for the update.

N/A

There shouldn't be any side effects w/ this approach, the changes to otel's API were renaming/moving things (packages, functions, etc.).  The only option removed was stdout Quantile aggregation (stdout.WithQuantiles) but I do not think this will have side effects.  See https://github.com/open-telemetry/opentelemetry-go/commit/49f699d65742e144cf19b5dd28f3d3a0891bf200#diff-2b283a7fb9f9b66e31a2b51a9ae9cad3599650a633f02fea9a956c4f6a714c6c

N/A

Fixes: #nnn 

**Description**


cmd/skaffold/skaffold.go
```diff
@@ -22,16 +22,40 @@ import (
...
 
 	"github.com/sirupsen/logrus"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/GoogleContainerTools/skaffold/cmd/skaffold/app"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/color"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/instrumentation"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/stdout"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
```
tejal29, line 32: Looks like these 3rd party libs need to be in the 3rd party import block

aaron-prindle: Thanks! This is fixed now using `gci` tool

briandealwis:
Can you verify with `SKAFFOLD_EXPORT_TO_STDOUT` whether the quantiles is equivalent?  Or that it's ingested appropriately with Cloud Monitoring?

aaron-prindle:
The Quantile piece is only used for the stdout exporter which is not used in data ingestion for Cloud Monitoring.  The way quantiles were used w/ the stdout exporter was not relevant to our logging setup.  This will have no import on data export/ingestion, metrics will still work as expected.  I have verified that this setup still works as expected pointing this at my own project and firing metrics there:

Screenshot 2021-04-30 3 35 39 PM

image

briandealwis (approved):

PR #5780, opened 2021-05-04
[V3] (part 1) Change runner/v3 to runner/v2. Update v3 flag to v2

yuwenma:
**Related**: #5727 
**Merge before**: #5781

PR #5781, opened 2021-05-04
[v3] (part 2) Move the v1 runner components to `pkg/skaffold/runner/v1`.

yuwenma:
Fixes: #nnn 
**Related**: #5727 
**Merge after**: #5780 

**Description**
Changes include:
- imports changes
- public/private functions and attributes
- Simplified the runner.Tester
-


cmd/skaffold/app/cmd/runner.go
```diff
@@ -35,6 +35,7 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/parser"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner/runcontext"
+	v1 "github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner/v1"
```
MarlonGamez, line 38: do we still want to reference this as `runner`? or `runnerV1`?

yuwenma:
Rebased after #5780 is merged.

MarlonGamez (commented):
Few questions:
- Should `pkg/skaffold/runner/build_deploy.go` be moved to `pkg/skaffold/runner/v1/build_deploy.go`?
- Should `pkg/skaffold/runner/v1/build_deploy_test.go` be deleted?
- Should `pkg/skaffold/runner/prune.go` be moved as well?

Both v1 and v2 will share the `build_deploy` logic, so it does not need to be moved (It calls a Deploy (in Deploy.go) function which will be different between v1 and v2).
 
`pkg/skaffold/runner/v1/build_deploy_test.go`  is moved under v1 to make sure the build_deploy for v1 works. V2 will have its own build_deploy_test.go with different testing data.

prune.go is still needed.

cmd/skaffold/app/cmd/runner.go
yuwenma: Thank you. Done

yuwenma: The `make linters` fails if we don't use the alias "v1".

more context: golangci-lint has local-prefix enabled. https://github.com/golangci/golangci-lint/issues/209

yuwenma:
> Few questions:
> 
> * Should `pkg/skaffold/runner/build_deploy.go` be moved to `pkg/skaffold/runner/v1/build_deploy.go`?

Both v1 and v2 will share the `build_deploy` logic, so it does not need to be moved (It calls a Deploy (in Deploy.go) function which will be different between v1 and v2).
 
> * Should `pkg/skaffold/runner/v1/build_deploy_test.go` be deleted?

`pkg/skaffold/runner/v1/build_deploy_test.go`  is moved under v1 to make sure the build_deploy for v1 works. V2 will have its own build_deploy_test.go with different testing data.


> * Should `pkg/skaffold/runner/prune.go` be moved as well?
prune.go is still needed.

tejal29 (approved):
Dicussed offline

1) We want to go to a structure 
```
runner/
  - base/
  - v1/
  - v2/
```


To do that, @tejal29 and @yuwenma will explore the using
1) `util.VersionedConfig`
2) Add interfaces for `Artifact` and `Pipeline` and get rid of latest_v1.Artifact and latest_v1.Pipeline

PR #5783, opened 2021-05-04
Emit deploy subtask events for API V2

MarlonGamez:
Fixes: #5782 
**Related**: #5423 

**Description**
This PR changes some of the get deployer logic to always use the deployer mux, as it reduces the complexity/LoC needed for emitting deploy protos. It also makes the change to emit deploy protos.


gsquared94 (approved):
lgtm!

PR #5799, opened 2021-05-06
[Code style] Fix snake case import package "latest_v1" to "latestV1"

yuwenma:


MarlonGamez (dismissed):
it looks like there are a few references that got missed, think they should all be pointed out in unit test failures

tejal29 (approved):

Issue #5704, opened 2021-04-19
Skaffold Debug Portforwarding with Watch Mode stopped working after 1st iteration


quoctruong:
### Expected behavior
After I make a change when using `skaffold debug` with watch mode on, portforwarding should still work.

### Actual behavior
Whenever I try to connect to the new port established, the connection breaks.

### Information

- Skaffold version: 1.21
- Operating system: macOS (10.15.7)
- Installed via: Google Cloud SDK
- Contents of skaffold.yaml:

```yaml
# To learn more about the skaffold.yaml schema visit
# https://skaffold.dev/docs/references/yaml/
apiVersion: skaffold/v2beta4
kind: Config
build:
  tagPolicy:
    sha256: {}
  # defines where to find the code at build time and where to push the resulting image
  artifacts:
  - image: nodejs-guestbook-backend
    context: src/backend
  - image: nodejs-guestbook-frontend
    context: src/frontend
# defines the Kubernetes manifests to deploy on each run
deploy:
  kubectl:
    manifests:
    - ./kubernetes-manifests/*.service.yaml
    - ./kubernetes-manifests/*.deployment.yaml
profiles:
# use the cloudbuild profile to build images using Google Cloud Build
- name: cloudbuild
  build:
    googleCloudBuild: {}
- name: buildpacks
  build:
    artifacts:
    - image: nodejs-guestbook-backend
      context: src/backend
      buildpack:
        builder: "gcr.io/buildpacks/builder:v1"
    - image: nodejs-guestbook-frontend
      context: src/frontend
      buildpack:
        builder: "gcr.io/buildpacks/builder:v1"
```

### Steps to reproduce the behavior

1. Clone NodeJS Guestbook sample from [here](https://github.com/GoogleCloudPlatform/cloud-code-samples/tree/master/nodejs/nodejs-guestbook)
2. Run Skaffold command:
```
skaffold debug --enable-rpc -v debug --port-forward --rpc-http-port 62422 --filename /Users/quoct/guestbook-3/skaffold.yaml --default-repo gcr.io/gcloud-powershell-testing --auto-build --auto-sync --auto-deploy
```
3. Wait for skaffold to finish deploying. Run `curl http://localhost:<backendport>/json/version` where `<backendport>` is the local port to which skaffold portforward the backend container's remote port `9229`. Most likely, this will be either `9229` or `9230`. You should get a response like:
```
{
  "Browser": "node.js/v12.22.1",
[...]

PR #5804, opened 2021-05-07
Avoid aliasing in image configuration

briandealwis:
Fixes: #5704

**Description**
In examining a debug trace, we discovered that the retrieved image configuration for the unchanged image was actually different:
- round 1:
  >    DEBU[0015] Retrieved local image configuration for gcr.io/gcloud-powershell-testing/nodejs-guestbook-frontend:latest@sha256:fdf561285af541e5db2483694966368287c1baced16f2c6180bcf06cfd328c06: {false false false [] <nil>  [node app.js] [PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin NODE_VERSION=12.22.1 YARN_VERSION=1.22.5 WRAPPER_VERBOSE=debug]  sha256:ce9a93883c04ba2b41e72a472b955dbaa171a00d7defe5c5b98ed280bd9d22cd map[] [] false false false  map[] /frontend map[] false false   []} 

- round 2 (notice the entrypoint now has an `--inspect=0.0.0.0:9229`)
  > DEBU[0131] Retrieved local image configuration for gcr.io/gcloud-powershell-testing/nodejs-guestbook-frontend:latest@sha256:fdf561285af541e5db2483694966368287c1baced16f2c6180bcf06cfd328c06: {false false false [] <nil>  [node --inspect=0.0.0.0:9229] [PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin NODE_VERSION=12.22.1 YARN_VERSION=1.22.5 WRAPPER_VERBOSE=debug]  sha256:ce9a93883c04ba2b41e72a472b955dbaa171a00d7defe5c5b98ed280bd9d22cd map[] [] false false false  map[] /frontend map[] false false   []} 

The problem is due to aliasing of underlying arrays when copying slices.  We retrieve the image manifest and configuration for the image from the local docker daemon:
https://github.com/GoogleContainerTools/skaffold/blob/01feff8c4dccaa6d76072abac1fe9ed6494da866/pkg/skaffold/debug/debug.go#L114

and then copy the values, but the slices and maps retain pointers to their underlying arrays:
https://github.com/GoogleContainerTools/skaffold/blob/01feff8c4dccaa6d76072abac1fe9ed6494da866/pkg/skaffold/debug/debug.go#L120-L129

This struct is then changed with new arguments values copied into the array:
https://github.com/GoogleContainerTools/skaffold/blob/01feff8c4dccaa6d76072abac1fe9ed6494da866/pkg/skaffold/debug/transform_nodejs.go#L196-L203

The solution here is to duplicate the arrays and maps taken from the image configuration.

### DIFF
--- a/pkg/skaffold/debug/debug.go
+++ b/pkg/skaffold/debug/debug.go
@@ -158,5 +158,4 @@ func dupMap(s map[string]string) map[string]string {
 		dup[k] = v
 	}
 	return dup
-
 }
--- a/pkg/skaffold/debug/debug.go
+++ b/pkg/skaffold/debug/debug.go
@@ -119,12 +119,13 @@ func retrieveImageConfiguration(ctx context.Context, artifact *graph.Artifact, i
 
 	config := manifest.Config
 	logrus.Debugf("Retrieved local image configuration for %v: %v", artifact.Tag, config)
+	// need to duplicate slices as apiClient caches requests
 	return imageConfiguration{
 		artifact:   artifact.ImageName,
 		env:        envAsMap(config.Env),
-		entrypoint: config.Entrypoint,
-		arguments:  config.Cmd,
-		labels:     config.Labels,
+		entrypoint: dupArray(config.Entrypoint),
+		arguments:  dupArray(config.Cmd),
+		labels:     dupMap(config.Labels),
 		workingDir: config.WorkingDir,
 	}, nil
 }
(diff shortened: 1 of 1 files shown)


briandealwis:
Verified by hand with `skaffold debug --auto-build --auto-sync` in `examples/nodejs`.  Opened #5806 to make a test to verify this situation.

tejal29:
If I  understand correctly, 
In round 2, the entry point should be `node app.js`  and then in the debug workflow round 2, the `--inspect 0.0.0.0:XXXX` should be added where `XXXX` is the right port. In this case the port is `9230` because `9229` was taken by backend in round 2?

briandealwis:
That correct: on the second round, the image configuration appeared to already have an `--inspect` argument, and so the debug transforms didn't make a change.

An alternative solution would be to change the code, like `rewriteNodeCommandLine()`, to use `util.StrSliceInsert()` which does make a new copy of the provided slice.

nkubala (approved):
seems good to me, restarting a few CI jobs to see if we can get around the pesky rate limit

tejal29 (approved):

Issue #5802, opened 2021-05-07
Port forwarding is not working with namespaces with Templated Fields


itaischw:
Hi, 
I'm working with Skaffold and helm deploy.
The issue I see is that when you specify under deploy.helm.releases.namespace with a templated field,
and you want to do a user port forwarding, i.e. map port 22 to 5022, you can't use a templated field to set the 
portForward.namespace since it's not supporting templated field.
If you don't set the namespace under the portForward.namespace you will get a warning:
"Skipping the port forwarding resource ... because namespace is not specified"

### Expected behavior
The expected behaviour here is to either make the port forwarding use the custom namespace under the resource it is forwarding.
OR support templated fields for the portForward.namespace as well.

### Actual behaviour
Getting a warning "Skipping the port forwarding resource ... because namespace is not specified"
or if you set it with the templated field you will get an error for not finding the resource in the namespace without the templated field extraction 
i.e. my-namespace-{{ .USER }}

### Information

- Skaffold version: v1.19.0
- Operating system: MacOS
- Installed via: skaffold.dev |  Cloud Code 
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta8
kind: Config
build: 
  artifacts:
    - image: my-image    
deploy:  
  helm:
    releases:
      - name: my-deployment
        chartPath: helm
        version: v2        
        valuesFiles:
          - helm/values/values.yaml
        artifactOverrides:
          image: my-image
        namespace: "my-namespace-{{ .USER }}"
        createNamespace: true
        upgradeOnChange: true

portForward:
  - resourceType: StatefulSet
    resourceName: tps-monitoring
    namespace: "my-namespace-{{ .USER }}"
    port: 22
    localPort: 5022

```

### Steps to reproduce the behavior

1. Deploy any helm you want to a namespace with a templated field.
2. Try to port forward  any port and set the namespace under the portForward with the same namespace as your deployment with a templated field.

itaischw:
@aaron-prindle I created a PR to support the above, feel free to review too.

PR #5808, opened 2021-05-10
Add support for Port forwarding with namespaces with Templated Fields

itaischw:
Fixes: #5802 

**Description**
As described in #5802 user defined PortForward wasn't supporting a namespace with templated fields.
I added the support to allow users to add in their skaffold.yaml port forwarding with a templated field namespace.
So they can match a namespace in the port forward to a  namespace in the deploy section.
Since the deply.helm.releases.namespace was already supported that brought some inconsistency and not allowing to port forward deployment with a namespace with templated fields in it

> deploy:  
  helm:
    releases:
      - name: my-deployment
        chartPath: helm
        version: v2        
        valuesFiles:
          - helm/values/values.yaml
        artifactOverrides:
          image: my-image
        **namespace: "my-namespace-{{ .FOO }}"**

> portForward:
   resourceType: StatefulSet
      resourceName: resourceName
    **namespace: "my-namespace-{{ .FOO }}"**
    port: 22
    localPort: 5022

I also added the new field support to the documentation under the templated field section.

### DIFF
--- a/pkg/skaffold/kubernetes/portforward/resource_forwarder.go
+++ b/pkg/skaffold/kubernetes/portforward/resource_forwarder.go
@@ -69,12 +69,21 @@ func (p *ResourceForwarder) Start(ctx context.Context, namespaces []string) erro
 		for _, pf := range p.userDefinedResources {
 			if pf.Namespace == "" {
 				pf.Namespace = namespaces[0]
+			} else {
+				err := applyNamespaceWithTemplate(pf)
+				if err != nil {
+					return err
+				}
 			}
 		}
 	} else {
 		var validResources []*latestV1.PortForwardResource
 		for _, pf := range p.userDefinedResources {
 			if pf.Namespace != "" {
+				err := applyNamespaceWithTemplate(pf)
+				if err != nil {
+					return err
+				}
 				validResources = append(validResources, pf)
 			} else {
 				logrus.Warnf("Skipping the port forwarding resource %s/%s because namespace is not specified", pf.Type, pf.Name)
@@ -92,6 +101,15 @@ func (p *ResourceForwarder) Start(ctx context.Context, namespaces []string) erro
 		serviceResources = found
 	}
 	p.portForwardResources(ctx, append(p.userDefinedResources, serviceResources...))
+	return nil
+}
+
+func applyNamespaceWithTemplate(resource *latestV1.PortForwardResource) error {
+	namespace, err := util.ExpandEnvTemplateOrFail(resource.Namespace, nil)
+	if err != nil {
+		return fmt.Errorf("cannot parse the namespace template on user defined port forwarder: %w", err)
+	}
+	resource.Namespace = namespace
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


itaischw:
> @googlebot I signed it!

MarlonGamez (commented):
thanks for opening this @itaischw! This is looking good, I just have one tiny nit comment. Either way this should be good to merge. I'm rerunning CI on this as it looks like it flaked but I can approve once it's green

itaischw:
@MarlonGamez can you assist with understanding why this build failed? seems like something unrelated to my code

pkg/skaffold/kubernetes/portforward/resource_forwarder.go
```diff
@@ -67,14 +67,23 @@ func NewUserDefinedForwarder(entryManager *EntryManager, userDefinedResources []
 func (p *ResourceForwarder) Start(ctx context.Context, namespaces []string) error {
 	if len(namespaces) == 1 {
 		for _, pf := range p.userDefinedResources {
-			if pf.Namespace == "" {
+			if pf.Namespace != "" {
+				err := applyNamespaceWithTemplate(pf)
+				if err != nil {
```
briandealwis, line 72: Minor nit: since you're not using `err` elsewhere you can inline this and save a line
```suggestion
				if err := applyNamespaceWithTemplate(pf); err != nil {
```

pkg/skaffold/kubernetes/portforward/resource_forwarder.go
```diff
@@ -67,14 +67,23 @@ func NewUserDefinedForwarder(entryManager *EntryManager, userDefinedResources []
...
+			} else {
 				pf.Namespace = namespaces[0]
 			}
 		}
 	} else {
 		var validResources []*latestV1.PortForwardResource
 		for _, pf := range p.userDefinedResources {
 			if pf.Namespace != "" {
+				err := applyNamespaceWithTemplate(pf)
+				if err != nil {
```
briandealwis, line 84: Can inline this and save a line
```suggestion
				if err := applyNamespaceWithTemplate(pf); err != nil {
```

itaischw: @briandealwis done!

itaischw:
@tejal29 can you add again the kokoro:run so it will run again, I can't seems to be able to add labels to the PR

pkg/skaffold/kubernetes/portforward/resource_forwarder.go
itaischw: @briandealwis can you approve?

MarlonGamez (approved):
@itaischw thanks for working on this 👍🏼 lgtm

pkg/skaffold/kubernetes/portforward/resource_forwarder.go
briandealwis: I'm a Smalltalker by background, and Smalltalk encourages small methods.  But separate functions/methods do impose a bit of cognitive overhead as it requires switching contexts when reading through the code to understand what it's doing.  In this case, the only real benefit is a shared text string, which isn't a big win.  And when I first looked at the function itself, I remember checking whether you were using it with other objects, particularly with live objects within the skaffold pipeline.

Anywys, I won't object to it.

PR #5815, opened 2021-05-11
fix skaffold test to use minikube docker-env docker context

### LINKED ISSUE #5696: Skaffold test should be run with the Docker socket of minikube when available.
### Expected behavior
When a user has minikube, we do some trickery to ensure that their images are built with minikube's docker daemon by using the values from `minikube docker-env`. We should do the same when running custom command tests as it will be common for images from the build phase to be used for testing.

### Actual behavior
The hosts docker socket is used for any docker run commands in a custom test causing dependent images to be missing.

### Information

- Skaffold version:  v1.21.0
- Operating system: MacOS arm64
- Installed via: Cloud Code
- Contents of skaffold.yaml:

```yaml
  
apiVersion: skaffold/v2beta13
kind: Config
build:
  artifacts:
    - image: test-image
      context: .
      docker:
        dockerfile: hack/test-image/Dockerfile
test:
  # Unit tests, remotely vis Cloud Build
  - image: sample-app-backend-dev
    custom:
    - command: docker run $IMAGE go test
  - image: test-image
    custom:
    - command: docker run $IMAGE ./hack/validate.sh
```

1. Clone https://github.com/viglesiasce/sample-app/tree/base-pattern
2. `minikube start`
3. `skaffold dev`
4. Tests will fail because they cant find the images

aaron-prindle:
Fixes #5696, fixing skaffold test code to use minikube docker env in test environement command context when a minikube cluster is being used.

This approach re-uses the minikube detection and docker context setup used by skaffold builders which is in pkg/skaffold/docker/client.go.  This way the build environment env vars will be identicla for both the build and test without having to duplicate code. The downside here is that the method used creates a docker client as well that is not used for test but this is likely ok because the docker client is a singleton (sync.Once) and as such should not require additional overhead.

I considered copying over some of the minikube specific docker-env logic from pkg/skaffold/docker/client.go and removing the client creation pieces but essentially the minikube logic would be in two places and build and test would use different image env related code when the issue is mainly around them not using the same code here.  It might make sense to refactor the docker client code to smaller pieces but I didn't want to change the public API and I think what is here should work fine.

There should be no side affects to this approach.  This should make it so the env vars used for build and test are aligned, most importantly for the minikube case.

N/A


tejal29 (approved):

pkg/skaffold/test/custom/custom.go
```diff
@@ -190,6 +193,14 @@ func (ct *Runner) getEnv(imageTag string) ([]string, error) {
 	}
 
 	envs = append(envs, util.OSEnviron()...)
+
+	// add minikube docker env vars to command env context if minikube cluster detected
+	localDaemon, err := docker.NewAPIClient(ct.cfg)
+	if err != nil {
+		return nil, fmt.Errorf("getting docker client information: %w", err)
```
tejal29, line 200: Lets make this an actionable error message.
See here to add a new distinct error code and add it here https://github.com/GoogleContainerTools/skaffold/blob/master/DEVELOPMENT.md#adding-actionable-error-messages-to-code

Depending on the error message, if `minikube docker-env` returns error, the actionable messages could be
1) If minikube is stopped and not running.  You can try reuse the error code [`CHECK_MINIKUBE_STATUS `](https://github.com/GoogleContainerTools/skaffold/blob/master/proto/enums/enums.proto#L454)
2) docker is stopped and not running.

tejal29: I am ok doing this in a follow up PR

aaron-prindle: Will do in a followup PR, tracking the work in #5820

Issue #5758, opened 2021-04-29
Add `skaffold mutate` command to support minimum features required for CC 


tejal29:
In this task

- [ ] P0 Support the use case to add a profile with new build env `gcb`  
- [ ] P1 Expose if a particular Dockerfile Artifact can be run via Kaniko. Maybe use a Metadata API maybe in `BuildMetadata`
      Not all DockerArtifacts can be run with kaniko. Some features are not supported in kaniko.
       See notes here https://github.com/tejal29/skaffold/pull/7/files#diff-c3bed1e9b531e926022ffef7a604e2045e899afe03ca1be23de8a3be568c9f8cR398
- [ ] P1 Support the use case to add a profile with new build env cluster and artifact type kaniko like this.
   ```
  artifacts:
    - image: skaffold-example
      kaniko:
        cache: {}
  cluster:
    namespace: default
```
- [ ] P1 Automatically plumb through gcloud auth if image push is true & registry is GCR.
   sample [code](https://github.com/tejal29/skaffold/pull/7/files#diff-122dbdbd5378fcfe71d3253fc132095e1756a908d4cfb406c4da5cba7b944edaR191)

tejal29:
@marlon-gamez do you want to close this issue and create a new issue for the remaining tasks?

tejal29:
Closing this. Please create individual issues for the remaining tasks. 

Looks like using Kaniko is a good to have feature but not required.

PR #5824, opened 2021-05-12
add `skaffold inspect build-env add googleCloudBuild` command

gsquared94:
Related: #5758
Add commands `skaffold inspect build-env add` and `skaffold inspect build-env add googleCloudBuild`


MarlonGamez (commented):
Lookin good, I have a couple questions relating to organization:
- do we want to store all flags for the different `build-env add` commands in that struct? It might get a bit bloated later.
- do we want to create a package for `pkg/skaffold/inspect/build_env`? The build_env.go file will probably be very bloated if we put all `build_env` functionality in there. wdyt?

gsquared94:
> * do we want to store all flags for the different `build-env add` commands in that struct? It might get a bit bloated later.

we can split it later if it gets too big
> * do we want to create a package for `pkg/skaffold/inspect/build_env`? The build_env.go file will probably be very bloated if we put all `build_env` functionality in there. wdyt?

done!

MarlonGamez (approved):
lgtm 👍🏼

Issue #5369, opened 2021-02-08
Create an application Log event endpoint


tejal29:
1. Create application log event endpoint. 
2.  Create an application log event as per the definition
```protobuf
message ApplicationLogEvent {
   string containerName = 1; // container that the log came from
   string podName = 2; // pod that the log came from
   string message = 3; // contents of the log
}
```

3. All application logs will be available via this end point..

MarlonGamez:
starting on this

PR #5829, opened 2021-05-12
emit applicationLogEvents on event endpoint

MarlonGamez:
**Related**: #5369 

**Description**
This PR adds emission of `applicationLogEvent`s through the event API V2.

User's will now be able to get logs from their application using the event API.

Following the original design doc, this functionality should exist on its own endpoint. I plan to do that in a follow up PR to keep the size of this one relatively small.

### DIFF
--- a/pkg/skaffold/event/v2/application_logs.go
+++ b/pkg/skaffold/event/v2/application_logs.go
@@ -20,12 +20,9 @@ import proto "github.com/GoogleContainerTools/skaffold/proto/v2"
 
 func ApplicationLog(podName, containerName, message string) {
 	handler.handleApplicationLogEvent(&proto.ApplicationLogEvent{
-		ContainerName:        podName,
-		PodName:              containerName,
-		Message:              message,
-		XXX_NoUnkeyedLiteral: struct{}{},
-		XXX_unrecognized:     nil,
-		XXX_sizecache:        0,
+		ContainerName: podName,
+		PodName:       containerName,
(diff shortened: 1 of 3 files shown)


pkg/skaffold/event/v2/application_logs.go
```diff
@@ -0,0 +1,35 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v2
+
+import proto "github.com/GoogleContainerTools/skaffold/proto/v2"
+
+func ApplicationLog(podName, containerName, message string) {
```
tejal29, line 21: I forgot if application logs have dev-iteration id in it.  Do you remember if we need to preserve the iteration id?

tejal29 (commented):
LGTM except for 1 quick question

pkg/skaffold/event/v2/application_logs.go
MarlonGamez: the original design doesn't have them, but I think we definitely should make the change to include them to make things easier for the CC team. I can open a todo issue for that and do it in a follow up PR to avoid cluttering this one

MarlonGamez: https://github.com/GoogleContainerTools/skaffold/issues/5837

tejal29 (approved):

PR #5846, opened 2021-05-14
propagate profiles across imported configs by default; disable using `propagate-profiles` flag

gsquared94:
Fixes: #5707 
**Related**: #5755 

**Description**

Change the behavior of `--profile` flag for multi-config projects:
 - Profiles specified by `--profile` are propagated across all imported configs recursively by default
 - This can be disabled by setting `--propagate-profiles` to `false` 

  Before:
Users needed to defined this `activeProfiles` stanza for each and every config dependency in order to activate `profiles` with the same name (and behavior) across files.
```
    activeProfiles:                                     
     - name: p1                               
       activatedBy: [p1] 
```
  After:
Users no longer have to specify `activeProfiles` stanza if they want to activate the same named profile across configs.

`propagate-profiles` can be changed from a bool to accept an allow and deny-list of profile names. This can be revisited if there is enough ask for that.

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -369,7 +369,7 @@ var flagRegistry = []Flag{
 	},
 	{
 		Name:          "propagate-profiles",
-		Usage:         "Set to false to disable propagating profiles set by the '--profile' flag across config dependencies",
+		Usage:         "Setting '--propogate-profiles=false' disables propagating profiles set by the '--profile' flag across config dependencies. This mean that only profiles defined directly in the target 'skaffold.yaml' file are activated.",
 		Value:         &opts.PropagateProfiles,
 		DefValue:      true,
 		FlagAddMethod: "BoolVar",
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -368,6 +368,15 @@ var flagRegistry = []Flag{
 		IsEnum:        true,
 	},
 	{
+		Name:          "propagate-profiles",
+		Usage:         "Set to false to disable propagating profiles set by the '--profile' flag across config dependencies",
+		Value:         &opts.PropagateProfiles,
+		DefValue:      true,
+		FlagAddMethod: "BoolVar",
+		DefinedOn:     []string{"dev", "run", "debug", "deploy", "render", "build", "delete", "diagnose"},
+		IsEnum:        true,
+	},
+	{
 		Name:          "trigger",
 		Usage:         "How is change detection triggered? (polling, notify, or manual)",
 		Value:         &opts.Trigger,
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -369,7 +369,7 @@ var flagRegistry = []Flag{
 	},
 	{
 		Name:          "propagate-profiles",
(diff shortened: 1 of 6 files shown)


cmd/skaffold/app/cmd/flags.go
```diff
@@ -367,6 +367,15 @@ var flagRegistry = []Flag{
 		DefinedOn:     []string{"dev", "run", "debug", "deploy", "render", "build", "delete", "diagnose"},
 		IsEnum:        true,
 	},
+	{
+		Name:          "propagate-profiles",
+		Usage:         "Set to false to disable propagating profiles set by the '--profile' flag across config dependencies",
```
aaron-prindle, line 372: nit. I personally find `Set to false to disable propagating profiles set by the '--profile' flag across config dependencies` a bit hard to read/understand.  One alternative (that is a bit long haha):
`Setting --propogate-profiles=false disables propagating profiles set by the '--profile' flag across config dependencies.  This makes the --profile flag value that is set used only for dependencies where a profile is not defined in the skaffold.yaml directly`

docs/content/en/docs/design/config.md
```diff
@@ -156,7 +156,9 @@ The remote config gets treated like a local config after substituting the path w
 
 ### Profile Activation in required configs
 
-Additionally the `activeProfiles` stanza can define the profiles to be activated in the required configs, via:
+Profiles specified by the `--profile` flag are also propagated to all  configurations imported as dependencies, if they define them. This behavior can be disabled by setting the `--propagate-profiles` flag to `false`.
+
+You can additionally set up more granular and conditional profile activations across dependencies through the `activeProfiles` stanza:
```
aaron-prindle, line 161: nit: it might helpful to add an example of how --propogate-profiles can be used w/ skaffold /examples directory if a user wants to see it in action easily

pkg/skaffold/parser/config_test.go
```diff
@@ -1205,6 +1206,86 @@ requires:
...
+`}, {name: "cfg01", requiresStanza: ""}}},
+				{path: "doc1/skaffold.yaml", configs: []mockCfg{{name: "cfg10", requiresStanza: `
+requires:
+  - path: ../doc2
+    configs: [cfg21]
+`}, {name: "cfg11", requiresStanza: ""}}},
+				{path: "doc2/skaffold.yaml", configs: []mockCfg{{name: "cfg20", requiresStanza: ""}, {name: "cfg21", requiresStanza: ""}}},
+			},
+			expected: func(base string) []*latestV1.SkaffoldConfig {
+				return []*latestV1.SkaffoldConfig{
```
aaron-prindle, line 1281: nit: config_test.go is a bit long atm with 56 test cases in `TestGetAllConfigs`.  From looking at it, this is likely justified as we want the full set of combos for each config option and the test framework is the same for all of them but wanted to point it out as it might be hard for newer users to read.

aaron-prindle (approved):
LGTM

cmd/skaffold/app/cmd/flags.go
gsquared94: done

pkg/skaffold/parser/config_test.go
gsquared94: I agree but didn't find any obvious way to simplify it. Do you have any ideas how this test can be better structured?

docs/content/en/docs/design/config.md
gsquared94: will take up as a followup item.

PR #5868, opened 2021-05-19
Move application logs to their own endpoint for API V2

MarlonGamez:
Fixes: #5369 

**Description**
Changes applicationLogEvents to have their own endpoint, rather than being put out on the event endpoint.
There are some refactors here to avoid code duplication and support future code that will be similar as well.


tejal29:
Assigning to myself since Nick is out this week

tejal29 (approved):

Issue #5841, opened 2021-05-13
Docs need an example of how to use .build.docker.buildArgs


torenware:
### Expected behavior
Based on [the docs on the site](https://skaffold.dev/docs/references/yaml/#build-artifacts-docker-buildArgs) I'd expect the following fragment to be legal. It isn't.

```yaml
    - image: my-repo/ticketer-client
      context: client
      docker:
        dockerfile: Dockerfile-production
        buildArgs:
          NEXT_PUBLIC_STRIPE_PUBLIC_KEY: {{ .NEXT_PUBLIC_STRIPE_PUBLIC_KEY }} 
          NEXT_PUBLIC_BASE_URI: ingress-nginx-controller.ingress-nginx
```

### Actual behavior
I get a parsing error:

```
parsing skaffold config: error parsing skaffold configuration file: unable to re-marshal YAML without dotted keys: unable to parse YAML: yaml: invalid map key: map[string]interface {}{".NEXT_PUBLIC_STRIPE_PUBLIC_KEY":interface {}(nil)}
```
Native English speaker I am, but that golang err, well, is not English.

### Information

- Skaffold version: 1.22 and also 1.24
- Operating system: MacOS Mojave 10.14.6
- Installed via: curl from website.
- Contents of skaffold.yaml:

relevant section included; removing the lines

```
        buildArgs:
          NEXT_PUBLIC_STRIPE_PUBLIC_KEY: {{ .NEXT_PUBLIC_STRIPE_PUBLIC_KEY }} 
          NEXT_PUBLIC_BASE_URI: ingress-nginx-controller.ingress-nginx
```

restores the file to working order, but does not pass on the buildArgs to my Dockerfile.

### Steps to reproduce the behavior

Original version of the repo (w/o the buildArg related lines) is at https://github.com/torenware/grider-micro-services.
skaffold.yaml is at the top level.

I have a .env.local file that looks like this:

```
NEXT_PUBLIC_STRIPE_PUBLIC_KEY=pk_test_51ImnsGKlT5z4...
NEXT_PUBLIC_BASE_URI=ingress-nginx-controller.ingress-nginx
NEXT_TELEMETRY_DISABLED=1
```
which is passed to skaffold as

``` bash
source client/.env.local && skaffold dev
```

briandealwis:
@torenware you need to escape your templated value as [it is not a plain scalar](https://www.yaml.info/learn/quote.html#noplain).  So something like the following should be fine:
```
[...]

PR #5872, opened 2021-05-20
Fix yaml reference rendering for object-type examples

briandealwis:
Fixes: #5841

**Description**
Our JS rendering of the `skaffold.yaml` schemas was swallowing the key-part when walking the key-value pairs in an object-style example.  The only examples of this type are the `buildArgs` fields of our builders.

**Before**
Screen Shot 2021-05-20 at 10 05 30 AM

**After**
Screen Shot 2021-05-20 at 10 05 44 AM


MarlonGamez (approved):
nice! Lgtm 👍🏼

Issue #5421, opened 2021-02-19
Skaffold schema objects are not marked as type=object


briandealwis:
Our JSON schema type definitions are not marked as `"type": "object"` and so array references etc can be filled in with numbers or other values.  For example, the following is deemed to be valid:

```
portForward:
 - "foo"
```

Our [schema generator](http://github.com/GoogleContainerTools/skaffold/blob/master/hack/schemas/main.go) needs to add a `type` field to the generated objects.

tejal29:
@tstromberg this bug is showing as p2 violation. It was created 4 days ago.
![image](https://user-images.githubusercontent.com/1534539/108921020-92c34580-75ea-11eb-934c-b1cc5a9cdded.png)

briandealwis:
Specifically: a definition without a type can hold any type.  Our [PortForwardResource](https://github.com/GoogleContainerTools/skaffold/blob/4eb8930752cc27d8016b61f7595440f7d8aaa8cc/docs/content/en/schemas/v2beta14.json#L2446-L2497) can be anything.

PR #5876, opened 2021-05-20
Schemas should emit type=object for object

briandealwis:
Fixes: #5421

**Description**
This PR causes the JSON schema generator to mark struct-types as `"type":"object"` in the resulting JSON scheme.  These `type` statements act as constraints.  Without this type, these definitions can be anything: objects, strings, numbers, null, etc.

### DIFF
--- a/docs/content/en/schemas/v1.json
+++ b/docs/content/en/schemas/v1.json
@@ -40,6 +40,7 @@
         "command"
       ],
       "additionalProperties": false,
+      "type": "object",
       "description": "criteria by which a profile is auto-activated.",
       "x-intellij-html-description": "criteria by which a profile is auto-activated."
     },
@@ -47,6 +48,7 @@
       "required": [
         "image"
       ],
+      "type": "object",
       "anyOf": [
         {
           "properties": {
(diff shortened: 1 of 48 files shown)


briandealwis:
Marking as do-not-merge as I'm trying to test the LSP to ensure it doesn't break anything.

aaron-prindle:
Seems that schema tests will need to be modified as well from Travis:
```
=== Failed Tests ===
/hack/schemas/TestGenerators
--- FAIL: TestGenerators (0.04s)
/hack/schemas/TestGenerators/inline
    main_test.go:73: string differ (-got, +want):   (
          	"""
          	... // 51 identical lines
          	      ],
          	      "additionalProperties": false,
        - 	      "type": "object",
          	      "description": "for testing the schema generator.",
          	      "x-intellij-html-description": "for testing the schema generator."
          	... // 4 identical lines
          	"""
          )
```

Additionally is there any concern around modifying old versions of the schema vs schemas going forward?  I don't think many users would be using/relying on incorrect definitions of objects but wanted to state it explicitly here.

aaron-prindle:
Is there any concern around modifying old versions of the schema vs schemas going forward?  I don't think many users would be using/relying on incorrect definitions of objects but wanted to state it explicitly here.

briandealwis:
I updated the schema files and rebuilt the LSP and it seems happy.

I don't think we need to be concerned about updating the past schemas: adding this field just constrains what could be used. The worst case is that invalid files will now be flagged as invalid where as they may have appeared to be valid previously. Skaffold would still have failed to load them.

aaron-prindle (approved):
LGTM

Issue #5871, opened 2021-05-20
Kaniko buildArgs are evaluated with environment but Docker buildArgs are not


briandealwis:
Noticed when fixing #5841 that the Kaniko `buildArgs` supports templating with an environment variable:
https://github.com/GoogleContainerTools/skaffold/blob/6c57b21b31d35e646df0352692e25af1c91d0aca/pkg/skaffold/build/cluster/kaniko.go#L51

But the Docker buildArgs does not:
https://github.com/GoogleContainerTools/skaffold/blob/6c57b21b31d35e646df0352692e25af1c91d0aca/pkg/skaffold/build/gcb/docker.go#L71

Fix would be to
- [ ] Change EvalBuildArgs to EvalBuildArgsWithEnv
- [ ] Fix the schema examples in pkg/skaffold/schema/latest to match the Kaniko builder example

torenware:
Just posted WIP.  I do not believe this fixes the underlying issue, since experiments with the test binary do not do the environment variable substitution; the behavior is the same as before the change.  In addition, there were issues with make test that suggest that whatever mechanism is used to mock Dockerfiles was not working for me correctly. You'll notice I checked in a fake Dockerfile with some ARG statements, which for whatever reason *did* allow some of those tests to pass, save one.

For me to go further, I expect I'd need some way to get visibility into what's actually executing. I'd need to "printf the s**t out it".  Are there docs on how to get access to to what's written to the logrus logs?  My reading of the code suggests that the right routines are being invoked, but still, no substitution of the `{{ .ENV_VAR }}` style templates get down to Docker.

torenware:
Some questions on linting. I notice that I failed the linter:

```
INFO [runner] linters took 2m22.825227725s with stages: goanalysis_metalinter: 2m2.267780476s, unused: 20.365257514s 

pkg/skaffold/build/gcb/docker.go:23: File is not `gci`-ed with -local github.com/GoogleContainerTools/skaffold (gci)

	cloudbuild "google.golang.org/api/cloudbuild/v1"

pkg/skaffold/build/gcb/docker.go:28: File is not `gci`-ed with -local github.com/GoogleContainerTools/skaffold (gci)

	v1 "k8s.io/api/core/v1"

pkg/skaffold/build/gcb/docker.go:24: File is not `goimports`-ed with -local github.com/GoogleContainerTools/skaffold (goimports)
```
Are there tools I need to run before checking in code? Where's a good doc to look at for this?

[...]

Issue #5877, opened 2021-05-20
Consolidate to single style for specifying environment-style parameters (env, buildArgs)


briandealwis:
We have three styles of specifying environment-style variables in the `skaffold.yaml`.  It's inconsistent and confusing for users (#5841).

### Map-style
Seen in Docker and Kaniko's `buildArgs`:
```
        // BuildArgs are arguments passed to the docker build.
        // For example: `{"key1": "value1", "key2": "value2"}`.
        BuildArgs map[string]*string `yaml:"buildArgs,omitempty"`
```
The use of a pointer here allows using YAML's `null` to unset a value
```
  docker:
     buildArgs:
       name: value
       other: null
```

### Array-style
Seen in Buildpacks's `env`:
```
        // Env are environment variables, in the `key=value` form,  passed to the build.
        // Values can use the go template syntax.
        // For example: `["key1=value1", "key2=value2", "key3={{.ENV_VARIABLE}}"]`.
        Env []string `yaml:"env,omitempty"`
```
 
### Key-Value pairs (`k8s.io/api/core/v1.EnvVar`)

Seen in Kaniko's `env`:
```
        // Env are environment variables passed to the kaniko pod.
        // It also accepts environment variables via the go template syntax.
        // For example: `[{"name": "key1", "value": "value1"}, {"name": "key2", "value": "value2"}, {"name": "key3", "value": "'{{.ENV_VARIABLE}}'"}]`.
        Env []v1.EnvVar `yaml:"env,omitempty"`
```

---

We have other fields that are in a similar state:
- Kaniko `label` is a `map[string]*string`
- Cluster's `annotations` is a `map[string]string`

PR #5884, opened 2021-05-21
Issue #5871 -- Doc change and unit test to verify docker buildArgs support templates

torenware:
WIP #5871 

**Description**

First pass on the issue.  I'd expect that acceptance tests will fail when I create the PR.

I've switched  EvalBuildArgs to EvalBuildArgsWithEnv, merging in enough from the Kaniko implementation so that docker.go actually compiles. 

Some issues:

* Working with the built skaffold, I have verified that this does NOT resolve the issue. Behavior in the app is the same as before the change as near as I can tell.
* Odd behavior in `make test`:  a number of docker related tests failed, with errors related to being unable to find Dockerfile.  As an experiment, I added a simple Dockerfile to the same directory, and to my surprise, after adding the missed ARG statements, all of the tests but one now pass. FWIW :-) I'd guess there's a mocking mechanism for Dockerfiles which is not being invoked for some reason. Couldn't figure that out.
* If someone can tell me how to run __only__ the tests for /pkg/skaffold/build/gcb, much obliged. Tried a couple of the obvious things with prefixing `SKAFFOLD_TEST_PACKAGES` to the invocation of make, still got a complete set of tests.


pkg/skaffold/schema/latest/v1/config.go
```diff
@@ -1208,6 +1208,11 @@ type DockerArtifact struct {
 	// For example: `{"key1": "value1", "key2": "value2"}`.
```
briandealwis, line 1208: This example should include a templated field.

pkg/skaffold/schema/latest/v1/config.go
```diff
@@ -1208,6 +1208,11 @@ type DockerArtifact struct {
...
 	BuildArgs map[string]*string `yaml:"buildArgs,omitempty"`
 
+	// Env are environment variables passed to the docker build.
+	// It also accepts environment variables via the go template syntax.
+	// For example: `[{"name": "key1", "value": "value1"}, {"name": "key2", "value": "value2"}, {"name": "key3", "value": "'{{.ENV_VARIABLE}}'"}]`.
+	Env []v1.EnvVar `yaml:"env,omitempty"`
```
briandealwis, line 1214: Docker doesn't propagate environment fields, so this doesn't really make sense to be included here.

briandealwis (changes requested):
I don't think Docker artifacts should support an `env` field: Kaniko does support setting environment variables for the build's pod, but Docker does not propagate environment variables into the container: that's the purpose of the build-args.

Plus `v1.EnvVar` is a Kubernetes object and shouldn't be included in the docker/GCB build — I opened #5877 to refactor this.

That will simplify this implementation.

pkg/skaffold/schema/latest/v1/config.go
torenware: The irony is that there's no need to modify the config for Docker artifacts. At least running Skaffold locally, the file `pkg/skaffold/build/gcb/docker.go` is not called at all. Instead, most of the work is done in pkg/skaffold/docker/image.go.  func (l *localDaemon) Build() calls EvalBuildArgs(), which turns out to be just fine; the missing env parameter isn't really well named, since it's used for non-environmental variables that need to be injected into templates.

I learned this by setting up delve with VS Code, and once I put in a .env file into the top skaffold/ directory of the repo, to my shock the templates took up the environment automatically. master / HEAD is not broken in this respect. But outside of the debugger, broken. 

I believe the problem is with how skaffold populates the environment. I can't get .env files to work with skaffold, pretty much no matter where I put them. Do you know how .env files are implemented? I can't find any loader code for the life of me; neither of the two common golang modules used for loading env files is called anywhere, and I could not find any code that was parsing the files and inserting them into the process environment. I'm guessing that VS Code exports the file in code, and then calls fork(), thus passing on the environment pristine. Whatever skaffold is currently doing does not work as well.

torenware: OK, my theory seems sound. I wrote a script to parse my .env file and exported the variable. Do that, and calling skaffold dev resolves my build arg template. 

So how can we fix the way that skaffold initializes its environment?

pkg/skaffold/schema/latest/v1/config.go
torenware: The latest commit does this.  I've also removed the schema change to DockerArtifact.

torenware: There's an issue with modifying [pkg/skaffold/schema/latest/v1/config.go](./pkg/skaffold/schema/latest/v1/config.go), since it annoys one of the "check" scripts. Doesn't want it modified since latest is assigned to an already released version.

pkg/skaffold/schema/latest/v1/config.go
torenware: I added a non-trivial test of the templating feature. The style of the test is a little different from most of the table driven tests used for docker, but was needed to actually exercise `EvalBuildArgs()`.  Short version: the feature works fine in 1.24, and likely earlier than that.

pkg/skaffold/schema/latest/v1/config.go
torenware: "but Docker does not propagate environment variables into the container: that's the purpose of the build-args."  The use case here is NextJS, which has "magic" environment variables that get built into server-side rendereding containers. Which is a PITA. But it is what it is.

torenware: Build now passes; @briandealwis's requested changes have been made.

pkg/skaffold/schema/latest/v1/config.go
torenware: This PR meandered too much. I've taken the final version and put it in as [PR-5901](https://github.com/GoogleContainerTools/skaffold/pull/5901), which should be much easier to understand. Leaving this one open for now so people know the background of it, since this is my first issue in the Skaffold repo.

briandealwis:
Closing — thanks for digging into this @torenware.

PR #5891, opened 2021-05-21
Failing to delete source archive should not fail GCB builds

briandealwis:
@dhodun mentioned that an error deleting a source archive from a GCB build failed the overall build.
```
cleaning up source tar after build: storage: object doesn't exist
```

This should be logged, but it is not a build failure.

It's unclear how this source archive was deleted.  Currently we name the source archives with a randomly-generated ID, so theoretically there could be a collision, though @dhodun was unable to find evidence 

This PR makes two changes:
- it logs any deletion errors rather than fail the build
- it uses a UUID to name the source archive to avoid collisions

- Google Cloud Builder builds no longer treat errors deleting the source archive as a failure


aaron-prindle (approved):
LGTM

aaron-prindle:
Also not sure if it is interesting but it seems that `cloud_build.go` is the only file in `skaffold/pkg/skaffold/build/gcb/` that does not have an associated `_test.go` file.  Might make sense to add a tracking issue for more testing here if we see more issues.

PR #5894, opened 2021-05-21
Refactor to use new SkaffoldWriter type

MarlonGamez:
**Related**: #5370 

**Description**
Introduces a new type, `SkaffoldWriter`. This replaces `colorableWriter` as the top level `io.Writer` type used by skaffold. It holds one main writer, which will most commonly be a `colorableWriter`, and a secondary writer, which will be used to write output out as events through skaffold API V2. The case for using this over `io.MultiWriter` is that this will allow us to implement functions that perform manipulations on the underlying writers that will be necessary for future implementation of API V2.

Implement event writer type and use in this struct.

### DIFF
--- a/pkg/skaffold/build/buildpacks/lifecycle.go
+++ b/pkg/skaffold/build/buildpacks/lifecycle.go
@@ -91,7 +91,7 @@ func (b *Builder) build(ctx context.Context, out io.Writer, a *latestV1.Artifact
 
 	builderImage, runImage, pullPolicy := resolveDependencyImages(artifact, b.artifacts, a.Dependencies, b.pushImages)
 
-	if err := runPackBuildFunc(ctx, output.GetWriter(out), b.localDocker, pack.BuildOptions{
+	if err := runPackBuildFunc(ctx, output.GetUnderlyingWriter(out), b.localDocker, pack.BuildOptions{
 		AppPath:      workspace,
 		Builder:      builderImage,
 		RunImage:     runImage,
--- a/pkg/skaffold/build/docker/docker.go
+++ b/pkg/skaffold/build/docker/docker.go
@@ -48,7 +48,7 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, a *latestV1.Artifact
 	var imageID string
 
 	if b.useCLI || b.useBuildKit {
-		imageID, err = b.dockerCLIBuild(ctx, output.GetWriter(out), a.Workspace, dockerfile, a.ArtifactType.DockerArtifact, opts)
+		imageID, err = b.dockerCLIBuild(ctx, output.GetUnderlyingWriter(out), a.Workspace, dockerfile, a.ArtifactType.DockerArtifact, opts)
 	} else {
 		imageID, err = b.localDocker.Build(ctx, out, a.Workspace, a.ImageName, a.ArtifactType.DockerArtifact, opts)
 	}
--- a/pkg/skaffold/output/output.go
+++ b/pkg/skaffold/output/output.go
@@ -18,6 +18,7 @@ package output
 
 import (
 	"io"
+	"io/ioutil"
 	"os"
 )
 
@@ -44,7 +45,7 @@ func SetupOutput(out io.Writer, defaultColor int, forceColors bool) io.Writer {
 	return SkaffoldWriter{
 		MainWriter: SetupColors(out, defaultColor, forceColors),
 		// TODO(marlongamez): Replace this once event writer is implemented
-		EventWriter: io.Discard,
+		EventWriter: ioutil.Discard,
 	}
 }
 
--- a/pkg/skaffold/output/output.go
+++ b/pkg/skaffold/output/output.go
@@ -42,6 +42,10 @@ func (s skaffoldWriter) Write(p []byte) (int, error) {
 }
 
 func GetWriter(out io.Writer, defaultColor int, forceColors bool) io.Writer {
+	if _, isSW := out.(skaffoldWriter); isSW {
+		return out
+	}
+
 	return skaffoldWriter{
 		MainWriter: SetupColors(out, defaultColor, forceColors),
 		// TODO(marlongamez): Replace this once event writer is implemented
--- a/pkg/skaffold/output/output.go
+++ b/pkg/skaffold/output/output.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2021 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package output
+
+import (
+	"io"
+	"os"
+)
+
+type SkaffoldWriter struct {
+	MainWriter  io.Writer
+	EventWriter io.Writer
+}
+
+func (s SkaffoldWriter) Write(p []byte) (int, error) {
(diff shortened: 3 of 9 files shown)


sampath:
great PR! Can't wait for API v2!

pkg/skaffold/output/output.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (s SkaffoldWriter) Write(p []byte) (int, error) {
+	for _, w := range []io.Writer{s.MainWriter, s.EventWriter} {
+		n, err := w.Write(p)
+		if err != nil {
+			return n, err
+		}
+		if n < len(p) {
+			return n, io.ErrShortWrite
+		}
```
briandealwis, line 38: So `n < |p|` isn't an error: it happens pretty often when writing to a network stream.

pkg/skaffold/output/output.go
```diff
@@ -0,0 +1,75 @@
...
+type SkaffoldWriter struct {
+	MainWriter  io.Writer
+	EventWriter io.Writer
+}
+
+func (s SkaffoldWriter) Write(p []byte) (int, error) {
+	for _, w := range []io.Writer{s.MainWriter, s.EventWriter} {
+		n, err := w.Write(p)
+		if err != nil {
+			return n, err
```
briandealwis, line 34: Should we ignore errors on the non-MainWriter?

pkg/skaffold/output/output.go
```diff
@@ -0,0 +1,75 @@
...
+func SetupOutput(out io.Writer, defaultColor int, forceColors bool) io.Writer {
+	return SkaffoldWriter{
+		MainWriter: SetupColors(out, defaultColor, forceColors),
+		// TODO(marlongamez): Replace this once event writer is implemented
+		EventWriter: ioutil.Discard,
+	}
+}
+
+func IsStdout(out io.Writer) bool {
+	sw, isSW := out.(SkaffoldWriter)
```
briandealwis, line 53: This doesn't do the right thing if a `colorableWriter` is passed in.  You could structure it as:
```
if sw, isSW := out.(SkaffoldWriter); isSW {
  out = sw.MainWriter
}
if cw, isCW := out.(colorableWriter); isSW {
  out = cw.Writer
}
return out == os.Stdout
```

pkg/skaffold/output/output.go
```diff
@@ -0,0 +1,75 @@
...
+		if isCW {
+			return cw.Writer == os.Stdout
+		}
+		return sw.MainWriter == os.Stdout
+	}
+	return out == os.Stdout
+}
+
+// GetWriter returns the underlying writer if out is a colorableWriter
+func GetWriter(out io.Writer) io.Writer {
```
briandealwis, line 65: And structure this similarly to the above to handle the case where `out` is a `colorableWriter`.

pkg/skaffold/output/output_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+func TestGetWriter(t *testing.T) {
+	tests := []struct {
+		description string
+		out         io.Writer
+		expected    io.Writer
+	}{
+		{
```
briandealwis, line 68: Add some cases for `out: colorableWriter{}`.  And maybe for `io.Discard` to verify that non-stdout is handled too?

pkg/skaffold/output/output_test.go
```diff
@@ -0,0 +1,86 @@
...
+			description: "out nil",
+			out:         nil,
+		},
+		{
+			description: "out bytes buffer",
+			out:         new(bytes.Buffer),
+		},
+		{
+			description: "colorable std out passed",
```
briandealwis, line 48: Plus where a `out: NewColorWriter(os.Stdout)`, and `SkaffoldWriter` and `colorableWriter` with a non-`os.Stdout` writer (`io.Discard`?)

pkg/skaffold/output/output.go
MarlonGamez: Do you think I should be handling this case at all then? `ErrShortWrite means that a write accepted fewer bytes than requested but failed to return an explicit error.` according to docs, so I thought I should handle it, but maybe not

pkg/skaffold/output/output.go
MarlonGamez: I could see the case for that. I'll change this to focus the checks on the mainwriter

pkg/skaffold/output/output.go
```diff
@@ -0,0 +1,75 @@
...
+
+package output
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+)
+
+type SkaffoldWriter struct {
```
gsquared94, line 25: can the struct and its fields be unexported? I don't think these need to be directly accessed other than via methods.

pkg/skaffold/output/output.go
```diff
@@ -0,0 +1,75 @@
...
+	if n != len(p) {
+		return n, io.ErrShortWrite
+	}
+
+	s.EventWriter.Write(p)
+
+	return len(p), nil
+}
+
+func SetupOutput(out io.Writer, defaultColor int, forceColors bool) io.Writer {
```
gsquared94, line 44: nit: How about `output.Setup` instead of `output.SetupOutput` to reduce stutter?

super nit: `Setup` seems to suggest the method has side-effects, but if it's only returning the type could it be called `GetWriter` or something?

pkg/skaffold/build/buildpacks/lifecycle.go
```diff
@@ -91,7 +91,7 @@ func (b *Builder) build(ctx context.Context, out io.Writer, a *latestV1.Artifact
 
 	builderImage, runImage, pullPolicy := resolveDependencyImages(artifact, b.artifacts, a.Dependencies, b.pushImages)
 
-	if err := runPackBuildFunc(ctx, output.GetWriter(out), b.localDocker, pack.BuildOptions{
+	if err := runPackBuildFunc(ctx, output.GetUnderlyingWriter(out), b.localDocker, pack.BuildOptions{
```
briandealwis, line 94: So doesn't this mean the output from this command won't go through our event logging?

Maybe we do this if there is no event writer?

pkg/skaffold/output/output.go
```diff
@@ -0,0 +1,75 @@
...
+	if n != len(p) {
+		return n, io.ErrShortWrite
+	}
+
+	s.EventWriter.Write(p)
+
+	return len(p), nil
+}
+
+func GetWriter(out io.Writer, defaultColor int, forceColors bool) io.Writer {
```
briandealwis, line 44: Should we try to detect if `out` is already a `SkaffoldWriter`?

pkg/skaffold/build/docker/docker.go
```diff
@@ -48,7 +48,7 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, a *latestV1.Artifact
 	var imageID string
 
 	if b.useCLI || b.useBuildKit {
-		imageID, err = b.dockerCLIBuild(ctx, output.GetWriter(out), a.Workspace, dockerfile, a.ArtifactType.DockerArtifact, opts)
+		imageID, err = b.dockerCLIBuild(ctx, output.GetUnderlyingWriter(out), a.Workspace, dockerfile, a.ArtifactType.DockerArtifact, opts)
```
briandealwis, line 51: Same as the pack run case

pkg/skaffold/build/buildpacks/lifecycle.go
MarlonGamez: I was actually curious as to why we did this, and was going to address this in a follow up, as I believe we should be able to use our writer here. I'll probably need to make some more changes to the output setup that's happening in these functions

pkg/skaffold/output/output.go
MarlonGamez: good idea, i'll update with this

tejal29 (approved):

pkg/skaffold/build/buildpacks/lifecycle.go
briandealwis: The other thing to look at is the https://github.com/creack/pty package

PR #5903, opened 2021-05-25
Allow caching from previously built artifacts on GCB

### LINKED ISSUE #5879: Google Cloud Build - cacheFrom to support dynamic images and tags
### Expected behavior
Skaffold should allow dynamic image naming and tagging for the 'cacheFrom' field in Docker (particularly for Google Cloud Build use cases where there is no Docker cache to hit unless you use this field).

Dynamic tagging is particularly needed here, though the image may need to be dynamic as well. A couple options:
1) Tag it the same as the Cloud Build push image/tag combo (works pretty well with gitCommit tagger, less so with dateTime)
2) Tag it with the "last" full SHA and tag that was built using skaffold (I think preferred)

### Actual behavior
Skaffold doesn't treat the 'cacheFrom' field in Cloud Build as an image name or dynamic in any way.

In general, this makes the cacheFrom field only useful for "latest" tag, which needs to be added to your build artifacts manually. Or you are stuck using SHA256 tagger which is generally not recommended.

We have many devs working on the same cluster and GCR repo using their own namespaces, so we need cacheFrom to be as close to the last container that dev built as possible. The only time using their last build isn't ideal is when they change branches, which will be small percentage of the time.

### Information

- Skaffold version: 1.24.1
- Contents of skaffold.yaml:

```yaml
kind: Config
build:
    tagPolicy:
        gitCommit: {}
  artifacts:
  - image: eu.gcr.io/my-project-id/my-image
    context: .
    docker:
      dockerfile: Dockerfile
      cacheFrom:
      - eu.gcr.io/my-project-id/my-image
  googleCloudBuild:
    projectId: my-project-id
```

Similar to: https://github.com/GoogleContainerTools/skaffold/issues/2536

briandealwis:
Fixes #5879

**Description**
Cause GCB Docker `cacheFrom` references to the artifact image (which is untagged) to be replaced with the tagged image (possibly affected by the default-repo).

For example:
```
build:
  googleCloudBuild:
    projectId: k8s-skaffold
  artifacts:
  - image: skaffold-example
    docker:
      cacheFrom:
      - skaffold-example
      - gcr.io/k8s-skaffold/skaffold-example:latest # used if the tagged image is not found
```

Allow disabling by setting `SKAFFOLD_DISABLE_GCB_CACHE_ADJUSTMENT` in the environment.

Note: Docker image caching is a bit complex, and this [StackOverflow answer](https://stackoverflow.com/q/54574821/600339) provides helpful details.

- Docker builds on GoogleCloudBuild can cache from the previous build by including the artifact `image` in the `cacheFrom`.

cc: @dhodun

### DIFF
--- a/docs/content/en/docs/pipeline-stages/builders/docker.md
+++ b/docs/content/en/docs/pipeline-stages/builders/docker.md
@@ -121,6 +121,10 @@ The following options can optionally be configured:
 
 {{< schema root="GoogleCloudBuild" >}}
 
+The `googleCloudBuild` builder replaces cache references to the
+artifact image with the tagged image to allow caching from the
+previously built image.
+
 **Example**
 
 The following `build` section, instructs Skaffold to build a
--- a/docs/content/en/samples/builders/gcb.yaml
+++ b/docs/content/en/samples/builders/gcb.yaml
@@ -1,5 +1,10 @@
 build:
   artifacts:
   - image: gcr.io/k8s-skaffold/example
+    docker:
+      cacheFrom:
+      # googleCloudBuild replaces cache references to the artifact image with
+      # the tagged image reference, useful for caching from the previous build.
+      - gcr.io/k8s-skaffold/example
   googleCloudBuild:
     projectId: YOUR-GCP-PROJECT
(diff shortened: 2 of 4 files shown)


MarlonGamez (approved):
lgtm 👍🏼

PR #5913, opened 2021-05-27
Update documentation to remove duplicates

shyamalmadura:
Duplicated installation URLs were removed to avoid possible confusions for users.

Fixes: N/A 
**Related**: Duplicate entries in the 

**Description**

### DIFF
--- a/docs/content/en/docs/install/_index.md
+++ b/docs/content/en/docs/install/_index.md
@@ -22,15 +22,8 @@ Your use of this software is subject to the [Google Privacy Policy](https://poli
 The latest **stable** binaries can be found here:
 
 https://storage.googleapis.com/skaffold/releases/latest/skaffold-linux-amd64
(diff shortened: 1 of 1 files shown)


aaron-prindle:
Hi @shyamalmadura, thanks for your contribution here.  To clarify, the lines that were removed in this PR were not duplicates, they are 2 different sets of instructions for users based on their specific architecture - AMD64 OR ARM64:
image.  

What might be helpful would be to add some additional clarification or perhaps some kind of architecture selection button/widget that then populates the commands.  As such I don't think this PR as is makes sense to keep as we do hope to have both of these options available but I believe you have a point that it might be confusing for some users.

aaron-prindle:
Closing this for now due to the reasons above, feel free to re-open or submit another PR if anything was missed here. Thanks

shyamalmadura:
> @googlebot I signed it!

PR #5915, opened 2021-05-27
Add step field for `BuildSubtaskEvent` to represent the different parts of a build for an artifact

MarlonGamez:
**Related**: #5423 

**Description**
Adds a new field to `BuildSubtaskEvent` to represent which part of the build phase an artifact is in (ex: Build, Cache, Push). Also emits new events to represents cache checks for an artifact.

**User facing changes**
Users will now see events to represent the cache check phase of skaffold.


proto/v2/skaffold.proto
```diff
@@ -192,8 +192,9 @@ message BuildSubtaskEvent {
     string id = 1; // id of the subtask which will be used in SkaffoldLog
     string task_id = 2; // id of the task of skaffold that this event came from
     string artifact = 3; // artifact name
-    string status = 4; // artifact build status oneof: InProgress, Completed, Failed
-    ActionableErr actionableErr = 5; // actionable error message
+    string step = 4; // which step of the build for the artifact oneof: Cache, Build, Push
```
tejal29, line 195: Just a note: Since v2 is alpha, it is ok to change the  position of the field. Usually, proto should always be backward compatible and should not change the field number.

tejal29 (approved):

Issue #5925, opened 2021-06-02
`docker build --secret` does not support `dst`


briandealwis:
Our Docker builder's [`secret` object](https://skaffold.dev/docs/references/yaml/#build-artifacts-docker-secret) takes a `dst` field (#4731), but it is not supported by `docker build --secret`: that field is actually specified from within the `Dockerfile`.

### Expected behavior

Using a filled in `secret` should succeed:
```
secret:
    id: npmrc
    src: "/Users/bdealwis/.npmrc"
    dst: /app/npmrc
```

### Actual behavior
```
$ skaffold build
Starting build...
Building [skaffold-example]...
could not parse secrets: [id=npmrc,src=/Users/bdealwis/.npmrc,dst=/app/npmrc]: unexpected key 'dst' in 'dst=/app/npmrc'
running build: exit status 1

$ docker build --secret id=npmrc,src=$HOME/.npmrc,dst=/app/npmrc .
could not parse secrets: [id=npmrc,src=/Users/bdealwis/.npmrc,dst=/app/npmrc]: unexpected key 'dst' in 'dst=/app/npmrc'
```

### Information

- Skaffold version: 1.23.0

PR #5926, opened 2021-06-02
New schema version: v2beta17

briandealwis:

PR #5927, opened 2021-06-02
Remove unsupported `docker.secret.dst` field

briandealwis:
Fixes: #5925
Merge after: #5926

**Description**
`docker build --secret` does not actually support a `dst`.  SIlently remove the field since its use causes errors.

The `docker` builder's `secret` object supported an optional `dst` field which was not supported by Buildkit.

### DIFF
--- a/pkg/skaffold/schema/v2beta16/upgrade.go
+++ b/pkg/skaffold/schema/v2beta16/upgrade.go
@@ -17,11 +17,11 @@ limitations under the License.
 package v2beta16
 
 import (
+	"github.com/sirupsen/logrus"
+
 	next "github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest/v1"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
 	pkgutil "github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
-
-	"github.com/sirupsen/logrus"
 )
 
 // Upgrade upgrades a configuration to the next version.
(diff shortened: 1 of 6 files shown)


pkg/skaffold/schema/v2beta16/upgrade.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
+	pkgutil "github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+)
+
+// Upgrade upgrades a configuration to the next version.
+// Config changes from v2beta16 to v2beta17
+func (c *SkaffoldConfig) Upgrade() (util.VersionedConfig, error) {
+	var newConfig next.SkaffoldConfig
+	pkgutil.CloneThroughJSON(c, &newConfig)
+	newConfig.APIVersion = next.Version
```
tejal29, line 30: shd we warn users about `dst` being removed if its set?

tejal29 (approved):
1 nit.

pkg/skaffold/schema/v2beta16/upgrade.go
briandealwis: I don’t think any body can have been using it as Docker fails (see actual behaviour in #5925).

briandealwis: But ok I’ll log a warning

tejal29 (approved):

PR #5935, opened 2021-06-03
add command `skaffold inspect build-env modify googleCloudBuild`

gsquared94:
Related: #5758 
**Description**

add command `skaffold inspect build-env modify googleCloudBuild` to modify the build environment definitions of profiles and default pipelines.

```
❯ skaffold inspect build-env modify googleCloudBuild --help
Modify an existing GoogleCloudBuild build environment definition.
Without the '--profile' flag, the default pipeline's build environment
definition is modified. With the '--profile' flag it'll modify the build
environment definition for all profile pipelines that match the given profile
name. 
In these respective scenarios, it will fail if the build env definition for the
default pipeline or the named profile is not GoogleCloudBuild. To create a new
build environment definition use 'skaffold inspect build-env add' command
instead. 
Use the '--module' filter to specify the individual module to target. Otherwise,
it'll be applied to all modules defined in the target file. Also, with the
'--profile' flag if the target config imports other configs as dependencies,
then it'll modify profiles that match the given name in all imported configs

Examples:
  # Modify an existing profile named 'gcb' targeting the builder
'googleCloudBuild' against the GCP project ID '1234'.
  skaffold inspect build-env modify googleCloudBuild --profile gcb --projectID
1234 -f skaffold.yaml

Options:
      --concurrency=-1: number of artifacts to build concurrently. 0 means
"no-limit"
      --diskSizeGb=0: Disk size of the VM that runs the build
      --machineType='': Type of VM that runs the build
      --projectId='': ID of the Cloud Platform Project.
      --timeout='': Build timeout (in seconds)

Usage:
  skaffold inspect build-env modify googleCloudBuild [flags] [options]

Use "skaffold options" for a list of global command-line options (applies to all
commands).
```


tejal29 (commented):
I feel, just having this command is not sufficient. 
We also need to provide a way for IDEs to detect if googleCloudBuild profile exists.

```
$skaffold inspect build-env list googleCloudBuild
gcb-dev
gcb-prod
gcb-staging

$skaffold inspect build-env modify googleCloudBuild -p gcb.
```

However, we can do this in another PR

pkg/skaffold/inspect/buildEnv/add_gcb.go
```diff
@@ -71,6 +71,57 @@ func AddGcbBuildEnv(ctx context.Context, out io.Writer, opts inspect.Options) er
...
+			profileFound = true
+			if cfg.Profiles[index].Build.GoogleCloudBuild == nil {
+				if opts.Strict {
+					return formatter.WriteErr(inspect.BuildEnvNotFound(inspect.BuildEnvs.GoogleCloudBuild, cfg.SourceFile, opts.Profile))
+				}
+				continue
+			}
+			cfg.Profiles[index].Build.GoogleCloudBuild = constructGcbDefinition(cfg.Profiles[index].Build.GoogleCloudBuild, opts.BuildEnvOptions)
+			cfg.Profiles[index].Build.LocalBuild = nil
+			cfg.Profiles[index].Build.Cluster = nil
```
tejal29, line 116: I was envisioning the given profile to be already a GCB profile and this command will error out if its not a GCB profile.  From the implementation looks like, it will force the profile build env to be GCB and override `local` or `cluster`

gsquared94: sorry could you take another look, I think it's doing what you're expecting.

gsquared94:
> I feel, just having this command is not sufficient.
> We also need to provide a way for IDEs to detect if googleCloudBuild profile exists.
> 
> ```
> $skaffold inspect build-env list googleCloudBuild
> gcb-dev
> gcb-prod
> gcb-staging
> 
> $skaffold inspect build-env modify googleCloudBuild -p gcb.
> ```
> 
> However, we can do this in another PR

We have that in:
```
skaffold profiles list --builder-type googleCloudBuild
```

proto/enums/enums.proto
```diff
@@ -412,6 +412,10 @@ enum StatusCode {
     INSPECT_UNKNOWN_ERR = 1301;
     // Trying to add new build environment that already exists
     INSPECT_BUILD_ENV_ALREADY_EXISTS_ERR = 1302;
+    // Trying to modify build environment that doesn't exist
+    INSPECT_BUILD_ENV_NOT_FOUND_ERR = 1303;
```
tejal29, line 416: A profile can only have 1 build env set right? By default its local?
So, this error code should be `INSPECT_BUILD_ENV_INCORRECT_TYPE` or trying to modify a build env of incorrect type?

tejal29 (approved):

PR #5937, opened 2021-06-03
Add event logger type and function to set event context for writer

MarlonGamez:
**Related**: #5368 

**Description**
Adds event logger type which will turn things passed to it's write function into `SkaffoldLogEvent`s to be sent through the event API v2.

No functional changes here, just adding the functions that will be tied to execution later. Also removes `pkg/skaffold/event/v2`'s reliance on `pkg/skaffold/event`'s Config interface, instead having it's own. This removes an import cycle that was occuring

Hook these up to an endpoint and set proper event contexts throughout skaffold execution

### DIFF
--- a/pkg/skaffold/event/v2/config.go
+++ b/pkg/skaffold/event/v2/config.go
@@ -1,0 +1,29 @@
+/*
+Copyright 2021 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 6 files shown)


tejal29:
Can you add some notes on how to test this out? Shd we see the log message on skaffoldLog endpoint?

MarlonGamez:
@tejal29 this currently isn't hooked up to anything yet, I just wanted to add the functions and types for review to split things up a bit. A follow up to this will actually hook things up and setup certain things like the `skaffoldLogs` endpoint

PR #5942, opened 2021-06-04
[v3]  Add validator in render v2.

yuwenma:
Fixes: #nnn 
**Related**: #5673 
**Merge after**: #5940 

**Description**
- Add the render validator and write the skaffold validation rules to Kptfile pipeline.
- Fixed corner cases when `render.generate` is not given in skaffold.yaml
- More test coverage on Render configs.

To reviewers: 
This PR is diff-based from 5940. Please only review [the last commit](https://github.com/GoogleContainerTools/skaffold/commit/4c4890263ef430ee48538ef370c7cc03f8e61b70)

### DIFF
--- a/pkg/skaffold/render/renderer/renderer.go
+++ b/pkg/skaffold/render/renderer/renderer.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -30,6 +31,7 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/manifest"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/render/generate"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/render/kptfile"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/render/validate"
 	latestV2 "github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest/v2"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 )
@@ -44,17 +46,37 @@ type Renderer interface {
 }
 
 // NewSkaffoldRenderer creates a new Renderer object from the latestV2 API schema.
-func NewSkaffoldRenderer(config *latestV2.RenderConfig, workingDir string) Renderer {
+func NewSkaffoldRenderer(config *latestV2.RenderConfig, workingDir string) (Renderer, error) {
 	// TODO(yuwenma): return instance of kpt-managed mode or skaffold-managed mode defer to the config.Path fields.
 	// The alpha implementation only has skaffold-managed mode.
 	// TODO(yuwenma): The current work directory may not be accurate if users use --filepath flag.
 	hydrationDir := filepath.Join(workingDir, DefaultHydrationDir)
-	generator := generate.NewGenerator(workingDir, *config.Generate)
-	return &SkaffoldRenderer{Generator: *generator, workingDir: workingDir, hydrationDir: hydrationDir}
+
+	var generator *generate.Generator
+	if config.Generate == nil {
+		// If render.generate is not given, default to current working directory.
+		defaultManifests := filepath.Join(workingDir, "*.yaml")
+		generator = generate.NewGenerator(workingDir, latestV2.Generate{Manifests: []string{defaultManifests}})
+	} else {
+		generator = generate.NewGenerator(workingDir, *config.Generate)
+	}
+
+	var validator *validate.Validator
+	if config.Validate != nil {
+		var err error
+		validator, err = validate.NewValidator(*config.Validate)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		validator, _ = validate.NewValidator([]latestV2.Validator{})
+	}
+	return &SkaffoldRenderer{Generator: *generator, Validator: *validator, workingDir: workingDir, hydrationDir: hydrationDir}, nil
 }
 
 type SkaffoldRenderer struct {
 	generate.Generator
+	validate.Validator
 	workingDir   string
 	hydrationDir string
 	labels       map[string]string
@@ -68,14 +90,16 @@ func (r *SkaffoldRenderer) prepareHydrationDir(ctx context.Context) error {
 	if _, err := os.Stat(r.hydrationDir); os.IsNotExist(err) {
 		logrus.Debugf("creating render directory: %v", r.hydrationDir)
 		if err := os.MkdirAll(r.hydrationDir, os.ModePerm); err != nil {
-			return fmt.Errorf("creating cache directory for hydration: %w", err)
+			return fmt.Errorf("creating render directory for hydration: %w", err)
 		}
 	}
 	kptFilePath := filepath.Join(r.hydrationDir, kptfile.KptFileName)
 	if _, err := os.Stat(kptFilePath); os.IsNotExist(err) {
 		cmd := exec.CommandContext(ctx, "kpt", "pkg", "init", r.hydrationDir)
 		if _, err := util.RunCmdOut(cmd); err != nil {
-			return err
+			// TODO: user error. need manual init
+			return fmt.Errorf("unable to initialize kpt directory in %v, please manually run `kpt pkg init %v`",
+				kptFilePath, kptFilePath)
 		}
 	}
 	return nil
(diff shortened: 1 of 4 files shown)


tejal29 (approved):

pkg/skaffold/render/validate/validate.go
```diff
@@ -0,0 +1,51 @@
...
+package validate
+
+import (
+	"fmt"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/render/kptfile"
+	latestV2 "github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest/v2"
+)
+
+var ValidatorWhitelist = map[string]kptfile.Function{
```
tejal29, line 25: we could make this private?
```suggestion
var validatorWhitelist = map[string]kptfile.Function{
```

pkg/skaffold/render/kptfile/pipeline_type.go
```diff
@@ -0,0 +1,294 @@
...
+
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+const (
+	KptFileName       = "Kptfile"
+	KptFileKind       = "Kptfile"
+	KptFileGroup      = "kpt.dev"
+	KptFileVersion    = "v1alpha2"
+	KptFileAPIVersion = KptFileGroup + "/" + KptFileVersion
```
tejal29: consider making this constants private.

pkg/skaffold/render/renderer/renderer.go
```diff
@@ -19,56 +19,83 @@ import (
...
-	generator := generate.NewGenerator(workingDir, *config.Generate)
-	return &SkaffoldRenderer{Generator: *generator, workingDir: workingDir}
+	// TODO(yuwenma): The current work directory may not be accurate if users use --filepath flag.
+	hydrationDir := filepath.Join(workingDir, DefaultHydrationDir)
+
+	var generator *generate.Generator
+	if config.Generate == nil {
+		// If render.generate is not given, default to current working directory.
+		defaultManifests := filepath.Join(workingDir, "*.yaml")
+		generator = generate.NewGenerator(workingDir, latestV2.Generate{Manifests: []string{defaultManifests}})
```
tejal29, line 54: nice!

pkg/skaffold/render/renderer/renderer.go
```diff
@@ -19,56 +19,83 @@ import (
...
+
+	var validator *validate.Validator
+	if config.Validate != nil {
+		var err error
+		validator, err = validate.NewValidator(*config.Validate)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		validator, _ = validate.NewValidator([]latestV2.Validator{})
```
tejal29, line 67: What does this validator do? is it a nil validator?

pkg/skaffold/render/renderer/renderer.go
```diff
@@ -19,56 +19,83 @@ import (
...
-	// TODO(yuwenma): The current work directory may not be accurate if users use --filepath flag.
-	outputPath := filepath.Join(r.workingDir, DefaultHydrationDir)
-	logrus.Debugf("creating render directory: %v", outputPath)
-
-	if err := os.MkdirAll(outputPath, os.ModePerm); err != nil {
-		return fmt.Errorf("creating cache directory for hydration: %w", err)
+	if _, err := os.Stat(r.hydrationDir); os.IsNotExist(err) {
+		logrus.Debugf("creating render directory: %v", r.hydrationDir)
+		if err := os.MkdirAll(r.hydrationDir, os.ModePerm); err != nil {
+			return fmt.Errorf("creating cache directory for hydration: %w", err)
```
tejal29:
```suggestion
			return fmt.Errorf("creating render directory for hydration: %w", err)
```

tejal29: Is this a fatal error? if its just a cache dir, we can still continue without caching?

pkg/skaffold/render/renderer/renderer.go
```diff
@@ -19,56 +19,83 @@ import (
...
+		}
 	}
-	if _, err := os.Stat(filepath.Join(outputPath, KptFileName)); os.IsNotExist(err) {
-		cmd := exec.CommandContext(ctx, "kpt", "pkg", "init", outputPath)
+	kptFilePath := filepath.Join(r.hydrationDir, kptfile.KptFileName)
+	if _, err := os.Stat(kptFilePath); os.IsNotExist(err) {
+		cmd := exec.CommandContext(ctx, "kpt", "pkg", "init", r.hydrationDir)
 		if _, err := util.RunCmdOut(cmd); err != nil {
 			return err
```
tejal29: Is this a kpt user error ? Can we define function and error codes like https://github.com/GoogleContainerTools/skaffold/blob/ca4886fcfbca61a8904190e25328b6ce9abdbd84/pkg/skaffold/deploy/kustomize/errors.go#L24 to define a kpt user error

pkg/skaffold/render/renderer/renderer.go
```diff
@@ -92,11 +119,36 @@ func (r *SkaffoldRenderer) Render(ctx context.Context, out io.Writer, builds []g
...
+	// Read the existing Kptfile content. Kptfile is guaranteed to be exist in prepareHydrationDir.
+	kptFilePath := filepath.Join(r.hydrationDir, kptfile.KptFileName)
+	file, err := os.Open(kptFilePath)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	kfConfig := &kptfile.KptFile{}
+	if err := yaml.NewDecoder(file).Decode(&kfConfig); err != nil {
+		return fmt.Errorf("unable to parse %v\n, please check if the kptfile is updated to new apiVersion > v1alpha2",
```
tejal29, line 128: thanks for creating an actionable error. Can you use sErrors.New() and add error code as well as suggestion code to this error. 
Docs here
https://github.com/GoogleContainerTools/skaffold/blob/master/DEVELOPMENT.md#adding-actionable-error-messages-to-code

The advantage of this this is, IDEs can consume the suggestion code and decide to show an actionable error message to the user as is or relevant one.

tejal29 (commented):
general LGTM.
Few qeustions

pkg/skaffold/render/renderer/renderer.go
yuwenma: Init the validator whose validate functions will be updated if the filewatcher detects a change in the skaffold.yaml.

pkg/skaffold/render/renderer/renderer.go
yuwenma: since kpt only supports in-place hydration, we do need a tmp dir to hold the DRY configs so as the hydrated config won't pollute the user working dir.

pkg/skaffold/render/validate/validate.go
yuwenma: done.

pkg/skaffold/render/renderer/renderer.go
yuwenma: can't guarantee it is a user error or not. I updated the err message with user actions.

pkg/skaffold/render/renderer/renderer.go
yuwenma: oh, we do not want to add the err. If the Decode fails, it basically means the yaml is invalid, and no more details will be given from the Decode return error. That's why we add specific user actions to help users triage.

yuwenma:
Updated. Please take another look @tejal29

pkg/skaffold/render/renderer/renderer.go
tejal29: we still need an error code with an actionable error message `KPT_INVALID_YAML` or `KPT_USER_ERROR` so we can track these in skaffold and get metrics on it. 
Right, if a skaffold dev fails, the error code logged via metrics pipeline  will be `DEPLOY_UNKNOWN`. We are trying to reduce these overall.

pkg/skaffold/render/renderer/renderer.go
tejal29: Is there a way to identify all possible if this was a `kpt pkg init` error or user config error?
Does kpt return unique exit codes based on different errors?
We are trying to attach an error code for all cases.  You can also define `KPT_PKG_INIT` as an error code and return here.

pkg/skaffold/render/renderer/renderer.go
yuwenma: I see. I added some TODO comments (here and other errors in ./render). Will add the Suggestion and error code in the following CL.

pkg/skaffold/render/renderer/renderer.go
yuwenma: that's a good point. I'm not sure if kpt returns the different exit codes or not. We may need more user data to know how likely a user config error may happen in `kpt pkg init` or `kpt fn`

pkg/skaffold/render/renderer/renderer.go
yuwenma: @tejal29 Here's the proto PR https://github.com/GoogleContainerTools/skaffold/pull/5953

tejal29 (approved):

PR #5675, opened 2021-04-13
Add taskevents Test, StatusCheck, and fix duplicates for Deploy

MarlonGamez:
**Related**: #5368  #5422 

**Description**
Adds emission of TaskEvent protos for Test and StatusCheck phases. Also fixes duplicate emission of Deploy TaskEvents 

**User facing changes (remove if N/A)**
User will now see TaskEvents for phases mentioned

**Follow-up Work (remove if N/A)**
Add emission for PortForwarding

PR #5949, opened 2021-06-04
Redo deduplication of deploy taskevents

MarlonGamez:
**Related**: #5368 

**Description**
Redoes the changes made by https://github.com/GoogleContainerTools/skaffold/pull/5675, that I believe were removed by accident by one of the `pkg/skaffold/runner` refactors

Users should not see duped deploy `TaskEvent` protos

### DIFF
--- a/pkg/skaffold/runner/v1/deploy.go
+++ b/pkg/skaffold/runner/v1/deploy.go
@@ -41,7 +41,6 @@ import (
 
 // DeployAndLog deploys a list of already built artifacts and optionally show the logs.
 func (r *SkaffoldRunner) DeployAndLog(ctx context.Context, out io.Writer, artifacts []graph.Artifact) error {
-	eventV2.TaskInProgress(constants.Deploy, "Deploy to cluster")
 
 	// Update which images are logged.
 	r.AddTagsToPodSelector(artifacts)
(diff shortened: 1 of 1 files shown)


tejal29 (approved):

pkg/skaffold/runner/v1/deploy.go
```diff
@@ -53,7 +52,6 @@ func (r *SkaffoldRunner) DeployAndLog(ctx context.Context, out io.Writer, artifa
 	logger.SetSince(time.Now())
 	// First deploy
 	if err := r.Deploy(ctx, out, artifacts); err != nil {
-		eventV2.TaskFailed(constants.Deploy, err)
```
tejal29, line 56: Wouldn't this cause the events to be not sent?

tejal29 (commented):
question about deleted lines

pkg/skaffold/runner/v1/deploy.go
MarlonGamez: It shouldn't as the `runner.Deploy` function has these same function calls. I originally made a mistake as I thought that `runner.DeployAndLog` and `runner.Deploy` were completely separate, but realized that `runner.Deploy` always gets called, either independently or within `runner.DeployAndLog`. That's why I made the original PR, and this is just redoing that PR

tejal29 (approved):

PR #5965, opened 2021-06-07
use event writer in skaffold execution

MarlonGamez:
**Related**: #5370 

**Description**
Sets up the main `skaffoldWriter` to also use an event writer so that output gets written on the `/skaffoldLogs` API endpoint as well. Also sets up some event writer contexts for build and deploy phases so that the output events can be categorized.

users can now hit the `/skaffoldLogs` endpoint to see output from skaffold being emitted as events.

Set up more calls to `WithEventContext` so that things are further organized.


pkg/skaffold/build/scheduler.go
```diff
@@ -107,6 +108,7 @@ func (s *scheduler) build(ctx context.Context, tags tag.ImageTags, i int) error
 	}
 	defer closeFn()
 
+	w = output.WithEventContext(w, constants.Build, strconv.Itoa(eventV2.GetArtifactID(a)), "skaffold")
```
tejal29, line 111: Nice!

pkg/skaffold/output/output.go
```diff
@@ -52,7 +51,7 @@ func GetWriter(out io.Writer, defaultColor int, forceColors bool) io.Writer {
 	return skaffoldWriter{
 		MainWriter: SetupColors(out, defaultColor, forceColors),
 		// TODO(marlongamez): Replace this once event writer is implemented
-		EventWriter: ioutil.Discard,
+		EventWriter: eventV2.NewLogger(constants.DevLoop, "-1", "skaffold"),
```
tejal29, line 54: Do you want to define a constant for "topLevelIdx" or "IdxNone" and use it?

MarlonGamez: will do 👍🏼

tejal29 (approved):

Issue #5407, opened 2021-02-17
Buildpacks builder should support mounting read/write volumes


briandealwis:
Pack 0.15.0 supports mounting read-writable volumes.  This would be very handy for mounting a user's Maven's `~/.m2` or other shared resources.

```
pack build <image-name> [flags]
...
    --volume stringArray       Mount host volume into the build container, in the form '<host path>:<target path>[:<mode>]'.
                               Repeat for each volume in order,
                                 or supply once by comma-separated list

```

We should add the ability for users to specify a source and mount point, and whether the mount should be read-only or read-write.

briandealwis:
PR coming up…

Issue #5231, opened 2021-01-14
Travis tests fail with toomanyrequests


briandealwis:
Docker rate limiting seen on kind, but k3d too:
```
=== Failed Tests ===
/integration/TestDebug/kubectl
        Building [skaffold-debug-java]...
...
        [ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:2.7.1:dockerBuild (default-cli) on project hello-java: 429 Too Many Requests
        [ERROR] GET https://registry-1.docker.io/v2/library/openjdk/manifests/sha256:637c8f0b39977a8228853d0aecbfccbd13db8f75a225d4be7c88d56b84744b9b
        [ERROR] {
        [ERROR]   "errors": [
        [ERROR]     {
        [ERROR]       "code": "TOOMANYREQUESTS",
        [ERROR]       "message": "You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"
        [ERROR]     }
        [ERROR]   ]
        [ERROR] }
```

`kind` is straightforward to configure to use Travis' undocumented registry mirror.  `k3d` fails in odd ways.

briandealwis:
```
=== RUN   TestFilterWithDebugging
time="2021-01-14T15:45:27Z" level=info msg="[skaffold render --default-repo gcr.io/k8s-skaffold]"
    debug_test.go:117: skaffold render: exit status 1, 
    panic.go:617: executing build: unable to stream build output: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit. Please fix the Dockerfile and try again..
```


Hmm but also seeing this error in in our [`checks`](https://api.travis-ci.com/v3/job/471791803/log.txt):
```
RUN hack/test-generated-proto.sh
Sending build context to Docker daemon  194.6kB


Step 1/21 : FROM golang:1.13.4-alpine3.10 AS generate-files
1.13.4-alpine3.10: Pulling from library/golang
[...]
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
Unable to find image 'gen-proto:latest' locally
docker: Error response from daemon: pull access denied for gen-proto, repository does not exist or may require 'docker login'.
See 'docker run --help'.

Generated proto files aren't updated. Please run ./hack/generate-proto.sh
```

[...]

PR #5972, opened 2021-06-08
Buildpacks builder supports mounting read/write volumes (experimental)

briandealwis:
Fixes: #5407 

**Description**
This PR adds support for mounting volumes in buildpacks artifacts, as per `pack build --volume`.  This functionality allows mounting both volumes and local files/directories during the build process. 

#### Possible Uses

This feature, with with a few key environment variables, can allow developers to perform more aggressive caching *between* buildpacks builds using language-specific settings:

- The Golang _module cache_ can be set with [`GOMODCACHE`](https://golang.org/ref/mod#module-cache)
- The Pip wheel cache can be set with `PIP_CACHE_DIR` (pip uses a skaffold-like [mapping of command-like flags to environment variables](https://pip.pypa.io/en/stable/user_guide/#environment-variables)); there's a similar [`PIPENV_CACHE_DIR`](https://github.com/pypa/pipenv/blob/66713a7bcf25fa4d31344df0e1e484a99e562d70/pipenv/environments.py#L87) for `pipenv`
- The npm download location can be set with the [`npm_config_cache`](https://docs.npmjs.com/cli/v7/using-npm/config#cache) [environment variable](https://docs.npmjs.com/cli/v7/using-npm/config#environment-variables)
- The yarn cache directory can be set with the [`YARN_CACHE_FOLDER`](https://classic.yarnpkg.com/en/docs/cli/cache/) environment variable.
- The Maven local repository can be set via the [`MAVEN_OPTS=-Dmaven.repo.local=xxx`](https://maven.apache.org/guides/mini/guide-configuring-maven.html#configuring-your-local-repository) (the property is not actually documented officially)
- Gradle doesn't have a single setting for controlling where downloads are cached, but the Gradle User Home (`~/.gradle`) can be set via the [`GRADLE_USER_HOME`](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_environment_variables) environment variable

Using these mounts is somewhat complicated as Docker normally configures volume- and bind-mounts as owned by root (https://github.com/moby/moby/issues/2259):
  -  _bind mounts_, for local files and directories, generally work for root containers, but can have mixed results with permission problems with non-root containers such as CNB builder images.  And the behaviour seems to vary on the platform (linux vs macOS/Windows).
  - _volume mounts_ retain the ownership defined in the volume.

So the volume needs some preparation for use in a non-root container like the following (assuming the user:group is 1000:1000):
```
docker volume create skaffold-buildpacks-cache
docker run --rm -v skaffold-buildpacks-cache:/skbpcache busybox chown 1000:1000 /skbpcache
```

CNB builder images, like `gcr.io/buildpacks/builder:v1`, define environment variables `CNB_USER` and `CNB_GROUP` to encode the UID and GID used for the build.

So for example, we can enable caching for a Java image with the GCP Buildpacks builder like:
```diff
--- examples/buildpacks-java/skaffold.yaml
+++ examples/buildpacks-java/skaffold.yaml
   - image: skaffold-buildpacks
     buildpacks:
       builder: "gcr.io/buildpacks/builder:v1"
+      trustBuilder: true
+      volumes:
+      - host: skaffold-buildpacks-cache
+        target: /skbpcache
+        options: "rw"
       env:
       - GOOGLE_RUNTIME_VERSION=8
+      - GOOGLE_BUILD_ARGS=-Dmaven.repo.local=/skbpcache
```

This seems like a good candidate for automating with Skaffold.

- Buildpacks builder supports mounting read/write volumes (experimental)

### DIFF
--- a/pkg/skaffold/build/buildpacks/lifecycle.go
+++ b/pkg/skaffold/build/buildpacks/lifecycle.go
@@ -72,6 +72,12 @@ func (b *Builder) build(ctx context.Context, out io.Writer, a *latestV1.Artifact
 	if err != nil {
 		return "", fmt.Errorf("unable to evaluate env variables: %w", err)
 	}
+
+	cc, err := containerConfig(artifact)
+	if err != nil {
+		return "", fmt.Errorf("%q: %w", a.ImageName, err)
+	}
+
 	// List buildpacks to be used for the build.
 	// Those specified in the skaffold.yaml replace those in the project.toml.
 	buildpacks := artifact.Buildpacks
@@ -92,14 +98,15 @@ func (b *Builder) build(ctx context.Context, out io.Writer, a *latestV1.Artifact
 	builderImage, runImage, pullPolicy := resolveDependencyImages(artifact, b.artifacts, a.Dependencies, b.pushImages)
 
 	if err := runPackBuildFunc(ctx, output.GetUnderlyingWriter(out), b.localDocker, pack.BuildOptions{
-		AppPath:      workspace,
-		Builder:      builderImage,
-		RunImage:     runImage,
-		Buildpacks:   buildpacks,
-		Env:          env,
-		Image:        latest,
-		PullPolicy:   pullPolicy,
-		TrustBuilder: artifact.TrustBuilder,
+		AppPath:         workspace,
+		Builder:         builderImage,
+		RunImage:        runImage,
+		Buildpacks:      buildpacks,
+		Env:             env,
+		Image:           latest,
+		PullPolicy:      pullPolicy,
+		TrustBuilder:    artifact.TrustBuilder,
+		ContainerConfig: cc,
 		// TODO(dgageot): Support project.toml include/exclude.
 		// FileFilter: func(string) bool { return true },
 	}); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/skaffold/build/buildpacks/lifecycle.go
```diff
@@ -231,3 +238,23 @@ func resolveDependencyImages(artifact *latestV1.BuildpackArtifact, r ArtifactRes
 
 	return builderImage, runImage, pullPolicy
 }
+
+func containerConfig(artifact *latestV1.BuildpackArtifact) (pack.ContainerConfig, error) {
+	var vols []string
+	if artifact.Volumes != nil {
+		for _, v := range *artifact.Volumes {
+			if v.Host == "" || v.Target == "" {
+				// in case these slip by the JSON schema
```
tejal29, line 247: Should we do some sanity check to make sure, the env variable is set to the target location? Could be done as a follow up PR. 

But following checks
1) Warning: You have volume mount set, but no env variable to use the volume?
2) Error: If there is a mismatch in the target specified and the env variable set.
We can define a allowlist of variables you listed in the PR description.

tejal29 (approved):
LGTM

pkg/skaffold/build/buildpacks/lifecycle.go
briandealwis: I think this risks a lot of false positives.  And there may be situations where just the mount itself is enough to trigger behavioural changes.

briandealwis:
Committing: test failures due to #5231.

PR #5978, opened 2021-06-08
Use golang:1.15-alpine in proto gen

briandealwis:
Related to: #5231

**Description**
Update `golang` image and protoc/protobuf used in the proto check to be consistent with other images being used in our tests.  Take down another source of `toomanyrequests` errors 🤞

### DIFF
--- a/hack/proto/Dockerfile
+++ b/hack/proto/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.13.4-alpine3.10 AS generate-files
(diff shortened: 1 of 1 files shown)


aaron-prindle (approved):
LGTM.  Seems Travis is failing atm though with:
```
The command '/bin/sh -c apk add --no-cache protobuf=3.6.1-r1 unzip jq moreutils' returned a non-zero code: 1
Unable to find image 'gen-proto:latest' locally
docker: Error response from daemon: pull access denied for gen-proto, repository does not exist or may require 'docker login'.
See 'docker run --help'.
```

tejal29 (approved):

PR #5993, opened 2021-06-09
Clear `ApplicationLogEvent` and `SkaffoldLogEvent` buffers on new dev iteration

MarlonGamez:
**Related**: #5368 #5370 

**Description**
This change resets the buffers containing `SkaffoldLogEvent`s and `ApplicationLogEvent`s on a new dev iteration, to prevent increased RAM usage over time.

### DIFF
--- a/pkg/skaffold/event/v2/event.go
+++ b/pkg/skaffold/event/v2/event.go
@@ -305,8 +305,12 @@ func AutoTriggerDiff(phase constants.Phase, val bool) (bool, error) {
 }
 
 func TaskInProgress(task constants.Phase, description string) {
+	// Special casing to increment iteration and clear application and skaffold logs
 	if task == constants.DevLoop {
 		handler.iteration++
+
+		handler.applicationLogs = nil
+		handler.skaffoldLogs = nil
 	}
 
 	handler.handleTaskEvent(&proto.TaskEvent{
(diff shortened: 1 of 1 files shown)


tejal29 (commented):
should we allocate new memory instead of nil to make sure there is no nil point exception

MarlonGamez:
@tejal29 I'm unsure, since there are no fields of a slice that we can access so I don't think that could happen. I think that functionally, a `nil` and empty slice would behave the same. Iterating over it just results in zero iterations, like an empty slice. 
This way is just cleaner to me over writing out a slice literal. I'm happy to change if you prefer tho!

tejal29:
> @tejal29 I'm unsure, since there are no fields of a slice that we can access so I don't think that could happen. I think that functionally, a `nil` and empty slice would behave the same. Iterating over it just results in zero iterations, like an empty slice.
> This way is just cleaner to me over writing out a slice literal. I'm happy to change if you prefer tho!

ah. I guess iterating would be fine. I had looked up usage last time and looks like we pass the pointer to these fields. 
```
return ev.forEach(&ev.applicationLogListeners, &ev.applicationLogs, &ev.applicationLogsLock, callback)
```
My concern is this might end up into nil pointer exception?

MarlonGamez:
@tejal29 made the change to use slice literals 👍🏼

tejal29 (approved):

PR #6006, opened 2021-06-10
Emit `PortFowardEvent`s for API V2

MarlonGamez:
**Related**: #5368 

**Description**
This PR brings over the functionality from API V1 that emits events when a resource is forwarded. The event contains information about the forwarded resource.

### DIFF
--- a/pkg/skaffold/kubernetes/portforward/entry_manager.go
+++ b/pkg/skaffold/kubernetes/portforward/entry_manager.go
@@ -30,7 +30,6 @@ import (
 
 var (
 	portForwardEvent = func(entry *portForwardEntry) {
-		// TODO priyawadhwa@, change event API to accept ports of type int
 		event.PortForwarded(
 			int32(entry.localPort),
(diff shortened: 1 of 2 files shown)


pkg/skaffold/kubernetes/portforward/entry_manager.go
```diff
@@ -41,6 +42,19 @@ var (
 			entry.resource.Name,
 			entry.resource.Address)
 	}
+	portForwardEventV2 = func(entry *portForwardEntry) {
+		// TODO priyawadhwa@, change event API to accept ports of type int
```
tejal29, line 46: Is this TODO still relevant?
If yes, we should remove @priyawadhwa from here.

MarlonGamez: whoops, that shouldn't be there. Removed the other as well as the event already accepts `int32`

tejal29 (approved):

Issue #5427, opened 2021-02-20
Skaffold reports pod is ready based on phase not readiness state


briandealwis:
In CC-VSC, I saw in the output:
```
Status check started
Resource pod/mongodb-deployment-587b66548b-nmgsh status updated to In Progress
Resource pod/nodetodo-deployment-5dcbd4795b-94sqb status updated to In Progress
Resource pod/mongodb-deployment-587b66548b-nmgsh status completed successfully
Resource deployment/mongodb-deployment status completed successfully
Resource deployment/nodetodo-deployment status updated to In Progress
Resource pod/nodetodo-deployment-5dcbd4795b-94sqb status completed successfully
```

So pod/nodetodo-deployment-5dcbd4795b-94sqb appeared to be running.  But `kubectl` disagrees:
```
$ kubectl get all
NAME                                       READY   STATUS    RESTARTS   AGE
pod/mongodb-deployment-587b66548b-nmgsh    1/1     Running   0          92s
pod/nodetodo-deployment-5dcbd4795b-94sqb   0/1     Running   0          92s

NAME                        TYPE           CLUSTER-IP       EXTERNAL-IP   PORT(S)        AGE
service/kubernetes          ClusterIP      10.96.0.1        <none>        443/TCP        16d
service/mongodb-service     ClusterIP      10.98.37.129     <none>        27017/TCP      92s
service/nodetodo-external   LoadBalancer   10.104.156.208   <pending>     80:32354/TCP   92s

NAME                                  READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/mongodb-deployment    1/1     1            1           92s
deployment.apps/nodetodo-deployment   0/1     1            0           92s

NAME                                             DESIRED   CURRENT   READY   AGE
replicaset.apps/mongodb-deployment-587b66548b    1         1         1       92s
replicaset.apps/nodetodo-deployment-5dcbd4795b   1         1         0       92s

$ kubectl describe pod/nodetodo-deployment-5dcbd4795b-94sqb
...
  Warning  Unhealthy  32s (x4 over 4m38s)  kubelet            Readiness probe failed: Get "http://172.17.0.5:3000/todo": EOF
  Warning  Unhealthy  32s (x3 over 4m32s)  kubelet            Liveness probe failed: Get "http://172.17.0.5:3000/todo": EOF
  Normal   Killing    32s                  kubelet            Container nodetodo failed liveness probe, will be restarted
[...]

PR #6010, opened 2021-06-11
fix status-check to return success only on exact success criteria match

gsquared94:
Fixes: #5427  

**Description**

This PR changes the logic that deals with deciding when to declare that a pod is _successfully_ deployed in Skaffold's `status check` phase.

Previously, we returned `success` if none of the pod condition type was `PodScheduled`. This didn't work for cases when the event transitions do not have a `PodScheduled` type event but a `PodReady` type event with `Status="False"`.
https://github.com/GoogleContainerTools/skaffold/blob/815984c068f232a548d2b8ba002207530af8004e/pkg/diag/validator/pod.go#L121-L154

This refactor returns `success` only when we detect the last condition transition (`PodReady` type with status `True`). The behavior for `PodScheduled` type event is extended for `PodReady` and `ContainersReady` type with status `False`.

This fixes the original repro in #5427 against https://github.com/VeerMuchandi/nodejs-todo-cloudshell where we get the correct status check events when run from VSCode Cloud Code IDE:

```
Status check started
Resource pod/mongodb-deployment-7d7c46b57-b9pzj status updated to In Progress
Resource pod/nodetodo-deployment-7f49f845b-429t5 status completed successfully
Resource pod/mongodb-deployment-7d7c46b57-b9pzj status completed successfully
Resource deployment/mongodb-deployment status completed successfully
Resource deployment/nodetodo-deployment status updated to In Progress
Resource pod/nodetodo-deployment-7f49f845b-429t5 status failed with Readiness probe failed: Get http://10.244.0.26:3000/todo: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
Resource deployment/nodetodo-deployment status completed successfully
Status check succeeded
```

### DIFF
--- a/pkg/diag/validator/pod.go
+++ b/pkg/diag/validator/pod.go
@@ -120,15 +120,19 @@ func (p *PodValidator) getPodStatus(pod *v1.Pod) *podStatus {
 
 func getPodStatus(pod *v1.Pod) (proto.StatusCode, []string, error) {
 	// See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions
+
+	// If the event type PodReady with status True is found then we return success immediately
 	if isPodReady(pod) {
 		return proto.StatusCode_STATUSCHECK_SUCCESS, nil, nil
 	}
+	// If the event type PodScheduled with status False is found then we check if it is due to taints and tolerations.
 	if c, ok := isPodNotScheduled(pod); ok {
 		logrus.Debugf("Pod %q not scheduled: checking tolerations", pod.Name)
 		sc, err := getUntoleratedTaints(c.Reason, c.Message)
 		return sc, nil, err
 	}
-
+	// we can check the container status if the pod has been scheduled successfully. This can be determined by having the event
+	// PodScheduled with status True, or a ContainerReady or PodReady event with status False.
 	if isPodScheduledButNotReady(pod) {
 		logrus.Debugf("Pod %q scheduled but not ready: checking container statuses", pod.Name)
 		// TODO(dgageot): Add EphemeralContainerStatuses
@@ -159,7 +163,7 @@ func getPodStatus(pod *v1.Pod) (proto.StatusCode, []string, error) {
 
 func isPodReady(pod *v1.Pod) bool {
 	for _, c := range pod.Status.Conditions {
-		if c.Type == v1.ContainersReady && c.Status == v1.ConditionTrue {
+		if c.Type == v1.PodReady && c.Status == v1.ConditionTrue {
 			return true
 		}
 	}
@@ -178,6 +182,9 @@ func isPodNotScheduled(pod *v1.Pod) (v1.PodCondition, bool) {
 func isPodScheduledButNotReady(pod *v1.Pod) bool {
 	for _, c := range pod.Status.Conditions {
 		if c.Type == v1.PodScheduled && c.Status == v1.ConditionTrue {
+			return true
+		}
+		if c.Type == v1.ContainersReady && c.Status == v1.ConditionFalse {
 			return true
 		}
 		if c.Type == v1.PodReady && c.Status == v1.ConditionFalse {
--- a/pkg/diag/validator/pod.go
+++ b/pkg/diag/validator/pod.go
@@ -120,37 +120,80 @@ func (p *PodValidator) getPodStatus(pod *v1.Pod) *podStatus {
 
 func getPodStatus(pod *v1.Pod) (proto.StatusCode, []string, error) {
 	// See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions
-	for _, c := range pod.Status.Conditions {
-		if c.Type == v1.PodScheduled {
-			switch c.Status {
-			case v1.ConditionFalse:
-				logrus.Debugf("Pod %q not scheduled: checking tolerations", pod.Name)
-				sc, err := getUntoleratedTaints(c.Reason, c.Message)
-				return sc, nil, err
-			case v1.ConditionTrue:
-				logrus.Debugf("Pod %q scheduled: checking container statuses", pod.Name)
-				// TODO(dgageot): Add EphemeralContainerStatuses
-				cs := append(pod.Status.InitContainerStatuses, pod.Status.ContainerStatuses...)
-				// See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states
-				statusCode, logs, err := getContainerStatus(pod, cs)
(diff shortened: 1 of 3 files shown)


tejal29:
Will this also fix  #5966 ?

pkg/diag/validator/pod.go
```diff
@@ -120,37 +120,80 @@ func (p *PodValidator) getPodStatus(pod *v1.Pod) *podStatus {
...
+	return proto.StatusCode_STATUSCHECK_UNKNOWN, nil, fmt.Errorf("unable to determine current service state of pod %q", pod.Name)
+}
+
+func isPodReady(pod *v1.Pod) bool {
+	for _, c := range pod.Status.Conditions {
+		if c.Type == v1.ContainersReady && c.Status == v1.ConditionTrue {
+			return true
+		}
+	}
+	return false
```
briandealwis, line 166: Use the pod's Ready status: the containers could be ready, but the pod itself might not yet be.

pkg/diag/validator/pod.go
```diff
@@ -120,37 +120,80 @@ func (p *PodValidator) getPodStatus(pod *v1.Pod) *podStatus {
...
+		}
+	}
+	return v1.PodCondition{}, false
+}
+
+func isPodScheduledButNotReady(pod *v1.Pod) bool {
+	for _, c := range pod.Status.Conditions {
+		if c.Type == v1.PodScheduled && c.Status == v1.ConditionTrue {
+			return true
+		}
```
briandealwis, line 182: This doesn't seem right: if we find the pod is scheduled, we'll return true even if the pod is not ready.  I think you need to track both the scheduled status and ready status.

It might be easier if you created some helper functions like `func podCondition(pod *v1.Pod, type v1. v1.PodConditionType) v1.Condition`.  Then you could recast some of these as `return podCondition(pod, v1.PodScheduled) == v1.ConditionTrue && podCondition(pod, v1.PodReady) != v1.ConditionTrue`

pkg/diag/validator/pod.go
```diff
@@ -120,37 +120,80 @@ func (p *PodValidator) getPodStatus(pod *v1.Pod) *podStatus {
...
+		cs := append(pod.Status.InitContainerStatuses, pod.Status.ContainerStatuses...)
+		// See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states
+		statusCode, logs, err := getContainerStatus(pod, cs)
+		if statusCode == proto.StatusCode_STATUSCHECK_POD_INITIALIZING {
+			// Determine if an init container is still running and fetch the init logs.
+			for _, c := range pod.Status.InitContainerStatuses {
+				if c.State.Waiting != nil {
+					return statusCode, []string{}, fmt.Errorf("waiting for init container %s to start", c.Name)
+				} else if c.State.Running != nil {
+					return statusCode, getPodLogs(pod, c.Name), fmt.Errorf("waiting for init container %s to complete", c.Name)
```
briandealwis, line 144: I wonder if we're not better treating InitContainers separately from the regular Containers, rather than trying to special case the handling in here.

briandealwis (commented):
Your refactor makes the code much more clear!  I've made a few suggestions for improving it even further.  I think these helper functions would benefit from some specific tests to be certain they're correctly identifying the desired cases.

pkg/diag/validator/pod.go
```diff
@@ -120,37 +120,80 @@ func (p *PodValidator) getPodStatus(pod *v1.Pod) *podStatus {
...
-				if statusCode == proto.StatusCode_STATUSCHECK_POD_INITIALIZING {
-					// Determine if an init container is still running and fetch the init logs.
-					for _, c := range pod.Status.InitContainerStatuses {
-						if c.State.Waiting != nil {
-							return statusCode, []string{}, fmt.Errorf("waiting for init container %s to start", c.Name)
-						} else if c.State.Running != nil {
-							return statusCode, getPodLogs(pod, c.Name), fmt.Errorf("waiting for init container %s to complete", c.Name)
-						}
-					}
+	if isPodReady(pod) {
```
briandealwis, line 123: It would also be useful to explain in prose the different conditions, especially odd cases like `the pod can be ready, but a container could be not ready because it terminated.`

pkg/diag/validator/pod.go
gsquared94: > This doesn't seem right: if we find the pod is scheduled, we'll return true even if the pod is not ready. 

we've already checked that the pod is not ready in the first method `isPodReady`. So here we're just checking if there's any `PodScheduled` true or `PodReady` false event so that we can fetch the container logs for it.

pkg/diag/validator/pod.go
gsquared94: I can take a stab at reworking the InitContainer logic as a follow-up, but don't want to make any additional changes in this PR that aren't related to the original issue.

pkg/diag/validator/pod.go
gsquared94: done

gsquared94:
> Your refactor makes the code much more clear! I've made a few suggestions for improving it even further. I think these helper functions would benefit from some specific tests to be certain they're correctly identifying the desired cases.

Pushed changes based on your comments. PTAL.

pkg/diag/validator/pod.go
```diff
@@ -120,37 +120,87 @@ func (p *PodValidator) getPodStatus(pod *v1.Pod) *podStatus {
...
-					for _, c := range pod.Status.InitContainerStatuses {
-						if c.State.Waiting != nil {
-							return statusCode, []string{}, fmt.Errorf("waiting for init container %s to start", c.Name)
-						} else if c.State.Running != nil {
-							return statusCode, getPodLogs(pod, c.Name), fmt.Errorf("waiting for init container %s to complete", c.Name)
-						}
-					}
+
+	// If the event type PodReady with status True is found then we return success immediately
+	if isPodReady(pod) {
```
tejal29, line 125: Nice! Like the helper methods.

pkg/diag/validator/pod.go
gsquared94: done.

pkg/diag/validator/pod.go
tejal29: I think the reason i handled it separately is because, 
1) Pods are in "Initialized" state only if  init containers have started and waiting to be completed. 
We want to show init container logs to users if pod is not ready due to init containers.

We only show regular container logs in status check phase if they failed or terminated. Otherwise, regular container logs are available after status check.

tejal29 (approved):

gsquared94:
the codecov/patch diff report is bogus
image

PR #6023, opened 2021-06-16
fix: gcb api throttling retry backoff not implemented correctly

gsquared94:
We intended to retry GCB status requests with exponential backoff when the requests start getting throttled. However it seems we never had the backoff working.

cc @dhodun might help with the GCB 429 errors you were experiencing with multiple parallel builds.

### DIFF
--- a/pkg/skaffold/build/gcb/cloud_build.go
+++ b/pkg/skaffold/build/gcb/cloud_build.go
@@ -151,7 +151,7 @@ watch:
 		logrus.Debugf("current offset %d", offset)
 		backoff := NewStatusBackoff()
 		if waitErr := wait.Poll(backoff.Duration, RetryTimeout, func() (bool, error) {
-			backoff.Step()
+			time.Sleep(backoff.Step())
 			cb, err = cbclient.Projects.Builds.Get(projectID, remoteID).Do()
 			if err == nil {
 				return true, nil


pkg/skaffold/build/gcb/cloud_build.go
```diff
@@ -151,7 +151,7 @@ watch:
 		logrus.Debugf("current offset %d", offset)
 		backoff := NewStatusBackoff()
 		if waitErr := wait.Poll(backoff.Duration, RetryTimeout, func() (bool, error) {
-			backoff.Step()
+			time.Sleep(backoff.Step())
```
briandealwis, line 154: Poll doesn't seem the right function here: it does a fixed wait.  And we're not checking the context here either.

[`wait.ExponentialBackoffWithContext()`](https://github.com/kubernetes/apimachinery/blob/v0.21.1/pkg/util/wait/wait.go#L610) looks like the function we should be using?

Honestly, this whole loop should be refactored into a set of methods.  It's hard to trace the logic, and I worry that there's a chance that we cut off the logs (do we know we retrieve the remainder of the logs when `cb.Status == StatusSuccess` or `StatusFailure`?

And shouldn't we delete the source archive on failure or any other termination condition?

briandealwis: And you are refactoring it in #6005 (where is my brain :-D)

gsquared94: ya 😅 I thought to just patch this bug in the interim.

tejal29 (approved):

Issue #5843, opened 2021-05-13
Add MetaData API event


tejal29:
Implement metadata API for June release.

tejal29:
I feel this is done @marlon-gamez. Closing this. Please re-open if anything needs to be addressed.

Closed by #5916 and #5918

PR #6033, opened 2021-06-17
Refactor runID usage, populate runID field in metadata event

MarlonGamez:
**Related**: #5843 

**Description**
This PR refactors the runID variable to be passed through the main RunContext of skaffold, rather than used through a package variable. It also populates the runID field in the metadata event that gets sent through the event API.

### DIFF
--- a/pkg/skaffold/deploy/label/labeller.go
+++ b/pkg/skaffold/deploy/label/labeller.go
@@ -19,16 +19,12 @@ package label
 import (
 	"fmt"
 	"strings"
-
-	"github.com/google/uuid"
 )
 
 const (
 	K8sManagedByLabelKey = "app.kubernetes.io/managed-by"
 	RunIDLabel           = "skaffold.dev/run-id"
 )
-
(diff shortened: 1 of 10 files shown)


pkg/skaffold/deploy/label/labeller.go
```diff
@@ -28,7 +28,7 @@ const (
 	RunIDLabel           = "skaffold.dev/run-id"
 )
 
-var runID = uuid.New().String()
+var RunID = uuid.New().String()
```
briandealwis, line 31: I have to admit to not being a fan of globals like this.  Is there no way to have this on the run context, and have the value be fed into the labeller?

MarlonGamez: seems like this is doable 👍🏼 I'll work on that

MarlonGamez: done

tejal29 (approved):

PR #6035, opened 2021-06-17
Log when unable to retrieve access token

briandealwis:
There have been a couple of users having issues with the GCB builder.  From working with one user, I believe they're encountering an error where we attempt to retrieve the access token with `gcloud auth print-access-token`.  If there is an error, we currently silently fail.  This PR logs the error similar to how we log subsequent errors.

### DIFF
--- a/pkg/skaffold/gcp/auth.go
+++ b/pkg/skaffold/gcp/auth.go
@@ -64,6 +64,8 @@ func activeUserCredentials() (*google.Credentials, error) {
 		cmd := exec.Command("gcloud", "auth", "print-access-token", "--format=json")
 		body, err := util.RunCmdOut(cmd)
 		if err != nil {
+			logrus.Infof("unable to retrieve gcloud access token: %v", err)
+			logrus.Infof("falling back to application default credentials")
 			credsErr = fmt.Errorf("retrieving gcloud access token: %w", err)
 			return
 		}


pkg/skaffold/gcp/auth.go
```diff
@@ -64,6 +64,8 @@ func activeUserCredentials() (*google.Credentials, error) {
 		cmd := exec.Command("gcloud", "auth", "print-access-token", "--format=json")
 		body, err := util.RunCmdOut(cmd)
 		if err != nil {
+			logrus.Infof("unable to retrieve gcloud access token: %v", err)
+			logrus.Infof("falling back to application default credentials")
 			credsErr = fmt.Errorf("retrieving gcloud access token: %w", err)
```
tejal29, line 69: Should this instead be an Actionable Error? `sErrors.Error`?

briandealwis: Probably not, since we fall through to the client library, which should default to the ADC?  The underlying error does show an actionable error from gcloud.  This should really be be a warning, not info.

briandealwis:
```
Starting build...
INFO[0002] unable to retrieve gcloud access token: running [gcloud auth print-access-token --format=json]
 - stdout: ""
 - stderr: "ERROR: (gcloud.auth.print-access-token) There was a problem refreshing your current auth tokens: ('invalid_grant: Invalid grant: account not found', {'error': 'invalid_grant', 'error_description': 'Invalid grant: account not found'})\nPlease run:\n\n  $ gcloud auth login\n\nto obtain new credentials.\n\nIf you have already logged in with a different account:\n\n    $ gcloud config set account ACCOUNT\n\nto select an already authenticated account to use.\n"
 - cause: exit status 1 
INFO[0002] falling back to application default credentials 
```

It's not pretty though.

tejal29 (approved):

pkg/skaffold/gcp/auth.go
briandealwis: `gcloud` isn't outputting the error as JSON; filed b/191668532 to track.

PR #6038, opened 2021-06-17
add gh actions early success if change is docs only

### LINKED ISSUE #6032: Add Code to not run Github Actions on docs-only changes, possibly something more general to stop actions from running on a PR altogether
Currently the Github Actions in the skaffold repo run on docs only changes.  Travis is setup to not do this but this functionality was not ported over.  This bug tracks adding either:
- Similar logic taken from Travis to halt job execution on Github Actions
- Add label based "stop execution" of the Github Action (more general, possibly more cumbersome as job might run once before label is applied)

aaron-prindle:
### What is the problem being solved?
Fixes #6032, add code to not run Github Actions on docs-only changes. This PR:
    - Updates linux and osx github actions to check if the PR only changes docs/* files and if so, pass the test w/o running the full test suite

Github actions does not have support for exiting the job from a step easily. See:
    - https://stackoverflow.com/questions/60589373/how-to-force-to-exit-in-github-actions-step
    - https://github.com/actions/runner/issues/662

### Why is this the best approach?
The approach here uses an env var and checks this env var in all subsequent steps, this way no single command fails and tests are skipped as desired.  The downside is that all subsequent steps need to have this check but there is no better solution at the moment from my investigation

### What other approaches did you consider?
I considered directly using the the 'conclusion' value github actions provides, eg:
  - "if: steps.s1.conclusion == 'failure'"
  but I believe that if a step fails then the test fails which is not what we want.

### What side effects will this approach have?
There shouldn't be any side effects w/ this approach, all steps have a conditional on the docs-only check and should work correctly as before in the non-docs file are updated case

### What future work remains to be done?
N/A.  Future work generally might include a tag that also skips testing but is not directly related

### DIFF
--- a/.github/workflows/integration-darwin.yml
+++ b/.github/workflows/integration-darwin.yml
@@ -18,12 +18,25 @@ jobs:
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    # Skip integration tests for `docs`-only changes (only works for PR-based dev workflows like Skaffold's).
+    # NOTE: grep output is stored in env var with `|| true` as the run command cannot fail or action will fail
+    - name: Check if only docs changes were made in this PR
+      run: |
+        echo ${{ github.event.before }}
+        echo ${{ github.event.after }}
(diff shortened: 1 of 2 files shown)


aaron-prindle:
See #6039 for validation that this works. You can see in the Github Actions on the PR there that the linux tests indeed succeed immediatly when a docs only change is made

tejal29 (approved):

Issue #6041, opened 2021-06-18
Add ko builder


halvards:
Tracking issue for the effort to add the `ko` builder.

See the [design proposal](../blob/master/docs/design_proposals/ko-builder.md) for further details.

Related: #5611, #1958

halvards:
Related PRs in [google/ko](https://github.com/google/ko):

- [Enable embedding of ko publish #348](https://github.com/google/ko/pull/348)
- [Add flag and PublishOption for destination repo #351](https://github.com/google/ko/pull/351)
- [Enable setting the working dir for the go tool #365](https://github.com/google/ko/pull/365)
- [Add support for Go build flags #340](https://github.com/google/ko/pull/340)

tejal29:
Thanks for creating the issue!

aaron-prindle:
ko builder is added now 🎊! Marking closed(fixed)

PR #6042, opened 2021-06-18
Add dependency on ko

halvards:
**Description**

Small steps, this is just to verify the impact of changes to the
dependency graph.

**Related**: #6041


halvards:
I've added a small section in the design proposal to explain the implementation approach of the ko builder. The approach is to implement the ko builder as a series of small PRs that can be merged one-by-one. No user-visible changes will be visible until the feature is ready.

See: #6046

tejal29 (approved):

tejal29:
Looking forward to the ko builder functionality!

PR #6046, opened 2021-06-21
Updates to ko builder design proposal to add implementation approach

halvards:
**Tracking**: #6041 

**Description**
Update resolved questions and add section for implementation approach.

Specifically, document that the implementation will be a series of small PRs that can be merged one-by-one, rather than all at once when the feature is done. The PRs will not surface any new user-visible behavior until the feature is ready to be released.

### DIFF
--- a/docs/design_proposals/ko-builder.md
+++ b/docs/design_proposals/ko-builder.md
@@ -423,7 +423,7 @@ The steps roughly outlined:
     func PublishImages(ctx context.Context, importpaths []string, pub publish.Interface, b build.Interface) (map[string]name.Reference, error)
     ```
 
-    Add build and publish options to support Skaffold config propagaring to
+    Add build and publish options to support Skaffold config propagating to
     ko. In the package `github.com/google/ko/pkg/commands/options`:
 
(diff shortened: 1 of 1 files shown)


docs/design_proposals/ko-builder.md
```diff
@@ -285,87 +290,176 @@ maps directly to this value.
...
     __Not Yet Resolved__
 
-8.  Should we default dependency paths to `{"go.mod", "**.go"}` instead of
-    `{"."}`.?
-
-    The former is a useful default for many (most?) Go apps, and it's used
-    in the `custom` example. The latter is the default for some other builders.
+4.  Should the ko builder be the default for `skaffold init`, instead of
+    buildpacks, for Go apps, when there's no Dockerfile and no Bazel workspace
+    file?
```
MarlonGamez, line 347: I think this would be a good options. The way `init` treats buildpacks currently is pretty much like a catch all if we can't find a more specific builder to use for the project. So ko would be a good option to use first in this case

MarlonGamez (changes requested):
one tiny typo and response to a question

MarlonGamez (approved):

PR #6064, opened 2021-06-22
Ensure events are serialized

briandealwis:
@matthewmichihara reported seeing out-of-order events.  Our event code fires off a separate goroutine for each event which can lead to out-of-sequence delivery.  Change the event delivery to write directly to the event channel to ensure events are delivered in order.


MarlonGamez (approved):
lgtm, thanks!

PR #6065, opened 2021-06-22
Fill id field for `StatusCheckSubtaskEvent`s

MarlonGamez:
**Related**: #5368 

**Description**
This PR fills the id field of `StatusCheckSubtaskEvent`s, with each id corresponding to a resource name. Also fixes the description for the status check task event to match the tense of other descriptions.


pkg/skaffold/event/v2/status_check.go
```diff
@@ -18,11 +18,14 @@ package v2
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
 	proto "github.com/GoogleContainerTools/skaffold/proto/v2"
 )
 
+var resourceIDs = map[string]int{}
```
gsquared94, line 27: is there a way to avoid this static map? I added a `ID()` method to `resource.Deployment` which I think can be used here.

pkg/skaffold/event/v2/status_check.go
```diff
@@ -62,7 +65,17 @@ func ResourceStatusCheckEventUpdated(r string, ae proto.ActionableErr) {
 	})
 }
 
+func getResourceID(resource string) int {
+	if _, contains := resourceIDs[resource]; !contains {
```
tejal29, line 69:
```suggestion
	if _, ok := resourceIDs[resource]; !ok {
```

tejal29 (commented):
quick nit

pkg/skaffold/event/v2/status_check.go
MarlonGamez: I think I can just use the resource name as the id, since it shouldn't make a difference on the ide side and that's what the integer ids would map to anyway

pkg/skaffold/event/v2/status_check.go
```diff
@@ -63,6 +63,8 @@ func ResourceStatusCheckEventUpdated(r string, ae proto.ActionableErr) {
 }
 
 func (ev *eventHandler) handleStatusCheckSubtaskEvent(e *proto.StatusCheckSubtaskEvent) {
+	e.Id = e.Resource
```
gsquared94, line 66: nit: do this where `e` is being instantiated

gsquared94 (approved):

PR #6066, opened 2021-06-22
working cloud profiler export for skaffold

aaron-prindle:
### Description
Adds `cloud profiler` code to skaffold that is able to upload profile information to GCP's `Cloud Profiler` using user's own gcloud application default credentials.  

### Alternatives
This could be done as a build tag and not compiled into skaffold binary but the upside is not worth it becuause:
- The binary size difference is negligile
  - skaffold `master` branch @ HEAD
    - ```
      aprindle@aprindle ~/skaffold  [master]$ ls -lh out/skaffold
      -rwxr-x--- 1 aprindle primarygroup 61M Jun 22 14:52 out/skaffold
      ```
  - skaffold `fix-5789_cloud-profiler` @ HEAD (this PR)
    - ```
      aprindle@aprindle ~/skaffold  [master]$ ls -lh out/skaffold
      -rwxr-x--- 1 aprindle primarygroup 62M Jun 21 21:46
      ```
- `make linters` checks would require adding tweaks/hacks as they do not support parsing two of the same function currently as build tags would require

### Considerations
  Similar to `SKAFFOLD_TRACE`, the `SKAFFOLD_PROFILE` functionality will be used in cron tests around skaffold's performance, do not affect skaffold's performance unless enabled (as would be expected) and add negligible size to the skaffold binary (~61M->~62M)


tejal29:
> Similar to `SKAFFOLD_TRACE`, the `SKAFFOLD_PROFILE` functionality will be used in cron tests around skaffold's performance, do not affect skaffold's performance unless enabled (as would be expected) and add negligible size to the skaffold binary

`SKAFFOLD_PROFILE` is already used for `--profile` flag. Can we change this name to `SKAFFOLD_PROFILER` ?

cmd/skaffold/skaffold.go
```diff
@@ -19,20 +19,38 @@ package main
...
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/output"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/version"
 )
 
 type ExitCoder interface {
 	ExitCode() int
 }
 
 func main() {
+	if _, ok := os.LookupEnv("SKAFFOLD_PROFILE"); ok {
```
tejal29, line 39:
```suggestion
	if _, ok := os.LookupEnv("SKAFFOLD_PROFILER"); ok {
```

See here https://skaffold.dev/docs/references/cli/#skaffold-apply ( search for `SKAFFOLD_PROFILE`)

cmd/skaffold/skaffold.go
```diff
@@ -19,20 +19,38 @@ package main
...
+		err := profiler.Start(profiler.Config{
+			Service:              "skaffold_examples_microservices",
```
tejal29, line 41: What does is service mean here? Sorry I don't have context on `cloud.google.com/go/profiler`?
Shd this service be deployed in the project?

tejal29 (changes requested):
1  nit and more explanation.

cmd/skaffold/skaffold.go
aaron-prindle: Changed s/SKAFFOLD_PROFILE/SKAFFOLD_PROFILER

cmd/skaffold/skaffold.go
aaron-prindle: Changed this to be configurable via env var - SKAFFOLD_PROFILER_SERVICE

tejal29 (approved):

Issue #6067, opened 2021-06-22
Allow tests to be run on any image change


viglesiasce:
### Expected behavior
When defining tests allow the image reference to be either empty or not provided. This would cause tests to run on every iteration. I'd like to be able to setup remote modules with shared tests steps but at the moment you can't do that because the module would have to explicitly call out an image.

### Actual behavior
Image field must be set in the test section so tests cannot be run or shared across repos.

gsquared94:
from discussion: the intention in an ongoing skaffold dev session is to have a single run of a `custom` test when `n` artifacts get rebuilt and not a run against every built artifact (what was attempted in #6070).

viglesiasce:
Ack. That sounds right to me.

PR #6070, opened 2021-06-23
make `test[i].imageName` optional in the skaffold config

gsquared94:
Fixes: #6067 

**Description**

Make the `imageName` field under `test` optional. 

Users can skip specifying `imageName` if they intend for the test to run against all artifacts built in the current devloop.


tejal29:
After some discussion on the thread internal, maybe we need to make test.image optional for custom?

Issue #5986, opened 2021-06-09
Update proto toolchain


briandealwis:
Our toolchain in `hack/proto/Dockerfile` is rather behind, but a simple update of versions doesn't work as the .pb.go files aren't generated.

- update protoc to 3.17.3: if we used a debian image (`golang:1.15` instead of `-alpine`) then we could use the downloaded protoc, but debian's `jq` is 1.5
- update grpc-gateway to v2 — the protoc-gen-swagger has been renamed to protc-gen-openapiv2, but it seems the proto needs some additional annotations
- verify that use of `jq` to rewrite of the generated swagger doc (`title` &rarr; `description`) is necessary?

briandealwis:
I suspect this `jq` malarkey can be handled by setting some [specific options in the proto](https://stackoverflow.com/a/52654286/600339).  I'm not clear what the title/description issues were in the generated swagger though.


This is where the proto is turned into the swagger implementation:
https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-openapiv2/internal/genopenapi/template.go

PR #6073, opened 2021-06-23
Use non alpine image and protoc 3.17.3 in proto generation

MarlonGamez:
**Related**: #5986 

**Description**
This PR makes changes to `/hack/proto/Dockerfile` to have it use a non alpine image, in order to be able to use the downloaded `protoc` binary. Also updates `protoc` version to `3.17.3`

Update to use newer version of `protoc-gen-grpc-gateway` and `protoc-gen-openapiv2`

### DIFF
--- a/hack/proto/Dockerfile
+++ b/hack/proto/Dockerfile
@@ -12,13 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.15-alpine AS generate-files
-RUN apk add --no-cache protobuf=3.13.0-r2 unzip jq moreutils
+FROM golang:1.15 AS generate-files
+RUN apt-get update && apt-get install -y unzip moreutils
 
 WORKDIR /protoc
-ENV PROTOC_VERSION=3.13.0
+ENV PROTOC_VERSION=3.17.3
 RUN wget -O protoc.zip https://github.com/google/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-linux-x86_64.zip
 RUN unzip protoc.zip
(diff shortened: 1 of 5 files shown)


hack/proto/Dockerfile
```diff
@@ -75,8 +76,9 @@ RUN protoc \
   v2/*.proto
 
 # this is a hack - seemingly grpc-gateway-swagger-gen is sometimes generating titles when they should be descriptions
-RUN jq 'walk(if type == "object" and has("title") then .description = ([.title, .description] | map(values) | join ("\n")) | del(.title) else .  end)' v1/skaffold.swagger.json | sponge v1/skaffold.swagger.json
-RUN jq 'walk(if type == "object" and has("title") then .description = ([.title, .description] | map(values) | join ("\n")) | del(.title) else .  end)' v2/skaffold.swagger.json | sponge v2/skaffold.swagger.json
+RUN wget -O jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 && chmod +x ./jq
```
briandealwis, line 79: It shouldn't be this hard to find an image with relatively recent toolsets 😢 
I wonder if grpc-gateway-swagger-gen has fixed this issue, or if it's a configuration issue on our part.

briandealwis (approved):

hack/proto/Dockerfile
MarlonGamez: I'll be looking more into this while testing out the newer versions of `protoc-gen-openapiv2`

breun:
You can also run `protoc` binaries on Alpine Linux if you install Alpine's `gcompat` package ("The GNU C Library compatibility layer for musl"), which provides `ld-linux-x86-64.so.2`.

PR #6075, opened 2021-06-24
Minor cleanup to structure tests

briandealwis:
Inline consts since values were inconsistently referenced, and avoid repetition.


gsquared94 (approved):
lgtm!

PR #6077, opened 2021-06-25
[v2] Major refactoring for v2 runner/runcontext

yuwenma:
**Related**: #5673 

**Description**    

- Seperate runcontext.Runcontext to V1 and V2, each runcontext holds its own v1 and v2 schema
- Store schema configs in their versioned runner. This decouple the versioned functions in cmd (where latestV1 is referred).
- Add Renderer to withTiming


yuwenma:
Moved to #6081

Issue #6074, opened 2021-06-24
Port forward not working for multiple `deploy` stanzas or `multi-config` projects


gsquared94:
From `master` branch, on `examples/getting-started` run:
```
skaffold dev
```

### Expected behavior
Single port-forward request per service successful.

### Actual behavior

Skaffold keeps trying to port-forward and gets stuck in a loop

```
[leeroy-app] 2021/06/24 12:55:44 leeroy app server ready
[leeroy-web] 2021/06/24 12:55:45 leeroy web server ready
Port forwarding deployment/leeroy-app in namespace default, remote port http -> 127.0.0.1:9001
Port forwarding deployment/leeroy-app in namespace default, remote port http -> 127.0.0.1:9002
failed to port forward deployment-leeroy-web-default-8080, port 9000 is taken, retrying...
Port forwarding deployment/leeroy-web in namespace default, remote port 8080 -> 127.0.0.1:9000
failed to port forward deployment-leeroy-web-default-8080, port 9000 is taken, retrying...
failed to port forward deployment-leeroy-web-default-8080, port 9000 is taken, retrying...
failed to port forward deployment-leeroy-web-default-8080, port 9000 is taken, retrying...
failed to port forward deployment-leeroy-web-default-8080, port 9000 is taken, retrying...
....
```
### Information

- Skaffold version: 
- Operating system: 
- Installed via: 
- Contents of skaffold.yaml:

```yaml
<paste your skaffold.yaml here>
```

### Steps to reproduce the behavior

1. a clonable repository with the sample skaffold project
2. `skaffold <command>`
3. ...

gsquared94:
cc. @tejal29 @nkubala 
I think this could be a regression from the recent refactors to `portforwarder` and `podselector`.

PR #6090, opened 2021-06-28
fix: port-forward for multi-config projects

gsquared94:
Fixes: #6074  

**Description**

The issue described in #6074 happens for multi-config projects or even a single config project defining multiple deployers (say `kubectl` and `helm`). This was caused by instantiating multiple `PortForwardManager` structs on every call to `GetKubernetesAccessor` method. These multiple structs each tried to port-forward all user defined ports and only one of them succeeded while the others reported failures.

This PR replaces that with a `map` keyed on the `kubecontext`. This way the same `PortForwardManager` is returned for all kubernetes deployments against the same cluster.

### DIFF
--- a/pkg/skaffold/access/provider.go
+++ b/pkg/skaffold/access/provider.go
@@ -17,8 +17,6 @@ limitations under the License.
 package access
 
 import (
-	"sync"
-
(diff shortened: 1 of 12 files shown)


MarlonGamez (approved):
lgtm

Issue #6100, opened 2021-06-09
Python debugging entrypoint rewrite incorrect if it contains parenthesis.


j-windsor:
**Description of the issue**:
on `skaffold debug` was getting the error `> [contacts-85bf9984d6-hlfkp contacts] /bin/sh: 1: Syntax error: "(" unexpected`, so I escaped the parenthesis, which fixed `debug`, but caused `dev` to error with `> [contacts-8668d9cdbb-f9xch contacts] Failed to parse 'create_app\\(\\)' as an attribute name or function call`. It seems like the rewrite of the entrypoint during `debug` is causing some string escaping issues?

**Expected behavior**:
I do not have to escape the parenthesis and it will work in both dev and debug.

**Steps to reproduce**: 
1. use https://github.com/GoogleCloudPlatform/bank-of-anthos/pull/501 (note that the parenthesis are not escaped)
2. `skaffold dev` works :)
3. `skaffold debug` does not work :(

**Environment**: 
OS: macOS v11.4
Skaffold: v1.25.0

PR #6101, opened 2021-06-28
Escape parentheses in shJoin

briandealwis:
Fixes: #6100

**Description**
In `skaffold debug`, we rewrite container command-lines to run under the applicable language runtime debugger.  We support "sh -c"-style command-lines that result from Dockerfile `RUN command args` by splitting and then re-joining the command-lines.  Our [re-joining helper](https://github.com/GoogleContainerTools/skaffold/blob/044c65fddcd0dcaaacca35c08391d1a8beb4dd1e/pkg/skaffold/debug/transform.go#L522-L538) tries to do minimal escaping, escaping only whitespace and a double-quote.  But command-lines such as [Bank of Anthos's `contacts`] container then fail as they include parentheses, which are interpreted by the shell:

    gunicorn -b :$PORT --threads 4 --log-config /logging.conf --log-level=$LOG_LEVEL "contacts:create_app()"

This PR causes our re-join to quote parentheses, brackets, and braces, and single-quotes too.

This quoting is imperfect as we don't track information such as the `contacts:create_app()` being quoted.

### DIFF
--- a/pkg/skaffold/debug/transform.go
+++ b/pkg/skaffold/debug/transform.go
@@ -527,7 +527,7 @@ func shJoin(args []string) string {
 		if i > 0 {
 			result += " "
 		}
-		if strings.ContainsAny(arg, " \t\r\n\"") {
+		if strings.ContainsAny(arg, " \t\r\n\"'()[]{}") {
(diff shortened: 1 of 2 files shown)


MarlonGamez (approved):
lgtm

PR #6104, opened 2021-06-29
Fetch and set up secret for skaffold-metrics project

tejal29:
In this PR
1) Use secret from newly created  project
2) Drop the 2 weeks validity as longer keys exist.

No code change required for instrumentation as project id is [derived](https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/instrumentation/export.go#L122) from the secret itself.

Earlier dashboard can be used as k8s-skaffold is defines as scoped project https://cloud.google.com/monitoring/settings#concept-scope

### DIFF
--- a/deploy/setup-secret.sh
+++ b/deploy/setup-secret.sh
@@ -17,6 +17,7 @@
 set -x
 # set default project id
 PROJECT_ID="k8s-skaffold"
+METRICS_PROJECT_ID="skaffold-metrics"
 KEY_FILE="./secrets/keys.json"
 BUCKET_ID="k8s-skaffold-secrets"
 LATEST_GCS_PATH="keys.json"
@@ -30,7 +31,7 @@ done
 
 function download_existing_key() {
   # Download a valid key created within the past two weeks.
-  KEY_IDS=$(gcloud iam service-accounts keys list --iam-account=metrics-writer@k8s-skaffold.iam.gserviceaccount.com --project=k8s-skaffold --managed-by=user --filter="validAfterTime>-P2W" --format="value(name)")
+  KEY_IDS=$(gcloud iam service-accounts keys list --iam-account=metrics-writer@${METRICS_PROJECT_ID}.iam.gserviceaccount.com --project=${METRICS_PROJECT_ID} --managed-by=user --format="value(name)")
   while read -r KEY_ID
   do
     if gsutil cp gs://${BUCKET_ID}/${KEY_ID}.json ${KEY_FILE}; then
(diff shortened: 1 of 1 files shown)


deploy/setup-secret.sh
```diff
@@ -30,10 +31,10 @@ done
 
 function download_existing_key() {
   # Download a valid key created within the past two weeks.
-  KEY_IDS=$(gcloud iam service-accounts keys list --iam-account=metrics-writer@k8s-skaffold.iam.gserviceaccount.com --project=k8s-skaffold --managed-by=user --filter="validAfterTime>-P2W" --format="value(name)")
+  KEY_IDS=$(gcloud iam service-accounts keys list --iam-account=metrics-writer@${METRICS_PROJECT_ID}.iam.gserviceaccount.com --project=${METRICS_PROJECT_ID} --managed-by=user --format="value(name)")
   while read -r KEY_ID
   do
-    if gsutil cp gs://${BUCKET_ID}/${KEY_ID}.json ${KEY_FILE}; then
+    if gsutil cp gs://${BUCKET_ID}/${LATEST_GCS_PATH}.json ${KEY_FILE}; then
```
aaron-prindle, line 37: LATEST_GCS_PATH=keys.json so I think this will be `keys.json`.json

deploy/setup-secret.sh
```diff
@@ -30,10 +31,10 @@ done
 
 function download_existing_key() {
   # Download a valid key created within the past two weeks.
-  KEY_IDS=$(gcloud iam service-accounts keys list --iam-account=metrics-writer@k8s-skaffold.iam.gserviceaccount.com --project=k8s-skaffold --managed-by=user --filter="validAfterTime>-P2W" --format="value(name)")
+  KEY_IDS=$(gcloud iam service-accounts keys list --iam-account=metrics-writer@${METRICS_PROJECT_ID}.iam.gserviceaccount.com --project=${METRICS_PROJECT_ID} --managed-by=user --format="value(name)")
```
aaron-prindle, line 34: nit: is this still needed? `--filter="validAfterTime>-P2W"` I'm guessing that w/ the new project and expiration date it isn't, just verifying

deploy/setup-secret.sh
tejal29: fixed.

deploy/setup-secret.sh
tejal29: yeah. new keys live longer and hence we don't need that check.

aaron-prindle (approved):
LGTM

PR #6108, opened 2021-06-29
log metrics upload failure and write to file instead.

tejal29:
While debugging new secret set up, realized we were swallowing the error message when uploads fail


aaron-prindle (approved):
LGTM

tejal29:
Kokoro build failure due to fetching kaniko project image.

PR #6111, opened 2021-06-29
Add information about workspace and dockerfile to artifact metadata

MarlonGamez:
**Description**
This PR adds information about an artifacts `context`, and in some cases `dockerfile`, to metadata events that get emitted through skaffold's event API.


pkg/skaffold/event/v2/metadata.go
```diff
@@ -82,7 +82,8 @@ func getArtifacts(b latestV1.BuildConfig) []*proto.BuildMetadata_Artifact {
 	result := []*proto.BuildMetadata_Artifact{}
 	for _, a := range b.Artifacts {
 		artifact := &proto.BuildMetadata_Artifact{
-			Name: a.ImageName,
+			Name:    a.ImageName,
+			Context: a.Workspace,
```
gsquared94, line 86: `a.Workspace` might just be `.` or a relative path if not set explicitly in the root config. If that's alright, then lgtm.

gsquared94 (approved):

pkg/skaffold/event/v2/metadata.go
MarlonGamez: I think that's alright, I can speak further with IDE folks regarding this

Issue #5968, opened 2021-06-08
Sync only monitors pre-configured namespaces


jt-helsinki:
### Expected behavior

When a file in a watched directory changes the file is built into an image, the image is deployed into the cluster. 

### Actual behavior

Running `skaffold run` works. As does the initial build/deploy of the `skaffold dev` command. The cluster is all up and running and it looks good. The correct image versions (as generated by Skaffold) are being used.

The problem happens when a file changes. 

1. File change is detected which kicks off a build
2. Skaffold tries to sync the changes, this looks ok until a sync error is shown. 
    ```
    DEBU[0328] getting client config for kubeContext: `minikube` 
    WARN[0328] Skipping deploy due to sync error: copying files: didn't sync any files 
    Watching for changes...
    ```
This seems to be a recurring issue based on similar issues posted in the tracker. However the solution is not clear. 

A more useful error message would be helpful. 

### Information

- Skaffold version: v1.25.0

- Operating system: Mac OS 11.3.1 Big Sur
- Installed via: Homebrew

- Multi module maven project. Java, Spring boot. Actuator

Jib plugin in `pom.xml`

```
<plugin>
    <groupId>com.google.cloud.tools</groupId>
    <artifactId>jib-maven-plugin</artifactId>
    <version>3.0.0</version>
    <executions>
        <execution>
            <phase>package</phase>
            <goals>
                <goal>dockerBuild</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```

- Contents of skaffold.yaml:
```yaml
apiVersion: skaffold/v2beta16
kind: Config
build:
  artifacts:
  - image: base-java-image
    context: ./docker/base-images/java-spring
    docker:
      dockerfile: Dockerfile

  - image: authentication-service
    context: ./service
    docker:
    jib:
      project: authentication-service
      args:
        - -Pspring-devtools
        - -Pskip-tests
        - -Pbuild-docker-images-jib
    sync:
      auto: true
    requires:
      - image: base-java-image

  - image: data-manager-service
    context: ./service
    jib:
      project: data-manager-service
      args:
        - -Pspring-devtools
        - -Pskip-tests
        - -Pbuild-docker-images-jib
    sync:
      auto: true
    requires:
      - image: base-java-image

  - image: router-service
    context: ./service
    jib:
      project: router-service
      args:
      - -Pspring-devtools
      - -Pskip-tests
      - -Pbuild-docker-images-jib
    sync:
      auto: true
[...]

PR #6135, opened 2021-07-05
Fetch namespaces at time of sync

briandealwis:
Fixes: #5968

**Description**
The `podSyncer` is initialized with the set of namespaces from the configuration, and thus misses any namespace updates from the deployment.
https://github.com/GoogleContainerTools/skaffold/blob/823896bdd7464d79886664b32f56948957ea5f65/pkg/skaffold/sync/provider.go#L41-L49

This patch causes the config to be passed to the `podSyncer` such that it processes the known namespaces at sync time.


gsquared94 (approved):
lgtm

Issue #5960, opened 2021-06-07
skaffold zsh command completion ends with `ShellCompDirectiveNoFileComp`


Nightreaver:
### Steps to reproduce

- using zsh
- using `source <(skaffold completion zsh)` to load command completion
- hitting [TAB] during commands should print the actual commands, without ( I guess ) the ShellComp... output which can also be tabbed through

### Expected behavior

hitting [TAB] during commands should print only available commands/options

### Actual behavior

hitting [TAB] during commands print additional output. ie

```
~ skaffold deploy [TAB]
:0 Completion ended with directive: ShellCompDirectiveNoFileComp
```

hitting [TAB] again will result in

```
~ skaffold deploy :0
:0 Completion ended with directive: ShellCompDirectiveNoFileComp
```

and again

```
~ skaffold deploy Completion\ ended\ with\ directive:\ ShellCompDirectiveNoFileComp
:0 Completion ended with directive: ShellCompDirectiveNoFileComp
```

which are no valid command options

### Information

- Skaffold version: 1.25.0
- Operating system: Debian 20.04 LTS
- Installed via: binary download

gsquared94:
able to repro on master branch.

marckhouzam:
I've taken the liberty to investigate this issue. It turns out that skaffold sets both stdout and stderr to the same output stream in a `PersistentPreRun`.  This causes a problem for Cobra's `__complete` command which outputs a debugging line on stderr which is normally ignored, but since this line appears on stdout for skaffold, it appears as a (invalid) completion.

Skaffold does this here https://github.com/GoogleContainerTools/skaffold/blob/master/cmd/skaffold/app/cmd/cmd.go#L91

Is this behaviour on purpose?  I happens to be using the deprecated Cobra function `SetOutput()` which sets both stdout and stderr at the same time.  Should Cobra's `SetOut()` be used instead?  
See:
https://github.com/spf13/cobra/blob/9a432671fd847f0faa5a5e4d9f9350ae289db2ac/command.go#L238-L250

Another option is to change this behaviour just when the `__complete` command is called.

PR #6138, opened 2021-07-06
Make completion work again

briandealwis:
Fixes: #5960 

This PR prevents redirecting output for the internal `spf13/cobra` commands used for handling completion support.  Use `cmd.Name()` as `cmd.Use` has [`[command-line]` appended to the name](https://github.com/spf13/cobra/blob/de187e874d1ca382320088f8f6d76333408e5c2e/completions.go#L151) for these internal commands.


gsquared94 (approved):
nice, lgtm!

PR #6139, opened 2021-07-06
Set specific permissions for workflows

briandealwis:
Explicitly set permissions on our github actions to [earn a star](https://github.com/ossf/scorecard/blob/main/checks/checks.md#token-permissions).  We can then reduce our default permissions in the project settings.


aaron-prindle (approved):
LGTM

PR #6144, opened 2021-07-07
[v2] Add events and handler for render stage

yuwenma:
**Related**: #5673 
**Cherry-pick**: Cherry-pick back to v2 branch.

**Description**
Add events and event handlers for the whole render process. The reason not to add the proto in v2 is because the `pkg/skaffold/event/event.go` file only imports v1 proto.


MarlonGamez (approved):
lgtm 👍🏼

PR #6159, opened 2021-07-08
hooks: add hooks schema, some interface refactor

gsquared94:
### Hooks PR 1 of n

Related: #1441, #6147

Please refer to #6147 for the complete set of changes. This PR only modifies the following:
- add new config structs, excluded from the yaml parser.
- add `PushImage() bool` method to the `PipelineBuilder` interface

### DIFF
--- a/docs/content/en/schemas/v2beta19.json
+++ b/docs/content/en/schemas/v2beta19.json
@@ -576,6 +576,34 @@
       "description": "contains all the configuration for the build steps.",
       "x-intellij-html-description": "contains all the configuration for the build steps."
     },
+    "BuildHooks": {
+      "properties": {
+        "after": {
+          "items": {
+            "$ref": "#/definitions/HostHook"
+          },
+          "type": "array",
+          "description": "describes the list of lifecycle hooks to execute *after* each artifact build step.",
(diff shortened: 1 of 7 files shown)


yuwenma:
LGTM! Once merged, this PR would need to be cherry-picked to v2 featured branch (schema needs to be added to v3alpha)

yuwenma (approved):

gsquared94:
the patch code coverage will be made up for in follow up PRs, since this PR was only for adding the schema definition.

tejal29:
@gsquared94 Was there a reason why we added `PushImage` to the Builder method?
Somewhere during code review, i remember @nkubala mentioning we should move the code to push to deployer instead of run context.

gsquared94:
> @gsquared94 Was there a reason why we added `PushImage` to the Builder method?

yes, the `PipelineBuilder` is the interface for the build environment `gcb`, `cluster` and `local`. So `PushImage` should be a property of the `PipelineBuilder`. 
> Somewhere during code review, i remember @nkubala mentioning we should move the code to push to deployer instead of run context.
I don't think that'd work if we want to keep `build` i.e  (build, tag & push) as an independent command separate from `deploy`.

PR #6160, opened 2021-07-08
hooks:  `sync` refactor

gsquared94:
### Hooks PR 2 of n

Related: #1441, #6147, #6159

Please refer to #6147 for the complete set of changes. This PR only modifies the following:
- add `v1.Artifact` reference to `podSyncer`
we want to be able to reference the `v1.Artifact.Sync.Hooks` struct from within the podSyncer which run the pre and post sync hooks.
- add `io.Writer` parameter to `Sync` method 
we need an output writer for the hooks executor.
~~Merge after #6159, only review last commit.~~

### DIFF
--- a/pkg/skaffold/sync/sync_test.go
+++ b/pkg/skaffold/sync/sync_test.go
@@ -758,7 +758,9 @@ func TestNewSyncItem(t *testing.T) {
 			})
 
 			actual, err := NewItem(ctx, test.artifact, test.evt, test.builds, &mockConfig{}, 0)
-
+			if test.expected != nil {
+				test.expected.Artifact = test.artifact
+			}
 			t.CheckErrorAndDeepEqual(test.shouldErr, err, test.expected, actual)
 		})
 	}
--- a/pkg/skaffold/runner/v1/dev.go
+++ b/pkg/skaffold/runner/v1/dev.go
@@ -92,7 +92,7 @@ func (r *SkaffoldRunner) doDev(ctx context.Context, out io.Writer) error {
 			output.Default.Fprintf(out, "Syncing %d files for %s\n", fileCount, s.Image)
 			fileSyncInProgress(fileCount, s.Image)
 
-			if err := r.deployer.GetSyncer().Sync(childCtx, s); err != nil {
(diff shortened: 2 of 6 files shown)


MarlonGamez (approved):
lgtm!

pkg/skaffold/sync/sync_test.go
```diff
@@ -758,7 +758,9 @@ func TestNewSyncItem(t *testing.T) {
 			})
 
 			actual, err := NewItem(ctx, test.artifact, test.evt, test.builds, &mockConfig{}, 0)
-
+			if test.expected != nil {
```
yuwenma, line 761: Normally we change each test data (I know it may be more work) in lieu of adding additional logic in the unittest, so as we don't need a unittest for unittests. 🤷

yuwenma (approved):
a minor comment

tejal29:
> ### Hooks PR 2 of n
> Related: #1441, #6147, #6159
> 
> Please refer to #6147 for the complete set of changes. This PR only modifies the following:
> 
> * add `v1.Artifact` reference to `podSyncer`
> * add `io.Writer` parameter to `Sync` method
> 
> ~Merge after #6159, only review last commit.~

Can you please add motivation behind this interface change?
Why do we need out to the Syncer.Sync interface? How was out passed before?

Same for `Artifact`

tejal29 (changes requested):
Please add motivation for the change

gsquared94:
> Can you please add motivation behind this interface change?
> Why do we need out to the Syncer.Sync interface? How was out passed before?
> 
> Same for `Artifact`

added to the description:

- add `v1.Artifact` reference to `podSyncer`
we want to be able to reference the `v1.Artifact.Hooks` struct from within the `podSyncer` which run the pre and post sync hooks.
- add `io.Writer` parameter to `Sync` method
we need an output writer for the hooks executor.

Previously the `Syncer` did not log any output other than logrus DEBUG messages. The `v1.Artifact` struct also wasn't required and it only needed the tagged image name.

tejal29 (approved):

Issue #6166, opened 2021-07-08
To ask helm users to fill a new helm deployer specific survey


tejal29:
TODO fill more.

tejal29:
Add a framework to show surveys for particular usages,

tejal29:
Design at #6185

tejal29:
Only Pr remaining is #6197 which is not required but good to go in this release

PR #6186, opened 2021-07-11
design proposal to show user survey other than Hats

tejal29:
related to #6166 

This is a design proposal to show user survey (other than Hats) to skaffold users.

### DIFF
--- a/docs/design_proposals/user_survey_hooks.md
+++ b/docs/design_proposals/user_survey_hooks.md
@@ -25,12 +25,15 @@ type config struct {
 	// promptText is shown to the user and should be formatted so each line should fit in < 80 characters.
 	// For example: `As a Helm user, we are requesting your feedback on a proposed change to Skaffold's integration with Helm.`
 	promptText   string
+	// startsAt mentions the date after the users survey should be prompted. This will ensure, skaffold team can finalize the survey 
+	// even after release date.
+	startsAt time.Time
 	// expiresAt places a time limit of the user survey. As users are only prompted every two weeks
 	// by design, this time limit should be at least 4 weeks after the upcoming release date to account
 	// for release propagation lag to Cloud SDK and Cloud Shell.
 	expiresAt    time.Time
 	isRelevantFn func([]util.VersionedConfig) bool
-	surveyURL         string
+	URL         string
 }
 
 ```
@@ -41,22 +44,23 @@ The survey config has two key fields
 {
   id: helmID,
   expiresAt: time.Date(2021, time.August, 14, 00, 00, 00, 0, time.UTC),
-  isRelevantFn: func(cfgs []util.VersionedConfig) bool {
+  isRelevantFn: func(cfgs []util.VersionedConfig, command string) bool {
 	for _, cfg := range cfgs {
-		v1Cfg := cfg.(*latestV1.SkaffoldConfig)
-		if h := v1Cfg.Deploy.HelmDeploy; h != nil {
-			return true
+		if v1Cfg, ok  := cfg.(*latestV1.SkaffoldConfig) ; ok {
+		    if h := v1Cfg.Deploy.HelmDeploy; h != nil {
+			    return true
 	        }
-        }
+	    }    
+    }
 	return false
  },
-link: helmURL,
+URL: helmURL,
 },
 
 ```
 For multi-module users, we could use something like the following:
 ```
-  isRelevantFn: func(cfgs []util.VersionedConfig) bool {
+  isRelevantFn: func(cfgs []util.VersionedConfig, _ string) bool {
 	return len(cfgs) > 1
  },
 ```
@@ -111,6 +115,8 @@ Since, HaTS survey never expire,
 ## Alternate Methods
 Other methods to get feedback is manually via pinging slack users to fill in a survey.
 However, this requires a core member to manually remind users from time to time to fill in a survey.
+Another disadvantage is users could form a biased opinion due to the questions asked in the user survey and 
+rate low on the NPS score.
 
 
 ## Implementation plan
--- a/docs/design_proposals/user_survey_hooks.md
+++ b/docs/design_proposals/user_survey_hooks.md
@@ -1,0 +1,135 @@
+# Title
+
+* Author(s): Tejal Desai
+* Design Shepherd: Brian de Alwis
+* Date: 07/11/2021
+* Status: Proposed
+
+## Background
+
+Currently, we get feedback on skaffold via skaffold HaTs survey. 
+However, in this survey we don't get feedback on new or existing features. 
+We also can't get enough information from our existing customers on how they are using certain features in case we want to make any changes.
+e.g.
+With render v2, we are changing how deploy works. This may affect our existing helm deployer users. 
+See [#6166](https://github.com/GoogleContainerTools/skaffold/issues/6166)
+## Design
+
+### User survey Definition
+In order to make sure only relevant surveys are shown to user, we have added a survey config.
+```
+type config struct {
+	id           string
+	promptText   string
+	expiresAt    time.Time
+	isRelevantFn func([]util.VersionedConfig) bool
+	link         string
+}
+
+```
+The survey config has two key fields
+1) expiresAt - This decided the lifetime of the survey. e.g. if we are collecting user feedback to get feedback on helm deployer re-design, then it makes sense to target the survey towards helm deployer user for a limited time until the re-design requirement phase is complete.
+2) isRelevantFn - This function determines if the user is a target audience for the survey. For the above example, it only makes sense to show the survey for helm deployer users, like this.
+```
+{
+  id: helmID,
+  expiresAt: time.Date(2021, time.August, 14, 00, 00, 00, 0, time.UTC),
+  isRelevantFn: func(cfgs []util.VersionedConfig) bool {
+	for _, cfg := range cfgs {
+		v1Cfg := cfg.(*latestV1.SkaffoldConfig)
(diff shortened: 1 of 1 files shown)


docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+With render v2, we are changing how deploy works. This may affect our existing helm deployer users. 
+See [#6166](https://github.com/GoogleContainerTools/skaffold/issues/6166)
+## Design
+
+### User survey Definition
+In order to make sure only relevant surveys are shown to user, we have added a survey config.
+```
+type config struct {
+	id           string
+	promptText   string
```
briandealwis, line 23: Worth describing what is expected here.  Is this the only text to be shown, or will we add other boilerplate?
```suggestion
	// promptText is shown to the user and should be formatted so each line should fit in < 80 characters.
	// For example: `As a Helm user, we are requesting your feedback on a proposed change to Skaffold's integration with Helm.`
	promptText   string
```

docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+The survey config has two key fields
+1) expiresAt - This decided the lifetime of the survey. e.g. if we are collecting user feedback to get feedback on helm deployer re-design, then it makes sense to target the survey towards helm deployer user for a limited time until the re-design requirement phase is complete.
+2) isRelevantFn - This function determines if the user is a target audience for the survey. For the above example, it only makes sense to show the survey for helm deployer users, like this.
+```
+{
+  id: helmID,
+  expiresAt: time.Date(2021, time.August, 14, 00, 00, 00, 0, time.UTC),
+  isRelevantFn: func(cfgs []util.VersionedConfig) bool {
+	for _, cfg := range cfgs {
+		v1Cfg := cfg.(*latestV1.SkaffoldConfig)
```
briandealwis, line 39: I think this will panic if not v1 :-)

docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+ },
+```
+
+### How to prompt users to take user surveys
+
+#### Rules to show survey prompt
+When prompting users with surveys, we need to keep in mind 2 important rules
+1) Don't prompt users too often.
+   Currently, we only prompt users to fill in HaTs survey every two weeks until they fill.
+2) Don't prompt users if they have already taken the survey.
```
briandealwis, line 63:
```suggestion
1) *Don't prompt users too often.*
   Currently, we only prompt users to fill in HaTs survey every two weeks until they fill.
2) *Don't prompt users if they have already taken the survey.*
```

docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+   Currently, we only prompt users to fill in Hats Survey if they haven't taken it in last 3 months.
```
briandealwis, line 64:
```suggestion
   Currently, we only prompt users to fill in HaTS Survey if they haven't taken it in last 3 months.
```

docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+
+For non HaTs surveys or user surveys, we will follow the same rules,
```
briandealwis, line 66:
```suggestion
For non HaTS surveys or user surveys, we will follow the same rules,
```

docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+1) Prompt users to fill in the survey once 2 weeks if its relevant to them until they fill it.
```
briandealwis, line 67: Do you think we should revise this down to 1 week?  What if we have two surveys going (e.g., Helm and `test`).  Or will we prompt them for each possible survey?

docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+
+```
+
+Alternate options:
+1) Another option is to show the Hats survey right after user survey. 
+   Look into if we can keep the same Hats response Form.
+
+## Metrics
+1) Measure sample responses to user surveys
+2) Monitor the number of Hats responses per month is not reducing.
```
briandealwis, line 123:
```suggestion
### Implementation details.
The current `ShouldDisplaySurveyPrompt` returns true only if
1) If survey prompt is not disabled and
2) If HaTS survey was not taken in last 3 months (stored in  `SurveyConfig.LastTaken`) and
3) If survey prompt was not shown `SurveyConfig.LastPrompted` in last 2 weeks.

This behavior will be changed to `ShouldDisplaySurveyPrompt` returns true if
1) If survey prompt is not disabled and
2)  If survey prompt was not shown `SurveyConfig.LastPrompted` in last 2 weeks and
3) If there is an active relevant user survey is available or HaTS survey was not taken in last 3 months

If both active relevant user survey is available or HaTS survey was not taken, then Skaffold will give preference to user survey over hats survey.
Since, HaTS survey never expire,
- if user takes the user survey this time, HaTS survey will be prompted the next time.
-  if the user never takes the user survey, the HaTS survey will be prompted once the user survey expires.
   **Note User surveys should not run longer than a quarter so that we don't reduce the volume of HaTS Survey,
```
pkg survey/config

def init() {
  for _, s := range surveys {
     if s.id != hats.id {
       if s.expiresAt.IsZero() || s.expiresAt.Sub(time.Now) > 90days
          panic(fmt.Errorf("survey %s is running longer than 90 days. This will starve users from being shown the hats survey."))
    }
  }
}

```

Alternate options:
1) Another option is to show the HaTS survey right after user survey. 
   Look into if we can keep the same HaTS response Form.

## Metrics
1) Measure sample responses to user surveys
2) Monitor the number of HaTS responses per month is not reducing.
```

docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+          panic(fmt.Errorf("survey %s is running longer than 90 days. This will starve users from being shown the hats survey."))
+    }
+  }
+}
+
+```
+
+Alternate options:
+1) Another option is to show the Hats survey right after user survey. 
+   Look into if we can keep the same Hats response Form.
```
briandealwis, line 119: The user survey may colour their response to the HaTS (_they're doing what!? I'm not going to promote Skaffold!_)

docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+See [#6166](https://github.com/GoogleContainerTools/skaffold/issues/6166)
+## Design
+
+### User survey Definition
+In order to make sure only relevant surveys are shown to user, we have added a survey config.
+```
+type config struct {
+	id           string
+	promptText   string
+	expiresAt    time.Time
```
briandealwis, line 24:
```suggestion
	// expiresAt places a time limit of the user survey. As users are only prompted every two weeks
	// by design, this time limit should be at least 4 weeks after the upcoming release date to account
	// for release propagation lag to Cloud SDK and Cloud Shell.
	expiresAt    time.Time
```

docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+```
+something similar can be imaged for multi-module users
+```
+  isRelevantFn: func(cfgs []util.VersionedConfig) bool {
+	return len(cfgs) > 1
+ },
+```
+
+### How to prompt users to take user surveys
+
```
briandealwis, line 58: Maybe show an example of how the current HaTS prompt looks like, and how the proposal will look like.

docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+## Background
+
+Currently, we get feedback on skaffold via skaffold HaTs survey. 
+However, in this survey we don't get feedback on new or existing features. 
+We also can't get enough information from our existing customers on how they are using certain features in case we want to make any changes.
+e.g.
+With render v2, we are changing how deploy works. This may affect our existing helm deployer users. 
+See [#6166](https://github.com/GoogleContainerTools/skaffold/issues/6166)
+## Design
```
briandealwis, line 16:
```suggestion

## Design

This document proposes to extend the existing HaTS survey framework to incorporate a set of user surveys.
```

briandealwis (approved):
Nice!  A few suggestions and very minor nits.

docs/design_proposals/user_survey_hooks.md
```diff
@@ -0,0 +1,135 @@
...
+
+```
+The survey config has two key fields
+1) expiresAt - This decided the lifetime of the survey. e.g. if we are collecting user feedback to get feedback on helm deployer re-design, then it makes sense to target the survey towards helm deployer user for a limited time until the re-design requirement phase is complete.
+2) isRelevantFn - This function determines if the user is a target audience for the survey. For the above example, it only makes sense to show the survey for helm deployer users, like this.
+```
+{
+  id: helmID,
+  expiresAt: time.Date(2021, time.August, 14, 00, 00, 00, 0, time.UTC),
+  isRelevantFn: func(cfgs []util.VersionedConfig) bool {
```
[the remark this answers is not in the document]
tejal29: yep!

docs/design_proposals/user_survey_hooks.md
tejal29: makes sense. Having them separate would remove any bias.

Issue #6189, opened 2021-07-12
Retire kube context based global config


tejal29:
With Multimodule feature, skaffold can now deploy to  multiple contexts.

We need to re-evalute loading global config based off active context.

PR #6191, opened 2021-07-13
refactor: Read globalConfig instead of kubecontext config for survey config

tejal29:
relates to epic issue #6166 
relates to #6189  
However not completely fixes it. We still need to revisit all occurrences of `GetConfigForCurrentKubectx` and maybe remove them.

Also note, we only update global config for all survey fields. https://github.com/GoogleContainerTools/skaffold/blob/db90d3815cc3657144cb1b23bbcc79f2eea7279f/pkg/skaffold/survey/survey.go#L101


gsquared94 (approved):
lgtm

PR #6194, opened 2021-07-13
Fix survey active logic

tejal29:
relates to #6166 
In #6185 missed the check to make sure the survey start date is in past.

### DIFF
--- a/pkg/skaffold/survey/config.go
+++ b/pkg/skaffold/survey/config.go
@@ -60,7 +60,8 @@ type config struct {
 }
 
 func (s config) isActive() bool {
-	return s.expiresAt.IsZero() || s.expiresAt.After(time.Now())
+	return s.expiresAt.IsZero() ||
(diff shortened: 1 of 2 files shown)


pkg/skaffold/survey/config.go
```diff
@@ -60,7 +60,8 @@ type config struct {
 }
 
 func (s config) isActive() bool {
-	return s.expiresAt.IsZero() || s.expiresAt.After(time.Now())
+	return s.expiresAt.IsZero() ||
+		(s.startsAt.Before(time.Now()) && s.expiresAt.After(time.Now()))
```
briandealwis, line 64: Funny: I found this really hard to reason through — we must think of time differently!

I'd have written this as "we're after the start and before the end":

    time.Now().After(s.startsAt) && time.Now().Before(s.expiresAt)

though there's a small boundary condition there that your formulation deals with 😆

briandealwis (approved):

PR #6208, opened 2021-07-14
Remove unneeded `Origin` field from `SkaffoldLogEvent`

MarlonGamez:
**Description**
This PR removes the `Origin` field from the `SkaffoldLogEvent` type. This field was originally thought to be needed, but IDE teams have said it won't be.


tejal29 (commented):
fyi: we are changing the field number in the proto which is a backward incompatible change.
However, this is not used in prod and hence ok to modify.

tejal29 (approved):

MarlonGamez:
@tejal29 thanks for the review, I figured it'd be okay since it's still WIP. Thanks for clarification

proto/v2/skaffold.proto
```diff
@@ -187,9 +187,8 @@ message MetaEvent {
 message SkaffoldLogEvent {
     string task_id = 1; // id of the task of skaffold that this log came from
     string subtask_id = 2; // id of the subtask that the log came from
-    string origin = 3; // which tool the output came from ex: skaffold, docker
-    enums.LogLevel level = 4; // log level
-    string message = 5; // contents of the log
+    enums.LogLevel level = 3; // log level
+    string message = 4; // contents of the log
```
briandealwis, line 191: I'm on the fence about this change.  @matthewmichihara has the new output been published anywhere (e.g., nightlies or insiders) and been enabled?  If not are you ok with refreshing and regenerating the protos?  Otherwise I think it's better to mark field 3 / origin as reserved.

matthewmichihara: thanks for checking brian. we haven't really published the new output logging anywhere yet, but i'd vote to just mark field 3 as reserved just to be safe.

MarlonGamez: sounds good, will do

PR #6234, opened 2021-07-16
Set output event context in cache check, tag generation, status check, port forward

MarlonGamez:
related: #5368 

**Description**
This PR makes more calls to `output.WithEventContext()` in order to have `SkaffoldLogEvent`s properly tagged for those parts of execution. This helps IDEs with sorting output.


tejal29 (approved):
We should think about adding an integration test for v2 events api for a sample app like microservices.

Issue #4081, opened 2020-05-01
Image not recognized in crd k8s manifest


simonfr:
When I try to launch `skaffold dev` with custom crd image field in manifest is not found by skaffold.

### Expected behavior
Image should be found by skaffold

### Actual behavior
Image not found so --> Not used by the deployment error

### Information

- Skaffold version: v1.8.0
- Operating system: Ubuntu 18.04 LTS
- Contents of skaffold.yaml: 

```yaml
apiVersion: skaffold/v2beta2
kind: Config
build:
  artifacts:
    - image: localhost:5000/myimage
      context: .
      custom:
        buildCommand: faas-cli build -f api.yml
  local:
    push: true
deploy:
  kubectl:
    manifests:
      - "k8s/*.yaml"
portForward:
- resourceType: Service
  resourceName: gateway
  namespace: openfaas
  port: 8080
```

- Contents of k8s/function.yaml:
```yaml
apiVersion: openfaas.com/v1alpha2
kind: Function
metadata:
  name: myfunction
  namespace: openfaas-fn
spec:
  name: my-fn
  image: localhost:5000/myimage
```

- What I see :
```
WARN[0000] image [localhost:5000/myimage] is not used by the deployment
```

### Steps to reproduce the behavior

1. a clonable repository with the sample skaffold project
2. `skaffold dev`

balopat:
Thank you for opening @simonfr! 

We should think about how to whitelist CRDs in Skaffold. 
One dirty-quick fix would be to add `"Function": true` to this map: 
https://github.com/GoogleContainerTools/skaffold/blob/master/pkg/skaffold/deploy/kubectl/visitor.go#L33
However, I would like to see at least the kind to be qualified with the API group. There could be other `Function` CRDs out there, that users might not want to put through image replacement by skaffold.

simonfr:
Thank you @balopat 
May be it's possible to add to current [skaffold.yaml](https://skaffold.dev/docs/references/yaml/) reference a new map in deploy.kubectl.transformableKinds where we can enable, disable and add parameters to existing recursivelyTransformableKinds map

whyvez:
I've hit the same issue trying to use Skaffold to build and deploy an Argo WorkflowTemplate.

foobarbecue:
@whyvez, I have exactly the same problem. Did you find a workaround?

lukebond:
[...]

PR #6236, opened 2021-07-16
[design proposal] Add config option 'deploy.config.transformableAllowList'

shawnzhu:
Supports: #4081 

**Description**
This is important for working with CRDs that needs transform by skaffold.

It introduces new schema change so user may need to update the `skaffold.yaml` to pick up this new feature

### DIFF
--- a/docs/design_proposals/configurable-transformable-allowlist.md
+++ b/docs/design_proposals/configurable-transformable-allowlist.md
@@ -1,0 +1,74 @@
+# Configurable transformableAllowList
+
+* Author(s): Ke Zhu (@shawnzhu)
+* Design Shepherd: Tejal Desai (tejal29)
+* Date: 2021-07-16
+* Status: Draft
+
+## Objectives
+
+Configurable transformableAllowList for transforming manifests.
+
+## Background
+
+Skaffold can only transform manifests from a non-extensible allowlist. When
+using any CRD out of this allowlist, skaffold can not transform it.
+
+Open issues concerning this problem:
+
+* Image not recognized in crd k8s manifest ([#4081](https://github.com/GoogleContainerTools/skaffold/issues/4081))
+
+There was comments out of the above issue to make this allowlist extensible.
+
+The goal of this document is to create an agreement on the configuration option
+and specification to extend `skaffold.kubernetes.manifest.transformableAllowList`.
+
+## Design
+
+Add configuration `deploy.config.transformableAllowList` in `skaffold.yaml`:
+
+Notice that any new configuration option will be appended to existing allowlist. 
+
+### Detailed discussion
+
+Option in `skaffold.yaml`
+
+```YAML
+deploy:
+  config:
+    transformableAllowList:
+    - Group: example.com
+      Kind: Application
+    - Group: argoproj.io
+      Kind: Workflow
+    - Group: tekton.dev
+      Kind: Task
+```
+
+## Open issues/Questions
+
+Since it is an allowlist, neither options could disable transformation on any
+built-in resource like `ReplicaSet` or `Deployment`.
+
+Is there any need to work out a deny list?
+
+## Implementation plan
+
+1. `pkg/skaffold/schema/latest/v1/config.go` - Add config option
+`transformableAllowList` to `DeployConfig`.
+2. `pkg/skaffold/kubernetes/manifest/visitor.go` - Add new parameter `transformableAllowList` 
+to `*ManifestList.Visit()` by appending it to existing coded `transformableAllowList`
+3. `pkg/skaffold/kubernetes/manifest/images.go` - Add new parameter to `*ManifestList.ReplaceImages()`
+to support given `transformableAllowList`
+4. Instrument each deployer to use the new parameter `transformableAllowList`
+
+## Integration test plan
+
+Please describe what new test cases you are going to consider.
+
+1.  Unit and integration tests for `visitor.go`.
+
+    The integration tests should be written to catch situations such as this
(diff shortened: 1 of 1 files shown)


docs/design_proposals/configurable-transformable-allowlist.md
```diff
@@ -0,0 +1,91 @@
...
+  - Group: argoproj.io
+    Kind: Workflow
+  - Group: tekton.dev
+    Kind: Task
+```
+
+Option in global Skaffold config
+
+```YAML
+global:
```
tejal29, line 55: Global config setting is not sharable and I would rather not support this here right now.

docs/design_proposals/configurable-transformable-allowlist.md
```diff
@@ -0,0 +1,91 @@
...
+| 2.         |Global Skaffold config|give users the possibility to extend the default transformableAllowList globally or per project.|
+
+Notice that any new configuration option will be appended to existing allowlist. 
+
+### Detailed discussion
+
+Option in `skaffold.yaml`
+
+```YAML
+deploy:
```
tejal29, line 42: I don't have a good suggestion for this. Not a fan of putting this `deploy`  since this is applicable to deploy and debug.
maybe something like `deploySettings` or `config`. I am not sure.

docs/design_proposals/configurable-transformable-allowlist.md
```diff
@@ -0,0 +1,91 @@
+# Configurable transformableAllowList
+
+* Author(s): Ke Zhu (@shawnzhu)
+* Design Shepherd: \<skaffold-core-team-member\>
```
tejal29, line 4:
```suggestion
* Design Shepherd: Tejal Desai (tejal29)
```

tejal29 (commented):
cc @viglesiasce for design input.

docs/design_proposals/configurable-transformable-allowlist.md
shawnzhu: make sense. let me focus on portable configuration only via `skaffold.yaml`

shawnzhu:
@tejal29 would you please review again?

tejal29:
> @tejal29 would you please review again?

Thanks @shawnzhu. Bringing this in today's team meeting. Will update my review based on discussion there.

tejal29:
We missed covering this in Wed's meeting. I will update the DD soon. Thanks!

docs/design_proposals/configurable-transformable-allowlist.md
```diff
@@ -0,0 +1,74 @@
...
+### Detailed discussion
+
+Option in `skaffold.yaml`
+
+```YAML
+deploy:
+  config:
+    transformableAllowList:
+    - Group: example.com
+      Kind: Application
```
briandealwis, line 41: It would be nice if this could be more compact.  And we should also be able to specify versions here too as they may have dramatically different shapes.  In Kubernetes, these objects are referred to as [_resource types_](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions), and they have versions.  

Then I think we should allow some configurability to determine whether (and where) we can perform certain transforms.  Currently Skaffold:
  - *applies labels:* Skaffold looks for any field named `metadata` and add a `labels` block if missing. This action breaks CRDs, which is why we blocked this behind an allowlist.
  - *rewrite images:* Skaffold looks for any field named `image`

So I suggest something like the following for the simple cases (rewrite everything based on field names):
```
- type: pod # no group, implicitly all versions
- type: batch/Job # group, implicitly all versions
- type: extensions/v1beta1/Deployment  # specific version
```

And we might try something like adding `images` and `labels` fields that provide a list of JSON-path-like paths. Maybe with `*` support?
```
- type: openfaas.com/v1/Function
  images: [spec.image]
  labels: [spec.metadata.labels, spec.labels]    # https://www.openfaas.com/blog/manage-functions-with-kubectl/
- type: apps/v1/Deployment
  images: [spec.template.spec.initContainers.*.image, spec.template.spec.containers.*.image]
  labels: [spec.metadata.labels, spec.template.metadata.labels]
```

Nit: these fields should be lower-cased in keeping with the rest (k8s yaml uses lower case)

briandealwis (commented):
Thanks for this proposal!  One comment below.

docs/design_proposals/configurable-transformable-allowlist.md
shawnzhu: I like such compact configuration and the explict specification for `images` and `labels`.

How about this:

1. adopt the compact configuration
2. implement applies labels so it only rewrite allowed labels of resource types in allowlist
3. only rewrite image of resource types in allowlist

Any guidance on the implementation like which module to change would also be appreicated!

shawnzhu: https://github.com/GoogleContainerTools/skaffold/pull/6236/commits/6a048ee83deff7d30d75edd54f2652147d980feb

tejal29 (approved):

tejal29:
@shawnzhu Sorry for the delay! LGTM!

Do you want to start working on it?

shawnzhu:
@tejal29 Thanks! and yes, let me try.

briandealwis:
Hi @shawnzhu — do you have the bandwidth to continue on with this effort?  Can I help?

shawnzhu:
@briandealwis I'm glad you can help. I was learning how to parse 3 different expressions of resource types with k8s API, but haven't figure out a clue yet. if you can provide pointers or start work on one expression of resource type, that would be awesome.

PR #6238, opened 2021-07-19
Make skaffold reproducible

jelly:
Use SOURCE_DATE_EPOCH to set the previous build date when rebuilding a
built binary.

Motivation: https://reproducible-builds.org/

**Description**

To make this package reproducible in Arch Linux and other distributions, we require SOURCE_DATE_EPOCH to be read if set, so when we are rebuilding an old package we can set the old build date which removes the build variation.

The problem can be spotted here https://reproducible.archlinux.org/api/v0/builds/137544/diffoscope

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,8 +43,15 @@ ifeq "$(strip $(VERSION))" ""
 	override VERSION = $(shell git describe --always --tags --dirty)
 endif
 
+DATE_FMT = +%Y-%m-%dT%H:%M:%SZ
+ifdef SOURCE_DATE_EPOCH
+    BUILD_DATE ?= $(shell date -u -d "@$(SOURCE_DATE_EPOCH)" "$(DATE_FMT)" 2>/dev/null || date -u -r "$(SOURCE_DATE_EPOCH)" "$(DATE_FMT)" 2>/dev/null || date -u "$(DATE_FMT)")
+else
+    BUILD_DATE ?= $(shell date "$(DATE_FMT)")
+endif
+
 GO_LDFLAGS = -X $(VERSION_PACKAGE).version=$(VERSION)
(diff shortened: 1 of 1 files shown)


tejal29:
@jelly just curious, how would you know what  should be the value of `SOURCE_DATE_EPOCH` when trying to reproduce the build for a date in past?

jelly:
> @jelly just curious, how would you know what should be the value of `SOURCE_DATE_EPOCH` when trying to reproduce the build for a date in past?

Arch Linux ships skaffold as an [Arch Linux package](https://archlinux.org/packages/community/x86_64/skaffold/) when building we record the build time and store this with other meta-data about the build environment in a .BUILDINFO file.

When we reproduce the build package, we extract the meta-data and re-construct the build environment and set the SOURCE_DATE_EPOCH variable to the build time of the package. This makes the reproduced package to be build with the exact same time as the original package.

More information can be found in these links:
https://man.archlinux.org/man/core/pacman/makepkg.8.en#REPRODUCIBILITY
https://reproducible-builds.org/docs/source-date-epoch/

tejal29 (approved):

tejal29:
> > @jelly just curious, how would you know what should be the value of `SOURCE_DATE_EPOCH` when trying to reproduce the build for a date in past?
> 
> Arch Linux ships skaffold as an [Arch Linux package](https://archlinux.org/packages/community/x86_64/skaffold/) when building we record the build time and store this with other meta-data about the build environment in a .BUILDINFO file.
> 
> When we reproduce the build package, we extract the meta-data and re-construct the build environment and set the SOURCE_DATE_EPOCH variable to the build time of the package. This makes the reproduced package to be build with the exact same time as the original package.
> 
> More information can be found in these links:
> https://man.archlinux.org/man/core/pacman/makepkg.8.en#REPRODUCIBILITY
> https://reproducible-builds.org/docs/source-date-epoch/

Thanks for the explanation.

Issue #6060, opened 2021-06-22
Skaffold should offer a command-line flag to cause execution to be performed alongside the Skaffold config file.


aaron-prindle:
(taken from design doc by @briandealwis)

### Introduction
Skaffold has a mismatch between the execution models of:
the root module vs dependent modules
the command-line and the IDEs
We see issues such as https://github.com/GoogleCloudPlatform/cloud-code-vscode/issues/379 / https://github.com/GoogleContainerTools/skaffold/issues/5355, and more.

### Background / Motivation
- Skaffold normally resolves paths from the working directory where it was invoked
  - Paths in `skaffold build -f dir/to/skaffold.yaml` are not resolved from `dir/to`
  - This is used by Cloud Code and `gcloud code` for referencing generated skaffold.yaml placed in $TMP
- This behaviour follows the common Unix/CLI standards:
  - `docker build -f dev/Dockerfile .` files are resolved from `.` (or the path specified), not `dev`
  - `go build`: the go.mod is resolved from cwd
  - `make -f dev/Makefile all`
- …except this isn't the case for referenced modules, where Skaffold appears to run from the directory hosting the skaffold.yaml
- When the IDEs launch Skaffold, they currently set the working directory to the directory containing the skaffold.yaml.  But this is not visible in the UI and so users can be confused when they try to run Skaffold from the command-line.

### Design / Proposal
Skaffold should offer a command-line flag to cause execution to be performed alongside the Skaffold config file.

#### Alternative flag to `-f` (new flag `-F`) to implicitly use module-mode
```
$ skaffold build -F path/to/skaffold.yaml
```
Skaffold will effectively set its working directory to be alongside the specified config file. This flag would be mutually exclusive with `-f`.
- Pros:
  - Looks similar to `-f`
  - Supports alternative config file names
- Cons:
  - Looks similar to `-f`
  - Inconsistent with modules as a module only uses a file named `skaffold.yaml`

### Alternative Solutions
See alternative solutions here - https://gist.github.com/aaron-prindle/386cc4d0018e55c5615859584ffbb8fa

aaron-prindle:
Adding @maggieneterval @ekupershlak @plumpy for partnership tracking

aaron-prindle:
@maggieneterval can you add perhaps a log snippet, example, etc. with an example of the issue you are seeing currently and verify that the `-F` flag design above would fix the issue?  Just want to make sure this will fix the issue.  Thanks

[...]

PR #6245, opened 2021-07-19
Fix Workdir error when --filename flag is used.

yuwenma:
**Related**: #5673 

**Description**
When --filename (skaffold.yaml) is used, the workdir shall be switched to where the custom skaffold.yaml lives. But current skaffold does not support switching the workdir.  

@dgageot

### DIFF
--- a/pkg/skaffold/runner/runcontext/context.go
+++ b/pkg/skaffold/runner/runcontext/context.go
@@ -19,6 +19,7 @@ package runcontext
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/google/uuid"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


pkg/skaffold/runner/runcontext/context.go
```diff
@@ -208,10 +209,14 @@ func GetRunContext(opts config.SkaffoldOptions, configs []schemaUtil.VersionedCo
...
-	if err != nil {
-		return nil, fmt.Errorf("finding current directory: %w", err)
+	var cwd string
+	if opts.ConfigurationFile == "" {
+		cwd, err = os.Getwd()
+		if err != nil {
+			return nil, fmt.Errorf("finding current directory: %w", err)
+		}
+	} else {
+		cwd = filepath.Dir(opts.ConfigurationFile)
```
gsquared94, line 219: I tried to get this change through once before but couldn't because the Cloudcode IDEs have made this assumption in several places.

gsquared94: go/skaffold-run-from-here is the correct alternative.

yuwenma: 🤦‍♀️ 🤦‍♀️ 🤦‍♀️

yuwenma:
See more context here https://docs.google.com/document/d/16vo43mEywdV47Dx_awwfA7hXCGi7J0b9G78srxsfQhw/edit?ts=60f70250&resourcekey=0-PLFsBoJG5-dzKCzrXhyK7g

briandealwis:
See #6060 which proposes the least-bad solution identified.

PR #6272, opened 2021-07-22
Add state to event handler to track `logrus` output events

MarlonGamez:
**Related**: #5368 

**Description**
This PR adds a `task` field to the `eventHandler` type to track what the current phase of skaffold is. This is used to populate the `SkaffoldLogEvent`s that come from `logrus` output.

### DIFF
--- a/pkg/skaffold/event/v2/event.go
+++ b/pkg/skaffold/event/v2/event.go
@@ -52,6 +52,7 @@ var handler = newHandler()
 func newHandler() *eventHandler {
 	h := &eventHandler{
 		eventChan: make(chan *proto.Event),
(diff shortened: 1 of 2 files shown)


nkubala (approved):

pkg/skaffold/event/v2/event.go
```diff
@@ -319,6 +321,7 @@ func TaskInProgress(task constants.Phase, description string) {
 		handler.skaffoldLogs = []proto.Event{}
 	}
 
+	handler.task = task
```
gsquared94, line 324: these assignments will run into race condition

Issue #6254, opened 2021-07-20
invalid skaffold config with v1.27.0


MarcoSchaar:
Hi 
I'm facing some issues while trying to run existing builds with updated version v1.27.0

### Expected behavior
successful operation like with v1.26.0

### Actual behavior
```
$ skaffold run --profile build-app '--filename=./skaffold.yaml'
invalid skaffold config: skaffold config named "/home/jenkins/agent/workspace/<folder>/skaffold.yaml" referenced file "k8s/*.yaml" that could not be found. Verify that file "k8s/*.yaml" referenced in config "/home/jenkins/agent/workspace/<folder>/skaffold.yaml" exists and the path and naming are correct.
```
### Information

- Skaffold version: 1.27.0
- Operating system: alpine 3.14
- Installed via: skaffold.dev
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta18
kind: Config
...
profiles:
- name: build-app
  build:
    artifacts:
    - image: <image>
      context: <folder>
      kaniko:
        dockerfile: Dockerfile 
        image: gcr.io/kaniko-project/executor:debug
        volumeMounts:
        - name: kaniko-secret
          readonly: false
          mountpath: /kaniko/.docker
          subpath: ""
          mountpropagation: null
          subpathexpr: ""
        - name: kaniko-secret
          readonly: false
          mountpath: /kaniko/ssl/certs/<cert>
          subpath: <cert>
          mountpropagation: null
          subpathexpr: ""
    tagPolicy:
      envTemplate:
        template: '{{.APP_VERSION}}'
    cluster:
      pullSecretName: kaniko-secret
      namespace: jenkins
```

### Steps to reproduce the behavior

1. a clonable repository with the sample skaffold project
2. `skaffold run --profile <profile>`
3. ...
```
ska$ ffold run --profile build-app '--filename=./skaffold.yaml' -v debug
level=info msg="Skaffold &{Version:v1.27.0 ConfigVersion:skaffold/v2beta18 GitVersion: GitCommit:1f46f249c832bb2a99e3285ad327647c95ff4bb9 BuildDate:2021-06-29T21:37:08Z GoVersion:go1.14.14 Compiler:gc Platform:linux/amd64 User:}"
level=info msg="Loaded Skaffold defaults from \"/home/build/.skaffold/config\""
level=debug msg="parsed 1 configs from configuration file /home/jenkins/agent/workspace/<folder>/skaffold.yaml"
[...]

PR #6294, opened 2021-07-26
skip manifest validation for default `kubectl deployer`

gsquared94:
Fixes: #6254 

**Description**

In #6043 we introduced validation for the `kubectl` deployer manifests. This breaks backward compatibility for users who were using `skaffold run` without any deployer defined. Although that doesn't seem like a valid usecase for the `run` command, for versions <= v1.26.0, `skaffold run` command worked the same as `skaffold build` command in the absense of a `deployer` stanza in the skaffold config. 

The default deployer (`./k8s/*`) gets added when no explicit deployer is defined, which now fails the validation introduced in #6043.

To keep backward compat behavior, this PR skips the manifest validations when only the default kubectl deployer is defined.

### DIFF
--- a/pkg/skaffold/schema/validation/validation.go
+++ b/pkg/skaffold/schema/validation/validation.go
@@ -27,9 +27,11 @@ import (
 
 	"github.com/docker/docker/api/types"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


aaron-prindle (approved):
LGTM

PR #6311, opened 2021-07-27
Pin `go-licenses` version to avoid go 1.16.x dependencies

MarlonGamez:
**Description**
CI is failing because of `go-licenses`'s dependency on go 1.16.x packages. This PR fixes that by pinning to a previous version without those dependencies.


briandealwis (approved):
Can you add a TODO to remove this once we fix up our builds?

PR #6326, opened 2021-07-28
Set up `logrus.Logger` in `output.WithEventContext`, improve usage of `output.GetWriter()`

MarlonGamez:
**Related**: #5368 

**Description**
This PR changes adds setup of `logrus.Logger`s to `WithEventContext()` function. It modifies the `output.skaffoldWriter` type to contain information about the verbosity level, which is then used when creating the loggers that are returned. Currently, these returned loggers aren't being used just yet. I'll be opening a follow up PR to utilize these in functions that make calls to `logrus` functions.

Plump `logrus.Logger`s returned from `WithEventContext()` into functions that will be making `logrus` calls.

### DIFF
--- a/cmd/skaffold/app/cmd/cmd.go
+++ b/cmd/skaffold/app/cmd/cmd.go
@@ -85,14 +85,15 @@ func NewSkaffoldCommand(out, errOut io.Writer) *cobra.Command {
 			// These are used for command completion and send debug messages on stderr.
 			if cmd.Name() != cobra.ShellCompRequestCmd && cmd.Name() != cobra.ShellCompNoDescRequestCmd {
 				instrumentation.SetCommand(cmd.Name())
-				lvl, _ := logrus.ParseLevel(v)
+				lvl, err := logrus.ParseLevel(v)
+				if err != nil {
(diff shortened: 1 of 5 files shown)


cmd/skaffold/app/cmd/cmd.go
```diff
@@ -85,13 +85,15 @@ func NewSkaffoldCommand(out, errOut io.Writer) *cobra.Command {
 			// These are used for command completion and send debug messages on stderr.
 			if cmd.Name() != cobra.ShellCompRequestCmd && cmd.Name() != cobra.ShellCompNoDescRequestCmd {
 				instrumentation.SetCommand(cmd.Name())
-				out := output.GetWriter(out, defaultColor, forceColors, timestamps)
+				lvl, err := logrus.ParseLevel(v)
+				if err != nil {
+					return fmt.Errorf("parsing log level: %w", err)
+				}
+				out := output.GetWriter(out, defaultColor, forceColors, timestamps, lvl)
```
gsquared94, line 92: I get confused across `skaffoldWriter`, `colorableWriter`, `io.Writer` and `logrus`. 😅

PR #6329, opened 2021-07-28
Plumb logrus loggers into Deploy and Render phases

MarlonGamez:
**Related**: #5368 
**Merge After**: https://github.com/GoogleContainerTools/skaffold/pull/6326

**Description**
This PR plumbs `*logrus.Logger` type into deploy phases of skaffold in order to allow `logrus.` calls to instead used a passed in logger that will have event context for the event api.


pkg/skaffold/deploy/docker/deploy.go
```diff
@@ -92,7 +92,7 @@ func (d *Deployer) TrackContainerFromBuild(build graph.Artifact, id string) {
 
 // Deploy deploys built artifacts by creating containers in the local docker daemon
 // from each artifact's image.
-func (d *Deployer) Deploy(ctx context.Context, out io.Writer, builds []graph.Artifact) error {
+func (d *Deployer) Deploy(ctx context.Context, out io.Writer, log *logrus.Logger, builds []graph.Artifact) error {
```
gsquared94, line 95: I think we should maintain a static `logrus.Logger` reference instead of passing it around.

gsquared94: https://github.com/sirupsen/logrus/issues/391 seems to suggest that multiple `*logrus.Logger` instances aren't concurrency-safe.

MarlonGamez: you're right, I think I need to keep searching for a better solution to this problem. I'm going to close this for now

Issue #6316, opened 2021-07-28
Skaffold 1.28.1 doesn't override kubecontext via SKAFFOLD_KUBE_CONTEXT


artemkoru:
### Expected behavior

When I set the SKAFFOLD_KUBE_CONTEXT environment variable, I expect that the value `deploy.kubeContext` in skaffold.yaml to be ignored and skaffold will deploy my application to the kubecontext that is set in the SKAFFOLD_KUBE_CONTEXT variable. This works fine in versions <= 1.26.0.

### Actual behavior

Skaffold ignores SKAFFOLD_KUBE_CONTEXT environment variable and uses the value `deploy.kubeContext` from skaffold.yaml.

### Information

- Skaffold version: 1.28.1, 1.27.0
- Operating system: Darwin 20.5.0
- Installed via:
curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/v1.28.1/skaffold-darwin-amd64 && \
sudo install skaffold /usr/local/bin/

- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta7
kind: Config
metadata:
  name: skaffod-kubecontext-mismatch
deploy:
  kubeContext: kubecontext-1
  helm:
    flags:
      install:
      - "--timeout=5m"
      upgrade:
      - "--timeout=5m"
    releases:
      - name: "skaffod-kubecontext-mismatch"
        chartPath: charts/web-app
        wait: true
        setValueTemplates:
          image.name: nginx
        valuesFiles:
          - helm/values.yaml
```

### Steps to reproduce the behavior

1. Clone git repo: https://github.com/artemkoru/skaffod-kubecontext-mismatch
2. Run in console: `export SKAFFOLD_KUBE_CONTEXT=kubecontext-2;skaffold deploy`
3. Make sure skaffold deployed your app in the wrong kubecontext (kubecontext-1)

Logs:
```bash
➜  skaffod-kubecontext-mismatch skaffold deploy -vdebug                                                                                          
INFO[0000] Activated kube-context "kubecontext-2"          
INFO[0000] Skaffold &{Version:v1.28.1 ConfigVersion:skaffold/v2beta19 GitVersion: GitCommit:7b855e136dc0f9cc5544ed2808c9d5feb767118c BuildDate:2021-07-20T17:51:49Z GoVersion:go1.14.14 Compiler:gc Platform:darwin/amd64 User:} 
[...]

PR #6331, opened 2021-07-29
fix: kubeContext override via flag

gsquared94:
Fixes: #6316 
#6024 made some fixes around setting `kubeContext` but reversed the order of preference between `kubeContext` set via flag and set via deploy config

**Description**
This PR ensures the following order:
- flag or env variable overrides the current set kubeContext and also the deploy config kubeContext
- deploy config kubeContext overrides the current set kubeContext
- if neither flag, env var nor config define kubeContext then the current set value is used.

### DIFF
--- a/pkg/skaffold/runner/deployer.go
+++ b/pkg/skaffold/runner/deployer.go
@@ -41,7 +41,8 @@ type deployerCtx struct {
 }
 
 func (d *deployerCtx) GetKubeContext() string {
-	if d.deploy.KubeContext != "" {
+	// if the kubeContext is not overridden by CLI flag or env. variable then use the value provided in config.
+	if d.RunContext.IsDefaultKubeContext() && d.deploy.KubeContext != "" {
 		return d.deploy.KubeContext
 	}
 	return d.RunContext.GetKubeContext()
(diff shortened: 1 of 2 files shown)


pkg/skaffold/runner/deployer.go
```diff
@@ -41,7 +41,8 @@ type deployerCtx struct {
 }
 
 func (d *deployerCtx) GetKubeContext() string {
-	if d.deploy.KubeContext != "" {
+	// if the kubeContext is not overridden by CLI flag or env. variable then use the value provided in config.
```
aaron-prindle, line 44: nit:  It might make sense to add some E2E/integration test that can verify that the ordering is preserved going forward

aaron-prindle (approved):
LGTM

aaron-prindle:
nit: It might make sense to add some E2E/integration test that can verify that the ordering is preserved going forward

gsquared94:
> nit: It might make sense to add some E2E/integration test that can verify that the ordering is preserved going forward

I was gonna add UT, then absentmindedly merged the approved PR. I'll raise another PR with tests.

Issue #6312, opened 2021-07-27
`dev --port-forward` with user-defined ports may result in services being allocated random ports


briandealwis:
Seen in Skaffold HEAD, but present back to v1.24.0.

@dhodun has a multi-module app with two services configured by user-defined port-forwards (UDPFs) with specific local ports.  The UDPFs look like:
```
portForward:
- resourceType: service
  resourceName: rules-engine
  port: 80
  localPort: 8887
```

He reported an odd situation where `skaffold dev --port-forward` with would frequently set one of the services that should be bound to port 8887 was sometimes being bound to port 4503.  As I write this, I realized there have been some other user reports of this nature.

In digging into the issue, the problem is that `skaffold dev --port-forward` creates multiple forwarders to handle the various types of forwarders (services, user ports, pods, etc): https://github.com/GoogleContainerTools/skaffold/blob/55f5860137ef74c8e7d935835a2dd59b98a6f63d/pkg/skaffold/kubernetes/portforward/forwarder_manager.go#L63-L91
and the [UDPF and services forwarders are different instances of  `ResourceForwarder`](https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/kubernetes/portforward/resource_forwarder.go#L50-L65).  The services forwarders retrieves the list of services and spoofs up a port-forwarding resource but with localPort=0 so that a random'ish port is allocated.

The entries-manager attempts to coalesce port-forward requests by mapping requests for the same resource+port to the same entry.  So a UDPF and a services forward for the same service will map to the same service.

Each of these forwarders is started in turn, but the `ResourceForwarder`s actually establish the individual port-forwards in a goroutine:
https://github.com/GoogleContainerTools/skaffold/blob/55f5860137ef74c8e7d935835a2dd59b98a6f63d/pkg/skaffold/kubernetes/portforward/resource_forwarder.go#L128-L134

The underlying forwarding mechanism, the `KubectlForwarder`, uses a goroutine to launch `kubectl port-forward` and then blocks on a channel to retrieve the launch result:
https://github.com/GoogleContainerTools/skaffold/blob/55f5860137ef74c8e7d935835a2dd59b98a6f63d/pkg/skaffold/kubernetes/portforward/kubectl_forwarder.go#L81-L85

So it's possible that the UDPF-forwards (with a specific local port) become blocked while the services-forwards (with local-port = 0) go forward.    And if the services-forwards happens first, then the real entry will have a random'ish port allocated.

[...]

Issue #5763, opened 2021-04-30
skaffold portForward not working with multiple ports


ashraf-revo:
iam getting this logs   
```
Port forwarding Service/streamer in namespace default, remote port 80 -> 127.0.0.1:8082
Port forwarding service/rtsp in namespace default, remote port 554 -> 127.0.0.1:4503
Port forwarding pod/streamer-84c6b9567d-l8qqt in namespace default, remote port 5005 -> 127.0.0.1:5005
```

### Expected behavior
i expect service/rtsp to run on port 8083

### Actual behavior
but i found it run on another port 4503   this bug only happen when iam using intellj idea cloud code plugin
### Information

- Skaffold version:  run `v1.16.0` 
- Operating system:  ubuntu 
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta7
kind: Config
metadata:
  name: streamer
build:
  artifacts:
  - image: ashraf1abdelrasool/streamer
    context: .
    jib:
      project: org.revo:streamer
deploy:
  kubectl:
    manifests:
    - target/jkube/*.yml
portForward:
  - port: 80
    localPort: 8082
    resourceType: Service
    resourceName: streamer
  - port: 554
    localPort: 8083
    resourceType: Service
    resourceName: rtsp

```

MarlonGamez:
hi @ashraf-revo, thanks for opening this issue. Just to clarify, you don't run into this issue when simply running skaffold through the command line?

ashraf-revo:
When run skaffold dev it work fine the problem with intelj idea only , it only map the first port in the portforward list

briandealwis:
@ashraf-revo is it possible there is something listening on port 8083?  Does it work f you change your `skaffold.yaml` to use a different port?

ashraf-revo:
Ot only map the first port , so of I switch the order of the array of portforward to start with the 554:8083 it will work but the other port will not work , so it only map the first port

PR #6332, opened 2021-07-29
ResourceForwarder should wait for its forwards

briandealwis:
Fixes: #6312
Fixes: #5763

**Description**
The `ForwarderManager` carefully [sets up a sequence of forwarders](https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/kubernetes/portforward/forwarder_manager.go#L71-L82), such that user-defined port-forwards are handled before any other forwards, and then [starts each in turn](https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/kubernetes/portforward/forwarder_manager.go#L145-L146).  But the `ResourceForwarder.Start()` actually [processed its forwards in a goroutine](https://github.com/GoogleContainerTools/skaffold/blob/27758a9a7b2e56d9c797bbee0790ce6db45a43ae/pkg/skaffold/kubernetes/portforward/resource_forwarder.go#L128-L134).  As a result, the `ResourceForwarder` processing user-defined port-forwards could execute out-of-order with the `ResourceForwarder` for services.  So a user-defined port-forward for a service could actually be processed after a services-derived port-forward such that the allocated port is different than the requested port.

This PR changes the `ResourceForwarder` startup to process its forwards in parallel but waits for their completion. This ensures the forwarders are processed in order and any user-defined `portForwards` have their `localPort` respected (if possible).


pkg/skaffold/kubernetes/portforward/resource_forwarder.go
```diff
@@ -126,11 +127,15 @@ func (p *ResourceForwarder) Stop() {
 
 // Port forward each resource individually in a goroutine
 func (p *ResourceForwarder) portForwardResources(ctx context.Context, resources []*latestV1.PortForwardResource) {
-	go func() {
-		for _, r := range resources {
-			p.portForwardResource(ctx, *r)
-		}
-	}()
+	var wg sync.WaitGroup
```
gsquared94, line 130: this change LGTM. The code in `main` seems like a bug since the comment says `Port forward each resource individually in a goroutine` but it isn't doing that.
I took a quick look to see if port forwarder can work with concurrent requests, this block seems a bit fishy:
https://github.com/GoogleContainerTools/skaffold/blob/2d36770ca26b9c3175b1cbd165816dfa7db22a58/pkg/skaffold/kubernetes/portforward/entry_manager.go#L120-L125

It doesn't lock properly since the `Load` and `Store` are done separately

briandealwis: Good point.

briandealwis: Rewrote to just use a sync.Map.

gsquared94 (approved):

PR #6338, opened 2021-08-02
 [v2] Fix un-registered render manifest dependencies.

yuwenma:
**Related**: #5673 

**Description**
 Root cause: The render manfiests are registered to the filewatcher before "Render" func is called. However, the renderer has its dependencies calculated and refreshed in each "Render" call.
    
 This PR separates the "Render" and "manfiestDependencies" completely.


yuwenma:
To the wrong branch.

PR #6356, opened 2021-08-03
Add Log helper function to output package

MarlonGamez:
**Related**: #5368 

**Description**
This PR adds a `Log()` function to wrap around `logrus.WithFields()`. This function takes an `io.Writer` and extracts information related to `task` and `subtask` for the event stream. We can use this for logging in the future, to properly set fields when sending out `SkaffoldLogEvent`s

### DIFF
--- a/examples/microservices/skaffold.yaml
+++ b/examples/microservices/skaffold.yaml
@@ -32,3 +32,4 @@ profiles:
   - name: gcb
     build:
       googleCloudBuild: {}
+# hello there
--- a/examples/microservices/skaffold.yaml
+++ b/examples/microservices/skaffold.yaml
@@ -32,4 +32,3 @@ profiles:
   - name: gcb
     build:
       googleCloudBuild: {}
-# hello there
--- a/pkg/skaffold/output/output.go
+++ b/pkg/skaffold/output/output.go
@@ -32,6 +32,8 @@ const TimestampFormat = "2006-01-02 15:04:05"
 type skaffoldWriter struct {
 	MainWriter  io.Writer
 	EventWriter io.Writer
+	task        constants.Phase
+	subtask     string
 
 	timestamps bool
 }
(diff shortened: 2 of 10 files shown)


examples/microservices/skaffold.yaml
```diff
@@ -32,3 +32,4 @@ profiles:
   - name: gcb
     build:
       googleCloudBuild: {}
+# hello there
```
aaron-prindle, line 35: nit: remove

pkg/skaffold/output/output_test.go
```diff
@@ -217,3 +217,36 @@ func TestWriteWithTimeStamps(t *testing.T) {
...
+				task:    constants.Build,
+				subtask: "test",
+			},
+			expectedTask:    constants.Build,
+			expectedSubtask: "test",
+		},
+		{
+			name:            "non skaffoldWriter",
+			writer:          ioutil.Discard,
+			expectedTask:    constants.DevLoop,
```
aaron-prindle, line 240: Why is the expected task for `non skaffold writer` - `constants.DevLoop`?

aaron-prindle: Ah I see now

aaron-prindle (approved):
LGTM.  I believe that there were some accidental changes made to `examples/microservices/skaffold.yaml` that should be cleanup up but that is it

pkg/skaffold/output/output.go
```diff
@@ -101,14 +103,33 @@ func GetUnderlyingWriter(out io.Writer) io.Writer {
...
+	if isSW {
+		return logrus.WithFields(logrus.Fields{
+			"task":    sw.task,
+			"subtask": sw.subtask,
+		})
 	}
 
-	return out, nil
+	return logrus.WithFields(logrus.Fields{
+		"task":    constants.DevLoop,
```
aaron-prindle, line 132: nit: Perhaps add a comment as to why `task: constants.DevLoop` makes sense here

tejal29 (approved):

examples/microservices/skaffold.yaml
MarlonGamez: whoops 🤦‍♂️

pkg/skaffold/output/output.go
MarlonGamez: done 👍🏼

PR #6365, opened 2021-08-05
Add eventV2.BuildCancelled

### LINKED ISSUE #6307: Clearer indication of skipped builds in logging V2 API
When building multiple containers, Skaffold will abort builds when one container fails.

The V2 API currently outputs build sub-task log event like the following for the aborted container:

"Building [container-name]"...

Which leads to an exeperience like this in cloud code:
image

This makes it difficult to understand what happened. As discussed in chat, we propose that Skaffold output an additional log event for this containers of the form (rough wording):

> Container [aborted-container-name] cancelled due to failure building [failed-build-container-name]

tejal29:
Fixes #6307

Add build cancelled if any dependent artifact build fails.

TODO:
- [x] Add build cancelled when dependent artifacts don't fail but other build fails.

### DIFF
--- a/pkg/skaffold/build/scheduler.go
+++ b/pkg/skaffold/build/scheduler.go
@@ -86,7 +86,7 @@ func (s *scheduler) build(ctx context.Context, tags tag.ImageTags, i int) error 
 	if err != nil {
 		// `waitForDependencies` only returns `context.Canceled` error
 		event.BuildCanceled(a.ImageName)
-		eventV2.BuildCanceled(a.ImageName, fmt.Errorf("one or more artifact builds failed"))
+		eventV2.BuildCanceled(a.ImageName, err)
 		return err
 	}
 	release := s.concurrencySem.acquire()
@@ -113,7 +113,11 @@ func (s *scheduler) build(ctx context.Context, tags tag.ImageTags, i int) error 
 	finalTag, err := performBuild(ctx, w, tags, a, s.artifactBuilder)
 	if err != nil {
 		event.BuildFailed(a.ImageName, err)
-		eventV2.BuildFailed(a.ImageName, err)
+		if ctx.Err() == context.Canceled {
+			eventV2.BuildCanceled(a.ImageName, err)
+		} else {
+			eventV2.BuildFailed(a.ImageName, err)
+		}
 		endTrace(instrumentation.TraceEndError(err))
 		return err
 	}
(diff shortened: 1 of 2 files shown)


tejal29:
@gsquared94 Needed some help on this. 

We want to send a `BuildCancelled` event to IDEs (see  #6307) when a build is cancelled.

I found the code where a dependent artifact build gets cancelled. 
ie. if `base` fails, then `frontend` artifact build will be cancelled

If two builds are running in parallel e.g. `frontend` and `backend`. 
1) If `frontend` fails first, and `backend` build is in progress, it should get cancelled. 
Where is the code where this `backend` build context is cancelled and I can send a `eventV2.BuildCancelled`?

MarlonGamez (approved):
LGTM, thanks @tejal29

gsquared94:
> If two builds are running in parallel e.g. `frontend` and `backend`.
> 
>     1. If `frontend` fails first, and `backend` build is in progress, it should get cancelled.
>        Where can i find the code where the `backend` build context is cancelled and I can send a `eventV2.BuildCancelled`?

All artifact builds (parallel or sequential) share the same `context.Context` created with `errgroup.WithContext` which "is canceled the first time a function passed to Go returns a non-nil error or the first time Wait returns, whichever occurs first."
So you should be able to check if the context is cancelled [here](https://github.com/GoogleContainerTools/skaffold/blob/ea599399ca433892ef2c8b1cd48b313a187699b3/pkg/skaffold/build/scheduler.go#L113) and log a `eventV2.BuildCancelled` conditionally instead of the `eventV2.BuildFailed`

tejal29:
Can't review my own PR, but LGTM!

pkg/skaffold/build/scheduler.go
```diff
@@ -112,7 +113,12 @@ func (s *scheduler) build(ctx context.Context, tags tag.ImageTags, i int) error
 	finalTag, err := performBuild(ctx, w, tags, a, s.artifactBuilder)
 	if err != nil {
 		event.BuildFailed(a.ImageName, err)
-		eventV2.BuildFailed(a.ImageName, err)
+		if ctx.Err() == context.Canceled {
```
gsquared94, line 116: use `errors.Is(ctx.Err(), context.Canceled)` instead

MarlonGamez: oops, thanks for pointing this out

MarlonGamez (approved):

Issue #6134, opened 2021-07-05
`build --push=false` may still try to push


briandealwis:
With #5708, we added a new `--push` flag to override the `build.local.push` value set in the `skaffold.yaml`.  But `--push=false` still seems to push when I don't have a `~/.kube/config`:
```
$ mv ~/.kube/config ~/.kube/config.old
$ cd examples/getting-started
$ skaffold build --push=false --cache-artifacts=false -vdebug
INFO[0000] Skaffold &{Version:v1.26.0 ConfigVersion:skaffold/v2beta17 GitVersion: GitCommit:851bcab1a12274966a846cbb95b0cf364f45531c BuildDate:2021-06-09T02:00:32Z GoVersion:go1.16beta1 Compiler:gc Platform:darwin/arm64 User:} 
INFO[0000] Loaded Skaffold defaults from "/Users/bdealwis/.skaffold/config" 
DEBU[0000] parsed 1 configs from configuration file /Users/bdealwis/Projects/Skaffold/repo-skaffold/examples/getting-started/skaffold.yaml 
DEBU[0000] Defaulting build type to local build         
INFO[0000] Using kubectl context:                       
DEBU[0000] Running command: [minikube version --output=json] 
DEBU[0000] Command output: [{"commit":"c61663e942ec43b20e8e70839dcca52e44cd85ae","minikubeVersion":"v1.20.0"}
] 
DEBU[0000] setting Docker user agent to skaffold-v1.26.0 
DEBU[0000] Using builder: local                         
DEBU[0000] push value not present in NewBuilder, defaulting to true because cluster.PushImages is true 
INFO[0000] build concurrency first set to 1 parsed from *local.Builder[0] 
...
Starting build...
Building [skaffold-example]...
...
Successfully built 59a7923c0b74
Successfully tagged skaffold-example:v1.27.0-10-gd8c2374c1-dirty
The push refers to repository [docker.io/library/skaffold-example]
cef509a10e7a: Preparing
5e04d10b60a4: Preparing
DEBU[0002] Running command: [tput colors]               
DEBU[0002] Command output: [256
]                       
denied: requested access to the resource is denied
```

It looks like `builder.NewBuilder()` needs some additional configuration:
https://github.com/GoogleContainerTools/skaffold/blob/823896bdd7464d79886664b32f56948957ea5f65/pkg/skaffold/build/local/types.go#L89-L95

PR #6366, opened 2021-08-05
fix build --push=false for missing kubeconfig

verbanicm:
Fixes: #6134 

**Description**
- Map push flag from run context to build context for local builder config
- Push flag from CLI takes precedence over other push config
- Otherwise existing logic prevails (use build config file, if nil, use cluster push flag)


pkg/skaffold/build/local/types.go
```diff
@@ -85,12 +86,17 @@ func NewBuilder(bCtx BuilderContext, buildCfg *latestV1.LocalBuild) (*Builder, e
...
 
 	cluster := bCtx.GetCluster()
+	pushFlag := bCtx.PushImages()
 
 	var pushImages bool
-	if buildCfg.Push == nil {
+	switch {
+	case pushFlag.Value() != nil:
+		logrus.Debugf("push value set via skaffold build --push flag, --push=%t", *pushFlag.Value())
+		pushImages = *pushFlag.Value()
```
MarlonGamez, line 95: tiniest nit: keep ordering of log message and `pushImages` assignment consistent with the `case` below this one

MarlonGamez (approved):
Looks good to me!

pkg/skaffold/build/local/types.go
verbanicm: Fixed

PR #6367, opened 2021-08-05
[v2] Add builtin kpt (Go1.16)

yuwenma:
**Related**: #5673 

**Description**
Some skaffold CMDs (e.g. `render`, `dev`, `run`) requires kpt CLI to hydrate the manifests. In case users do not have `kpt` or the right `kpt` version installed, we embed the kpt executable in the skaffold CLI, and installs for users if and only if a kpt executable is required but not found. This gives current skaffold users a smoother experience when switching to v2 

The matching kpt OS and Arch will be decided in skaffold build (`make install`) and copied to pkg/skaffold/render/embed/kpt before building skaffold.   

This builtin kpt uses the Go 1.16 "embed" feature    

PR is tested manually. No unittests (since it requires different OS and ARC), integration tests are needed.

### DIFF
--- a/pkg/skaffold/render/embed/embed.go
+++ b/pkg/skaffold/render/embed/embed.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2021 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package embed
+
+import (
+	_ "embed" //nolint https://github.com/golangci/golangci-lint/issues/1727
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+//go:embed kpt
+var kpt []byte
+
+// UseBuiltinKpt guarantees the `kpt` always exists in $PATH/bin. `kpt` is needed in skaffold render.
+func UseBuiltinKpt() error {
+	err := os.WriteFile(getKptInstallPath(), kpt, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("unable to add builtin kpt")
+	}
+	return nil
+}
+
+func getKptInstallPath() string {
+	paths := filepath.SplitList(os.Getenv("PATH"))
+	dir := paths[0]
+	if dir == "" {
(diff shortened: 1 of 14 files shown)


briandealwis:
Discussed with @yuwenma separately, but embedding kpt (a ~50mb) binary is not really a workable solution.  We have a similar request around `kompose` for `skaffold init --compose-file`.

We do include a reasonably up-to-date kpt in the `gcr.io/k8s-skaffold/skaffold` image.

pkg/skaffold/render/embed/embed.go
```diff
@@ -0,0 +1,45 @@
...
+func UseBuiltinKpt() error {
+	err := os.WriteFile(getKptInstallPath(), kpt, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("unable to add builtin kpt")
+	}
+	return nil
+}
+
+func getKptInstallPath() string {
+	paths := filepath.SplitList(os.Getenv("PATH"))
```
ahmetb, line 38: as a user, I'd prefer skaffold not to install something to PATH[0] on my computer. PATH[0] has no significance as far as I know. on my machine it points to `/Users/ahmetb/.krew/bin` :) and most devs will make similar edits to PATH.

pkg/skaffold/render/embed/embed.go
```diff
@@ -0,0 +1,45 @@
...
+	"os"
+	"path/filepath"
+)
+
+//go:embed kpt
+var kpt []byte
+
+// UseBuiltinKpt guarantees the `kpt` always exists in $PATH/bin. `kpt` is needed in skaffold render.
+func UseBuiltinKpt() error {
+	err := os.WriteFile(getKptInstallPath(), kpt, os.ModePerm)
```
ahmetb, line 30: would it be better to

1. create tmpdir
2. write kpt there + chmod
3. update running process's env via os.Setenv and add tmpdir
4. now the os/exec should be running kpt via tmpdir?

aaron-prindle:
Currently this is blocked on - https://github.com/GoogleContainerTools/skaffold/issues/6790. @briandealwis do we have plans on prioritizing updating skaffold's CI & release process to use go1.16 in an upcoming milestone?

briandealwis:
Turning Skaffold from a 50MB &rarr; 100MB+ binary is not a workable solution and needs to be rethought.  We can have a skaffold+kpt .tar.gz/.zip bundle.

SSGoku369 (commented):
ok

SSGoku369:
ok

SSGoku369 (commented):
ok

SSGoku369:
ok

SSGoku369 (commented):
ok

SSGoku369 (commented):
ok

tejal29:
closing this PR. Lets address this once we decide to distribute kpt via skaffold.

Issue #6336, opened 2021-07-31
logs are not printed with --tail when kubeContext is set and its different from current-context 


weltonrodrigo:
### Expected behavior

Running ```skaffold run --tail``` with `deploy.kubeContext` set should print logs after deployment stabilization. 

### Actual behavior
No log is printed.

```shell
skaffold run --tail -v debug
```

```
Starting deploy...
DEBU[0008] getting client config for kubeContext: ``    
DEBU[0009] Running command: [kubectl version --client -ojson] 
DEBU[0009] Command output: [{
  "clientVersion": {
    "major": "1",
    "minor": "21",
    "gitVersion": "v1.21.2",
    "gitCommit": "092fbfbf53427de67cac1e9fa54aaa09a28371d7",
    "gitTreeState": "clean",
    "buildDate": "2021-06-16T12:52:14Z",
    "goVersion": "go1.16.5",
    "compiler": "gc",
    "platform": "darwin/amd64"
  }
}
] 
DEBU[0009] Running command: [kustomize build kubernetes] 
DEBU[0010] Command output: [REDACTED] 
DEBU[0010] manifests with tagged images: apiVersion: v1
[REDACTED]
DEBU[0010] manifests with labels apiVersion: v1
  [REDACTED]
DEBU[0010] Running command: [kubectl --context foobar --namespace my_namespace get -f - --ignore-not-found -ojson] 
DEBU[0010] Command output: [REDACTED]
DEBU[0010] 2 manifests to deploy. 2 are updated or new  
DEBU[0010] Running command: [kubectl --context foobar --namespace my_namespace apply -f -] 
 - configmap/ajeitadin-config-d8k4696hch configured
 - deployment.apps/nifi-ajeitadin configured
INFO[0011] Deploy completed in 2.993 seconds            
Waiting for deployments to stabilize...
DEBU[0011] getting client config for kubeContext: ``    
Deployments stabilized in 95.335945ms
DEBU[0011] getting client config for kubeContext: ``    
Press Ctrl+C to exit
```
### Information

- Skaffold version:  v1.28.1
- Operating system: Macos 11.4
- Installed via: homebrew
- Contents of skaffold.yaml: v1.28.1

```yaml
apiVersion: skaffold/v2beta16
kind: Config
metadata:
  name: redacted
build:
  artifacts:
  - image: image:tag
    docker:
      dockerfile: Dockerfile
deploy:
  kubeContext: foobar
  kustomize:
    paths:
[...]

Issue #6078, opened 2021-06-25
Specifying a `kubeContext` in the skaffold config doesn't work if current context is set to a local cluster


gsquared94:
Using this modified `skaffold.yaml` from the `<repo>/examples/getting-started` project:

```yaml
apiVersion: skaffold/v2beta18
kind: Config
build:
  artifacts:
  - image: skaffold-example
deploy:
  kubeContext: gke_gaghosh-project-1_us-central1-c_fc1
  kubectl:
    manifests:
      - k8s-*
```

with my current kubecontext set to a local `kind` cluster.

### Expected behavior
`skaffold dev` builds the image, pushes it to the remote registry and then deploys to the remote cluster `gke_gaghosh-project-1_us-central1-c_fc1
` specified in the `skaffold.yaml` file.
### Actual behavior
```
❯ skaffold dev
Listing files to watch...
 - skaffold-example
Generating tags...
 - skaffold-example -> skaffold-example:v1.26.0-41-gc326595dd-dirty
Checking cache...
 - skaffold-example: Found. Tagging
Starting test...
Tags used in deployment:
 - skaffold-example -> skaffold-example:a9216dc058bacea3d905f96574b2247114fd08f32dadf8e412f270d302e00de0
Starting deploy...
 - pod/getting-started created
Waiting for deployments to stabilize...
Deployments stabilized in 1.102 second
Press Ctrl+C to exit
Watching for changes...
rpc error: code = Unknown desc = failed to pull and unpack image "docker.io/library/skaffold-example:a9216dc058bacea3d905f96574b2247114fd08f32dadf8e412f270d302e00de0": failed to resolve reference "docker.io/library/skaffold-example:a9216dc058bacea3d905f96574b2247114fd08f32dadf8e412f270d302e00de0": pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```
### Information
If the `build.push` setting is not specified then skaffold tries to guess if it should push the built images based on the locally set kubecontext (or that provided by `--kubecontext` flag), even though the deploy definition in the skaffold config can specify a remote cluster.

PR #6368, opened 2021-08-05
fix: make `client.Client` kubernetesContext aware

gsquared94:
Fixes: #6336, #6078 

**Description**

This PR mainly changes the signature for functions `client.Client` and `client.DynamicClient` to require a `kubeContext`. The previous behavior of just using the statically set `kubeContext` is fulfilled where required using a new `client.DefaultClient` function.

All other changes are from needing to plumb the correct `kubeContext` across. This is done by either adding a function or method parameter or struct field.

This change should finally fix all scenarios overriding the kubecontext in the `deploy.kubeContext` field in skaffold config.


MarlonGamez (approved):
looks good to me 👍🏼

PR #6370, opened 2021-08-05
fix status check showing unhealthy pods from prev iteration

tejal29:
fixes https://github.com/GoogleCloudPlatform/cloud-code-intellij-internal/issues/4323

In this PR

1) Use owner reference to query pods for a deployment. 
A `Deployment` manages `ReplicaSets`.  When a deployment is updates, a new `ReplicaSet` object is created. 
A `ReplicaSet`'s purpose is to maintain a stable set of replica Pods running at any given time.
Pods are controlled/owned by `ReplicaSet`

When a deployment is updated, a new replicaset is created which controls pods.
This is evident by running `kubectl describe` 
<details> 

```
➜  go-guestbook git:(master) ✗ kubectl describe deployment/go-guestbook-frontend
Name:                   go-guestbook-frontend
...
OldReplicaSets:  <none>
NewReplicaSet:   go-guestbook-frontend-869d5b5fb4 (1/1 replicas created)

```

Make a change to frontend code 
```
➜  go-guestbook git:(master) ✗ kubectl describe deployment/go-guestbook-frontend
...
OldReplicaSets:  <none>
NewReplicaSet:   go-guestbook-frontend-5c8664b8cb (1/1 replicas created)
```
</details

This replica set is used to manage pods.  The appear in the `Pod.Metadata.OwnerReference` 

```
➜  go-guestbook git:(master) ✗ kubectl describe po/go-guestbook-frontend-869d5b5fb4-nsljc
Name:         go-guestbook-frontend-869d5b5fb4-nsljc
Labels:       app=guestbook
              app.kubernetes.io/managed-by=skaffold
              pod-template-hash=869d5b5fb4
              skaffold.dev/run-id=afc3bb01-54bf-4331-b854-3f9591a57ae0
              tier=frontend
Controlled By:  ReplicaSet/go-guestbook-frontend-869d5b5fb4

```

Pods from previous iteration have a different ReplicaSet id. 

This approach is correct because in case a deployment is not updated in a subsequent iteration, k8s scheduler does not spin up new pods for it.  Previous iteration pods keep running. 

Previously we were ignoring status check for previous seen pods. As a result of this, there were no events propagated to these pods. See testing notes -> https://github.com/GoogleContainerTools/skaffold/pull/6370#issuecomment-894339538 

With the updated logic, we rely on 
1) Current replica set for the deployment.
2) Fetch pods belonging to current replica set. 

If a deployment is updated and pods from previous iteration still exist, the diagnose will filter out of pods from previous iteration as they don't belong to updated replica set .

<details>
<summary> Prev approach </summary>
In this PR
 - [ ] Add `ignore` field to pod Validator to ignore pods matching a filter in given namespace
 - [ ] Add `prevPods` field to `kubernetes.StatusMonitor` which will collect all pods from previous iteration
 - [ ] Update the `prevPods` in `status.Monitor.Reset` method.
</details>

### DIFF
--- a/pkg/skaffold/kubernetes/status/status_check.go
+++ b/pkg/skaffold/kubernetes/status/status_check.go
@@ -29,6 +29,7 @@ import (
 	"golang.org/x/sync/singleflight"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	deploymentutil "k8s.io/kubectl/pkg/util/deployment"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/diag"
 	"github.com/GoogleContainerTools/skaffold/pkg/diag/validator"
@@ -56,6 +57,9 @@ var (
 
 	// report resource status for pending resources 5 seconds.
 	reportStatusTime = 5 * time.Second
+
+	// testing
+	getReplicaSet = deploymentutil.GetAllReplicaSets
 )
 
 const (
@@ -84,7 +88,6 @@ type Monitor struct {
 	deadlineSeconds int
 	muteLogs        bool
 	seenResources   resource.Group
-	prevResources   resource.Group
 	singleRun       singleflight.Group
 	namespaces      *[]string
 	kubeContext     string
@@ -98,7 +101,6 @@ func NewStatusMonitor(cfg Config, labeller *label.DefaultLabeller, namespaces *[
 		labeller:        labeller,
 		deadlineSeconds: cfg.StatusCheckDeadlineSeconds(),
 		seenResources:   make(resource.Group),
-		prevResources:   make(resource.Group),
 		singleRun:       singleflight.Group{},
 		namespaces:      namespaces,
 		kubeContext:     cfg.GetKubeContext(),
@@ -132,12 +134,6 @@ func (s *Monitor) check(ctx context.Context, out io.Writer) error {
 }
 
 func (s *Monitor) Reset() {
-	s.prevResources.Reset()
-	for k := range s.seenResources {
-		if parts := strings.Split(k, ":"); len(parts) > 0 {
-			s.prevResources.AddID(parts[0])
-		}
-	}
 	s.seenResources.Reset()
 }
 
@@ -150,7 +146,7 @@ func (s *Monitor) statusCheck(ctx context.Context, out io.Writer) (proto.StatusC
 	deployments := make([]*resource.Deployment, 0)
 	for _, n := range *s.namespaces {
 		newDeployments, err := getDeployments(ctx, client, n, s.labeller,
-			getDeadline(s.deadlineSeconds), s.prevResources)
+			getDeadline(s.deadlineSeconds))
 		if err != nil {
 			return proto.StatusCode_STATUSCHECK_DEPLOYMENT_FETCH_ERR, fmt.Errorf("could not fetch deployments: %w", err)
 		}
@@ -193,22 +189,15 @@ func (s *Monitor) statusCheck(ctx context.Context, out io.Writer) (proto.StatusC
 	// Wait for all deployment statuses to be fetched
 	wg.Wait()
 	cancel()
-	// update seen all pods retrieved.
-	s.updateSeenResources(deployments)
 	return getSkaffoldDeployStatus(c, deployments)
 }
 
-func getDeployments(ctx context.Context, client kubernetes.Interface, ns string, l *label.DefaultLabeller, deadlineDuration time.Duration, prevPods resource.Group) ([]*resource.Deployment, error) {
+func getDeployments(ctx context.Context, client kubernetes.Interface, ns string, l *label.DefaultLabeller, deadlineDuration time.Duration) ([]*resource.Deployment, error) {
 	deps, err := client.AppsV1().Deployments(ns).List(ctx, metav1.ListOptions{
 		LabelSelector: l.RunIDSelector(),
 	})
 	if err != nil {
 		return nil, fmt.Errorf("could not fetch deployments: %w", err)
-	}
-
-	var ignore []string
-	for k := range prevPods {
-		ignore = append(ignore, k)
 	}
 
 	deployments := make([]*resource.Deployment, len(deps.Items))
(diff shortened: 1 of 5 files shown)


pkg/skaffold/kubernetes/status/status_check.go
```diff
@@ -183,17 +191,24 @@ func (s *Monitor) statusCheck(ctx context.Context, out io.Writer) (proto.StatusC
...
+func getDeployments(ctx context.Context, client kubernetes.Interface, ns string, l *label.DefaultLabeller, deadlineDuration time.Duration, prevPods resource.Group) ([]*resource.Deployment, error) {
 	deps, err := client.AppsV1().Deployments(ns).List(ctx, metav1.ListOptions{
 		LabelSelector: l.RunIDSelector(),
 	})
 	if err != nil {
 		return nil, fmt.Errorf("could not fetch deployments: %w", err)
 	}
 
+	var ignore []string
+	for k := range prevPods {
```
aaron-prindle, line 208: nit: Are `prevPods` only the pods that did not come up successfully the prev iteration?  Would `unhealthyPods` possibly be more descriptive/accurate here?  I would think only unhealthy pods are added to the ignore list.

aaron-prindle (approved):
LGTM

pkg/skaffold/kubernetes/status/status_check.go
aaron-prindle: From this I'm a bit confused, is the idea to ignore all pods from prev iteration or only the unhealthy pods from the prev iteration (as per the title?)  Are they the same thing?  (this wouldn't be obvious to me as I would think the prev itertion some might have been healthy, is the idea to not report any logs until all are healthy?)

MarlonGamez: Previous pods would just be any pod from the previous iteration. This change is mostly so we ignore them in the event API. CLI users shouldn't see any difference as we don't print updates from pods normally anyway, however we do send out events for them. So this will help IDE's not report pod failures when we tear down the old pods.

tejal29: nope. They could be unhealthy or healthy pods.  
use case we have in mind is [this](https://github.com/GoogleCloudPlatform/cloud-code-intellij-internal/issues/4323#issue-951100782)

Here if you see as we are tearing down previous iteration due to `kubectl apply` event (healthy and unhealthy) related to those pods (since they have the same run-id) are shown.

MarlonGamez (commented):
LGTM and seems good when testing locally, however maybe let's hold off on merging until IDEs have tested with it. Will update with their response

MarlonGamez:
@etanshaul said he can do some testing from this branch, will wait and see if he runs into anything

etanshaul:
Looks like this fixed it!

**Before - second iteration** (notice the 2 frontend pods, presumably one of them from the previous iteration)
Screen Shot 2021-08-06 at 11 12 27 AM

**After - second iteration**
Screen Shot 2021-08-06 at 11 15 47 AM

Thanks @MarlonGamez

etanshaul:
The double `java-guestbook-frontend` pods under `Stream application logs` is interesting though...

etanshaul:
In addition to the above, I also noticed another oddity:

The set of deploy status sub-tasks we get from this branch is inconsistent across iterations (even when changing the exact same service):

Iteration 5
Screen Shot 2021-08-06 at 11 27 42 AM

Iteration 6
Screen Shot 2021-08-06 at 11 28 17 AM

It looks like the original issue of displaying statuses from previous iterations is gone, but would you expect something like this to happen?

tejal29:
> It looks like the original issue of displaying statuses from previous iterations is gone, but would you expect something like this to happen?

For iteration 6 did you only change `backend`? 
Skaffold uses `kubeclt apply` and hence we don't have control on which deployments get updated. 
Another factor for this discrepancy is,
1) We poll deployments and pods after 1 s
In case, `kubectl apply <all resources>` updated deployment `frontend` before 1s, we don't fetch status for pods for this deployment.

It shdn't be a huge code change for fetching pods. do you think it would make things easier ?

Thanks
Tejal

etanshaul:
> For iteration 6 did you only change backend?

In both cases, I only changed the `frontend`.

> Another factor for this discrepancy is ..
> It shdn't be a huge code change for fetching pods. do you think it would make things easier ?

This might explain it. I can make the same change (in the same file) over and over again and each iteration I may see a different set of status nodes.

Is your suggestion to always fetch status for pods regardless of whether or not the status check finishes in 1 sec? (e.g. poll immediately or something)

tejal29:
> Is your suggestion to always fetch status for pods regardless of whether or not the status check finishes in 1 sec? (e.g. poll immediately or something)

Yes we will fetch pods for a deployment along with their statuses even if it stabilized the first time we check deployment status.

Right now, what we do is
```
if `kubectl rollout status dep/frontend` is success:
   send ResourceStatusCheckEvent successful for dep/frontend
else:
   1) Fetch pods for dep/frontend
   2) Check statuses for pods
   3) for pod  in pods:
         send ResourceStatusCheckEvent successful for pod/frontend-xxxxx

```

etanshaul:
Got it! It _seems_ to me to make sense to also fetch pod status like you do for the deployment. Is there a particular reason you chose not to originally?

tejal29:
> Got it! It _seems_ to me to make sense to also fetch pod status like you do for the deployment. Is there a particular reason you chose not to originally?

No reason. If the deployments is stabilized no need to diagnose the pods. The decision design was for helping users surface issues in deployments

tejal29:
Let me give this a spin on this branch. We can test the change today or Monday

etanshaul:
> If the deployments is stabilized no need to diagnose the pods.

That's true. For the sake of our UI though I _think_ it's less confusing if the behavior appears consistent.

> Let me give this a spin on this branch. We can test the change today or Monday

Great

etanshaul:
@tejal29 let me know if you'd like me to give this another test on my end

tejal29:
@etanshaul Even after  tejal29/skaffold#17 (which once github has recovered I will create a PR against remote), I see the issue you mentioned about 2 frontend nodes in stream log. 

The reason being, this deleted container is still tracked.

image

etanshaul:
> The reason being, this deleted container is still tracked.

Dumb question - can't we just deleted containers from being emitted from the eventing?

tejal29:
> > The reason being, this deleted container is still tracked.
> 
> Dumb question - can't we just exclude deleted containers from being emitted from the eventing?

Yeah, we don't know which containers are going to get deleted before hand. Keeping this sync after deploy happens is a large impact change and might need some re-work.

PR #6211, opened 2021-07-14
added support for kaniko pushretry

elnoro:
Fixes: #5967  
**Related**:  none
**Merge before/after**:  none

**Description**
Added support for --push-retry flag in Kaniko. Nothing major, just filled in a new constant, a new unit test and changed the api docs. 

**User facing changes (remove if N/A)**

New configuration option pushRetry in the kaniko section.

PR #4900, opened 2020-10-11
Extended Kaniko support

ricardo-larosa:
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/4852 
**Related**: https://github.com/GoogleContainerTools/skaffold/pull/4853

**Description**

Implement kaniko flags as skaffold config items.

**User facing changes (remove if N/A)**
Support for `additionalFlags` aka
```yaml
        flags: 
```
is *removed*

**Example:** 
***Before***
```yaml
    artifacts:
    - image: registry.io/image:tag
      kaniko:
        flags: 
          - --snapshotMode=redo
          - --cleanup=true
          - --verbosity=debug
```
***Now***
```yaml
    artifacts:
    - image: registry.io/image:tag
      kaniko:
        snapshotMode: redo
        cleanup: true
        verbosity: debug
```

**Limitations/Scope**

Does not implement kaniko `--git` tag nor related.

**Notes**
It looks long but is due to the number of flags implemented.

PR #6380, opened 2021-08-07
Add support for Kaniko flag --image-fs-extract-retry

RabidFire:
Hi! My first pull request here 👋 

**Related**: 
- PR I used for reference: #6211
- PR that added the flag to Kaniko: https://github.com/GoogleContainerTools/kaniko/pull/1685

**Description**
Adds support for a newly introduced Kaniko flag, `--image-fs-extract-retry`. This is important for GCB users, as builds fail often and need to be retried by hand without this flag.

**User facing changes**
New configuration option `imageFSExtractRetry` in the kaniko section.

_Side Note:_ I wanted to play with the `--cache-copy-layers` flag as well, but this is also missing. Perhaps the previous architecture where the user could add additionalFlags themselves was better ([changed here](https://github.com/GoogleContainerTools/skaffold/pull/4900/files)). It would definitely be useful to be able to add custom flags without waiting for a skaffold release.

### DIFF
--- a/docs/content/en/schemas/v2beta21.json
+++ b/docs/content/en/schemas/v2beta21.json
@@ -2146,6 +2146,11 @@
           "description": "Docker image used by the Kaniko pod. Defaults to the latest released version of `gcr.io/kaniko-project/executor`.",
           "x-intellij-html-description": "Docker image used by the Kaniko pod. Defaults to the latest released version of <code>gcr.io/kaniko-project/executor</code>."
         },
+        "imageFSExtractRetry": {
+          "type": "string",
+          "description": "number of retries that should happen for extracting an image filesystem.",
+          "x-intellij-html-description": "number of retries that should happen for extracting an image filesystem."
+        },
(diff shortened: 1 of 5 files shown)


gsquared94 (approved):
looks good to me!

gsquared94:
> Perhaps the previous architecture where the user could add additionalFlags themselves was better (changed here). It would definitely be useful to be able to add custom flags without waiting for a skaffold release.

@tejal29 should we bring back `additionalFlags` to support arbitrary flags in addition to the typed ones?

tejal29 (approved):

tejal29:
> > Perhaps the previous architecture where the user could add additionalFlags themselves was better (changed here). It would definitely be useful to be able to add custom flags without waiting for a skaffold release.
> 
> @tejal29 should we bring back `additionalFlags` to support arbitrary flags in addition to the typed ones?

we should definitely discuss this and keep is same across all tool config (docker, kubectl, etc)

Issue #5867, opened 2021-05-19
Skaffold dev/debug should exit with error code 128 when incorrect flags are passed


sujit-kamireddy:

tejal29:
@sujit-kamireddy do you have a timeline in mind so we can plan for this accordingly.

sujit-kamireddy:
We will be taking up our feature in H2. It will be great if we can get it by July/August time frame.

tejal29:
Sounds good @sujit-kamireddy. Scheduled it for mid Aug release.

ahmetb:
I will take a stab at this, I am trying to understand why `dev` and `debug` only? would we want to make this a cli-wide behavior or just these two commands because presumably Cloud Code is integrating with those?

ahmetb:
Also is there any specific reason to stick to 128? I suspect it's best practice to document this behavior eventually, and in that case, we should probably pick the right value. Is there established convention around 128 being _invalid arguments_ (I am not finding much on it, e.g. [here](https://tldp.org/LDP/abs/html/exitcodes.html)).

tejal29:
Thanks @ahmetb.
Some context around this from my prev discussion with @sujit-kamireddy.
We should make is cli-wide behavior.  Currently, we exit with `1` exit code. 
The 
```
~/workspace/skaffold/out/skaffold dev -d gcr.io/tejal-gke1 -v=trace --help1
unknown flag: --help1
See 'skaffold dev --help' for usage.
```


Skaffold uses cobra library and we use `NoArgs` and `ExactArgs` function. 
See cmd/command.go.
Skaffold does some error message handling [here](https://github.com/GoogleContainerTools/skaffold/blob/main/cmd/skaffold/skaffold.go#L73) and return 0 or 1 error code. 
It would be great if we parse the error message and set error code to 127 (std error code when invalid args are set) if `invalid args` or `unknown flag` error is present

PR #6401, opened 2021-08-10
Set exit code 127 for CLI validation errors

ahmetb:
Parsing error strings from cobra.Execute to detect invalid commands and
arguments to fail with exit code 127.

Pending: discussion whether we need to document this as a feature or a hidden
property of the program.

Fixes: #5867

### DIFF
--- a/cmd/skaffold/app/skaffold_test.go
+++ b/cmd/skaffold/app/skaffold_test.go
@@ -83,3 +83,24 @@ func TestSkaffoldCmdline_MainUnknownCommand(t *testing.T) {
 		t.CheckError(true, err)
 	})
 }
+
+func TestMain_InvalidUsageExitCode(t *testing.T) {
+	testutil.Run(t, "unknown command", func(t *testutil.T) {
+		// --interactive=false removes the update check and survey prompt.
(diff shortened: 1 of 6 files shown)


cmd/skaffold/app/skaffold_test.go
```diff
@@ -83,3 +83,24 @@ func TestSkaffoldCmdline_MainUnknownCommand(t *testing.T) {
...
 	})
 }
+
+func TestMain_InvalidUsageExitCode(t *testing.T) {
+	testutil.Run(t, "unknown command", func(t *testutil.T) {
+		// --interactive=false removes the update check and survey prompt.
+		t.Override(&os.Args, []string{"skaffold", "unknown", "--interactive=false"})
+		err := Run(ioutil.Discard, ioutil.Discard)
+		t.CheckErrorAndExitCode(127, err)
+	})
```
tejal29, line 93: For sake of completeness, can you add a test case for
- nil error exit code should be 0
- an error other than what matches the exit code 1

tejal29: never mind, i see it is covered in the test above.

tejal29 (approved):

PR #6404, opened 2021-08-10
introduce PostRunHooks for cmd builder

ahmetb:
This helps clean up some deduplication we had in cmd builder to call
shutdownAPIServer. Now it can be done more modularly and more post-RunE() hooks
can be added (as cobra's own PostRunE() does not work if RunE returns error).

Thought it might help, but I might've also overdone it. No hard feelings if
rejected.


gsquared94 (approved):
lgtm

PR #6407, opened 2021-08-10
mute application logs when a new iteration starts.

tejal29:
Mute application logs when a new dev iteration starts. 
Before:
Skaffold would read logs from pod and send these to application log endpoint. Skaffold will **mute logs** on CLI.
 Due to this, 
1) If you make a change to frontend, it will trigger a re-build.  
2) while its building, old frontend pod is still running 
3) now go to the portforwarded frontend url and make a request. 
4) IDE users will see a logs from prev pod which served the request and new pod which got initialized
The two pods shown in IDE view confuses the users. 

https://user-images.githubusercontent.com/1534539/128949393-0c0a502b-5ad4-4d04-95cc-04b89f84ab3a.mov

Resolution:
1) Mute logs keep the behaviour same on CLI and event api v2.


MarlonGamez (approved):
nice! lgtm

Issue #6378, opened 2021-08-06
Previous iteration application logs being emitted


matthewmichihara:
Skaffold built from: 5ceca63
Cloud Code built from: GoogleCloudPlatform/cloud-code-intellij-internal@8276037

(This might be working as intended, and not a bug)

![image](https://user-images.githubusercontent.com/79073/128540092-6b58db6f-2e97-42c9-8c55-80ead7ead0c3.png)

The above screenshot shows the second iteration of developing the go-guestbook sample app. Notice that in the `Stream application logs` tree node, there are two `go-guestbook-backend-*` nodes. This app only has 1 backend pod and 1 frontend pod, so we'd expect there to only be 1 backend application log node here.

What I think is happening is that the previous iteration's backend logs are still being received in addition to the current iteration's.

etanshaul:
this looks very similar, or the same underlying issue, as this comment: https://github.com/GoogleContainerTools/skaffold/pull/6370#issuecomment-894335725

tejal29:
This can be closed now.

PR #6410, opened 2021-08-11
Move deployment controller check to pod validator

tejal29:
Fixes NPE seen in our integration tests https://app.travis-ci.com/github/GoogleContainerTools/skaffold/jobs/530583442#L1835

If a cluster is busy, it takes longer for a replicate set for a deployment to be created.
 #6378 add logic to fetch pods for a deployment corresponding to a `NewReplicaSet` field. However at the time a deployment is created, replica set might not exist.  The `controller` object could be null [here](https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/kubernetes/status/status_check.go#L212) 

Moving this logic to Validator so at the time of fetching pods, we could check if a deployment has a replicate set. If it does, then fetch pods and filter based on replica set.

### DIFF
--- a/pkg/diag/validator/pod.go
+++ b/pkg/diag/validator/pod.go
@@ -24,11 +24,13 @@ import (
 	"strings"
 
 	"github.com/sirupsen/logrus"
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	deploymentutil "k8s.io/kubectl/pkg/util/deployment"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/diag/recommender"
 	"github.com/GoogleContainerTools/skaffold/proto/v1"
@@ -58,7 +60,8 @@ var (
 	runContainerRe = regexp.MustCompile(errorPrefix)
 	taintsRe       = regexp.MustCompile(taintsExp)
 	// for testing
-	runCli = executeCLI
+	runCli        = executeCLI
+	getReplicaSet = deploymentutil.GetAllReplicaSets
 
 	unknownConditionsOrSuccess = map[proto.StatusCode]struct{}{
 		proto.StatusCode_STATUSCHECK_UNKNOWN:                   {},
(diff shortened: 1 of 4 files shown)


pkg/diag/validator/pod.go
```diff
@@ -71,27 +74,37 @@ var (
...
-func NewPodValidator(k kubernetes.Interface, controller metav1.Object) *PodValidator {
+func NewPodValidator(k kubernetes.Interface, d appsv1.Deployment) *PodValidator {
 	rs := []Recommender{recommender.ContainerError{}}
-	return &PodValidator{k: k, recos: rs, controller: controller}
+	return &PodValidator{k: k, recos: rs, depObj: d}
 }
 
 // Validate implements the Validate method for Validator interface
 func (p *PodValidator) Validate(ctx context.Context, ns string, opts metav1.ListOptions) ([]Resource, error) {
+	var controller *appsv1.ReplicaSet
```
gsquared94, line 90: `controller` could be declared directly where initialized on line 99.

pkg/diag/validator/pod.go
```diff
@@ -71,27 +74,36 @@ var (
 
 // PodValidator implements the Validator interface for Pods
 type PodValidator struct {
-	k          kubernetes.Interface
-	controller metav1.Object
-	recos      []Recommender
+	k      kubernetes.Interface
+	depObj appsv1.Deployment
```
gsquared94, line 78: is there a more generic type than `appsv1.Deployment` that we can use here. Thinking if we extend the status-check to `statefulsets` in the future, is there a way this code wouldn't need to be refactored again.

tejal29: We could refactor this later. Right now I do want to get this in since we don't do any resource check for statefulsets.

pkg/diag/validator/pod_test.go
```diff
@@ -861,3 +870,8 @@ type result struct {
 	iScheduledNotReady bool
 	isUnknown          bool
 }
+
+func truePtr() *bool {
```
gsquared94, line 874: can use `util.BoolPtr(true)` ?

gsquared94 (approved):
lgtm

pkg/diag/validator/pod_test.go
tejal29: I thought about it.  We can't because the util is defined in `pkg/skaffold/util`
1) `diag` should not depend on `skaffold`. Right thing to do would be to add pkg/diag/util  or move this to `pkg/util` so both libs can share the uitl code.
2) I didnt check but it could have round trip dependencies.

PR #6419, opened 2021-08-11
Change `logrus` calls in `pkg/` and `cmd/` to use `log.Entry(ctx)`

MarlonGamez:
**Related**: #5368 

**Description**
This PR is mostly a large refactor to begin using`pkg/skaffold/output/log` to do logging, rather than making calls directly to `logrus.*`

`log.Entry(context.Context)` has been introduced to extract information from the given context and construct a `logrus.Entry` from it. Functions that don't currently have access to a `context.Context` will make calls using `context.Background()` for now.


pkg/skaffold/build/bazel/dependencies.go
```diff
@@ -49,7 +48,7 @@ func GetDependencies(ctx context.Context, dir string, a *latestV1.BazelArtifact)
 
 	go func() {
 		<-timer.C
-		once.Do(func() { logrus.Warnln("Retrieving Bazel dependencies can take a long time the first time") })
+		once.Do(func() { log.Entry(ctx).Warn("Retrieving Bazel dependencies can take a long time the first time") })
```
tejal29, line 51: Do we need to append a new line at the end due to `Warnln` -> `Warn` change?

pkg/skaffold/initializer/testdata/init/microservices/leeroy-web/web.go
```diff
@@ -2,9 +2,8 @@ package main
 
 import (
 	"io"
-	"net/http"
```
tejal29, line 5: please revert `pkg/skaffold/initializer/testdata`

tejal29 (commented):
LGTM!

1) Revert few files.
2) Will take a look at the `diag` package logging more.

pkg/skaffold/build/bazel/dependencies.go
MarlonGamez: This was a mistake but it shouldn't actually have any effect. The only difference in how the `ln` vs non `ln` logging functions work is how they append multiple arguments to one another. Since there's only one arg being passed here it shouldn't effect anything

MarlonGamez: I did find a couple spots where I should've used `ln` but didn't, and fixed those ones. Thanks for pointing out 👍🏼

tejal29 (approved):

PR #6428, opened 2021-08-13
Fix #6408 - Add `--load-images` flag to deploy

### LINKED ISSUE #6408: Provide a way for users to force loading image to cluster local cluster (KinD, K3d) using `skaffold deploy`
### Expected behavior

`skaffold deploy --images=local/image:latest`

Successfully deploys image to KinD or K3d cluster

### Actual behavior

Because image wasn't built via this runner, we dont load the image to local cluster causing errors. It seems KinD specifically adds docker.io to images without a host. When using `kind load docker-image` it also appends docker.io to the local image on the nodes allowing this scenario to be successful.

Example error:
```
Failed to pull image "local/image:latest": rpc error: code = Unknown desc = failed to pull and unpack image "docker.io/local/image:latest": failed to resolve reference "docker.io/local/image:latest": pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

### Solution

Suggested adding `--load-image` flag to `deploy` command to force loading image to cluster

See issue #6379, PR #6396

verbanicm:
Fixes: #6408, #6379 

**Description**
- Added `--load-images` flag for the `deploy` command to force load images to KinD or K3d local clusters
- Since deploy command has no context for what was built, this was the best option to let the user tell skaffold to push the image rather than a manual process outside of skaffold

- Documentation will be updated to reflect changes for new command line flag for `deploy` command


tejal29 (approved):
LGTM. 
Havent tried this manually yet

pkg/skaffold/runner/v1/deploy.go
```diff
@@ -111,6 +111,11 @@ See https://skaffold.dev/docs/pipeline-stages/taggers/#how-tagging-works`)
 		}
 	}
 
+	// if --load-images=true, load all images into cluster
+	if r.runCtx.ForceLoadImages() {
+		localAndBuiltImages = artifacts
```
briandealwis, line 116: This doesn't seem right: these images may be remote images and they should be pulled into the cluster through the normal mechanism.

I think you've diagnosed that our `localImages` is missing images.  Is the problem with `isImageLocal()` or elsewhere?

verbanicm: The issue was that since the images weren't built by this runner, IE they were built by a previous build command and passed as artifacts file, the loop above doesnt append anything to `localAndBuiltImages` (r.wasBuilt returns false). Therefore the images dont get pushed to cluster.

My solution was to let the user tell us that we need to load these images. I see what you are saying about images could be remote, I can move the logic up into the for loop and check `if r.runCtx.ForceLoadImages() || r.wasBuilt` then append to `localAndBuiltImages`.

Does this make more sense since we would only consider local images then?

PR #6436, opened 2021-08-16
hooks: add metrics

gsquared94:
This PR adds instrumentation for `lifecycle hooks` usage. It collects the number of hooks configured per phase.

### DIFF
--- a/pkg/skaffold/instrumentation/export.go
+++ b/pkg/skaffold/instrumentation/export.go
@@ -167,6 +167,8 @@ func createMetrics(ctx context.Context, meter skaffoldMeter) {
 	m := global.Meter("skaffold")
 
 	// cloud monitoring only supports string type labels
+	// cloud monitoring only supports 10 labels per metric descriptor
+	// be careful when appending new values to this `labels` slice
 	labels := []attribute.KeyValue{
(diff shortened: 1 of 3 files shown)


pkg/skaffold/instrumentation/export.go
```diff
@@ -167,6 +167,8 @@ func createMetrics(ctx context.Context, meter skaffoldMeter) {
 	m := global.Meter("skaffold")
 
 	// cloud monitoring only supports string type labels
+	// cloud monitoring only supports 10 labels per metric descriptor
+	// be careful when appending new values to this `labels` slice
```
MarlonGamez, line 171: do we track how many labels we are using so far and what they're being used for?

MarlonGamez (approved):
LGTM 👍🏼

pkg/skaffold/instrumentation/export.go
gsquared94: I meant this as a comment for the immediate `labels` slice on line 172

gsquared94: but you're right, we don't have any document for the exact set across all metrics

Issue #5869, opened 2021-05-19
Deploy section in profile does not fully replace deployment configuration


ekuefler:
### Expected behavior

According to https://skaffold.dev/docs/environment/profiles/, "the specified build, test and deploy configuration in [the profile] will replace the build, test and deploy sections declared in the main section of skaffold.yaml". I expect that if I specify a deploy section in a profile, it should just use that and not retain any information form the main section.

### Actual behavior

Skaffold runs the `kubectl` command defined in the main section even though I only specified `kustomize` in my profile. I can suppress this if I explicitly set `kubectl: {}` in my profile.

### Information

- Skaffold version: v1.23.0
- Operating system: Linux
- Installed via: skaffold.dev
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta14
kind: Config
metadata:
  name: environments

build:
  tagPolicy:
    sha256: {}
  artifacts:
  - image: gcr.io/devops-environments/manager
    docker:
      dockerfile: cmd/manager/Dockerfile
deploy:
  kubectl:
    remoteManifests:
    - environments-system:deployment/manager

profiles:
- name: release
  activation:
  - command: render
  test:
  - image: gcr.io/devops-environments/manager
    custom:
    - command: make test
  deploy:
#    kubectl: {}   # Uncomment to fix
    kustomize:
      paths:
      - config/release
```

### Steps to reproduce the behavior

1. Run `skaffold render` using the above configuration
2. Observe that Skaffold tries to read a remote manifest from the context

### Log output

```
time="2021-05-19T23:51:02Z" level=info msg="Skaffold &{Version:v1.23.0 ConfigVersion:skaffold/v2beta15 GitVersion: GitCommit:e8f3c652112c338e75e03497bc8ab09b9081142d BuildDate:2021-04-28T00:53:10Z GoVersion:go1.14.14 Compiler:gc Platform:linux/amd64}"
time="2021-05-19T23:51:02Z" level=info msg="Loaded Skaffold defaults from \"/usr/local/google/home/ekuefler/.skaffold/config\""
time="2021-05-19T23:51:02Z" level=debug msg="config version out of date: upgrading to latest \"skaffold/v2beta15\""
time="2021-05-19T23:51:02Z" level=debug msg="parsed 1 configs from configuration file /usr/local/google/home/ekuefler/environments/skaffold.yaml"
time="2021-05-19T23:51:02Z" level=info msg="applying profile: release"
[...]

PR #6445, opened 2021-08-16
Change wording describing how profiles work

MarlonGamez:
**Related**: #5869

**Description**
This PR changes some of the wording used when describing how profile activation works.


MarlonGamez:
Should I add an example to showcase how exactly this might cause issues for the user? Like #5869 ?

tejal29 (approved):

tejal29:
> Should I add an example to showcase how exactly this might cause issues for the user? Like #5869 ?

You could say something like "To completely replace a config section see <link>" and  point to the Json Patch example below

Issue #6424, opened 2021-08-12
[Flake] TestDevGracefulCancel no output for 10 mins


tejal29:
```
=== RUN   TestDevGracefulCancel/getting-started
time="2021-08-12T20:45:55Z" level=info msg="Ran [skaffold dev --namespace skaffoldbnrts --default-repo gcr.io/k8s-skaffold] in 32.464 seconds"
time="2021-08-12T20:45:55Z" level=info msg="Namespace: skaffoldvjhqk"
time="2021-08-12T20:45:55Z" level=info msg="Running [skaffold dev --namespace skaffoldvjhqk --default-repo gcr.io/k8s-skaffold] in examples/getting-started"
time="2021-08-12T20:45:55Z" level=info msg="Waiting for pods [getting-started] to be ready"
Listing files to watch...
 - skaffold-example
Generating tags...
 - skaffold-example -> gcr.io/k8s-skaffold/skaffold-example:v1.29.0-40-g50a585d
Checking cache...
 - skaffold-example: Found Locally
Starting test...
Tags used in deployment:
 - skaffold-example -> gcr.io/k8s-skaffold/skaffold-example:55673b440103e61c7db744acfb318e6027d024e45e3da530a384c2a5962c6f48
Starting deploy...
Loading images into kind cluster nodes...
 - gcr.io/k8s-skaffold/skaffold-example:55673b440103e61c7db744acfb318e6027d024e45e3da530a384c2a5962c6f48 -> Loaded
Images loaded in 920.236524ms
 - pod/getting-started created
time="2021-08-12T20:45:57Z" level=info msg="Still waiting for pods [getting-started]"
Waiting for deployments to stabilize...
time="2021-08-12T20:45:57Z" level=info msg="Still waiting for pods [getting-started]"
Deployments stabilized in 49.670341ms
Waiting for deployments to stabilize...
time="2021-08-12T20:45:57Z" level=info msg="Still waiting for pods [getting-started]"
Deployments stabilized in 16.485819ms
Press Ctrl+C to exit
Watching for changes...
time="2021-08-12T20:45:59Z" level=info msg="Pods marked as ready: map[getting-started:true]"
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
```

tejal29:
happened twice on same PR

[...]

PR #6446, opened 2021-08-16
add skaffold inspect tests list command

### LINKED ISSUE #6355: Add command/functionality to skaffold to verify whether tests will run for a given skaffold.yaml
For projects that use skaffold as a dependency it can be helpful to understand what tests a skaffold.yaml have defined and are applicable for a given profile,etc.  This bug tracks extending `skaffold inspect` command to inspect the complete skaffold config (aggregated from all relevant `skaffold.yaml` files) to verify if a test will occur.

aaron-prindle:
###  What is the problem being solved?
Fixes #6355, adding skaffold command to verify if tests will be run.  This is done by extending the `skaffold inspect` command to have a new subcommand `skaffold inspect tests` and `skaffold inspect tests list`.  This is needed as an upstream project that depends on skaffold requires the ability to introspect from a users skaffold configuration (all `skaffold.yamls`, modules, profiles, etc.) whether a test will run or not.  

###  Why is this the best approach?
This approach is best as it re-uses the previously created `skaffold inspect` command, the goal if which (inspecting skaffold configuration and outputting usable json) is nearly identical to what is desired for #6355. The only small issue is that currently this outputs a flat list of json objects which represent both `CustomTest` and `StructureTest` tests.  The upstream project mentioned they only require the ability to know `will tests run for these configs+modules+profiles`.  This can be done though via some jq or other ingestion of the json, for example `jq -e '.tests | select(length != 0)' inspect-tests-list.json` (returns exitCode: 4 if array is empty)

### What other approaches did you consider?
Other approaches included:
- `skaffold test --dry-run` surface (vs `skaffold inspect tests lists`) to see what tests might be run.  A bit cludgy espectially considering `skaffold inspect already exists`
- Making `skaffold inspect tests lists` surface have custom outputter that could output bools to specifically answer the question required w/ no json parsing - `will tests run for these configs+modules+profiles`

###  What side effects will this approach have?
There shouldn't be any side effects w/ this approach as this is a new command and the `skaffold inspect` command itself is hidden from users and not widely used by most skaffold users

###  What future work remains to be done?
Future work includes:
    - possibly adding `skaffold inspect [build|deploy] list` to flesh out `skaffold inspect` for all phases
    - possibly adding more outputters/flag-opts for specific `skaffold inspect` use cases

### DIFF
--- a/cmd/skaffold/app/cmd/inspect_tests.go
+++ b/cmd/skaffold/app/cmd/inspect_tests.go
@@ -1,0 +1,57 @@
+/*
+Copyright 2021 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"context"
+	"io"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/inspect"
+	tests "github.com/GoogleContainerTools/skaffold/pkg/skaffold/inspect/tests"
+)
+
+func cmdTests() *cobra.Command {
+	return NewCmd("tests").
+		WithDescription("View skaffold test information").
+		WithPersistentFlagAdder(cmdTestsFlags).
+		WithCommands(cmdTestsList())
+}
+
+func cmdTestsList() *cobra.Command {
+	return NewCmd("list").
+		WithExample("Get list of tests", "inspect tests list --format json").
+		WithExample("Get list of tests targeting a specific configuration", "inspect tests list --profile local --format json").
+		WithDescription("Print the list of tests that would be run for a given configuration (default skaffold configuration, specific module, specific profile, etc).").
+		NoArgs(listTests)
+}
+
+func listTests(ctx context.Context, out io.Writer) error {
+	return tests.PrintTestsList(ctx, out, inspect.Options{
+		Filename:     inspectFlags.filename,
+		RepoCacheDir: inspectFlags.repoCacheDir,
+		OutFormat:    inspectFlags.outFormat,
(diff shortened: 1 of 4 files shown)


aaron-prindle:
@gsquared94 should I also add some filtering options for buildEnvs as well similar to `skaffold inspect profiles list`?  Not sure I understand if tests can be tied to buildEnvs, it didn't seem so from my analysis

gsquared94:
> @gsquared94 should I also add some filtering options for buildEnvs as well similar to `skaffold inspect profiles list`? Not sure I understand if tests can be tied to buildEnvs, it didn't seem so from my analysis

you're right, tests can't be tied to build-env

cmd/skaffold/app/cmd/inspect_tests.go
```diff
@@ -0,0 +1,56 @@
...
+		WithDescription("Interact with configuration tests").
+		WithPersistentFlagAdder(cmdTestsFlags).
+		WithCommands(cmdTestsList())
+}
+
+func cmdTestsList() *cobra.Command {
+	return NewCmd("list").
+		WithExample("Get list of tests", "inspect tests list --format json").
+		WithExample("Get list of tests targeting GoogleCloudBuild environment", "inspect tests list --build-env googleCloudBuild --format json").
+		WithDescription("Print the list of profile names that can be invoked with the --profile flag in other skaffold commands.").
```
gsquared94, line 41: description seems incorrect for the command.

gsquared94: also this could use a `--profile` flag similar to the `--module` flag?

aaron-prindle: fixed both, added better description (forgot this sorry) and added `--profile` flag w/ similar description to buildEnv (where `--profile` flag is also used)

pkg/skaffold/inspect/types.go
```diff
@@ -128,3 +128,18 @@ func GetBuildEnv(t *latestV1.BuildType) BuildEnv {
 		return BuildEnvs.Local
 	}
 }
+
+// ================
```
gsquared94, line 132: remove commented code?

gsquared94 (approved):
lgtm

aaron-prindle:
Submitted multiple times, getting repeated known test flakes:
- https://github.com/GoogleContainerTools/skaffold/issues/6450
- https://github.com/GoogleContainerTools/skaffold/issues/6424

Testing via `make test` passes locally and as this makes no changes to other skaffold command this means the new functionality is correct w/o changing any other components.

As change adds a new command with tests that does not affect any other skaffold functionality, I am merging

PR #6437, opened 2021-08-16
Refine ko builder behavior for main packages

halvards:
**Description**
Decide on how ko builder users should specify the location of `package main`.

Previous discussion: https://github.com/GoogleContainerTools/skaffold/pull/6054#discussion_r660727851

Tracking: #6041
Related: #6054, #6286

PR #6447, opened 2021-08-17
Implement Target support for ko builder

halvards:
**Description**

This implements support for the `Target` config field for the ko builder.

The `Target` field allows users to specify a target for `go build`. This is necessary where the main package is not in the context directory.

Creates a `testdata` directory under `./pkg/skaffold/build/ko` with test samples. This is to avoid depending on test samples from other packages.

Tracking: #6041
Related: #6054, #6437

### DIFF
--- a/pkg/skaffold/build/ko/schema/temporary.go
+++ b/pkg/skaffold/build/ko/schema/temporary.go
@@ -36,6 +36,14 @@ type KoArtifact struct {
 	// Defaults to `all` to build for all platforms supported by the
 	// base image.
 	Platforms []string `yaml:"platforms,omitempty"`
+
+	// Target is the location of the main package.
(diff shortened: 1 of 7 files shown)


pkg/skaffold/build/ko/schema/temporary.go
```diff
@@ -36,6 +36,14 @@ type KoArtifact struct {
 	// Defaults to `all` to build for all platforms supported by the
 	// base image.
 	Platforms []string `yaml:"platforms,omitempty"`
+
```
tejal29, line 39: 👍 for temporary schema changes file.

tejal29 (approved):

pkg/skaffold/build/ko/schema/temporary.go
halvards: 👍

PR #6210, opened 2021-07-14
add flag `sync-remote-cache`

gsquared94:
Fixes: #6145 

**Description**

This PR implements a flag `sync-remote-cache` which has three values
```
	// clone missing repositories and sync them on each skaffold run
	always = "always"
	// clone missing repositories but do not sync found repositories
	missing = "missing"
	// do not clone missing repositories, and do not sync found repositories
	never = "never"
```

This flag allows users to turn off syncing their remote modules on every run of `skaffold`.
It supersedes the value set in the config.

PR #6453, opened 2021-08-18
fix: do not sync remote configs for specific commands

gsquared94:
**Description**

In https://github.com/GoogleContainerTools/skaffold/pull/6210 we added the flag `sync-remote-cache`. This PR changes the default value of this flag to `missing` for the commands `inspect`, `diagnose` and `fix`. This will speed up these commands for projects using remote skaffold modules.


MarlonGamez (approved):
lgtm 👍🏼

PR #6468, opened 2021-08-19
Plumb `context.Context` down into functions that use `util/cmd` functions

MarlonGamez:
Related: #5368 

**Description**
This PR plumbs `context.Context` parameter into functions in order to have a `context.Context` variable in `util.RunCmd` and related functions

### DIFF
--- a/hack/versions/pkg/schema/git.go
+++ b/hack/versions/pkg/schema/git.go
@@ -73,7 +73,7 @@ func (g *git) diffWithBaseline(path string) ([]byte, error) {
 
 func (g *git) run(args ...string) ([]byte, error) {
 	cmd := exec.Command(g.path, args...)
-	out, err := util.RunCmdOut(cmd)
+	out, err := util.RunCmdOut(ctx, cmd)
 	if err != nil {
 		return nil, fmt.Errorf("failed running %v: %s\n%s", cmd.Args, err, string(out))
 	}
--- a/hack/versions/pkg/schema/git.go
+++ b/hack/versions/pkg/schema/git.go
@@ -17,6 +17,7 @@ limitations under the License.
 package schema
 
 import (
+	"context"
 	"fmt"
 	"os/exec"
 	"strings"
@@ -73,7 +74,7 @@ func (g *git) diffWithBaseline(path string) ([]byte, error) {
 
 func (g *git) run(args ...string) ([]byte, error) {
 	cmd := exec.Command(g.path, args...)
-	out, err := util.RunCmdOut(ctx, cmd)
+	out, err := util.RunCmdOut(context.Background(), cmd)
 	if err != nil {
 		return nil, fmt.Errorf("failed running %v: %s\n%s", cmd.Args, err, string(out))
 	}
--- a/pkg/skaffold/event/v2/logger.go
+++ b/pkg/skaffold/event/v2/logger.go
@@ -59,16 +59,10 @@ func (ev *eventHandler) handleSkaffoldLogEvent(e *proto.SkaffoldLogEvent) {
 }
 
 // logHook is an implementation of logrus.Hook used to send SkaffoldLogEvents
-type logHook struct {
-	task    constants.Phase
-	subtask string
-}
+type logHook struct{}
 
-func NewLogHook(task constants.Phase, subtask string) logrus.Hook {
-	return logHook{
-		task:    task,
-		subtask: subtask,
-	}
+func NewLogHook() logrus.Hook {
+	return logHook{}
 }
 
 // Levels returns all levels as we want to send events for all levels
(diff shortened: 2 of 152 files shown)


hack/versions/pkg/schema/git.go
```diff
@@ -73,7 +74,7 @@ func (g *git) diffWithBaseline(path string) ([]byte, error) {
 
 func (g *git) run(args ...string) ([]byte, error) {
 	cmd := exec.Command(g.path, args...)
-	out, err := util.RunCmdOut(cmd)
+	out, err := util.RunCmdOut(context.Background(), cmd)
```
tejal29, line 77: There is another PR to move replace context.Background to context.TODO.
Do you want to sync with that and see why `TODO` is preferred?

From docs, looks like Background context is never cancelled. 
https://pkg.go.dev/context#Background

tejal29 (approved):

tejal29:
Code changes Look good. 
Trying this out.

tejal29:
After applying a diff and some validation I can now see debug logs in relevant nodes. 
e.g. skaffold deploy logs are not seen in deploy node for `skaffold` run config with debug level logging.

image



Diff i applied:
```
diff --git a/pkg/skaffold/event/v2/logger.go b/pkg/skaffold/event/v2/logger.go
index d38e68a8c..bf19f8e9c 100644
--- a/pkg/skaffold/event/v2/logger.go
+++ b/pkg/skaffold/event/v2/logger.go
@@ -86,9 +86,16 @@ func (h logHook) Levels() []logrus.Level {
 
 // Fire constructs a SkaffoldLogEvent and sends it to the event channel
 func (h logHook) Fire(entry *logrus.Entry) error {
+       task := constants.DevLoop
+       if s, ok := entry.Data["task"]; ok {
+               fmt.Println("found for entry", entry.Message)
+               task = s.(constants.Phase)
+       } else {
+               fmt.Println("not found for ", entry.Message, entry.Data)
+       }
        handler.handleSkaffoldLogEvent(&proto.SkaffoldLogEvent{
-               TaskId:    fmt.Sprintf("%s-%d", h.task, handler.iteration),
-               SubtaskId: h.subtask,
+               TaskId:    fmt.Sprintf("%s-%d", task, handler.iteration),
+               SubtaskId: entry.Data["subtask"].(string),
                Level:     levelFromEntry(entry),
                Message:   entry.Message,
        })
```

tejal29:
Testing for BUILD scenario. 

Seeing `mvn build` command in the build node for frontend and backend.

image

tejal29:
Portfowarding logs are in the right section
image

tejal29:
Follow up cleaning tasks
1) `output.WithContext` should also return the `traceContext` and `endTrace` or `instrumentation.StartTrace` should preserve context values.
right now, `instrumentation.StartTrace(ctx, "Cleanup")` clears all context values if tracing if enabled.

pkg/skaffold/event/v2/logger.go
```diff
@@ -87,8 +81,8 @@ func (h logHook) Levels() []logrus.Level {
 // Fire constructs a SkaffoldLogEvent and sends it to the event channel
 func (h logHook) Fire(entry *logrus.Entry) error {
 	handler.handleSkaffoldLogEvent(&proto.SkaffoldLogEvent{
-		TaskId:    fmt.Sprintf("%s-%d", h.task, handler.iteration),
-		SubtaskId: h.subtask,
+		TaskId:    fmt.Sprintf("%s-%d", entry.Data["task"], handler.iteration),
```
tejal29, line 84: Can you verify if this will not cause a null pointer exception

tejal29: (if the key does not exists)

MarlonGamez: my bad, just pushed changes to fix this

tejal29 (approved):
LGTM for Monday in advance.

gsquared94:
can this be merged @MarlonGamez ?

hack/versions/pkg/schema/git.go
gsquared94: I think that `context.TODO` is a hint to the dev that if ever a proper context becomes available in future by a change in the function parameters or something, then plumb that through. Whereas `context.Background` is a deliberate use even if a parent context is available. That's what I understood from @ahmetb's PR.

ahmetb: yeah that's the case. ideally a context.Background() is for a call stack that can be cancelled from a central place, e.g. Ctrl-C. Background threads with separate lifecycle (as the name also says) get a new Background context.

If you don't plumb yet but intend do plumb someday, you can use context.TODO, which is identical to Background() functionally, only different in semantics and indicates your intention.

PR #6469, opened 2021-08-20
Wrap tester in `SkaffoldRunner` to improve `SkaffoldLogEvent` labelling

MarlonGamez:
Fixes: #5368 

**Description**
This PR adds proper labelling of `SkaffoldLogEvent`s that come from Skaffold's testing phase. I've also added a `Test` function onto the `SkaffoldRunner` rather than embedding a tester. This allows us to use `output.WithEventContext` in a way that's more consistent with how we use it for build and deploy phases.

`SkaffoldLogEvent`s starting from `Starting Test...` up to, but not including, `Tags used in deployment` should now have proper task information set.

Tests were not added for `SkaffoldRunner.Test` as there is no real logic there, it's simply calling the underlying data structures Test function.

### DIFF
--- a/pkg/skaffold/runner/v1/dev.go
+++ b/pkg/skaffold/runner/v1/dev.go
@@ -259,7 +259,7 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
 	for i := range artifacts {
 		artifact := artifacts[i]
 		if err := r.monitor.Register(
-			func() ([]string, error) { return r.Tester.TestDependencies(artifact) },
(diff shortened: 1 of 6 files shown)


tejal29:
The code changes look good. Can you please add the new event changes?

tejal29 (commented):
Please specify the event difference.

MarlonGamez:
@tejal29 `SkaffoldLogEvent`s starting from `Starting Test...` up to, but not including, `Tags used in deployment` should now have proper task information set. I updated the PR description to include this information 👍🏼

tejal29 (approved):

tejal29 (approved):

PR #6493, opened 2021-08-23
Fix skaffold inspect tests flags and help message

aaron-prindle:
skaffold inspect tests did not display its flag in its help message initially, the changes here fix this.  Additionally the profile and modules flags were not wired up/plumbed properly initially.  They are now tested and working as expected

### DIFF
--- a/pkg/skaffold/inspect/types.go
+++ b/pkg/skaffold/inspect/types.go
@@ -37,6 +37,15 @@ type Options struct {
 	ModulesOptions
 	ProfilesOptions
 	BuildEnvOptions
+	TestsOptions
+}
+
+// TestsOptions holds flag values for various `skaffold inspect tests` commands
+type TestsOptions struct {
+	// Profiles is the slice of profile names to activate.
+	TestsProfiles []string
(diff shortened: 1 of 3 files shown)


pkg/skaffold/inspect/types.go
```diff
@@ -37,6 +37,15 @@ type Options struct {
 	ModulesOptions
 	ProfilesOptions
 	BuildEnvOptions
+	TestsOptions
+}
+
+// TestsOptions holds flag values for various `skaffold inspect tests` commands
+type TestsOptions struct {
+	// Profiles is the slice of profile names to activate.
+	TestsProfiles []string
```
aaron-prindle, line 46: Note: name `TestsProfiles` was used (vs `Profiles`) as the name `Profiles` is used in the `BuildEnvs` options which would make `Profiles` ambiguous when used as `opts.Profiles`

gsquared94 (approved):

aaron-prindle:
Kokoro is failing for unrelated reason, this change is to a new command and it does not impact the GCP tests that kokoro runs.

Issue #6492, opened 2021-08-23
Inconsistent line breaks in LogEvent messages (V2 API)


etanshaul:
See https://github.com/GoogleCloudPlatform/cloud-code-intellij-internal/issues/4563

To repro:
- Using the latest version of skaffold built from HEAD (latest commit at the time of opening this issue)
- Run `skaffold dev` with `--verbosity debug` on the Cloud Code java-guestbook sample
-  Observe the output LogEvents from the V2 API

![image](https://user-images.githubusercontent.com/1735744/130515762-89443afb-a32e-495c-9b6b-e0080f99d7d7.png)

Debug level message - no line break
```
skaffoldLogEvent {
  task_id: "Build-1"
  subtask_id: "-1"
  level: INFO
  message: "Tags generated in258.279\302\265s"
}
```

default message - line break
```
skaffoldLogEvent {
  task_id: "Build-1"
  subtask_id: "java-guestbook-backend"
  level: UNKNOWN_ENUM_VALUE_LogLevel_-1
  message: "\033[32mFound Remotely\033[0m\n"
}
```

There is also some seemingly incorrect data in the `level` field.

PR #6506, opened 2021-08-24
Add newlines to `SkaffoldLogEvent`s that come from `logrus`

MarlonGamez:
**Related**: https://github.com/GoogleContainerTools/skaffold/issues/6492

**Description**
This PR adds line breaks to the end of messages that come from logrus and get sent out through `SkaffoldLogEvent`s. 
Before, we sent `entry.Message`, but now we will send `fmt.Sprintf("%s\n", entry.Message)`

Also fixes a few log lines that should be the `ln` form of their function call.

**Follow-up Work (remove if N/A)**
Another PR will be opened to add a default log level for basic output that comes from skaffold. That PR will close #6492

PR #6507, opened 2021-08-24
Add new `enums.LogLevel` to indicate skaffold's standard output

MarlonGamez:
Related: https://github.com/GoogleContainerTools/skaffold/pull/6506
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/6492

**Description**
This PR adds a new LogLevel `STANDARD` which is meant to indicate that the associated output is coming from Skaffold's normal output, rather than from `logrus`. This is a step up from just setting `-1` as the `level` value, which results in odd output.


proto/enums/enums.proto
```diff
@@ -99,6 +99,8 @@ enum LogLevel {
     PANIC = 5;
     // Trace Level
     TRACE = 6;
+    // Skaffold's standard output level
```
ahmetb, line 102: I think the "Skaffold standard output" doesn't precisely communicate to the reader.
Could something like "User-visible output level" work better (prob ok to keep the enum as `STANDARD`)?

MarlonGamez: Sounds good to me 👍🏼

tejal29 (approved):

Issue #6481, opened 2021-08-20
[Docs fixit] [technical writers review] Multiple links to the same content


gsquared94:
<img width="842" alt="image" src="https://user-images.githubusercontent.com/39389231/130270402-a8cb5adc-648b-4d57-8acb-f2356c400406.png">

https://skaffold.dev/docs/tutorials/

This behavior is found on several pages.

briandealwis:
The `docsy` theme started auto-generating these subpage links.  They're always added at the bottom of the page though, but maybe that's configurable?  They can be turned off, but maybe we should just take advantage of them and remove our hand-rolled lists.

PR #6540, opened 2021-08-30
docs: fix duplicate subpage listing

gsquared94:
Fixes: #6481

### DIFF
--- a/docs/content/en/docs/tutorials/_index.md
+++ b/docs/content/en/docs/tutorials/_index.md
@@ -2,6 +2,7 @@
 title: "Tutorials"
 linkTitle: "Tutorials"
 weight: 90
+no_list: true
 ---
 
 See the [Github Examples page](https://github.com/GoogleContainerTools/skaffold/tree/main/examples) for examples.
--- a/docs/content/en/docs/tutorials/_index.md
+++ b/docs/content/en/docs/tutorials/_index.md
@@ -2,7 +2,7 @@
 title: "Tutorials"
 linkTitle: "Tutorials"
 weight: 90
-no_list: true
+simple_list: true
 ---
 
 See the [Github Examples page](https://github.com/GoogleContainerTools/skaffold/tree/main/examples) for examples.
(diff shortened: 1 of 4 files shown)


MarlonGamez:
I think this is an interesting one because our manually created lists don't seem to match up 100%, but the auto generated ones do. Do you think we should continue with the manual lists and try to maintain them?

briandealwis:
I'm with @MarlonGamez — let Hugo/docsy do the heavy lifting

gsquared94:
> I'm with @MarlonGamez — let Hugo/docsy do the heavy lifting

done 👍

docs/content/en/docs/tutorials/_index.md
```diff
@@ -22,10 +23,3 @@ When deploying to a remote cluster you have to point Skaffold to your default im
 ### What's next
```
briandealwis, line 23: It's odd seeing a `what's next` before the actual content.  TBH, I think we should:

1. turn the default-repo into a callout `{{ <alert> }}` — but really this should be incorporated into each of the tutorial pages
2. move the survey into the footer shown on each page
2. get rid of most of the remaining text

briandealwis (commented):
Thanks for cleaning this up.  I have some more suggestions 😱

docs/content/en/docs/tutorials/_index.md
gsquared94: ya, those definitely seem out of place

gsquared94: how about now?
http://34.94.56.85:1313/docs/tutorials/

moved Survey to footer and named it `Feedback`
image

gsquared94:
> I think this is an interesting one because our manually created lists don't seem to match up 100%, but the auto generated ones do. Do you think we should continue with the manual lists and try to maintain them?

@MarlonGamez updated where feasible to use auto list, PTAL.

MarlonGamez (approved):
Looks good to me!

PR #6544, opened 2021-08-30
Add per OS unit test, 'make linters' and 'make checks' per-pr tests to gh actions

aaron-prindle:
Add per OS unit test, 'make linters' and 'make checks' per-pr tests to gh actions. Step towards migrating off of travis

### DIFF
--- a/.github/workflows/integration-linux.yml
+++ b/.github/workflows/integration-linux.yml
@@ -1,4 +1,4 @@
-name: per-pr (linux)
+name: PR integration tests (linux)
 
 # Triggers the workflow on push or pull request events
 on: [push, pull_request]
@@ -8,7 +8,7 @@ permissions: read-all
 jobs:
 
   build:
-    name: PR integration tests
+    name: PR integration tests (linux)
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/linters-checks-and-unit-tests-linux.yml
+++ b/.github/workflows/linters-checks-and-unit-tests-linux.yml
@@ -1,4 +1,4 @@
-name: PR integration tests (darwin)
+name: PR linters, checks, and unit tests (linux)
 
 # Triggers the workflow on push or pull request events
 on: [push, pull_request]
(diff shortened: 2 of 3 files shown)


.github/workflows/linters-and-checks-linux.yml
```diff
@@ -0,0 +1,31 @@
+name: per-pr (linux)
```
nkubala, line 1: while we're in here, can we change the names of these tests to something a bit easier to read in the test matrix? every test starts with `per-pr` so they're kinda hard to tell apart. how about `Linters and Checks`?

.github/workflows/integration-darwin.yml
```diff
@@ -8,7 +8,7 @@ permissions: read-all
 jobs:
 
   build:
-    name: PR integration tests
+    name: PR unit and integration tests
```
nkubala, line 11: would it be difficult to make these separate jobs? two benefits we get from that:
1) unit/integration tests can run in parallel
2) unit test failure more easily distinguishable from integration test passing, and vice versa

.github/workflows/linters-and-checks-linux.yml
aaron-prindle: I'm fine w/ changing them.  The idea is that in the github UI, the name will stutter a bit with the naming suggested:
Current naming:
image

New naming:
Linters and Checks / PR linters, checks, and unit tests

.github/workflows/integration-darwin.yml
aaron-prindle: I've made this change and split them out.  Initially I thought we might want the unit tests to block the running of an integration test (thinking that if the unit tests failed the integration tests would definitely fail).  I think it makes more sense to run them in parallel though, esp with the integration tests being the time bottleneck.

.github/workflows/linters-and-checks-linux.yml
aaron-prindle: I've opted for verbose names for both fields - `name:` and `build:` so we will have the stutter but it will be easier to tell tests apart in all UIs (Actions UI, etc.)

briandealwis (approved):
LGTM: we can tweak naming in subsequent PRs

PR #6545, opened 2021-08-30
Add option to wait for connection to `/v2/events` on gRPC/HTTP server

MarlonGamez:
**Description**
This PR adds the option for the user to have skaffold pause execution on initialization of the gRPC/HTTP server. Execution will continue when a request to `/v2/events` is made. This will help ensure that users of the server can connect before Skaffold execution ends early in some cases.

This PR also moves the call to `event.SendErrorMessageOnce` to the server shutdown callback function, which ensures that the error message gets sent out before the server is shut down.

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -183,6 +183,15 @@ var flagRegistry = []Flag{
 		IsEnum:        true,
 	},
 	{
+		Name:          "wait-for-connection",
+		Usage:         "Blocks execution until the /v2/events gRPC/HTTP endpoint is hit",
+		Value:         &opts.WaitForConnection,
+		DefValue:      false,
+		FlagAddMethod: "BoolVar",
+		DefinedOn:     []string{"dev", "debug"},
+		IsEnum:        true,
+	},
+	{
 		Name:          "event-log-file",
 		Usage:         "Save Skaffold events to the provided file after skaffold has finished executing, requires --enable-rpc=true",
 		Hidden:        true,
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -188,7 +188,7 @@ var flagRegistry = []Flag{
 		Value:         &opts.WaitForConnection,
 		DefValue:      false,
 		FlagAddMethod: "BoolVar",
-		DefinedOn:     []string{"dev", "debug"},
+		DefinedOn:     []string{"dev", "build", "run", "debug", "deploy", "render", "apply", "test"},
 		IsEnum:        true,
 	},
 	{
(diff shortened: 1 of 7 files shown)


cmd/skaffold/app/cmd/flags.go
```diff
@@ -182,6 +182,15 @@ var flagRegistry = []Flag{
 		DefinedOn:     []string{"dev", "build", "run", "debug", "deploy", "render", "apply", "test"},
 		IsEnum:        true,
 	},
+	{
+		Name:          "wait-for-connection",
+		Usage:         "Blocks execution until the /v2/events gRPC/HTTP endpoint is hit",
+		Value:         &opts.WaitForConnection,
+		DefValue:      false,
+		FlagAddMethod: "BoolVar",
+		DefinedOn:     []string{"dev", "debug"},
```
aaron-prindle, line 191: Should the commands this is `DefinedOn` here be the same as those for `--enable-rpc`?  Might make sense as any command can initialize the gRPC/HTTP server and emit events

cmd/skaffold/skaffold.go
```diff
@@ -57,7 +55,7 @@ func main() {
 			// for the entire skaffold run here. It's possible SetupColors() was never called, so call it again
 			// before we print an error to get the right coloring.
 			errOut := output.SetupColors(context.Background(), os.Stderr, output.DefaultColorCode, false)
-			eventV2.SendErrorMessageOnce(constants.DevLoop, constants.SubtaskIDNone, err)
+			//eventV2.SendErrorMessageOnce(constants.DevLoop, constants.SubtaskIDNone, err)
```
aaron-prindle, line 58: remove comment?

aaron-prindle (approved):
LGTM, seems this flag could be defined on more commands and it looks like a commented out line was left (see comments)

cmd/skaffold/app/cmd/flags.go
MarlonGamez: yeah, this was definitely a brain fart of mine 😅 thanks for catching

cmd/skaffold/skaffold.go
MarlonGamez: thanks for catching this !

pkg/skaffold/event/v2/event.go
```diff
@@ -173,6 +180,9 @@ func (ev *eventHandler) forEach(listeners *[]*listener, log *[]proto.Event, lock
 }
 
 func (ev *eventHandler) forEachEvent(callback func(*proto.Event) error) error {
+	if len(handler.wait) != cap(handler.wait) {
```
gsquared94, line 183: you can run into a race condition from multiple goroutines and have a deadlock here.

MarlonGamez: Updated this to use a select statement instead

nkubala (approved):

PR #6548, opened 2021-08-31
Run CodeQL analysis nightly and cache the go modules cache

briandealwis:
The CodeQL analysis was set up to run weekly rather than nightly.  We can also speed up the build by caching the Go build cache and modules cache as per [the actions/cache docs](https://github.com/actions/cache/blob/main/examples.md#go---modules).


tejal29 (approved):
thanks for doing this!

Issue #6442, opened 2021-08-16
Multiple scrolls when displaying skaffold.yaml page using Chrome on Windows


kourtneyshort:
### Expected behavior

Page display is responsive - I can read the text without excessive scrolling

### Actual behavior

Horizontal scroll and two vertical scrolls
![image](https://user-images.githubusercontent.com/47219253/129588078-62bf338a-710c-48b7-a749-191a0df9532b.png)

### Information

- Skaffold version: N/A
- Operating system: Windows 10 Enterprise
- Installed via: N/A
- Contents of skaffold.yaml: N/A

### Steps to reproduce the behavior
1. Open https://skaffold.dev/docs/references/yaml/ in Chrome on Windows. 32" monitor is split across two windows.

PR #6559, opened 2021-09-02
docs: use dropdown menu for selecting version

gsquared94:
Fixes: #6442 
This PR replaces the huge list of versions with a dropdown menu.
## Before:
image

## After:
image

### DIFF
--- a/docs/layouts/partials/page-meta-links.html
+++ b/docs/layouts/partials/page-meta-links.html
@@ -9,9 +9,8 @@
 {{ if in .Permalink "/docs/references/yaml"}}
     <br>
     <div>Schema Versions</div>
-    <div id="schema_links">
-
-    </div>
+    <select id="schema_links" onchange="selectSchema(this)">
+    </select>
     <ul id="schema_list" hidden>
     {{ range (readDir "/content/en/schemas") }}
         <li>{{ trim .Name ".json" }}</li>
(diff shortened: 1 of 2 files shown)


ahmetb:
@gsquared94 I would approach this cautiously.
I don't think search engines will click to that dropdown and wait for JS to execute, so any sub-page that is not directly linked from an `<a href` will no longer be indexed.

Is that desirable?

gsquared94:
> @gsquared94 I would approach this cautiously.
> I don't think search engines will click to that dropdown and wait for JS to execute, so any sub-page that is not directly linked from an `<a href` will no longer be indexed.
> 
> Is that desirable?

@ahmetb  I can't say that I even considered this. I'm not too sure about what the benefits are of the different version pages being indexed. 
I definitely want to change this to a dropdown menu but must admit that I'm not very good with frontend. Is there a better way for achieving this?

ahmetb:
1. I note that we already use some sort of JS loading this skaffod.yaml page itself as well as those different pages. 

    For example, this [cache](https://webcache.googleusercontent.com/search?q=cache:i4JxLMYZ2Q4J:https://skaffold.dev/docs/references/yaml/+&cd=1&hl=en&ct=clnk&gl=us) does have no content of skaffold.yaml fields indexed (hence they cannot be used as search keywords). 

    > image


2. But with this change, you'll also lose some results from https://www.google.com/search?q=%22skaffold.yaml%22+%22version+of+the+configuration%22 as these separate pages will not be recognized by the search bots and also since they are not in /sitemap.xml

    > image

I'd say this is not making the situation worse, so go for it.

aaron-prindle (approved):
LGTM

PR #6556, opened 2021-09-01
CNCF cloudevent support for skaffold events

baleshk:
Fixes: #4259

**Description**
In this pull request, we have added support for cloudevents standards on a different route /v3/events.

PR #6567, opened 2021-09-03
Update to grpc-gateway v2

### LINKED ISSUE #5986: Update proto toolchain

briandealwis:
Fixes #5986
**Related**: #6556

**Description**
This PR separates out an upgrade of `grpc-gateway` that @baleshk incorporated into his CloudEvents PR in #6556.

#### Protobufs must be passed by-reference and not by-value

A side-effect is that I had to change our code pass protobuf objects by reference instead of by value (more below). The newer proto generation now marks protobuf objects with a [`DoNotCopy` marker](https://godoc.org/google.golang.org/protobuf/internal/pragma#DoNotCopy) to generate warnings when they are copied.  Protobuf objects are not meant to be copied by value: they include an internal field that is meant to be accessed atomic access, and [this field has been there for a while](https://github.com/golang/protobuf/issues/1155#issuecomment-668171322).

#### Swagger/OpenAPI output has changed

@baleshk noted that [error handling is removed in the newer gateway](https://github.com/GoogleContainerTools/skaffold/pull/6556#discussion_r701590870). As a result the swagger output is changed.  
The [gRPC Gateway migration guide](https://grpc-ecosystem.github.io/grpc-gateway/docs/development/grpc-gateway_v2_migration_guide/#error-handling-configuration-has-been-overhauled) notes that they now use the RPC status return.

This change is for the better:

OLD:
swagger-old

NEW:
skaffold-new

----

- Swagger API docs have changed

Filed #6600 to track follow-on work: the Swagger editor complains about our Swagger output as we're missing a version field, and we have a lot of badly named types (`protoState`) that would be nice to clear up.


hack/generate-proto.sh
```diff
@@ -16,16 +16,18 @@
 
 set -e
 
-docker build -t gen-proto -f hack/proto/Dockerfile --target generate-files proto
+docker build --platform linux/amd64 -t gen-proto -f hack/proto/Dockerfile --target generate-files proto
```
baleshk, line 19: Do we really need to mention the platform switch?

briandealwis: Oops.  I was trying to build from an M1 and it was failing.  I'll back this out.

aaron-prindle:
CI/CD failing with lint error:
```
pkg/skaffold/event/event.go:112:2: commentFormatting: put a space between `//` and comment text (gocritic)
	// fields are known to be nil'd out when empty
	^
FAILED hack/golangci-lint.sh
```

The kokoro failure is the known flake - 
```
    panic.go:617: time="2021-09-13T15:49:17Z" level=warning msg="Assuming the secret e2esecret is mounted inside Kaniko pod with the filename kaniko-secret. If your secret is mounted at different path, please specify using config key `pullSecretPath`.\nSee https://skaffold.dev/docs/references/yaml/#build-cluster-pullSecretPath" subtask=-1 task=Build
        Get "https://34.135.186.128:5000/v2/": dial tcp 34.135.186.128:5000: i/o timeout; Get "http://34.135.186.128:5000/v2/": dial tcp 34.135.186.128:5000: i/o timeout
--- FAIL: TestBuildKanikoInsecureRegistry (693.43s)
```

briandealwis:
Thanks @aaron-prindle — I left that code in to see as the integration tests were failing when cloning the state using the JSON marshaller as `State.StatusCheckState.Resources` would be nil'd out when empty.  Seems to be ok using protobuf's `Clone()` though 🎉

briandealwis:
Marking as blocked: hold off until next release.

proto/google/api/annotations.proto
```diff
@@ -0,0 +1,31 @@
+// Copyright 2015 Google LLC
```
MarlonGamez, line 1: will these files in `proto/google/api/` need to be included in the repo now?

MarlonGamez (commented):
Couple quick questions but everything looks good

proto/google/api/annotations.proto
briandealwis: grpc-gateway v2.3.0 removed the `third_party` files; [the easiest work-around is to copy them in](https://github.com/grpc-ecosystem/grpc-gateway/issues/1935#issuecomment-803572170).

pkg/skaffold/errors/err_def.go
```diff
@@ -31,39 +31,41 @@ type Error interface {
 
 type ErrDef struct {
 	err error
-	ae  proto.ActionableErr
+	ae  *proto.ActionableErr
 }
 
-func (e ErrDef) Error() string {
+var _ error = (*ErrDef)(nil)
```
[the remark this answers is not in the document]
briandealwis: It's a somewhat odd explicit compile-time check that `*ErrDef` implements the `error` interface.

pkg/skaffold/errors/errors.go
```diff
@@ -58,8 +58,11 @@ func ActionableErrV2(cfg interface{}, phase constants.Phase, err error) *protoV2
 	errCode, suggestions := getErrorCodeFromError(cfg, phase, err)
 	suggestionsV2 := make([]*protoV2.Suggestion, len(suggestions))
 	for i, suggestion := range suggestions {
-		converted := protoV2.Suggestion(*suggestion)
-		suggestionsV2[i] = &converted
+		var suggestions2 protoV2.Suggestion
+		suggestions2.Action = suggestion.Action
+		suggestions2.SuggestionCode = suggestion.SuggestionCode
+		suggestions2.Action = suggestion.Action
+		suggestionsV2[i] = &suggestions2
```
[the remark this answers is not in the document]
briandealwis: Are there other places?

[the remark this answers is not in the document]
MarlonGamez: I think there are just 2, here and in `func V2fromV1()` a bit further down in the same file

pkg/skaffold/errors/err_def.go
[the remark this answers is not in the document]
MarlonGamez: ah okay, thanks for clarification

proto/google/api/annotations.proto
MarlonGamez: sounds good, thanks for the info!

MarlonGamez (approved):
LGTM. Thanks for the clarifications @briandealwis

Issue #6564, opened 2021-09-03
--iterative-status-check broken in 1.31.0 (modules are always deployed sequentially, and not all at once)


akostic-kostile:
### Expected behavior
When doing `skaffold dev --module one,two,three --iterative-status-check=false` I would expect all modules to be deployed at once, and not sequentially.

### Actual behavior
Modules are deployed sequentially regardless of value of `--iterative-status-check`

### Information

- Skaffold version: 1.31.0 
- Operating system: MacOS Big Sur 11.5.2 
- Installed via: brew 
- Contents of skaffold.yaml:

Can't paste the yaml as it belongs to the company I work for, but the project is composed of 7 modules out of which 4 are being built and deployed, while 3 are third party, and they are only being deployed.

### Steps to reproduce the behavior

1. `skaffold dev --module one,two,three`
2. Observe the modules being deployed one after another

This is a regression bug, this worked fine in 1.30.0

aaron-prindle:
I was able to reproduce this regression using skaffold/examples/multi-config-microservices comparing v1.31.0 and v1.29.0.  I saw the same regression in v1.30.0 although the bug here outlines that this only happens in v1.31.0 so I'm not sure what might be different in the setup, etc.:

Repro steps:

See parallel deploy in v1.29.0:
```
$ cd skaffold/examples/multi-config-microservices
$ skaffold-v1.29.0 dev --module app-config,web-config --iterative-status-check=false
...<dev output>
Tags used in deployment:
 - base -> base:5463cd546c093eb8af26dc6794a339efe54d3b3a6b04e7433f4b7ab9cedf1380
 - leeroy-app -> leeroy-app:e22b15e750340ae6504e808b7d7197cbe121d315ddb0a6519ae014e40e8327e7
 - leeroy-web -> leeroy-web:e9c010ecf5fc434475fdd43f4b67ee740742d2bec68532f4001a382d82775897
Starting deploy...
 - service/leeroy-app created
 - deployment.apps/leeroy-app created
 - deployment.apps/leeroy-web created
Waiting for deployments to stabilize...
 - deployment/leeroy-app is ready. [1/2 deployment(s) still pending]
 - deployment/leeroy-web is ready.
Deployments stabilized in 2.107 seconds
```

See iterative deploy in v1.30.0 and v1.31.0:
```
```
$ skaffold-v1.30.0 dev --module app-config,web-config --iterative-status-check=false
...<dev output>
Tags used in deployment:
[...]

PR #6574, opened 2021-09-08
fix: iterative-status-check only for deployers defining hooks

gsquared94:
Fixes: #6564 

**Description**

- Enable iterative status checks for deployers that define at least one deploy hook.
- The regression is caused by enabling iterative status checks erroneously for all deployers that satisfy the hook runner interface.


aaron-prindle (approved):
LGTM

Issue #6572, opened 2021-09-07
Build hooks sometimes don't work


ahmetb:
### Expected behavior

I tried `examples/lifecycle-hooks` and tried to apply the same to `examples/getting-started`.

So modify the examples/getting-started/skaffold.yaml like this:
```diff
diff --git a/examples/getting-started/skaffold.yaml b/examples/getting-started/skaffold.yaml
index c0b13239a..038c7258d 100644
--- a/examples/getting-started/skaffold.yaml
+++ b/examples/getting-started/skaffold.yaml
@@ -3,6 +3,13 @@ kind: Config
 build:
   artifacts:
   - image: skaffold-example
+    hooks:
+      before:
+      - command: ["env"]
+        os: [darwin, linux]
+      after:
+      - command: ["env"]
+        os: [darwin, linux]
 deploy:
   kubectl:
     manifests:
```

I expect `skaffold build -v debug` logs to show stuff like:

```
Starting pre-build hooks...
DEBU[0000] Running command: [env]
```

### Actual behavior

The hook is not being executed according to logs.

### Information

- Skaffold version: v1.31.0
- Operating system: macOS
- Installed via: Brew
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta22
kind: Config
build:
  artifacts:
  - image: skaffold-example
    hooks:
      before:
      - command: ["env"]
        os: [darwin, linux]
      after:
      - command: ["env"]
        os: [darwin, linux]
deploy:
  kubectl:
    manifests:
      - k8s-*
```

### Steps to reproduce the behavior

1. apply the diff above
2. `skaffold build -v debug`
3. output doesn't show any hooks being executed

gsquared94:
Are you seeing the following loglines:
```
Checking cache...
DEBU[0003] Found dependencies for dockerfile: [{main.go /go true}]  subtask=-1 task=DevLoop
DEBU[0003] push value not present in isImageLocal(), defaulting to false because cluster.PushImages is false  subtask=-1 task=DevLoop
 - skaffold-example: Found. Tagging
```

The build hooks only run if the artifact gets built. It's possible that the artifact is being found in the skaffold cache. Can you try running:
```
skaffold build -v debug --cache-artifacts=false
```

ahmetb:
@gsquared94 that was it, I'll treat this as a docs gap and add a notice [here](https://skaffold.dev/docs/pipeline-stages/lifecycle-hooks/#before-build-and-after-build).

PR #6577, opened 2021-09-08
docs: mention hooks don't run on cached artifacts

ahmetb:
Fixes: #6572 
**Related**: #6572

### DIFF
--- a/docs/content/en/docs/pipeline-stages/lifecycle-hooks.md
+++ b/docs/content/en/docs/pipeline-stages/lifecycle-hooks.md
@@ -18,9 +18,9 @@ Host hooks are executed on the runner and can be defined for the following phase
 
 ### `before-build` and `after-build`
 
-Build hooks are executed if the build phase runs. If the resulting artifact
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/pipeline-stages/lifecycle-hooks.md
```diff
@@ -18,6 +18,10 @@ Host hooks are executed on the runner and can be defined for the following phase
 
 ### `before-build` and `after-build`
 
+Build hooks are executed if the build phase runs. If the resulting artifact
+is cached, the hooks are not executed. To test the build hooks, run Skaffold
+with `--cache-artifacts=false` option.
```
briandealwis, line 23: This text implies that build hooks are run for the _phase_, rather than for each artifact.

```suggestion
Build hooks are executed before and after each artifact is built. 
If an artifact is not built, such as happens when the image was found in the Skaffold image cache, then the build hooks will not be executed.
To force the build hooks, run Skaffold with `--cache-artifacts=false` option.
```

gsquared94 (approved):

Issue #6533, opened 2021-08-27
Persist V2 logs in a file if `--event-log-file` is present


tejal29:
Persist v2 events to log file. Add suffix `.v2` to filename so  v2 and v1 event log file can be differentiated

PR #6581, opened 2021-09-09
Persist EventV2 logs when `--event-log-file` exists

verbanicm:
Fixes: #6533 

**Description**
Added code to persist EventV2 logs in a separate log file with extension `.v2` if the `--event-log-file` flag is set


tejal29 (approved):
Thanks for the quick fix!

PR #6585, opened 2021-09-10
Updates skaffold.yaml example in helm section

gersonsosa:
**Description**
Updates the configuration example to match YAML schema, using artifacts stanza and image property


tejal29 (approved):

tejal29:
Thanks for catching this! 
@gsquared94 I remember we do make sure doc samples are verified against latest schema. 
Should we move this to `samples`?

gsquared94:
> Thanks for catching this!
> @gsquared94 I remember we do make sure doc samples are verified against latest schema.
> Should we move this to `samples`?

Filed https://github.com/GoogleContainerTools/skaffold/issues/6589

Issue #6586, opened 2021-09-10
Debug container manager must treat pod deletion as container termination


briandealwis:
@emmanuelbaah noticed that `DebuggingContainerEvent.Start`may not always be matched with a corresponding `.Terminated`.  I can see the pod container state turning to terminated when I hammer the server with `kubectl get -o yaml pod/xxxx`.  But from tracing our podWatcher events, I only see an event reporting the pod being deleted, with no intermediate events showing the container status being set to `terminated`.  I hazard that the podWatcher may coalesce events, which makes sense.

<details>
<summary>skaffold debug -vtrace</summary>
```
[server] 2021-09-09 21:38:43.894  INFO 1 --- [extShutdownHook] o.e.jetty.server.AbstractConnector       : Stopped ServerConnector@37d80fe7{HTTP/1.1, (http/1.1)}{0.0.0.0:8080}
[server] 2021-09-09 21:38:43.897  INFO 1 --- [extShutdownHook] org.eclipse.jetty.server.session         : node0 Stopped scavenging
[server] 2021-09-09 21:38:43.916  INFO 1 --- [extShutdownHook] o.e.j.s.h.ContextHandler.application     : Destroying Spring FrameworkServlet 'dispatcherServlet'
[server] 2021-09-09 21:38:43.921  INFO 1 --- [extShutdownHook] o.e.jetty.server.handler.ContextHandler  : Stopped o.s.b.w.e.j.JettyEmbeddedWebAppContext@14faa38c{application,/,[file:///tmp/jetty-docbase.8080.6384822271742410622/],STOPPED}
[server] 2021-09-09 21:38:43.937  INFO 1 --- [extShutdownHook] o.s.s.concurrent.ThreadPoolTaskExecutor  : Shutting down ExecutorService 'applicationTaskExecutor'
TRAC[0298] pod event: MODIFIED: default/java-hello-world-848ffb9cf7-27qfr: phase:Running server<waiting>   subtask=-1 task=DevLoop
TRAC[0301] was able to obtain INADDR_ANY:5005            subtask=-1 task=DevLoop
TRAC[0301] was able to obtain 127.0.0.1:5005             subtask=-1 task=DevLoop
port forwarding pod-java-hello-world-6dcdd5d77-2m9lr-default-5005 recovered on port 5005
[...]

PR #6587, opened 2021-09-10
fix: treat pod deletion as if its containers were terminated

briandealwis:
Fixes: #6586

**Description**
- Treat pod-deletions as equivalent to terminating all containers within the pod.
- Rewrite the debug container-manager tests to test different sequences of events to ensure that we handle container terminations and pod deletions equivalently


pkg/skaffold/kubernetes/watcher.go
```diff
@@ -114,6 +129,24 @@ func (w *podWatcher) Start(kubeContext string, namespaces []string) (func(), err
 					continue
 				}
 
+				if l.Logger.IsLevelEnabled(logrus.TraceLevel) {
```
gsquared94, line 132: is this check `l.Logger.IsLevelEnabled(logrus.TraceLevel)` required? Doesn't `l.Trace` already handle it? Is it to save the cost of constructing the string?

briandealwis: Yeah, probably premature optimization: I thought the podWatcher fired more events than it actually does.  This dereferencing seemed a bit distasteful but our output logger just returns a logrus object anyways.

briandealwis: `pkg/skaffold/kubernetes/watcher.go:25:2: Do not use github.com/sirupsen/logrus package, use output.log.Entry instead` 😭

gsquared94 (approved):

PR #6595, opened 2021-09-13
Ko: disable compiler optimizations and inlining when debugging

halvards:
**Description**
Disable Go compiler optimizations and inlining when building an artifact with the ko builder, and the Skaffold run mode is debug.

Use ko's existing support: https://github.com/google/ko/blob/7f145a7e1057f2f2a099827ab9d7e66a31cd1868/pkg/build/gobuild.go#L709-L712

**Tracking**: #6041
**Related**: #6563, #6569


MarlonGamez (approved):
looks good to me 👍🏼

PR #6597, opened 2021-09-13
Add image label support to ko builder

halvards:
**Description**
This enables users to label images built by the ko builder, similar to the `LABEL` instruction in `Dockerfile`s.

The Open Container Initiative Image Format Specification lists a number of [pre-defined annotation (label) keys](https://github.com/opencontainers/image-spec/blob/main/annotations.md). Some image registries use these, An example is [GitHub Container Registry](https://docs.github.com/en/packages/learn-github-packages/connecting-a-repository-to-a-package#connecting-a-repository-to-a-container-image-using-the-command-line), which uses the `org.opencontainers.image.source` label to connect a Git repository to a container image.

**Tracking:** #6041

### DIFF
--- a/pkg/skaffold/build/ko/builder.go
+++ b/pkg/skaffold/build/ko/builder.go
@@ -21,6 +21,7 @@ package ko
 // the real schema in pkg/skaffold/schema/latest/v1.
 import (
 	"context"
+	"fmt"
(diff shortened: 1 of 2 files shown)


MarlonGamez (approved):
Looks good to me!

PR #6612, opened 2021-09-16
fix: make `useBuildkit` field nullable across all config versions

### LINKED ISSUE #4936: Enable Docker BuildKit by default

gsquared94:
fix: setting `useBuildkit: false` actually turns off buildkit for docker builder

**Related**: #4936

**Description**

Currently setting `useBuildkit: false` in the `skaffold.yaml` config doesn't actually disable buildkit. This PR fixes that by setting `DOCKER_BUILDKIT =0` env variable when `useBuildkit: false` is set in the config.

In `pkg/skaffold/build/docker/docker.go `:
```diff
-	if b.useBuildKit {
-		cmd.Env = append(cmd.Env, "DOCKER_BUILDKIT=1")
+	if b.useBuildKit != nil {
+		if *b.useBuildKit {
+			cmd.Env = append(cmd.Env, "DOCKER_BUILDKIT=1")
+		} else {
+			cmd.Env = append(cmd.Env, "DOCKER_BUILDKIT=0")
+		}
	}
``` 

This change in itself would however break backwards compatibility since for previous versions of the config even if `useBuildkit` property is not set, it'd default to `false` and now start disabling buildkit which is not how it worked previously (so that bug essentially became a feature, that if `useBuildkit` is unspecified it should use the Docker default).

So we additionally make the config change to all config versions, so that a missing `useBuildkit` field can be identified separately from an explicitly set `useBuildkit: false`

There are no user facing changes other than if users were setting `useBuildkit: false` in the config, skaffold will now actually disable buildkit.

Fix #4936 after this is merged.

> Changes discussed with @briandealwis 

> The `make checks` will be failing since we generally prohibit making changes to already released configs. But this PR only changes a single field from `bool` to `*bool`.

### DIFF
--- a/pkg/skaffold/build/docker/docker.go
+++ b/pkg/skaffold/build/docker/docker.go
@@ -54,7 +54,7 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, a *latestV1.Artifact
 
 	var imageID string
 
-	if b.useCLI || b.useBuildKit {
+	if b.useCLI || (b.useBuildKit != nil && *b.useBuildKit) {
 		imageID, err = b.dockerCLIBuild(ctx, output.GetUnderlyingWriter(out), a.Workspace, dockerfile, a.ArtifactType.DockerArtifact, opts)
 	} else {
 		imageID, err = b.localDocker.Build(ctx, out, a.Workspace, a.ImageName, a.ArtifactType.DockerArtifact, opts)
@@ -91,8 +91,12 @@ func (b *Builder) dockerCLIBuild(ctx context.Context, out io.Writer, workspace s
 
 	cmd := exec.CommandContext(ctx, "docker", args...)
 	cmd.Env = append(util.OSEnviron(), b.localDocker.ExtraEnv()...)
-	if b.useBuildKit {
-		cmd.Env = append(cmd.Env, "DOCKER_BUILDKIT=1")
+	if b.useBuildKit != nil {
+		if *b.useBuildKit {
+			cmd.Env = append(cmd.Env, "DOCKER_BUILDKIT=1")
+		} else {
+			cmd.Env = append(cmd.Env, "DOCKER_BUILDKIT=0")
+		}
 	}
 	cmd.Stdout = out
 	cmd.Stderr = out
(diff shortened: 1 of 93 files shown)


pkg/skaffold/build/docker/docker.go
```diff
@@ -54,7 +54,7 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, a *latestV1.Artifact
 
 	var imageID string
 
-	if b.useCLI || b.useBuildKit {
+	if b.useCLI || (b.useBuildKit != nil && *b.useBuildKit) {
```
briandealwis, line 57: Worth a comment here to say that we ignore the useCLI setting if buildkit is requested as we currently use the Docker CLI to use buildkit. We could consider a different approach in the future.

pkg/skaffold/build/docker/docker_test.go
```diff
@@ -135,7 +135,7 @@ func TestDockerCLIBuild(t *testing.T) {
 					test.err,
 				)
 				t.Override(&util.DefaultExecCommand, mockCmd)
-			} else if test.localBuild.UseBuildkit || test.localBuild.UseDockerCLI {
+			} else if (test.localBuild.UseBuildkit != nil && *test.localBuild.UseBuildkit) || test.localBuild.UseDockerCLI {
```
briandealwis, line 138: This feels like it should be a `want` field in the test: `wantDockerCLI`.

briandealwis, line 138: This feels like it should be a `want` field in the test: `wantDockerCLI`.

briandealwis (approved):

pkg/skaffold/build/docker/docker_test.go
gsquared94: done.

gsquared94: done

nkubala (approved):

Issue #6615, opened 2021-09-17
New version failed while copying previous upgrade.go


verbanicm:
Getting errors when attempting to generate a new version

### Expected behavior

./hack/new_version.sh cleanly upgrades the schemas to a new version

### Actual behavior

Error is thrown when existing upgrade logic in upgrade.go is copied over

PR #6616, opened 2021-09-17
Fix new version generation

verbanicm:
Fixes: #6615  

**Description**
- Creates a template to use when generating a new skaffold version as opposed to copying existing upgrade files
- Uses template appropriate in `/hack/new_version.sh` script

### DIFF
--- a/hack/versions/cmd/new/templates/upgrade.template.go
+++ b/hack/versions/cmd/new/templates/upgrade.template.go
@@ -1,0 +1,38 @@
+/*
+Copyright 2020 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package %PREV_VERSION%
+
+import (
+	next "github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest/v1"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
+	pkgutil "github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+)
+
+// Upgrade upgrades a configuration to the next version.
+// Config changes from %PREV_VERSION% to %NEXT_VERSION%
+func (c *SkaffoldConfig) Upgrade() (util.VersionedConfig, error) {
+	var newConfig next.SkaffoldConfig
+	pkgutil.CloneThroughJSON(c, &newConfig)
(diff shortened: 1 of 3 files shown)


hack/versions/cmd/new/templates/upgrade.template
```diff
@@ -0,0 +1,38 @@
+/*
+Copyright 2020 The Skaffold Authors
```
briandealwis, line 2:
```suggestion
Copyright 2021 The Skaffold Authors
```

hack/versions/cmd/new/templates/upgrade_test.template
```diff
@@ -0,0 +1,201 @@
...
+
+import (
+	"testing"
+
+	next "github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest/v1"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/yaml"
+	"github.com/GoogleContainerTools/skaffold/testutil"
+)
+
+func TestUpgrade(t *testing.T) {
```
briandealwis, line 27: This is the one problem: this upgrade template should use the `expected` from the previous version.  We could extract this out with a reasonably easy sed script, I think?
```
sed -n '/expected := `apiVersion:/,/^`/p

Comment (verbanicm) [reply hack/versions/cmd/new/templates/upgrade_test.template:27]: I am using "template variables" and replacing the them with the appropriate values. I thought the expected should be the next version after the upgrade no?

The verify function should then used the expected version from the latest/v1 to validate the upgrade test. (maybe I am misunderstanding?)

File: hack/versions/cmd/new/templates/upgrade_test.template [hunk 3]
```diff
@@ -0,0 +1,201 @@
...
+	yaml := `apiVersion: skaffold/%PREV_VERSION%
```
verbanicm, line 28: See here

hack/versions/cmd/new/templates/upgrade_test.template
```diff
@@ -0,0 +1,201 @@
...
+          dockerfile: path/to/Dockerfile
+      local:
+        push: false
+    deploy:
+      kubectl:
+        manifests:
+        - k8s-*
+      kustomize: {}
+`
+	expected := `apiVersion: skaffold/%NEXT_VERSION%
```
verbanicm, line 107: See here

hack/versions/cmd/new/templates/upgrade_test.template
```diff
@@ -0,0 +1,201 @@
...
+kind: Config
+build:
+  artifacts:
+  - image: gcr.io/k8s-skaffold/skaffold-example
+    docker:
+      dockerfile: path/to/Dockerfile
+      secret:
+        id: id
+        src: /file.txt
```
briandealwis, line 116: This is no longer current, right?

hack/versions/cmd/new/templates/upgrade_test.template
briandealwis: But because this is a fixed template, future changes need to be reflected in here.  For example, the before section includes the secret-as-a-single-item:
```
      secret:
        id: id
        src: /file.txt
```

But from now on, this should be an array, right?

PR #6617, opened 2021-09-17
Ko builder: Implement file watch dependencies

halvards:
**Description**
Add support for the `dependencies` section in the ko builder config.

Use `**/*.go` as the default file watch pattern, this prevents unintended image rebuilds based on changes to files in `.git/`, `.idea/`, and `*.swap` files and similar.

For this pattern to be effective, the changes in #6605 are required.

**Tracking:** #6041
**Related:** #6605

### DIFF
--- a/pkg/skaffold/build/ko/dependencies.go
+++ b/pkg/skaffold/build/ko/dependencies.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2021 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ko
+
+// TODO(halvards)[09/14/2021]: Replace the latestV1 import path with the
+// real schema import path once the contents of ./schema has been added to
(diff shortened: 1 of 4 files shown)


pkg/skaffold/build/ko/dependencies.go
```diff
@@ -0,0 +1,45 @@
...
+	if a.Dependencies == nil || (a.Dependencies.Paths == nil && a.Dependencies.Ignore == nil) {
+		a.Dependencies = defaultKoDependencies()
+	}
+	return list.Files(workspace, a.Dependencies.Paths, a.Dependencies.Ignore)
+}
+
+// defaultKoDependencies behavior is to watch all files in the context directory
+func defaultKoDependencies() *latestV1.KoDependencies {
+	return &latestV1.KoDependencies{
+		Paths:  []string{"."},
```
tejal29, line 42: We shd stay away from having `.` as default dependency for following reasons
1) A change in `.git/*.lock` file will trigger a dev loop
2) A change in any hidden files (like .swap files or .ides) will trigger dev loop.

Can we instead add `**/*.go` as default?

tejal29 (commented):
LGTM except default dependency

pkg/skaffold/build/ko/dependencies.go
halvards: Thanks @tejal29, good point! I've pushed a new commit that uses `**/*.go` as the default. We'll need the globstar support in #6605 to make this pattern work well.

As a side note, the buildpacks and custom builders both use `.` as the default (https://github.com/GoogleContainerTools/skaffold/blob/26dc1b8936bde06692d53dabb0b6fbee78cd14ae/pkg/skaffold/schema/defaults/defaults.go#L233). For those builders we can't really list patterns, since it could be any number of languages. Should we consider a list of default excludes for those builders?

tejal29 (approved):
Thanks!

PR #6618, opened 2021-09-19
fix: remove "managed-by" fixing helm conflict (fixes #6421)

### LINKED ISSUE #6421: [Helm] "app.kubernetes.io/managed-by" must equal "Helm": current value is "skaffold"
### Expected behavior

`skaffold run` succeeds multiple times.

### Actual behavior

```
> skaffold run
...
Error: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "foobar" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: key "app.kubernetes.io/managed-by" must equal "Helm": current value is "skaffold"
deploying "fn-drupal-operator": install: exit status 1

```

It appears the underlying helm execution cannot handle the modified "app.kubernetes.io/managed-by" value for some resources. This causes multiple executions of `skaffold run` to fail in some cases.

It seems the resources it struggles with are core k8s resources in the default namespace. This is just a guess, but we have many resources that it doesn't have any issue with.

##### Workaround

```bash
kubectl api-resources --verbs=list \
  | grep 'k8s\.io' | cut -d' ' -f1 \
  | xargs -P8 -n 1 kubectl get --show-kind --ignore-not-found -l app.kubernetes.io/managed-by=skaffold -o name \
  | xargs kubectl delete
skaffold run
```

### Information
```
helm version
version.BuildInfo{Version:"v3.6.3", GitCommit:"d506314abfb5d21419df8c7e7e68012379db2354", GitTreeState:"dirty", GoVersion:"go1.16.5"}
```

- Skaffold version: v1.29.0
- Operating system: macOS 11.5
- Installed via: Homebrew
- Contents of skaffold.yaml:

TBD; too much to put here

1. a clonable repository with the sample skaffold project
2. `skaffold <command>`
3. ...

I will try to reproduce this on an example I can share. So far I haven't been able to reproduce outside of an internal project.

glennpratt:
Fixes: #6421

**Description**
remove "managed-by" fixing helm conflict

remove "managed-by"

### DIFF
--- a/integration/render_test.go
+++ b/integration/render_test.go
@@ -686,7 +686,7 @@ resources:
 kind: Pod
 metadata:
   labels:
-    app.kubernetes.io/managed-by: SOMEDYNAMICVALUE
+    skaffold.dev/run-id: SOMEDYNAMICVALUE
     this-is-from: kustomization.yaml
   name: my-pod-123
 spec:
--- a/integration/render_test.go
+++ b/integration/render_test.go
@@ -584,7 +584,6 @@ spec:
 kind: Pod
 metadata:
   labels:
-    app.kubernetes.io/managed-by: SOMEDYNAMICVALUE
     skaffold.dev/run-id: SOMEDYNAMICVALUE
   name: my-pod-123
 spec:
@@ -688,7 +687,6 @@ kind: Pod
 metadata:
   labels:
     app.kubernetes.io/managed-by: SOMEDYNAMICVALUE
-    skaffold.dev/run-id: SOMEDYNAMICVALUE
     this-is-from: kustomization.yaml
   name: my-pod-123
 spec:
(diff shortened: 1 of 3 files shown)


MarlonGamez:
hey @glennpratt , thanks for opening this. It looks good! I think there is just a gofmt error that needs to be resolved 👍🏼 I'll approve and merge once that's all good

integration/render_test.go
```diff
@@ -688,7 +687,6 @@ kind: Pod
 metadata:
   labels:
     app.kubernetes.io/managed-by: SOMEDYNAMICVALUE
-    skaffold.dev/run-id: SOMEDYNAMICVALUE
```
MarlonGamez, line 691: didn't catch this at first but looks like this is the wrong line being removed here right?

MarlonGamez (commented):
caught a small error with the updated integration test

integration/render_test.go
glennpratt: Oof, thanks for catching that. Fixed.

MarlonGamez (approved):
looks good to me 👍🏼

PR #6634, opened 2021-09-23
fix: update render help as render does not build

briandealwis:
Render doesn't build.

cc @dhodun


nkubala (approved):
we'd probably do well to also update our documentation explaining *when* render builds, because sometimes it does (`--digest-source=local`)

briandealwis:
Closing: it looks like I am completely incorrect and render may do a build:
https://github.com/GoogleContainerTools/skaffold/blob/b5c809fc0c080d43301969dbaf25495fa7da16f0/cmd/skaffold/app/cmd/render.go#L66-L78

PR #6643, opened 2021-09-28
fix: Wait for context cancel in k8s pod watcher

halvards:
**Description**
Make the Kubernetes pod watcher context aware, and select on `context.Done()`. This means we can stop waiting, and acting on, pod events when the context has been cancelled.

Remove waiting on `context.Done()` in the Kubernetes log aggregator, container manager, and pod port forwarder. This is to eliminate the chance that the pod watcher sends a `PodEvent` on a channel without a waiting receiver.

This should help avoid deadlocks that can occur when pod watcher event receivers stop reading from the channel that they've registered with the pod watcher.

We still close the channels on the receiver side, which could increase the chances of regression and re-occurrence of this issue.

Also uses a RWMutex in the pod watcher, though we could move this change to a separate PR.

Fixes: #6424

### DIFF
--- a/pkg/skaffold/kubernetes/debugging/container_manager.go
+++ b/pkg/skaffold/kubernetes/debugging/container_manager.go
@@ -67,21 +67,23 @@ func (d *ContainerManager) Start(ctx context.Context) error {
 	}
 
 	d.podWatcher.Register(d.events)
-	stopWatcher, err := d.podWatcher.Start(d.kubeContext, *d.namespaces)
+	stopWatcher, err := d.podWatcher.Start(ctx, d.kubeContext, *d.namespaces)
 	if err != nil {
 		return err
 	}
-	d.stopWatcher = stopWatcher
 
 	go func() {
 		defer stopWatcher()
-
+		l := log.Entry(ctx)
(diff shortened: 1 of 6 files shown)


halvards:
Marking as Draft so I can see the behavior on kokoro.

halvards:
Integration test fails on kokoro at a different point: https://github.com/GoogleContainerTools/skaffold/blob/151b8d373d1eafcfe559fa1b302907c893cdda1a/integration/dev_test.go#L123

Full logs: https://source.cloud.google.com/results/invocations/22b8290e-9bab-4aaf-8f68-533a32ed0ca9/targets/skaffold%2Fpresubmit/log

There was a similar failure in the integration tests running on GitHub Actions for a previous commit: https://github.com/GoogleContainerTools/skaffold/pull/6643/checks?check_run_id=3729355176
That commit had additional logging enabled for `TestDevGracefulCancel()`, which may have slowed it down sufficiently to expose the issue.

Is there more than one issue?

MarlonGamez (approved):
Looks good to me 👍🏼 Thanks @halvards !

ahmetb:
mad props! thanks for getting to the bottom of this.

Issue #6629, opened 2021-09-22
Support declarative mechanism for deploying to Cloud Run


briandealwis:
Relates to #6686
Ensure that Skaffold supports deploying Cloud Run services.

- [x] Add image replacement for identified KCC group 
- [ ] Add status check for cloud run KCC

briandealwis:
cc @ekupershlak

tejal29:
Reopening as @briandealwis mentioned we might need to follow up.

tejal29:
We opened #6709 for health check for KCC resource. closing this one.

PR #6645, opened 2021-09-28
feat: add preliminary support for Config Connector service KRM

briandealwis:
**Related**: #6629 

**Description**
Add support for Config Connector's preliminary `Service` type.  As the exact group is still TBD, add support for wildcarded group/kinds.  The wildcards are expressed using regexps to avoid taking on a dependency for a glob package.

The group should be filled in at some point once it is settled.

### DIFF
--- a/pkg/skaffold/kubernetes/manifest/visitor.go
+++ b/pkg/skaffold/kubernetes/manifest/visitor.go
@@ -18,6 +18,7 @@ package manifest
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


aaron-prindle (approved):
LGTM.  For related `Future Work`, do any docs need to be updated as well?  Might make sense to make an issue or track that somewhere.

briandealwis:
#6629 can serve that purpose.

PR #6647, opened 2021-09-29
feat: Add ko artifact type to schema

halvards:
**Description**
This change moves the ko builder config from the temporary schema location into the `v1` and `v2` schemas.

The `key` in the YAML tag for `KoArtifact` in the `ArtifactType` struct is `-`, so the config is not exposed to users.

Tracking: #6041


halvards:
Realized a comment in the schema was incorrect. Converting this PR to Draft while I fix it.

tejal29 (approved):

Issue #6273, opened 2021-07-22
kubectl should use `--dry-run=server` 


briandealwis:
This issue proposes to:
1. switch the `kubectl` deployer to use `--dry-run=server` rather than the current default of `--dry-run=client`
2. expose this dry-run setting as a flag in the `skaffold.yaml`.  

`--dry-run=server` is very useful for clusters like GKE Autopilot that perform additional validation, including erroring on settings that violate its constraints.

In case this causes deployments to fail that worked fine with `--dry-run=client`, we should provide an environment variable to restore the old behaviour like `SKAFFOLD_KUBECTL_DRY_RUN` that can be set to `client`, `server`, or `none`.

Relevant: #4094

briandealwis:
A few thoughts:
  - `--dry-run=server` may fail if the user has insufficient permissions — but an `apply` would fail then too
  - should have an option, probably in the global config, to default dry-run to client-only.

MarlonGamez:
@briandealwis so do you think we should change the default for this? Or should we first just introduce a change to have it be configurable before possibly changing the default?

briandealwis:
I think we should change the default so that users can benefit from the warnings returned by the Kubernetes API — @ahmetb [shows a nice example](https://twitter.com/ahmetb/status/1437816662212763657).

briandealwis:
Closing: we already get validation reports from the `kubectl apply`.

PR #6648, opened 2021-09-29
feat: allow user to configure manifest validation by setting deploy.kubectl.dryRun

MarlonGamez:
**Related**: #6273 

**Description**
This PR introduces a new config option that allows users to change how their manifests are validated.
Users can now configure `deploy.kubectl.dryRun` in their `skaffold.yaml`. Values can be `none`, `server`, or `client`.
This will simply change the value that is passed to `kubectl create --dry-run=`.

As per #6273, we should change the default value to `server`. I will do that in a follow up PR.


pkg/skaffold/deploy/kubectl/cli.go
```diff
@@ -205,7 +207,12 @@ func (c *CLI) ReadManifests(ctx context.Context, manifests []string) (manifest.M
 		return nil, versionGetErr(err)
 	}
 	if compTo1_18 >= 0 {
-		dryRun += "=client"
+		// Always use `client` on cleanup to avoid errors from resources already existing on cluster
```
briandealwis, line 210: This is unexpected.  Is there a pointer for more details?  Just some resources?

pkg/skaffold/deploy/kubectl/cli.go
```diff
@@ -192,8 +194,8 @@ func (c *CLI) WaitForDeletions(ctx context.Context, out io.Writer, manifests man
 	}
 }
 
-// ReadManifests reads a list of manifests in yaml format.
-func (c *CLI) ReadManifests(ctx context.Context, manifests []string) (manifest.ManifestList, error) {
+// ReadManifests reads a list of manifests in yaml format and validates them by running `kubectl create --dry-run` .
+func (c *CLI) ReadManifests(ctx context.Context, manifests []string, cleanup bool) (manifest.ManifestList, error) {
```
briandealwis, line 198: Could you document the meaning of cleanup too.

pkg/skaffold/deploy/kubectl/kubectl_test.go
```diff
@@ -219,6 +219,56 @@ func TestKubectlDeploy(t *testing.T) {
 			shouldErr:        true,
 			waitForDeletions: true,
 		},
+		{
```
briandealwis, line 222: Should include a test for `none` too

pkg/skaffold/deploy/kubectl/cli.go
```diff
@@ -62,12 +63,13 @@ type Config interface {
 	PipelineForImage(imageName string) (latestV1.Pipeline, bool)
 }
 
-func NewCLI(cfg Config, flags latestV1.KubectlFlags, defaultNamespace string) CLI {
+func NewCLI(cfg Config, flags latestV1.KubectlFlags, defaultNamespace string, dryRun string) CLI {
```
briandealwis, line 66: Add some doc here.  If dryRun == "" then uses "client".

pkg/skaffold/schema/defaults/defaults.go
```diff
@@ -208,6 +209,12 @@ func setDefaultLogsConfig(c *latestV1.SkaffoldConfig) {
 	}
 }
 
+func setDefaultKubectlDryRun(c *latestV1.SkaffoldConfig) {
+	if c.Deploy.KubectlDeploy != nil && c.Deploy.KubectlDeploy.DryRun == "" {
+		c.Deploy.KubectlDeploy.DryRun = "client"
+	}
+}
```
briandealwis, line 216: Do we need this? You're already handling `""` in the CLI code itself, and it provides false comfort since you're also passing in `""` explicitly from the Kustomize code.

pkg/skaffold/schema/latest/v1/config.go
```diff
@@ -568,6 +568,13 @@ type KubectlDeploy struct {
 	// DefaultNamespace is the default namespace passed to kubectl on deployment if no other override is given.
 	DefaultNamespace *string `yaml:"defaultNamespace,omitempty"`
 
+	// DryRun is the value kubectl will use with `kubectl create --dry-run=`. Valid values are:
+	// `none`: validate manifests using `kubectl create --dry-run=none`.
+	// `server`: validate manifests using `kubectl create --dry-run=server`.
+	// `client` (default): validate manifests using `kubectl create --dry-run=client`.
```
briandealwis, line 574: I think it makes sense to order the default first, and place `none` last?

pkg/skaffold/schema/latest/v1/config.go
```diff
@@ -568,6 +568,13 @@ type KubectlDeploy struct {
 	// DefaultNamespace is the default namespace passed to kubectl on deployment if no other override is given.
 	DefaultNamespace *string `yaml:"defaultNamespace,omitempty"`
 
+	// DryRun is the value kubectl will use with `kubectl create --dry-run=`. Valid values are:
```
briandealwis, line 571: Would you mind fixing the schema enum collector to support an optional trailing `:`?
https://github.com/GoogleContainerTools/skaffold/blob/151b8d373d1eafcfe559fa1b302907c893cdda1a/hack/schemas/main.go#L52

docs/content/en/schemas/v2beta24.json
```diff
@@ -2586,6 +2586,12 @@
           "description": "default namespace passed to kubectl on deployment if no other override is given.",
           "x-intellij-html-description": "default namespace passed to kubectl on deployment if no other override is given."
         },
+        "dryRun": {
+          "type": "string",
+          "description": "value kubectl will use with `kubectl create --dry-run=`. Valid values are: `none`: validate manifests using `kubectl create --dry-run=none`. `server`: validate manifests using `kubectl create --dry-run=server`. `client` (default): validate manifests using `kubectl create --dry-run=client`.",
+          "x-intellij-html-description": "value kubectl will use with <code>kubectl create --dry-run=</code>. Valid values are: <code>none</code>: validate manifests using <code>kubectl create --dry-run=none</code>. <code>server</code>: validate manifests using <code>kubectl create --dry-run=server</code>. <code>client</code> (default): validate manifests using <code>kubectl create --dry-run=client</code>.",
+          "default": "client"
```
briandealwis, line 2593: Hmm the enum values should be showing up here.

briandealwis (approved):
A few nits

pkg/skaffold/schema/latest/v1/config.go
briandealwis: Maybe we should do this separately as it looks like there are a bunch of similar enums that aren't picked up.

pkg/skaffold/deploy/kubectl/cli.go
MarlonGamez: [This stack overflow answer](https://stackoverflow.com/questions/47369351/kubectl-apply-vs-kubectl-create) helped me understand what I was seeing here a bit more.

Basically since we are using `kubectl create` to validate the manifests, in the case of us cleaning up, since we validate in this way, if we validate using `--dry-run=server` then we'll receive an error saying the resources already exist on cluster.

pkg/skaffold/deploy/kubectl/cli.go
MarlonGamez: Will do. I'll also move the empty string check to here rather than the ReadManifests function. That way it's already set to "client" in the case that the value ever gets read anywhere else.

pkg/skaffold/schema/defaults/defaults.go
MarlonGamez: that's true, it doesn't look like this is needed. I'll remove this

pkg/skaffold/schema/latest/v1/config.go
MarlonGamez: yup, I've fixed for this here for now by removing the `:`. I'll open up another PR to allow a trailing `:`

MarlonGamez: https://github.com/GoogleContainerTools/skaffold/pull/6651

pkg/skaffold/deploy/kubectl/cli.go
briandealwis: Oh! I didn't realize we ran this as `kubectl create`.  I wonder why.

MarlonGamez:
Holding off on merging this change as we're awaiting more discussion on the topic.
CC @sivakku @briandealwis

MarlonGamez:
closing this as we've decided that it's unneeded at the moment

PR #6649, opened 2021-09-29
doc: document Helm deployer's IMAGE_NAME<N>, IMAGE_TAG<N>, IMAGE_DIGEST<N>

snickell:
I can't find any documentation (other than references in issues by skaffold devs) to these. Without this, its very hard to figure out how to use the `IMAGE_NAME` template etc for more than a single artifact.

### DIFF
--- a/docs/content/en/docs/environment/templating.md
+++ b/docs/content/en/docs/environment/templating.md
@@ -31,4 +31,6 @@ _Please note, this list is not exhaustive._
 List of variables that are available for templating:
 
 * all environment variables passed to the Skaffold process at startup
-* `IMAGE_NAME` - the artifacts' image name - the [image name rewriting]({{< relref "/docs/environment/image-registries.md" >}}) acts after the template is calculated
+* `IMAGE_NAME`, `IMAGE_TAG`, `IMAGE_DIGEST` - the first (by order of declaration in `build.artifacts`) artifacts' image name, tag, and sha256 digest. Note: the [image name rewriting]({{< relref "/docs/environment/image-registries.md" >}}) acts after the template is calculated.
+* `IMAGE_NAME2`, `IMAGE_TAG2`, `IMAGE_DIGEST2` - the 2nd artifacts' image name, tag, and sha256 digest
+* `IMAGE_NAMEN`, `IMAGE_TAGN`, `IMAGE_DIGESTN` - the Nth artifacts' image name, tag, and sha256 digest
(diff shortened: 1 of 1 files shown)


snickell:
@googlebot I signed it!

briandealwis (commented):
Thanks @snickell. May I ask how you use these values?

From what I can tell, these `IMAGE_XXX<N>` values are [only supported for the Helm deployer](https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/deploy/helm/args.go#L68-L77).

The custom builder supports `IMAGE_NAME` and `IMAGE_TAG`, but doesn't support the `IMAGE_DIGEST`.  And this isn't documented.

The Kaniko builder supports `IMAGE_REPO`, `IMAGE_NAME`, and `IMAGE_TAG`, but the _REPO and _NAME are different.

briandealwis (approved):

snickell:
@briandealwis 

TL;DR While the `helmStrategy` does reflect the conventional recommendation for how container image references be formatted as a `values.yaml` JSON tree, not all upstream charts use that recommendation, making it impossible to override their values with skaffold.

Basically, in our case, an upstream helm chart (JupyterHub: https://github.com/jupyterhub/zero-to-jupyterhub-k8s) required build artifacts to be inserted into the helm `values.yaml` JSON tree in a format that was both strict (they enabled JSON schema validation, which in turn prevents this from being just a warning but there's now no way to deploy our system with skaffold until fixed), and not producable  by skaffold's available `imageStrategy`s.

By explicitly using `gitCommit: { variant: AbbrevCommitSha }` we can use `setValueTemplates` to construct and inject the value JSON tree in the shape required by the upstream z2jh chart:
```
      setValueTemplates:
        # Why not use artifactOverrides? We have to match name/tag value format required by the jupyterhub chart
        jupyterhub.singleuser.image.name: "{{.IMAGE_NAME}}"
        jupyterhub.singleuser.image.tag: "{{.IMAGE_TAG}}@{{.IMAGE_DIGEST}}"
        jupyterhub.hub.image.name: "{{.IMAGE_NAME2}}"
        jupyterhub.hub.image.tag: "{{.IMAGE_TAG2}}@{{.IMAGE_DIGEST2}}"
```

Our preference, not surprisingly, would have been for `buildArtifacts` to itself have the option to output in the format expected by our upstream chart (which we can't change), perhaps some sort of JSON template system, so we could work with ANY required build artifacts values shape/format. Without that, this workaround was required to continue using skaffold to build our with our chart.

briandealwis:
@snickell the tracking issue is #6706.

PR #6664, opened 2021-10-01
feat: Add distinct error codes for GCB failures

tejal29:
As I was debugging internal issue where specifying a `workerpool` in google could build section does not create a GCB build, I noticed we did not have distinct error codes for GCB failures.


pkg/skaffold/build/gcb/cloud_build.go
```diff
@@ -115,30 +136,46 @@ func (b *Builder) buildArtifactWithCloudBuild(ctx context.Context, out io.Writer
...
+			ErrCode: proto.StatusCode_BUILD_GCB_UPLOAD_TO_GCS_ERR,
+			Message: fmt.Sprintf("uploading source archive: %s", err),
+		})
 	}
 
 	buildSpec, err := b.buildSpec(ctx, artifact, tag, cbBucket, buildObject)
 	if err != nil {
-		return "", fmt.Errorf("could not create build description: %w", err)
+		return "", sErrors.NewErrorWithStatusCode(&proto.ActionableErr{
+			ErrCode: proto.StatusCode_BUILD_GCB_GET_BUILD_DESCRIPTOR_ERR,
```
briandealwis, line 157: I wonder if this might be better called GENERATE rather than GET (but avoiding CREATE due to BUILD_GCB_CREATE_BUILD_ERR below)

briandealwis (approved):

Issue #6203, opened 2021-07-14
`skaffold init --generate-manifests` failing to generate manifests for Jib project


briandealwis:
### Actual behavior

```
$ cd examples/jib
$ rm -rf k8s skaffold.yaml
$ skaffold init --generate-manifests
? Which builders would you like to create kubernetes resources for?  [Use arrows to move, space to select, <right> to all, <left> to none, type to filter]
> [ ]  Buildpacks (pom.xml)
  [ ]  Jib Maven Plugin (org.skaffold:hello-spring-boot, pom.xml)
<select Jib Maven Plugin>
one or more valid Kubernetes manifests are required to run skaffold
```

### Expected behavior

Expected to have Kubernetes manifests generated.

Workaround: `skaffold init --generate-manifests --XXenableBuildpacksInit=false` is successful.

It looks like this logic in `defaultBuildInitializer.resolveBuilderImages()` to call `generateArtifactInfos()` needs to happen regardless of the number of builders.
https://github.com/GoogleContainerTools/skaffold/blob/7b18b11ba98ea0e48c0dc500d0b8f48e94867978/pkg/skaffold/initializer/build/resolve.go#L38-L44

### Information

- Skaffold version: HEAD; reported on Slack with 1.27.0
- Operating system: darwin

MarlonGamez:
@briandealwis I think the issue here is that hitting enter without selecting any options will cause this to happen. Space needs to be used to select, then enter to move forward, which is a bit of confusing UX.

sivakku:
i used up & down arrow to choose the builder option and see that one of the builder is picked but it didn't seem to register and as a result it failed.

``` $ '/Users/sivakk/Library/Application Support/cloud-code/installer/google-cloud-sdk/bin/skaffold' init --generate-manifests
? Which builders would you like to create kubernetes resources for?                                                                  one or more valid Kubernetes manifests are required to run skaffold
sivakk@sivakk-macbookpro ~/dev/cloudcode-projects/java-server-demos/jboss/wildfly/helloworld (main) $ '/Users/sivakk/Library/Application Support/cloud-code/installer/google-cloud-sdk/bin/skaffold' init --generate-manifests
? Which builders would you like to create kubernetes resources for?                                                                  one or more valid Kubernetes manifests are required to run skaffold ```

MarlonGamez:
@sivakku did you use `space` to select the option that you wanted to generate for?
When unselected, it looks like so:
[...]

PR #6665, opened 2021-10-01
fix: change init builder-prompt to a single-select

briandealwis:
Fixes: #6203

**Description**
Turn the builder selection prompt into a single-select to avoid puzzling failure on empty-selection (the default) as well as the puzzling result from selecting multiple builders.


tejal29 (approved):

briandealwis:
We need to revert this as it doesn't handle multiple images:
Screen Shot 2021-10-04 at 12 17 20 PM

PR #6755, opened 2021-10-21
ci: add cancel-workflow-action functionality to all github workflows

### LINKED ISSUE #6745: Use the cancel-workflow-action to prevent previous workflows from running.
As per internal docs, we should be cancelling any previous run to prevent from multiple actions for the same PR when dev make commits in quick succession.

Please follow this https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre/67939898#67939898 and them to all the workflows

Assigning this to you @aaron-prindle since you are primary on call

aaron-prindle:
fixes #6745

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -10,6 +10,10 @@ on:
 
 permissions: read-all
 
+concurrency:
+  group: build-${{ github.event.pull_request.number || github.ref }}-${{github.workflow}}
+  cancel-in-progress: true
+
 jobs:
   analyze:
     name: Analyze
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -7,6 +7,10 @@ on:
 
 permissions: read-all
 
+concurrency:
+  group: build-${{ github.event.pull_request.number || github.ref }}-${{github.workflow}}
+  cancel-in-progress: true
+
 jobs:
   check:
     name: Check if release commit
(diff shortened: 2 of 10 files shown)


briandealwis (dismissed):

MarlonGamez:
@aaron-prindle is there a reason these changes would make the CI not run? or is that just a coincidence

tejal29 (approved):

briandealwis:
Uh oh, I don't think this can work as-is.  The [`concurrency` docs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency) say:
> Concurrency ensures that only a single job or workflow using the same concurrency group will run at a time. A concurrency group can be any string or expression. The expression can only use the github context. For more information about expressions, see "Expressions."
> ...
> When a concurrent job or workflow is queued, if another job or workflow using the same concurrency group in the repository is in progress, the queued job or workflow will be pending. Any previously pending job or workflow in the concurrency group will be canceled. To also cancel any currently running job or workflow in the same concurrency group, specify cancel-in-progress: true.

Since we're using the same string for each workflow here, only one will run: the others will be cancelled or made pending.

I think we might be able to append `${{github.workflow}}`?  The doc for concurrency doesn't say that it needs to have no whitespaces:
```
concurrency:
  group: build-${{ github.event.pull_request.number || github.ref }}-${{github.workflow}}
  cancel-in-progress: true
```

aaron-prindle:
Made suggested change, seems like it is working now.  Merged this in my branch to make sure it will take, CI is passing  with this change merged on a test pr - https://github.com/aaron-prindle/skaffold/pull/87

PR #6767, opened 2021-10-25
fix(helm): handle templated namespaces consistently

### LINKED ISSUE #6749: Skaffold dev with a templated helm namespace fails while querying a non-existent namespace
### Expected behavior
When I use a scaffold to deploy to a specific templated namespace, I expect that I will require permission to access resources in that namespace only and that the deploy will succeed
### Actual behavior
Skaffold tries to access resources in a namespace as if it's not templated and if I don't have permission to do so, it fails completely.

### Information

- Skaffold version: v1.33.0
- Operating system: MacOS Big Sur
- Installed via: Homebrew
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta13
kind: Config
deploy:
  helm:
    releases:
      - name: test
        remoteChart: bitnami/nginx
        namespace: "dev-{{ .SKAFFOLD_USERNAME }}"
```

1. Clone a repo https://github.com/vladimir-avinkin/skaffold-bug
2. Create a kubernetes namespace dev-123
3. Create a kubernetes user or rolebinding for an existing user with privileges only for the dev-123 namespace
2. run `TEST_NAMESPACE=123 skaffold dev` as a user described in step 3
3. See the error which looks like

```
...
Cleaning up...
release "test" uninstalled
could not fetch deployments: could not fetch deployments: deployments.apps is forbidden: User "redacted" cannot list resource "deployments" in API group "apps" in the namespace "dev-{{ .TEST_NAMESPACE }}"
```

Usually, if the user is running scaffold as a superuser it does not result in a bug, because most of the calls involved in a bug are kubernetes List requests, which just return an empty list which skaffold iterates over, so. No resources in mangled namespaces - no problems

*Below are my speculation, but I'm fairly sure I'm right*

I think the problem is that we grab the namespaces before we template them here:
https://github.com/GoogleContainerTools/skaffold/blob/f47aa416e09287c30cfaae2cbf07c36dc267740d/pkg/skaffold/deploy/helm/deploy.go#L152
and them pass them to a statusMonitor here:
https://github.com/GoogleContainerTools/skaffold/blob/f47aa416e09287c30cfaae2cbf07c36dc267740d/pkg/skaffold/deploy/helm/deploy.go#L165

And then wrong namespaces end up here in *s.namespaces:
https://github.com/GoogleContainerTools/skaffold/blob/f47aa416e09287c30cfaae2cbf07c36dc267740d/pkg/skaffold/kubernetes/status/status_check.go#L145

Which is fine if you are a super user, but skaffold fails if you don't have a right to access resources in non-existent namespaces.

Naive approach to fix it seemed to work for me, all unit tests also seemed to pass
```
diff --git a/pkg/skaffold/deploy/util/namespaces.go b/pkg/skaffold/deploy/util/namespaces.go
index 83af160..3bcaaf2 100644
--- a/pkg/skaffold/deploy/util/namespaces.go
+++ b/pkg/skaffold/deploy/util/namespaces.go
@@ -18,6 +18,7 @@ package util

 import (
        "fmt"
+       "github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
        "sort"

        kubectx "github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/context"
@@ -69,7 +70,9 @@ func collectHelmReleasesNamespaces(pipelines []latestV1.Pipeline) []string {
                if cfg.Deploy.HelmDeploy != nil {
                        for _, release := range cfg.Deploy.HelmDeploy.Releases {
                                if release.Namespace != "" {
-                                       namespaces = append(namespaces, release.Namespace)
+                                       if templatedNamespace, err := util.ExpandEnvTemplateOrFail(release.Namespace, nil); err != nil {
+                                               namespaces = append(namespaces, templatedNamespace)
+                                       }
                                }
                        }
                }
```

aaron-prindle:
fixes #6749 

Currently if there is a templated namespace with helm that does not currently exist in the cluster, skaffold will fail even though a regular deploy would succeed.  As mentioned in #6749, this is because currently skaffold fetches the namespaces before we template which results in this failure.  The fix here is to template the helm namespaces before fetching them.

### DIFF
--- a/pkg/skaffold/deploy/util/namespaces.go
+++ b/pkg/skaffold/deploy/util/namespaces.go
@@ -22,6 +22,7 @@ import (
 
 	kubectx "github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/context"
 	latestV1 "github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest/v1"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 )
 
 // GetAllPodNamespaces lists the namespaces that should be watched.
@@ -49,7 +50,11 @@ func GetAllPodNamespaces(configNamespace string, pipelines []latestV1.Pipeline) 
 	}
 
 	// Set additional namespaces each helm release referenced
-	for _, namespace := range collectHelmReleasesNamespaces(pipelines) {
+	helmReleasesNamespaces, err := collectHelmReleasesNamespaces(pipelines)
+	if err != nil {
+		return nil, fmt.Errorf("collecting helm releases namespaces: %w", err)
+	}
+	for _, namespace := range helmReleasesNamespaces {
 		nsMap[namespace] = true
 	}
 
@@ -63,16 +68,20 @@ func GetAllPodNamespaces(configNamespace string, pipelines []latestV1.Pipeline) 
 	return namespaces, nil
 }
 
-func collectHelmReleasesNamespaces(pipelines []latestV1.Pipeline) []string {
+func collectHelmReleasesNamespaces(pipelines []latestV1.Pipeline) ([]string, error) {
 	var namespaces []string
 	for _, cfg := range pipelines {
 		if cfg.Deploy.HelmDeploy != nil {
 			for _, release := range cfg.Deploy.HelmDeploy.Releases {
 				if release.Namespace != "" {
-					namespaces = append(namespaces, release.Namespace)
+					templatedNamespace, err := util.ExpandEnvTemplateOrFail(release.Namespace, nil)
+					if err != nil {
+						return []string{}, fmt.Errorf("cannot parse the release namespace template: %w", err)
+					}
+					namespaces = append(namespaces, templatedNamespace)
 				}
 			}
 		}
 	}
-	return namespaces
+	return namespaces, nil
 }
--- a/pkg/skaffold/deploy/util/namespaces_test.go
+++ b/pkg/skaffold/deploy/util/namespaces_test.go
@@ -1,0 +1,96 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"testing"
+
+	latestV1 "github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest/v1"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
+	"github.com/GoogleContainerTools/skaffold/testutil"
+)
+
+func TestCollectHelmReleasesNamespaces(t *testing.T) {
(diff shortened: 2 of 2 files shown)


tejal29:
> Is there a way to test this?

Either some unit tests or manual testing notes could help.

pkg/skaffold/deploy/util/namespaces.go
```diff
@@ -69,7 +70,11 @@ func collectHelmReleasesNamespaces(pipelines []latestV1.Pipeline) []string {
 		if cfg.Deploy.HelmDeploy != nil {
 			for _, release := range cfg.Deploy.HelmDeploy.Releases {
 				if release.Namespace != "" {
-					namespaces = append(namespaces, release.Namespace)
```
briandealwis, line 72: This should use the same approach as in `releaseNamespace()`, and it should surface any errors back up.
https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/deploy/helm/util.go#L159-L164

briandealwis:
Some unit tests of `collectHelmReleasesNamespaces()` looks easy enough?

aaron-prindle:
I've added the suggested changes and unit test for `collectHelmReleasesNamespaces`.  I'm a bit confused though as to the suggested code change made.  The other referenced logic at https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/deploy/helm/util.go#L159-L164 is referencing both the helm deployer namespace and the helm release namespace with casing and an error case.  I just wanted to verify that the logic currently:
```
			for _, release := range cfg.Deploy.HelmDeploy.Releases {
				if release.Namespace != "" {
					templatedNamespace, err := util.ExpandEnvTemplateOrFail(release.Namespace, nil)
					if err != nil {
						return []string{}, fmt.Errorf("cannot parse the release namespace template: %w", err)
					}
					namespaces = append(namespaces, templatedNamespace)
				}
			}
```

There was a suggested edit made that added:
```
                                       else {
						namespaces = append(namespaces, release.Namespace)
					}	
```
but I'm a bit confused as this is what I believe should be an error case.  Thanks.

briandealwis:
I was wrong with my comment.  It happens more often than I'd like :-)

The important thing is that we treat inputs uniformly.  It would seem to make sense to have `collectHelmReleasesNamespaces()` and [`Deployer.releaseNamespace()`](https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/deploy/helm/util.go#L156) using the same function?

aaron-prindle:
I'm not sure I understand, I believe the logic for namespace gathering from pipelines done in `collectcollectHelmReleasesNamespaces` is fundamentally different than the logic for `Deployer.releaseNamespace`.  Reverting to the deployer namespace is only done in `Deployer.releaseNamespace`.  IIUC, the only shared logic would be something like:
```
func expandNamespace(ns string) (string, error) {
	expandedNs, err := util.ExpandEnvTemplateOrFail(r.Namespace, nil)
	if err != nil {
		return "", fmt.Errorf("cannot parse the release namespace template: %w", err)
	}
	return expandedNs, nil
}
```

which is essentially `util.ExpandEnvTemplateOrFail`

aaron-prindle:
Is there anything else to be done here?  Currently I believe the implementation and tests are sufficient, was there any concern around better re-use of logic for `Deployer.releaseNamespace` and `collectHelmReleasesNamespaces`?  From my comment above, I don't believe there is a worthwhile amount of logic to reuse but I may not understand correctly.

briandealwis:
We have `util.Expand*` calls in a few places in the code now and pulling it into a single place might reduce the likelihood of errors, like this issue, in the future.  I mean, who would expect to find Helm-specific code outside of `deploy/helm`?

briandealwis (approved):

PR #6782, opened 2021-10-30
Add Labels to Metadata

jlewi:
**Description**
* Add labels to metadata so that skaffold files can have labels identifying
  the config.

* Using labels is useful for being able to match skaffold files.
* Concretely, we want to use labels to identify the different environments
  corresponding to a skaffold file (e.g. dev/prod etc...). These labels
  our used by our CD tooling to identify the skaffold files which need to
  be built.

After this change users can include labels in their skaffold.yaml files; e.g.

```
kind: Config
metadata:
  name: some config
  labels:
     dev: "true"
```

### DIFF
--- a/pkg/skaffold/schema/latest/v2/config.go
+++ b/pkg/skaffold/schema/latest/v2/config.go
@@ -62,6 +62,13 @@ type SkaffoldConfig struct {
 type Metadata struct {
 	// Name is an identifier for the project.
 	Name string `yaml:"name,omitempty"`
+
+	// Labels is a map of labels identifying the project.
+	Labels map[string]string `yaml:"labels,omitempty"`
+
+	// Annotations is a map of annotations providing additional
+	// metadata about the project.
+	Annotations map[string]string `yaml:"annotations,omitempty"`
 }
 
 // Pipeline describes a Skaffold pipeline.


jlewi:
Looks like the linter is failing because I updated v2beta24

```
Invalid changes:
----------------
diff --git a/pkg/skaffold/schema/v2beta24/config.go b/pkg/skaffold/schema/v2beta24/config.go
time="2021-10-30T22:51:36Z" level=fatal msg="structural changes detected"
index dcb198ef0..1eea3ba3a 100755
--- a/pkg/skaffold/schema/v2beta24/config.go
+++ b/pkg/skaffold/schema/v2beta24/config.go
@@ -58,6 +58,9 @@ type SkaffoldConfig struct {
 type Metadata struct {
 	// Name is an identifier for the project.
 	Name string `yaml:"name,omitempty"`
+
+	// Labels is a map of labels identifying the project.
+	Labels map[string]string `yaml:"labels,omitempty"`
 }
 
 // Pipeline describes a Skaffold pipeline.
exit status 1
FAILED hack/check-schema-changes.sh
```

I would expect this to be a backwards compatible change since its purely additive. 

What would be the correct way to introduce this change? Should I create a new version?

aaron-prindle:
@jlewi thanks for the submission here.  Yes, creating a new schema version would be the correct thing to do in this case

briandealwis:
As a rule, we don't make changes to previously published schema versions.

jlewi:
Apologies; I should have checked the development guide first.
https://github.com/GoogleContainerTools/skaffold/blob/main/DEVELOPMENT.md#making-a-config-change

Might take me a couple days to follow up on this. Feel free to mark this PR as draft or wip.

jlewi:
This updated. PTAL.

aaron-prindle (approved):
LGTM.  Thanks adding this

PR #6792, opened 2021-11-02
docs(ko): Document the ko builder

halvards:
**Description**
User documentation for the ko builder.

**User facing changes**
None. The site builder ignores the file until we are ready to release the ko builder feature.

Tracking: #6041

### DIFF
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -38,6 +38,9 @@ pygmentsStyle = "tango"
 # Enable Emojis
 enableEmoji = true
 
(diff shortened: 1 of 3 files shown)


MarlonGamez (approved):
This looks good to me. Mostly did a scan for typos/grammar, but the content itself is good as well ✅

Since we can't use the docs-preview label here, if anyone else would like to check these changes out, you'll have to checkout this PR locally and use `make preview-docs` after commenting out the `ignoreFiles = ['ko.md']` line

PR #6798, opened 2021-11-04
docs(ko): Templating in `flags` and `ldflags`

halvards:
Document how environment variable templating of `flags` and `ldflags` is handled by `ko` (rather than by Skaffold).

Keeping this consistent with `ko` simplifies the transition to Skaffold for existing `ko` users. The implementation in `ko` also matches that of GoReleaser.

Note that this is inconsistent with other templating support in Skaffold.

Also clarify the ko builder support for the `.ko.yaml` config file.

Tracking: #6041


tejal29 (approved):

tejal29:
hmm I can't see the Ko docs here http://34.94.106.191:1313/docs/pipeline-stages/builders/

PR #6815, opened 2021-11-09
fix(log): Send Go std `log` to `logrus`, and output `ggcr` logs

halvards:
This change directs log messages sent to Go's standard `log` package to `logrus` instead.

Also, previously, Skaffold discarded log messages from [`go-containerregistry`](https://github.com/google/go-containerregistry) (a.k.a. `ggcr`).
This change enables `ggcr` log messages, with severity tied to Skaffold's log level.

Related: #6041

### DIFF
--- a/pkg/skaffold/output/log/log.go
+++ b/pkg/skaffold/output/log/log.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	stdlog "log"
 
 	ggcrlogs "github.com/google/go-containerregistry/pkg/logs"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


briandealwis (commented):
LGTM. @MarlonGamez any concerns?

halvards:
The unit test failure is puzzling, I haven't been able to reproduce it locally.

halvards:
I have an idea for how to improve this to cover other libraries that log to Go's standard `log` package (e.g., `ko`), so I'll update this PR to cover that.

Marking as Draft in the meantime.

pkg/skaffold/output/log/log.go
```diff
@@ -96,5 +98,41 @@ func SetupLogs(stdErr io.Writer, level string, timestamp bool, hook logrus.Hook)
...
+		// Discard logs if we're not interested.
+		stdLogger.SetOutput(io.Discard)
+	}
+}
+
+// setupGGCRLogging enables go-containerregistry logging, mapping its levels to logrus.
+// The mapping is:
+// - ggcr Warn -> logrus Error
+// - ggcr Progress -> logrus Info
+// - ggcr Debug -> logrus Trace
```
MarlonGamez, line 122: Is there a reason we send `Warn` to `logrus.Error` rather than to `logrus.Warn`, and `Debug` to `logrus.Trace` rather than `logrus.Debug`?

MarlonGamez (commented):
Sorry, I know this is draft but just had a QQ

pkg/skaffold/output/log/log.go
halvards: I'm happy to change if we prefer.

My thinking was as follows:
- `ggcr` defines `Warn` as "non-fatal errors": https://github.com/google/go-containerregistry/blob/main/pkg/logs/logs.go#L24
- `ggcr` `Debug` logging is _very_ verbose, it includes all HTTP requests and responses, including non-sensitive headers and full payload.

halvards: `s/full payload/full payload except for image layers/`

MarlonGamez: that makes sense to me! Just wanted some clarification as it wasn't immediately obvious to me

halvards:
Ok, this one is (finally) ready.

tejal29:
I just want to make sure, this works well with Cloud code output logging.  Lets try to test this PR from cloud code side in tomorrow's bug bash.

halvards:
Just rebased the commits.

tejal29 (approved):

PR #6654, opened 2021-09-30
feat(ko): Add `KO` enum value to `BuilderType`

halvards:
**Description**
Make the Skaffold API aware of the ko builder.

Tracking: #6041

PR #6820, opened 2021-11-09
feat(ko): Enable the ko builder in the API

halvards:
Tracking: #6041
Obsolete: #6654


MarlonGamez (changes requested):
Not much to say here, looks good!

Can we add a little piece of code here to make sure that ko built artifacts get accounted for in our metadata events?
https://github.com/GoogleContainerTools/skaffold/blob/ea2b8b576112e49fc066291c100228ce61b4588a/pkg/skaffold/event/v2/metadata.go#L83-L111

halvards:
Thanks @MarlonGamez, I was a little quick on this one! Have pushed an update.

MarlonGamez (approved):
LGTM ✅ Thanks for addressing the metadata changes

PR #6821, opened 2021-11-09
fix(ko): Fall back to build configs in `.ko.yaml`

halvards:
ko build configs are optional, and can either be specified in `skaffold.yaml` or [`.ko.yaml`](https://github.com/google/ko#overriding-go-build-settings).

Prior to this change, the Skaffold ko builder would ignore any build configs present in `.ko.yaml`. This happened because we always passed a build config map with one entry, corresponding to the artifact being built.

With this change, build configs from `.ko.yaml` will be used as a fallback if the artifact config in `skaffold.yaml` does not specify any of the fields that are part of the build config. This is helpful to existing `ko` users who want to adopt Skaffold.

We achieve the fallback behavior by passing an empty build config map, which hits this conditional in `ko`: https://github.com/google/ko/blob/0015a815375d456baed846579ec02cf936b99307/pkg/commands/options/build.go#L120

Tracking: #6041

### DIFF
--- a/docs/content/en/docs/pipeline-stages/builders/ko.md
+++ b/docs/content/en/docs/pipeline-stages/builders/ko.md
@@ -212,9 +212,18 @@ Useful tips for existing `ko` users:
   [`default-repo` functionality]({{< relref "/docs/environment/image-registries" >}}).
   The ko builder does _not_ read the `KO_DOCKER_REPO` environment variable.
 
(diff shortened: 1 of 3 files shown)


halvards:
Forgot to update the docs. Converting to draft while I do that.

tejal29 (approved):

PR #6826, opened 2021-11-10
docs(ko): Improve ko docs for existing ko users

halvards:
Provide more examples of equivalent workflows between `ko` and the Skaffold ko builder.

Also add a note to the custom builder example that links to the (to be published) ko builder documentation.

Tracking: #6041

### DIFF
--- a/docs/content/en/docs/pipeline-stages/builders/ko.md
+++ b/docs/content/en/docs/pipeline-stages/builders/ko.md
@@ -155,6 +155,7 @@ e.g.:
 ```yaml
     ko:
       flags:
+      - -mod=vendor
       - -v
 ```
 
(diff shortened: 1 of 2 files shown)


docs/content/en/docs/pipeline-stages/builders/ko.md
```diff
@@ -216,27 +217,115 @@ Useful tips for existing `ko` users:
...
+  kind: Config
+  build:
+    artifacts:
+  EOF
+  grep -rho "ko://$(go list -m)[^\"]*" ./config/ | sort | uniq | xargs -Iimg echo -e "  - image: img\n    ko: {}" >> skaffold.yaml
+  cat << EOF >> skaffold.yaml
+    local:
+      concurrency: 0
+    tagPolicy:
+      sha256: {}
```
briandealwis, line 235: We can remove this tagPolicy

docs/content/en/docs/pipeline-stages/builders/ko.md
```diff
@@ -216,27 +217,115 @@ Useful tips for existing `ko` users:
...
 
   ```shell
-  ko resolve --filename k8s/*.yaml > out.yaml
+  cat << EOF > skaffold.yaml
+  apiVersion: skaffold/v2beta26
+  kind: Config
+  build:
+    artifacts:
+  EOF
+  grep -rho "ko://$(go list -m)[^\"]*" ./config/ | sort | uniq | xargs -Iimg echo -e "  - image: img\n    ko: {}" >> skaffold.yaml
```
briandealwis, line 230: We need to teach `skaffold init` how to recognize a Ko set up.  Can you mark this snippet as being a temporary measure?

integration/examples/custom/README.md
```diff
@@ -1,5 +1,10 @@
 ### Example: use the custom builder with ko
 
+**Note:** Skaffold now includes a
+[`ko` builder](https://skaffold.dev/docs/pipeline-stages/builders/ko/).
+When you use the `ko` builder, you do not need to provide a custom build shell
+script or install the `ko` binary.
```
briandealwis, line 6: We should replace this example with the buildx example.

briandealwis (approved):

docs/content/en/docs/pipeline-stages/builders/ko.md
halvards: The reason I added it is because this is a section specifically to help existing `ko` users. `ko` users tend to always use digests rather than tags, and this setting means that when Skaffold renders manifests, it'll render the same way as `ko`.

docs/content/en/docs/pipeline-stages/builders/ko.md
halvards: Good point, and `init` support is in our list of future features. I've added a note as you suggest for now.

integration/examples/custom/README.md
halvards: Agree. I was thinking to keep it for the first ko builder release, so users can compare. Then we can remove it from the next release.

docs/content/en/docs/pipeline-stages/builders/ko.md
briandealwis: The downside to the `sha256:` tagger is that it always tags as `latest`.  [`inputDigest`](https://skaffold.dev/docs/pipeline-stages/taggers/#inputdigest-uses-a-digest-of-the-artifact-source-to-tag-images) might be better.

(We have `sha256` throughout our examples.  I have a dangling patch somewhere to rename it to `latest` — maybe that would be good to bring in for Skaffold 2.0.)

PR #6829, opened 2021-11-10
test: fix integration test for stateful-sets

gsquared94:

### DIFF
--- a/integration/apply_test.go
+++ b/integration/apply_test.go
@@ -94,6 +94,7 @@ func TestApplyStatusCheckFailure(t *testing.T) {
 	}
 	for _, test := range tests {
 		testutil.Run(t, test.description, func(t *testutil.T) {
+			MarkIntegrationTest(t.T, NeedsGcp)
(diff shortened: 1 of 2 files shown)


pkg/diag/validator/statefulset.go
```diff
@@ -19,29 +19,32 @@ package validator
...
-	return &statefulSetPodsSelector{k, d}
+func NewStatefulSetPodsSelector(k kubernetes.Interface, name string) PodSelector {
+	return &statefulSetPodsSelector{k, name}
 }
 
 func (s *statefulSetPodsSelector) Select(ctx context.Context, ns string, opts metav1.ListOptions) ([]v1.Pod, error) {
+	ss, err := s.k.AppsV1().StatefulSets(ns).Get(ctx, s.name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
```
tejal29, line 40: Is there a reason we are fetching the same object twice?

gsquared94: reverted back

PR #6830, opened 2021-11-10
fix: update windows ci description to be correct

aaron-prindle:


.github/workflows/integration-windows.yml
```diff
@@ -28,6 +28,6 @@ jobs:
         fetch-depth: 0
 
     # TODO(aaron-prindle) skip integration tests for doc only changes on Windows.  Figure out the proper syntax, etc.
-    - name: Run windows unit tests and integration tests
+    - name: Run skaffold unit tests on windows
```
tejal29, line 31: do you want to use the same packages as darwin integration test?
```
  go test -short -timeout 60s  ./pkg/skaffold/... ./cmd/... ./hack/... ./pkg/webhook/...
```

aaron-prindle: I can update this to be the same as `darwin`.  For more context here, this is the same as what was in the travis.yml setup for windows as is currently:
```
    - os: windows
      name: "Windows unit"
      script:
        - go test -short -timeout 60s ./...
```

aaron-prindle: I've updated the windows ci tests to do a smaller subset of unit tests (the same as `darwin` as suggested)

tejal29: ah. I wonder how things were working before. 🤷

gsquared94 (approved):

Issue #6835, opened 2021-11-11
ko builder sideload displays both correct and incorrect tag when using default tag policy


halvards:
### Expected behavior

Using the ko builder with the default tag policy results in the output displaying the correct tag name.

### Actual behavior

The image builds correctly, and the correct tag is applied.

However, using the default tag policy results in the output displaying both the correct tag name and an image name with the wrong tag.

Using the `sha256` tag policy results in the correct tag being displayed.

### Information

Build from `main` at commit a9d715237873e19a22295159128ab9d85d7eb87b

Output:

```
$ skaffold build --cache-artifacts=false --push=false

Generating tags...
 - skaffold-ko -> australia-southeast1-docker.pkg.dev/test/default/skaffold-ko:v1.34.0-42-ga9d715237-dirty
Starting build...
Found [minikube] context, using local docker daemon.
Building [skaffold-ko]...
2021/11/11 22:49:46 Using base gcr.io/distroless/static:nonroot for github.com/GoogleContainerTools/skaffold/examples/ko
2021/11/11 22:49:50 Building github.com/GoogleContainerTools/skaffold/examples/ko for linux/amd64
2021/11/11 22:49:51 Loading australia-southeast1-docker.pkg.dev/test/default/skaffold-ko:83b8d5fec80f5c121ff0f4365f19a679ca1bb674c8533dd7189995c882da9d13
2021/11/11 22:49:51 Loaded australia-southeast1-docker.pkg.dev/test/default/skaffold-ko:83b8d5fec80f5c121ff0f4365f19a679ca1bb674c8533dd7189995c882da9d13
2021/11/11 22:49:51 Adding tag v1.34.0-42-ga9d715237-dirty
2021/11/11 22:49:51 Added tag v1.34.0-42-ga9d715237-dirty
australia-southeast1-docker.pkg.dev//default/skaffold-ko:83b8d5fec80f5c121ff0f4365f19a679ca1bb674c8533dd7189995c882da9d13
```

The image `australia-southeast1-docker.pkg.dev//default/skaffold-ko:v1.34.0-42-ga9d715237-dirty` is correctly built and exists in the local Docker daemon after the build is done.

briandealwis:
The final image ref also seems to be missing the `test` repository component:
`....dev//default/skaffold-ko...` vs
`...dev/test/default/skaffold-ko...`

halvards:
That may have been me editing out the project ID... :-)

PR #6836, opened 2021-11-11
fix(ko): Display correct tag for sideloaded images

halvards:
Display the expected tag after successfully building a ko image, for both sideloaded and published images.

Also handles the `tagPolicy.sha256` scenarios.

Fixes: #6835
Tracking: #6041

### DIFF
--- a/pkg/skaffold/build/ko/build.go
+++ b/pkg/skaffold/build/ko/build.go
@@ -52,7 +52,13 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, a *latestV1.Artifact
 	if err != nil {
 		return "", fmt.Errorf("could not build and publish ko image %q: %w", a.ImageName, err)
 	}
-	fmt.Fprintln(out, imageRef.Name())
+
+	outputRef := ref
+	if strings.HasSuffix(ref, ":latest") {
+		// handle tagPolicy sha256
+		outputRef = imageRef.Name()
+	}
+	fmt.Fprintln(out, outputRef)
 
 	return b.getImageIdentifier(ctx, imageRef, ref)
 }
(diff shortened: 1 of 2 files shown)


pkg/skaffold/build/ko/build.go
```diff
@@ -52,7 +52,13 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, a *latestV1.Artifact
...
 		return "", fmt.Errorf("could not build and publish ko image %q: %w", a.ImageName, err)
 	}
-	fmt.Fprintln(out, imageRef.Name())
+
+	outputRef := ref
+	if strings.HasSuffix(ref, ":latest") {
+		// handle tagPolicy sha256
+		outputRef = imageRef.Name()
+	}
+	fmt.Fprintln(out, outputRef)
```
briandealwis, line 61: We're not outputting the image ref for the other builders.  Why is this necessary?

halvards: Good question. `ko` itself prints it out at the end, and I've found that helpful, so this was just mimicking that behavior. And when using the Docker builder, the image name is printed at the end, even though Skaffold doesn't do it.

briandealwis: ok makes sense

pkg/skaffold/build/ko/build.go
```diff
@@ -52,7 +52,13 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, a *latestV1.Artifact
 	if err != nil {
 		return "", fmt.Errorf("could not build and publish ko image %q: %w", a.ImageName, err)
 	}
-	fmt.Fprintln(out, imageRef.Name())
+
+	outputRef := ref
+	if strings.HasSuffix(ref, ":latest") {
```
briandealwis, line 57: So does Ko only do use the digest for :latest?

halvards:
Marking as draft until I have time to get back to this one.

halvards:
Closing in favor of #6928

PR #6838, opened 2021-11-11
fix: `deploy --skip-render` not applying skaffold labels, causes status check to not work

gsquared94:
**Related**: https://github.com/GoogleContainerTools/skaffold/pull/6674

`skaffold deploy` invoked with the `--skip-render` flag should still apply skaffold labels so that status checker can function correctly.


nkubala (approved):
would be nice to have a quick integration test for this. we could just copy one of the `TestDeploy` test cases and throw this flag in to the invocation

PR #6841, opened 2021-11-11
fix: make kcc status-check less aggressive

### LINKED ISSUE #6842: kcc/status: Ready condition with non-empty message is treated as error
In skaffold v1.34 we are offering status checking for Kubernetes Config Connector.

However, if a user has a YAML file that with CRDs that have a dependencies to each other, then some resources that are in-progress of creation (currently failing due to its dependency not being Ready, but will eventually succeed with retries in CNRM controller) are causing Skaffold healthchecks to fail early because of this logic:

https://github.com/GoogleContainerTools/skaffold/blob/c70d294c50d4d9ee64c2bb4fd17083a2670897d0/pkg/diag/validator/config_connector.go#L71-L80

We should either remove this check and wait for the status to set to a failed kstatus by CNRM/KCC controller – or we should update kstatus package to surface the `Reason` field so we do not fail as the resource is being actuated.

Repro  (copied from https://cloud.google.com/config-connector/docs/reference/resource-docs/sql/sqlinstance#sample_yamls) 

```yaml
apiVersion: serviceusage.cnrm.cloud.google.com/v1beta1
kind: Service
metadata:
   name: sqladmin.googleapis.com
---
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLDatabase
metadata:
  labels:
    label-one: "value-one"
  name: sqldatabase-x5
spec:
  charset: utf8mb4
  collation: utf8mb4_bin
  instanceRef:
    name: sqldatabase-dep-x5
---
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLInstance
metadata:
  name: sqldatabase-dep-x5
spec:
  region: us-central1
  databaseVersion: MYSQL_5_7
  settings:
    tier: db-n1-standard-1

```

Status on the `sqldatabase.sql.cnrm.cloud.google.com` object currently looks like:

```
status:
  conditions:
  - lastTransitionTime: "2021-11-11T18:59:07Z"
    message: reference SQLInstance default/sqldatabase-dep-x5 is not ready
    reason: DependencyNotReady
    status: "False"
    type: Ready
```

cc: @gsquared94

gsquared94:
Fixes #6842
We cannot reliably use `message` field to determine if the resource has failed deployment. We're making it less aggressive.

### DIFF
--- a/integration/testdata/apply/skaffold.yaml
+++ b/integration/testdata/apply/skaffold.yaml
@@ -12,4 +12,5 @@ profiles:
   - name: configconnector
     deploy:
       kubectl:
+        statusCheckDeadlineSeconds: 10
         manifests: ["configconnector.yaml"]
--- a/pkg/diag/validator/config_connector.go
+++ b/pkg/diag/validator/config_connector.go
@@ -69,16 +69,11 @@ func getResourceStatus(res unstructured.Unstructured) (kstatus.Status, *proto.Ac
 	case kstatus.CurrentStatus:
 		ae = proto.ActionableErr{ErrCode: proto.StatusCode_STATUSCHECK_SUCCESS}
 	case kstatus.InProgressStatus:
-		if result.Message == "" {
-			ae = proto.ActionableErr{ErrCode: proto.StatusCode_STATUSCHECK_CONFIG_CONNECTOR_IN_PROGRESS, Message: result.Message}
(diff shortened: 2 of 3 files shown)


tejal29 (approved):

ahmetb:
addresses #6842

PR #6846, opened 2021-11-11
fix: Add skaffold internal error and return that instead of user cancelled

tejal29:
Relates to #5424

VSC nightly [job](https://fusion2.corp.google.com/invocations/78bd64a7-265b-4aca-b19d-453a550d338f/log)  is seeing a bunch of errors where status check returns "USER_CANCELLED" when the deployment failed.
The deployments could have failed due to
1) resource failed to stabilize within the given time. 
In this case, the [logic](https://github.com/GoogleContainerTools/skaffold/blob/8f67e8d4bc5e7117eea45d0a63982aa311147f6a/pkg/skaffold/kubernetes/status/resource/deployment.go#L440) returns the exit code seen by `Resource.resource` i.e. pods in case of deployments and statefulset, KCC resource Instance in case of KCC.

Previously, the code set`USER_CANCELLED` as failure code. In this PR, I have done the following changes

1) Code change to only return `USER_CANCELLED` if all resources status check was cancelled.
 To detect this condition add `cancelled` to the counter.

2) Add a new error code `STATUSCHECK_INTERNAL_ERROR` when resource fails but we don't find an offending error code.


pkg/skaffold/kubernetes/status/status_check.go
```diff
@@ -382,18 +382,25 @@ func pollResourceStatus(ctx context.Context, cfg kubectl.Config, r *resource.Res
...
 		return proto.StatusCode_STATUSCHECK_SUCCESS, nil
 	}
 	err := fmt.Errorf("%d/%d deployment(s) failed", c.failed, c.total)
+	if ctxErr == context.DeadlineExceeded {
+		return proto.StatusCode_STATUSCHECK_DEADLINE_EXCEEDED, err
+	} else {
+		if ctxErr == context.Canceled {
+			return proto.StatusCode_STATUSCHECK_USER_CANCELLED, err
+		}
+	}
```
tejal29, line 396:
```suggestion
	} else if ctxErr == context.Canceled {
		return proto.StatusCode_STATUSCHECK_USER_CANCELLED, err
	}
```

aaron-prindle (dismissed):
LGTM

pkg/skaffold/kubernetes/status/status_check.go
```diff
@@ -243,8 +246,7 @@ func (s *monitor) statusCheck(ctx context.Context, out io.Writer) (proto.StatusC
 
 	// Wait for all deployment statuses to be fetched
 	wg.Wait()
-	cancel()
```
tejal29, line 246: There is a `defer cancel` at L223

tejal29:
==== Testing Notes =====

1) Return status code other than success for unstable deployment

```
$ LOCAL=true make && cd integration/testdata/unstable-deployment
```

```
$ ../../../out/skaffold dev -d gcr.io/tejal-gke1 -v=debug
DEBU[0010] Fetching logs for container unstable-deployment-7cbcd465c7-z898k/incorrect-example  subtask=-1 task=DevLoop
 - deployment/unstable-deployment: BackOff: Back-off restarting failed container
    - pod/unstable-deployment-7cbcd465c7-z898k: BackOff: Back-off restarting failed container
DEBU[0011] marking resource failed due to error code STATUSCHECK_CONTAINER_RESTARTING  subtask=-1 task=Deploy
 - deployment/unstable-deployment: container incorrect-example is backing off waiting to restart
    - pod/unstable-deployment-7cbcd465c7-z898k: container incorrect-example is backing off waiting to restart
 - deployment/unstable-deployment failed. Error: container incorrect-example is backing off waiting to restart.
DEBU[0011] setting skaffold deploy status to STATUSCHECK_CONTAINER_RESTARTING.  subtask=-1 task=Deploy

```


2) Return status code cancelled when you hit control C in status check phase.

```
$ ../../../out/skaffold dev -d gcr.io/tejal-gke1 -v=debug

INFO[0004] Deploy completed in 1.661 second              subtask=-1 task=Deploy
DEBU[0004] getting client config for kubeContext: `gke_tejal-gke1_us-central1-c_dump-apis`  subtask=-1 task=DevLoop
DEBU[0004] getting client config for kubeContext: `gke_tejal-gke1_us-central1-c_dump-apis`  subtask=-1 task=DevLoop
DEBU[0005] checking status deployment/unstable-deployment  subtask=-1 task=Deploy
^CDEBU[0005] marking resource status check cancelledSTATUSCHECK_USER_CANCELLED  subtask=-1 task=Deploy
DEBU[0005] set

```

pkg/skaffold/kubernetes/status/status_check.go
```diff
@@ -220,17 +221,19 @@ func (s *monitor) statusCheck(ctx context.Context, out io.Writer) (proto.StatusC
...
+	var exitStatusCode proto.StatusCode
 
 	for _, d := range resources {
 		wg.Add(1)
 		go func(r *resource.Resource) {
 			defer wg.Done()
 			// keep updating the resource status until it fails/succeeds/times out
 			pollResourceStatus(ctx, s.cfg, r)
-			rcCopy := c.markProcessed(r.Status().Error())
+			rcCopy := c.markProcessed(ctx, r.StatusCode())
```
briandealwis, line 232: Can we move the logic here and in `getSkaffoldDeployStatus()` into `counter`?  This scattering of logic is awkward:
- `markProcessed` can check if `ctx.Err() == context.Canceled` to determine if the user cancelled
- have `markProcessed` return `(counter, failed bool)` to replace the `resourceFailed()` check on line 235
- move the exitStatusCode determination into `counter`
- line 249 would `return c.result()`

tejal29: We can't use `markProcessed can check if ctx.Err() == context.Canceled to determine if the user cancelled` since, we fail fast and cancel context if a deployment failed.
So at this point, its difficult to detect if a user hit cntrl C Vs, code cancelling all the status check for all resources.

tejal29: I had suggestions 2-4 implemented but then the unit tests were becoming more complex. Maybe the `counter` need to be renamed to `deployState`.

tejal29: Ok, so i remember, why we cant have counter store the `exitStatusCode`. The counter gets updated in a go routine. 
The exit code is stored in `resources`. It makes sense to declare another mutex to update the exit status code safely.

tejal29: Ok, 
I am addressed 1 and 2 from the suggestions. 
For 3, 4: 
 - I have declared a exitStatusCode in `checkStatus` and updated the code to only capture first failure.
 -  kept the `getDeployStatus` as is.

pkg/skaffold/kubernetes/status/status_check.go
```diff
@@ -220,17 +221,23 @@ func (s *monitor) statusCheck(ctx context.Context, out io.Writer) (proto.StatusC
...
+			// keep updating the resource status until it fails/succeeds/times out/cancelled.
 			pollResourceStatus(ctx, s.cfg, r)
-			rcCopy := c.markProcessed(r.Status().Error())
+			rcCopy, failed := c.markProcessed(ctx, r.StatusCode())
 			s.printStatusCheckSummary(out, r, rcCopy)
-			// if one deployment fails, cancel status checks for all deployments.
-			if r.Status().Error() != nil && r.StatusCode() != proto.StatusCode_STATUSCHECK_USER_CANCELLED {
+			// if a resource fails, cancel status checks for all resources to fail fast
+			// and capture the first failed exit code.
+			if failed {
```
briandealwis, line 237: It's odd that cancel is not considered a failure — worth documenting here.

pkg/skaffold/kubernetes/status/status_check.go
```diff
@@ -382,18 +388,23 @@ func pollResourceStatus(ctx context.Context, cfg kubectl.Config, r *resource.Res
...
 	}
+	// construct an error message and return appropriate error code
 	err := fmt.Errorf("%d/%d deployment(s) failed", c.failed, c.total)
-	for _, r := range rs {
-		if r.StatusCode() != proto.StatusCode_STATUSCHECK_SUCCESS &&
-			r.StatusCode() != proto.StatusCode_STATUSCHECK_USER_CANCELLED {
-			return r.StatusCode(), err
-		}
+	if sc == proto.StatusCode_STATUSCHECK_SUCCESS || sc == 0 {
+		log.Entry(ctx).Debugf("found statuscode %s. setting skaffold deploy status to STATUSCHECK_INTERNAL_ERROR.", sc)
```
briandealwis, line 403: I wonder if we should have a `panicIfDev()` function that will fail for non-release builds.

proto/enums/enums.proto
```diff
@@ -307,6 +307,8 @@ enum StatusCode {
     STATUSCHECK_CONTAINER_WAITING_UNKNOWN = 503;
     // Container event reason unknown
     STATUSCHECK_UNKNOWN_EVENT = 509;
+    // Status Check internal error
+    STATUSCHECK_INTERNAL_ERROR = 514;
```
briandealwis, line 311: Could we make this a general Skaffold internal error and not just statuscheck?  Then we could capture `panic()` and notify about these too.

briandealwis (approved):

Issue #6810, opened 2021-11-07
`skaffold build` fails when attempting to docker COPY an empty directory


aaron-prindle:
Curently if a Dockerfile attempt to COPY an empty directory into a file, skaffold build will fail even though this is valid.  This makes some valid Dockerfiles not usable with skaffold which is an issue as we hope to fully support all Dockerfiles.  Example steps to reproduce this issue taken from - https://github.com/GoogleContainerTools/skaffold/pull/6796#issuecomment-962691340:

Steps to reproduce:
cd into `skaffold/examples/docker-deploy/bert/`
```
mkdir test
# add following copy to Dockerfile
COPY test .
cd ..
skaffold build
Generating tags...
 - bert -> bert:151b8d3-dirty
 - ernie -> ernie:151b8d3
Checking cache...
 - bert: Not found. Building
 - ernie: Found Locally
Starting build...
Found [ed-ks2] context, using local docker daemon.
Building [bert]...
Sending build context to Docker daemon  3.072kB
Step 1/9 : FROM golang:1.15 as builder
 ---> 40349a2425ef
Step 2/9 : COPY main.go .
 ---> Using cache
 ---> 5bcddb464805
Step 3/9 : ARG SKAFFOLD_GO_GCFLAGS
 ---> Using cache
 ---> 0fe720d5c3d6
Step 4/9 : RUN go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -o /app main.go
 ---> Using cache
 ---> b0fa60be1cbd
Step 5/9 : FROM alpine:3.10
 ---> e7b300aee9f9
Step 6/9 : ENV GOTRACEBACK=single
 ---> Using cache
 ---> e850d4805844
Step 7/9 : COPY test .
unable to stream build output: COPY failed: file not found in build context or excluded by .dockerignore: stat test: file does not exist
```
Verify directory exists:
```
ls bert/
Dockerfile   main.go   src   test
```

Now, add a single file to the directory and try again:
```
touch bert/test/testfile
 skaffold build
Generating tags...
 - bert -> bert:151b8d3-dirty
 - ernie -> ernie:151b8d3
Checking cache...
 - bert: Not found. Building
 - ernie: Found Locally
Starting build...
Found [ed-ks2] context, using local docker daemon.
Building [bert]...
Sending build context to Docker daemon  3.584kB
Step 1/9 : FROM golang:1.15 as builder
 ---> 40349a2425ef
[...]

PR #6796, opened 2021-11-04
fix: Quotes in dockerfiles env vars break copy dependency checks

kadern0:
Fixes #6762
Signed-off-by: Pablo Caderno <kaderno@gmail.com>

**Description**

Fixes the issue with quoted variables and "COPY" commands.

Before:
```
skaffold build
Generating tags...
 - bert -> bert:151b8d3-dirty
 - ernie -> ernie:151b8d3
Checking cache...
 - bert: Error checking cache.
getting hash for artifact "bert": getting dependencies for "bert": file pattern [./src/"some_dir"] must match at least one file
```
After:
```
skaffold build
Generating tags...
 - bert -> bert:151b8d3-dirty
 - ernie -> ernie:151b8d3
Checking cache...
 - bert: Not found. Building
 - ernie: Found Locally
Starting build...
Found [ed-ks2] context, using local docker daemon.
Building [bert]...
Sending build context to Docker daemon  7.168kB
Step 1/20 : FROM golang:1.15 as builder
 ---> 40349a2425ef
Step 2/20 : COPY main.go .
 ---> Using cache
 ---> 5bcddb464805
Step 3/20 : ARG SKAFFOLD_GO_GCFLAGS
 ---> Using cache
 ---> 0fe720d5c3d6
Step 4/20 : RUN go build -gcflags="${SKAFFOLD_GO_GCFLAGS}" -o /app main.go
 ---> Using cache
 ---> b0fa60be1cbd
Step 5/20 : FROM alpine:3.10
 ---> e7b300aee9f9
Step 6/20 : ENV MY_NAME="John Doe" MY_DOG=Rex\ The\ Dog     MY_CAT=fluffy
 ---> Using cache
 ---> 0a651a254ade
Step 7/20 : ENV GOTRACEBACK=single
 ---> Using cache
 ---> 0588ba733bf7
Step 8/20 : ENV DIR2=some_dir
 ---> Using cache
 ---> ac1901e9b83c
Step 9/20 : ENV DIRNAME="some_dir"
 ---> Using cache
 ---> 0a5beef983fa
Step 10/20 : ENV NOBREAK="asdf'asdf"
 ---> Running in 48d6824c02ad
 ---> 17df958a420a
Step 11/20 : COPY $NOBREAK .
 ---> 1fcafb39acb1
Step 12/20 : COPY a .
 ---> 428b63829342
Step 13/20 : COPY src/ .
 ---> 4e55fcb5d5c5
[...]

PR #6851, opened 2021-11-12
fix: skaffold build fails when attempting to COPY an empty dir

kadern0:
Signed-off-by: Pablo Caderno <kaderno@gmail.com>

Fixes: #6810 
Related: #6796 

**Description**

Refactored function WalkWorkspace to also run over directories. If directory wasn't empty, by copying the file it would create also the directory, but if the directory is empty there is some code region that couldn't be reached the way it was written.

The code I've added has some duplication that I'm not a huge fan of that could be easily replaced if there's an easy/better way to figure out that a directory is empty.

If this is merged, the test introduced within #6796 can be deleted as this would replace it.

### DIFF
--- a/pkg/skaffold/docker/dependencies.go
+++ b/pkg/skaffold/docker/dependencies.go
@@ -251,7 +251,7 @@ func readDockerignore(workspace string, absDockerfilePath string) ([]string, err
 }
 
 // WalkWorkspace walks the given host directories and records all files found.
-// Note: if you change this function, you might also want to modify `walkWorkspaceWithDestinations`.
+// Note: if you change this function, you might also want to modify walkWorkspaceWithDestinations.
 func WalkWorkspace(workspace string, excludes, deps []string) (map[string]bool, error) {
 	dockerIgnored, err := NewDockerIgnorePredicate(workspace, excludes)
 	if err != nil {
@@ -264,12 +264,6 @@ func WalkWorkspace(workspace string, excludes, deps []string) (map[string]bool, 
 		absFrom := filepath.Join(workspace, dep)
 
 		keepFile := func(path string, info walk.Dirent) (bool, error) {
-			// This region was never reached
-			// Always keep root folders.
-			// if info.IsDir() && path == absFrom {
-			// 	return false, nil
-			// }
-
 			// Ignore non empty dirs
 			if info.IsDir() && !util.IsEmptyDir(path) {
 				return false, nil
--- a/pkg/skaffold/docker/dependencies.go
+++ b/pkg/skaffold/docker/dependencies.go
@@ -264,9 +264,15 @@ func WalkWorkspace(workspace string, excludes, deps []string) (map[string]bool, 
 		absFrom := filepath.Join(workspace, dep)
 
 		keepFile := func(path string, info walk.Dirent) (bool, error) {
+			// This region was never reached
 			// Always keep root folders.
-			if info.IsDir() && path == absFrom {
-				return true, nil
+			// if info.IsDir() && path == absFrom {
+			// 	return false, nil
+			// }
+
+			// Ignore non empty dirs
+			if info.IsDir() && !util.IsEmptyDir(path) {
+				return false, nil
 			}
 
 			ignored, err := dockerIgnored(path, info)
@@ -277,7 +283,7 @@ func WalkWorkspace(workspace string, excludes, deps []string) (map[string]bool, 
 			return !ignored, nil
 		}
 
-		if err := walk.From(absFrom).Unsorted().When(keepFile).WhenIsFile().Do(func(path string, info walk.Dirent) error {
+		if err := walk.From(absFrom).Unsorted().When(keepFile).Do(func(path string, info walk.Dirent) error {
 			relPath, err := filepath.Rel(workspace, path)
 			if err != nil {
 				return err
@@ -289,6 +295,5 @@ func WalkWorkspace(workspace string, excludes, deps []string) (map[string]bool, 
 			return nil, fmt.Errorf("walking %q: %w", absFrom, err)
 		}
 	}
-
 	return files, nil
 }
--- a/pkg/skaffold/util/util.go
+++ b/pkg/skaffold/util/util.go
@@ -23,6 +23,7 @@ import (
 	"crypto/rand"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -285,6 +286,20 @@ func IsDir(path string) bool {
 	info, err := os.Stat(path)
 	// err could be permission-related
 	return (err == nil || !os.IsNotExist(err)) && info.IsDir()
+}
+
+// IsEmptyDir returns true for empty directories otherwise false
+func IsEmptyDir(path string) bool {
+	d, err := os.Open(path)
+	if err != nil {
+		return false
+	}
+	defer d.Close()
+	if _, err := d.ReadDir(1); err == io.EOF {
+		return true
+	}
+	return false
+
 }
 
 // IsHiddenDir returns if a directory is hidden.
--- a/pkg/skaffold/util/util.go
+++ b/pkg/skaffold/util/util.go
@@ -299,7 +299,6 @@ func IsEmptyDir(path string) bool {
 		return true
 	}
 	return false
-
 }
 
 // IsHiddenDir returns if a directory is hidden.
--- a/integration/run_test.go
+++ b/integration/run_test.go
@@ -178,7 +178,7 @@ func TestRun(t *testing.T) {
 		t.Run(test.description, func(t *testing.T) {
 			ns, client := SetupNamespace(t)
 			args := append(test.args, "--cache-artifacts=false")
-			if test.dir == "emptydir" {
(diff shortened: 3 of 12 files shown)


pkg/skaffold/docker/dependencies.go
```diff
@@ -283,7 +283,26 @@ func WalkWorkspace(workspace string, excludes, deps []string) (map[string]bool,
...
+			relPath, err := filepath.Rel(workspace, path)
+			if err != nil {
+				return err
+			}
+			var keepDir = true
+			for item := range files {
+				if strings.Contains(item, relPath) {
+					keepDir = false
+				}
+			}
```
briandealwis, line 298: This will be very expensive and also suffers from false positives.  Consider the following:
```
test/  <-- empty directory
testing/README.txt
src/main/test/Test.java
```

`testing/README.txt` contains `test`
`src/main/test/Main.java` contains `test`

There are a few approaches I can think of here:
1. Change the original walk to process both files and directories.  If the path is a directory, check if the directory has contents, and if empty then add the directory to `files`.  This could entail adding a `walk.IsEmpty(dir string) bool` funciton, whose implementation would use [godirwalk's NewScanner](https://github.com/karrick/godirwalk/blob/f64d91544257ab0cc1cca44f22600925a1b9ddae/examples/remove-empty-directories/main.go#L50-L66).
2. Do two passes, where the first pass collects the parent directories of files to a separate `directoriesWithChildren` map. Then walk the directories in the second pass and only add directories to `files` if they weren't in `directoriesWithChildren`.
3. Add a new `DoThen(Action, postDir Action)` action to our `walk` interface that adds a post-directory action using [`godirwalk`'s PostChildrenCallback](https://github.com/karrick/godirwalk/tree/f64d91544257ab0cc1cca44f22600925a1b9ddae#configurable-post-children-callback).  Then do the same `directoriesWithChildren` building as in Option 2 but then check empty-dirs in this post-children callback.

kadern0: Thank you very much for your comment. I agree with you about the extra cost of going over all the files for each directory (although preventing false positives should be easy to do). 

My initial idea was using a function to check if the directory is empty (so basically the first approach you described). That would avoid having to make two passes. 

I'm going to have a look at the godirwalk to see how it's being done. I'll let you know if I get stuck.

pkg/skaffold/docker/dependencies.go
```diff
@@ -260,10 +259,17 @@ func WalkWorkspace(workspace string, excludes, deps []string) (map[string]bool,
...
 
 		keepFile := func(path string, info walk.Dirent) (bool, error) {
+			// This region was never reached
 			// Always keep root folders.
-			if info.IsDir() && path == absFrom {
-				return true, nil
+			// if info.IsDir() && path == absFrom {
+			// 	return false, nil
+			// }
+
```
briandealwis, line 267: Remove this code (I'm not sure what line 262 is referring to)

briandealwis (commented):
Thanks for the contribution!  Two nits to address.

pkg/skaffold/util/util.go
```diff
@@ -287,6 +287,14 @@ func IsDir(path string) bool {
 	return (err == nil || !os.IsNotExist(err)) && info.IsDir()
 }
 
+func IsEmptyDir(path string) bool {
```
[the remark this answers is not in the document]
kadern0: Wanted to know if this approach was good first. I'll add testing.

pkg/skaffold/docker/dependencies.go
kadern0: I wanted to show  (for discussion) that this `if` statement always evaluates to false since the `keepfile` function was only being called for files and not directories. Also, if we add the root directory tests fail so I initially set it to false but then I realized we don't need that part at all (hence commented out and ready to be removed if there's an agreement)

briandealwis: Ah makes sense 👍

pkg/skaffold/util/util.go
```diff
@@ -287,6 +287,14 @@ func IsDir(path string) bool {
...
+	files, err := os.ReadDir(path)
+	if err != nil {
+		return false
+	}
+	return len(files) == 0
```
briandealwis, line 295: I think it would be worth optimizing this.  The [implementation of `os.ReadDir(string)`](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/os/dir.go;l=115-125) does the following:
```golang
func ReadDir(name string) ([]DirEntry, error) {
	f, err := Open(name)
	if err != nil {
		return nil, err
	}
	defer f.Close()

	dirs, err := f.ReadDir(-1)
	sort.Slice(dirs, func(i, j int) bool { return dirs[i].Name() < dirs[j].Name() })
	return dirs, err
}
```

We don't care about the full list nor about sorting.  So we could just use:
```suggestion
	d, err := os.Open(name)
	if err != nil {
		return false
	}
	defer d.Close()
	ents, err := d.ReadDir(1)
	if err != nil {
		return false
	}
	return len(ents) == 0
```

kadern0: Agree. That will be more efficient for this use case.

pkg/skaffold/util/util.go
```diff
@@ -287,6 +287,20 @@ func IsDir(path string) bool {
...
 }
 
+// IsEmptyDir returns true for empty directories otherwise false
+func IsEmptyDir(path string) bool {
+	d, err := os.Open(path)
+	if err != nil {
+		return false
+	}
+	defer d.Close()
+	ents, err := d.ReadDir(0)
```
briandealwis, line 297: According to the [docs](https://pkg.go.dev/os#File.ReadDir), `n ≤ 0` will cause all of the entries to be read.  Setting `n = 1` will ensure that we only read the first entry.

kadern0: You're right but setting it to 1 will also return error for empty directories so it was failing. I think the documentation is not too clear:
```
If n <= 0, ReadDir returns all the DirEntry records remaining in the directory. When it succeeds, it returns a nil error (not io.EOF). 
```

You can test it yourself really quick. This is the output when I set `n = 1`:
```
=== RUN   TestIsEmptyDir
    util_test.go:457: bool differ (-got, +want):   bool(
        - 	false,
        + 	true,
          )
--- FAIL: TestIsEmptyDir (0.00s)
```

briandealwis: I'm sure I tested this but I deleted my test case.  Trying again, it fails just like you say 🤦 

Easy fix though:
```
	if _, err := d.ReadDir(1); err == io.EOF {
		return true
	}
	return false
```

kadern0: I like this one, thanks! :+1:

pkg/skaffold/docker/dependencies.go
briandealwis: These comments need to be cleaned up.

kadern0:
@briandealwis, according to this comment on the `WalkWorkspace` function: `// Note: if you change this function, you might also want to modify `walkWorkspaceWithDestinations`.`  we might still have some work to do. 

On a quick look, this function seems to also ignore empty directories. I'm going to add a `WIP` to the PR and see what else I can find.

kadern0:
Hi @briandealwis , I think I'm going to need some help here. I think this code change fixes the issue, but I cannot create an appropriate test for it. I was trying to modify the integration test but I'm struggling with the empty directory creation since you cannot add empty dirs on git (which I did not know until working on this). The error thrown on the integration tests means that the `testdir` is not being created where it was supposed to so I'm wondering how to do that properly. I'll be offline for 10 days so I won't be able to work on this until then.

SSGoku369 (commented):
ok

SSGoku369:
ok

MarlonGamez:
hi @kadern0, sorry for disappearing on this PR for so long. Are you still interested on working on this? We can try to help out, however, we're planning on a code freeze for a bit this upcoming week in order to prepare for a large merge we're doing. If you're still interested in working on this, we can leave it open and pick it up once that merge is done

kadern0:
Hi @MarlonGamez, happy to work on this after the merge, although I think I need a hand to create a proper test for this.

aaron-prindle:
Friendly ping @kadern0, sorry for the delay with the PR here.  Is there any change you can resubmit your changes such that we can get CI green and we can get this over the finish line?  Thanks!

kadern0:
Thanks for the ping. I'm finishing my holidays, should be able to work on
it towards the end of next week.

kadern0:
@aaron-prindle, I've rebased the branch and quickly performed some tests locally. Whilst it seems to be doing what's expected, I've removed the WIP from the PR so all the tests are triggered although some further inspection would be appreciated.

aaron-prindle:
Seems there is currently a test failure related to the test added:
```
    util.go:246: Timed out waiting for pods [empty-dir] in namespace "skaffoldpwspk"
    --- FAIL: TestRun/copying-empty-directory (325.34s)
```

aaron-prindle (approved):
LGTM.  Thanks for your work here @kadern0 and sorry for the delays!!

PR #6860, opened 2021-11-15
feat: add line & col info to Artifact schema and related validations

aaron-prindle:
This PR adds Line and Column information to the `Artifact` type in the skaffold schema.  This is one step towards making validations and additional LSP/IDE based features faster/have-better-UX as the skaffold schema elements will have associated line numbers in their source file so we can better direct users.  Additional work might need to be done to hide these fields (perhaps in docs, etc), my understanding is that using the yaml tag like the following:
```
Line int `yaml:"-"`
```
should be sufficient for marshalling and unmarshalling but I need to investigate more to fully verify

### DIFF
--- a/pkg/skaffold/schema/latest/v1/config.go
+++ b/pkg/skaffold/schema/latest/v1/config.go
@@ -877,6 +877,12 @@ type Artifact struct {
 
 	// LifecycleHooks describes a set of lifecycle hooks that are executed before and after each build of the target artifact.
 	LifecycleHooks BuildHooks `yaml:"hooks,omitempty"`
+
+	// Line is the line number for the source skaffold.yaml file which has this artifact configuration info
+	Line int `yaml:"-"`
+
+	// Column is the column number for the source skaffold.yaml file for the artifact congfiguration info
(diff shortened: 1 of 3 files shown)


pkg/skaffold/schema/latest/v1/config.go
```diff
@@ -877,6 +877,12 @@ type Artifact struct {
 
 	// LifecycleHooks describes a set of lifecycle hooks that are executed before and after each build of the target artifact.
 	LifecycleHooks BuildHooks `yaml:"hooks,omitempty"`
+
+	// Line is the line number for the source skaffold.yaml file which has this artifact configuration info
+	Line int `yaml:"-"`
+
+	// Column is the column number for the source skaffold.yaml file for the artifact congfiguration info
+	Column int `yaml:"-"`
 }
```
MarlonGamez, line 886: I wonder if it's worth moving these into a struct together. If we're adding lint rules for the artifact schema, I'm assuming that we'll probably be adding lint rules for other schema as well in the future. In that case, we could just drop that struct into whatever type we're adding lint rules for. Thoughts?

PR #6865, opened 2021-11-16
chore(deps): update to kompose 1.26

briandealwis:
Update `kompose` to 1.26.


briandealwis:
Converting to draft: I'm not sure we actually have a test of kompose in action.

tejal29 (changes requested):
Please make the same change to `Dockerfile.deps.lts`

tejal29 (approved):
NVM there is no compose in LTS image. I  confused Kompose with Kustomize.

Issue #6870, opened 2021-11-17
Skaffold apply deployments never stabilize on Skaffold 1.35  


edwardthiele:
Using the manifest file "manifest.xml"
```
apiVersion: v1
kind: Pod
metadata:
  name: getting-started
spec:
  containers:
  - image: k8s.gcr.io/echoserver:1.4
    name: echoserver
```

with the command:
```
/usr/local/bin/skaffold apply 'manifest.yaml'
```

skaffold apply 'manifest.yaml'
Starting deploy...
 - pod/getting-started configured
Waiting for deployments to stabilize...
 - pods: could not stabilize within 10m0s
 - pods failed. Error: could not stabilize within 10m0s.

The actual deployment appears to work as expected

### Expected behavior

The deployment reports success 

### Actual behavior

The command times out

### Information

- Skaffold version: 1.35
- Operating system: Linux
- Installed via: skaffold.dev

briandealwis:
I've been able to reproduce this by deploying to a GKE Autopilot cluster.  In digging into the issue, I've found that we call into `checkStandalonePodsStatus` and `len(r.resources) == 0` and so we cycle:
https://github.com/GoogleContainerTools/skaffold/blob/9f56bbd6b90128c984d6fb6191dec948fc25822c/pkg/skaffold/kubernetes/status/resource/deployment.go#L144-L147

```
DEBU[0015] about to call CheckStatus on pods             subtask=-1 task=DevLoop
DEBU[0015] checkStandalonePodsStatus                     subtask=-1 task=DevLoop
DEBU[0015] update status: errCode:STATUSCHECK_STANDALONE_PODS_PENDING  subtask=-1 task=DevLoop
DEBU[0015] finished call to CheckStatus on pods          subtask=-1 task=DevLoop
DEBU[0016] about to call CheckStatus on pods             subtask=-1 task=DevLoop
DEBU[0016] checkStandalonePodsStatus                     subtask=-1 task=DevLoop
DEBU[0016] update status: errCode:STATUSCHECK_STANDALONE_PODS_PENDING  subtask=-1 task=DevLoop
DEBU[0016] finished call to CheckStatus on pods          subtask=-1 task=DevLoop
DEBU[0017] about to call CheckStatus on pods             subtask=-1 task=DevLoop
DEBU[0017] checkStandalonePodsStatus                     subtask=-1 task=DevLoop
[...]

PR #6875, opened 2021-11-18
fix: apply cmd should run `kubectl create --dry-run` to get the `ManifestList`

gsquared94:
Fixes: #6870


briandealwis (approved):
Good find!

Issue #6348, opened 2021-08-03
Upgrade CI to Go 1.16


yuwenma:
### Expected behavior
`os.WriteFile` should be supported.

### Actual behavior
Travis CI reports  `undefined: os.WriteFile`

### Information
Travis CI runs on Go1.15 and cannot support some Go1.16  features like `os.WriteFile`.
This disallow us to develop on Go1.16 and misses some cool features https://golang.org/doc/go1.16

aaron-prindle:
We have officially migrated off of Travis to GH Actions.  It might make sense to close/rename this issue

yuwenma:
THanks Aaron.

MarlonGamez:
closing this since we've moved our release process and are using go `1.17+`

PR #6882, opened 2021-11-19
ci: update `deploy/cross/Dockerfile` to use `neilotoole/xcgo:go1.17` as base image

MarlonGamez:
**Related**: https://github.com/GoogleContainerTools/skaffold/issues/6348

**Description**
This PR removes `dockercore/golang-cross` as the base image for our cross compilation Dockerfile, as this is limiting us to using `go 1.14`, which is pretty outdated at this point and is causing us issues with building skaffold. It now uses `neilotoole/xcgo:go1.17`, which currently uses `go 1.17.2`.

This image is quite a bit larger than our previous one (1.16 GB compared to 687 MB), but it will actually allow us to use newer versions of go

### DIFF
--- a/deploy/cross/Dockerfile
+++ b/deploy/cross/Dockerfile
@@ -12,17 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM dockercore/golang-cross:1.13.15 as base
-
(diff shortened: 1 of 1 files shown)


MarlonGamez:
Closing this PR as we've discussed offline and don't want to use this image. We will prioritize moving onto kokoro instead, which will allow us to do what we need

Issue #6861, opened 2021-11-15
Skaffold should output which kube-context + cluster is being used


briandealwis:
It should be more clear for users (and people trying to offer support by looking at logs!) which kube-context and cluster is being used.
```
$ skaffold dev
Listing files to watch...
 - skaffold-example
Generating tags...
 - skaffold-example -> skaffold-example:v1.34.0-60-gef8185519
Checking cache...
 - skaffold-example: Not found. Building
Starting build...
Building [skaffold-example]...
```

This information is shown with `-vinfo`, but most users do not run at this level:
```
$ skaffold dev -v info
INFO[0000] Skaffold &{Version:v1.34.0-60-gef8185519 ConfigVersion:skaffold/v2beta26 GitVersion: GitCommit:ef818551948c82de81ef4308aef7dbcb36d3dede BuildDate:2021-11-15T15:47:08Z GoVersion:go1.17.2 Compiler:gc Platform:darwin/arm64 User:}  subtask=-1 task=DevLoop
INFO[0000] Loaded Skaffold defaults from "/Users/bdealwis/.skaffold/config"  subtask=-1 task=DevLoop
INFO[0000] Using kubectl context: gke_k8s-skaffold_us-west1_test-cluster  subtask=-1 task=DevLoop
INFO[0001] build concurrency first set to 1 parsed from *local.Builder[0]  subtask=-1 task=DevLoop
```

OmSaran:
Can I take a stab at this?

aaron-prindle:
@OmSaran that would be great if you want to add this feature.  If you could create a PR with the feature the Skaffold maintainers would be happy to review and get the feature in

riyamittal009:
@OmSaran and I have taken a stab at this issue, please see PR #8587

majd102-p:
Hi  @GoogleContainerTools/skaffold-team and @briandealwis I've submitted PR #10069 to address this. Everything is ready on my end (including the os.Stderr update and CLA). Ready for review and for the workflows to be triggered whenever you have a moment. Thanks

PR #6886, opened 2021-11-21
feat: output which kube-context and cluster is being used

kadern0:
Signed-off-by: Pablo Caderno <kaderno@gmail.com>

Fixes: #6861 

New output:
```
$skaffold dev
Using kubectl context: minikube
Using cluster: minikube
Listing files to watch...
 - skaffold-helm
Generating tags...
 - skaffold-helm -> skaffold-helm:151b8d3-dirty
Checking cache...
 - skaffold-helm: Found Locally

```

It might look better if both are printed in a single line although it would be a little less readable I think.

### DIFF
--- a/pkg/skaffold/deploy/kubectl/kubectl.go
+++ b/pkg/skaffold/deploy/kubectl/kubectl.go
@@ -40,6 +40,7 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/hooks"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/instrumentation"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes"
+	kubectx "github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/context"
 	k8slogger "github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/logger"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/manifest"
 	kstatus "github.com/GoogleContainerTools/skaffold/pkg/skaffold/kubernetes/status"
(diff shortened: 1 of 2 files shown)


pkg/skaffold/runner/runcontext/context.go
```diff
@@ -235,8 +235,9 @@ func GetRunContext(ctx context.Context, opts config.SkaffoldOptions, configs []s
 		return nil, fmt.Errorf("getting current cluster context: %w", err)
 	}
 	kubeContext := kubeConfig.CurrentContext
-	log.Entry(context.TODO()).Infof("Using kubectl context: %s", kubeContext)
 
+	fmt.Printf("Using kubectl context: %s\n", kubeContext)
+	fmt.Printf("Using cluster: %s\n", kubeConfig.Contexts[kubeContext].Cluster)
```
briandealwis, line 240: So we should use `output.Default` to get the default colouring of console messages:
```
	output.Default.Fprintf(out, "Using kubectl context: %s\n", kubeContext)
	output.Default.Fprintf(out, "Using cluster: %s\n", kubeConfig.Contexts[kubeContext].Cluster)
```

That'll require updating callers to pass through the output `io.Writer`.

I then tried running `skaffold dev --kube-context blahblah` and got a panic as I had no `blahblah` context!  We apparently were never validating the contexts 😱   Would you be up for modifying the `CurrentConfig()` to validate the context and returning an error if not found?

kadern0: I was expecting this to happen :) I'll give it a go and let you know if I get stuck. 

Are both changes OK within the same commit?

tejal29:
@kadern0 The changes look good. Some of the integration tests are failing. Do you need help fixing these?

kadern0:
@tejal29, I've been offline for a few days. I'll have a look at the tests and I'll fix the issues :)

tejal29:
sounds good. Thanks!

kadern0:
@briandealwis , @tejal29 sorry to bother you but after quite some tries it seems I'm not able to make any progress on this. Initially I was trying to output the info from the `context.go` but it turned out, such output had to be a valid yaml because it's validated later on so it didn't seem the right place. After that, I moved it to `kubectl.go` and whilst it kind of works, there are some cases when it fails all the time. The way I get the context is broken on Windows and Mac and I can't easily test that since I'm using Linux. Any idea how could I solve that?

SSGoku369 (commented):
ok

SSGoku369:
ok

aaron-prindle:
This requires further discussion, currently this PR does not handle all of the kubecontext locations and edge cases.  Closing this PR for now

PR #6888, opened 2021-11-21
feat: show image tag in console logs

davidxia:
for ease of debugging.

Before

```
Checking cache...
 - my-image: Not found. Building
```

After

```
Checking cache...
 - my-image:351bcbd-dirty: Not found. Building
```


davidxia:
I see. thanks for explaining!

PR #6890, opened 2021-11-22
fix: correct English wording in error message

davidxia:
Should be "Try running" instead of "Trying running."


briandealwis (approved):
Thanks!

PR #6893, opened 2021-11-22
doc(debug): add why breakpoints may fail and clean up manual configuration

briandealwis:
@quoctruong, @etanshaul: should we add anything else here for Go, like `-trimpath`?

### DIFF
--- a/docs/content/en/docs/workflows/debug.md
+++ b/docs/content/en/docs/workflows/debug.md
@@ -420,7 +420,7 @@ Go, Python, and JavaScript debugging in IDEA Ultimate requires the
 corresponding language plugin to be installed. If it is not installed,
 the debugger will not attach and breakpoints will not be hit.
 
-**Breakpoints on startup path are not hit.**
+**Breakpoints are not hit during program startup.**
 Skaffold configures the containers to run-on-start (sometimes called
 _continue_ mode), such that the containers do not wait for the
 debugger to connect.  The container will have likely finished its
@@ -429,10 +429,10 @@ the breakpoints, and so breakpoints in the normal startup path are
 unlikely to be hit.
 
 **File mapping misconfiguration between local and container filesystems.**
-IDE debugger integrations generally require configuring a _source map_
+Some language runtimes require configuring the IDE with a _source map_
 to map files in the container to their corresponding local files.
 If this mapping is incorrect, then the remote debugging runtime
-will not be able to find the source file for the breakpoint.
+will not be able to find the source file for the breakpoint. The Cloud Code plugins generally prompt the user when required.
 
 - Some builds use multi-stage dockerfiles to compile the binary using
   a build-time image and then copy the binary into a smaller image
--- a/docs/content/en/docs/workflows/debug.md
+++ b/docs/content/en/docs/workflows/debug.md
@@ -405,16 +405,46 @@ WARN[0005] Image "image-name" not configured for debugging: unable to determine 
 
 See the language runtime section details on how container images are recognized.
 
+### Why aren't my breakpoints being hit?
+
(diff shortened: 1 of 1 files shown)


etanshaul:
What do you think about adding IDE specific instructions. Some JB specific things I can think of:

- GoLand (and Ultimate + Go plugin) does not support source mapping. Rather, the application must be a Go modules based app AND the user must explicitly configure the IDE to use Go modules in the IDE settings.
- Go, Python, and Javascript debugging in IDEA Ultimate requires the corresponding language plugin to be installed. If it is not installed, the debugger will not attach and breakpoints will not be hit (though the `skaffold debug` session will still start up).

I don't recall what we found with regards to `-trimpath`. I vaguely remember a conversation around this where it turned out to be useful in IntelliJ but I don't recall the specifics. Will update if I can dig it up.

briandealwis:
PTAL @etanshaul

docs/content/en/docs/workflows/debug.md
```diff
@@ -405,16 +413,52 @@ WARN[0005] Image "image-name" not configured for debugging: unable to determine
...
 See the language runtime section details on how container images are recognized.
 
+### Why aren't my breakpoints being hit?
+
+**Missing required language extension.**
+Go, Python, and JavaScript debugging in IDEA Ultimate requires the
+corresponding language plugin to be installed. If it is not installed,
+the debugger will not attach and breakpoints will not be hit.
+
+**Breakpoints on startup path are not hit.**
```
tejal29, line 423: What do you mean by startup path here?

docs/content/en/docs/workflows/debug.md
```diff
@@ -405,16 +413,52 @@ WARN[0005] Image "image-name" not configured for debugging: unable to determine
...
+debugger to connect.  The container will have likely finished its
+startup by the time the debugger is able to connect and configure
+the breakpoints, and so breakpoints in the normal startup path are
+unlikely to be hit.
+
+**File mapping misconfiguration between local and container filesystems.**
+IDE debugger integrations generally require configuring a _source map_
+to map files in the container to their corresponding local files.
+If this mapping is incorrect, then the remote debugging runtime
+will not be able to find the source file for the breakpoint.
```
tejal29, line 435: Is there a way to manually fix this mapping ? any pointers on how users can look into this setting and where it is configured.

docs/content/en/docs/workflows/debug.md
```diff
@@ -405,16 +413,52 @@ WARN[0005] Image "image-name" not configured for debugging: unable to determine
...
+
+- Some builds use multi-stage dockerfiles to compile the binary using
+  a build-time image and then copy the binary into a smaller image
+  intended for deployment.  In these cases, the source mapping must
+  use the build-time paths.  For example, many Go builds use a `golang`
```
tejal29, line 440: ah looks like you have documented resolution here.

tejal29: Can we add more details around how to fix it?

docs/content/en/docs/workflows/debug.md
briandealwis: I meant the container execution initialization path (e.g., from main() to listening)

docs/content/en/docs/workflows/debug.md
briandealwis: Cloud Code generally prompts in these situations.  It really is IDE-by-IDE

docs/content/en/docs/workflows/debug.md
briandealwis: Retitled: let me know if that's still unclear.

tejal29 (approved):

PR #6896, opened 2021-11-23
docs: how to skip image building in hot-reload

davidxia:
example. Took me a while to figure this out. Might be useful
for others.

### DIFF
--- a/examples/hot-reload/README.md
+++ b/examples/hot-reload/README.md
@@ -18,3 +18,16 @@ skaffold dev
 * Make some changes to `python/src/app.py`:
     * The file will be synchronized to the cluster
     * `flask` will restart the application
+
+#### Skip image build entirely
+
+If you have a pre-built image and want skaffold to skip building your image entirely
+and only sync code to the cluster, add the following to `skaffold.yaml`.
+
+```yaml
+build:
+  local:
+    tryImportMissing: true
+```
+
+Then add `--tag [image-tag]` to your `skaffold dev` command.


aaron-prindle (approved):
Thanks for updating the docs here!  LGTM

davidxia:
@aaron-prindle thanks for the quick review! While I have you, do you know how to completely skip having a Dockerfile and Docker daemon as well?

My use case is I'm trying to provide a tool to easily create cloud-based dev envs for users who might not understand or have installed anything related to Docker, e.g. ML researchers who just want to get an environment and start coding. I want to give them a pre-built image with pre-installed tools/libs/deps. I just want skaffold to sync code from the local directory in which its run to the remote container without needing a Dockerfile (don't need any image to be built) or Docker daemon (don't want to require users to have that installed).

Is there a way to do this?

P.S. Feel free to chat with me (dxia) on kubernetes.slack.com as well.

briandealwis (changes requested):
hey @davidxia, I don't think this addition benefits the example.  There are some subtleties to `tryImportMissing` that isn't captured here (specifically around tags).   It would be better written up as a tutorial in `docs/content/en/docs/tutorials`.

Aside: Changes to examples also need to be copied into `integration/examples`.  The copies in `examples` are for the current release whereas `integration/examples` are for the next release.  `examples` is replaced with `integration/examples` on the next release.

davidxia:
@briandealwis Thanks for pointing out. I see the tutorial main page has a link to the GitHub dir of examples and links to several tutorial pages. Do you think my use case would be useful as a separate example in this repo (don’t think I have enough content to make it an actual tutorial page)?

If so, I could put what I have in https://github.com/davidxia/skaffold-no-docker-just-sync in there with caveat that skaffold will always restart the Pod on startup. I have the hard requirement that Pods that are running are not restarted but only found and synced to by Skaffold. But maybe my example can help others.

PR #6906, opened 2021-11-24
ci: add kokoro scripts for releases

MarlonGamez:
**Related**: https://github.com/GoogleContainerTools/skaffold/issues/6348

**Description**
This PR modifies the `$(BUILD_DIR)/$(PROJECT)-%:` target of our Makefile to build things locally, rather then using docker to build the binaries. This will allow us to build these binaries in kokoro using this target without having to use docker and the `dockercore/golang-cross` image.


Makefile
```diff
@@ -118,13 +119,33 @@ $(BUILD_DIR)/$(PROJECT)-%: $(STATIK_FILES) $(GO_FILES) $(BUILD_DIR) deploy/cross
 	(cd `dirname $@`; shasum -a 256 `basename $@`) | tee $@.sha256
 	file $@ || true
 
+.PHONY: kokoro-cross
+kokoro-cross: $(foreach platform, $(SUPPORTED_PLATFORMS), $(KOKORO_BUILD_DIR)/$(PROJECT)-$(platform))
+
+$(KOKORO_BUILD_DIR)/$(PROJECT)-%: $(STATIK_FILES) $(GO_FILES) $(KOKORO_BUILD_DIR)
```
briandealwis, line 125: You'll need to revise the values on lines 68-73

Makefile
```diff
@@ -118,13 +119,33 @@ $(BUILD_DIR)/$(PROJECT)-%: $(STATIK_FILES) $(GO_FILES) $(BUILD_DIR) deploy/cross
...
+	$(eval os = $(firstword $(subst -, ,$*)))
+	$(eval arch = $(lastword $(subst -, ,$(subst .exe,,$*))))
+	$(eval ldflags = $(GO_LDFLAGS) $(patsubst %,-extldflags \"%\",$(LDFLAGS_$(os))))
+	$(eval tags = $(GO_BUILD_TAGS_$(os)) $(GO_BUILD_TAGS_$(os)_$(arch)))
+	GOOS=$(os) GOARCH=$(arch) CGO_ENABLED=1 go build -tags "$(tags)" -ldflags "$(ldflags)" -o $@ ./cmd/skaffold
+	(cd `dirname $@`; shasum -a 256 `basename $@`) | tee $@.sha256
+
+.PHONY: upload-kokoro-release
+upload-kokoro-release: $(BUILD_DIR)/VERSION
+	gsutil -m cp $(KOKORO_ARTIFACTS_DIR)/build/$(PROJECT)-* $(GSC_RELEASE_PATH)/
```
briandealwis, line 135: Do you think GSC was a typo (GCS)? 😄   Dare we fix it?

hack/kokoro/release.sh
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+export GO_VERSION=1.17.3
+
+# Replace go installation with version of our choosing
+rm -Rf /usr/local/go && mkdir /usr/local/go
+curl --fail --show-error --silent --location https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz \
```
briandealwis, line 23: It would be good to compare the SHA256 values to a known value.

Makefile
```diff
@@ -118,13 +119,33 @@ $(BUILD_DIR)/$(PROJECT)-%: $(STATIK_FILES) $(GO_FILES) $(BUILD_DIR) deploy/cross
 	(cd `dirname $@`; shasum -a 256 `basename $@`) | tee $@.sha256
 	file $@ || true
```
briandealwis, line 120: So we could just replace this `cross`, since we won't be using it anymore.  Then there's no kokoro-way and cross-way — there's just the one.

Makefile
```diff
@@ -118,13 +119,33 @@ $(BUILD_DIR)/$(PROJECT)-%: $(STATIK_FILES) $(GO_FILES) $(BUILD_DIR) deploy/cross
...
+$(KOKORO_BUILD_DIR)/$(PROJECT)-%: $(STATIK_FILES) $(GO_FILES) $(KOKORO_BUILD_DIR)
+	$(eval os = $(firstword $(subst -, ,$*)))
+	$(eval arch = $(lastword $(subst -, ,$(subst .exe,,$*))))
+	$(eval ldflags = $(GO_LDFLAGS) $(patsubst %,-extldflags \"%\",$(LDFLAGS_$(os))))
+	$(eval tags = $(GO_BUILD_TAGS_$(os)) $(GO_BUILD_TAGS_$(os)_$(arch)))
+	GOOS=$(os) GOARCH=$(arch) CGO_ENABLED=1 go build -tags "$(tags)" -ldflags "$(ldflags)" -o $@ ./cmd/skaffold
+	(cd `dirname $@`; shasum -a 256 `basename $@`) | tee $@.sha256
+
+.PHONY: upload-kokoro-release
+upload-kokoro-release: $(BUILD_DIR)/VERSION
```
briandealwis, line 134: I was thinking that if we're replacing the release process to use kokoro, then we should *remove* the release stuff here so as to prevent someone from accidentally releasing outside of kokoro.  So maybe this copying process should be moved into hack/kokoro/?

Makefile
MarlonGamez: ah yeah, I think I'll just remove
```
GO_VERSION_darwin_arm64 = 1.16beta1
CGO_ENABLED_darwin_arm64 = 0
```
since we should use a consistent version now and always enable CGO

Makefile
MarlonGamez: I'll move these into a makefile in the `hack/kokoro` directory, that way we can still take advantage of some of the handy makefile features

hack/kokoro/release.sh
MarlonGamez: done 👍🏼

Makefile
MarlonGamez: I was going to make a follow up PR to remove the code related to our old release pipeline once we can verify that the kokoro way works for us. Would you like me to remove it now?

hack/kokoro/release.sh
```diff
@@ -0,0 +1,36 @@
...
+curl --fail --show-error --silent --location https://go.dev/dl/go${GO_VERSION}.darwin-amd64.tar.gz -o go${GO_VERSION}.darwin-amd64.tar.gz
+if [ "$(shasum -a 256 go${GO_VERSION}.darwin-amd64.tar.gz | cut -d ' ' -f 1)" != ${GO_VERSION_SHA} ]; then
+  echo "failed to verify sha of downloaded go installation"
+  exit 1
+fi
+
+tar xz --directory=/usr/local/go --strip-components=1
+
+# Make binaries
+make -f Makefile -f hack/kokoro/Makefile kokoro-cross
```
briandealwis, line 33: Doesn't this need to define SUPPORTED_PLATFORMS and PROJECT?

But do we need to have a separate makefile?  Can't we just use `make cross` and then copy `out/*` instead of `build/*`?

hack/kokoro/upload.sh
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+make -f Makefile -f hack/kokoro/Makefile upload-kokoro-release
```
briandealwis, line 19: Isn't there a danger of using make that the file versions may be earlier than VERSION and so we don't upload the files?  Couldn't we just copy in the `gsutil cp` commands into here?

hack/kokoro/release.sh
MarlonGamez: we can't use make cross as it uses docker to build these binaries

Makefile
```diff
@@ -66,7 +66,7 @@ GO_BUILD_TAGS_windows = release
 
 # darwin/arm64 requires Go 1.16beta1 or later; dockercore/golang-cross
 # doesn't have a recent macOS toolchain so disable CGO and use
-# github.com/rjeczalik/notify's kqueue support. 
+# github.com/rjeczalik/notify's kqueue support.
 GO_VERSION_darwin_arm64 = 1.16beta1
 CGO_ENABLED_darwin_arm64 = 0
```
briandealwis, line 71: We can delete this

briandealwis (approved):

Issue #6126, opened 2021-07-02
`skaffold render` sometimes exits early


ekuefler:
### Expected behavior

`skaffold render` should finish its build and write its output file before exiting.

### Actual behavior

Rarely, the `skaffold` process appears to exit (with a 0 status code) while the build is still in progress, without writing any output.

### Information

- Skaffold version: v1.25.0
- Operating system: Linux (via Google Cloud Build)
- Installed via: skaffold.dev
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta14
kind: Config
metadata:
  name: environments

build:
  local:
    tryImportMissing: true
    push: true
  tagPolicy:
    inputDigest: {}
  artifacts:
  - image: gcr.io/devops-environments/manager
    docker:
      dockerfile: cmd/manager/Dockerfile
deploy:
  kubectl:
    remoteManifests:
    - environments-system:deployment/manager

profiles:
- name: release
  activation:
  - command: render
  build:
    local:
      push: true
      concurrency: 0
    tagPolicy:
      customTemplate:
        template: "{{.DATE}}-{{.GIT}}"
        components:
          - name: DATE
            dateTime:
              format: 20060102-150405
              timezone: America/Los_Angeles
          - name: GIT
            gitCommit:
              variant: AbbrevCommitSha
    artifacts:
    - image: gcr.io/devops-environments/clustergateway
      docker:
        dockerfile: cmd/clustergateway/Dockerfile
    - image: gcr.io/devops-environments/clusterreporter
      docker:
        dockerfile: cmd/clusterreporter/Dockerfile
    - image: gcr.io/devops-environments/manager
      docker:
        dockerfile: cmd/manager/Dockerfile
    - image: gcr.io/devops-environments/resourcebackup
      docker:
        dockerfile: cmd/resourcebackup/Dockerfile
    - image: gcr.io/devops-environments/sshpasswordchecker
      docker:
        dockerfile: cmd/sshpasswordchecker/Dockerfile
    - image: gcr.io/devops-environments/volumereclaimer
      docker:
        dockerfile: cmd/volumereclaimer/Dockerfile
    - image: gcr.io/devops-environments/webhookserver
      docker:
        dockerfile: cmd/webhookserver/Dockerfile
  deploy:
    kubectl: {}
    kustomize:
      paths:
      - config/release
```

### Notes

[...]

PR #6910, opened 2021-11-25
fix: distinguish docker build errors from streaming errors

briandealwis:
In trying to reproduce #6126, I noticed that docker build errors are oddly prefixed with "`unable to stream build output:`".
```
Canceled build for leeroy-app
unable to stream build output: The command '/bin/sh -c exit 1' returned a non-zero code: 1. Please fix the Dockerfile and try again..
```

It turns out that the Docker code for parsing and interpreting the JSON message stream from the Docker daemon parses build failures as a `JSONError` object, which behaves like an `error`.  This PR changes our error-handling code to return such errors directly.

### DIFF
--- a/pkg/skaffold/docker/image.go
+++ b/pkg/skaffold/docker/image.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


briandealwis:
What's odd here is that this error can appear long after the output started.  Maybe `docker reports:` is a better prefix?

pkg/skaffold/docker/image.go
```diff
@@ -364,6 +365,10 @@ func (l *localDaemon) Build(ctx context.Context, out io.Writer, workspace string
 	}
 
 	if err := streamDockerMessages(out, resp.Body, auxCallback); err != nil {
+		var jm *jsonmessage.JSONError
+		if errors.As(err, &jm) {
+			return "", err
```
briandealwis, line 370: Maybe this should be 
```suggestion
			return "", fmt.Errorf("docker build failure: %w", err)
```

tejal29: I like it better too.

tejal29 (approved):

tejal29:
Not sure, why `blocked` label was added. Feel free to remove it!

briandealwis:
I had it blocked pending the additional message.

PR #6911, opened 2021-11-25
fix: report build failures inline for failed concurrent builds

briandealwis:
Fixes: #6126 

**Description**
This PR causes Skaffold to report build successes and failures inline as part of the build logs, just as we report cancelled builds.
```
 ---> 40349a2425ef
Step 3/8 : RUN exit 1
 ---> Running in cf6bf19cb78e
Build [leeroy-web] failed: The command '/bin/sh -c exit 1' returned a non-zero code: 1
Building [leeroy-app]...
```

Previously such an error would be passed up and reported by the [top-level command instance](https://github.com/GoogleContainerTools/skaffold/blob/6e95e61c14607df62fbed0133d716160345d39d6/cmd/skaffold/app/skaffold.go#L59-L60).  This works for serial builds (`concurrency = 1`, the default with local builds) and single-artifact builds as the failed build would be the last build reported.  But it makes diagnosing builds failures very difficult with concurrent builds as Skaffold buffers the logs of each build and reports them as each completes.  When a build fails, the log is copied out and any other builds in progress are cancelled. These cancelled logs may appear to have been interrupted mid-stream.  And the actual build failure is easily lost in the sea of logs.

This change does not report the actual error on serial or single-artifact builds to avoid a visual stutter as [Skaffold also reports the build failure in its final output](https://github.com/GoogleContainerTools/skaffold/blob/6e95e61c14607df62fbed0133d716160345d39d6/cmd/skaffold/app/skaffold.go#L55-L60), and leads to output like:
```
Building [base]...
Sending build context to Docker daemon  2.048kB
Step 1/4 : FROM gcr.io/distroless/base
 ---> 64bbfbb81976
Step 2/4 : RUN exit 1
 ---> Running in b605bfb58ba0
Build [base] failed: unable to stream build output: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/bin/sh": stat /bin/sh: no such file or directory: unknown. Please fix the Dockerfile and try again..
unable to stream build output: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/bin/sh": stat /bin/sh: no such file or directory: unknown. Please fix the Dockerfile and try again..
```

### DIFF
--- a/pkg/skaffold/build/scheduler.go
+++ b/pkg/skaffold/build/scheduler.go
@@ -45,6 +45,7 @@ type scheduler struct {
 	logger          logAggregator
 	results         ArtifactStore
 	concurrencySem  countingSemaphore
+	reportFailure   bool
 }
 
 func newScheduler(artifacts []*latestV1.Artifact, artifactBuilder ArtifactBuilder, concurrency int, out io.Writer, store ArtifactStore) *scheduler {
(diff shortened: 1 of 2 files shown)


tejal29 (approved):
fix tests and merrge!

tejal29 (approved):

PR #5723, opened 2021-04-26
Fix Helm deployment check to only retrieve deployed YAML

briandealwis:
Fixes: #5484

**Description**
The Helm deployer calls `helm get all <chart>` to get the release record for the deployed chart and then parses these release record to get the Kubernetes YAML.  

<details>
<summary>Example Helm release record (`helm get all skaffold-helm-default`)</summary>

```console
$ helm get all skaffold-helm-default
NAME: skaffold-helm-default
LAST DEPLOYED: Sat Apr 24 00:22:07 2021
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
USER-SUPPLIED VALUES:
image: gcr.io/k8s-skaffold/skaffold-helm:0a46319f93bc0edf6bea77fa3e9ab287979d641939645abe1f45d9503550e93b

COMPUTED VALUES:
image: gcr.io/k8s-skaffold/skaffold-helm:0a46319f93bc0edf6bea77fa3e9ab287979d641939645abe1f45d9503550e93b
ingress:
  annotations: null
  enabled: true
  hosts: null
  tls: null
replicaCount: 1
resources: {}
service:
  externalPort: 80
  internalPort: 80
  name: nginx
  type: ClusterIP

HOOKS:
MANIFEST:
---
# Source: skaffold-helm/templates/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: skaffold-helm-default-skaffold-helm
  labels:
    app: skaffold-helm
    chart: skaffold-helm-0.1.0
    release: skaffold-helm-default
    heritage: Helm
spec:
  type: ClusterIP
  ports:
    - port: 80
      targetPort: 80
      protocol: TCP
      name: nginx
  selector:
    app: skaffold-helm
    release: skaffold-helm-default
---
# Source: skaffold-helm/templates/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: skaffold-helm-default
  labels:
    app: skaffold-helm
    chart: skaffold-helm-0.1.0
    release: skaffold-helm-default
    heritage: Helm
spec:
  replicas: 1
  selector:
    matchLabels:
      app: skaffold-helm
      release: skaffold-helm-default
  template:
    metadata:
      labels:
        app: skaffold-helm
        release: skaffold-helm-default
    spec:
      containers:
        - name: skaffold-helm
[...]

Issue #5484, opened 2021-03-03
Helm chart containing NOTES.txt leads to the "error parsing object from string: collecting namespaces: reading Kubernetes YAML:..."


prokher:
### Expected behavior

Helm charts with NOTES.txt file should work well and does not produce any extra errors in the log. 

### Actual behavior

Using Skaffold with Helm chart with NOTES.txt (containing just "Hello world!") leads to the error:
```
error parsing object from string: collecting namespaces: reading Kubernetes YAML: yaml: line 24: could not find expected ':' 
```

Removing NOTES.txt from the char fixes the issue.

### Information

- Helm's NOTES.txt file: https://helm.sh/docs/chart_template_guide/notes_files/
- Skaffold version: v1.20.0
- Operating system: Ubuntu 20.04
- Installed via: skaffold.dev

### Steps to reproduce the behavior

1. Create a Helm chart with NOTES.txt.
2. Configure Skaffold to use this Helm chart.
3. `skaffold dev`

### Extra

As far as I see, the problem is in the function `pkg/skaffold/deploy/helm/parse.go:parseReleaseInfo()` which expects an output of `helm get all <release name>` to be valid YAML file which is not the case when there is NOTES.txt in the Helm chart.

briandealwis:
@prokher would you be able to submit a patch?

prokher:
@briandealwis, unfortunately, no. I just made assumption about the place where error happens. Fixing this is a different story, for that I need to better understand decomposition, at least the local one.

briandealwis:
I think we can just call `helm get all <xxx> --template '{{.Release.Manifest}}'` to return just the Kubernetes YAML.

PR #6913, opened 2021-11-26
refactor: remove vestiges of parsing release info text from helm deployment

briandealwis:
This PR is a minor refactoring of the Helm post-deployment parsing to reflect the reality after #5723.

Prior to #5723, our Helm deployer would invoke `helm get all <chart>` to obtain the deployed manifests.  This output some semi-structured text with release information, followed by the manifests.  We discarded the release info and then parsed out the remaining YAML.  But this could fail (#5484) as there was no clean separation between the release info and the YAML.
<details>
<summary>example showing output from `helm get all skaffold-helm`</summary>

```
$ helm get all skaffold-helm -n namespace
NAME: skaffold-helm
LAST DEPLOYED: Fri Nov 26 12:59:05 2021
NAMESPACE: namespace
STATUS: deployed
REVISION: 1
TEST SUITE: None
USER-SUPPLIED VALUES:
null

COMPUTED VALUES:
image: skaffold-helm:latest
replicaCount: 2

HOOKS:
MANIFEST:
apiVersion: apps/v1
kind: Deployment
metadata:
[...elided...]
```

</details>

Since we don't actually use the Helm release info text, #5723 changed our deployer to instead use `helm get all <chart> --template {{.Release.Manifest}}` to only output the Kubernetes manifests.  But there were some vestiges of the release-info parsing still present.

This PR removes the last bits of this release-info parsing.  In so doing, I discovered that our tests weren't providing an actual deployed manifest to `helm get all`, and so the namespaces were not being properly reported, and so our deploy-with-namespace tests weren't actually checking that the right namespaces were being reported!

### DIFF
--- a/pkg/skaffold/deploy/helm/deploy.go
+++ b/pkg/skaffold/deploy/helm/deploy.go
@@ -579,7 +579,7 @@ func (h *Deployer) deployRelease(ctx context.Context, out io.Writer, releaseName
 		return nil, userErr("get release", err)
 	}
 
-	artifacts := parseReleaseInfo(opts.namespace, bufio.NewReader(&b))
+	artifacts := parseReleaseManifests(opts.namespace, bufio.NewReader(&b))
 	return artifacts, nil
 }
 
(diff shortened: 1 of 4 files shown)


MarlonGamez (approved):
LGTM

PR #6923, opened 2021-12-01
refactor(lint): remove unused regexp linter

aaron-prindle:
This PR removes the unused RegExp linter and related logic.  As I am working on some additional refactoring, it makes sense to remove this unused linter to minimize the amount of work involved.  We can add it back later if it is required (additional files need checking that we don't have a schema parser for, etc.)


MarlonGamez (approved):
lgtm

PR #6928, opened 2021-12-03
fix(ko): Do not print image name to stdout

halvards:
With this change, the `ko` builder no longer prints the image name to `stdout` by default.

Originally, this was added to the `ko` builder to mimic the behavior of the `ko` CLI. Other Skaffold image builders do not print the image name in this way.

The reason this is useful for existing `ko` CLI users is that some documented workflows rely on capturing the image name from `stdout`, see https://github.com/google/ko/blob/v0.9.3/README.md#build-an-image.

After some investigation, the better option seems to be to rely on the existing `--quiet` and `--output` Skaffold flags to format the output of the image name(s).

This change also updates the `ko` builder documentation to show existing `ko` CLI users how to capture the image name by providing an appropriate Go template to `--output`.

Fixes: #6835
Closes: #6836
Tracking: #6041


aaron-prindle (approved):
LGTM

MarlonGamez (approved):

Issue #6929, opened 2021-12-03
Docker builds with cliFlags should be build with docker CLI


briandealwis:
### Expected behavior

A skaffold.yaml with a docker build artifact specifying `cliFlags` should always use the `docker` CLI, overriding the `useDockerCLI` setting.
```
kind: Config
apiVersion: skaffold/v2beta24
build:
  artifacts:
  - image: example
    docker:
      cliFlags: ["--platform", "linux/amd64"]
```

### Actual behavior
```
$ skaffold build -v debug
...
Starting build...
Found [minikube] context, using local docker daemon.
Building [example]...
DEBU[0001] Running docker build: context: ., dockerfile: Dockerfile  subtask=example task=Build
DEBU[0003] Found dependencies for dockerfile: []         subtask=-1 task=DevLoop
Sending build context to Docker daemon  2.048kB
```

### Information

- Skaffold version: v1.34.0 (and still in HEAD)

PR #6930, opened 2021-12-03
fix: docker build for artifact with cliFlags should use docker CLI

briandealwis:
Fixes: #6929

**Description**
A Docker build for an artifact that sets `cliFlags` should always do a CLI build.


aaron-prindle (approved):
LGTM

Issue #6916, opened 2021-11-29
FR: `ko` builder to support envvars in `labels` config field


halvards:
### Desired behavior

When using Skaffold to publish a "releaseable" container image in a CI build, users often want to add image labels with information such as release tag, git SHA, and more.

Example: https://github.com/google/k8s-digester/blob/main/.github/workflows/release.yml#L136

### Current behavior

The `ko` builder [`labels` config field](https://skaffold.dev/docs/references/yaml/#build-artifacts-ko-labels) doesn't support expansion of environment variables.

### Design considerations

The [`flags` and `ldflags`](https://skaffold.dev/docs/references/yaml/#build-artifacts-ko-flags) `ko` builder config fields support expansion of environment variables. However, this expansion is currently handled by the upstream [`ko` implementation](https://skaffold.dev/docs/pipeline-stages/builders/ko/#build-flags), which uses a slightly [different syntax](https://github.com/google/ko#overriding-go-build-settings) to other [Skaffold config fields that support envvar expansion](https://skaffold.dev/docs/environment/templating/).

Skaffold: `{{.FOO}}`

`ko`: `{{.Env.FOO}}`

Would we accept different envvar handling and syntax for different fields within the same builder config?

Or should we consider moving envvar expansion of the `flags` and `ldflags` fields to Skaffold?

Tracking: #6041

PR #6944, opened 2021-12-07
feat(ko): Enable templating of `labels` and `env`

halvards:
This change enables `ko` builder users to substitute environment variable values in the `labels` and `env` config fields.

These fields are used for image labels and build-time environment variables, respectively.

Envvar templating of image labels can be used to add information such as the Git commit SHA to the image, see #6916.

Also, environment variable expansion of `flags` and `ldflags` in the `ko` builder configuration now supports Skaffold's templating syntax, for consistency.

For backwards compatibility, `ko`'s templating syntax still works with `flags` and `ldflags`:

Skaffold: `{{.FOO}}`

`ko`: `{{.Env.FOO}}`

Tracking: #6041
Fixes: #6916

### DIFF
--- a/pkg/skaffold/build/ko/builder_test.go
+++ b/pkg/skaffold/build/ko/builder_test.go
@@ -17,10 +17,14 @@ limitations under the License.
 package ko
 
 import (
+	"fmt"
+	"os"
 	"path/filepath"
 	"testing"
 
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/ko/pkg/build"
+	"github.com/google/ko/pkg/commands/options"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
 	latestV1 "github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest/v1"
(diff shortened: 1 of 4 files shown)


pkg/skaffold/build/ko/builder_test.go
```diff
@@ -46,128 +51,108 @@ func TestBuildOptions(t *testing.T) {
...
 			},
-			wantLabels: []string{"foo=bar", "frob=baz"},
 		},
 	}
 	for _, test := range tests {
 		testutil.Run(t, test.description, func(t *testutil.T) {
-			bo, err := buildOptions(&test.artifact, test.runMode)
+			os.Setenv(testKoBuildOptionsEnvVar, test.envVarValue)
+			gotBo, err := buildOptions(&test.artifact, test.runMode)
+			os.Unsetenv(testKoBuildOptionsEnvVar)
```
tejal29, line 150: lets call this in defer.

pkg/skaffold/build/ko/builder_test.go
```diff
@@ -46,128 +51,108 @@ func TestBuildOptions(t *testing.T) {
...
-			},
-			wantWorkingDirectory: "my-app-subdirectory",
-		},
-		{
-			description: "source dir",
-			artifact: latestV1.Artifact{
-				ArtifactType: latestV1.ArtifactType{
-					KoArtifact: &latestV1.KoArtifact{
-						Dir: "my-go-mod-is-here",
+			envVarValue: "baz",
```
tejal29, line 90: shd we add a test case for `{{.Env.baz}}`  here?

pkg/skaffold/build/ko/builder_test.go
halvards: Good point, done!

pkg/skaffold/build/ko/builder_test.go
halvards: Added.

tejal29 (approved):

PR #6955, opened 2021-12-09
feat(lint): added lib for getting yaml line #s for a given schema struct, sub-element and sub-fields.  supports patches & profiles

aaron-prindle:
This PR adds functionality for getting the line & column info for a skaffold schema object/field.  This is done by adding a new `YAMLInfos` field to SkaffoldConfigEntry:
```
type SkaffoldConfigEntry struct {
	*latestV1.SkaffoldConfig
	SourceFile   string
	SourceIndex  int
	IsRootConfig bool
	IsRemote     bool
	YAMLInfos    *configlocations.YAMLInfos
}
```
which has the information/functionality required to query the location of any field in the `*latestV1.SkaffoldConfig`.  Here is an example of how this can be used:

skaffold.yaml file:
```
apiVersion: skaffold/v2beta21
kind: Config
build:
  artifacts:
    - image: leeroy-web
```

`Locate` usage:
```
artifactImageNameLocation := c.YAMLInfos.LocateField(c.Build.Artifacts[i], "ImageName")
# artifactImageNameLocation.SourceFile, artifactImageNameLocation.StartLine, artifactImageNameLocation.EndLine, etc. now accessible  
```

The method used here has 2 phases.  During the initial construction of the `*latestV1.SkaffoldConfig` information related to what fields were overwritten by profiles and patch profiles is stored in `fieldPathElementsOverrodeByProfile` (these are the changes made in `pkg/skaffold/parser/config.go`). This map contains entries that map from the directory representation of a yaml field - ex:`/build/artifacts/0` to the override information of that field.  So something like `/build/artifacts/0` -> overriden by profileName:"profile-name" and patchProfile[0] (the first patch profile in the list of patch profiles)

This map is then passed to the `Parse` logic in `pkg/skaffold/configlocations/types.go` which uses reflection to recursively go through the populated `*latestV1.SkaffoldConfig`  as well as the kyaml Nodes (yaml nodes w/ location information), mapping each obj/field to the corresponding yaml node.  If a node is overwritten by a profile/patch, the logic there assigns the correct node from the `fieldPathElementsOverrodeByProfile` information

### DIFF
--- a/pkg/skaffold/parser/config.go
+++ b/pkg/skaffold/parser/config.go
@@ -28,6 +28,7 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/git"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/output/log"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/parser/configlocations"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/defaults"
 	sErrors "github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/errors"
@@ -80,7 +81,7 @@ func GetAllConfigs(ctx context.Context, opts config.SkaffoldOptions) ([]schemaUt
 func GetConfigSet(ctx context.Context, opts config.SkaffoldOptions) (SkaffoldConfigSet, error) {
 	cOpts := configOpts{file: opts.ConfigurationFile, selection: nil, profiles: opts.Profiles, isRequired: false, isDependency: false}
 	r := newRecord()
-	cfgs, err := getConfigs(ctx, cOpts, opts, r)
+	cfgs, fieldsOverrodeByProfile, err := getConfigs(ctx, cOpts, opts, r)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


pkg/skaffold/parser/config.go
```diff
@@ -80,7 +81,7 @@ func GetAllConfigs(ctx context.Context, opts config.SkaffoldOptions) ([]schemaUt
 func GetConfigSet(ctx context.Context, opts config.SkaffoldOptions) (SkaffoldConfigSet, error) {
 	cOpts := configOpts{file: opts.ConfigurationFile, selection: nil, profiles: opts.Profiles, isRequired: false, isDependency: false}
 	r := newRecord()
-	cfgs, err := getConfigs(ctx, cOpts, opts, r)
+	cfgs, fieldPathElementsOverrodeByProfile, err := getConfigs(ctx, cOpts, opts, r)
```
MarlonGamez, line 84: could this just be `fieldsOverrodeByProfile`? The name is a bit long/cumbersome right now and the meaning of `fieldPathElements` isn't entirely clear

pkg/skaffold/configlocations/types.go
```diff
@@ -0,0 +1,308 @@
+/*
```
MarlonGamez, line 1: could this package be put until `pkg/skaffold/config/packagename` or `pkg/skaffold/parser/packagename`? Might make it easier to find when looking for related code

pkg/skaffold/schema/profiles.go
```diff
@@ -309,44 +329,76 @@ func overlayStructField(config interface{}, profile interface{}) interface{} {
...
+			return overlayOneOfField(profileName, fieldPath, fieldPathElementsOverrodeByProfile, config, profile)
 		}
-		return overlayStructField(config, profile)
+		return overlayStructField(profileName, fieldPath, fieldPathElementsOverrodeByProfile, config, profile)
 	case reflect.Slice:
 		// either return the values provided in the profile, or the original values if none were provided.
 		if v.Len() == 0 {
 			return config
 		}
+		fieldPathElementsOverrodeByProfile["/"+strings.Join(fieldPath, "/")] = configlocations.YAMLOverrideInfo{
```
MarlonGamez, line 370: I think it might be easier/cleaner to replace these `strings.Join()` calls with [`path.Join()`](https://pkg.go.dev/path#Join) since it automatically joins with `/`

pkg/skaffold/parser/config.go
aaron-prindle: I have changed this to be s/fieldPathElementsOverrodeByProfile/fieldsOverrodeByProfile.  This was the initial name but I changed it when I realized I would need paths to fields (eg: /build/artifacts/0/image) vs the name of a field (image).  This was to let readers know that the keys are paths.  I am fine w/ this change though, it is done now

pkg/skaffold/schema/profiles.go
aaron-prindle: Done

pkg/skaffold/configlocations/types.go
aaron-prindle: moved `/pkg/skaffold/configlocations` under `parser` -> `pkg/skaffold/parser/configlocations`

pkg/skaffold/parser/configlocations/configlocations.go
```diff
@@ -0,0 +1,309 @@
...
+	if kind != reflect.Ptr {
+		log.Entry(context.TODO()).Infof("non pointer object passed to Locate: %v of type %T", obj, obj)
+		return MissingLocation()
+	}
+	if _, ok := m.YamlInfos[reflect.ValueOf(obj).Pointer()]; !ok {
+		log.Entry(context.TODO()).Infof("no map entry found when attempting Locate for %v of type %T and pointer: %d", obj, obj, reflect.ValueOf(obj).Pointer())
+		return MissingLocation()
+	}
+	node := m.YamlInfos[reflect.ValueOf(obj).Pointer()][key]
+	if _, ok := m.YamlInfos[reflect.ValueOf(obj).Pointer()][key]; !ok {
```
MarlonGamez, line 112: i think we don't need to do 2 reads here
```suggestion
	node, ok := m.YamlInfos[reflect.ValueOf(obj).Pointer()][key]
	if !ok {
```

MarlonGamez (approved):
lgtm, just have one nit but not necessary

pkg/skaffold/parser/configlocations/configlocations.go
aaron-prindle: Good catch, done

PR #6970, opened 2021-12-17
fix: bugs for skaffold.yaml files read in via stdin

aaron-prindle:
This fix modifies the behavior for 2 pieces of how skaffold.yaml's read in via stdin work:
1. Makes it so that the `SourceFile` field for a SkaffoldConfigEntry is `-` for files read in via `stdin` instead of /abs/path/skaffold.yaml/- which is what the value is currently. 
2. Makes it so that `util.ReadConfiguration` is idempotent for skaffold.yaml files read in via `stdin` by storing the []byte from `os.Stdin` that are read in.  Currently once the `Reader` that is os.Stdin is read, subsequent calls to `util.ReadConfiguration` are empty for `stdin` skaffold.yaml files

These changes are required for supporting `stdin` for skaffold.yaml's with the line # support that https://github.com/GoogleContainerTools/skaffold/pull/6955 provides

### DIFF
--- a/pkg/skaffold/parser/config.go
+++ b/pkg/skaffold/parser/config.go
@@ -107,7 +107,7 @@ func getConfigs(ctx context.Context, cfgOpts configOpts, opts config.SkaffoldOpt
 		return nil, sErrors.ConfigParsingError(err)
 	}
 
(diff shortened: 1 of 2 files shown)


MarlonGamez (approved):
lgtm

Issue #4954, opened 2020-10-27
AbbrevCommitSha tagPolicy not respected with 50+ images


zmarouf:
### Expected behavior
skaffold build with an AbbrevCommitSha tagPolicy tags all the images with the abbreviated commits.

### Actual behavior
Some of the images get tagged with "latest" when there are more than 50 images to process.
![image](https://user-images.githubusercontent.com/4321056/97309837-dfe61380-18a5-11eb-9c00-dd7b104dd0ee.png)

### Information

- Skaffold version: 1.15.0
- Operating system: macOS 10.15.7 (Darwin Kernel Version 19.6.0)
- Contents of skaffold.yaml:

```yaml
---
apiVersion: skaffold/v1
kind: Config
profiles:
  - name: local
    build:
      tagPolicy:
        gitCommit:
          variant: AbbrevCommitSha
      local:
        push: false
      artifacts:
        - image: local/zmarouf/61
          docker:
            dockerfile: ./61/Dockerfile
        - image: local/zmarouf/59
          docker:
            dockerfile: ./59/Dockerfile
        - image: local/zmarouf/66
          docker:
            dockerfile: ./66/Dockerfile
        - image: local/zmarouf/50
          docker:
            dockerfile: ./50/Dockerfile
        - image: local/zmarouf/68
          docker:
            dockerfile: ./68/Dockerfile
        - image: local/zmarouf/57
          docker:
            dockerfile: ./57/Dockerfile
        - image: local/zmarouf/32
          docker:
            dockerfile: ./32/Dockerfile
        - image: local/zmarouf/35
          docker:
            dockerfile: ./35/Dockerfile
        - image: local/zmarouf/69
          docker:
            dockerfile: ./69/Dockerfile
        - image: local/zmarouf/56
          docker:
            dockerfile: ./56/Dockerfile
        - image: local/zmarouf/51
          docker:
            dockerfile: ./51/Dockerfile
        - image: local/zmarouf/58
          docker:
            dockerfile: ./58/Dockerfile
        - image: local/zmarouf/67
          docker:
            dockerfile: ./67/Dockerfile
        - image: local/zmarouf/60
          docker:
            dockerfile: ./60/Dockerfile
        - image: local/zmarouf/34
          docker:
            dockerfile: ./34/Dockerfile
        - image: local/zmarouf/33
          docker:
            dockerfile: ./33/Dockerfile
        - image: local/zmarouf/20
          docker:
[...]

PR #6976, opened 2021-12-22
fix: limit number of simultaneous taggers to GOMAXPROCS

briandealwis:
Fixes: #4954

**Description**
#4954 demonstrates a pathological case with a build with dozens of artifacts using the gitCommit tagger.  This causes a flood of calls to `git describe`, some of which fail due to `too many open files` and result in using the fall-back tagger.
```
 - local/zmarouf/78 -> DEBU[0001] generating tag: unable to find git commit: starting command /usr/bin/git describe --tags --always: fork/exec /usr/bin/git: too many open files  subtask=local/zmarouf/78 task=Build
DEBU[0001] Using a fall-back tagger                      subtask=local/zmarouf/78 task=Build
```

This patch uses a counting semaphore (using golang.org/x/sync's weighted semaphore) to restrict the number of simultaneous calls to `GOMAXPROCS` (initialized to the number of CPUs).  It's possible that a system may support more than this number of simultaneous calls, but our taggers ideally shouldn't be that slow.

A future optimization would be to either have the taggers use memoization somehow to allow reusing previously-computed values, or provide a tagger lifecycle to allow taggers to cache and discarding caches.

### DIFF
--- a/pkg/skaffold/runner/build.go
+++ b/pkg/skaffold/runner/build.go
@@ -24,6 +24,8 @@ import (
 	"runtime"
 	"time"
 
+	"golang.org/x/sync/semaphore"
+
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/cache"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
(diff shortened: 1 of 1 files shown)


aaron-prindle:
Seems there are lint errors:
```
pkg/skaffold/runner/build.go:25: File is not `gci`-ed with -local github.com/GoogleContainerTools/skaffold (gci)
	"time"
pkg/skaffold/runner/build.go:39: File is not `gci`-ed with -local github.com/GoogleContainerTools/skaffold (gci)

	"golang.org/x/sync/semaphore"
FAILED hack/golangci-lint.sh
```

aaron-prindle (approved):

PR #6977, opened 2021-12-22
feat(lsp): add virtual file system support to ReadConfiguration and use this in LSP for buffer files and lint

aaron-prindle:
this change modifies ReadConfiguration to use an afero Virtual file system - https://github.com/spf13/afero under the hood.  By default it uses an `OS` filesystem (identical to what was used prior under the hood) but this is modified when using the LSP to a memory mapped filesystem so that the LSP can use buffered file information (vs only linting on save, it can lint on each keystroke even if unsaved).  Using a VFS also allows for possibly refactoring tests to use in memory skaffold.yaml files (vs on disk) for speedup

One possible issue is that currently the `Fs` var is a global value and it is possible to have a race as there is no mutex.  As this is only changed for the LSP, I think this should be ok.  Adding a mutex might possibly slowdown any future requirements/desire for parellelizing config/file reading so it is not done atm

### DIFF
--- a/pkg/skaffold/lsp/document_manager.go
+++ b/pkg/skaffold/lsp/document_manager.go
@@ -17,35 +17,31 @@ limitations under the License.
 package lsp
 
 import (
-	"sync"
+	"github.com/spf13/afero"
 )
 
-// DocumentManager manages syncing documents for the LSP server
+// DocumentManager manages syncing memMapFs for the LSP server
 type DocumentManager struct {
-	documents map[string]string
-	mtx       sync.RWMutex
+	memMapFs afero.Fs
 }
 
 // NewDocumentManager creates a new DocumentManager object
-func NewDocumentManager() *DocumentManager {
+func NewDocumentManager(fs afero.Fs) *DocumentManager {
 	return &DocumentManager{
-		documents: make(map[string]string),
+		memMapFs: fs,
 	}
 }
 
-// UpdateDocument updates the string value for a document
+// UpdateDocument updates the string value for a memMapFs
 func (m *DocumentManager) UpdateDocument(documentURI string, doc string) {
-	m.mtx.Lock()
-	defer m.mtx.Unlock()
-	m.documents[documentURI] = doc
+	afero.WriteFile(m.memMapFs, documentURI, []byte(doc), 0644)
 }
 
-// GetDocument gets the string value for a document
+// GetDocument gets the string value for a memMapFs
 func (m *DocumentManager) GetDocument(documentURI string) string {
-	m.mtx.RLock()
-	defer m.mtx.RUnlock()
-	if doc, ok := m.documents[documentURI]; ok {
-		return doc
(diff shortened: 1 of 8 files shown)


pkg/skaffold/lsp/document_manager.go
```diff
@@ -17,35 +17,31 @@ limitations under the License.
...
 	}
 }
 
-// UpdateDocument updates the string value for a document
+// UpdateDocument updates the string value for a memMapFs
 func (m *DocumentManager) UpdateDocument(documentURI string, doc string) {
-	m.mtx.Lock()
-	defer m.mtx.Unlock()
-	m.documents[documentURI] = doc
+	afero.WriteFile(m.memMapFs, documentURI, []byte(doc), 0644)
```
MarlonGamez, line 37: super nit: can save some characters by using `m.memMapFs.WriteFile()` and `m.memMapFs.ReadFile()` down below. Also makes these calls more consistent with things like `os.ReadFile()`

MarlonGamez (commented):
Looking good, basically just have a nit that could apply to most of the afero function calls

pkg/skaffold/lsp/document_manager.go
aaron-prindle: IIUC, these convience methods are not built into the FS interface provides:
```
type Fs interface {
	// Create creates a file in the filesystem, returning the file and an
	// error, if any happens.
	Create(name string) (File, error)

	// Mkdir creates a directory in the filesystem, return an error if any
	// happens.
	Mkdir(name string, perm os.FileMode) error

	// MkdirAll creates a directory path and all parents that does not exist
	// yet.
	MkdirAll(path string, perm os.FileMode) error

	// Open opens a file, returning it or an error, if any happens.
	Open(name string) (File, error)

	// OpenFile opens a file using the given flags and the given mode.
	OpenFile(name string, flag int, perm os.FileMode) (File, error)

	// Remove removes a file identified by name, returning an error, if any
	// happens.
	Remove(name string) error

	// RemoveAll removes a directory path and any children it contains. It
	// does not fail if the path does not exist (return nil).
	RemoveAll(path string) error

	// Rename renames a file.
	Rename(oldname, newname string) error

	// Stat returns a FileInfo describing the named file, or an error, if any
	// happens.
	Stat(name string) (os.FileInfo, error)

	// The name of this FileSystem
	Name() string

	// Chmod changes the mode of the named file to mode.
	Chmod(name string, mode os.FileMode) error

	// Chown changes the uid and gid of the named file.
	Chown(name string, uid, gid int) error

	//Chtimes changes the access and modification times of the named file
	Chtimes(name string, atime time.Time, mtime time.Time) error
}

```
The WriteFile and ReadFile methods are helper wrapper and actually need to be called via `afero.*`

MarlonGamez: ah, my bad, I missed that. Thanks for clarification

MarlonGamez (approved):
lgtm 👍🏼

PR #7000, opened 2022-01-10
fix(ci): fix lts image build

MarlonGamez:
This PR removes the `release-lts` make target's dependency on the `cross` target. The `cross target is being removed as a dependency from targets as it's functionality is being moved to kokoro.


aaron-prindle (approved):
LGTM

tejal29 (approved):

Issue #6989, opened 2022-01-04
Add tagPolicy: gitCommit: variant: Branches


snickell:
Since https://github.com/GoogleContainerTools/skaffold/pull/6904 has been merged, `cacheFrom` will automatically munge an not-explicitly-tagged image name to append the current tagPolicy. This makes it potentially very useful for easy layer-build-caching of dev branches..... if the gitCommit tagPolicy permitted a Branch variant.

In a dev context when building large docker build on a non-local machine relying on `cacheFrom` pulls rather than a local build cache, image tags are important for exporting and importing the most-relevant-to-current build cache, and git tags are generally not relevant.

I would propose that:
```
tagPolicy:
  gitCommit:
    variant: 'Branches'
```

behave just like `variant: 'Tags'`, but will check if the current commit is part of a branch, and use that for the image tag, rather than using the git tag.

snickell:
As an alternative approach, perhaps the set of "skaffold created special env vars" could be expanded to include git branch, git commit, etc: https://skaffold.dev/docs/environment/templating/

This would allow very nice control of git related tagging (including composite tags) using the `envTemplate` variant

MarlonGamez:
@snickell thanks for opening up this issue, I don't know that this will be prioritized by the team but we're always open to PRs :)

snickell:
I'd be up to implement a PR if there was Skaffold team consensus that (at least in principal) this was a desirable feature

gsquared94:
> I would propose that:
> 
> ```
> tagPolicy:
>   gitCommit:
>     variant: 'Branches'
> ```
> 
> behave just like `variant: 'Tags'`, but will check if the current commit is part of a branch, and use that for the image tag, rather than using the git tag.

> I'd be up to implement a PR if there was Skaffold team consensus that (at least in principal) this was a desirable feature

I'd say go for it 👍

vikas027:
> I'd be up to implement a PR if there was Skaffold team consensus that (at least in principal) this was a desirable feature

Hey @snickell ,  This is indeed a popular request in the [slack](https://kubernetes.slack.com/archives/CABQMSZA6) community :)

snickell:
See #7006 for a PR that implements the proposal here

cyraid:
Sorry for the extremely late bump, but is there a way to specify a specific branch? Like if you always want to get the latest commit of "release"?

PR #7006, opened 2022-01-12
feat: add 'branches' git tag variant

snickell:
Fixes: #6989 

**Description**
Implements a new git tagpolicy variant `Branches`, which can be used as follows:
```
tagPolicy:
  gitCommit:
    variant: 'Branches'
```

When this tagPolicy is active, when currently tracking a git branch, the branch's name (`git branch --show-current`) with slashes replaced with underscores will be used as the tag, otherwise the current short sha commit will be used.

### DIFF
--- a/pkg/skaffold/tag/git_commit.go
+++ b/pkg/skaffold/tag/git_commit.go
@@ -44,6 +44,7 @@ var variants = map[string]func(context.Context, string) (string, error){
 	"abbrevcommitsha": gitAbbrevcommitsha,
 	"treesha":         gitTreesha,
 	"abbrevtreesha":   gitAbbrevtreesha,
+	"branches":        gitBranches,
 }
 
 // NewGitCommit creates a new git commit tagger. It fails if the tagger variant is invalid.
@@ -133,6 +134,15 @@ func gitAbbrevtreesha(ctx context.Context, workingDir string) (string, error) {
 	}
 
 	return runGit(ctx, workingDir, "rev-parse", "--short", "HEAD:"+gitPath+"/")
+}
+
+func gitBranches(ctx context.Context, workingDir string) (string, error) {
+	gitBranch, err := runGit(ctx, workingDir, "branch", "--show-current")
+	if err != nil {
+		return gitAbbrevcommitsha(ctx, workingDir)
+	}
+
+	return gitBranch, nil
 }
 
 func getGitPathToWorkdir(ctx context.Context, workingDir string) (string, error) {
--- a/pkg/skaffold/tag/git_commit_test.go
+++ b/pkg/skaffold/tag/git_commit_test.go
@@ -46,6 +46,7 @@ func TestGitCommit_GenerateTag(t *testing.T) {
 		variantAbbrevCommitSha string
 		variantTreeSha         string
 		variantAbbrevTreeSha   string
+		variantBranches        string
 		createGitRepo          func(string)
 		subDir                 string
 		shouldErr              bool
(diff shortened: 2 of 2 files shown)


pkg/skaffold/tag/git_commit.go
```diff
@@ -135,6 +136,15 @@ func gitAbbrevtreesha(ctx context.Context, workingDir string) (string, error) {
 	return runGit(ctx, workingDir, "rev-parse", "--short", "HEAD:"+gitPath+"/")
 }
 
+func gitBranches(ctx context.Context, workingDir string) (string, error) {
+	gitBranch, err := runGit(ctx, workingDir, "branch", "--show-current")
+	if err != nil {
+		return gitAbbrevcommitsha(ctx, workingDir)
```
gsquared94, line 142: can you add a warning message here that `branches` tagger failed and we're falling back to `abbrevcommitsha`

pkg/skaffold/tag/git_commit_test.go
```diff
@@ -85,12 +88,29 @@ func TestGitCommit_GenerateTag(t *testing.T) {
...
+			createGitRepo: func(dir string) {
+				gitInit(t, dir).
+					write("source.go", "code").
+					add("source.go").
+					commit("initial").
+					branch("branch")
+			},
+		},
+		{
+			description:            "dirty  with tag containing a slash",
```
gsquared94, line 107: `dirty worktree` ?

pkg/skaffold/tag/git_commit_test.go
```diff
@@ -85,12 +88,29 @@ func TestGitCommit_GenerateTag(t *testing.T) {
 			},
 		},
 		{
-			description:            "dirty worktree with tag containing a slash",
+			description:            "clean worktree with a branch",
```
gsquared94, line 91: can you add a test case for branch name with a slash like `branch/foo`? The tag should then be `branch_foo`.

gsquared94 (approved):
lgtm, except one typo and one good to have testcase.

pkg/skaffold/tag/git_commit_test.go
snickell: great idea!

pkg/skaffold/tag/git_commit.go
snickell: Is this a warning condition? e.g. afaik the `Tags` variant doesn't consider it a warning when it falls back to the abbrevcommitsha, the only difference is that they found a `git` command that automatically falls back (that's the --always), whereas I couldn't find a single git shell command, so I created one by explicitly falling back

snickell:
I'm trying to figure out why the linter check is breaking here:
```
Error: pkg/skaffold/tag/git_commit_test.go:761:48: (*gitRepo).branch - result 0 (*github.com/GoogleContainerTools/skaffold/pkg/skaffold/tag.gitRepo) is never used (unparam)
make: *** [Makefile:126: linters] Error 1
func (g *gitRepo) branch(newBranchName string) *gitRepo {
```

Unfortunately this is my first go PR, I can't quite see how `branch()` is referencing `*gitRepo` (or not) any differently than the parallel functions like `commit()`..... could somebody with more go experience take a look and suggest?

snickell:
Oh I see, its that nothing is (currently) using the `*gitRepo` typed return value of `branch()`, because its always the last in our existing tests. I'm adding a no-op `.commit('second')` after one of the branches to provide a ref and please the linter.

snickell:
OK, I think this might be ready for merge pending review by @tejal29

pkg/skaffold/tag/git_commit.go
gsquared94: looks good as is 👍

snickell:
@gsquared94 Do the schemas need manual editing to add the branch tag policy, or is that something that's automatically done on the skaffold build system side?

snickell:
Aloha @gsquared94 / @briandealwis , I was wondering if my PR not editing the schemas was a mistake? Or are schemas going to update automatically at some point?

briandealwis:
@snickell the schemas and the docs will need to be updated separately.

snickell:
OK, I'll open a separate PR for that, sorry I thought those might be auto-generated by a github action or something, my bad!

Issue #6995, opened 2022-01-06
Port forwarding does not work with ko builder


norbjd:
### Expected behavior

When using [ko builder](https://skaffold.dev/docs/pipeline-stages/builders/ko/), any of the following commands:

```
skaffold debug
skaffold debug --port-forward
skaffold debug --port-forward=debug
skaffold debug --port-forward=pods
skaffold debug --port-forward=user
skaffold run --port-forward
```

should expose the debugging port (default 56268) locally, i.e. [port forwarding](https://skaffold.dev/docs/pipeline-stages/port-forwarding/) should work.

### Actual behavior

None of the previous commands port forwards the debugging port, nor deployment port (in the case of `skaffold debug --port-forward=pods`, `skaffold debug --port-forward=user`, or `skaffold run --port-forward`).

It is like skaffold **never port forwards anything with ko builder**.

### Information

- Skaffold version: v1.35.1
- Operating system: ArchLinux 5.15.12-arch1-1
- Installed via: pacman (https://archlinux.org/packages/community/x86_64/skaffold/)
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta26
kind: Config
metadata:
  name: test-skaffold
build:
  artifacts:
  - image: ko://test-skaffold
    ko: {}
deploy:
  kubectl:
    manifests:
    - deployment.yaml
```

### Steps to reproduce the behavior

1. Clone repository https://github.com/norbjd/skaffold-ko-builder-port-forwarding-does-not-work
2. `cd ko && SKAFFOLD_DEFAULT_REPO=<replace with repo> skaffold debug`

Logs does not show errors, but does not port forward debug port (I have also tried with `skaffold debug --port-forward`).

It is noticeable that debug port is correctly exposed in deployment (see `containerPort: 56268` in resulting deployment):

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  name: test-skaffold
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test-skaffold
  strategy: {}
  template:
    metadata:
      annotations:
        debug.cloud.google.com/config: '{"test-skaffold":{"artifact":"ko://test-skaffold","runtime":"go","workingDir":"/home/nonroot","ports":{"dlv":56268}}}'
      creationTimestamp: null
      labels:
        app: test-skaffold
    spec:
      containers:
      - command:
        - /dbg/go/bin/dlv
        - exec
[...]

PR #7009, opened 2022-01-12
fix(ko): Debug port forwarding for `ko://` images

halvards:
This change ensures that images with the `ko://` prefix have the `dlv` debug port, and other default ports, forwarded by default when running `skaffold debug`.

The image build process sanitizes the image name by removing the `ko://` prefix and lowercasing the Go import path. This means that the built image name doesn't match the placeholder values in the Kubernetes manifest files.

This change sanitizes the image placeholder values from the Kubernetes resource files when matching with the built image names, which in turn means that Skaffold sets up port forwarding.

Fixes: #6995
Tracking: #6041


gsquared94 (approved):
lgtm

PR #7010, opened 2022-01-12
fix(ko): Default repo for pushing `ko://` images

halvards:
When using the `ko://` scheme prefix for the Skaffold image name, the rest of the image name that follows the prefix, is a Go import path.

This import path maps to the Git repository location (typically), and won't match an image registry name (e.g., `ko://github.com/org/repo` vs `gcr.io/project_id`). Currently, the result is a hard-to-understand error resulting from a failed push to the non-existent host called `ko` (see #6933).

With this change, if _all_ of the following are true, the build results in an error that instructs the user to set the default repo:

- the image name in `skaffold.yaml` uses the `ko://` prefix; and
- the Skaffold default repo is _not_ set; and
- the image is being pushed to a registry (rather than sideloaded to a Docker daemon).

Fixes: #6933
Tracking: #6041

### DIFF
--- a/pkg/skaffold/build/ko/build.go
+++ b/pkg/skaffold/build/ko/build.go
@@ -37,6 +37,9 @@ import (
 // identifier. The image identifier is the tag or digest for pushed images, or
 // the docker image ID for sideloaded images.
 func (b *Builder) Build(ctx context.Context, out io.Writer, a *latestV1.Artifact, ref string) (string, error) {
+	if b.pushImages && strings.HasPrefix(ref, build.StrictScheme) {
+		return "", fmt.Errorf("default repo must be set when using the 'ko://' prefix and pushing to a registry: %s, see https://skaffold.dev/docs/environment/image-registries/", a.ImageName)
+	}
 	koBuilder, err := b.newKoBuilder(ctx, a)
 	if err != nil {
 		return "", fmt.Errorf("error creating ko builder: %w", err)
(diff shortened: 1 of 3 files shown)


pkg/skaffold/build/ko/build.go
```diff
@@ -37,6 +37,9 @@ import (
 // identifier. The image identifier is the tag or digest for pushed images, or
 // the docker image ID for sideloaded images.
 func (b *Builder) Build(ctx context.Context, out io.Writer, a *latestV1.Artifact, ref string) (string, error) {
+	if b.pushImages && strings.HasPrefix(ref, build.StrictScheme) {
+		return "", fmt.Errorf("default repo must be set when using the 'ko://' prefix and pushing to a registry: %s, see https://skaffold.dev/docs/environment/image-registries/", a.ImageName)
```
aaron-prindle, line 41: nit: perhaps this information should be on the docs page as well - eg: `NOTE: the ko builder requires using the default repo flag as ...`

aaron-prindle (approved):
LGTM

pkg/skaffold/build/ko/build.go
halvards: Thanks @aaron-prindle, good point. I just pushed another commit to update the docs.

MarlonGamez (approved):
lgtm. I wonder if in the future it would be better to have this validation be done at the beginning of execution failure sooner for the user

PR #7011, opened 2022-01-13
fix(labelling): hotfix for issues related to deploy erroring due to nested yaml field overriding in StatefulSet w/ PVC

aaron-prindle:
Hotfix for an issue that is being seen in which when skaffold attempts to redeploy manifests with nested objects and "immutable" fields, skaffold fails as it will attempt to override the label/image when doing so will result in a deployment error.

Minimal repro:

Skaffold.yaml
```
apiVersion: skaffold/v2beta12
kind: Config
deploy:
  kubectl:
    manifests:
      - k8-*
```
k8-pod.yaml:

```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: web
spec:
  selector:
    matchLabels:
      app: echoserver # has to match .spec.template.metadata.labels
  serviceName: "echoserver"
  replicas: 3 # by default is 1
  template:
    metadata:
      labels:
        app: echoserver 
    spec:
      terminationGracePeriodSeconds: 10
      containers:
      - image: k8s.gcr.io/echoserver:1.4
        name: echoserver
        volumeMounts:
        - name: www
          mountPath: /usr/share/nginx/html
  volumeClaimTemplates:
  - metadata:
      name: www
    spec:
      accessModes: [ "ReadWriteOnce" ]
      resources:
        requests:
          storage: 1Gi
```
run "skaffold deploy" twice w/o this PR fails with:
```
aprindle@aprindle ~/repro-6416 $ skaffold deploy
Tags used in deployment:
Starting deploy...
 - The StatefulSet "web" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy' and 'minReadySeconds' are forbidden
kubectl apply: exit status 1[
```

works w/ this change.

A more extensible/robust approach to fixing this issue is in progress as well that will implement something similar to the proposed idea from - https://github.com/GoogleContainerTools/skaffold/pull/6236 (found here - https://github.com/GoogleContainerTools/skaffold/blob/main/docs/design_proposals/configurable-transformable-allowlist.md)

### DIFF
--- a/pkg/skaffold/kubernetes/manifest/visitor.go
+++ b/pkg/skaffold/kubernetes/manifest/visitor.go
@@ -18,6 +18,7 @@ package manifest
 
 import (
 	"fmt"
+	"path"
 
 	apimachinery "k8s.io/apimachinery/pkg/runtime/schema"
 
@@ -46,7 +47,7 @@ var transformableAllowlist = map[apimachinery.GroupKind]bool{
 type FieldVisitor interface {
 	// Visit is called for each transformable key contained in the object and may apply transformations/aggregations on it.
 	// It should return true to allow recursive traversal or false when the entry was transformed.
-	Visit(object map[string]interface{}, key string, value interface{}) bool
+	Visit(path string, object map[string]interface{}, key string, value interface{}) bool
 }
 
 // Visit recursively visits all transformable object fields within the manifests and lets the visitor apply transformations/aggregations on them.
@@ -81,7 +82,7 @@ func traverseManifestFields(manifest map[string]interface{}, visitor FieldVisito
 	if shouldTransformManifest(manifest) {
 		visitor = &recursiveVisitorDecorator{visitor}
 	}
-	visitFields(manifest, visitor)
+	visitFields("/", manifest, visitor)
 }
 
 func shouldTransformManifest(manifest map[string]interface{}) bool {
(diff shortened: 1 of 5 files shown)


pkg/skaffold/kubernetes/manifest/visitor.go
```diff
@@ -123,23 +124,27 @@ type recursiveVisitorDecorator struct {
...
 }
 
 // visitFields traverses all fields and calls the visitor for each.
-func visitFields(o interface{}, visitor FieldVisitor) {
+func visitFields(jsonpath string, o interface{}, visitor FieldVisitor) {
 	switch entries := o.(type) {
 	case []interface{}:
 		for _, v := range entries {
-			visitFields(v, visitor)
+			visitFields(jsonpath, v, visitor)
```
briandealwis, line 139: You need to update `jsonpath` here too, right?

pkg/skaffold/kubernetes/manifest/visitor.go
```diff
@@ -123,23 +124,27 @@ type recursiveVisitorDecorator struct {
...
+func (d *recursiveVisitorDecorator) Visit(path string, o map[string]interface{}, k string, v interface{}) bool {
+	if d.delegate.Visit(path, o, k, v) {
+		visitFields(path, v, d)
 	}
 	return false
 }
 
 // visitFields traverses all fields and calls the visitor for each.
-func visitFields(o interface{}, visitor FieldVisitor) {
+func visitFields(jsonpath string, o interface{}, visitor FieldVisitor) {
```
briandealwis, line 135: Worth adding doc for the `jsonpath`: a `/`-delimited path representing the fields navigated to this point

briandealwis (approved):
LGTM with one caveat/question: is `jsonpath` a valid JSONPath string?  If not, could you rename `jsonpath` to `navpath` or something similar to avoid any confusion that this the contents is actually a JSONPath string?

pkg/skaffold/kubernetes/manifest/visitor.go
aaron-prindle: No, this case comes up when a list or map is encountered in the json.  The []interface{} here is always a map of maps and when we actually traverse those the other case is called.  Additionally what would you append to jsonpath in this case, `fmt.Sprintf("%v", v)`?  The other case hits every field in the json allowing us to have jsonpaths to every field.

pkg/skaffold/kubernetes/manifest/visitor.go
aaron-prindle: Done!

pkg/skaffold/kubernetes/manifest/visitor.go
briandealwis: Could you add a comment to this effect?  Even if `// this case covers lists so we don't update the path`

aaron-prindle: Good idea, done

PR #7014, opened 2022-01-13
fix: unblock cloud build edge builds

MarlonGamez:
This PR unblocks cloud build per commit jobs that create our skaffold:edge images.

It removes the `cross` make target dependency and upload stages from the relevant make targets


aaron-prindle (approved):
LGTM

Issue #6120, opened 2021-07-01
Enabling Cross-platform builds with Skaffold


gsquared94:
# Enabling Cross-platform builds with Skaffold

## Local build environment
The local build environment delegates to the individual artifact builder to decide the list of supported platforms.
- [x] Docker builder supports cross-platform
- [x] Custom builder supports cross-platform and multi-platform
- [x] Jib supports cross-platform and multi-platform
- [x] ~~Buildpacks builder only supports `linux/amd64`~~ (pending https://github.com/buildpacks/lifecycle/issues/435)
- [x] ~~Bazel supports cross-platform~~ (out of scope)
- [x] Kaniko supports cross-platform depending on the platform configuration for the nodes of the target kubernetes cluster.

## Google Cloud Build build environment

- [x] Docker builder supports cross-platform
- [x] Jib supports cross-platform and multi-platform

## Cluster build environment

- [x] The Cluster build environment supports cross-platform depending on the platform configuration for the nodes of the target kubernetes cluster.

tejal29:
This is now being de-prioritized. Please move to Dec timeline.

snickell:
Any sense what's required on this still after the merger of #7021?

gsquared94:
> Any sense what's required on this still after the merger of #7021?

updated issue description.

Issue #5504, opened 2021-03-08
Artifact hashing should incorporate the builder


briandealwis:
In #5492, our custom builder example wasn't actually being built as the artifact was resolved in the artifact cache (same files, same name).  We should also incorporate more details into the artifact hash, such as the artifact builder type, to prevent such false positives.

briandealwis:
Workaround: build with `--cache-artifacts=false`.

PR #7021, opened 2022-01-18
feat: implement cross-platform build support for local docker builder & custom builder

gsquared94:
Related: #6120 

**Description**

_Summary_

This PR implements constructs for recognizing build platforms and implements doing cross platform build for local docker builder (see `integration/examples/cross-platform-build`) and multi platform build for local custom builder (see `integration/examples/custom-buildx`).

_What's new_
* Skaffold `dev` and `debug` will check an active kubernetes context and attempt to build for the platform matching the node platform.
For instance, now you can run `skaffold dev` or `skaffold debug` from an M1 Macbook (arm64) against a minikube or GKE cluster (amd64) without changing anything in your `skaffold.yaml` file.
* The platforms can be specified in the `skaffold.yaml` file via the `build.platforms` field which sets the target build platform for all the artifacts:
```yaml
build:
  platforms: ["linux/amd64", "linux/arm64"]
  artifacts:
  - image: skaffold-example1
  - image: skaffold-example2
```
* The set of platforms can also be specified via the command line flag `--platform`, e.g. `--platform=linux/amd64,linux/arm64` which will override the value set in the config at `build.platforms`.

* Individual artifacts can set the `build.artifact[i].platforms` field that overrides all previous set values for that artifact. 
```yaml
build:
  artifacts:
  - image: skaffold-example1
    platforms: ["linux/amd64"]
  - image: skaffold-example2
    platforms: ["linux/arm64"]
```

* Skaffold artifact cache will not match two images for the same artifact if they're built for different platforms. (related #5504)

---

**Note: Multiplatform builds are not yet implemented. So, you can set at most one platform at a time**


docs/content/en/docs/references/cli/_index.md
```diff
@@ -413,6 +415,7 @@ Options:
   -n, --namespace='': Run deployments in the specified namespace
       --no-prune=false: Skip removing images and containers built by Skaffold
       --no-prune-children=false: Skip removing layers reused by Skaffold
+      --platform=[]: The platform to target for the build artifacts
```
tejal29, line 418: Should we validate platforms? e.g. `arm64, amd` etc.

gsquared94: we validate it here: https://github.com/GoogleContainerTools/skaffold/blob/0fd9e46a7928dcdc6c3a139b5612fb9331bfc889/pkg/skaffold/platform/platform.go#L88-L91

halvards:
For builders that already support multi-platform builds (e.g., ko, Jib), would artifact-specific platform config override the value of the top-level `build.platform`?

halvards:
Kubernetes clusters can be heterogeneous, with nodes of differing platforms. In this case, should we consider building for all platforms discovered?

gsquared94:
> Kubernetes clusters can be heterogeneous, with nodes of differing platforms. In this case, should we consider building for all platforms discovered?

yes, for `skaffold run` we build for all platforms detected from the cluster. For `dev` and `debug` we choose one, preferably that which matches the host machine. 

You can see read more [here](http://go/skaffold-multi-platforms)

halvards:
On a somewhat related note, looking at prior art, `ko` allows the value `all` in its `--platform` flag, to build for all platforms supported by the base image: https://github.com/google/ko#multi-platform-images

The `all` value is also allowed in the Skaffold ko builder: https://skaffold.dev/docs/references/yaml/#build-artifacts-ko-platforms

For the `build.platforms` field, should we consider supporting `all`, with similar semantics?

pkg/skaffold/build/gcb/types.go
```diff
@@ -120,3 +122,7 @@ func (b *Builder) Prune(ctx context.Context, out io.Writer) error {
 func (b *Builder) PushImages() bool {
 	return true
 }
+
+func (b *Builder) SupportedPlatforms() platform.Matcher {
+	return platform.Matcher{Platforms: []v1.Platform{{OS: "linux", Architecture: "amd64"}}}
```
tejal29, line 127: Let's confirm if this is the only supported platform.

pkg/skaffold/platform/platform.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	res := Matcher{Platforms: pl}
+	log.Entry(context.TODO()).Debugf("intersect matchers %q and %q, result %q", m, other, res)
+	return res
+}
+
+func Parse(ps []string) (Matcher, error) {
+	var sl []specs.Platform
+	for _, p := range ps {
+		platform, err := platforms.Parse(p)
```
tejal29, line 88: Should we validate this command line value by defining a custom flag type and call it in the setter.

gsquared94: I tried doing it that way, but that entailed duplicating `platforms.Parse` function otherwise I ran into cyclic module references. IMO we are validating the cli values here which happens during creating the `SkaffoldRunner` so it should be alright without any custom flag types.

gsquared94:
the failed `kokoro` test is very strange. I've run the same test against our integration test cluster and it succeeds every time. I'd request a review despite this failing test as the current set of changes shouldn't affect this testcase.
@tejal29 @briandealwis

pkg/skaffold/build/gcb/types.go
gsquared94: it isn't. But this is just to block users trying out specifying platforms for GCB builders that aren't yet implemented. Will be removed after implementing it on a builder by builder basis.

cmd/skaffold/app/cmd/flags.go
```diff
@@ -384,6 +384,14 @@ var flagRegistry = []Flag{
 		FlagAddMethod: "StringVar",
 		DefinedOn:     []string{"build", "debug", "dev", "run", "deploy"},
 	},
+	{
+		Name:          "platform",
+		Usage:         "The platform to target for the build artifacts",
```
MarlonGamez, line 389: should we point to docs here to give the user more information about accepted values?

MarlonGamez (approved):
Lgtm. Just have one question below. Also, I think we should change the title or commit message message when merging to specify which builders this adds support for

snickell:
@gsquared94 this is amazing, and very timely....for me 🤣: coincidentally, our monorepo built and passed tests for the first time on arm64 3 hours after you merged this, a backburner 1.5 year endeavor.... and here's suddenly skaffold all ready to take us the rest of the way to multi-arch builds w/o having to use a buildx shim 🙏🏽+🥅

chris13524:
Very nice!

Is there a PR or tracking issue for supporting multi-platform builds? This PR says multi-platform builds are not yet supported.

gsquared94:
> Very nice!
> 
> Is there a PR or tracking issue for supporting multi-platform builds? This PR says multi-platform builds are not yet supported.

It's tracked in https://github.com/GoogleContainerTools/skaffold/issues/6120

snickell:
It looks like this didn't make it all the way into a new `gcr.io/k8s-skaffold/skaffold:edge` yet, maybe due to the failing windows test? 
image

I've been testing this locally, and its working great for me on both linux/arm64 and linux/amd64 to build for the other platform! Very cool, thanks @gsquared94 again

gsquared94:
> It looks like this didn't make it all the way into a new `gcr.io/k8s-skaffold/skaffold:edge` yet, maybe due to the failing windows test? 

We've fixed the broken windows UT. This should be available on the image `gcr.io/k8s-skaffold/skaffold:edge` now @snickell

gsquared94:
@chris13524 skaffold multi-platform build support is almost complete. https://github.com/GoogleContainerTools/skaffold/issues/7082

You can use the bleeding-edge version right now, should be out with the next release.

PR #7043, opened 2022-01-22
[Do Not Commit] run Windows tests with -v to show testing output

briandealwis:


briandealwis:
And now it's working 😡

Issue #7051, opened 2022-01-24
Skaffold last-prompted updated despite update-check is false


Guillaume-Mayer:
Hi,

I turned update-check off, in Skaffold global settings, because I use my package manager to update.
I try to track my dot files with git and I added ~/.skaffold/config to them.

### Expected behavior

When update-check is false, update.last-prompted should not be updated (I think)

### Actual behavior

update.last-prompted is updated, I think each time I use Skaffold

### Information

- Skaffold version: v1.35.2
- Operating system: Linux 5.16
- Installed via: pacman
- Contents of ~/.skaffold/config:

```yaml
global:
  local-cluster: true
  update-check: false
  survey:
    last-prompted: "2022-01-19T17:56:58-03:00"
  collect-metrics: false
  update:
    last-prompted: "2022-01-24T20:29:00-03:00"
kubeContexts: []
```

### Steps to reproduce the behavior

1. `skaffold config set -g update-check false`
2. Backup ~/.skaffold/config
3. Run a skaffold run  o skaffold build
4. Check ~/.skaffodl/config diffs with backup

Guillaume-Mayer:
This is really a minor issue obviously.
I could try to fix it myself if you confirm me that it is not the expected behavior.
Be fine

tejal29:
Thanks @Guillaume-Mayer for finding this issue. You are right. The `last-prompted` should not be updated. Feel free to create a PR.

Thanks!

tomas-mota:
I'll try to fix it. Remember to label 'good first issue' for other new comers :)

Guillaume-Mayer:
Thanks @tomas-mota I'm not familiar enough with Go, I tried to add an if but then test failed and I didn't investigate more

PR #7067, opened 2022-01-31
fix: dont update last-prompted if check-update is false

tomas-mota:
Fixes: #7051

**Description**
When check-update is set to false, the config property last-prompted was being updated even though the user is not prompted. This has been fixed.

**Notes**
Also removed a test regarding the survey prompt that was being ran as part of the update message unit tests (coverage not affected)

### DIFF
--- a/pkg/skaffold/config/util.go
+++ b/pkg/skaffold/config/util.go
@@ -387,6 +387,10 @@ func IsUpdateCheckEnabled(configfile string) bool {
 	if err != nil {
 		return true
 	}
+	return IsUpdateCheckEnabledForContext(cfg)
+}
+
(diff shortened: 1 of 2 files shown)


briandealwis:
I think this looks reasonable — wdyt @tejal29?

tejal29:
> I think this looks reasonable — wdyt @tejal29?

Agree. We did discuss on the issue opened for the same.

tejal29 (approved):

tomas-mota:
The windows unit test falure looks like a flake, should we just re-run it @tejal29?

tomas-mota:
Just ran 'go test -short -timeout 60s  ./pkg/skaffold/... ./cmd/... ./hack/... ./pkg/webhook/...' on a windows machine and all tests passed fine :/

PR #7069, opened 2022-02-02
fix: always reset cached repo if sync is true

tomas-mota:
Fixes: #6598 

**Description**
As described in the issue and mentioned in the docs, the indented behaviour of the git sync flag seems to be that, if true, the repo gets reset on every run. This PR removes the check for changes in the cached repo, always hard resetting when sync is true.

This is just a possible solution, which avoids breaking the configuration by changing the type of the flag from bool to string.

If we assume that no users should have sync as true _and_ make manual changes to the cached repos, this has no user facing changes


MarlonGamez:
hey @tomas-mota , thanks for taking the time to work on this! It's much appreciated. These changes look pretty good to me, I'll run our CI to make sure we're not breaking anything

tomas-mota:
Not sure why the integration test failed, it does not seem to be related to the feature at hand. Could you give me a hand in debugging this?

MarlonGamez:
I can try and figure out what's going on here, I don't think it should be related to this PR so I might need to see if another PR that got merged broke something

tomas-mota:
Should we try to rerun in order to understand whether or not it was a fluke?

MarlonGamez:
Hey @tomas-mota, thanks for being patient with this. I just merged a change yesterday to fix that integration test failure, so if you rebase your PR then that issue should be fixed :)

tejal29 (approved):
thanks @MarlonGamez and @tomas-mota for your patience. The kokoro job timedout to Kaniko test. I am going to merge this in

PR #7089, opened 2022-02-09
fix: correctly set the default value of `StatusCheck` to nil

sbe-genomics:
**Description**

The `statusCheck` option in the `deploy` config section was not working as expected. The value from the skaffold config was never picked up, however it was doing the right thing when `--status-check` was used on the command line.

[This](https://github.com/GoogleContainerTools/skaffold/blob/20ae7f64ccdb9fcf40fb18e41f3cd9a86eb312ed/pkg/skaffold/runner/deployer.go#L53) method was checking if the `StatusCheck` value was defined via the command line, and if so returning the value it was set to. However since the flag was set to incorrectly default to true, that method would always return true if `--status-check=` was not set on the command line.

This PR sets the default value of `StatusCheck` to `nil`, as this lines up with its usage and [type](https://github.com/GoogleContainerTools/skaffold/blob/20ae7f64ccdb9fcf40fb18e41f3cd9a86eb312ed/pkg/skaffold/config/options.go#L95) `BoolOrUndefined`

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -316,11 +316,10 @@ var flagRegistry = []Flag{
 		Name:          "status-check",
 		Usage:         "Wait for deployed resources to stabilize",
 		Value:         &opts.StatusCheck,
-		DefValue:      true,
+		DefValue:      nil,
 		FlagAddMethod: "Var",
 		DefinedOn:     []string{"dev", "debug", "deploy", "run", "apply"},
 		IsEnum:        true,
-		NoOptDefVal:   "true",
 	},
 	{
 		Name:          "iterative-status-check",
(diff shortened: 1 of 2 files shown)


MarlonGamez:
hey @sbe-genomics , thanks for catching this and creating a fix! Would you mind rebasing this PR? the integration test failure is unrelated and we've merged a fix into main

sbe-genomics:
@MarlonGamez done, thanks!

MarlonGamez:
Merging this change as the integration test failure is unrelated/needs to be fixed on our end

cmd/skaffold/app/cmd/flags.go
```diff
@@ -316,11 +316,10 @@ var flagRegistry = []Flag{
 		Name:          "status-check",
 		Usage:         "Wait for deployed resources to stabilize",
 		Value:         &opts.StatusCheck,
-		DefValue:      true,
+		DefValue:      nil,
 		FlagAddMethod: "Var",
 		DefinedOn:     []string{"dev", "debug", "deploy", "run", "apply"},
 		IsEnum:        true,
-		NoOptDefVal:   "true",
```
gsquared94, line 323: This line needs to be reverted

Issue #7084, opened 2022-02-08
Disabling profiles does not work


vladimir-avinkin:
### Expected behavior
According to documentation (in `skaffold --help`) prepending a minus sign before a profile name should disable it, instead, we get an error.
### Actual behavior
When trying to disable a profile we get an error that says that a profile with a prepended minus is not declared.

### Information

- Skaffold version: v1.35.2
- Operating system: Darwin MacBook-Pro-Vladimir.local 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:07:06 PST 2021; root:xnu-7195.81.3~1/RELEASE_X86_64 x86_64
- Installed via: homebrew
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta13
kind: Config
profiles:
  - name: default
  - name: test
```

Below are my speculations:
The initial feature was introduced here https://github.com/GoogleContainerTools/skaffold/pull/4054 and then probably the logic got duplicated somewhere along the way, probably two places that conflict with each other are here:
https://github.com/GoogleContainerTools/skaffold/blob/20ae7f64ccdb9fcf40fb18e41f3cd9a86eb312ed/pkg/skaffold/parser/config.go#L96
https://github.com/GoogleContainerTools/skaffold/blob/20ae7f64ccdb9fcf40fb18e41f3cd9a86eb312ed/pkg/skaffold/schema/profiles.go#L130

### Steps to reproduce the behavior

1. bare skaffold.yaml contents of which are stated above is enough to reproduce
2. `skaffold dev -p -test`
3. According to documentation prepending a minus sign before a profile name should disable it, instead we got an error
```
skaffold dev -p -test
parsing skaffold config: profile selection ["-test"] did not match those defined in any configurations. Check that values specified in the "--profile" or "-p" flags are valid profile names.
```

tomas-mota:
I'll take a look at this :)

tomas-mota:
@vladimir-avinkin the error message should definitely be more clear, but it seems like you are trying to disable the 'test' profile, even though no profile was activated. As you can read [here](https://skaffold.dev/docs/environment/profiles/), no profile is enabled by default, you have to activate it either through the command or setting triggers in the yaml.

Perhaps what would make sense is that if you exclude a profile that was not activated anyways, it simply continues without erroring (perhaps this was what you actually meant and I misunderstood)

PR #7099, opened 2022-02-10
fix: check for profile existing

tomas-mota:
Fixes: #7084

**Description**
The check for whether or not a profile exists was being made against the applied profiles. These already take into consideration profiles remove using the '-' notation in the command, so the removed profile would not be found in the list.

### DIFF
--- a/pkg/skaffold/parser/config.go
+++ b/pkg/skaffold/parser/config.go
@@ -92,10 +92,6 @@ func GetConfigSet(ctx context.Context, opts config.SkaffoldOptions) (SkaffoldCon
 		return nil, sErrors.ZeroConfigsParsedErr(opts.ConfigurationFile)
 	}
 
-	if unmatched := unmatchedProfiles(r.appliedProfiles, opts.Profiles); len(unmatched) != 0 {
-		return nil, sErrors.ConfigProfilesNotMatchedErr(unmatched)
-	}
-
 	for _, c := range cfgs {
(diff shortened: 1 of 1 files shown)


MarlonGamez (approved):
LGTM ✅ Tested locally to make sure this works. Would probably be good to introduce an integration test for this at some point in the future

@tomas-mota if you're willing to do add a test that would be awesome, but if not that's understandable. We can always save that for a follow up PR

gsquared94 (changes requested):
This change is checking for the valid profiles at every config parsed. However the profile names are scoped within the space of all configs (explicit and imported). For instance, this branch fails for `integration/examples/multi-config-microservices` with the following `skaffold.yaml`:
```yaml
apiVersion: skaffold/v2beta27
kind: Config
requires:
- path: ./leeroy-app
- path: ./leeroy-web
profiles:
  - name: foo
```

I think we can make only the following change to `unmatchedProfiles` method and solve this issue:
```diff
func unmatchedProfiles(activatedProfiles map[string]string, allProfiles []string) []string {
	var allActivated []string
	for _, profiles := range activatedProfiles {
		allActivated = append(allActivated, strings.Split(profiles, ",")...)
	}
	var unmatched []string
	for _, p := range allProfiles {
+		if strings.HasPrefix(p, "-") {
+			continue
+		}
		if !stringslice.Contains(allActivated, p) {
			unmatched = append(unmatched, p)
		}
	}
	return unmatched
}

```

gsquared94:
Closing in favor of https://github.com/GoogleContainerTools/skaffold/pull/7107

PR #7107, opened 2022-02-14
fix: correctly handle excluded profiles

tomas-mota:
Fixes: #7084
**Alternative implemetation to PR #7099**

**Description**
As mentioned in #7099, that implementation is checking for every profile loaded, which does not work. The problem is that since each file is processed in a "bubble", we never have a full picture of all the profiles in order to check the command line flags. I added such a list to the record, such that we can check all the opts profiles against the config files after all files have been loaded.

What do you think of this solution @gsquared94?

This is sort of WIP, I can try to add an integration test in a follow up PR, as requested by @MarlonGamez


gsquared94 (approved):
lgtm!

PR #7109, opened 2022-02-15
[V2] refactor: remove pointers from render code. 

tejal29:
Removing pointers from render package. 
Generators, Validators and Transformers don't need to be pointer and can be empty structs.

All the above objects struct fields are arrays, string or boolean. Converting these to object will avoid any NPE


MarlonGamez (approved):
LGTM ✅

Issue #7111, opened 2022-02-15
Skaffold 1.36 fails if config file named `skaffold.yml`, but still works for `skaffold.yaml`


AndiDog:
**Important:** I already have a fix for this ready soon. Just publishing the issue already for visibility. This seems to come from changes where `buildMapOfSchemaObjPointerToYAMLInfos` and eventually `config.ReadConfiguration` is called with an absolute path, not a filename (note that the term "filename" means the name, not the path of a file).

### Expected behavior

Skaffold works with existing config file with `.yml` file extension.

### Actual behavior

```
$ rm -rf /tmp/project \
  && cd /tmp \
  && mkdir project \
  && cd project \
  && printf 'apiVersion: skaffold/v2beta8\nkind: Config\n' >skaffold.yml \
  && skaffold version \
  && skaffold run
v1.36.0
parsing skaffold config: error parsing skaffold configuration file: open /private/tmp/project/skaffold.yaml: no such file or directory
```

### Information

- Skaffold version: v1.36.0
- Operating system: macOS 12.2.1
- Installed via: Homebrew
- Contents of skaffold.yaml: any valid content

### Steps to reproduce the behavior

See above

PR #7112, opened 2022-02-15
fix: parsing alternative config filename `skaffold.yml` by supporting absolute paths in `config.ReadConfiguration`

AndiDog:
Fixes: #7111

**Description**
It seems some recent change introduced calling the now-fixed function with absolute paths. For example, the call stack via `buildMapOfSchemaObjPointerToYAMLInfos` does this. The logical solution for me was to fix the config parsing function to accept a path, not a filename.

I didn't know where best to add a test for this, and there hasn't been one before. See #7111 for how I manually tested that it works.

### DIFF
--- a/pkg/skaffold/util/config.go
+++ b/pkg/skaffold/util/config.go
@@ -35,11 +35,11 @@ var stdin []byte
 
 // ReadConfiguration reads a `skaffold.yaml` configuration and
 // returns its content.
-func ReadConfiguration(filename string) ([]byte, error) {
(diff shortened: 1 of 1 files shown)


MarlonGamez (commented):
hi @AndiDog , thanks for the contribution! These changes look good. Would you mind just changing the commit message to follow the pattern of `fix: ...`? We've recently standardized our commit messages according to `conventionalcommits.org`

gsquared94 (approved):
lgtm

Issue #7024, opened 2022-01-19
Skaffold does not cope well with buildkit used with --output (invalid reference format)


ifalex:
### Expected behavior
When using buildkit in conjunction with --output skaffold should not expect that docker will tag the build.

### Actual behavior
Docker buildkit used with --output does not produce any tag for the build. Skaffold on the other hand is not aware about this behavior and expects a tag failing the process.
```
Error parsing reference: "" is not a valid repository/tag: invalid reference format
```
### Information

- Skaffold version: 
```
v1.35.2
```
- Operating system: 
Running from a docker container based on:
```CentOS Linux release 7.9.2009 (Core)  (Linux 6e75c71d7fd8 4.18.0-348.7.1.el8_5.x86_64```
- Installed via: 
```skaffold.dev```
- Contents of skaffold.yaml:

```yaml
---
apiVersion: skaffold/v2beta26
kind: Config
build:
  local:
    push: false
    tryImportMissing: true
    useBuildkit: true
    useDockerCLI: true
  tagPolicy:
    envTemplate:
      template: '{{.GIT_COMMIT}}'
  artifacts:
  - image: my-docker-image
    context: source/my-docker-image
    docker:
      buildArgs:
        GIT_COMMIT: 75c25fd
        GIT_BRANCH: <branch_name>
      noCache: false
      target: export-test-results
      cliFlags:
      - --output
      - /tmp/output_dir
```

### Steps to reproduce the behavior

1. Using a multistage docker build with a target and running skaffold on that target to get the output files (unit tests at build time)
https://docs.docker.com/engine/reference/commandline/build/#custom-build-outputs

```
FROM scratch as export-test-results
COPY --from=build /my-app/build/reports/jacoco/* /
```
From skaffold perspective this will resume to following skaffold build execution:

2. `skaffold build -f /tmp/skaffold-my-image-export-test-results.yaml --quiet=false --cache-artifacts=false --default-repo=my-docker.repository`
Which will fail with:
```
Error parsing reference: "" is not a valid repository/tag: invalid reference format
```
By using vdebug I got the command that was executed de-facto by skaffold.
```
[...]

PR #7120, opened 2022-02-16
fix:  skaffold's assumption for image tag when building via buildkit and custom output

elnoro:
Fixes: #7024 

**Description**
The original issue is about using local docker buildkit build. When building with buildkit and custom output, image is not tagged by default. Empty image id is passed to `ImageAPIClient.ImageTag` from the docker package, which fails.

This [line](https://github.com/GoogleContainerTools/skaffold/blob/55bea853ef2e81e4fc5b15322d6a04997176d1d0/pkg/skaffold/docker/image.go#L555) suggests that an empty image id is a valid use case in some circumstances. 

I simply added a new condition to the method - if image id is empty, docker `ImageTag` is not called and simply returns empty string instead. 
Unit test included. 

**Before**
When using buildkit and custom output, skaffold returned the following error
`Error parsing reference: "" is not a valid repository/tag: invalid reference format`
**After**
When using buildkit and custom output, skaffold successfully completes build. 

Custom output is produced in both cases, as it's exported by the docker cli, not by skaffold

### DIFF
--- a/pkg/skaffold/docker/image.go
+++ b/pkg/skaffold/docker/image.go
@@ -540,6 +540,10 @@ func (l *localDaemon) TagWithImageID(ctx context.Context, ref string, imageID st
 		return "", err
 	}
 
+	if imageID == "" {
+		log.Entry(ctx).Debugf("generating tag for %s: empty image id\n", ref)
+		return "", nil
+	}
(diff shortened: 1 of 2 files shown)


tejal29:
re-running the test suite.

Issue #7125, opened 2022-02-17
The --default-repo flag no longer seems to be respected in 1.36.0


todd-andrew-procore:
### Expected behavior

On 1.35.0, 1.35.2, and previous versions; `skaffold run --default-repo localhost:5999` is able to push docker images to the local Docker registry running at `localhost:5999`

### Actual behavior

On 1.36.0, the `--default-repo` flag and the `SKAFFOLD_DEFAULT_REPO` environment variable seem to be ignored in favor of the global config.

`skaffold run --default-repo localhost:5999`

results in:

```
build [docker-registry/my-app] failed: could not push image "localhost:5000/docker-registry/my-app:0.158.000-4-ge8bafc2-dirty": Get "http://localhost:5000/v2/": dial tcp [::1]:5000: connect: connection refused
error Command failed with exit code 1.
```

### Information

- Skaffold version: 1.36.0
- Operating system: MacOs Big Sur 11.6.3
- Installed via: asdf
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta27
kind: Config
build:
  artifacts:
  - image: docker-repo/my-app
    docker:
      buildArgs:
        APP_ENV: ${APP_ENV}
      cliFlags:
      - --progress
      - plain
  insecureRegistries:
  - localhost:{{.REGISTRY_PORT:-5999}}
  local:
    useBuildkit: true
deploy:
  helm:
    releases:
    - name: my-app
      chartPath: .chart/
      artifactOverrides:
        image: docker-repo/my-app
      namespace: my-app
      createNamespace: true
      imageStrategy:
        helm: {}
```

### Steps to reproduce the behavior

1. Local KinD cluster
2. Local Docker registry running at localhost:5999
3. Run `skaffold fix --overwrite` to update our config
4. `skaffold run --default-repo localhost:5999`

I don't think logs are needed as the error should be easily reproducible. If logs are needed, I can provide snippets.

tomas-mota:
I'll look into this 🙂

MarlonGamez:
hi @tomas-mota, would you like this assigned to you? We'd like to have this fixed for our next release so please let me know if you want any help or want someone else to take this up :)

tomas-mota:
Hi, I have in fact been trying to track it down, let me know if you have any hints :D Think I'll figure it out before March, so you can assign it to me.

MarlonGamez:
I think the piece of logic here basically shouldn't happen
[...]

PR #6985, opened 2021-12-31
feat: fetch default-repo from local cluster

shaxbee:
Fixes: #6558 

**Description**
If no default-repo is specified via CLI or config and local cluster is being used try to look up config map describing local registry following [KEP-1755](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/generic/1755-communicating-a-local-registry).

Sample script for setting up local registry for Kind: https://kind.sigs.k8s.io/docs/user/local-registry/

Demo:
* Clone [github.com/shaxbee/todo-app-skaffold](https://github.com/shaxbee/todo-app-skaffold)
* `make bootstrap`
* `skaffold run --force`

You will observe that local registry is automatically used as default-repo and images are pushed instead of being loaded without need to use `skaffold config`.

Issue #7121, opened 2022-02-16
[Test Failure] TestBuildKanikoInsecureRegistry is timing our consistently 


tejal29:
The latest version of kaniko is broken due to auth issues. 

1) Pinning our test to older version to see if it fixes the Test. 
   - It did not fix the test. 

`TestBuildKanikoInsecureRegistry` needs more investigation. Skipping it for now.

ericzzzzzzz:
re-open this due to the test is being flaky

PR #7144, opened 2022-02-27
fix: choose cli default-repo over config file

tomas-mota:
Fixes:  #7125

**Description**
Fixes error introducing in pr #6985. Logic was introduced there to give precedence to config value instead of cli.

### DIFF
--- a/pkg/skaffold/config/util.go
+++ b/pkg/skaffold/config/util.go
@@ -257,10 +257,6 @@ func GetCluster(ctx context.Context, opts GetClusterOpts) (Cluster, error) {
 
 	var defaultRepo = opts.DefaultRepo
 
-	if defaultRepo.Value() != nil && cfg.DefaultRepo != "" {
-		defaultRepo = NewStringOrUndefined(&cfg.DefaultRepo)
-	}
-
 	if local && defaultRepo.Value() == nil {
 		registry, err := DiscoverLocalRegistry(ctx, kubeContext)
 		switch {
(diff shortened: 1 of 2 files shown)


pkg/skaffold/config/util.go
```diff
@@ -257,10 +257,6 @@ func GetCluster(ctx context.Context, opts GetClusterOpts) (Cluster, error) {
 
 	var defaultRepo = opts.DefaultRepo
 
-	if defaultRepo.Value() != nil && cfg.DefaultRepo != "" {
-		defaultRepo = NewStringOrUndefined(&cfg.DefaultRepo)
-	}
-
 	if local && defaultRepo.Value() == nil {
```
gsquared94, line 260: is `cfg.DefaultRepo` being checked somewhere else? Otherwise I think the author intended this to say:
```
if defaultRepo.Value() == nil && cfg.DefaultRepo != "" {
		defaultRepo = NewStringOrUndefined(&cfg.DefaultRepo)
}
```

MarlonGamez: the reason I believe this piece of code is not needed is because typically when we call the following function:
https://github.com/GoogleContainerTools/skaffold/blob/1dbc7a2c8746201ccff73a54928a57d687eb020a/pkg/skaffold/deploy/util/util.go#L34-L52

We grab the value from the config and compare things then. This is how things usually worked before these few lines were added I believe

gsquared94: sounds good

tomas-mota:
Could the PR actually have broken something in that test? I'm not familiar with kaniko, so I'd have a hard time debugging this

tejal29:
There is an open issue #7121. Not your fault!!

gsquared94 (approved):
lgtm

Issue #7181, opened 2022-03-10
build concurrency from skaffold.yaml is not respected


ssmall:
### Expected behavior

Build concurrency is set to 5

### Actual behavior

Build concurrency is set to 1

### Information

- Skaffold version: v1.36.1
- Operating system: macOS 12.2.1 (darwin/arm64)
- Installed via: Homebrew
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta18
kind: Config
build:
  local:
    concurrency: 5
requires:
  - path: ./src/frontend
  - path: ./src/backend
```

(this is just a modified version of https://github.com/GoogleCloudPlatform/cloud-code-samples/blob/v1/golang/go-guestbook/skaffold.yaml)

### Steps to reproduce the behavior

1.  `git clone https://github.com/GoogleCloudPlatform/cloud-code-samples`
1. `cd golang/go-guestbook`
3. `skaffold dev`

Relevant debug logs snippet:
```
DEBU[0001] setting Docker user agent to skaffold-v1.36.1  subtask=-1 task=DevLoop
INFO[0001] Using minikube docker daemon at tcp://127.0.0.1:55676  subtask=-1 task=DevLoop
DEBU[0001] CLI platforms provided: ""                    subtask=-1 task=DevLoop
DEBU[0001] getting client config for kubeContext: `minikube`  subtask=-1 task=DevLoop
DEBU[0001] platforms detected from active kubernetes cluster nodes: "linux/arm64"  subtask=-1 task=DevLoop
DEBU[0001] platforms selected for artifact "java-guestbook-frontend": "linux/arm64"  subtask=-1 task=DevLoop
DEBU[0001] platforms selected for artifact "java-guestbook-backend": "linux/arm64"  subtask=-1 task=DevLoop
DEBU[0001] Using builder: local                          subtask=-1 task=DevLoop
DEBU[0001] push value not present in NewBuilder, defaulting to false because cluster.PushImages is false  subtask=-1 task=DevLoop
INFO[0001] build concurrency first set to 1 parsed from *local.Builder[0]  subtask=-1 task=DevLoop
DEBU[0001] Using builder: local                          subtask=-1 task=DevLoop
DEBU[0001] push value not present in NewBuilder, defaulting to false because cluster.PushImages is false  subtask=-1 task=DevLoop
[...]

PR #7182, opened 2022-03-11
fix: ignore concurrency if not specified

gsquared94:
Fixes: #7181 

**Description**

The build concurrency should be set to the minimum defined concurrency across all configs. Previously, imported configs that didn't define concurrency were defaulting to the value of 1, thereby lowering the overall concurrency to 1.
We remove setting this default value in the config, rather if the resolved build concurrency is defaulted to 1 at the end.

**Testing instructions**
- In project `examples/multi-config-microservices` set the root `skaffold.yaml` config's build concurrency.
```yaml
apiVersion: skaffold/v2beta28
kind: Config
build:
  local:
    concurrency: 2
requires:
- path: ./leeroy-app
- path: ./leeroy-web
```
- Run `skaffold build -v DEBUG`. The output should have the following log lines showing that the build concurrency was updated to that value:
```
INFO[0000] build concurrency first set to 2 parsed from *local.Builder[3]  subtask=-1 task=DevLoop
INFO[0000] final build concurrency value is 2            subtask=-1 task=DevLoop
```

### DIFF
--- a/pkg/skaffold/build/builder_mux.go
+++ b/pkg/skaffold/build/builder_mux.go
@@ -149,10 +149,10 @@ func getConcurrency(pbs []PipelineBuilder, cliConcurrency int) int {
 	}
 	minConcurrency := -1
 	for i, b := range pbs {
-		if b.Concurrency() == nil {
-			continue
+		concurrency := 1
+		if b.Concurrency() != nil {
+			concurrency = *b.Concurrency()
 		}
-		concurrency := *b.Concurrency()
 		// set mux concurrency to be the minimum of all builders' concurrency. (concurrency = 0 means unlimited)
 		switch {
 		case minConcurrency < 0:
--- a/pkg/skaffold/build/builder_mux.go
+++ b/pkg/skaffold/build/builder_mux.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"reflect"
 
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/constants"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/graph"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/hooks"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/output/log"
@@ -65,20 +66,29 @@ func NewBuilderMux(cfg Config, store ArtifactStore, builder func(p latestV1.Pipe
 
 		if cfg.BuildConcurrency() >= 0 {
 			minConcurrency = cfg.BuildConcurrency()
-		} else {
-			concurrency := b.Concurrency()
-			// set mux concurrency to be the minimum of all builders' concurrency. (concurrency = 0 means unlimited)
-			switch {
-			case minConcurrency < 0:
-				minConcurrency = concurrency
-				log.Entry(context.TODO()).Infof("build concurrency first set to %d parsed from %s[%d]", minConcurrency, reflect.TypeOf(b).String(), i)
-			case concurrency > 0 && (minConcurrency == 0 || concurrency < minConcurrency):
-				minConcurrency = concurrency
-				log.Entry(context.TODO()).Infof("build concurrency updated to %d parsed from %s[%d]", minConcurrency, reflect.TypeOf(b).String(), i)
-			default:
-				log.Entry(context.TODO()).Infof("build concurrency value %d parsed from %s[%d] is ignored since it's not less than previously set value %d", concurrency, reflect.TypeOf(b).String(), i, minConcurrency)
-			}
+			continue
 		}
+
+		if b.Concurrency() == nil {
+			continue
+		}
+		concurrency := *b.Concurrency()
+
+		// set mux concurrency to be the minimum of all builders' concurrency. (concurrency = 0 means unlimited)
+		switch {
+		case minConcurrency < 0:
+			minConcurrency = concurrency
+			log.Entry(context.TODO()).Infof("build concurrency first set to %d parsed from %s[%d]", minConcurrency, reflect.TypeOf(b).String(), i)
+		case concurrency > 0 && (minConcurrency == 0 || concurrency < minConcurrency):
+			minConcurrency = concurrency
+			log.Entry(context.TODO()).Infof("build concurrency updated to %d parsed from %s[%d]", minConcurrency, reflect.TypeOf(b).String(), i)
+		default:
+			log.Entry(context.TODO()).Infof("build concurrency value %d parsed from %s[%d] is ignored since it's not less than previously set value %d", concurrency, reflect.TypeOf(b).String(), i, minConcurrency)
+		}
+	}
+	if minConcurrency < 0 {
+		minConcurrency = constants.DefaultLocalConcurrency // set default concurrency to 1.
+		log.Entry(context.TODO()).Infof("build concurrency set to default value of %d", minConcurrency)
 	}
 	log.Entry(context.TODO()).Infof("final build concurrency value is %d", minConcurrency)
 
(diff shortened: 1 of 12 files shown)


pkg/skaffold/build/builder_mux.go
```diff
@@ -65,20 +66,29 @@ func NewBuilderMux(cfg Config, store ArtifactStore, builder func(p latestV1.Pipe
...
-			}
+			continue
 		}
+
+		if b.Concurrency() == nil {
+			continue
+		}
+		concurrency := *b.Concurrency()
+
+		// set mux concurrency to be the minimum of all builders' concurrency. (concurrency = 0 means unlimited)
```
briandealwis, line 77: Add that `0` merits special treatment: if there are two configs A and B, and A has concurrency 0 and B has concurrency 1, then the "minimum" should be 1 since it's the most stringent.

pkg/skaffold/build/builder_mux.go
```diff
@@ -65,20 +66,29 @@ func NewBuilderMux(cfg Config, store ArtifactStore, builder func(p latestV1.Pipe
...
-				log.Entry(context.TODO()).Infof("build concurrency updated to %d parsed from %s[%d]", minConcurrency, reflect.TypeOf(b).String(), i)
-			default:
-				log.Entry(context.TODO()).Infof("build concurrency value %d parsed from %s[%d] is ignored since it's not less than previously set value %d", concurrency, reflect.TypeOf(b).String(), i, minConcurrency)
-			}
+			continue
 		}
+
+		if b.Concurrency() == nil {
+			continue
+		}
```
briandealwis, line 74: This doesn't seem to be equivalent.  Consider if I have two modules, A and B.  Both have local build-envs.  A does not specify a build concurrency.  B specified concurrency 0.  I believe that prior to this change, A would have had a default concurrency of 1 set which would have resulted in the minConcurrency==1 (serialized).  Whereas now A would be skippedI think we will have minConcurrency 0.

Honestly, I think we should separate this out into a function whose results can be tested for different scenarios.

briandealwis (commented):
WDYT about extracting the minConcurrency logic into a separately testable function?

pkg/skaffold/build/builder_mux_test.go
```diff
@@ -92,6 +92,71 @@ func TestNewBuilderMux(t *testing.T) {
...
+		description         string
+		pbs                 []PipelineBuilder
+		cliConcurrency      int
+		expectedConcurrency int
+	}{
+		{
+			description: "default concurrency - builder and cli concurrency unset.",
+			pbs: []PipelineBuilder{
+				&mockPipelineBuilder{nil, "local"},
+				&mockPipelineBuilder{nil, "gcb"},
```
briandealwis, line 106: I'm thinking it might be worth explicitly specifying field names — it wasn't clear to me that "local" and "gcb" were the builder-type (go/go-style/decisions#literal-field-names).

pkg/skaffold/build/builder_mux.go
tejal29: @briandealwis I verified via this test [here](https://github.com/GoogleContainerTools/skaffold/pull/7182/commits/f8904bd6575a22bca86fa7bb4c078fe13d9ea3b2#diff-1dd11874c1929244aebdbb3fe85c5120f832b2c8b17f53be4276af55a0dce90aR122) that if B specified concurrency 0 and A does not set the concurrency, then it default to 1.

pkg/skaffold/build/builder_mux_test.go
```diff
@@ -92,6 +92,83 @@ func TestNewBuilderMux(t *testing.T) {
...
+		{
+			description: "min non-zero concurrency",
+			pbs: []PipelineBuilder{
+				&mockPipelineBuilder{util.IntPtr(0), "local"},
+				&mockPipelineBuilder{util.IntPtr(3), "gcb"},
+				&mockPipelineBuilder{util.IntPtr(2), "gcb"},
+			},
+			cliConcurrency:      -1,
+			expectedConcurrency: 2,
+		},
```
briandealwis, line 162: I think you need a test for all-GCB.  I suspec the code will return 1 but it should return 0 for this case.

pkg/skaffold/build/builder_mux_test.go
```diff
@@ -92,6 +92,83 @@ func TestNewBuilderMux(t *testing.T) {
...
+			description: "default concurrency - builder and cli concurrency unset.",
+			pbs: []PipelineBuilder{
+				&mockPipelineBuilder{nil, "local"},
+				&mockPipelineBuilder{nil, "gcb"},
+			},
+			cliConcurrency:      -1,
+			expectedConcurrency: 1,
+		},
+		{
+			description: "builder concurrency set to less cli concurrency",
```
MarlonGamez, line 112:
```suggestion
			description: "builder concurrency set to less than cli concurrency",
```

pkg/skaffold/build/builder_mux.go
```diff
@@ -169,3 +141,34 @@ func filterBuildEnvSupportedPlatforms(supported platform.Matcher, target platfor
...
+		case concurrency > 0 && (minConcurrency == 0 || concurrency < minConcurrency):
+			minConcurrency = concurrency
+			log.Entry(context.TODO()).Infof("build concurrency updated to %d parsed from %s[%d]", minConcurrency, reflect.TypeOf(b).String(), i)
+		default:
+			log.Entry(context.TODO()).Infof("build concurrency value %d parsed from %s[%d] is ignored since it's not less than previously set value %d", concurrency, reflect.TypeOf(b).String(), i, minConcurrency)
+		}
+	}
+	if minConcurrency < 0 {
+		log.Entry(context.TODO()).Infof("build concurrency set to default value of %d", minConcurrency)
+		return constants.DefaultLocalConcurrency // set default concurrency to 1.
```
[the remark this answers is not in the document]
tejal29: yeah. I had to read the config.go to understand, local build concurrency is a ptr while GCB and cluster is a int with default zero.

pkg/skaffold/build/builder_mux_test.go
tejal29: it's the default in IntelliJ. Will add the names.
image

MarlonGamez (commented):
don't have much to add that brian didn't already catch

pkg/skaffold/build/builder_mux.go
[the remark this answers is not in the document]
tejal29: done

pkg/skaffold/build/builder_mux_test.go
tejal29: Let me add one.

tejal29: So, in case of GCB, it not possible to have nil Concurrency. 
See GCB build definition.  The `Concurrency` is an actual int.

```
type GoogleCloudBuild struct {
	// ProjectID is the ID of your Cloud Platform Project.
	// If it is not provided, Skaffold will guess it from the image name.
	// For example, given the artifact image name `gcr.io/myproject/image`, Skaffold
	// will use the `myproject` GCP project.
	ProjectID string `yaml:"projectId,omitempty"`
....

	// Concurrency is how many artifacts can be built concurrently. 0 means "no-limit".
	// Defaults to `0`.
	Concurrency int `yaml:"concurrency,omitempty"`

}


```

pkg/skaffold/build/builder_mux_test.go
tejal29: done!

pkg/skaffold/build/builder_mux_test.go
MarlonGamez: @tejal29 should we still add a test case for it ?

tejal29: done.

Issue #7131, opened 2022-02-18
Implement `ko` builder beta features


halvards:
Tracking issue for the `ko` builder beta features.

See the [release plan](docs/design_proposals/ko-builder.md#release-plan) in the design proposal for further details.

Related: #6041 (`ko` builder alpha tracking issue)

PR #7193, opened 2022-03-15
refactor: Use new logrus.Logger instead of default

halvards:
Use a new `logrus.Logger` instance instead of the default logrus singleton.

Relying on the default singleton is problematic when dependencies (or transitive dependencies) also use logrus. When code paths in these dependencies modify the default singleton, it also impacts Skaffold logging.

I ran into this issue while trying to upgrade the `ko` dependency to v0.10, and there was unwanted error-level log output messages from `amazon-ecr-credential-helper`.

Context:
- https://github.com/google/ko/pull/586
- https://github.com/awslabs/amazon-ecr-credential-helper/pull/309

This change also moves us closer to not leaking the logger implementation dependency outside the `output/log` package.

Tracking: #7131

### DIFF
--- a/pkg/skaffold/output/log/log.go
+++ b/pkg/skaffold/output/log/log.go
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Includes code from github.com/sirupsen/logrus (MIT License)
+
 package log
 
 import (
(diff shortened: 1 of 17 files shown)


pkg/skaffold/output/log/log.go
```diff
@@ -28,100 +30,170 @@ import (
...
 }
 
+// New returns a new logrus.Logger.
+// We use a new instance instead of the default logrus singleton to avoid clashes with dependencies that also use logrus.
 func New() *logrus.Logger {
 	return logrus.New()
 }
 
 // KanikoLogLevel makes sure kaniko logs at least at Info level and at most Debug level (trace doesn't work with Kaniko)
 func KanikoLogLevel() logrus.Level {
```
halvards, line 142: This function doesn't appear to be called anywhere, except for the unit test. I left the function signature unchanged, in case I missed something.

briandealwis:
Adding @MarlonGamez as he's our logging expert

pkg/skaffold/output/log/log.go
MarlonGamez: Yeah, we've actually been using the global logrus logger, but we should definitely completely move off of that. I can make a follow up to this to have us use a local one that we own

MarlonGamez (approved):
thanks for making this change @halvards . As mentioned I can make a follow up to this to make further improvements

PR #7205, opened 2022-03-18
fix: add missing Argo CronWorkflow to transforms

kevin-hanselman:
**Description**
Other Argo Workflows CRDs with images are already allowed in the source. This PR adds the missing CronWorkflow CRD.

**User facing changes**
Skaffold will now appropriately transform the images mentioned in CronWorkflow resources.


MarlonGamez:
hi there @kevin-hanselman! Thanks for opening this. Before moving forward, would you be willing to try using our new config that allows you to specify kubernetes resources to be modifiable by skaffold? Our docs for this can be found [here](https://skaffold.dev/docs/tutorials/skaffold-resource-selector/).

MarlonGamez:
@kevin-hanselman I don't think we have any concern about it, just that we'd like to find a comfortable solution for people in the future to avoid them having to go through the friction of creating a PR for skaffold to work with their resources. You're right that we have the other argo resources here so I think it does make sense to add this one for this PR. I'll approve and merge this one.

Can I ask what you think would improve the awkwardness/surprise of the solution we have now?

kevin-hanselman:
> Can I ask what you think would improve the awkwardness/surprise of the solution we have now?

@MarlonGamez Merging this PR 😄 Really though: The surprise comes from expecting Skaffold to treat CronWorkflow resources the same as other Argo resources. The awkwardness comes from the workaround -- using a custom resource selector for CronWorkflow resources specifically. Another dev looking at my Skaffold config might ask: "Why aren't you adding other Argo resources to the resourceSelector section?" (That is, if I hadn't linked to this PR in the file. 😄)

I think the resource selector is a huge win, by the way. Please don't mistake this PR for criticism of that concept. It's the inconsistency of handling Argo resources that's the issue for me. If for some reason this PR wouldn't get merged, the next best solution, IMO, would be to remove all Argo resources from the built-in transforms and lean on the resource selector wholesale.

MarlonGamez:
Ahh okay, makes sense, thanks for clarifying! Appreciate the response :)

tejal29 (commented):
dummy FB

MarlonGamez (approved):
lgtm

PR #7213, opened 2022-03-21
fix: change GCB backoff check to use error code instead of checking the error

barp:
**Related**: #5888

**Description**
Current gcb implementation checks quota limit using string matching, this can break as recently the string returned from the server has changed causing the backoff mechanism to not work.

This PR change the check to be based on error code instead.

According to https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto

the relevant error code is 429 (as the HTTP error code is returned).

I have tested this to be the correct error code using the following program:

```go
	cbclient, err := cloudbuild.NewService(ctx)
	if err != nil {
		fmt.Printf("Service creation error")
		return
	}
	var wg sync.WaitGroup
	for i := 1; i < 10000; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			_, err := cbclient.Projects.Builds.Get(projectID, remoteID).Do()
			if err != nil {
				fmt.Printf("got gcb error %T\n", err)
				if apiErr, ok := err.(*googleapi.Error); ok {
					fmt.Printf("Detected error %d\n", apiErr.Code)
				}
			}
		}()
	}

	wg.Wait()
```

### DIFF
--- a/pkg/skaffold/build/gcb/cloud_build.go
+++ b/pkg/skaffold/build/gcb/cloud_build.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strings"
 	"time"
 
 	cstorage "cloud.google.com/go/storage"
@@ -195,7 +194,8 @@ watch:
 			if err == nil {
 				return true, nil
 			}
-			if strings.Contains(err.Error(), "Error 429: Quota exceeded for quota metric 'cloudbuild.googleapis.com/get_requests'") {
+			// Error code 429 is the error code for quota exceeded https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+			if apiErr, ok := err.(*googleapi.Error); ok && apiErr.Code == 429 {
 				// if we hit the rate limit, continue to retry
 				return false, nil
 			}


pkg/skaffold/build/gcb/cloud_build.go
```diff
@@ -195,8 +194,7 @@ watch:
 			if err == nil {
 				return true, nil
 			}
-			if strings.Contains(err.Error(), "Error 429: Quota exceeded for quota metric 'cloudbuild.googleapis.com/get_requests'") {
-				// if we hit the rate limit, continue to retry
```
MarlonGamez, line 199: I'd say we should leave this comment here or leave a new comment explaining this case since it's not really clear what this is to someone reading the code

MarlonGamez (changes requested):
Thanks for opening this PR @barp! This change looks good, I just have one comment

pkg/skaffold/build/gcb/cloud_build.go
barp: Hi, I have returned the comments and added another one on 429.

barp:
Hi, fixed the comment. Can you look again?

MarlonGamez (approved):
lgtm, thanks @barp

barp:
Can we rerun the test, the failure doesn't seem related to this change.

MarlonGamez:
Looks like the test is stuck in queue but can't be deleted or forced to run 🤔 could you try pushing an empty commit or something similar to retrigger the tests?

barp:
forced pushed, now its waiting for maintainer to approve.

barp:
Hi, do you know when the next release with this fix will be out?

Issue #7228, opened 2022-03-25
Incompatible output directory with io_bazel_rules_docker v0.23


sashamor:
### Expected behavior

`skaffold.yaml`

```yaml
build:
  artifacts:
  - image: reports
    bazel:
      target: //go/services/reports/cmd/server:server_img.tar
```

Expect that it builds & runs with `skaffold`.

### Actual behavior

`skaffold dev` gives:

```
Target //go/services/reports/cmd/server:server_img.tar up-to-date:
  bazel-out/k8-fastbuild-ST-4a519fd6d3e4/bin/go/services/reports/cmd/server/server_img.tar
INFO: Elapsed time: 84.321s, Critical Path: 54.36s
INFO: 572 processes: 210 remote cache hit, 30 internal, 332 linux-sandbox.
INFO: Build completed successfully, 572 total actions
INFO: Build completed successfully, 572 total actions
Building [reports]...
Build [reports] was canceled
build [reports] failed: reading image "/home/sasha/.cache/bazel/_bazel_sasha/ab67a788adc4b5b8d1470f6531fe3b59/execroot/__main__/bazel-out/k8-fastbuild/bin/go/services/reports/cmd/server/server_img.tar": open /home/sasha/.cache/bazel/_bazel_sasha/ab67a788adc4b5b8d1470f6531fe3b59/execroot/__main__/bazel-out/k8-fastbuild/bin/go/services/reports/cmd/server/server_img.tar: no such file or directory
```

It looks like it's expdecting `server_img` to be at `./bazel-out/k8-fastbuild/...`, but it's now at `./bazel-out/k8-fastbuild-ST-4a519fd6d3e4/...` since the latest `io_bazel_rules_docker` update.

### Information

- Skaffold version: v1.37.0
- Operating system: Linux workspace-678db7c888-t7jbq 5.4.0-1067-azure #70~18.04.1-Ubuntu SMP Thu Jan 13 19:46:01 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
- Installed via: Homebrew
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta19
kind: Config
metadata:
  name: paypaplane-dev-env
build:
  artifacts:
  - image: reports
    bazel:
      target: //go/services/reports/cmd/server:server_img.tar
```

### Steps to reproduce the behavior

1. Create skaffold project
2. `skaffold dev`
3. Above error.

tejal29:
[...]

PR #7251, opened 2022-03-31
fix: fix bazel-out path for `rules_docker-0.23.0`

sashamor:
Fixes: #7228 
**Related**: https://github.com/bazelbuild/rules_docker/pull/1963

**Description**

With the release of https://github.com/bazelbuild/rules_docker/releases/tag/v0.23.0, the output directory for bazel targets is calculated differently and is no longer a static path.

For example, a `container_image` target such as:
```
//foo/bar/server:server_img.tar
```

Previously was stored at:
```
bazel-out/k8-fastbuild/bin/foo/bar/server/server_img.tar
```

Internally, `skaffold` calculated this path by running `bazel info bazel-bin`, which returns the start of the path (e.g. `/home/...../execroot/__main__/bazel-out/k8-fastbuild/bin`), then appended the name of the rule with `:` replaced with `/` (e.g. `foo/bar/server/server_img.tar`). This results in the correct path as above.

However, with `rules_docker-0.23.0`, the output is now stored at a path like:
```
bazel-out/k8-fastbuild-ST-4a519fd6d3e4/bin/foo/bar/server/server_img.tar
```

So `bazel info` no longer returns the correct path prefix.

Using `bazel cquery` with a similar method to that described in https://github.com/bazelbuild/rules_pkg/tree/main/examples/where_is_my_output, as recommended in https://github.com/bazelbuild/rules_docker/pull/1963#issuecomment-1081197607, we can query Bazel for the actual output path to the target. Docker `.tar` output files are available in `cquery` as targets with a `.files` attribute, which contains the list of output files (there's only one), and each file has a `.path` attribute that gives the path to the file, which we can use internally in Skaffold. This is also backwards compatible with older versions of https://github.com/bazelbuild/rules_docker.

For example, the command:
```
bazel cquery //foo/bar/server:server_img.tar --output starlark --starlark:expr="target.files.to_list()[0].path"
```
May return:
```
bazel-out/k8-fastbuild/bin/foo/bar/server/server_img.tar
``` 
And with `rules_docker-0.23.0`, this may return something like:
```
bazel-out/k8-fastbuild-ST-4a519fd6d3e4/bin/foo/bar/server/server_img.tar
```
Which is the correct path to the target.


MarlonGamez:
hi @sashamor , Thanks for opening this change! It looks like you'll have to go in and update `TestBuildBazel` unit test as well, but other than that these changes look good :)

sashamor:
Thanks, fixed :) Also changed the arguments to no longer use nested quotes.

Tested locally with https://github.com/bazelbuild/rules_docker v0.23.0 and v0.22.0 and can confirm they both find the output tar correctly as expected.

Anthony-Bible:
I wanted to add another data point, I had a similar problem and it worked when running with your patch.

MarlonGamez:
@sashamor , thanks for making the change! I have a fix open for the github action failure we're seeing 
https://github.com/GoogleContainerTools/skaffold/pull/7263
Once this is merged, could you rebase this PR to retrigger the tests? I think it should be fixed then

MarlonGamez (approved):

sashamor:
Thanks for rebasing and merging :) Woohoo

pkg/skaffold/build/bazel/build.go
```diff
@@ -80,12 +79,11 @@ func (b *Builder) buildTar(ctx context.Context, out io.Writer, workspace string,
 		return "", fmt.Errorf("running command: %w", err)
 	}
 
-	bazelBin, err := bazelBin(ctx, workspace, a)
+	tarPath, err := bazelTarPath(ctx, workspace, a)
```
morozov, line 82: It looks like this breaks the scenarios where the artifact has a non-default context. The `bazel cquery` runs in the directory corresponding to the context and returns that tar path relative to that context but `docker.Push()` still runs in the current working directory and is now supplied with a path relative to the context.

Take the following example. There is a scaffold file in projects/acme/some-projects:
```yaml
apiVersion: skaffold/v2beta26
kind: Config
build:
  artifacts:
  - image: quay.io/acme/some-project
    context: ../../..
```
If one does:
```
cd projects/acme/some-projects && skaffold build
```
It will no longer work. It will fail with an error like:
```
build [quay.io/sugarcrm/blackhole] failed: reading image "bazel-out/darwin-fastbuild-ST-4a519fd6d3e4/bin/projects/some-project/docker.tar": open bazel-out/darwin-fastbuild-ST-4a519fd6d3e4/bin/projects/some-project//docker.tar: no such file or directory
```

pkg/skaffold/build/bazel/build.go
```diff
@@ -80,12 +79,11 @@ func (b *Builder) buildTar(ctx context.Context, out io.Writer, workspace string,
...
 	if err != nil {
-		return "", fmt.Errorf("getting path of bazel-bin: %w", err)
+		return "", fmt.Errorf("getting bazel tar path: %w", err)
 	}
 
-	tarPath := filepath.Join(bazelBin, buildTarPath(a.BuildTarget))
```
morozov, line 88: It looks like this breaks the scenarios where the artifact has a non-default context. The `bazel cquery` runs in the directory corresponding to the context and returns the tar path relative to that context but `docker.Push()` still runs in the current working directory and is now supplied with a path relative to the context.

Take the following example. There is a skaffold file in projects/acme/some-projects:
```yaml
apiVersion: skaffold/v2beta26
kind: Config
build:
  artifacts:
  - image: quay.io/acme/some-project
    context: ../../..
```
If one does:
```
cd projects/acme/some-projects && skaffold build
```
It will no longer work. It will fail with an error like:
```
build [quay.io/sugarcrm/blackhole] failed: reading image "bazel-out/darwin-fastbuild-ST-4a519fd6d3e4/bin/projects/some-project/docker.tar": open bazel-out/darwin-fastbuild-ST-4a519fd6d3e4/bin/projects/some-project//docker.tar: no such file or directory
```

morozov, line 88: It looks like this breaks the scenarios where the artifact has a non-default context and the `skaffold` command is executed from within a path different than the repository root.

The `bazel cquery` runs in the directory corresponding to the context and returns the tar path relative to that context but `docker.Push()` still runs in the current working directory and is now supplied with a path relative to the context, not relative to the current working directory.

Take the following example. There is a skaffold file in projects/acme/some-projects:
```yaml
apiVersion: skaffold/v2beta26
kind: Config
build:
  artifacts:
  - image: quay.io/acme/some-project
    context: ../../..
```
If one does:
```
cd projects/acme/some-projects && skaffold build
```
It will no longer work. It will fail with an error like:
```
build [quay.io/acme/some-project] failed: reading image "bazel-out/darwin-fastbuild-ST-4a519fd6d3e4/bin/projects/some-project/docker.tar": open bazel-out/darwin-fastbuild-ST-4a519fd6d3e4/bin/projects/some-project//docker.tar: no such file or directory
```

PR #7278, opened 2022-04-07
fix: add default value for status-check flag when no value is specified

MarlonGamez:
Fixes problem addressed here: https://github.com/GoogleContainerTools/skaffold/pull/7089/files#r845391034

`--status-check` can now be used without specifying `={true|false}`

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -320,6 +320,7 @@ var flagRegistry = []Flag{
 		FlagAddMethod: "Var",
 		DefinedOn:     []string{"dev", "debug", "deploy", "run", "apply"},
 		IsEnum:        true,
+		NoOptDefVal:   "true",
 	},
 	{
 		Name:          "iterative-status-check",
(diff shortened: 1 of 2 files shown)


briandealwis (commented):
Can we add a test for this?

MarlonGamez:
@briandealwis Yes, do you think we should add a test for all NoOptDefValue flags?

tejal29 (approved):

briandealwis:
This flag broke downstream consumers, so it seems worthy of some tests to ensure it doesn't happen again.

MarlonGamez:
@briandealwis I included a test in this PR to ensure that skaffold runs with these flags passed in with no value 👍🏼

byCedric:
Can we please leave this now as is, thanks for adding tests to make sure this doesn't break again. But now our pipeline broke twice:
1. For the previous change that added `--status-check {true,false}` - PR #7089
2. For this change that broke `--status-check true` - https://github.com/expo/snack/runs/7411459432?check_suite_focus=true#step:9:11

PR #7282, opened 2022-04-10
fix: do not replace every string matching image names

barata:
Skaffold does not behave consistently when transforming manifests. It will always replace fields inside Config Connector manifests that match some image name in the build. See the example below.

```yaml
image: image-name
something: image-name
image: do-no-match
```

Skaffold will produce the following output:

```yaml
image: <REPLACED>
something: <REPLACED>
image: do-no-match
```

However, the expected result would be:

```yaml
image: <REPLACED>
something: image-name
image: do-no-match
```

This bug only exists for Config Connector resources due to [this line](https://github.com/GoogleContainerTools/skaffold/compare/main...barata:main#diff-8e1e726f52eed31819c51b1de64e9c24bc22cbbc63b022edd299d9970b73ad60L67).

The fix will make the behavior consistent with other kinds of resources and only replace `image` fields, unless some rule inside `TransformAllowlist` tell otherwise. Many tests were added to illustrate the expected behavior.

Another minor bug in [this block](https://github.com/GoogleContainerTools/skaffold/pull/7282/files#diff-8e1e726f52eed31819c51b1de64e9c24bc22cbbc63b022edd299d9970b73ad60L84-L87) was fixed. The problem happens when there is a wildcard `.*` and some other field inside the same `image` definition. The previous code will just stop the loop when it finds a wildcard, which is not the desired state. See the example below.

```yaml
image: image-name
something: image-name
```

In case `Image: [".*", ".something"]` exists, Skaffold will not replace the field `something` because a wildcard is found first inside the loop. Tests were also added to cover this case.

### DIFF
--- a/pkg/skaffold/kubernetes/manifest/images.go
+++ b/pkg/skaffold/kubernetes/manifest/images.go
@@ -29,6 +29,8 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest"
 )
 
+const imageField = "image"
+
 type ResourceSelectorImages struct {
 	allowlist map[apimachinery.GroupKind]latest.ResourceFilter
 	denylist  map[apimachinery.GroupKind]latest.ResourceFilter
@@ -62,9 +64,12 @@ func (rsi *ResourceSelectorImages) allowByGroupKind(gk apimachinery.GroupKind) b
 }
 
 func (rsi *ResourceSelectorImages) allowByNavpath(gk apimachinery.GroupKind, navpath string, k string) (string, bool) {
+	matchedConfigConnectorImage := false
+
 	for _, w := range ConfigConnectorResourceSelector {
-		if w.Matches(gk.Group, gk.Kind) {
-			return "", true
+		if k == imageField && w.Matches(gk.Group, gk.Kind) {
+			matchedConfigConnectorImage = true
+			break
 		}
 	}
 
@@ -80,11 +85,10 @@ func (rsi *ResourceSelectorImages) allowByNavpath(gk apimachinery.GroupKind, nav
 	}
 
 	if rf, ok := rsi.allowlist[gk]; ok {
+		matchedConfigConnectorImage = false
+
 		for _, allowpath := range rf.Image {
-			if allowpath == ".*" {
-				if k != "image" {
-					return "", false
-				}
+			if allowpath == ".*" && k == imageField {
 				return "", true
 			}
 			if navpath == allowpath {
@@ -92,7 +96,7 @@ func (rsi *ResourceSelectorImages) allowByNavpath(gk apimachinery.GroupKind, nav
 			}
 		}
 	}
-	return "", false
+	return "", matchedConfigConnectorImage
 }
 
 // GetImages gathers a map of base image names to the image with its tag
@@ -107,7 +111,7 @@ type imageSaver struct {
 }
 
 func (is *imageSaver) Visit(gk apimachinery.GroupKind, navpath string, o map[string]interface{}, k string, v interface{}, rs ResourceSelector) bool {
-	if k != "image" {
+	if k != imageField {
 		return true
 	}
 
--- a/pkg/skaffold/kubernetes/manifest/images_test.go
+++ b/pkg/skaffold/kubernetes/manifest/images_test.go
@@ -20,8 +20,12 @@ import (
 	"context"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


MarlonGamez:
@barata thanks for opening up this PR to fix this. I think these changes look good, there's just a complaint from our linter. Would you be able to fix that? It just wants you to take a repeated string and put it into a variable for reuse

barata:
Cool, thanks for the review @MarlonGamez 
I just fixed the issue. Please let me know if you want me to do anything else.

pkg/skaffold/kubernetes/manifest/images.go
```diff
@@ -62,9 +64,12 @@ func (rsi *ResourceSelectorImages) allowByGroupKind(gk apimachinery.GroupKind) b
...
 
 func (rsi *ResourceSelectorImages) allowByNavpath(gk apimachinery.GroupKind, navpath string, k string) (string, bool) {
+	matchedConfigConnectorImage := false
+
 	for _, w := range ConfigConnectorResourceSelector {
-		if w.Matches(gk.Group, gk.Kind) {
-			return "", true
+		if k == imageField && w.Matches(gk.Group, gk.Kind) {
+			matchedConfigConnectorImage = true
+			break
```
MarlonGamez, line 72: Clarifying question: is the reason we don't return here anymore to let the denylist stop it if the user configures that?

MarlonGamez (commented):
just one quick question!

aaron-prindle:
@barata thank you for the PR!  This code was recently changed to allow allowlistying/denylisting for certain groupkinds.  Unfortunately Config Connector is a special case currently as it allows all kinds for a specific group.  I think this default behaviour is better than the current status for the reasons you have outlined.  The only caveat is that there are some config connector resources where what is essentially `image:` has a different naming schema for example a `BuildStep` which uses `name:` (instead of the image:) field.  This was initially why this was done but as you have mentioned the matching of all of the image tags has negative unintended consequences.  LGTM

I have added an issue - https://github.com/GoogleContainerTools/skaffold/issues/7305 to track adding some kind of whitelist to specific Config Connector resources that perhaps use a different field than 'image:' for images which skaffold might need to overwrite in those cases

aaron-prindle (approved):
LGTM

pkg/skaffold/kubernetes/manifest/images.go
barata: Yes, but not only that. You cannot return here before also evaluating the allowlist below. The default `.*` will replace `image` fields, but if the user defines a selector with some field name, `image` won't be replaced (unless the user definition includes the `image` field as well). This is the behavior for other non-Config Connector resources as well. I wrote some tests to verify this behavior.

barata:
> which uses `name:` (instead of the image:) field. This was initially why this was done but as you have mentioned the matching of all of the image tags has negative unintended consequences for users.

@aaron-prindle Yes, I realized that when I was browsing the code before submitting this PR. If you think it makes sense, we can also explicitly define a rule in the allowlist (inside vitisor.go) to cover cases where `image` is not the field to be replaced for Config Connector resources. It will work as expected together with this PR. But users can also define that rule themselves after this PR is merged.

Thanks for the review 👍

MarlonGamez:
sorry @barata , could you rebase this PR when you get the chance?

barata:
@MarlonGamez done :+1:

barata:
I rebased the PR again @tejal29

tejal29:
Thank you @barata Will get this in today.

tejal29:
Thanks @barata There are some lint errors in image_test.go

1) Imports not in order
2) lint. 

I wanted to update the PR myself but don't have permissions.  Here is the patch. you can simply run gofmt tool to obtain it. 
[patch](https://gist.github.com/tejal29/b7133eb4723965e3202b8b09482287a0)

barata:
@tejal29 done, thank you!

tejal29:
> @tejal29 done, thank you! Btw, I applied your patch because "go fmt" will indent the code as it was before.

oh shoot. Maybe I made a mistake then. Lets see if the linter is happy

tejal29:
Sorry looks like you have to run `gofmt`

barata:
np! done @tejal29

barata:
hi @tejal29 @MarlonGamez ! Is this going to be available in a future v1 release or only on v2?
If it is only for v2, to which branch should I submit a PR to hotfix this?

PR #7242, opened 2022-03-29
chore: merge v2-v1.37.0 into main

MarlonGamez:
**Description**
This PR is only here to see the potential changes of merging the v2 branch into main

PR #7290, opened 2022-04-12
chore: remove `v1` config package and move it to `latest` package.

gsquared94:
Consistently set alias `latest` for importing `"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/latest/v1"`.
This will reduce diff for PR #7242

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -70,7 +70,7 @@ to be submitted.
 
 Some changes to the skaffold code require a change to the skaffold config. These changes require a few extra steps:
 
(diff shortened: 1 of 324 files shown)


MarlonGamez:
@gsquared94, do we want to move the package up one level as well? `latest/v1` -> `latest`
If we do the same on the v2 branch then I think we can reduce the diff even more

gsquared94:
> @gsquared94, do we want to move the package up one level as well? `latest/v1` -> `latest` If we do the same on the v2 branch then I think we can reduce the diff even more

done.

MarlonGamez (approved):
@gsquared94 This looks good to me, on the v2 branch when we do this same change I think we can keep move up `latest/v2` to `latest`, and keep the `latest/v1` package where it is so we can reference when we need to.

Do you want me to go ahead with that change, or would you like to work on it?

PR #7291, opened 2022-04-12
Set permissions for GitHub actions

neilnaveen:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

 Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/verify-examples.yml
+++ b/.github/workflows/verify-examples.yml
@@ -7,6 +7,9 @@ concurrency:
   group: build-${{ github.event.pull_request.number || github.ref }}-${{github.workflow}}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   dependabot:
     name: Verify changes to examples


MarlonGamez:
@neilnaveen this definitely seems like a good change, should we do this on all of our jobs? Or is this only applicable to the one job that you modified?

MarlonGamez:
@neilnaveen this definitely seems like a good change, should we do this on all of our jobs? Or is this only applicable to the one job that you modified?

neilnaveen:
> @neilnaveen this definitely seems like a good change, should we do this on all of our jobs? Or is this only applicable to the one job that you modified?

Probably yes,  for now I was able to address this one. Thanks

.github/workflows/verify-examples.yml
```diff
@@ -7,6 +7,9 @@ concurrency:
   group: build-${{ github.event.pull_request.number || github.ref }}-${{github.workflow}}
   cancel-in-progress: true
 
+permissions:
+  contents: read
```
briandealwis, line 11: The following will make all scopes `read`, vs just contents.
```suggestion
permissions: read-all
```

briandealwis:
Thanks @neilnaveen for picking this up.

@MarlonGamez The other jobs have `permissions: read-all`.

briandealwis:
@neilnaveen you'll need to sign the CLA

briandealwis (approved):

PR #7301, opened 2022-04-14
fix: Do not ignore supplied platform when retagging an image

barp:
When building image using the --platform flag, if platform is not equal to the platform of the built image and skaffold decides it only needs to retag the image, the operation will fails because the platform do not match and will fail with 
```
getting image: no child with platform {Architecture:amd64 OS:linux OSVersion: OSFeatures:[] Variant: Features:[]} in index <image tag>
```

this PR tries to fix this issue.

### DIFF
--- a/pkg/skaffold/build/bazel/build.go
+++ b/pkg/skaffold/build/bazel/build.go
@@ -48,7 +48,7 @@ func (b *Builder) Build(ctx context.Context, out io.Writer, artifact *latest.Art
 	}
 
 	if b.pushImages {
-		return docker.Push(tarPath, tag, b.cfg)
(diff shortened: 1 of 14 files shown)


barp:
I will fix the tests once I get CR too see if the changes are acceptable.

gsquared94:
Thank you @barp for this PR. The changes definitely make sense. 

We are currently in the process of upgrading the `main` branch to `v2` schema version so the branch is a bit unstable atm. Is it possible for you to raise this PR against the branch `v1` instead of `main`? 

I can take up cherrypicking these changes onto `main` when we get there.

barp:
Hi, thanks for doing this. I was busy in the last few weeks.

Issue #6683, opened 2021-10-05
Add helm support for `skaffold init`


MarlonGamez:
Support getting helm projects up and running with `skaffold init`. This change is currently blocked by changes being made to `render` and the upcoming config v2 changes.

tejal29:
Assigning this to myself. 

This featured will be completed in the following 

| Description | On V1   | On V2 | 
| ------------- | -------|------ |
| Add manifest transformation in skaffold filter |  [ ] Add `--transform` flag to `skaffold filter` command |  [X] Copy the test over on v2 branch.  |
| add skaffold init for helm |  [ ]   |  [ ]  |
| add new helm31Deployer|  [ ] | [ ]  |

tejal29:
I just verified, Helm subcharts init is implemented. Tests are here. https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/initializer/helm_test.go#L81

PR #7317, opened 2022-04-15
chore: Add test for filter command

tejal29:
This is one of the tasks for #6683

Adding a test for filter command. This will copied over to v1 branch.


aaron-prindle (approved):
LGTM

Issue #7302, opened 2022-04-14
Suppot adding `--context-sub-path` option to kaniko executor can handle more directory structure


sshota0809:
This is a proposal.

To be able to specify `--context-sub-path` option used by kaniko executor in the `skaffold.yaml` can handle more directory structure.

### Expected behavior
I can specify `contextsubpath` option in `skaffold.yaml`.

```
apiVersion: skaffold/v2beta28
kind: Config
metadata:
  name: my-api
build:
  artifacts:
    - image: my-registory/my-api
      context: rootdir/
      kaniko:
        dockerfile: ../environments/my-api/app/docker/Dockerfile
        contextsubpath: my-api/
```

### Actual behavior
I have a repository that has a structure like

```
environments/my-api/app/
├── docker
│   └── Dockerfile
└── skaffold.yaml
my-api/
└── src/ (source code for my-api)
```

I made `skaffold.yaml` like following and run `skaffold build` in `environments/my-api/app/`.

```
apiVersion: skaffold/v2beta28
kind: Config
metadata:
  name: my-api
build:
  artifacts:
    - image: my-registory/my-api
      context: ../../../my-api
      kaniko:
        dockerfile: ../environments/my-api/app/docker/Dockerfile
...
```

But this fails with a kaniko executor error because skaffold only copies directories under `context` whereas `dockerfile` contains `../`.

```
Error: error resolving dockerfile path: please provide a valid path to a Dockerfile within the build context with --dockerfile
...
```

Ideally I would like to run a kaniko executor with `--context-sub-path` option through `skaffold.yaml` like

```
apiVersion: skaffold/v2beta28
kind: Config
metadata:
  name: my-api
build:
  artifacts:
    - image: my-registory/my-api
      context: ../../../
      kaniko:
        dockerfile: ../environments/my-api/app/docker/Dockerfile
        contextsubpath: my-api/
...
```

What do you think about this proposal. Thank you for your attention in advance.

tejal29:
Makes sense to us. @sshota0809 would you want to work on this. 
It will be very similar to this PR https://github.com/GoogleContainerTools/skaffold/pull/6703

1) Add a config to the Kaniko struct
2) Add the flag and value to the kaniko executor command if the config is defined. 

Let us know if we can help

sshota0809:
@tejal29 
Thank you for your response! I will try to create a PR.
If I need any help I would ask you it.

sshota0809:
@tejal29 
[...]

PR #7334, opened 2022-04-22
feat: support the contextSubPath param in the Kaniko Artifact

sshota0809:
Fixes: #7302 

**Description**
Support the `contextSubPath` param in the Kaniko Artifact.

**Implementation Strategy**

Before I implement this feature, I was going to add `--context-sub-path` param in Kaniko executor.
But when creating `tar` file, the file path of the dependencies turns relative path from context.

https://github.com/GoogleContainerTools/skaffold/blob/8ff6917f61f08795b1ed6a84010190521d63e0cd/pkg/skaffold/util/tar.go#L139

By the above effect, the dependencies is decompressed in Kaniko root context as `dir:///kaniko/buildcontext`.

For example,

* dir structure
```
environments/my-api/app/
├── docker
│   └── Dockerfile
└── skaffold.yaml
my-api/
└── src/ (source code for my-api)
```

* skaffold.yaml
```
apiVersion: skaffold/v3alpha1
kind: Config
metadata:
  name: my-api
build:
  artifacts:
    - image: my-registory/my-api
      context: ../../../
      kaniko:
        dockerfile: ../environments/my-api/app/docker/Dockerfile
        contextSubPath: my-api/
...
```

* dir structure in Kaniko executor pod after decompressing `tar` of dependencies.
```
# ls /kaniko/buildcontext
src/
environments/
```

In case of this example if I add `--context-sub-path my-api/` to Kaniko executor it fails to find context because `my-api/` dir doesn't exist in `/kaniko/buildcontext`.

So I implemented this feature by combining `contextSubPath` and `dockerfile` path like following.

* pkg/skaffold/build/kaniko/args.go
```
"--dockerfile", filepath.Join(artifact.ContextSubPath, artifact.DockerfilePath),
```

Please review and thanks a lot in advance.

### DIFF
--- a/pkg/skaffold/build/cache/hash.go
+++ b/pkg/skaffold/build/cache/hash.go
@@ -28,6 +28,7 @@ import (
 	"sort"
 
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/buildpacks"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/kaniko"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/config"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/docker"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/graph"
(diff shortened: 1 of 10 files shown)


tejal29 (approved):
LGTM

tejal29:
@sshota0809 there are some windows test failures. you might have to use `/` instead of filepath.Join to make skaffold more portable across os architectures.

sshota0809:
@tejal29 [tejal29](https://github.com/tejal29)

Thank you for your information. I've fixed it. When file path is passed to kaniko executor, it's correct that the separator is `/` so I used `filepath.ToSlash()`.

PR #7335, opened 2022-04-22
chore: format gotest output to easily navigate the test results

tejal29:
Came across this interesting project to format our test output and save some time. 
https://debugged.it/projects/gotestfmt/

Github Project:
https://github.com/haveyoudebuggedit/gotestfmt#github-actions

Try this on one github workfow job


janosdebugs:
Hey @tejal29 I just came across this PR, sadly gotestfmt doesn't, and can't support go test output without `-json` because the packages can't be correctly identified from the plain `go test` output.

Issue #7235, opened 2022-03-28
Investigate replacing statik with go:embed


briandealwis:
Skaffold uses github.com/rakyll/statik to embed a number of files in our binary including:
  - third-party licenses (`skaffold credits`)
  - skaffold.yaml schemas (`skaffold schema list` and `skaffold schema get`)

The minikube team saw [a 15MB reduction in their binary](https://github.com/kubernetes/minikube/pull/11671#issuecomment-862544939) from switching from `statik` to `go:embed`.

Details on `go:embed`: https://pkg.go.dev/embed

PR #7339, opened 2022-04-26
refactor: replace statik with embed

ericzzzzzzz:
Fixes: #7235 

**Description**
 - remove statik mod and related code
 - rename statik-generate.sh to embedded-files-generate.sh and move to generated files to fs/*_generated folders so we can embed those files into output binary. 
 - update MakeFile accordingly. 
 - read files from the embedded fs in schema/print, credit/export, instrumentations/export method.
 - create fs diretory to hold embedded files(licenses, schemas) generated from script. 
 - refactor credits/export method to use io/fs.WalkDir to traversal credit director.
 - update tests. 
 
**Binary size result**
  - before (86,850,656 bytes (86.9 MB on disk))
  - after (91,637,424 bytes (91.6 MB on disk)) 
  Not sure why it's increased 

**Test Plan**
This should not break the exiting behavior.. related testing
  - run ``make`` should collect license and schemas by default
  - run ``out/skaffold credits`` should be export licenses successfully run ``out/skaffold schema get skaffold/v1`` should export skaffold/v1 schema
  - update a source file then re-run make should not trigger license and schemas collecting
  - run ``make clean`` then and ``make `` should trigger license and schemas collecting
  - run ``Local=true make`` should skip license and schemas collecting

**Follow up Issues**
 - #7343 investigate options for replacing golint


ericzzzzzzz:
The linter seems doesn't recognize //go:embed special comment, the comment fails to pass lint check due to not adding space after //

tejal29:
Feel free to use any of the methods mentioned here https://golangci-lint.run/usage/false-positives/ to disable linting for that line or file.

hack/generate-embedded-files.sh
```diff
@@ -34,30 +31,21 @@ elif ! [ -x "$(command -v ${LICENSES})" ]; then
     # Also can't be easily installed from a vendor folder because it relies on non-go files
     # from a dependency.
     echo "Installing go-licenses"
-    pushd $(mktemp -d ${TMPDIR:-/tmp}/generate-statik.XXXXXX)
-    #TODO(marlongamez): unpin this version once we're able to build using go 1.16.x
     go mod init tmp; GOBIN=${BIN} go get github.com/google/go-licenses@9376cf9847a05cae04f4589fe4898b9bce37e684
```
tejal29, line 34: I think this is failing in the integration test. you will have to move to a temporary directory and run go mod init there.

tejal29 (commented):
Left some comments on addressing lint and fixing errors in the integration test

Pending:
1) Verify the output and compare it against main. 

Would be great if  you also verify it on your end and make sure there is no change in the exported credits.

tejal29:
I see there is an open issue on the lint https://github.com/golang/lint/issues/503. Looks like the linter is deprecated. We should switch to some the options mentioned here https://golangci-lint.run/usage/false-positives/

hack/generate-embedded-files.sh
ericzzzzzzz: ohhhh.. thank you!! let me try that

ericzzzzzzz:
> 
> 1. Verify the output and compare it against main.
> 
> Would be great if you also verify it on your end and make sure there is no change in the exported credits.
schema comparison looks good. 
@tejal29  I checked the credit output from this branch against main, here is the difference
  - Only in /main/skaffold-credits/github.com: rakyll (expected) This is the statik mod we're removing in this pr. 
  - Only in /main/skaffold-credits/github.com: hashicorp (not expected), checked the content of github.com/hashicorp generated by **go-licenses** , this one is a little different, **go-licenses** only extract LICENSE for other mod, but for hashicorp it looks like copy a lot of other stuff including some go files into output. From my understanding, go:embed is expected to embed resource files, if embedding a sub-folder contains go files, it may skip that folder. The behavior on go:embed seems reasonable and **go-licenses** should not extract other thing except license from a mod, should we raise an issue on go-licenses repo?

tejal29 (approved):
LGTM!

Fix lint and merge

PR #7350, opened 2022-04-29
feat: helm init without artifact image overrides.

tejal29:
relates to #6683
This is implementation of go/sk-helm-init. 

In this PR, 

1) Add a helm analyzer. This analyzer will find all the dirs container `Chart.yaml` as it walk through user workspace.
The analyzer will also find any yaml files in a dir with `Chart.yaml` to create a map of Chart dir => values file.
2) Add deploy/helm.go which will initialize the helm deploy config for a project.

Things to do
- [ ] remote charts
- [ ] Integration test for subcharts & more value files.


MarlonGamez (approved):
nice! This looks good to me

Issue #7265, opened 2022-04-06
Update Golang-based examples to use go.mod


briandealwis:
All of Skaffold's Go -based examples should use go.mod and probably be updated to the latest release of Go.

GOPATH builds create problems for Go debugging as the build-time filepaths are embedded into the binary and almost certainly don't map to the user's local file system.  For example, our [`examples/getting-started`](https://github.com/GoogleContainerTools/skaffold/tree/main/examples/getting-started) uses a multistage build.  The build part uses the `golang:1.15` image:
https://github.com/GoogleContainerTools/skaffold/blob/ba6ba4f0415b331c9af9c135b010626bc3106dff/examples/getting-started/Dockerfile#L1-L5
The COPY command copies the source (`main.go`) from the user's workspace into the container's current directory (spolier: it's `/go`, which I determined by inspecting the image config, but neither the user nor us have any indication of this), and builds the app into a binary in the root as `/app`.

The second stage builds the deploy-time image, which uses a different base image (alpine:3):
https://github.com/GoogleContainerTools/skaffold/blob/ba6ba4f0415b331c9af9c135b010626bc3106dff/examples/getting-started/Dockerfile#L7-L12
The COPY command copies the built binary /app from the first stage into this otherwise-disconnected deploy image.

Connecting with dlv to the running result shows that the source files are believed to be in `/go`, even though the actual deploy-time image does not have a `/go`:
 ```
$ dlv connect localhost:56268
(dlv) sources
/go/main.go
/usr/local/go/src/errors/errors.go
/usr/local/go/src/errors/wrap.go
/usr/local/go/src/fmt/doc.go
```

Using Go modules, on the other hand, changes the file prefix to be relative to the module name.  This simplifies the debugger's life.

Cloud Code for IntelliJ requires use of Go modules.

----

Implementation notes:
- must be done to `integration/examples/` and `examples/`
- all go.mod should have unique module names: just append the example directory name.
    - `gopls`, the Go LSP, underpinning a lot of IDE tooling, has conniptions if it finds multiple go.mod files with the same module name

**Comment by johnmanjiro13** [comment]
I try to use go.mod in the Dockerfile, but prefix doesn't change.
Do you have any idea?

I create a go.mod and changed Dockerfile of the [`examples/getting-started`](https://github.com/GoogleContainerTools/skaffold/tree/main/examples/getting-started) like below.

go.mod
```
module github.com/GoogleContainerTools/skaffold/examples/getting-started

[...]

PR #7361, opened 2022-04-30
fix: Update Golang-based examples to use go.mod

johnmanjiro13:
Fixes: #7265

**Description**
Golang-based multistage built examples don't use go.mod.
I fixed it by below.
- Add go.mod
- Use golang:1.18 for builder image
- Use `WORKDIR` because `$GOPATH` is `/go` (which is the default current directory) by default in the golang image.
- Pass `-trimpath` flag to get the file prefix relative to the module name in debugging (ref. https://github.com/GoogleContainerTools/skaffold/issues/7265#issuecomment-1112521737).

### DIFF
--- a/examples/compose/Dockerfile
+++ b/examples/compose/Dockerfile
@@ -1,8 +1,10 @@
-FROM golang:1.15 as builder
+FROM golang:1.18 as builder
+WORKDIR /code
 COPY main.go .
+COPY go.mod .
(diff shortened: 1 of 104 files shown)


tejal29:
@johnmanjiro13 , Thanks a lot for this PR.

Can you please only update the examples in integration dir. 
As part of our release process, the examples get replaced by integration/examples.

johnmanjiro13:
@tejal29 I got it.
I removed changes from `examples` dir.

tejal29 (approved):
LGTM!

Issue #6521, opened 2021-08-26
skaffold  logtail not support unique color  when concurrency != 1 for more than one microservice


beastpu:
skaffold  logtail not support unique color  when concurrency set to != 1 for more than one microservice

gsquared94:
I ran the `microservices` example with concurrency 1 and got unique color pod name prefixes
![image](https://user-images.githubusercontent.com/39389231/131358828-72b7746a-31bd-4b6e-bfdf-d0f6e3899ead.png)

Can you share a small example that reproduces the issue that you're seeing?

beastpu:
sorry~   the color problem will appear when  concurrency set to 0 not 1

gsquared94:
tested with `examples/microservices`. We lose all colors even for deployers, loggers and port-forward messages if the artifacts are built with concurrency != 1. 
@MarlonGamez I think this could be a regression from the recent `skaffoldWriter` and `colorableWriter` changes.

akostic-kostile:
+1, it's very difficult to follow logs on several services with everything being same color. As a small addition to this bug report I'd like to say that, even tho I `skaffold dev` with concurrency=0, in my case this is an intermittent bug, sometimes I get colored output, sometimes not. But in fairness colored outputs seems much more rare than colorless.

GeraldXv:
+1

PR #7362, opened 2022-05-01
fix: no colored output with concurrency!=1

beastpu:
Fixes: #6521

**Description**
lose all colors even for deployers, loggers and port-forward messages if the artifacts are built with concurrency != 1. 
support colored output with concurrency!=1.

**Changes**

**before**
when concurrency != 1,  color is disabled in skaffoldWriter
`io.Writer` is `os.pipe` type, so `isTerm` always return false

```
// SetupColors conditionally wraps the input `Writer` with a color enabled `Writer`
func SetupColors(ctx context.Context, out io.Writer, defaultColor int, forceColors bool) io.Writer {

	_, isTerm := term.IsTerminal(out)
	supportsColor, err := term.SupportsColor(ctx)

	if err != nil {
		log.Entry(context.TODO()).Debugf("error checking for color support: %v", err)
	}

	useColors := (isTerm && supportsColor) || forceColors
	if useColors {
		// Use EnableColorsStdout to enable use of color on Windows
		useColors = false // value is updated if color-enablement is successful
		colorable.EnableColorsStdout(&useColors)
	}

	colors.Disable(!useColors)
	......
        ......
}
```

**after**

i I passed the `forceColors` parameter as true in the function of  `(l *logAggregatorImpl)getWriter `, `(isTerm && supportsColor) || forceColors` always return true

### DIFF
--- a/pkg/skaffold/build/result.go
+++ b/pkg/skaffold/build/result.go
@@ -62,7 +62,7 @@ func (l *logAggregatorImpl) GetWriter(ctx context.Context) (io.Writer, func(), e
 
 	writer := io.Writer(w)
 	if output.IsColorable(l.out) {
-		writer = output.GetWriter(ctx, writer, output.DefaultColorCode, false, false)
+		writer = output.GetWriter(ctx, writer, output.DefaultColorCode, true, false)
 	}
 	ch := make(chan string, buffSize)
 	l.messages <- ch


MarlonGamez:
hi @beastpu , thanks for opening this PR! This looks good to me

MarlonGamez (approved):
lgtm!

PR #7383, opened 2022-05-06
feat: implement first round of `skaffold fix` for v2beta28 to v3alpha1 config

MarlonGamez:
**Description**
This PR implements basic `skaffold fix` for going from `2beta28` to `v3alpha1`. Some features are still missing that are required for full implementation of `skaffold fix`
- [ ] kustomize render config supports old `deploy.kustomize.buildArgs` config
- [ ] kubectl render config supports `deploy.kubectl.remoteManifests` config
- [ ] kpt render config is fully designed

until these things are resolved, `skaffold fix` can't be fully implemented, so I will follow up on those.

This PR also removes the `pkg/skaffold/schema/latest/v1` package and puts the relevant code into `pkg/skaffold/schema/v2beta28`, since we should start treating `v3alpha1` as the natural continuation of `v2beta28`

I've also made a change to ensure that `manifests.helm.releases` uses the same underlying type as `deploy.helm.releasese`, which it wasn't before.


tejal29 (approved):
LGTM! 

fix linting issue

aaron-prindle:
LGTM.  One nit I see is that currently the following is generated in the resulting fixed yaml:
```
deploy:
  kubectl: {}
```

Would it make sense to instead not generate anything?  I think this is ok but IIUC, the preffered yaml wouldn't have this empty entry that is the default

aaron-prindle (approved):

aaron-prindle:
Additionally it might make sense to return an error and print to stdout some error message when attempting to use w/ helm deployer

MarlonGamez:
> LGTM. One nit I see is that currently the following is generated in the resulting fixed yaml:
> 
> ```
> deploy:
>   kubectl: {}
> ```
> 
> Would it make sense to instead not generate anything? I think this is ok but IIUC, the preffered yaml wouldn't have this empty entry that is the default

I wasn't sure if we wanted to have implied deployer config when generating config for the user here, as they'd go from having explicitly written kubectl deploy config to having none in their resulting yaml. Wdyt? @aaron-prindle @tejal29

tejal29:
> I wasn't sure if we wanted to have implied deployer config when generating config for the user here, as they'd go from having explicitly written kubectl deploy config to having none in their resulting yaml. Wdyt? @aaron-prindle @tejal29

I like the reasoning. Users can delete it if they need by hand. I am ok with the current approach.

PR #7384, opened 2022-05-08
Fix #can-images-be-debugged-without-the-runtime-support-images link in debug.md doc

nelsonjchen:
**Description**
Just fixing some missing angle bracket to reference an anchor link


nelsonjchen:
image

[It looks like this at the moment.](https://skaffold.dev/docs/workflows/debug/)

Issue #7388, opened 2022-05-10
Ko documentation example is not valid 


hown3d:
Ko documentation has a wrong example regarding multi-platform builds:
Current YAML of [example ](https://skaffold.dev/docs/pipeline-stages/builders/ko/#multi-platform-images)is not valid, platforms array needs to be inside of **ko** map.

PR #7389, opened 2022-05-10
docs: fix ko multi-platform example

hown3d:
Fixes: #7388

**Description**
Configuration mistake for ko multiplatform build example. Didn't work when using it the old way.


aaron-prindle (approved):
LGTM!  Thanks for the fix

Issue #5892, opened 2021-05-21
Write out artifact cache at completion of each successful build


briandealwis:
@dhodun noted that when build a suite of artifacts, the artifact cache only seems to be updated after all builds complete successfully.  So after fixing the error, each artifact is then rebuilt even if it had been successfully built previously.  This matters when using a remote build.

We should write out the artifact cache after each successful build.

PR #7394, opened 2022-05-11
[WIP] fix: Write out artifact cache with successful builds even on task failure #5892

dhodun:
**Closes:** https://github.com/GoogleContainerTools/skaffold/issues/5892

**Description:** WIP PR for comments / feedback.

Caches successfully built artifacts when one build task fails, preventing rebuilds from reoccurring. Particularly helpful with remote builds.


dhodun:
Will port this to main after a few changes.

tejal29:
> **Closes:** #5892
> 
> **Description:** WIP PR for comments / feedback.
> 
> Caches successfully built artifacts when one build task fails, preventing rebuilds from reoccurring. Particularly helpful with remote builds.

Are you thinking of contributing this to CD users hence contributing to v1.37.X branch

aaron-prindle:
Talked w/ @dhodun  offline, their plan is to get feedback on the implemented changes on v1.37.2 as the tests were more stable in the v1 branch when they started but to migrate their changes over to submit the actual PR onto `main`.  This feature is not particular to Cloud Deploy

tejal29:
> Talked w/ @dhodun offline, their plan is to get feedback on the implemented changes on v1.37.2 as the tests were more stable in the v1 branch when they started this work and concurrently migrate their changes over to submit the actual PR onto `main`. This feature is not particular to Cloud Deploy or v1

If that is the case, can we change the base to v1 instead of v1.37.2? 
I am going to create a v1.39.0 which is also going to be off the same base.

pkg/skaffold/build/builder_mux.go
```diff
@@ -37,6 +37,11 @@ type BuilderMux struct {
 	byImageName map[string]PipelineBuilder
 	store       ArtifactStore
 	concurrency int
+	cache       Cache
+}
+
+type Cache interface {
+	AddArtifacts(ctx context.Context, bRes []graph.Artifact) error
```
gsquared94, line 44: can this instead now be `AddArtifacts(ctx context.Context, bRes graph.Artifact) error` ?

gsquared94:
changes look good, other than one comment.

pkg/skaffold/build/builder_mux.go
dhodun: The function still accepts a slice of `graph.Artifact` and will add them all, so plural is appriorate:
https://github.com/GoogleContainerTools/skaffold/blob/c9d3521350f661cc32f587dad1c7ef39da21cc4d/pkg/skaffold/build/cache/retrieve.go#L174

Technically it is only being called in the singleton case, so I could refactor the function to make singular?

https://github.com/GoogleContainerTools/skaffold/blob/c9d3521350f661cc32f587dad1c7ef39da21cc4d/pkg/skaffold/build/builder_mux.go#L110-L114

gsquared94: I think that we should refactor it to make it singular.

dhodun: Done.

gsquared94 (approved):
LGTM!

PR #7401, opened 2022-05-12
chore: add caching for tests

tejal29:
All our integration tests first run `skaffold build` to fail fast and then run the command to be tested. 

Before including run mode (some time before #4893) in the artifact hash,  the artifact hash for skaffold build and skaffold dev were same. 
The tests could rely on the built image from the previous build command. 
There was reason for adding RunModes in the artifact hash. 

I am adding a new RunMode `testing` which can be used in the testing. This will reduce the test times and subsequent skaffold commands will use the built artifact from the previous skaffold built.

Time reduced : [16 mins](https://github.com/GoogleContainerTools/skaffold/actions/workflows/integration-linux.yml?query=actor%3AMarlonGamez) to  < 9 min(https://github.com/GoogleContainerTools/skaffold/runs/6415058294?check_suite_focus=true)

### DIFF
--- a/integration/skaffold/helper.go
+++ b/integration/skaffold/helper.go
@@ -346,6 +346,7 @@ func (b *RunBuilder) cmd(ctx context.Context) *exec.Cmd {
 	if value, found := os.LookupEnv("SKAFFOLD_BINARY"); found {
 		skaffoldBinary = value
 	}
+	b.env = append(b.env, "SKAFFOLD_INT_TEST=true")
 	cmd := exec.CommandContext(ctx, skaffoldBinary, args...)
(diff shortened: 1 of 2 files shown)


gsquared94 (changes requested):
The `runMode` is used in more places that the artifact hash. We use it for adding different build args for the docker builder, environment variables for buildpacks in `dev` vs `debug` scenarios. For integration tests I think we should test the real scenarios. As such I think we shouldn't make this change. 

It also detracts from `runModes` being 1:1 with the skaffold commands.

tejal29:
Let's revist later.

Issue #7382, opened 2022-05-06
statusCheck in skaffold.yaml is being ignored


ansky:
I think I may be impacted by #5210 so I added `statusCheck: false` to my skaffold.yaml but deployment is still failing.  When I use `--status-check=false` on the command line it works.

### Expected behavior
Using statusCheck in skaffold.yaml works the same as --status-check on the command line.

### Actual behavior
`statusCheck: false` is being ignored.  It's the same behavior with or with out it.

### Information

- Skaffold version: v1.37.1
- Operating system: Debian GNU/Linux 11 (bullseye)
- Installed via: Cloud Shell
- Contents of skaffold.yaml:

```
apiVersion: skaffold/v2beta16
kind: Config
build:
  artifacts:
  - image: gcr.io/joelfriedman-gitea-to-deploy/github.com/ansky/gitea
    context: /workspace/source
    docker:
      dockerfile: Dockerfile
  tagPolicy:
    gitCommit: {}
  local: {}
deploy:
  statusCheck: false
  helm:
    releases:
    - name: gitea
      chartPath: /workspace/source/helm-chart
      valuesFiles:
      - /workspace/source/helm-chart/values.yaml
      artifactOverrides:
        imageKey: gcr.io/joelfriedman-gitea-to-deploy/github.com/ansky/gitea
      skipBuildDependencies: true
  logs:
    prefix: container
```

### Steps to reproduce the behavior

1. skaffold apply --filename='skaffold.yaml' 'manifest.yaml'

End of debug logs from running above...
```
...
DEBU[0003] 13manifests to deploy.13are updated or new    subtask=-1 task=DevLoop
DEBU[0003] Running command: [kubectl --context gke_joelfriedman-gitea-to-deploy_us-east1_gitea-staging apply -f -]  subtask=-1 task=DevLoop
 - serviceaccount/gitea-memcached unchanged
 - secret/gitea-postgresql unchanged
 - secret/gitea configured
 - secret/gitea-init configured
 - service/gitea-memcached configured
 - service/gitea-postgresql-headless configured
 - service/gitea-postgresql configured
 - service/gitea-http configured
 - service/gitea-ssh configured
 - Warning: Autopilot increased resource requests for Deployment default/gitea-memcached to meet requirements. See http://g.co/gke/autopilot-resources.
 - deployment.apps/gitea-memcached configured
[...]

PR #7403, opened 2022-05-13
fix: `apply` not using config set `kubeContext` or `statusCheck`

gsquared94:
Fixes: #7382 
**Description**

The `skaffold apply` code path was initializing a dummy `kubectlDeploy` without setting `kubeContext` or `statusCheck`.

**User facing changes **

Users will be able to set `kubeContext` and `statusCheck` in their `skaffold.yaml` file and apply it with `skaffold apply` command.


tejal29 (commented):
Do we need this changes at L111 where 
```
dCtx := &deployerCtx{runCtx, d}
```

gsquared94:
> Do we need this changes at L111 where
> 
> ```
> dCtx := &deployerCtx{runCtx, d}
> ```

we don't need it there because `d` has it set correctly.

tejal29 (approved):

PR #7421, opened 2022-05-18
feat: Add helm init images for multiple services project

tejal29:
Add helm init for images. 

In this PR, 
1) In the `GetImages` functionality, instead of using chart names which could be used in the template, I made the following changes
 - run `helm template --output-dir <tmpDir> -f values.yaml`
 - Parse all the templated k8s manifests in tmpDir and get images.

To try this out
1) LOCAL=true make install
2) git clone https://github.com/tejal29/helm-examples
3) cd go-guestbook
4) run 
```
skaffold init -f skaffold.yaml.out

diff skaffold.yaml skaffold.yaml.out
```
5) Run skaffold dev and make sure the example is running.
```
skaffold run -d <your registry> 
```


integration/init_test.go
```diff
@@ -52,18 +57,17 @@ func TestInitCompose(t *testing.T) {
 
 			initArgs := append([]string{"--force"}, test.args...)
 			skaffold.Init(initArgs...).InDir(test.dir).WithConfig("skaffold.yaml.out").RunOrFail(t.T)
-
 			checkGeneratedConfig(t, test.dir)
 
 			// Make sure the skaffold yaml and the kubernetes manifests created by kompose are ok
```
aaron-prindle, line 62: nit: this comment doesn't make sense anymore as this test is no longer just for compose/kompose testing but also for helm

integration/init_test.go
```diff
@@ -29,21 +29,26 @@ import (
...
+					args: []string{"--compose-file", "docker-compose.yaml"},
+				},
+		*/
 		{
-			name: "compose",
-			dir:  "testdata/init/compose",
-			args: []string{"--compose-file", "docker-compose.yaml"},
+			name: "helm init",
+			dir:  "testdata/init/helm-project",
+			args: []string{"--XXenableBuildpacksInit=false"},
```
aaron-prindle, line 51: nit: For users using `skaffold init` do we expect them to know to use this flag?

integration/testdata/init/helm-project/skaffold.yaml
```diff
@@ -0,0 +1,19 @@
...
+kind: Config
+metadata:
+  name: helm-project
+build:
+  artifacts:
+  - image: go-hello
+    context: hello-world/src
+    docker:
+      dockerfile: Dockerfile
+deploy:
```
aaron-prindle, line 11: Question: is there a way to generate the version of schema with a `manifests` field as well?  Is that something we think users might want?

aaron-prindle (approved):
LGTM

integration/init_test.go
tejal29: no. I added the flag to force disable detecting `Buildpacks` builder.  If I keep the flag, there are potentially 2 ways of building the image
1) Using docker
2) USing buildpacks. 
Users will be presented on the CLI with that option.  They could choose which builder to use.

The integration tests would fail with an error saying "user input needed to determine builder". Hence I disabled it.

Issue #7419, opened 2022-05-17
integration/TestDebug helm is failing


ericzzzzzzz:
#7412 commented out the test as it was blocking other prs, we need to find the root cause of why it's failing and fix it. 
The initial investigation seems that the manifests transformation for Helm is not happening. 
Here is the error log. 

=== Failed Tests ===
578/integration/TestDebug
579--- FAIL: TestDebug (346.31s)
580    helper.go:245: Ran [skaffold debug --namespace skaffold79m7c --default-repo [gcr.io/k8s-skaffold](http://gcr.io/k8s-skaffold)] in 4.165 seconds
581/integration/TestDebug/helm
582    helper.go:259: Running [skaffold build --default-repo [gcr.io/k8s-skaffold](http://gcr.io/k8s-skaffold)] in examples/helm-deployment
583    helper.go:259: Ran [skaffold build --default-repo [gcr.io/k8s-skaffold](http://gcr.io/k8s-skaffold)] in 44.797 seconds
584time="2022-05-17T18:49:57Z" level=info msg="Namespace: skaffold79m7c" subtask=-1 task=DevLoop
585    debug_test.go:86: Running [skaffold debug --namespace skaffold79m7c --default-repo [gcr.io/k8s-skaffold](http://gcr.io/k8s-skaffold)] in examples/helm-deployment
586time="2022-05-17T18:49:57Z" level=info msg="Waiting for deployments [skaffold-helm] to stabilize" subtask=-1 task=DevLoop
587time="2022-05-17T18:49:59Z" level=info msg="Deployment skaffold-helm: Generation 0/1, Replicas 0/2, Available 0/2" subtask=-1 task=DevLoop
588time="2022-05-17T18:49:59Z" level=info msg="Deployment skaffold-helm: Generation 0/1, Replicas 0/2, Available 0/2" subtask=-1 task=DevLoop
589time="2022-05-17T18:49:59Z" level=info msg="Deployment skaffold-helm: Generation 1/1, Replicas 0/2, Available 0/2" subtask=-1 task=DevLoop
590time="2022-05-17T18:49:59Z" level=info msg="Deployment skaffold-helm: Generation 1/1, Replicas 2/2, Available 0/2" subtask=-1 task=DevLoop
591time="2022-05-17T18:50:01Z" level=info msg="Deployment skaffold-helm: Generation 1/1, Replicas 2/2, Available 1/2" subtask=-1 task=DevLoop
[...]

PR #7397, opened 2022-05-11
chore: fix TestDockerDebug integration test

tejal29:
fixes: #7030

PR #7432, opened 2022-05-20
chore: fix integration/test_debug/helm test

ericzzzzzzz:
Fixes: #7419 

**Related** 
 - #7397 introduced testing mode which causes integration test_debug helm test failure 
 - we had a discussion about adding testing mode in #7401 , but the conclusion was not to add testing mode, not sure #7497 added testing mode on purpose, if it's accidentally added then we can remove it and this pr is not necessary. 

**Description**
 -  The test failed due to running with testing run mode while helm deployer relies on debug RunMode to determine if it adds --debugging filter
 - this change add a flag to bypass testing mode in integration test

### DIFF
--- a/integration/skaffold/helper.go
+++ b/integration/skaffold/helper.go
@@ -346,7 +346,6 @@ func (b *RunBuilder) cmd(ctx context.Context) *exec.Cmd {
 	if value, found := os.LookupEnv("SKAFFOLD_BINARY"); found {
 		skaffoldBinary = value
 	}
-	b.env = append(b.env, "SKAFFOLD_INT_TEST=true")
 	cmd := exec.CommandContext(ctx, skaffoldBinary, args...)
 	cmd.Env = append(removeSkaffoldEnvVariables(util.OSEnviron()), b.env...)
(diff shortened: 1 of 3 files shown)


integration/skaffold/helper.go
```diff
@@ -346,7 +353,9 @@ func (b *RunBuilder) cmd(ctx context.Context) *exec.Cmd {
 	if value, found := os.LookupEnv("SKAFFOLD_BINARY"); found {
 		skaffoldBinary = value
 	}
-	b.env = append(b.env, "SKAFFOLD_INT_TEST=true")
+	if !b.cmdRunMode {
+		b.env = append(b.env, "SKAFFOLD_INT_TEST=true")
+	}
```
tejal29, line 358: Thanks @ericzzzzzzz for catching this. I didnt mean to merge this in.

tejal29: Lets revert this change `SKAFFOLD_INT_TEST=true` and then rebase to see if just reverting fixes the issue

tejal29 (approved):
LGTM!

Issue #7342, opened 2022-04-27
Support region configuration for googleCloudBuild shared pool


henrybell:
Currently, the [`googleCloudBuild`](https://skaffold.dev/docs/references/yaml/#build-googleCloudBuild) configuration supports the non-regional/"global" Cloud Build shared pool, as well as the specification of a [private pool](https://skaffold.dev/docs/references/yaml/#build-googleCloudBuild-workerPool). For organizations whose Org Policy does not permit the use of the non-regional/"global" shared pool, it would be useful to be able to supply a [region](https://cloud.google.com/build/docs/locations) as part of the Cloud Build configuration, so that the build can run in the regional shared pool (i.e. without having to create and manage a private pool).

PR #7436, opened 2022-05-23
chore: add-schema-v2beta29

ericzzzzzzz:
Fixes: #7342 

**Description**
 - add a new version schema to support adding region configuration in GoogleCloudBuild in a following up pr

 - support region configuration in GoogleCloudBuild https://github.com/ericzzzzzzz/skaffold/tree/add-location-configuration-for-gcb


tejal29 (approved):
LGTM!

Issue #7240, opened 2022-03-29
Skaffold should warn when no pods have a matching image for sync


briandealwis:
Consider a setup where a user mistakenly fails to reference their artifact in the skaffold.yaml from a k8s manifest:

```
build:
  artifacts:
  - image: mynodejsapp
    context: .
    sync:
      infer: 
      - '**/*.js'  
```

and
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: mynodejsapp-deployment
...
spec:
      containers:
      - name: mynodejsapp
        image: busybox
        command: ["/bin/sh", "-c", "sleep 200"]
```

When they run `skaffold dev` and change a file, sync will indicate that a file is sync'd but no actual syncing will occur:
```
Starting deploy...
 - deployment.apps/mynodejsapp-deployment created
 - service/mynodejsapp-service created
Waiting for deployments to stabilize...
 - deployment/mynodejsapp-deployment is ready.
Deployments stabilized in 4.337 seconds
Press Ctrl+C to exit
Watching for changes...
Syncing 1 files for mynodejsapp:0354fe4f17a5cfd2c07171b2c7b34ed4dc6713e175acfe6423fec025ed9325d9
Watching for changes...
```

Sync should indicate when no pods are matched for the image.

Issue #7447, opened 2022-05-26
Files are synched more times than needed when running `skaffold dev`


renzodavid9:
When running `skaffold dev` with the sync feature enable, in a project with **N** different configs (different `skaffold.yaml` files), every time skaffold syncs a file, it will try to do it **N** times: it will look for the corresponding image to sync the file **N** times.

### Expected behavior
A file sync, in a multi config project, running `skaffold dev`, should look for the image to sync the file only once, independently of how many configs we have in the project.

### Actual behavior
Instead, when a file sync is triggered, skaffold will try to find the corresponding image more than once, running the logic more times than needed.

### Information

- Skaffold version: v1.38.0-63-ga1f4e4b85-dirty
- Operating system: MacOS Monterey
- Installed via: Local build using the `main` branch
- Contents of skaffold.yaml:

We can use the  example project [multi-config-microservices](https://github.com/GoogleContainerTools/skaffold/tree/main/examples/multi-config-microservices), enabling the `sync` feature:

`leeroy-app` skaffold.yaml file with `sync`:
```yaml
apiVersion: skaffold/v2beta28
kind: Config
metadata:
  name: app-config
requires:
- path: ../base
build:
  artifacts:
  - image: leeroy-app
    sync: 
        infer:
        - "**/*.go"
    requires:
    - image: base
      alias: BASE
deploy:
  kubectl:
    manifests:
    - kubernetes/*
portForward:
- resourceType: deployment
  resourceName: leeroy-app
  port: http
  localPort: 9001
```

`leeroy-web` skaffold.yaml file with `sync`:
```yaml
apiVersion: skaffold/v2beta28
kind: Config
metadata:
  name: web-config
requires:
- path: ../base
build:
  artifacts:
  - image: leeroy-web
    sync: 
      infer:
      - "**/*.go"
    requires:
    - image: base
      alias: BASE
deploy:
  kubectl:
    manifests:
    - kubernetes/*
portForward:
- resourceType: deployment
  resourceName: leeroy-web
  port: 8080
  localPort: 9000
```
The other yaml files in the example don't need a change.

### Steps to reproduce the behavior

1. Open the [multi-config-microservices](https://github.com/GoogleContainerTools/skaffold/tree/main/examples/multi-config-microservices) example
2. Modify the `skaffold.yaml` files listed above to enable the sync feature
3. Check the `for`loop in:
[...]

PR #7448, opened 2022-05-26
feat: add warning message when no pods are matched for an image doing sync

renzodavid9:
Fixes: #7240

**Follow-up Work**
- With this change, the warning message could appear multiple times in one dev loop, for a single file. [This issue](https://github.com/GoogleContainerTools/skaffold/issues/7447) was created to follow-up on the work needed to check that behavior.

### DIFF
--- a/pkg/skaffold/sync/syncer_mux.go
+++ b/pkg/skaffold/sync/syncer_mux.go
@@ -18,9 +18,12 @@ package sync
 
 import (
 	"context"
+	"fmt"
 	"io"
 
 	"github.com/pkg/errors"
+
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/output/log"
 )
 
 type SyncerMux []Syncer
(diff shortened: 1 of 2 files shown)


pkg/skaffold/sync/syncer_mux.go
```diff
@@ -33,13 +35,21 @@ func (s SyncerMux) Sync(ctx context.Context, out io.Writer, item *Item) error {
 		}
 	}
 
-	// Return an error only if all syncers fail
-	if len(errs) == len(s) {
-		var err error
+	if len(errs) > 0 {
+		err := fmt.Errorf("sync failed for artifact %q", item.Image)
```
MarlonGamez, line 39: do we need to wrap with this error? I think this might lead to something a bit awkward looking, i.e.
```
sync failed for artifact foo: copying files: sync failed for artifact foo
```
unless I'm misunderstanding the code path. Wdyt @renzodavid9 ?

MarlonGamez (approved):
one concern, other than that looks good! And I think there are just a couple linter errors to be addressed

pkg/skaffold/sync/syncer_mux.go
renzodavid9: You are right, the message will have all the errors found during sync. In this case, and due to #7447, the message will look something like this:

`Skipping deploy due to sync error:copying files: didn't sync any files: copying files: didn't sync any files: sync failed for artifact foo`

Not sure then if it's better to manage this error here differently (without doing the `for` for the wrapping) or just wait until we solve #7447 to avoid having the same message duplicated in the error.

MarlonGamez: I think maybe we can just leave it for now, and if we need to clean up because it's confusing we can do so later

ericzzzzzzz:
Hi @renzodavid9  I think you need to re-order imports to fix the linter issue.
change your imports to the following
```go
import (
	"context"
	"fmt"
	"io"

	"github.com/pkg/errors"

	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/output/log"
)
```

There are 3 groups here:
 - built-in packages
 - third party packages
 - skaffold packages

groups should be separated by an empty line.

PR #7450, opened 2022-05-27
chore: update CODEOWNERS

briandealwis:
Update with recent team changes.  I don't think we need a global owner since `skaffold-team` already owns this code.  I *think* that should address `skaffold-team` being requested on all reviews and let us benefit from GitHub's round-robin assignments.


aaron-prindle (approved):
LGTM

Issue #7442, opened 2022-05-25
Allow use of url manifests when specifying manifests


MarlonGamez:
### Expected behavior
Skaffold v2 should be able to apply URL manifests specified in `manifests.rawYaml`

### Actual behavior
Skaffold v2 doesn't currently have the piece of code needed for this ported over, so it doesn't work

PR #7459, opened 2022-05-30
WIP: feat: allow use of URL manifests

kadern0:
Fixes: #7442  

**Description**

This change allows using remote manifests as in:
```
apiVersion: skaffold/v2beta28
kind: Config
build:
  artifacts:
  - image: skaffold-example
deploy:
  kubectl:
    manifests:
      - https://raw.githubusercontent.com/GoogleContainerTools/skaffold/main/examples/getting-started/k8s-pod.yaml

```

**Follow-up Work**
 - This feature should be documented
 - Also I have a couple questions about temporary files/dirs, hence the WIP. Should I try to maintain the original filename if possible? (that was my initial approach and I've left some commented out code on purpose), although then I thought it is easier using tmp files. How does skaffold  deal with cleaning up the `ManifestTmpDir`? The first time I run this new code it was failing because there was already some file within that directory. So it might be useful having an additional subdirectory for the downloaded manifests (gcs and urls) and perform a cleanup before downloading anything. Opinions?

### DIFF
--- a/pkg/skaffold/deploy/kubectl/kubectl.go
+++ b/pkg/skaffold/deploy/kubectl/kubectl.go
@@ -288,10 +288,11 @@ func (k *Deployer) PostDeployHooks(ctx context.Context, out io.Writer) error {
 }
 
 func (k *Deployer) manifestFiles(manifests []string) ([]string, error) {
-	var nonURLManifests, gcsManifests []string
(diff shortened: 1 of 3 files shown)


kadern0:
Also adding tests would be useful.

PR #7468, opened 2022-05-31
chore: re-enableintegration/deploy_test, update test config

ericzzzzzzz:
Fixes:  integration/deploy_test, update test config

### DIFF
--- a/integration/examples/kustomize/skaffold.yaml
+++ b/integration/examples/kustomize/skaffold.yaml
@@ -1,4 +1,10 @@
 apiVersion: skaffold/v3alpha1
 kind: Config
+manifests:
+  kustomize:
+    paths:
+      - .
 deploy:
-  kustomize: {}
+  kustomize:
+    paths:
+      - .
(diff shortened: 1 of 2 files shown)


integration/examples/kustomize/skaffold.yaml
```diff
@@ -1,4 +1,10 @@
...
 kind: Config
+manifests:
+  kustomize:
+    paths:
+      - .
 deploy:
-  kustomize: {}
+  kustomize:
+    paths:
+      - .
```
MarlonGamez, line 10: I think we should be able to/want to do this with kustomize renderer -> kubectl deployer. If we can get rid of the kustomize deploy stuff here that would be great

MarlonGamez (changes requested):

integration/examples/kustomize/skaffold.yaml
ericzzzzzzz: ohhh I see! thanks!

MarlonGamez (approved):

PR #7501, opened 2022-06-02
fix: partition GCP integration tests

MarlonGamez:
**Description**
This PR is a WIP to get partitioned tests working in kokoro. This will help us unblock integration testing for v2

I've also added a log line for each test to say which split it's in, so you can more easily click into the integration test split you're looking for


aaron-prindle (approved):
LGTM!

PR #7520, opened 2022-06-08
feat: add hidden flag --fast-fail-status-check to allow toggling on/off the fast-fail behavior for status checking

aaron-prindle:


MarlonGamez (approved):
lgtm ✅

PR #7527, opened 2022-06-09
fix: plumb manifests through Apply to get functionality working

### LINKED ISSUE #7526: Fix issue in Skaffold V2 where currently no manifests are passed to Apply
Currently there is a TODO item
https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/runner/v2/apply.go#L32

to fix `Apply` s.t. manifests are parsed out of the hydration dir and passed to `Apply`.  Currently `Apply` is not fully implemented as no manifests are passed to the method currently

This unimplemented `Apply` funcitonality is blocking partners from implementing their work items, bumping this to be done in the current milestone

aaron-prindle:
Fixes #7526

### DIFF
--- a/cmd/skaffold/app/cmd/delete.go
+++ b/cmd/skaffold/app/cmd/delete.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	deployutil "github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy/util"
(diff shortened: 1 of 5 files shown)


MarlonGamez (approved):
LGTM, do we have tests for these? I know we have apply tests but I haven't investigated them enough to know how much they cover. We can always open an issue and add them later

Issue #7551, opened 2022-06-14
flake: TestDevAutoSyncAPITrigger


aaron-prindle:
Currently TestDevAutoSyncAPITrigger appears to flake with the error:
```
== RUN   TestDevAutoSyncAPITrigger
time="2022-06-14T19:40:16Z" level=trace msg="Partitions: " subtask=-1 task=DevLoop
time="2022-06-14T19:40:16Z" level=trace msg="P0 total: 300.000000, size: 6\n" subtask=-1 task=DevLoop
time="2022-06-14T19:40:16Z" level=trace msg="P1 total: 299.980000, size: 8\n" subtask=-1 task=DevLoop
time="2022-06-14T19:40:16Z" level=trace msg="P2 total: 299.990000, size: 15\n" subtask=-1 task=DevLoop
time="2022-06-14T19:40:16Z" level=trace msg="P3 total: 87.830000, size: 16\n" subtask=-1 task=DevLoop
    util.go:95: Test partition: 3
time="2022-06-14T19:40:16Z" level=debug msg="getting client config for kubeContext: ``" subtask=-1 task=DevLoop
time="2022-06-14T19:40:16Z" level=info msg="Namespace: skaffoldfhb8v" subtask=-1 task=DevLoop
    helper.go:259: Running [skaffold build --namespace skaffoldfhb8v -f skaffold-manual.yaml --default-repo gcr.io/k8s-skaffold] in testdata/file-sync
    helper.go:259: Ran [skaffold build --namespace skaffoldfhb8v -f skaffold-manual.yaml --default-repo gcr.io/k8s-skaffold] in 1.012 second
    sync_test.go:237: Running [skaffold dev --namespace skaffoldfhb8v -f skaffold-manual.yaml --default-repo gcr.io/k8s-skaffold --auto-sync=false --rpc-port 33115] in testdata/file-sync
    rpc_test.go:399: waiting for connection...
    rpc_test.go:399: waiting for connection...
time="2022-06-14T19:40:19Z" level=info msg="Waiting for pods [test-file-sync] to be ready" subtask=-1 task=DevLoop
time="2022-06-14T19:40:19Z" level=info msg="Still waiting for pods [test-file-sync]" subtask=-1 task=DevLoop
time="2022-06-14T19:40:20Z" level=info msg="Pods marked as ready: map[test-file-sync:true]" subtask=-1 task=DevLoop
panic: assignment to entry in nil map
goroutine 19 [running]:
[...]

PR #7553, opened 2022-06-15
chore: fix flaky test -TestDevAutoSyncAPITrigger

ericzzzzzzz:
Fixes: #7551  
**Related**:  
 - The same issue as #5615

**Test Flakyness Reason**
 - rpc.ClientAutoSync will invoke /proto.SkaffoldService/AutoSync method which calls [UpdateStateAutoSyncTrigger](https://github.com/ericzzzzzzz/skaffold/blob/TestDevAutoSyncAPITrigger/pkg/skaffold/event/event.go#L765-L769) method, this does deepCopy for handler.state but state.StatusCheckState.Resource will be set to nil if the original Resource is an empty map. 
 - [pollResourceStatus](https://github.com/ericzzzzzzz/skaffold/blob/TestDevAutoSyncAPITrigger/pkg/skaffold/kubernetes/status/status_check.go#L235) emits Event_ResourceStatusCheckEvent, if ``UpdateStateAutoSyncTrigger`` happens first then Resources map will become nil from empty map , so when event handler processes the event it will fail. If Event_ResourceStatusCheckEvent happen first, then things are ok. 

**Description**
 - assign a new map to ev.state.StatusCheckState.Resources if nil when handling Event_ResourceStatusCheckEvent

### DIFF
--- a/pkg/skaffold/event/event.go
+++ b/pkg/skaffold/event/event.go
@@ -654,6 +654,9 @@ func (ev *eventHandler) handleExec(f firedEvent) {
 		rse := e.ResourceStatusCheckEvent
 		rseName := rse.Resource
 		ev.stateLock.Lock()
+		if ev.state.StatusCheckState.Resources == nil {
+			ev.state.StatusCheckState.Resources = map[string]string{}
+		}
 		ev.state.StatusCheckState.Resources[rseName] = rse.Status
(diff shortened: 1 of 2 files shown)


aaron-prindle (approved):
Awesome!  LGTM!

Issue #7429, opened 2022-05-19
Support port forwarding for Cloud Run


bskaplan:
Additional feature for the Cloud Run Deployer: supporting port forwarding.

Gcloud has a built in proxy for forwarding traffic to the remote container, `gcloud beta run services proxy`, that automatically applies the necessary credentials for a service with public ingress but guarded by IAM invoker permissions.

PR #7556, opened 2022-06-15
Feat: Implement Services Port Forwarding for Cloud Run

bskaplan:
Fixes: #7429

**Description**
Feat: Support Port Forwarding with the Cloud Run deployer

The feature relies on gcloud being present on the user's machine and on the path. It can forward requests for services that are using IAM authentication if the currently logged in user has permission to access the service and the service is set to allow ingress from external sources.

The Cloud Run deployer now works with the --port-forward flag. Since Cloud Run only exposes a single port, only the service port is forwarded, not debug ports or user-defined ports.


aaron-prindle:
Still going through the code, in playing with this having some issues with some functionality:
- `skaffold dev` does not port forward when it should (was able to get things working with `skaffold run --port-forward=true`)
- unclear how to use Skaffold's `portForward` configuration to specify the port(s) to forward, in attempting the below with Cloud Run service using the `gcr.io/cloudrun/hello` image:
```
...
portForward:
- resourceType: Service
  resourceName: skaffold-demo-1
  port: 8080
  localPort: 9000
```
I was not able to get the actual port to change from local port 8080

- unrelated to this PR but delete/cleanup (for `skaffold dev`) expect the rendered manifests to be in the hydrationDir (by default, `.kpt-pipeline`).  As currently the Cloud Run setup does not generate any manifests there, delete/cleanup is not working as desired.

pkg/skaffold/deploy/cloudrun/accessor.go
```diff
@@ -0,0 +1,175 @@
...
+	cmd := exec.Command("gcloud", "--version")
+	return cmd.Run() == nil
+}
+
+type runProxyForwarder struct {
+	resources *resourceTracker
+}
+
+func (r *runProxyForwarder) Start(ctx context.Context, out io.Writer) error {
+	if !gcloudInstalled() {
```
aaron-prindle, line 124: likely makes sense to also check that the cloud run proxy is installed here and add a related actionable error as well

pkg/skaffold/deploy/cloudrun/accessor.go
```diff
@@ -0,0 +1,175 @@
...
+		output.Red.Fprintln(out, "gcloud not found on path. Unable to set up Cloud Run port forwarding")
+		return sErrors.NewError(fmt.Errorf("gcloud not found"), &proto.ActionableErr{ErrCode: proto.StatusCode_PORT_FORWARD_RUN_GCLOUD_NOT_FOUND})
+	}
+	for _, resource := range r.resources.resources {
+		if resource.port == 0 {
```
aaron-prindle, line 129: not sure the cause but resource.port always seems to be 0 even if the value is set via the `portForward` schema

pkg/skaffold/deploy/cloudrun/accessor.go
bskaplan: Hmm, this is a bit complex. If gcloud is installed manually, it seems it'll automatically install it. It only runs into a problem if you need root to install components or it was installed through a package manager. I can add the check, but it'll lead to a worse experience in one case.

pkg/skaffold/deploy/cloudrun/accessor.go
bskaplan: Added code to plumb through port overrides. It doesn't support changing the address, although I can see about adding that to gcloud as well. 

I did re-use the "service" resource type here. I can add a new type for cloud run services, but then I'd have to make sure the kubectl forwarder doesn't error out on cloud run resources.

pkg/skaffold/deploy/cloudrun/accessor.go
aaron-prindle: I think it is ok to leave as is, the context was mainly around wanting to make sure that this error is actionable as only actionable errors are properly bubbled up in our metrics pipeline s.t. Cloud Code can piece out how many users might hit this.

I'm not sure I understand how doing an explicit check leads to a worse experience in one case vs the other atm though

aaron-prindle:
Currently forwarding of type `service` for k8s Services on skaffold requires the `--port-forward=true` flag to be set.  K8s Deployments though are forwarded by default.  I believe it is likely better UX to have it so that vanilla:
`skaffold dev` 
forwards the ports of services for the Cloud Run deployer as it will make the default continuous interation not that useful if the service cannot be accessed easily locally.

Currently though:
`skaffold dev --port-forward` works.

It might make sense to file issues related to:
- changing `skaffold dev` default behaviour for port forward
- making it so that the Cloud Run Renderere/Deployer generate/copy yaml files to the `hydrationDir`
- support setting the target port via the Skaffold `portForward` schema and the gcloud proxy tool

bskaplan:
The third commit I added should support setting the target port via the portForward schema. It just doesn't support setting the address.

PR #7569, opened 2022-06-17
doc: remove image build from the render command decription

Shabirmean:
Fixes:
**Related**:  N/A
**Merge before/after**:  N/A

**Description**
- The public docs for the `skaffold render` command talks about *building images*
- However, the build step has been decoupled from the render command
- Thus, we must remove the information on `image builds` from the description to the `render` command


briandealwis (approved):
Thanks!

briandealwis (approved):

PR #7575, opened 2022-06-22
fix: verify now supports fully qualified image names, would error prior

### LINKED ISSUE #7570: skaffold verify uses image name as container name which causes issues using qualified images when running verify
error:

> creating container in local docker: Error response from daemon: Invalid container name ([us-east1-docker.pkg.dev/darien-sample-app/sample-repo/verify-hello-app](http://us-east1-docker.pkg.dev/aprindle-sample-app/sample-repo/verify-hello-app)), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed

aaron-prindle:
Fixes #7570


tejal29 (commented):
Can we add some tests. I am not sure why path.Base fixes this issue.

aaron-prindle:
The issue was that image names were being used as container names when called for images that skaffold did not build.  docker has a validation regexp on container names that image names do not meet (no "/" for example, etc.).  I have updated the existing integration tests to use a non-skaffold built image with a fully qualified path `docker.io/hello-world:latest` to trigger the current error which is now fixed w/ this PR

tejal29 (approved):

PR #7590, opened 2022-06-26
Replace deprecated ioutil pkg with os & io

abhinavnair:
**Description**
As of Go 1.16, the same functionality is now provided by package io or
package os, and those implementations should be preferred in new code.

So replacing all usage of `ioutil` pkg with `io` & `os`.


aaron-prindle (approved):
LGTM! Thanks!

PR #7599, opened 2022-06-28
test: cover the case when diag.Run encounters error

piotrostr:
**Description**
- Cover the remainder of the `diag.Run` method with an error handling test case

### DIFF
--- a/pkg/diag/diag_test.go
+++ b/pkg/diag/diag_test.go
@@ -45,7 +45,7 @@ func (m *mockValidator) Validate(_ context.Context, ns string, opts metav1.ListO
 }
 
 func (e *mockErrValidator) Validate(_ context.Context, ns string, opts metav1.ListOptions) ([]validator.Resource, error) {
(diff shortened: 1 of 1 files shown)


pkg/diag/diag_test.go
```diff
@@ -87,3 +96,33 @@ func TestRun(t *testing.T) {
...
+		expectedErr error
+	}{
+		{
+			description: "handles error",
+			shouldErr:   true,
+			labels: map[string]string{
+				"skaffold": "session",
+			},
+			ns:          []string{"foo"},
+			expectedErr: fmt.Errorf("following errors occurred someErr\n"),
```
aaron-prindle, line 115: nit: Seems our linter is mad w/ the error text:
```
Error: pkg/diag/diag_test.go:115:28: error strings should not be capitalized or end with punctuation or a newline (golint)
```

Fix this and things should be all green

piotrostr: thank you for the review, Aaron, should be fixed now!

Issue #7597, opened 2022-06-27
Remove Kustomize deployer


ericzzzzzzz:
### Discussion
 - Kustomize deployer is not used in v2, and we've moved Kustomize manifests logic into generate.go should we remove Kustomize deployer?

tejal29:
@ericzzzzzzz Can you verify if this is done already?

ericzzzzzzz:
@tejal29  just verified.. kustomize deployer has already been removed, we can close this out.

Issue #7595, opened 2022-06-27
Render doesn't respect namespace from cli 


ericzzzzzzz:
### Information
 - TestHelmRender currently using helm deployer.Render for testing render, we should change it to use helm renderer render method
 - after that change ``render with cli namespace`` and ``render with HelmRelease.Namespace and cli namespace`` test will fail
 - We need to inject cli namespace into helm Renderer struct to make Render method to use the namespace from cli for rendering if provided to make the helm render behavior aligned with v1 helm render

tejal29:
I was looking at this issue and trying to see what v1.37.2 behavior is for `skaffold render --namespace`

```
../../out/skaffold version
v1.37.2-113-g804abffed-dirty
```

CASE 1: If k8-pod.yaml has namespace set and `--namespace` is default, then render errors out.
```
➜  getting-started git:(render_cli) ✗ .../../skaffold1.39 render -d gcr.io/tejal-gke1 --namespace=test
kubectl create: running [kubectl --context gke_tejal-gke1_us-central1-c_dump-apis --namespace test create --dry-run=client -oyaml -f /Users/tejaldesai/workspace/skaffold/examples/getting-started/k8s-pod.yaml]
 - stdout: ""
 - stderr: "error: the namespace from the provided object \"default\" does not match the namespace \"test\". You must pass '--namespace=default' to perform this operation.\n"
 - cause: exit status 1
```
Similarly, skaffold dev or skaffold run will error out in the above scanrio.

CASE 2: If k8-pod.yaml has no namespace set `--namespace=test`, then the *rendered resource namespace is set to test**.
```
➜  getting-started git:(render_cli) ✗ ../../skaffold1.39 render -d gcr.io/tejal-gke1 --namespace=test
apiVersion: v1
kind: Pod
metadata:
  name: getting-started
  namespace: test
spec:
  containers:
  - image: gcr.io/tejal-gke1/skaffold-example:v1.37.2-114-g982a65188@sha256:4830d819b6f1ced78b64816608c3f1d9c5f1a0cceac51a8f6a2189fc97815f41
    name: getting-started

➜  getting-started git:(render_cli) ✗ 

```


My proposal here is to do the following,
1) render with the CLI namespace only if all resources namespaces are unset.

tejal29:
Manual Verification: 

on the PR branch 


1. Check out the branch and make.
 ```
git checkout render_cli 
Switched to branch 'render_cli'
LOCAL=true make
```
[...]

Issue #7567, opened 2022-06-16
--output flag not working as intended in V2


aaron-prindle:
--output flag not working as intended in V2, manifests always go to .kpt-pipeline/manifests.yaml regardless of specified location from --output flag

aaron-prindle:
Thanks @ericzzzzzzz, I am also seeing this working atm.  I had taken a note of this from a past friction log but it seems working now.  Closing the issue now, thanks!

PR #7603, opened 2022-06-28
chore: Remove render from deployer inteface

ericzzzzzzz:
Fixes: #7563  

**Description**
 - remove Render method from deployers implementations
 - refactor helm render unit tests to use Render method from helm renderer

 - #7595 
 - #7567 
 - #7597


pkg/skaffold/deploy/kustomize/kustomize_test.go
```diff
@@ -619,190 +618,6 @@ func TestKustomizeBuildCommandArgs(t *testing.T) {
 	}
 }
 
-func TestKustomizeRender(t *testing.T) {
```
tejal29, line 622: Are tests moved to some other place? We don't want to remove these tests

ericzzzzzzz: Related to kustomize deployer.. I see this https://github.com/ericzzzzzzz/skaffold/blob/main/pkg/skaffold/schema/v2beta28/upgrade.go#L60-L61 from schema upgrade .. if the comment is true should we remove kustomize deployer? However, I do see some breaking changes, like https://github.com/ericzzzzzzz/skaffold/blob/main/pkg/skaffold/schema/v2beta28/upgrade.go#L63-L65 also some fields in this struct
```go
// KustomizeDeploy *beta* uses the `kustomize` CLI to "patch" a deployment for a target environment.
type KustomizeDeploy struct {
	// KustomizePaths is the path to Kustomization files.
	// Defaults to `["."]`.
	KustomizePaths []string `yaml:"paths,omitempty" skaffold:"filepath"`

	// Flags are additional flags passed to `kubectl`.
	Flags KubectlFlags `yaml:"flags,omitempty"`

	// BuildArgs are additional args passed to `kustomize build`.
	BuildArgs []string `yaml:"buildArgs,omitempty"`

	// DefaultNamespace is the default namespace passed to kubectl on deployment if no other override is given.
	DefaultNamespace *string `yaml:"defaultNamespace,omitempty"`

	// LifecycleHooks describes a set of lifecycle hooks that are executed before and after every deploy.
	LifecycleHooks DeployHooks `yaml:"hooks,omitempty"`
}
nil the old kustomize deployer out seems breaking some deployer config and also LifecycleHooks. 
```

tejal29 (approved):

PR #7614, opened 2022-06-30
feat: add field imagePullSecrets to kaniko pod definition

yurchenkosv:
Fixes: #7615 
**Related**:  Relevant tracking issues, for context
Merge before/after: Dependent or prerequisite PRs

**Description**

I've been faced with no ability to build image via kaniko builder. This is because of build was started in private k8s cluster  which no access to internet and private docker proxy which needs requests to be authorized. Only way to get kaniko executor image is to set imagePullSecrets into pod definition. 
This PR is about to solve this case.

Changed yaml schema, to set imagePullSecret via config into pod.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -41,6 +41,7 @@ require (
 	github.com/docker/docker v20.10.14+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/dustin/go-humanize v1.0.0
+	github.com/fatih/semgroup v1.2.0
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-git/go-git/v5 v5.4.2
 	github.com/golang/glog v1.0.0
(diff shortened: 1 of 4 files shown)


tejal29:
@yurchenkosv Thanks for the fix. Looks like this is an issue in v1 skaffold versions. I am not sure how the PR you have fixes the issue. Can you please explain more?

yurchenkosv:
@tejal29 Thanks for the review. As I wrote earlier, there is a problem with the inability to build a docker image through kaniko when using a private docker proxy. Scaffold creates a pod in the cluster for the build, the pod uses 2 images (for the init container and for kaniko executor). The only way to get these images in my case is through a docker proxy, which requires authentication. The native k8s cluster mechanism assumes specifying imagePullSecret in the pod definition for such purposes. Unfortunately, scaffold does not provide the ability to pass imagePullSecret to pod via cluster config. My PR adds this ability.
If you have any other question about this PR, feel free to ask, I'm very interested in adding this code to the project.

tejal29:
@yurchenkosv friendly ping

tejal29:
@yurchenkosv friendly ping!

aaron-prindle:
@yurchenkosv friendly ping here.  I think you may need to rebase this PR as well now as it seems `go.sum,` has a conflict now

aaron-prindle:
Closing due to inactivity > 90 days.  Feel free to re-open or resubmit this change, thanks!

PR #7618, opened 2022-06-30
fix: handle project id correct for cloud run deploy

bskaplan:
**Description**
The Cloud Run Deployer is supposed to use the Project ID specified in the skaffold manifest if none is specified in the Service manifest, but it previously was failing to propagate the value from the skaffold manifest to the deployer.


aaron-prindle (approved):
LGTM

PR #7630, opened 2022-07-11
fix: handle skaffold label in Cloud Run revision template

bskaplan:
**Description**

Skaffold automatically adds a label, skaffold.dev/run-id to all k8s manifests it processes. Cloud Run tries to map its labels to GCP labels, but that label id is not a valid GCP label id so it needs to be renamed in order for the deploy to succeed.

This change was already being made to service.metadata.labels, but if service.spec.template.metadata is present, Skaffold will also add the label there so it's propagated to the Cloud Run Revision resource. This change renames the label in the second location if it's present.


aaron-prindle (approved):
LGTM!  Test failure is unrelated, merging

PR #7646, opened 2022-07-14
fix: add validation to verify container names to mirror k8s deployment validation

aaron-prindle:
Adds validation rule to enforce uniqueness across container[i].name entries as is done with the container[i].name entries in a k8s `Deployment`


MarlonGamez (approved):
lgtm!

Issue #7258, opened 2022-04-03
support ko in google cloud build


zmitry:
### Expected behavior
ko works in google cloud build

It seems that it's not that hard to add support ko on gcb, we need some default which we can use and some minor changes.

halvards:
Thanks for the input @zmitry! We are planning on enhancing the `ko` builder to add support for remote builds on Google Cloud Build.

Until then, you could take a look at the example `cloudbuild.yaml` for `ko` with Skaffold: https://github.com/GoogleContainerTools/skaffold/blob/main/examples/ko/cloudbuild.yaml#L41

Or feel free to submit a PR, if you'd like to see `gcb` support sooner 😄

zmitry:
hi @halvards, I have som mvp with ko that works, it seems that there are some questions that I don't know how to answer, if you could help me with that I might submit pr. For instance we need some default builder for ko, I managed to implement such with custom build of https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/ko but I'm not sure how to implement that for open source. 
Second option is where should I add configuration for koImage(similar to mavenImage and kanikoImage under builder configuration) should it go to new schema version v3beta1 or under v2beta28, can u help with that? 

Also default dependencies tracking for local ko is not quite right for cloud builder since it specifies **.go which doesn't work because we will need to publish go mod file as well.

halvards:
Thanks for looking into this @zmitry!

A new config field should be added to both `v1` and `v2` here: https://github.com/GoogleContainerTools/skaffold/tree/main/pkg/skaffold/schema/latest

`v2beta28` is already released, so a change here would need a new version (`v2beta29`).

Skaffold embeds `ko` as a library, so the default value of `koImage` should be `gcr.io/k8s-skaffold/skaffold`. You could also try this image with GCB: https://github.com/google/ko/pkgs/container/ko

As for dependency tracking, `go.mod` (and `go.sum`) is left out of the default configuration on purpose. Not all projects use `go.mod`, and Skaffold fails if a dependency pattern fails to match any files.

halvards:
This feature is now on the `main` branch and should be included in an upcoming release of Skaffold.

PR #7650, opened 2022-07-15
feat(ko): Google Cloud Build remote builds for the ko builder

halvards:
Enables remote builds on Google Cloud Build when using the Skaffold `ko` builder.

Since Skaffold embeds ko as a module, the Cloud Build step uses the `skaffold` binary to build and push the image. Since the `gcr.io/k8s-skaffold/skaffold` image contains the Go distribution, we can use this image on Cloud Build. This means that we don't need to create and maintain a separate `ko` builder image for Cloud Build.

Tracking: #7131
Related: #7258

### DIFF
--- a/pkg/skaffold/build/gcb/ko.go
+++ b/pkg/skaffold/build/gcb/ko.go
@@ -1,0 +1,143 @@
+/*
+Copyright 2022 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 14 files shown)


pkg/skaffold/build/gcb/ko.go
```diff
@@ -0,0 +1,143 @@
...
+				"\n",
+			)},
+			Env: skaffoldGCBEnv(env...),
+		}},
+	}
+}
+
+func skaffoldGCBEnv(env ...string) []string {
+	defaultEnv := []string{
+		"SKAFFOLD_DETECT_MINIKUBE=false",
```
tejal29, line 138: Nice!

tejal29 (approved):
LGTM!

Issue #7648, opened 2022-07-14
Skaffold delete is not working for kubectl deployer


ericzzzzzzz:
### Related issue
 - #7644

### Expected behavior
 - deployed resources should be cleaned up with ``skaffold delete`` command

### Actual behavior
 - deployed resources are not released after running ``skaffold delete`` command

### Steps to reproduce the behavior

1. checkout main branch, and run make to build binary. 
2. run ``skaffold run`` in ``example/getting_started`` folder,
3. run ``skaffold delete`` to delete resources. 
4. run ``kubectl get pod``, should see ``getting_started`` pods are still running.  However, the expected behavior is all resources in the run should be cleaned up. 

### information 
 - Kubeclt deployer works for v1 as its clean up method is able to read the manifests configured for this deployer by calling Dependencies() method, so cleanUp method is able to release resource defined in those manifests. 
 - In v2, manifests are not defined under deployer, and schema upgrade will nil out kubectl deployer for old version schema as well. Cleanup method is not calling Dependencies(). However, even we call Dependencies() method, it returns nothing, so it won't fix the problem. 
 - As team discussed, we've decided to take advantages of manifests config in skaffold.yaml and use the rendered manifests as source data for resource deletion in v2. The implementation will #7644 as well
 - The engineering work may be easier after #7572 merged into main.

Issue #7644, opened 2022-07-14
Skaffold dev does not clean up resource after run terminates for kubectl deployer


ericzzzzzzz:
### Expected behavior
 - deployed resources should be cleaned up after ``skaffold dev`` terminates 

### Actual behavior
 - deployed resources are not released after ``skaffold dev`` exits.

### Steps to reproduce the behavior

1. checkout main branch, and run make to build binary. 
2. run ``skaffold dev`` in ``example/getting_started`` folder,
3. after log streaming, exits with ``ctrl + c``
4. run ``kubectl get pod``, should see ``getting_started`` pods are still running.  However, the expected behavior is all resources in the run should be cleaned up. 

### information 
 - Kubeclt deployer works for v1 as its clean up method is able to read the manifests configured for this deployer by calling Dependencies() method, so cleanUp method is able to release resource defined in those manifests. 
 - In v2, manifests are not defined under deployer, and schema upgrade will nil out kubectl deployer for old version schema as well. Cleanup method is not calling Dependencies(). However, even we call Dependencies() method, it returns nothing, so it won't fix the problem. 
 - Instead, we could have a struct within kubectl deployer to maintain the deployed manifests, then we should be able to clean up deployment from that struct when running ``skaffold dev``.

ericzzzzzzz:
This will be fixed along with ``skaffold delete`` feature, we will be using manifests config for deployer cleanup methods.

PR #7657, opened 2022-07-19
fix: skaffold delete command and cleanup function

ericzzzzzzz:
Fixes: #7648 , #7644 

**Description**
 - fix ``skaffold delete`` command so users can use it to delete resources built from ``skaffold run``
 - fix ``deployer cleanup`` method, so resources can be cleaned up after dev command terminates
 - remove unnecessary code
 - add test util method for testing, deployments/pods deletions and deployer cleanup method.
 - add tests

 - fix ``skaffold delete`` command so users can use it to delete resources built from ``skaffold run``
 - fix ``deployer cleanup`` method, so resources can be cleaned up after dev command terminates

### DIFF
--- a/cmd/skaffold/app/cmd/delete.go
+++ b/cmd/skaffold/app/cmd/delete.go
@@ -24,6 +24,7 @@ import (
 	"github.com/spf13/cobra"
 
 	deployutil "github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy/util"
+	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/graph"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
 )
@@ -50,6 +51,11 @@ func doDelete(ctx context.Context, out io.Writer) error {
 		if err != nil {
 			return fmt.Errorf("getting manifests from hydration dir: %w", err)
 		}
-		return r.Cleanup(ctx, out, dryRun, manifests)
+
+		manifestListByConfig, err := r.Render(ctx, io.Discard, []graph.Artifact{}, false)
+		if err != nil {
+			return err
+		}
+		return r.Cleanup(ctx, out, dryRun, manifests, manifestListByConfig)
 	})
 }
--- a/cmd/skaffold/app/cmd/delete.go
+++ b/cmd/skaffold/app/cmd/delete.go
@@ -18,12 +18,10 @@ package cmd
 
 import (
 	"context"
-	"fmt"
 	"io"
 
 	"github.com/spf13/cobra"
 
-	deployutil "github.com/GoogleContainerTools/skaffold/pkg/skaffold/deploy/util"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/graph"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner"
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
@@ -47,15 +45,10 @@ func NewCmdDelete() *cobra.Command {
 
 func doDelete(ctx context.Context, out io.Writer) error {
 	return withRunner(ctx, out, func(r runner.Runner, configs []util.VersionedConfig) error {
-		manifests, err := deployutil.GetManifestsFromHydrationDir(ctx, opts)
-		if err != nil {
-			return fmt.Errorf("getting manifests from hydration dir: %w", err)
-		}
-
 		manifestListByConfig, err := r.Render(ctx, io.Discard, []graph.Artifact{}, false)
 		if err != nil {
 			return err
 		}
-		return r.Cleanup(ctx, out, dryRun, manifests, manifestListByConfig)
+		return r.Cleanup(ctx, out, dryRun, manifestListByConfig)
 	})
 }
--- a/integration/delete_test.go
+++ b/integration/delete_test.go
@@ -68,8 +68,6 @@ func TestDelete(t *testing.T) {
 			client.waitForDeploymentsToStabilizeWithTimeout(time.Minute*2, test.deployments...)
 
 			skaffold.Delete().InDir(test.dir).InNs(ns.Name).WithEnv(test.env).RunOrFail(t)
-			client.WaitForPodsDeletionWithTimeout(time.Minute*1, time.Second*5, test.pods...)
-			client.WaitForDeploymentDeletionWithTimeout(time.Minute*1, time.Second*5, test.deployments...)
 		})
 	}
 }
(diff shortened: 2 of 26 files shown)


pkg/skaffold/deploy/cloudrun/deploy.go
```diff
@@ -101,7 +101,7 @@ func (d *Deployer) Dependencies() ([]string, error) {
 }
 
 // Cleanup deletes the created Cloud Run services
-func (d *Deployer) Cleanup(ctx context.Context, out io.Writer, dryRun bool, manifests manifest.ManifestList) error {
+func (d *Deployer) Cleanup(ctx context.Context, out io.Writer, dryRun bool, manifests manifest.ManifestList, byConfig *manifest.ManifestListByConfig) error {
```
tejal29, line 104: Nice!

tejal29 (commented):
Things look good to me.

One quick question:

- Does the dev cycle clean up function used this new signature to pass the rendered manifests to delete? I was expecting to see some changes in runner/v2/dev.go

ericzzzzzzz:
> Things look good to me.
> 
> One quick question:
> 
> * Does the dev cycle clean up function used this new signature to pass the rendered manifests to delete? I was expecting to see some changes in runner/v2/dev.go

dev cycle calls SkaffoldRunner cleanup method with arguments context, io.Writer, offline, hydratedManifests (no change) , in that runner.cleanup method, we do the render work in this method then pass renderManifestsByConfig to delpoyers.cleanup() method, kind like the old way.. dev cycle doesn't pass deployer.dependencies to runner.cleanup() method. Yeah..after thinking a little bit, adding manifestsConfig arg to ``runner.cleanup`` signature and let ``delete command`` and ``dev cycle`` responsible for rendering before calling ``runner.cleanup`` makes more sense, that produces less surprise than hiding rendering part in ``runner.cleanup`` method. I'll update it.

pkg/skaffold/runner/runner.go
```diff
@@ -35,7 +35,7 @@ type Runner interface {
 	Apply(context.Context, io.Writer) error
 	ApplyDefaultRepo(tag string) (string, error)
 	Build(context.Context, io.Writer, []*latest.Artifact) ([]graph.Artifact, error)
-	Cleanup(context.Context, io.Writer, bool, manifest.ManifestList) error
+	Cleanup(context.Context, io.Writer, bool, manifest.ManifestList, *manifest.ManifestListByConfig) error
```
gsquared94, line 38: do you need both parameters `manifest.ManifestList` and `*manifest.ManifestListByConfig`?

cmd/skaffold/app/cmd/delete.go
```diff
@@ -50,6 +51,11 @@ func doDelete(ctx context.Context, out io.Writer) error {
 		if err != nil {
 			return fmt.Errorf("getting manifests from hydration dir: %w", err)
 		}
-		return r.Cleanup(ctx, out, dryRun, manifests)
+
+		manifestListByConfig, err := r.Render(ctx, io.Discard, []graph.Artifact{}, false)
```
gsquared94, line 55: don't you have to initialize the `graph.Artifact` correctly to get the correct render output?

integration/util.go
```diff
@@ -364,6 +366,77 @@ func (k *NSKubernetesClient) waitForDeploymentsToStabilizeWithTimeout(timeout ti
...
+
+func (k *NSKubernetesClient) WaitForPodsDeletionWithTimeout(timeout time.Duration, interval time.Duration, podNames ...string) {
+	k.waitForResourceDeletionWithTimeout(timeout, interval, "Pod", podNames...)
+}
+
+func (k *NSKubernetesClient) WaitForDeploymentDeletionWithTimeout(timeout time.Duration, interval time.Duration, deployments ...string) {
+	k.waitForResourceDeletionWithTimeout(timeout, interval, "Deployment", deployments...)
+}
+
+func (k *NSKubernetesClient) waitForResourceDeletionWithTimeout(timeout time.Duration, interval time.Duration, kind string, resources ...string) {
```
gsquared94, line 388: did you try something similar to [this](https://stackoverflow.com/a/71202736/4540147) using the event stream?

gsquared94 (changes requested):
Please check the review comments.

cmd/skaffold/app/cmd/delete.go
ericzzzzzzz: Do you mean passing the actually build outcome to this method? I might have missed something, but by my investigation we only need some identifier like kind, metadata.name for deleting kubernetes resources, we don't need the exact the same manifests used for deployment for deletion. We're doing similar thing in v1 as well [implementation](https://github.com/GoogleContainerTools/skaffold/blob/release/v1.37.2/pkg/skaffold/deploy/kubectl/kubectl.go#L493-L535). Without passing the build results, we are actually skipping override image tag, allowlist and something, that shouldn't change the deletion result, right?

ericzzzzzzz:
The new test is getting timeout.. may increate timeout limit or use another way for testing.

integration/delete_test.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Run(test.description, func(t *testing.T) {
+			ns, client := SetupNamespace(t)
+
+			args := append(test.args, "--cache-artifacts=false")
+			skaffold.Run(args...).InDir(test.dir).InNs(ns.Name).WithEnv(test.env).RunOrFail(t)
+
+			client.WaitForPodsReady(test.pods...)
+			client.waitForDeploymentsToStabilizeWithTimeout(time.Minute*2, test.deployments...)
+
+			skaffold.Delete().InDir(test.dir).InNs(ns.Name).WithEnv(test.env).RunOrFail(t)
```
tejal29, line 70: Lets convert this test to Smoke test instead. 
Check if there is not error in the command when skaffold.Delete is called. 

Remove verification of the whether is resources got deleted or not. 
These lines, test the functionality of kubectl delete.
Unit tests should cover if the right `kubectl delete` command was constructed by skaffold with correct cli arguments should suffice.

cmd/skaffold/app/cmd/delete.go
gsquared94: Got it. Then this is working the same as in `v1`.

pkg/skaffold/runner/runner.go
gsquared94: Can we try removing `manifest.ManifestList` as a parameter in this PR itself? Otherwise it just looks weird. I don't think that there should be any reference of `manifest.ManifestList` that can't use `manifest.ManifestListByConfig`

integration/delete_test.go
tejal29: another idea is to look at the kubectl events however, that too relies on cluster load. Events could get cleaned up and scheduled for deletion for k8s controller.

pkg/skaffold/runner/runner.go
ericzzzzzzz: done

integration/util.go
ericzzzzzzz: verification for deletion is not necessary, hence removed this part.

tejal29 (approved):

PR #7658, opened 2022-07-20
fix: Remove duplicate config from latest skaffold config

### LINKED ISSUE #7055: [v2] fix Skaffold v3 config duplication
While going through the skaffold v3alpha1 config, I found a bit of redundancy

1) `manifests`, `helmReleases` and `kustomizePaths` is duplicated in  render config as well respective deploy config i.e. `KubectlDeploy`, `HelmDeploy` and `KustomizeDeploy`.

2) Remote manifests config should be moved from `KubectlDeploy` to `manifests` config section.

tejal29:
relates to  #7055 
In this PR,  I have 
Removed duplicate config
- deploy.Kubectl.Manifest
- deploy.Kustomize.Paths

One main functional change this PR does is
- Addresses a long standing todo of cleaning up kubectl.RemoteManifests. During clean up, if a remote manifest is declared, skaffold queries the server to get the manifest. Removes it and then applies the remote manifest again without the image tag. 
The above change could be dangerous for there is no guarantee the remote manifest is created correctly when we call `kubectl apply` without the image tag.  Since there is no history why this was added, removing it made sense. If users see discrepancy, lets address this later

The `deploy.kubectl.RemoteManifests` stay in `kubectl` deployer config. The reason being, without specifying `kubectl` deployer, skaffold cannot deploy or update remote cluster manifests.


pkg/skaffold/schema/latest/config.go
```diff
@@ -721,18 +721,12 @@ type DockerDeploy struct {
 // KubectlDeploy *beta* uses a client side `kubectl apply` to deploy manifests.
```
aaron-prindle, line 721: nit: I believe that from the PR `deploy.Kustomize.Paths` should also be removed but this only touches the `deploy.Kubectl` configuration

aaron-prindle (approved):
LGTM, with one issue.  Currently this marks #7055 fixed but that issue also references removing deploy.Kustomize.paths which this PR does not.  I am not sure how you want to address this but it might make sense to either add thos changes here or not mark #7055 fixed until a PR is sent also removing that config

PR #7663, opened 2022-07-20
fix: Properly clean up and restart port forwarding for Cloud Run

bskaplan:
**Description**
Cloud Run port forwarding was previously adding a new port forward record every time it redeployed, so you'd end up with many port forwards for the same service as you updated your code with skaffold dev. Dedupe the port forwards for the same resource so we only start one.


aaron-prindle (approved):
LGTM

Issue #7600, opened 2022-06-28
Skaffold cannot find image tar with bazel symlinks disabled


arjantop-cai:
### Expected behavior
Skaffold uses the absolute path to the image tar

### Actual behavior
Skaffold fails to find the image because a relative path is used bu bazel is run with `build --symlink_prefix=/` which does not create the symlinks.
```
failed: reading image "bazel-out/darwin_arm64-fastbuild/bin/some_image.tar": open bazel-out/darwin_arm64-fastbuild/bin/some_image.tar: no such file or directory
```

### Information

- Skaffold version: v1.38.0
- Operating system: MacOS 12.3.1
- Installed via: Homebrew

### Steps to reproduce the behavior

1. Skaffold project with bazel image build
2. Add `build --symlink_prefix=/` to `.bazelrc`
3. Run skaffold

Issue #7281, opened 2022-04-09
Opening image tarball: open bazel-out/darwin../image.tar: no such file or directory


weivall:
Skaffold build throws an error after upgrading to v1.38.0.
 
### Information
Dir structure:
```
└── skaffold.yaml
    └── backend/
        |── WORKSPACE
       └── cmd/
           └── api/    
               └── BUILD

```

- Skaffold version: v1.38.0
- Operating system: Darwin MacBook-Pro-5.local 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:45:05 PDT 2022; root:xnu-8020.101.4~15/RELEASE_X86_64 x86_64
- Installed via: skaffold.dev
- Contents of skaffold.yaml:

```yaml
    build:
      artifacts:
        - image: api
          context: backend
          bazel:
            target: //cmd/api:image.tar

```

```

DEBU[0001] Running command: [bazel build //cmd/api:image.tar --color=yes]  subtask=api task=Build
Loading:
Loading: 0 packages loaded
Analyzing: target //cmd/api:image.tar (0 packages loaded, 0 targets configured)
INFO: Analyzed target //cmd/api:image.tar (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
[0 / 1] [Prepa] BazelWorkspaceStatusAction stable-status.txt
Target //cmd/api:image.tar up-to-date:
  bazel-out/darwin-fastbuild-ST-4a519fd6d3e4/bin/cmd/api/image.tar
INFO: Elapsed time: 0.376s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
DEBU[0001] Running command: [bazel cquery //cmd/api:image.tar --output starlark --starlark:expr target.files.to_list()[0].path]  subtask=api task=Build
DEBU[0001] Command output: [bazel-out/darwin-fastbuild-ST-4a519fd6d3e4/bin/cmd/api/image.tar
], stderr: Loading:
Loading: 0 packages loaded
Analyzing: target //cmd/api:image.tar (0 packages loaded, 0 targets configured)
INFO: Analyzed target //cmd/api:image.tar (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
INFO: Elapsed time: 0.262s
INFO: 0 processes.
INFO: Build completed successfully, 0 total actions
[...]

PR #7671, opened 2022-07-21
use absolute paths to the container_image.tar in bazel build.

reltuk:
Fixes: #7600
**Related**: #7251, #7281, #7430

**Description**

This PR changes the bazel build task to get an absolute path to the built container image. The correct path is the concatenation of the bazel `execution_root` and the result of the cquery for the output path of the configured target.

As per #7600, the current approach is broken if workspace symlinks are disabled. It's actually also broken if workspace symlinks are anything other than `bazel-`. The current approach is also broken if `skaffold.yaml` is not in the workspace root and `Artifact.Workspace` does not refer to the directory that contains the workspace symlinks. Previously, the `Artifact.Workspace` did not need to refer to the root of the workspace for a bazel build target, since the target can be an absolute path and it will build regardless of where bazel is invoked.

I believe using the absolute path is more robust and handles more use cases than the current approach.

**User facing changes**

* skaffold bazel builds and pushes with `--symlink_prefix` passed to bazel will work.
* skaffold bazel builds and pushes in directories that are not the root of the bazel workspace and where `context:` is not set to navigate to the root of the bazel workspace will work.

### DIFF
--- a/pkg/skaffold/build/bazel/build.go
+++ b/pkg/skaffold/build/bazel/build.go
@@ -85,7 +85,7 @@ func (b *Builder) buildTar(ctx context.Context, out io.Writer, workspace string,
 		return "", fmt.Errorf("getting bazel tar path: %w", err)
 	}
 
-	return filepath.Join(workspace, tarPath), nil
+	return tarPath, nil
 }
 
 func (b *Builder) loadImage(ctx context.Context, out io.Writer, tarPath string, a *latest.BazelArtifact, tag string) (string, error) {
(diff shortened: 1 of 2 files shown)


aaron-prindle:
Thanks for the PR @reltuk!  Seems there is currently a Skaffold unit test failing on Windows related to this change:
```
--- FAIL: TestBazelTarPath (0.03s)
    --- FAIL: TestBazelTarPath/EmptyExecutionRoot (0.02s)
        build_test.go:10[8](https://github.com/GoogleContainerTools/skaffold/runs/7527004738?check_suite_focus=true#step:6:9): string differ (-got, +want):   string(
            - 	`\absolute\path\bin`,
            + 	"/absolute/path/bin",
              )
    --- FAIL: TestBazelTarPath/AbsoluteExecutionRoot (0.02s)
        build_test.go:[12](https://github.com/GoogleContainerTools/skaffold/runs/7527004738?check_suite_focus=true#step:6:13)2: string differ (-got, +want):   strings.Join({
            - 	`\var\tmp\`,
            + 	"/var/tmp/",
              	"bazel-execution-roots",
            - 	`\abcdefg\execroot\workspace_name\bazel-bin\`,
            + 	"/abcdefg/execroot/workspace_name/bazel-bin/",
              	"darwin-fastbuild-ST-confighash",
            - 	`\path\to\`,
            + 	"/path/to/",
              	"bin",
              }, "")
```
once we fix that we should be able to merge!

reltuk:
@aaron-prindle Thanks for the feedback and for bringing the test failure on windows to my attention. I think filepath.Join is normalizing the path separator, but the mocked data in the test setup does not take that behavior into account. I suspect adding a normalizing call in the test setup will fix it. I’m currently traveling but will hopefully be able to fix this soon.

Thanks again!

reltuk:
@tejal29 Thank you so much for looking at this while I’m traveling! Much appreciated :)

Issue #7667, opened 2022-07-21
Skaffold V2 seems to hang indefinitely when deploying Pod K8s Objects (most of the time)


aaron-prindle:
When using the Skaffold V2 binary, it seems that status checking perhaps has a regression where when deploying Pods it can hang forever.  This is a regression from Skaffold V1.  Here are `/examples/*` applications where this occurs (attempt `skaffold dev` on one of them and it should hang waiting for deployments):
- examples/gcb-kaniko
- examples/gettting-started
- examples/jib-sync
- examples/getting-started-kaniko
- examples/simple-artifact-dependency
- examples//tagging-with-environmental-variables

When attempting to use `skaffold dev` on these exmaples (which deploy a `Pod`), the deploy will hang with:
```
Waiting for deployments to stabilize...
```

aaron-prindle:
Seems this r.resources is always 0 here which is incorrect which is causing this issue:
https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/kubernetes/status/resource/deployment.go#L145-L147
```
func (r *Resource) checkStandalonePodsStatus(ctx context.Context, cfg kubectl.Config) *proto.ActionableErr {
	if len(r.resources) == 0 {
		return &proto.ActionableErr{ErrCode: proto.StatusCode_STATUSCHECK_STANDALONE_PODS_PENDING}
	}
```

aaron-prindle:
Seems the root issue is that somehow in V2 the `r.namespaces` are empty for Pods when they should have a `default entry`.  This is causing this double for loop to do nothing:
https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/diag/diag.go#L75-L76
```
func (d *diag) Run(ctx context.Context) ([]validator.Resource, error) {
	var (
		res  []validator.Resource
		errs []error
	)
	// get selector from labels
	selector := labels.SelectorFromSet(d.labels)
	listOptions := metav1.ListOptions{
		LabelSelector: selector.String(),
	}

	for _, v := range d.validators {
		for _, ns := range d.namespaces {
                        # never getting here as d.namespaces is empty in V2 binary atm
			r, err := v.Validate(ctx, ns, listOptions)
```

tejal29:
I was able to figure out, why I never bumped into this issue. I had get `namesapce` in my active context. 
[...]

PR #7691, opened 2022-07-27
fix: Fix standalone pods and deployment health check pending.

tejal29:
fixes: #7667

After some walkthrough, i feel its safe to remove empty namespace check. I added few tests added and realizes, `ConsolidateNamespaces` will never add an empty namespace to the return value if no new namespace are found.

Here is the flow of how namespaces are updated in Deploy
```

func New() {

// first namespace is set to "". 
defaultNamespace := ""
// then preference to kubectl.CurrentContext.Namespace if set. 

// Here cfg.GetNamespace() is opt.Namespace which is set to "" by default. This method collects all 
// namespaces in the pipeline.
namespaces, err := deployutil.GetAllPodNamespaces(cfg.GetNamespace(), cfg.GetPipelines())
}

// If user has not set any namespace on CLI or in the skaffold.yaml config 
// or in kubectl.CurrentConfig, 
// namespaces = []string{""}


func (k Deployer) Deploy(...., manifests ManifestList, ...) {

// CollectNamespace never adds "" namesapce value to namespaces 
// See tests https://github.com/GoogleContainerTools/skaffold/blob/bcbdfe043c2f334f919fa2e6ae06aed4a7578486/pkg/skaffold/kubernetes/manifest/namespace_test.go#L131 
namespaces, err := manifests.CollectNamespaces()

k.trackNamespaces(namespaces)
}


func (k *Deployer) trackNamespaces(namespaces []string) {

 // Consolidate namespaces will delete "" empty namespace if namespaces is non empty.  
// So the only case when empty
// namespace value "" is set to deployer namespace is when 
// *k.namespace already has the value in New and rendered manifest don't have any new manifests.
   *k.namespaces = deployutil.ConsolidateNamespaces(*k.namespaces, namespaces)
}


```

### DIFF
--- a/pkg/diag/diag.go
+++ b/pkg/diag/diag.go
@@ -39,12 +39,8 @@ type diag struct {
 }
 
 func New(namespaces []string) Diagnose {
-	var ns []string
-	for _, n := range namespaces {
-		ns = append(ns, n)
-	}
 	return &diag{
-		namespaces: ns,
+		namespaces: namespaces,
 		labels:     map[string]string{},
 	}
 }
--- a/pkg/diag/diag.go
+++ b/pkg/diag/diag.go
@@ -41,9 +41,7 @@ type diag struct {
 func New(namespaces []string) Diagnose {
 	var ns []string
 	for _, n := range namespaces {
-		if n != "" {
-			ns = append(ns, n)
-		}
+		ns = append(ns, n)
 	}
 	return &diag{
 		namespaces: ns,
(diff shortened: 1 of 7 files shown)


aaron-prindle (approved):
LGTM!

aaron-prindle:
nit: perhaps as a followup issues/PR we can add integration tests that can catch these cases where our standalone health checks are pending indefinitely?

Issue #6952, opened 2021-12-09
Migrate Helm users using artifactOverrides


briandealwis:
Tracking bug: #6706 

#6949 removes the `artifactOverrides` field and instead configured Helm to use Skaffold as a post-renderer filter for hydrating the Helm-rendered manifests.  One requirement for this to work is that the images in the rendered manifests must correspond to either the `build.artifacts[].image` or to the final locations of the images (post-default-repo).  This may be a breaking change as some users may have been using image defaults that don't correspond to the image names in the `build.artifacts[].image` and so their deployments may cease to work.

We should convert uses of `artifactOverrides` into `setValues` where we populate the keys with the corresponding `artifact[].image` using the configured Helm strategy.

---

An alternative strategy would be to check for this situation in the schema-upgrade.  We could invoke `helm template` and then look for matched and unmatched image names.  If we found mismatches, Skaffold could warn the user or abort the upgrade and direct them to do a manual upgrade.  Warning seems inadequate as a missed image can result in a broken deployment (either referencing an image that does not exist or that is not the built image), and users may miss the warning.  Aborting and requiring an update would entail one-off toil for users, since they would need to examine their charts and `skaffold.yaml` to see if there is a discrepancy.  Since some users won't rewrite their `skaffold.yaml` — since we silently upgrade from previous schema versions — our upgrade process would also invoke `helm template` each time, which is not ideal.

PR #7707, opened 2022-08-01
chore: support `artifactOverrides` schema functionality from v2beta* in v3alpha* via `setValue*`

### LINKED ISSUE #7668: Skaffold V2 Helm Templated Fields Broken/Regression -- V2 breaks on V1 schemas using `artifactOverrides`
When attempting to run `skaffold dev` against the following example using the Skaffold V2 binary:
- examples/templated-fields

The run fails, erroring with
```
aprindle@aprindle ~/skaffold/examples/templated-fields  [main]$ skaffold dev --default-repo=gcr.io/aprindle-test-cluster
Listing files to watch...
 - skaffold-templated
Generating tags...
 - skaffold-templated -> gcr.io/aprindle-test-cluster/skaffold-templated:v1.38.0-143-gf56c65383
Checking cache...
 - skaffold-templated: Found. Tagging
Tags used in deployment:
 - skaffold-templated -> gcr.io/aprindle-test-cluster/skaffold-templated:v1.38.0-143-gf56c65383@sha256:155da19316252d528072101ee1521ad6056918f4a342763f2954d22688540d47
Starting deploy...
Helm release skaffold-templated not installed. Installing...
Error: Deployment.apps "skaffold-helm" is invalid: spec.template.spec.containers[0].image: Required value
Cleaning up...
release "skaffold-templated" uninstalled
deploying "skaffold-templated": install: exit status 1

```

When using the Skaffold V1 binary the runs succeeds.

aaron-prindle:
Fixes #7668 

Fixes https://github.com/GoogleContainerTools/skaffold/issues/6952
Uses the suggestion recommended by @briandealwis in the issue above to migrate `artifactOverrides` configuration to similar `setValues` and `setValueTemplates` configurations.  

Prior to this PR, users using `artifactOverrides` with v2beta* would have their projects broken as `artifactOverrides` config was dropped going v2beta* -> v3alpha1.  This PR parses the v2beta* `artifactsOverrides` and converts it to appropriate `setValues` and `setValueTemplates` config which v3alpha1 supports.  There still might be some possible errors with the currently implementation related to the error mentioned in the issue:
```
This may be a breaking change as some users may have been using image defaults that don't correspond to the image names in the build.artifacts[].image and so their deployments may cease to work.
```
```
 We could invoke helm template and then look for matched and unmatched image names. If we found mismatches, Skaffold could warn[or fail with error] the user or abort the upgrade and direct them to do a manual upgrade.
```
I will make a separate issue tracking the above case which still needs to be accounted for but this way the majority of `artifactOverrides` users will not be broken in the migration and won't have to manually update their `skaffold.yaml`

EDIT: Related to above ^^^^ statement.  Talked offline w/ Brian, the error referenced above would be an ongoing error and NOT a regression.  He is happy with the approach here and the above issue is not high priority (no regression)

NOTE: The changes to examples/templated-fields in this PR were lost when rebasing the v2 branch onto main from https://github.com/GoogleContainerTools/skaffold/issues/6952 and that is why they are added back here

### DIFF
--- a/examples/templated-fields/charts/values.yaml
+++ b/examples/templated-fields/charts/values.yaml
@@ -2,3 +2,4 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 replicaCount: 2
+image: skaffold-templated
--- a/examples/templated-fields/skaffold.yaml
+++ b/examples/templated-fields/skaffold.yaml
@@ -16,8 +16,6 @@ deploy:
     releases:
     - name: skaffold-templated
       chartPath: charts
-      artifactOverrides:
-        image: skaffold-templated
       setValueTemplates:
         imageRepo: "{{.IMAGE_REPO}}"
         imageTag: "{{.IMAGE_TAG}}"
--- a/pkg/skaffold/helm/args.go
+++ b/pkg/skaffold/helm/args.go
@@ -130,5 +130,9 @@ func envVarForImage(imageName string, digest string) map[string]string {
 	} else {
 		customMap["DIGEST_HEX"] = digest
 	}
+
+	// IMAGE_DOMAIN and IMAGE_REPO_NO_DOMAIN added for v2beta* 'helm+explicitRegistry' -> v3alpha* compatibility
+	customMap["IMAGE_DOMAIN"] = ref.Domain
+	customMap["IMAGE_REPO_NO_DOMAIN"] = strings.TrimPrefix(ref.BaseName, ref.Domain+"/")
 	return customMap
 }
--- a/pkg/skaffold/schema/v2beta29/upgrade.go
+++ b/pkg/skaffold/schema/v2beta29/upgrade.go
@@ -87,6 +87,43 @@ func upgradeOnePipeline(oldPipeline, newPipeline interface{}) error {
 		newPL.Render.Helm.Releases = newHelm.Releases
 		newPL.Render.Helm.Flags = newHelm.Flags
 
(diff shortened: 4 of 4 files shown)


tejal29:
Actually, we are getting rid of `artifactOverrides` in v3 schema. They were confusing and we don't need these with the new `helm template` post-renderer.
skaffold fix should emit some warning and we can point to doc https://docs.google.com/document/d/1g54UNCK-FYcVz2ZTmx4i3iNUjvWapnnzrtZ4yW8daE8/edit?resourcekey=0-SmqP33WG9HwYUerlr9qJ0A#heading=h.2nx6vufc97a3

aaron-prindle:
Talked offline, this issue supports the v3 schema removal of `artifactOverrides` by properly updating v2 schema `artifactOverrides` users to the correct `setValues|setValueTemplates` configuration for the v3 schema to do what their `artifactOverrides` config was doing prior

tejal29 (approved):

iosifnicolae2:
`setValueTemplates` does not work correctly when building multiple skaffold configs - https://github.com/GoogleContainerTools/skaffold/issues/8008

aaron-prindle:
@iosifnicolae2 can you paste a redacted `skaffold.yaml` showing the config that does not work?  I would love to repro this locally and start working on a fix, currently I am not sure what is missing in our current upgrade.go logic

iosifnicolae2:
@aaron-prindle I've added a GitHub repo and how to reproduce the issue in #8008

Thank you!

Issue #7047, opened 2022-01-24
Helm flags not working via skaffold


yurchenkosv:
### Expected behavior
skaffold execute helm upgrade with flags 
### Actual behavior
Error: unknown flag: --history-max 3
deploying "release_name": install: exit status 1
### Information

- Skaffold version:  v1.35.2
- Operating system:  MacOS 11.6.2
- Installed via:  Homebrew
- Contents of skaffold.yaml:

```yaml
Cannot share 
```

### Steps to reproduce the behavior

Add flags to upgrade stage of helm
```
        flags:
          upgrade:
            - '--history-max 3'
```
execute
`skaffold deploy --build-artifacts images.json`

### Logs
```
Running command: [helm --kube-context minikube upgrade release_name --history-max 3 /Users/<username>/Desktop/projects/<project_name>/build/helm/<chart_name> --set-string image.repository=repo,image.tag=119914028473db0e6c82c0862a399477b9bfb6a9b767b256b28bccffd5d48a22 -f /Users/<username>/Desktop/projects/<project_name>/build/helm/<chart_name>/values.local.yaml -f /Users/<username>/Desktop/projects/<project_name>/build/helm/<chart_name>/values.pwd.local.yaml]  subtask=0 task=Deploy
Error: unknown flag: --history-max 3
```

###  Additional info
Executing Helm command via console - got no errors.
Please, help to understand skaffold behavior and fix this case )

yurchenkosv:
I found out that a space in the flag leads to an error. 
With the following I got no errors and all works as expected
```
        flags:
          upgrade:
            - '--history-max=3'
```
I think, this could be added to documentation?

tejal29:
Thanks @yurchenkosv for finding this out. We will update the docs to reflect this.

JohnnyD45:
Hi Folks, I'll work on this and submit a PR later.

nayang7:
> Hi Folks, I'll work on this and submit a PR later.

Hey JohnnyD45, seems there's quite a long time, how about I work on this and submit a PR later.

JohnnyD45:
Sure @nayang7 , please go ahead.
And apologies folks for not following up on this task.

aaron-prindle:
Closing, the flag pattern used is common across many k8s/yaml specs.  Re-open if perhaps this warrants more inspection

PR #7725, opened 2022-08-03
Fix Helm flags not working via skaffold #7047

nayang7:
Fixes: #7047  

**Description**


nayang7:
will have a new PR

PR #7728, opened 2022-08-03
feat(ko): Inferred file sync for ko static assets

halvards:
This change adds inferred file sync mode to the ko builder.

Skaffold syncs static assets from the workspace to a running container, based on the `<importpath>/kodata/` convention: https://github.com/google/ko#static-assets

Tracking: #7131

### DIFF
--- a/docs/content/en/docs/pipeline-stages/builders/ko.md
+++ b/docs/content/en/docs/pipeline-stages/builders/ko.md
@@ -389,7 +389,46 @@ To learn more about how Skaffold debugs Go applications, read the
 
 ### File sync
 
-File `sync` is not supported while the ko builder feature is in Alpha.
+The `ko` builder can
+[sync files to a running container]({{< relref "/docs/pipeline-stages/filesync" >}})
+when you run `skaffold dev`.
+
+The sync feature for the `ko` builder only works for 
+[static assets bundled with the container image](https://github.com/google/ko#static-assets).
+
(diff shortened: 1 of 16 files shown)


docs/content/en/docs/pipeline-stages/builders/ko.md
```diff
@@ -389,7 +389,46 @@ To learn more about how Skaffold debugs Go applications, read the
 
 ### File sync
 
-File `sync` is not supported while the ko builder feature is in Alpha.
+The `ko` builder can
```
tejal29, line 392: Since this feature will only be available in V2, would it make sense to revert all doc changes here?

tejal29 (commented):
First pass review. Will look through the code in detail later this week.

docs/content/en/docs/pipeline-stages/builders/ko.md
halvards: We're not going to create any more v1.x releases from the `main` branch? I was thinking that this was perhaps the reason for the `docs` vs `docs-v2` split. If not, I'll revert the changes in `docs` 👍

tejal29: right now no. We need to clean up the docs and docs-v2 split. Depends on what @aaron-prindle has in mind. 
We do want the v1 and v2 docs to reside side by side so doc fixes could be made for users still using v1

tejal29 (approved):

PR #7750, opened 2022-08-05
fix: remove typo in manifests error message

aaron-prindle:


pkg/skaffold/kubernetes/manifest/errors.go
```diff
@@ -75,7 +75,7 @@ func nsSettingErr(err error) error {
 }
 
 func nsAlreadySetErr() error {
-	err := fmt.Errorf("namespace already set in the namesapce")
+	err := fmt.Errorf("namespace already set in the namespace")
```
gsquared94, line 78: should this be rephrased to something like the error code?
```
	err := fmt.Errorf("namespace field already set in the manifests")
```

aaron-prindle: Good idea, updated the text to be as you suggest

gsquared94 (approved):
lgtm

PR #7768, opened 2022-08-10
fix: render should be online for `dev`, `deploy` and `run`.

gsquared94:
Render should be `online` by default for `dev`, `deploy` and `run`. 
It seems to only be configurable for the `render` command with the `--offline` flag, [and even that is `false` by default](https://github.com/GoogleContainerTools/skaffold/blob/c3fe38d25d0ab68aadf56b1a905eb8477b48dcb9/cmd/skaffold/app/cmd/render.go#L45). This is also the behavior on the [`v1` branch](https://github.com/GoogleContainerTools/skaffold/blob/6f1ffc3b59ff9748f7f8891d51628e894c92d385/pkg/skaffold/deploy/kubectl/kubectl.go#L213). So this is a regression in `v2`

### DIFF
--- a/cmd/skaffold/app/cmd/deploy.go
+++ b/cmd/skaffold/app/cmd/deploy.go
@@ -52,7 +52,7 @@ func doDeploy(ctx context.Context, out io.Writer) error {
 			return err
 		}
 		// Render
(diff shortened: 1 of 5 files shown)


aaron-prindle (approved):
LGTM!

PR #7784, opened 2022-08-17
fix: remove required tag from cloud run region

bskaplan:
This can now be specified via flag so it's no longer required in the YAML.


aaron-prindle:
What happens here if `region` is not specified and the flag is not specified?  Will the Cloud Run deployer draw a default value from somewhere or fail with a message stating region needs to be specified?   We might need to add a validation rule and an actionable error here

bskaplan:
You're right, we should have validation on this. Is there somewhere in particular the validation should go? If not, I'll add it to cloudrun.NewDeployer.

aaron-prindle:
validations typically go here:
https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/schema/validation/validation.go

aaron-prindle (approved):
LGTM

PR #7785, opened 2022-08-17
feat: add hooks support to render schema and phase

### LINKED ISSUE #7749: [v2][REGRESSION] Need ability to do hooks in manifest section prior to rendering
### Expected behavior
* Able to perform actions pre-post render

Expected YAML allowed:

```yaml
apiVersion: skaffold/v3alpha1
kind: Config
metadata:
  name: skaffold-myapp-demo
profiles:
  - name: standalone-dev
    manifests:
      kustomize:
        paths:
          - kustomize/overlays/dev
        buildArgs: ["--enable-alpha-plugins", "--enable-exec"]
        hooks:
          before:
           - host:
               command: ["sh", "-c", "(/workspace/source/kustomize/overlays/$DEPLOYMENT_ENVIRONMENT; /usr/bin/kustomize-inject-env)"]
    deploy:
      kubectl:
        flags:
          disableValidation: false
        #defaultNamespace: mynamespace-dev
```

### Actual behavior

* Functionality regression from v1.x, not able to perform actions prior to running kustomize

### Information

- Skaffold version: v2.0.0-beta1
- Operating system: Linux
- Installed via: skaffold.dev
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v3alpha1
kind: Config
metadata:
  name: skaffold-myapp-demo
profiles:
  - name: standalone-dev
    manifests:
      kustomize:
        paths:
          - kustomize/overlays/dev
        buildArgs: ["--enable-alpha-plugins", "--enable-exec"]
    deploy:
      kubectl:
        flags:
          disableValidation: false
        #defaultNamespace: mynamespace-dev
        hooks:
          before:
           - host:
               command: ["sh", "-c", "(/workspace/source/kustomize/overlays/$DEPLOYMENT_ENVIRONMENT; /usr/bin/kustomize-inject-env)"]
```

```
$ SKAFFOLD_MODULE=skaffold-myapp-demo SKAFFOLD_PROFILE=standalone-dev skaffold2 render -f skaffold/dev/skaffold.yaml -v info
INFO[0000] Skaffold &{Version:v2.0.0-beta1 ConfigVersion:skaffold/v3alpha1 GitVersion: GitCommit:06ad0bd3d7bff1f53c6cd201f3f67b73435c95e8 BuildDate:2022-08-03T21:07:27Z GoVersion:go1.17 Compiler:gc Platform:linux/amd64 User:}  subtask=-1 task=DevLoop
INFO[0000] applying profile: standalone-dev              subtask=-1 task=DevLoop
INFO[0000] no values found in profile for field TagPolicy, using original config values  subtask=-1 task=DevLoop
INFO[0000] no values found in profile for field BuildType, using original config values  subtask=-1 task=DevLoop
INFO[0000] Using kubectl context: XXXXXXXX  subtask=-1 task=DevLoop
INFO[0000] Loaded Skaffold defaults from "/home/spstarr/.skaffold/config"  subtask=-1 task=DevLoop
INFO[0000] hydrated-dir does not exist, creating /build/myapp-demo/.kpt-pipeline  subtask=-1 task=DevLoop
INFO[0000] manifests hydration will take place in /build/myapp-demo/.kpt-pipeline  subtask=-1 task=DevLoop
INFO[0000] build concurrency set to default value of -1  subtask=-1 task=DevLoop
INFO[0000] Tags generated in 2.217µs                     subtask=-1 task=Build
INFO[0000] Starting render...                            subtask=-1 task=DevLoop
```

Invalid rendering, should be kustomize?

```
INFO[0000] rendering using kubectl                       subtask=0 task=Render
```

Seems like a typo :( and this error seems to be invalid, i dont see any namespace fields defined.. :confused: 

```
namespace already set in the namesapce. remove/unset 'namespace' field in manifests and try again.
```

aaron-prindle:
Fixes #7749 

Adds support for `render` phase Lifecycle hooks at 
`manifests.hooks`

NOTE: these hooks are pre/post the render phase and not per renderer

### DIFF
--- a/pkg/skaffold/schema/latest/config.go
+++ b/pkg/skaffold/schema/latest/config.go
@@ -601,6 +601,9 @@ type Generate struct {
 
 	// Kpt defines the kpt resources in the application.
 	Kpt []string `yaml:"kpt,omitempty" skaffold:"filepath"`
+
+	// LifecycleHooks describes a set of lifecycle hooks that are executed before and after every render.
+	LifecycleHooks RenderHooks `yaml:"hooks,omitempty"`
 }
 
 // Kustomize defines the paths to be modified with kustomize, along with
@@ -1520,6 +1523,20 @@ type SyncHooks struct {
 	PreHooks []SyncHookItem `yaml:"before,omitempty"`
 	// PostHooks describes the list of lifecycle hooks to execute *after* each artifact sync step.
 	PostHooks []SyncHookItem `yaml:"after,omitempty"`
+}
+
+// RenderHookItem describes a single lifecycle hook to execute before or after each deployer step.
+type RenderHookItem struct {
+	// HostHook describes a single lifecycle hook to run on the host machine.
+	HostHook *HostHook `yaml:"host,omitempty" yamltags:"oneOf=render_hook"`
+}
+
+// RenderHooks describes the list of lifecycle hooks to execute before and after each render step.
+type RenderHooks struct {
+	// PreHooks describes the list of lifecycle hooks to execute *before* each render step. Container hooks will only run if the container exists from a previous deployment step (for instance the successive iterations of a dev-loop during `skaffold dev`).
+	PreHooks []RenderHookItem `yaml:"before,omitempty"`
+	// PostHooks describes the list of lifecycle hooks to execute *after* each render step.
+	PostHooks []RenderHookItem `yaml:"after,omitempty"`
 }
 
 // DeployHookItem describes a single lifecycle hook to execute before or after each deployer step.
--- a/docs-v2/content/en/schemas/v3alpha1.json
+++ b/docs-v2/content/en/schemas/v3alpha1.json
@@ -3364,6 +3364,11 @@
           "description": "defines the helm charts used in the application. NOTE: Defines cherts in this section to render via helm but deployed via kubectl or kpt deployer. To use helm to deploy, please see deploy.helm section.",
           "x-intellij-html-description": "defines the helm charts used in the application. NOTE: Defines cherts in this section to render via helm but deployed via kubectl or kpt deployer. To use helm to deploy, please see deploy.helm section."
         },
+        "hooks": {
+          "$ref": "#/definitions/RenderHooks",
+          "description": "describes a set of lifecycle hooks that are executed before and after every render.",
+          "x-intellij-html-description": "describes a set of lifecycle hooks that are executed before and after every render."
+        },
         "kpt": {
           "items": {
             "type": "string"
@@ -3406,6 +3411,7 @@
         "kustomize",
         "helm",
         "kpt",
+        "hooks",
         "transform",
         "validate",
         "output"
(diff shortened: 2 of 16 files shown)


pkg/skaffold/schema/latest/config.go
```diff
@@ -612,6 +612,9 @@ type Kustomize struct {
 
 	// BuildArgs are additional args passed to `kustomize build`.
 	BuildArgs []string `yaml:"buildArgs,omitempty"`
+
+	// LifecycleHooks describes a set of lifecycle hooks that are executed before and after every render.
+	LifecycleHooks RenderHooks `yaml:"hooks,omitempty"`
```
tejal29, line 617: For an application with k8 manifests only, there is no way to specify the pre-render hooks. Is there a specific reason for it? 
Would it make sense to move it top level to like `Transform` and `Validate` in `RenderConfig`. 
Challenge would be to find a  good config name for it.

aaron-prindle: I agree that this is currently a gap in the code that is here.  My plan was to add a new schema field `rawk8sHooks` and use that as the hooks that run for `rawK8s`.  I also thought of making the hooks be a top level concept but I opted to not to this initially as it would not follow the same design as `deploy`

tejal29: makes sense.

pkg/skaffold/schema/latest/config.go
```diff
@@ -589,6 +589,9 @@ type Generate struct {
 	// RawK8s defines the raw kubernetes resources.
 	RawK8s []string `yaml:"rawYaml,omitempty" skaffold:"filepath"`
 
+	// RawK8sLifeCycleHooks describes a set of lifecycle hooks that are executed before and after every render.
+	RawK8sLifeCycleHooks RenderHooks `yaml:"rawYamlHooks,omitempty"`
```
gsquared94, line 593: maybe we can simplify this by only having a single `RenderHooks` inside `Generate` that gets handled in the `render_mux.go` file:
```diff
- type GroupRenderer []Renderer
+ type GroupRenderer struct {
+  renderers []Renderer
+ hooks RenderHooks
+}
``` 

WDYT?

aaron-prindle: I can make these changes but initially the idea was the keep the hooks per renderer philosophy similar to the hooks per deployer design vs hooks per render phase.  I will make this change and push them but it was unclear if this was acceptable or why similarly deploy doesn't have a top level hooks vs deployer level hooks

aaron-prindle: I've updated the code here to have a render phase set of hooks at `manifests.hooks` vs per renderer hooks

aaron-prindle: Have made the suggested change, updated the PR info to state:

Adds support for `render` phase Lifecycle hooks at 
`manifests.hooks`

NOTE: these hooks are pre/post the render phase and not per renderer

pkg/skaffold/schema/latest/config.go
```diff
@@ -1522,6 +1525,20 @@ type SyncHooks struct {
...
+	HostHook *HostHook `yaml:"host,omitempty" yamltags:"oneOf=render_hook"`
+}
+
+// RenderHooks describes the list of lifecycle hooks to execute before and after each render step.
+type RenderHooks struct {
+	// PreHooks describes the list of lifecycle hooks to execute *before* each render step. Container hooks will only run if the container exists from a previous deployment step (for instance the successive iterations of a dev-loop during `skaffold dev`).
+	PreHooks []RenderHookItem `yaml:"before,omitempty"`
+	// PostHooks describes the list of lifecycle hooks to execute *after* each render step.
+	PostHooks []RenderHookItem `yaml:"after,omitempty"`
+}
```
gsquared94, line 1540: you don't need `RenderHookItem`. You can just define:
```
type RenderHooks struct {
	// PreHooks describes the list of lifecycle hooks to execute *before* each render step. Container hooks will only run if the container exists from a previous deployment step (for instance the successive iterations of a dev-loop during `skaffold dev`).
	PreHooks []HostHook `yaml:"before,omitempty"`
	// PostHooks describes the list of lifecycle hooks to execute *after* each render step.
	PostHooks []HostHook `yaml:"after,omitempty"`
}
```
since `render` like `build` doesn't need `ContainerHook`s. See: https://github.com/GoogleContainerTools/skaffold/blob/0164731ae119370aedd71501eb5cc1dfa1ccb3fa/pkg/skaffold/schema/latest/config.go#L1502-L1507

gsquared94 (approved):
We can simplify the new config properties a bit, otherwise lgtm.

tejal29:
should this need a follow up PR for `skaffold fix` to move deploy hook to render or a message suggesting it?

pkg/skaffold/schema/latest/config.go
aaron-prindle: After talking w/ Cloud Deploy team, they mentioned they will likely need to add support for another Hook type that runs in a container (vs on the Host) at a later date.  As such I think it makes sense to leave this indirection for now as we will be adding more hook types for render in the future.

aaron-prindle:
Added #7803 to track updating/investigating if `skaffold fix` should translate pre-deploy hooks to pre/post render hooks or at least warn users of the ambiguity

Issue #4295, opened 2020-06-06
Make IMAGE_REPO, IMAGE_NAME, IMAGE_TAG available in build.artifacts[i].docker.buildArgs


javier-b-perez:
### Expected behavior

```yaml
apiVersion: skaffold/v2beta1
kind: Config
build:
  local:
    push: false
  artifacts:
  - image: gobuilder
    context: build/gobuilder
    docker:
      dockerfile: Dockerfile
  # These containers are using builders (like golang + tools) + src.
  - image: my-app
    docker:
      dockerfile: my-app/Dockerfile
      buildArgs:
        GOBUILDER: "{{.IMAGE_REPO}}/gobuilder:{{.IMAGE_TAG}}"
```

```shell
skaffold build --default-repo=gcr.io/my-project -t demo
```

### Actual behavior

```shell
- my-app: Error checking cache.
getting hash for artifact "my-app": getting dependencies for "my-app": parsing ONBUILD instructions: retrieving image "<no value>/gobuilder:<no value>": parsing reference [<no value>/gobuilder:<no value>]: could not parse reference: <no value>/gobuilder:<no value>
```

### Information

- Skaffold version: v1.10.1

### Steps to reproduce the behavior

1. `git clone https://github.com/javier-b-perez/skaffold-tags-demo.git`
2. `skaffold build --default-repo=gcr.io/my-project -t demo`

### Workaround

1. Update skaffold.yaml
```yaml
apiVersion: skaffold/v2beta1
kind: Config
build:
  local:
    push: false
  artifacts:
  - image: basebuilder
    docker:
      dockerfile: Dockerfile.base
  # These containers are using builders (like golang + tools) + src.
  - image: my-app
    docker:
      dockerfile: Dockerfile
      buildArgs:
        BASEBUILDER: "gcr.io/my-project/basebuilder:demo"
```

2. Build builder
```shell
skaffold build --default-repo=gcr.io/my-project -t demo -b basebuilder
```

3. Build app
```shell
skaffold build --default-repo=gcr.io/my-project
```

tejal29:
@javier-b-perez Unfortunately the pr #3974 linked in #3229 only added the `IMAGE_REPO`, `IMAGE_NAME` and `IMAGE_TAG` to kaniko builder.

[...]

Issue #7774, opened 2022-08-15
panic: runtime error: index out of range [0] with length 0


sjoerd222888:
### Expected behavior

Skaffold is able to build the container image locally, push it and deploy it 

### Actual behavior

Skaffold crashes:

```
% skaffold run -p dev
Generating tags...
- gcr.io/███████ -> gcr.io/███████:███████
Checking cache...
- gcr.io/███████: Found. Tagging
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/cache.needsRemoteTagging.Tag({{0x140013ca900, 0x40}, {0x14001181090, 0x46}, {0x140011800f0, 0x47}}, {0x105a2fd18, 0x14000f3d7a0}, 0x14001079780, {0x140010b6e40})
      /private/tmp/skaffold-20220624-80169-j89lh6/pkg/skaffold/build/cache/details.go:94 +0x218
github.com/GoogleContainerTools/skaffold/pkg/skaffold/build/cache.(*cache).Build(0x14001079780, {0x105a2fc70, 0x140001b2340}, {0x1059de580, 0x14000d04960}, 0x14000de5f80, {0x14000e1bc48, 0x1, 0x1}, {0x140010b6e40}, ...)
      /private/tmp/skaffold-20220624-80169-j89lh6/pkg/skaffold/build/cache/retrieve.go:85 +0xd04
github.com/GoogleContainerTools/skaffold/pkg/skaffold/runner.(*Builder).Build(0x140005d7380, {0x105a2fd18, 0x14000de5ec0}, {0x1059de580, 0x14000d04960}, {0x14000e1bc48, 0x1, 0x1})
      /private/tmp/skaffold-20220624-80169-j89lh6/pkg/skaffold/runner/build.go:105 +0x8fc
github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/cmd.doRun.func1({0x105a76820, 0x140005d7380}, {0x14001077150, 0x1, 0x1})
      /private/tmp/skaffold-20220624-80169-j89lh6/cmd/skaffold/app/cmd/run.go:45 +0x108
github.com/GoogleContainerTools/skaffold/cmd/skaffold/app/cmd.withRunner({0x105a2fc70, 0x140004b2240}, {0x1059de580, 0x14000aa07d0}, 0x1400061d540)
[...]

PR #7788, opened 2022-08-19
fix: Fill IMAGE_TAG,etc on Docker builds

0xADD1E:
Fixes: #4295

**Description**
Docs have long stated that these fields are available when using the local docker builder. This makes is to they actually are.

### DIFF
--- a/pkg/skaffold/build/cache/retrieve_test.go
+++ b/pkg/skaffold/build/cache/retrieve_test.go
@@ -142,6 +142,9 @@ func TestCacheBuildLocal(t *testing.T) {
 		t.Override(&docker.EvalBuildArgs, func(_ config.RunMode, _ string, _ string, args map[string]*string, _ map[string]*string) (map[string]*string, error) {
 			return args, nil
 		})
+		t.Override(&docker.EvalBuildArgsWithEnv, func(_ config.RunMode, _ string, _ string, args map[string]*string, _ map[string]*string, _ map[string]string) (map[string]*string, error) {
+			return args, nil
+		})
 
 		// Create cache
 		cfg := &mockConfig{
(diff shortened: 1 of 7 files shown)


aaron-prindle:
Thanks for the PR here!  Can you rebase this PR, I think recently one of the issues in the tests:
```
panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0
```
was fixed - 
https://github.com/GoogleContainerTools/skaffold/issues/7774

Additionally I believe there are some legitimate test failures where some expected values in tests might have to be changed in relation to the change made here.  To run the full set of unit tests locally you can use `make test`.  After we get these tests green we should be able to merge this!

0xADD1E:
Right - so finally got a chance to get back around to this. Looks like it was just "used to be mocking `EvalBuildArgs` (as we were only using `EvalBuildArgs`) but need to also mock `EvalBuildArgsWithEnv` in some places"

It does feel like there *should* be a more elegant way of doing this, particularly with the caching bits (perhaps using `EvalBuildArgsWithEnv` everywhere? That's not how Kaniko is doing it though) -- but I don't think I have a complete enough picture to know exactly what that would be.

In the mean while, I believe the tests should be passing, and the docs match the actual behaviour

aaron-prindle:
Thanks so much for your work here @0xADD1E!  Re:

> It does feel like there should be a more elegant way of doing this, particularly with the caching bits (perhaps using EvalBuildArgsWithEnv everywhere? That's not how Kaniko is doing it though) -- but I don't think I have a complete enough picture to know exactly what that would be.

I've created an issue  - https://github.com/GoogleContainerTools/skaffold/issues/7824  to explore refactoring this as I agree.  

Tests look good now, merging!  Thanks again!

aaron-prindle (approved):
LGTM!

Issue #7779, opened 2022-08-16
arm Image cannot be deployed to mixed nodes cluster


ericzzzzzzz:
### Expected behavior
 - running command ``skaffold dev  --platform linux/arm64 --default-repo=gcr.io/k8s-skaffold``  should successfully deploy image to a mixed cluster without any special configuration.

### Actual behavior
 - pods cannot be scheduled
 - log 
```bash
Waiting for deployments to stabilize...
 - pods: Unschedulable: 0/1 nodes available:
    - pod/getting-started: Unschedulable: 0/1 nodes available:
```
### Information

- Skaffold version:  main
- Operating system:  MAC OS 12.5
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v3alpha1
kind: Config
build:
  artifacts:
  - image: skaffold-example
    context: .
    docker:
      dockerfile: Dockerfile
      noCache: true
  tagPolicy:
    gitCommit: {}
  local:
    concurrency: 1
manifests:
  rawYaml:
  - k8s-*
profiles:
  - name: cloudbuild
    build:
      googleCloudBuild: {}
```

### Steps to reproduce the behavior
 - **connect kubectl to a mixed cluster** 
 - use integration/examples/cross-platform-builds example in skaffold project to run ``skaffold dev  --platform linux/arm64 --default-repo=gcr.io/k8s-skaffold`` 
 - Note that : adding the following toleration to k8s_pod.yaml manifest under spec can make deployment success  but we shouldn't any configuration .. 
  ```yaml
    tolerations:
    - key: kubernetes.io/arch
      operator: Equal
      value: arm64
      effect: NoSchedule
```

PR #7789, opened 2022-08-20
feat: add toleration for GKE ARM nodes taint

gsquared94:
Fixes: #7779 

**Description**

The GKE ARM nodes are tainted with `kubernetes.io/arch="arm64":NoSchedule` by default to prevent workloads from being scheduled. In this PR, Skaffold will automatically add the corresponding toleration to the manifest for ARM image Pods against GKE clusters. 
This can be turned off with the flag `--enable-gke-arm-node-toleration=false`.

**Testing Instruction**
* Use a GKE ARM node cluster (node pool version `1.23.7-gke.1400` or greater). Googlers can use [this](https://pantheon.corp.google.com/kubernetes/clusters/details/us-central1-a/integration-tests-hybird/details?project=k8s-skaffold) cluster
* In the `integration/examples/cross-platform-builds` project, with Skaffold built from `main` branch HEAD, run:
  ```
  skaffold dev --default-repo=gcr.io/k8s-skaffold --platform=linux/arm64
  ```
  This should fail due to the node Taint.
  ```
  Starting deploy...
   - pod/getting-started created
   - pods: Unschedulable: 0/1 nodes available: 
      - pod/getting-started: Unschedulable: 0/1 nodes available: 
   - pods: 0/3 nodes are available: 1 node(s) didn't match Pod's node affinity/selector, 2 node(s) had taint {kubernetes.io/arch: arm64}, that the pod didn't tolerate.
      - pod/getting-started: 0/3 nodes are available: 1 node(s) didn't match Pod's node affinity/selector, 2 node(s) had taint {kubernetes.io/arch: arm64}, that the pod didn't tolerate.
   - pods failed. Error: 0/3 nodes are available: 1 node(s) didn't match Pod's node affinity/selector, 2 node(s) had taint {kubernetes.io/arch: arm64}, that the pod didn't tolerate..
  ```
* Now, build Skaffold from this PR branch and run:
  ```
  skaffold dev --default-repo=gcr.io/k8s-skaffold --platform=linux/arm64
  ```
  This should now succeed.
  ```
  Starting deploy...
   - pod/getting-started created
   - pods is ready.
  Deployments stabilized in 6.941 seconds
  Press Ctrl+C to exit
  Watching for changes...
  [getting-started] Hello world! Running on linux/arm64
  ```


pkg/skaffold/kubernetes/manifest/affinity.go
```diff
@@ -45,21 +45,21 @@ const (
 	nodeArchitectureLabel    = "kubernetes.io/arch"
 )
 
-type ResourceSelectorAffinity struct {
+type ResourceSelectorPodSpec struct {
```
gsquared94, line 48: Because I'm reusing the same resource selector for `tolerations` as for `affinity`, I realise that it's better named as `ResourceSelectorPodSpec` than `ResourceSelectorAffinity`

pkg/skaffold/kubernetes/manifest/gke_toleration.go
```diff
@@ -0,0 +1,115 @@
...
+
+var gkeARMToleration = v1.Toleration{
+	Key:      "kubernetes.io/arch",
+	Operator: v1.TolerationOpEqual,
+	Value:    "arm64",
+	Effect:   v1.TaintEffectNoSchedule,
+}
+
+// SetGKEARMToleration adds a toleration for GKE ARM node taint `kubernetes.io/arch="arm64":NoSchedule`.
+func (l *ManifestList) SetGKEARMToleration(ctx context.Context, rs ResourceSelector, platforms PodPlatforms) (ManifestList, error) {
```
ericzzzzzzz, line 44: Do we have a standard naming style for writing abbreviation with Camel Case. We use upper case abbreviation in variables may cause ambiguity, especially when we have abbreviation sitting along with each other like this, GKEARM,  it's not very intuitive to see where to separate words.  Should we just treat abbreviations as lower case words, then this method name become SetGkeArmToleration, an exception might be we only use upper case abbreviation as the first word of a variable name, something like IOException, JSONParse, the words separation is also clear in this case, but in general take abbreviation as lower case words seem better.

ericzzzzzzz (approved):
test plan works as expected, approved. Left a minor comment about variable naming.

pkg/skaffold/kubernetes/manifest/gke_toleration.go
gsquared94: that's the go style guide for acronyms. See [here](go/go-style/decisions#initialisms)

ericzzzzzzz: Ohh ! thank you!！

PR #7800, opened 2022-08-23
feat: upgrade profiles when running skaffold fix

### LINKED ISSUE #7773: deploy.helm.releases -> manifests.helm.releases from config upgrade breaks profiles
`skaffold fix` (and config upgrade in general) from `v2beta29` to `v3alpha1` does not automatically update the profile paths. Thus, if the `skaffold.yaml` contains any profile patch that points to `deploy.helm.releases`, it would fail.

Example `skaffold.yaml`:

```yaml
apiVersion: skaffold/v2beta29
kind: Config
deploy:
  helm:
    releases:
      - name: skaffold-helm
        chartPath: charts
profiles:
  - name: customize
    patches:
    - op: add
      path: /deploy/helm/releases/0/namespace
      value: customized
```

Running `skaffold fix` produces:

```yaml
apiVersion: skaffold/v3alpha1
kind: Config
manifests:
  helm:
    releases:
    - name: skaffold-helm
      chartPath: charts
deploy:
  helm: {}
profiles:
- name: customize
  patches:
  - op: add
    path: /deploy/helm/releases/0/namespace  # path no longer valid, causing failure
    value: customized
```

The manual fix is to update the path above to `/manifests/helm/releases/0/namespace`, or moving helm back to `deploy` from `manifests`.

Unfortunately, profiles are ignored when upgrading config (pipeline-only thing), so it seems we can't really update these paths while upgrading :sweat_smile: asking here to see if there's any better solution to it

This would cause running skaffold `v2.0.0-beta1` with config `v2beta29` to fail due to invalid paths. Probably helm.flags is also affected, looking at the code.

### Information

- Skaffold version: v2.0.0-beta1
- Operating system: 
- Installed via: [binary](https://github.com/GoogleContainerTools/skaffold/releases)

### Others

[new-render-and-deploy.md](https://github.com/GoogleContainerTools/skaffold/blob/main/docs/design_proposals/new-render-and-deploy.md#helm-deployer) is a good read on the reasoning behind this change

tejal29:
fixes #7773 

In PR, add code to walk through profiles and upgrade the patches in profile as well pipeline config. 

Config considered so far
1) deploy.kubectl.manifests
2) deploy.kustomize.paths
3) deploy.kustomize.buildArgs
4) deploy.helm.releases

Testing instructions. 
1) Checkout this PR and LOCAL=true make
2) `cd integration/testdata/debug`
3) cp the following [skaffold.yaml](https://github.com/GoogleContainerTools/skaffold/blob/v1.39.2/integration/testdata/debug/skaffold.yaml)
via 
```
git fetch origin   v1.39.2
git restore --source  origin/v1.39.2 skaffold.yaml
```
3) run `skaffold fix --overwrite=true`
4) 
```
➜  debug git:(fix_sk_fix) ✗ ../../../out/skaffold fix --overwrite=true 
Backed up previous skaffold.yaml at  skaffold.yaml.v2
New config at version skaffold/v3alpha1 generated and written to skaffold.yaml
```

5) Verify the patches are upgraded correctly
```
➜  debug git:(fix_sk_fix) ✗ tail skaffold.yaml skaffold.yaml.v2               
==> skaffold.yaml <==
  - op: remove
    path: /manifests/rawYaml
  build:
    artifacts:
    - image: skaffold-debug-nodejs
      context: nodejs
  deploy:
    docker:
      images:
      - skaffold-debug-nodejs

==> skaffold.yaml.v2 <==
  patches:
  - op: remove
    path: /deploy/kubectl
  build:
    artifacts:
    - image: skaffold-debug-nodejs
      context: nodejs
  deploy:
    docker:
      images: [skaffold-debug-nodejs
```

### DIFF
--- a/pkg/skaffold/schema/v2beta29/upgrade.go
+++ b/pkg/skaffold/schema/v2beta29/upgrade.go
@@ -18,6 +18,8 @@ package v2beta29
 
 import (
 	"encoding/json"
+	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 
@@ -25,6 +27,13 @@ import (
 	"github.com/GoogleContainerTools/skaffold/pkg/skaffold/schema/util"
 	pkgutil "github.com/GoogleContainerTools/skaffold/pkg/skaffold/util"
 )
+
+var migrations = map[string]string{
+	"/deploy/kubectl/manifests":   "/manifests/rawYaml",
+	"/deploy/kustomize/paths":     "/manifests/kustomize/paths",
+	"/deploy/kustomize/buildArgs": "/manifests/kustomize/buildArgs",
+	"/deploy/helm/releases":       "/manifests/helm/releases",
+}
 
 // Upgrade upgrades a configuration to the next version.
 // v2beta29 is the last config version for skaffold v1, and future version will
@@ -35,7 +44,31 @@ func (c *SkaffoldConfig) Upgrade() (util.VersionedConfig, error) {
 	newConfig.APIVersion = next.Version
 
 	err := util.UpgradePipelines(c, &newConfig, upgradeOnePipeline)
-	return &newConfig, err
+	if err != nil {
+		return &newConfig, err
+	}
+
+	var newProfiles []next.Profile
+	// convert Profiles (should be the same)
+	if c.Profiles != nil {
+		pkgutil.CloneThroughJSON(c.Profiles, &newProfiles)
+	}
+
+	// Update profiles
+	for i, p := range c.Profiles {
+		err = upgradeOnePipeline(&p.Pipeline, &newProfiles[i].Pipeline)
+		if err != nil {
+			return &newConfig, err
+		}
+		err = upgradePatches(p.Patches, newProfiles[i].Patches)
+		if err != nil {
+			return &newConfig, err
+		}
+	}
+
+	newConfig.Profiles = newProfiles
+
+	return &newConfig, nil
 }
 
 func upgradeOnePipeline(oldPipeline, newPipeline interface{}) error {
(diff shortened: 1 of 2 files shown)


pkg/skaffold/schema/v2beta29/upgrade.go
```diff
@@ -124,6 +153,19 @@ func upgradeOnePipeline(oldPipeline, newPipeline interface{}) error {
...
+}
 
+func upgradePatches(olds []JSONPatch, news []next.JSONPatch) error {
+	for i, old := range olds {
+		for str, repStr := range migrations {
+			if strings.Contains(old.Path, str) {
+				news[i].Path = strings.Replace(old.Path, str, repStr, 0)
+			}
+			if strings.Contains(old.Path, "/deploy/kpt") {
+				fmt.Println("skip migrating kpt deploy sections. Please migrate these over manually")
```
aaron-prindle, line 166: should this be `Println` or `log.Entry`?  Also would it make sense to error in the case `/deploy/kpt` is found in a patch profile  stating the above message (is there ever a case where the old patch profile would continue to work) vs printing it out and continuing?  It might make sense not to error, just thinking through the # of error cases vs correct cases possible

tejal29: I was initially inclined to return an error. However, I was wondering if it made sense to convert the main skaffold.yaml.  A user would probably had not configured kpt as a deployer in main skaffold.yaml. (The code would error out when upgrading the pipeline)

e.g.
```
apiVersion: skaffold/v2beta29
kind: Config
deploy:
  manifests:
    k8s:
    - 'base-service/*.yaml'
    
profiles:
   - name: try-kpt
      patches:
       - op: add
          path: /deploy/kpt/
          value:
              <>
```

This could be an edge case and maybe I am over-engineering ?

gsquared94: we could still use `log.Entry.Warn` instead of `Println`

Issue #7787, opened 2022-08-18
Allow specification of serviceAccount in Cloud Build


thegoodlittlescript:
### Expected behavior

I expected to be able to set the service account used by cloud build, when building with `googleCloudBuild`. 

### Actual behavior

I found no such configuration in [the docs](https://skaffold.dev/docs/pipeline-stages/builders/#remotely-on-google-cloud-build). When I inspected my builds they were all building with the default cloudbuild service account, like `{project_number}@cloudbuild.gserviceaccount.com`

### Information

- Skaffold version: v1.35.1
- Operating system: Linux 1526a687ef0c 5.10.104-linuxkit...
- Installed via: skaffold docker image (gcr.io/k8s-skaffold/skaffold:v1.35.1)
- Contents of skaffold.yaml:

```yaml
    apiVersion: skaffold/v2beta17
    kind: Config
    build:
      artifacts:
      - image: runtime
        kaniko:
          cache: {}
          dockerfile: Dockerfile
      googleCloudBuild:
        timeout: 300s
      tagPolicy:
        gitCommit:
          variant: AbbrevCommitSha
```

### Steps to reproduce the behavior

Build via googleCloudBuild. Go look at the cloud build - the Execution Details indicate that the service account is the default one for cloud build. There appears to be no way to configure this.

thegoodlittlescript:
Hopefully this will be easy to implement.  Looks like it can be specified in the cloudbuild.yaml via the `serviceAccount` field.

https://cloud.google.com/build/docs/securing-builds/configure-user-specified-service-accounts

By way of rationale I ran into this when building images FROM an artifact registry repository in a different gcp project.  I ended up having to grant the permissions to read images twice; once to the service account running my pipeline, I believe because skaffold was checking the cache, and a second time to the cloud build service account so the image could be pulled during the build.  I wanted to simply reuse the service account from my pipeline, but I could not.

PR #7843, opened 2022-09-05
add: added service account to cloud build config

OladapoAjala:
Fixes: #7787  

**Description**

This PR adds a Service Account field to Google Cloud Build configurations. 
It allow users set which service account (different from the default) is used by
cloud build.


aaron-prindle:
Thanks for the PR @OladapoAjala!  Currently while this PR adds `serviceAccount` to the schema, it does not plumb the value through our GoogleCloudBuild inputs to the gcloud sdk (which calls the CloudBuild API).  If you could also plumb this value through the buildSpec here:
https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/build/gcb/spec.go#L39

Something like adding 
`buildSpec.ServiceAccount = serviceAccount`
to the end of the function then the serviceAccount will actually get propogated to the CloudBuild call

OladapoAjala:
Thanks @aaron-prindle I just pushed an updated code.

aaron-prindle (approved):
LGTM, seeing one test error that I believe is a flake (not related to the changes you have made).  I am restarting to test and hope it passes.  Once all the tests are green we can merge

PR #7844, opened 2022-09-06
chore: add new marker for integration tests need mixed nodes cluster

ericzzzzzzz:
**Description**
 - add a new testType -  NeedsHybridCluster for integration tests that need a mixed nodes cluster
 - Related build config will also be set on kokoro.

### DIFF
--- a/integration/util.go
+++ b/integration/util.go
@@ -49,10 +49,16 @@ type TestType int
 const (
 	CanRunWithoutGcp TestType = iota
 	NeedsGcp
+	NeedsHybridCluster
 )
 
 func MarkIntegrationTest(t *testing.T, testType TestType) {
 	t.Helper()
+	runOnHybridCluster := os.Getenv("HYBRID_ONLY") == "true"
+	if runOnHybridCluster != (NeedsHybridCluster == testType) {
+		t.Skipf("Skipping test due to testType doesn't match environment setup : runOnHybrid: %t, NeedsHybridCluster: %t", runOnHybridCluster, NeedsHybridCluster == testType)
+	}
+
 	if testing.Short() {
 		t.Skip("skipping integration test")
 	}


integration/util.go
```diff
@@ -49,10 +49,16 @@ type TestType int
 const (
 	CanRunWithoutGcp TestType = iota
 	NeedsGcp
+	NeedsHybridCluster
 )
 
 func MarkIntegrationTest(t *testing.T, testType TestType) {
 	t.Helper()
+	runOnHybridCluster := os.Getenv("HYBRID_ONLY") == "true"
+	if runOnHybridCluster != (NeedsHybridCluster == testType) {
```
gsquared94, line 58:
```suggestion
	if testType == NeedsHybridCluster && !runOnHybridCluster {
```

ericzzzzzzz: > 
I think we want mixed nodes cluster only runs tests that require it. There are 4 cases. 
 - NeedsHybridCluster && runOnHybridCluster is true ,  the condition true != true  is false, we're expecting cross-platform tests are running in this case, and test should not be skipped, other conditions will not stop the test from running.
 - NeedsHybridCluster && runOnHybridCluster is false, the condition true != false is true, the test should be skipped as the test cannot be running on the machine connected to single-platform cluster
 - Not NeedsHybridCluster && runOnHybridCluster is true, the condition false != true is true. although the machine connected to mixed nodes cluster may be able to run the test, we don't need it to run the test as the test should be covered in other CI presubmit runs, so this case should be skipped. 
 - Not NeedsHybridCluster && runOnHybridCluster is false, the condition false != false is false, the this will not be skipped by this condition, but other conditions will determine if test can be running in this case. 

Note that, the env variable gcp_only and hybrid_only are actually mutually exclusive, we can also merge them into one env variable as test_cluster, the values could be LOCAL, GCP_SINGLE_ONLY, GCP_HYBRID_ONLY, that also requires kokoro build config change as well, if that's preferred way.I can do the refactor and submit a cl for kokoro change.

gsquared94: discussed offline with Eric. We'll try to not require a new marker, rather target all NEEDS_GCP tests against different cluster types.

ericzzzzzzz: Closing the pr.

ericzzzzzzz:
Discussed offline with gaurav, we will specify cluster name on kokoro build config to determine which cluster test should be running, and we will run all gcp tests against different clusters.

PR #7846, opened 2022-09-06
chore: adding status check phase with v2 taskEvent

ericzzzzzzz:
### Description
 - This pr adds a v2 status check phase with taskEvent as we have that phase with v1 event.


aaron-prindle (approved):
LGTM!

PR #7851, opened 2022-09-08
Update ko docs to describe new default base image

imjasonh:
`ko` v0.12 updated the default base image to one built using apko instead of Bazel.

cc @halvards

### DIFF
--- a/docs/content/en/docs/pipeline-stages/builders/ko.md
+++ b/docs/content/en/docs/pipeline-stages/builders/ko.md
@@ -56,10 +56,9 @@ build:
 
 ### Base image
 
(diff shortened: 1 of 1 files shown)


halvards:
Thanks for spotting this @imjasonh!

Skaffold is still on ko v0.11.2, so we'll hold off on merging this until we move to v0.12.0 or later.

aaron-prindle:
@halvards do you have an opinion on bumping the ko version that skaffold uses?  Is this something it makes sense to create an issue for and bump (not sure what features, etc. ko `v0.12.0` has)?

Issue #7865, opened 2022-09-18
customTemplate tagger doesn't have any alias to inputDigest tagger


mikekamornikov:
### Expected behavior

The following `tagPolicy` is supported
```yaml
tagPolicy:
  customTemplate:
    template: test-{{.INPUT_DIGEST}}
```

### Actual behavior

`inputDigest` is supported only if it's explicitly added to components. While other taggers have aliases (`SHA`, `GIT`, `DATE`):

```yaml
tagPolicy:
  customTemplate:
    template: test-{{.SHA}}-{{.GIT}}-{{.DATE}}-{{.INPUT_DIGEST}}
    components:
    - name: INPUT_DIGEST
      inputDigest: {}
```

### Information

- Skaffold version: 1.39.2
- Operating system: macOS 12.4
- Installed via: downloaded from Github release assets
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta29
kind: Config
build:
  artifacts:
  - image: quay.io/org/project
    context: ../../..
    bazel:
      target: //projects/go/cmd/project:docker.tar
      args:
      - --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64
  tagPolicy:
    customTemplate:
      template: cxp-mothership-operator-{{.SHA}}-{{.GIT}}-{{.DATE}}-{{.INPUT_DIGEST}}
  local:
    useBuildkit: false
```

### Steps to reproduce the behavior

1. Modify `tagPolicy` in `skaffold.yaml`
2.  Run `skaffold build`

PR #7867, opened 2022-09-19
fix: add inputDigest tagger alias to customTemplate tagger

mikekamornikov:
Fixes: #7865

**Description**
It modifies `customTemplate` tagger to support `inputDigest` by default:
- Given template is parsed to understand which variables to expect. This way it's not required to initialize all possible taggers in advance. And it seriously simplifies testing.
- Taggers are created one by one giving higher priority to the ones defined in `components`.

### DIFF
--- a/pkg/skaffold/tag/custom_template.go
+++ b/pkg/skaffold/tag/custom_template.go
@@ -22,24 +22,28 @@ import (
 	"fmt"
 	"text/template"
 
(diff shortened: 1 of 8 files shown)


aaron-prindle (approved):
LGTM!  Thanks for this, this is great!  I have created https://github.com/GoogleContainerTools/skaffold/issues/7933 to track updating the Skaffold documentation to also include the new INPUT_DIGEST field

PR #7874, opened 2022-09-21
Add an example for helm chart with multiple images

filipenf:
Shows how to override multiple images in a helm template

**Description**
Documentation-only changes: Add a snippet showing how to override multiple images from skaffold-built artifacts. I was testing with this and thought an example in the doc would be useful. Thoughts?

### DIFF
--- a/docs-v2/content/en/docs/pipeline-stages/deployers/helm.md
+++ b/docs-v2/content/en/docs/pipeline-stages/deployers/helm.md
@@ -23,6 +23,79 @@ Skaffold no longer requires the intricate configuring of `artifactsOverride` and
 {{< /alert >}}
 
 
+## Image Configuration
+The normal Helm convention for defining image references is through the `values.yaml` file. Often, image information is configured through an `image` stanza in the values file, which might look something like this:
+
+```project_root/values.yaml```
+```yaml
+image:
+  repository: gcr.io/my-project/my-image
+  tag: v1.2.0
+  pullPolicy: IfNotPresent
+```
+
+This image would then be referenced in a templated resource file, maybe like this:
+
+```project_root/templates/deployment.yaml:```
+```yaml
+spec:
+  template:
+    spec:
+      containers:
+        - name: {{ .Chart.Name }}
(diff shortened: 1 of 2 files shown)


aaron-prindle:
Thanks for the submission @filipenf!  Currently you have updated:
`docs/content/en/docs/pipeline-stages/deployers/helm.md`

We have recently forked our docs into a `docs` and a `docs-v2` folder.  Can you also update the
docs-v2/content/en/docs/pipeline-stages/deployers/helm.md
file with this change as well?

Additionally our repo has a unit test to enforce [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).  Can you ammend your commit message to add a prefix - `docs:` eg:
`docs: Add an example for helm chart with multiple images`

filipenf:
@aaron-prindle thanks for the info, I just checked the helm.md in the v2 path you mentioned and I noticed that it is missing a few of the sections from the original file. Would you also like me to match those or should I just add the new bits?

aaron-prindle:
@filipenf just the new bits thanks! I believe the other differences are inherent to Skaffold v1 vs v2

filipenf:
@aaron-prindle I added just the new bits and updated the commit message, however I noticed that the whole `## Image Configuration` section is missing in the v2 doc. Is that expected?

aaron-prindle:
@filipenf ah, I see now.  Yes, if you wouldn't mind adding that section as well here that would be great.  Thanks for extra analysis here!

aaron-prindle (approved):
LGTM!  Thanks!

iosifnicolae2:
When using schema v3 `artifactOverrides` does not exist, instead I've used `setValueTemplates` as follow:
```yaml
manifests:
  helm:
    releases:
    - setValueTemplates:
        firstContainerImage.repo: '{{.IMAGE_REPO}}'
        firstContainerImage.tag: '{{.IMAGE_TAG}}'
        secondContainerImage.repo: '{{.IMAGE_REPO2}}'
        secondContainerImage.tag: '{{.IMAGE_TAG2}}'
        ...
```

PR #7884, opened 2022-09-23
fix: ensure project descriptor in Pack Build

willpxxr:
**Description**
Currently Skaffold BuildPack implementation does not correctly support most of the properties in BuildPacks project descriptor, as a user defined project descriptor is an optional value - but is not correctly passed into the build options during pack build.

A side affect of this is that build.include or build.exclude as documented here: https://buildpacks.io/docs/app-developer-guide/using-project-descriptor/ does not correctly ignore (or exclusively include) files.

This means that for projects with larges files ie. node_modules in a nodejs project will be incorrectly copied into the `/workspace` mount destination.

ie. Equivalent of Pack CLI commands:

```
...
[build]
exclude = [
   "ignore_me"
]
...
```

Using Pack CLI will correctly exclude matching files from the build container
```bash
pack build mycontainer:latest -B myBuilder:base --run-image myRunImage:base-cnb --trustbuilder -d project.toml
```

However Skaffolds implementation is Equivalent to
```bash
pack build mycontainer:latest -B myBuilder:base --run-image myRunImage:base-cnb --trustbuilder
```

As project.toml is used to extract certain fields, but other fields - such as exclude/include are essentially ignored.

### DIFF
--- a/pkg/skaffold/build/buildpacks/build_test.go
+++ b/pkg/skaffold/build/buildpacks/build_test.go
@@ -23,6 +23,7 @@ import (
 
 	pack "github.com/buildpacks/pack/pkg/client"
 	packcfg "github.com/buildpacks/pack/pkg/image"
+	packtypes "github.com/buildpacks/pack/pkg/project/types"
 	"github.com/docker/docker/api/types"
 	"github.com/google/go-cmp/cmp"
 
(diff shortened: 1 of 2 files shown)


aaron-prindle (approved):
LGTM, thanks for adding this support @willpxxr!

Issue #7711, opened 2022-08-01
Spurious warning when using minikube with "none" driver


gilbahat:
Skaffold is confused when running under local "none" minikube.

# minikube docker-env

❌  Exiting due to ENV_DRIVER_CONFLICT: 'none' driver does not support 'minikube docker-env' command

### Expected behavior

Skaffold ignores the specific error issued by minikube when the 'none' driver is used

### Actual behavior

Skaffold complains about the output of minikube docker-env but proceeds correctly:
WARN[0000] Could not get minikube docker env, falling back to local docker daemon: getting minikube env: running [/usr/local/bin/minikube docker-env --shell none -p minikube --user=skaffold]

note - this is kinda cosmetic but I wonder if it could be an issue with other nonstandard drivers such as ssh.

### Information

- Skaffold version: 1.39.1

### Steps to reproduce the behavior

1. install a none-driver minikube instance - https://minikube.sigs.k8s.io/docs/drivers/none/
2. run any skaffold build

tejal29:
## Background

Minikube  can be deployed as a VM, a container, or bare-metal. In case of container, user can choose the docker driver.  (list [here](https://minikube.sigs.k8s.io/docs/drivers/))

When minikube is started with docker driver, the k8s cluster is installed  into an existing Docker install.

Currently in skaffold, if we detect a user has minikube as their active kubecontext, we run a command  `minikube docker-env ` so that docker commands on the host machine build to minikube's docker daemon.
What this means, running a `docker build `will run inside minikube's docker daemon and the image will be available on the minikube cluster without a need to push.

Here is why, it is beneficial to use docker's deamon. https://stackoverflow.com/questions/52310599/what-does-minikube-docker-env-mean/52310892#52310892

## Problem: 
docker driver flow was the most used flow before k8s decided to [deprecate docker shim](https://kubernetes.io/blog/2022/01/07/kubernetes-is-moving-on-from-dockershim/#:~:text=We%20formally%20announced%20the%20dockershim,year%20after%20their%20announced%20deprecation.)  A lot of users started using different container runtimes or drivers when creating minikube cluster. 
Skaffold needs to 
1) verify if the minikube is using docker driver. There is a command to do so on the minikube docker driver page. 
2) If the driver is not other than docker, skaffold should use the `minikube image load` command which can load a image into the minikube cluster. https://minikube.sigs.k8s.io/docs/handbook/pushing/#7-loading-directly-to-in-cluster-container-runtime
[...]

PR #7904, opened 2022-09-30
feat: support-other-container-drivers

OladapoAjala:
Fixes: #7711  

**Description**

This PR fixes issue with using non docker driver with a minikube cluster.


pkg/skaffold/docker/client.go
```diff
@@ -105,9 +106,32 @@ type ExitCoder interface {
...
-// newMinikubeAPIClient returns a docker client using the environment variables
+// newMinikubeAPIClient returns a client using the environment variables
 // provided by minikube.
 func newMinikubeAPIClient(ctx context.Context, minikubeProfile string) ([]string, client.CommonAPIClient, error) {
+	driver, err := getMinikubeDriver(ctx, minikubeProfile)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if driver == "docker" {
```
tejal29, line 117: add a constant for "docker" in the same file.

pkg/skaffold/docker/client.go
```diff
@@ -105,9 +106,32 @@ type ExitCoder interface {
...
+		return newMinikubeAPIClientWithDockerDriver(ctx, minikubeProfile)
+	}
+
+	cmd, err := cluster.GetClient().MinikubeExec(ctx, "image", "load")
+	if err != nil {
+		return nil, nil, fmt.Errorf("executing minikube command: %w", err)
```
tejal29, line 123: lets have actionable errors for there. Add an action code and a suggestion to see `minikube logs` to figure out what is wrong.

pkg/skaffold/docker/client.go
```diff
@@ -175,6 +199,27 @@ func newMinikubeAPIClient(ctx context.Context, minikubeProfile string) ([]string
 	return environment, api, err
 }
 
+func getMinikubeDriver(ctx context.Context, minikubeProfile string) (string, error) {
+	if minikubeProfile == "" {
+		return "", fmt.Errorf("empty minikube profile")
+	}
```
tejal29, line 205: Minikube usually has a default profile "minikube".  I am not sure if Skaffold reads the default profile as "minikube" or just empty string from kubecontext. We will have the check the workflow and make sure it doesn't error out.

tejal29 (commented):
just few nits. 

Looks good overall. 
1. Adding error codes and re-assessing if we need constants.

pkg/skaffold/docker/client.go
tejal29: Note: minikube also provides exit codes. See example here https://github.com/GoogleContainerTools/skaffold/pull/7904/files#diff-f7852592ac185659a3acd7d1b0c5127703270e8c7143128f79cab2d24c02011eR140

Maybe we should refactor the code in the above lines and make a func to extract error codes from minikube errors

OladapoAjala:
Thanks, another thing again the test [TestNewMinikubeImageAPIClient](https://github.com/GoogleContainerTools/skaffold/blob/13a25fa33e826aeeefda4f0eebc532c676345a88/pkg/skaffold/docker/client_test.go#L63) fails because it expects a single call to `util.RunCmd` any suggestions how to fix it?

pkg/skaffold/docker/client.go
OladapoAjala: I think you added the wrong link here @tejal29

tejal29:
> Thanks, another thing again the test [TestNewMinikubeImageAPIClient](https://github.com/GoogleContainerTools/skaffold/blob/13a25fa33e826aeeefda4f0eebc532c676345a88/pkg/skaffold/docker/client_test.go#L63) fails because it expects a single call to `util.RunCmd` any suggestions how to fix it?

I think we have a utility to specify another command run `testutil.AddRunOut(`
There are a lot of examples of it. Please see other tests on how to use or explore the class.

pkg/skaffold/docker/client.go
tejal29: https://github.com/GoogleContainerTools/skaffold/blob/eefc9fa01555d972847aeddd11511c929007165b/pkg/skaffold/docker/client.go They are in the same file. Just a bit below. I think you since you updated the diff, the link does not work.

tejal29: https://github.com/GoogleContainerTools/skaffold/blob/eefc9fa01555d972847aeddd11511c929007165b/pkg/skaffold/docker/client.go#L117

aaron-prindle:
Currently when I actually attempt to use this PR for the use case mentioned in https://github.com/GoogleContainerTools/skaffold/issues/7711 I get the following:

```
aprindle@minikube-none:~/skaffold/examples/getting-started$ sudo skaffold-oladapo dev
invalid skaffold config: loading cluster image: exit status 14
```

I'm assumming this is because the `minikube none driver` does not support minikube load as this driver is essentially running kubernetes on the host directly meaning that the local/default docker env is used (vs other minikube drivers)

aaron-prindle:
Currently you are getting a `lint` error:
```
pkg/skaffold/docker/client.go:36: File is not `gci`-ed with --skip-generated -s standard,default,prefix(github.com/GoogleContainerTools/skaffold) (gci)
	"github.com/docker/docker/client"
	"github.com/docker/go-connections/tlsconfig"
```
You can resolve this by installing the `gci` tool here: https://github.com/daixiang0/gci#installation

and running
```
gci write --skip-generated -s standard,default,prefix(github.com/GoogleContainerTools/skaffold) pkg/skaffold/docker/client.go
```

OladapoAjala:
> Currently you are getting a `lint` error:
> 
> ```
> pkg/skaffold/docker/client.go:36: File is not `gci`-ed with --skip-generated -s standard,default,prefix(github.com/GoogleContainerTools/skaffold) (gci)
> 	"github.com/docker/docker/client"
> 	"github.com/docker/go-connections/tlsconfig"
> ```
> 
> You can resolve this by installing the `gci` tool here: https://github.com/daixiang0/gci#installation
> 
> and running
> 
> ```
> $ gci write --skip-generated -s standard,default,prefix(github.com/GoogleContainerTools/skaffold) pkg/skaffold/docker/client.go
> ```

Pasting the correct command here for future reference:

```
gci write --skip-generated -s "standard,default,prefix(github.com/GoogleContainerTools/skaffold)" pkg/skaffold/docker/client.go
```

aaron-prindle:
Thanks for updating the PR here @OladapoAjala, in running the unit tests locally on linux using minikube `v1.28.0` I don't seem to be seeing some test issues related to `minikube profile list -o json --user=skaffold` returning error code `80` consistently in tests.  Potential interesting things to note in debugging this:
- `windows` and `mac` tests only run unit tests and not the full suite of integration tests.  I don't believe these CI/CD tests have `minikube` installed which is likely why they error in attempting to locate (`~/.minikube/...` or unable to find `minikube profiles list ...`)

OladapoAjala:
@aaron-prindle are you suggesting to include an installation of minikube in the pipeline for windows and mac?

OladapoAjala:
@aaron-prindle @tejal29, I've updated this PR for cases where the cluster isn't managed by minikube. Some of the CI tests still fail at "install GCloud".

Kindly let me know what you think. Thanks

ericzzzzzzz:
> @aaron-prindle @tejal29, I've updated this PR for cases where the cluster isn't managed by minikube. Some of the CI tests still fail at "install GCloud".
> 
> Kindly let me know what you think. Thanks

Hi @OladapoAjala could you try to rebase on main? That gcloud install issue should be fixed now.

OladapoAjala:
Thanks @ericzzzzzzz

OladapoAjala:
@ericzzzzzzz @tejal29 @aaron-prindle could you please review this? 🤞🤞🤞

pkg/skaffold/docker/client.go
```diff
@@ -65,29 +71,46 @@ type Config interface {
...
+		if err != nil {
+			return nil, nil, err
+		}
+		return newMinikubeAPIClient(ctx, profile, driver)
 	}
+
 	return newEnvAPIClient()
 }
 
+func getMinikubeProfile(ctx context.Context) (string, string, error) {
```
ericzzzzzzz, line 95: Although the original implementation assumes that kubecontext name is never changed after a minikube profile is created and kubecontext(name) is the same as profile name, this is more likely to be true than the current one which assumes that  kubectl is configured to connect to the first profile. So I think we could use this method to get the driver with profile(kubecontext) from the result of ``minikube profile list -o json``,  I cannot find the command to get the detail of a profile, although we can ``cat ~/.minikube/profiles/{{profile}}/config.json`` to get the profile detail directly, that seems dangerous, I wouldn't suggest to do that.

pkg/skaffold/docker/client.go
```diff
@@ -105,9 +128,36 @@ type ExitCoder interface {
...
+func newMinikubeAPIClient(ctx context.Context, minikubeProfile string, minikubeDriver string) ([]string, client.CommonAPIClient, error) {
+	if minikubeDriver == noneDriver {
+		return newEnvAPIClient()
+	}
+
+	if minikubeDriver == dockerDriver {
+		return newMinikubeAPIClientWithDockerDriver(ctx, minikubeProfile)
+	}
+
+	cmd, err := cluster.GetClient().MinikubeExec(ctx, "image", "load")
```
ericzzzzzzz, line 142: We don't need this method here,  ``minikube image load`` require images argument(build result).. this would just produce an error if this part get executed, we could do similar stuff here https://github.com/GoogleContainerTools/skaffold/blob/b4fba3990841eb3477f987588c5e96a34d7f8d2d/pkg/skaffold/kubernetes/loader/load.go#L88-L115 before deploying stuff to local cluster when driver is neither ``none`` or ``docker`` , but as we're currently not sure how ``minikube docker-env`` would behave for other drivers, we would prefer not including this in this pr,  what do you think?

pkg/skaffold/docker/client.go
```diff
@@ -105,9 +128,36 @@ type ExitCoder interface {
...
+			return nil, nil, sErrors.MinikubeImageLoadError(minikubeExGuestError, err)
+		}
+		return nil, nil, err
+	}
+
+	err = util.RunCmd(ctx, cmd)
+	if err != nil {
+		return nil, nil, fmt.Errorf("loading cluster image: %w", err)
+	}
+	return nil, nil, nil
```
ericzzzzzzz, line 155: can we return ``newEnvAPIClient() `` here?

ericzzzzzzz (commented):
left some comments.

OladapoAjala:
Thanks I'd look into it 🙂

aaron-prindle:
Hey @OladapoAjala!  I think this might need a rebase now

aaron-prindle:
@OladapoAjala any updates here?  Definitely let us know if we can help in any way to get this through, thanks!

aaron-prindle:
Hey @OladapoAjala!  Hope all is well, just wanted to bump the PR here to see if you were still pursuing this and ask how we might be able to help with any questions, work-items, etc.  Thanks!

aaron-prindle:
Friendly ping @OladapoAjala, let me know if the Skaffold team can help in any way to get this PR across the finish line.  Thanks!

OladapoAjala:
@aaron-prindle thanks for pinging me on this, it appears I already gave up on this. I'd go ahead to close it and have a look around other issues when I'm more chanced.

Thanks.

Issue #6488, opened 2021-08-23
[Docs fixit] [technical writers review] [HaTS feedback] Pages should link to relevant github repo `examples`


gsquared94:
The [`examples`](https://github.com/GoogleContainerTools/skaffold/tree/main/examples) directory is a treasure trove of many direct usage examples for various concepts explained in the docs. We should be able to link the relevant examples directly from a documentation page.
The right side pane for every page could have an `examples` section similar to the `Applicable/Not applicable` section.
![image](https://user-images.githubusercontent.com/39389231/130473876-fdf5ed1e-8a94-47e1-b696-9b88c8f3d926.png)

maggieneterval:
I am working on this.

PR #7906, opened 2022-10-03
docs: Add examples to the info sidebar

maggieneterval:
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/6488

**Description**

Add relevant examples from the [examples directory](https://github.com/GoogleContainerTools/skaffold/tree/main/examples) to the sidebar of each page.

**User facing changes**

Before:

Screen Shot 2022-10-03 at 11 42 34 AM

After: 

Screen Shot 2022-10-03 at 11 38 11 AM

### DIFF
--- a/docs-v2/layouts/partials/info-panel.html
+++ b/docs-v2/layouts/partials/info-panel.html
@@ -9,6 +9,7 @@
 {{ $deploy := $data.deploy }}
 {{ $render := $data.render }}
 {{ $all := and $dev $debug $run $build $deploy $render }}
+{{ $examples := $data.examples }}
 
 <div class="alert infopanel" role="contentinfo">
     <table>
@@ -45,6 +46,17 @@
             </td>
         </tr>
         {{end}}
+        
+        {{ if $examples}}
+        <tr>
+            <td><b>Code Samples</b></td>
+            <td>
+                {{ range $example := $examples }}
+                <a href="{{$example.url}}"><code>{{$example.name}}</code></a>
+                {{ end }}
+            </td>
+        </tr>
+        {{ end }}
     </table>
 </div>
 
--- a/docs/data/maturity.json
+++ b/docs/data/maturity.json
@@ -19,7 +19,25 @@
     "area": "Build",
     "feature": "Cloud Native Buildpacks support",
     "maturity": "GA",
-    "description": "Skaffold natively support for artifacts built with Cloud Native Buildpacks"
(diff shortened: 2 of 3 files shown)


maggieneterval:
@aaron-prindle Could you please review this when you get a chance? Let me know if there are any presubmit steps or conventions I missed. Thank you!

docs-v2/layouts/partials/info-panel.html
```diff
@@ -45,6 +46,17 @@
             </td>
         </tr>
         {{end}}
+        
+        {{ if $examples}}
+        <tr>
+            <td><b>Examples</b></td>
```
aaron-prindle, line 52: nit: can this be "Code Samples" vs examples.  I am not sure if `Examples` is obvious to someone coming to these docs.

docs/data/maturity.json
```diff
@@ -19,7 +19,25 @@
     "area": "Build",
     "feature": "Cloud Native Buildpacks support",
     "maturity": "GA",
-    "description": "Skaffold natively support for artifacts built with Cloud Native Buildpacks"
+    "description": "Skaffold natively support for artifacts built with Cloud Native Buildpacks",
```
aaron-prindle, line 22: nit I think this `Skaffold natively supports artifacts built with Cloud Native Buildpacks`

(This isn't related to your changes but might be easy to slip in)

aaron-prindle:
nit: Can you also copy these changes over to `docs/` (as well as `docs-v2/`.  This changes would also help people looking at the v1 docs

docs-v2/layouts/partials/info-panel.html
maggieneterval: Makes sense; updated from "Examples" to "Code Samples" (here and in the v1 partial).

docs/data/maturity.json
maggieneterval: Updated.

maggieneterval:
> nit: Can you also copy these changes over to docs/ (as well as docs-v2/. This changes would also help people looking at the v1 docs

My understanding was that docs/layouts/partials/info-panel.html was the partial for v1, and docs-v2/layouts/partials/info-panel.html was the partial for v2. Both are updated; are there updates to any other files required for these changes to take effect in both v1 and v2?

aaron-prindle:
Sorry, my mistake - you are right,yYou did the correct work initially, I didn't realize both `docs/` and `docs-v2/` were updated.  LGTM, nothing else needs to be done there

Issue #7674, opened 2022-07-22
skaffold run on a kpt sample fails


tejal29:
skaffold v2 run on a kpt sample fails with error below
```
➜  kpt_demo git:(guestbook-sample-v2) ✗ skaffoldv2 run  -d gcr.io/tejal-gke1
Generating tags...
 - redis-slave -> gcr.io/tejal-gke1/redis-slave:latest
 - php-redis -> gcr.io/tejal-gke1/php-redis:latest
 - skaffold-helm -> gcr.io/tejal-gke1/skaffold-helm:latest
Checking cache...
 - redis-slave: Not found. Building
 - php-redis: Not found. Building
 - skaffold-helm: Found Remotely
Starting build...
Building [php-redis]...
Target platforms: [linux/amd64]
[+] Building 2.0s (13/13) FINISHED                                                                                                                                                   
 => [internal] load build definition from Dockerfile                                                                                                                            0.0s
 => => transferring dockerfile: 1.28kB                                                                                                                                          0.0s
 => [internal] load .dockerignore                                                                                                                                               0.0s
 => => transferring context: 2B                                                                                                                                                 0.0s
 => [internal] load metadata for docker.io/library/php:5-apache                                                                                                                 0.4s
 => [1/8] FROM docker.io/library/php:5-apache@sha256:0a40fd273961b99d8afe69a61a68c73c04bc0caa9de384d3b2dd9e7986eec86d                                                           0.0s
 => [internal] load build context                                                                                                                                               0.0s
 => => transferring context: 99B                                                                                                                                                0.0s
 => CACHED [2/8] RUN apt-get update                                                                                                                                             0.0s
 => CACHED [3/8] RUN pear channel-discover pear.nrk.io                                                                                                                          0.0s
 => [4/8] RUN sed -i 's#ErrorLog /proc/self/fd/2#ErrorLog "|$/bin/cat 1>\&2"#' /etc/apache2/apache2.conf                                                                        0.4s
 => [5/8] RUN sed -i 's#CustomLog /proc/self/fd/1 combined#CustomLog "|/bin/cat" combined#' /etc/apache2/apache2.conf                                                           0.6s
 => [6/8] ADD guestbook.php /var/www/html/guestbook.php                                                                                                                         0.0s
 => [7/8] ADD controllers.js /var/www/html/controllers.js                                                                                                                       0.0s
 => [8/8] ADD index.html /var/www/html/index.html                                                                                                                               0.0s
[...]

PR #7909, opened 2022-10-04
fix: kpt render with subdirectories

renzodavid9:
Fixes: #7674

**Description**
This change set the proper value to the Kpt hydration directory in the Kpt renderer, so the created manifests are placed in the proper `.kpt-pipeline` folder where they can be managed by skaffold automatically.

Also, we are throwing a new error when a Kpt deployer is detected, but the version is greater than the required.


pkg/skaffold/schema/validation/validation.go
```diff
@@ -87,6 +88,7 @@ func ProcessToErrorWithLocation(configs parser.SkaffoldConfigSet, validateConfig
 		errs = append(errs, validateCustomTest(config, config.Test)...)
 		errs = append(errs, validateGCBConfig(config, config.Build)...)
 		errs = append(errs, validateVerifyTests(config, config.Verify)...)
+		errs = append(errs, validateKptDeployerVersion(config, config.Deploy)...)
```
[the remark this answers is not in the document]
aaron-prindle: +1, this is the better place to do this (vs where Helm does this).  One issue might be that currently this will disallow something like:
`skaffold render` + kpt deployer exists but this would actually work properly (only `dev`, `deploy`, etc. would actually give us an issue)

pkg/skaffold/schema/validation/samples_test.go
```diff
@@ -55,9 +57,13 @@ func TestParseExamples(t *testing.T) {
 // Samples are skaffold.yaml fragments that are used
 // in the documentation.
 func TestParseSamples(t *testing.T) {
-	paths, err := walk.From(samplesRoot).WhenIsFile().CollectPaths()
-	if err != nil {
-		t.Fatalf("unable to list samples in %q", samplesRoot)
+	// paths, err := walk.From(samplesRoot).WhenIsFile().CollectPaths()
```
gsquared94, line 60: why is this code commented out?

pkg/skaffold/schema/validation/validation.go
```diff
@@ -99,6 +101,18 @@ func ProcessToErrorWithLocation(configs parser.SkaffoldConfigSet, validateConfig
 	return errs
 }
 
+func validateKptDeployerVersion(cfg *parser.SkaffoldConfigEntry, dc latest.DeployConfig) (cfgErrs []ErrorWithLocation) {
+	if dc.KptDeploy != nil {
```
gsquared94, line 105: nit: invert conditional
```
if dc.KptDeploy == nil {
  return
}
...
```

pkg/skaffold/schema/validation/samples_test.go
renzodavid9: My bad. Fixed

pkg/skaffold/schema/validation/validation.go
renzodavid9: I changed it to do an early return, thanks! Done.

aaron-prindle (approved):
LGTM!

pkg/skaffold/schema/validation/validation.go
```diff
@@ -99,6 +101,25 @@ func ProcessToErrorWithLocation(configs parser.SkaffoldConfigSet, validateConfig
...
+	kptRendererWillBeCreated := rc.Kpt != nil || rc.Transform != nil || rc.Validate != nil
+
+	if kptRendererWillBeCreated {
+		if err := kpt.CheckIsProperBinVersion(context.TODO()); err != nil {
+			cfgErrs = append(cfgErrs, ErrorWithLocation{
+				Error:    err,
+				Location: cfg.YAMLInfos.LocateField(cfg, "Render"),
+			})
+		}
+	}
```
gsquared94, line 118:
```suggestion
	if rc.Kpt == nil && rc.Transform == nil && rc.Validate == nil { // no kpt renderer created
		return
	}
	if err := kpt.CheckIsProperBinVersion(context.TODO()); err != nil {
		cfgErrs = append(cfgErrs, ErrorWithLocation{
			Error:    err,
			Location: cfg.YAMLInfos.LocateField(cfg, "Render"),
		})
	}
```

gsquared94: reduce nesting.

renzodavid9: Done!

Issue #7284, opened 2022-04-11
Clean up order should be reverse of install order


stanislav-zaprudskiy:
### Expected behavior

`skaffold dev` cleans up in the reverse order of its install order.

### Actual behavior

`skaffold dev` cleans up in the same order (?) as its install order.

### Information

- Skaffold version: `v1.38.0`
- Operating system: `macOS 12.3.1`
- Installed via: skaffold.dev
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta28
kind: Config
metadata:
  name: all
requires:
  - configs:
      - postgres-operator-crds
      - postgres-operator

---
apiVersion: skaffold/v2beta28
kind: Config
metadata:
  name: postgres-operator-crds
deploy:
  helm:
    flags:
      install:
        - "--skip-crds"
      upgrade:
        - "--skip-crds"
    releases:
      - name: postgres-operator-crds
        chartPath: ./helm-charts/postgres-operator-crds
        wait: true

---
apiVersion: skaffold/v2beta28
kind: Config
metadata:
  name: postgres-operator
deploy:
  helm:
    flags:
      install:
        - "--skip-crds"
      upgrade:
        - "--skip-crds"
    releases:
      - name: postgres-operator
        remoteChart: postgres-operator
        repo: https://opensource.zalando.com/postgres-operator/charts/postgres-operator/
        version: 1.7.1
        namespace: awx
        createNamespace: true
        wait: true

```

### Steps to reproduce the behavior

The idea is that CRDs are installed first, and then the actual CRs are deployed - which works fine for the initial installation. However, when clean up of `skaffold dev` runs, it seems to be executing clean up in the same order as install: first `postgres-operator-crds` is removed, but then the removal of `postgres-operator` fails as CRDs aren't there anymore.

```console
$ skaffold dev --module all
Listing files to watch...
Generating tags...
Checking cache...
Tags used in deployment:
Starting deploy...
Helm release postgres-operator-crds not installed. Installing...
NAME: postgres-operator-crds
LAST DEPLOYED: Mon Apr 11 21:18:24 2022
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
Helm release postgres-operator not installed. Installing...
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
[...]

PR #7925, opened 2022-10-06
fix: reverse order of deployers during cleanup (#7284)

chronicc:
Fixes: #7284 

**Description**

Add a new deployer slice during cleanup function of the DeployerMux which is reversed in it's order. This change should be covered by the existing unittests since no new function was added.

Before:
Modules are undeployed in the same order as they are deployed. Visible in the console as

```shell
^CCleaning up...
release "postgresql" uninstalled
 - deployment.apps "example-django" deleted
INFO[0016] Cleanup completed in 886.10607ms              subtask=-1 task=DevLoop

```

After:
Modules are undeployed in reverse order.

```shell
^CCleaning up...
 - deployment.apps "example-django" deleted
release "postgresql" uninstalled
INFO[0328] Cleanup completed in 1.038 second             subtask=-1 task=DevLoop
```

### DIFF
--- a/pkg/skaffold/deploy/deploy_mux.go
+++ b/pkg/skaffold/deploy/deploy_mux.go
@@ -57,6 +57,14 @@ func (m DeployerMux) GetDeployers() []Deployer {
 	return m.deployers
 }
 
+func (m DeployerMux) GetDeployersInverse() []Deployer {
+	inverse := m.deployers
+	for i, j := 0, len(inverse)-1; i < j; i, j = i+1, j-1 {
+		inverse[i], inverse[j] = inverse[j], inverse[i]
+	}
+	return inverse
+}
+
 func (m DeployerMux) GetAccessor() access.Accessor {
 	var accessors access.AccessorMux
 	for _, deployer := range m.deployers {
@@ -161,7 +169,9 @@ func (m DeployerMux) Dependencies() ([]string, error) {
 }
 
 func (m DeployerMux) Cleanup(ctx context.Context, w io.Writer, dryRun bool, manifestsByConfig manifest.ManifestListByConfig) error {
-	for _, deployer := range m.deployers {
+	// Reverse order of deployers for cleanup to ensure resources
+	// are removed before their definitions are removed.
+	for _, deployer := range m.GetDeployersInverse() {
 		ctx, endTrace := instrumentation.StartTrace(ctx, "Cleanup")
 		if dryRun {
 			output.Yellow.Fprintln(w, "Following resources would be deleted:")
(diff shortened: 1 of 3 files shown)


pkg/skaffold/deploy/deploy_mux.go
```diff
@@ -161,6 +161,12 @@ func (m DeployerMux) Dependencies() ([]string, error) {
 }
 
 func (m DeployerMux) Cleanup(ctx context.Context, w io.Writer, dryRun bool, manifestsByConfig manifest.ManifestListByConfig) error {
+	// Reverse order of deployers for cleanup to ensure resources
+	// are removed before their definitions are removed.
+	revDeployers := m.deployers
+	for i, j := 0, len(revDeployers)-1; i < j; i, j = i+1, j-1 {
+		revDeployers[i], revDeployers[j] = revDeployers[j], revDeployers[i]
+	}
```
tejal29, line 169: Thanks. Was wondering if there is a utility method which already does that.

chronicc: That's a good point. Searching the internet I did not find any lib that does this. I was also supprised the slice type has no internal method since this is such a common task. Searching the skaffold source code for `for i, j` and for `for .*,.*:=.*,len` does not give any results.

The golang completion on the other side provides a `slice.reverse!` method which automatically creates the for loop. Maybe that's why there is no lib or function.

I can put such a function somewhere in the skaffold source code if you wish. I need help finding the location where such a function would be placed best.

gsquared94: You can write this function here:
https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/util/util.go

chronicc: Looking into this now.

chronicc: @gsquared94 I tried adding a function to the file you suggested. However when I did this, I created a dependency lopp between util.go and deploy.go.

I also tried create a generic slice reverse function in utils but since the go.mod file says to use at least go 1.17, I could not use generics.

In the end I added it to DeployMux since this is the only place using this function for now.

gsquared94:
@chronicc can you sign the cla?

> New Contributors: Update this check after signing the CLA by clicking [here](https://cla.developers.google.com/github/rescan?pr=GoogleContainerTools%2Fskaffold%2Fpull%2F7925&key=%01%D8%8E%A4%B2%94%F5%15%25%C8%F7%05%B0W%9A%2An9D%8A+&expiry=2022-11-26).

tejal29 (approved):

chronicc:
Sorry for pushing again. I've fixed the linting erros and added a file to the gitignore for asdf.

Issue #7934, opened 2022-10-11
Skaffold v2 skaffold init command failing


emmanuelbaah:
### Expected behavior
I would expect that the following command would work with skaffold v2 since it works with skaffold v1 - `skaffold init --filename /var/folders/25/4ys18_8d4_s33c213nb0dk5r00v4vj/T/skaffold5537756292734259627.tmp --force --artifact='{"builder":"Docker","payload":{"path":"Dockerfile"},"image":"gcr.io/baah-cloud-clode/java-cloud-run-skaffold-v2-sandbox"}' --XXenableJibInit --skip-deploy`
<img width="1746" alt="image" src="https://user-images.githubusercontent.com/16374713/195140007-c33f5006-2b5c-4c14-b01c-82fb508228e6.png">

### Actual behavior
Instead I get the following error - `one or more valid Kubernetes manifests are required to run skaffold`
<img width="1740" alt="image" src="https://user-images.githubusercontent.com/16374713/195139581-fbd66b89-3886-4535-b5d2-324b4c0ebf46.png">

Even after adding the `` flag i get the same error
<img width="1735" alt="image" src="https://user-images.githubusercontent.com/16374713/195139700-2d258bab-c433-4fe9-822e-ae196906092c.png">

### Information

- Skaffold version: v2.0.0-beta3-17-gc50ca8ab5
- Operating system: MacOS
- Installed via: `curl -Lo skaffold https://storage.googleapis.com/skaffold/builds/latest/v2/skaffold-darwin-amd64 && \
sudo install skaffold /usr/local/bin/`
- Contents of skaffold.yaml: N/A

```yaml

```

### Steps to reproduce the behavior

1. Run the above command with skaffold v1
2. Then repeat with skaffold v2

### Related information
This was discovered while testing Skaffold v2 with Cloud Code IntelliJ for the `Cloud Run: Deploy` run configuration.

emmanuelbaah:
Is there an equivalent skaffold init command i can run in v2? ie maybe the syntax changed unbeknownst to me. Or is the behavior expected that the exact command should work across v1 and v2?

gsquared94:
> Is there an equivalent skaffold init command i can run in v2? ie maybe the syntax changed unbeknownst to me. Or is the behavior expected that the exact command should work across v1 and v2?

The exact command should work. I think I have narrowed down the issue to a regression in Skaffold, working on a fix.

PR #7936, opened 2022-10-11
fix: allow `init` without `render` and `deploy`

gsquared94:
Fixes: #7934 
Specifying `--skip-deploy` should skip creating both `render` and `deploy` stanzas in the `skaffold.yaml` file


aaron-prindle (approved):
LGTM!

PR #7961, opened 2022-10-20
chore: add generic `util.Ptr` function

gsquared94:
Replace `util.StringPtr`, `util.IntPtr` and `util.BoolPtr` with a generic `util.Ptr` 🥹
```
func Ptr[T any](t T) *T {
	o := t
	return &o
}
```


aaron-prindle (approved):
LGTM.  It's glorius! 🥇

PR #7968, opened 2022-10-21
docs: Add a copy of the buildpacks node tutorial without a skaffold.yaml

maggieneterval:
**Description**

Creates a copy of the buildpacks node tutorial without a `skaffold.yaml`, so that it can be used in a tutorial that has users run `skaffold init` to generate the `skaffold.yaml`.

@aaron-prindle LMK what you think!

### DIFF
--- a/examples/buildpacks-node-tutorial/.gitignore
+++ b/examples/buildpacks-node-tutorial/.gitignore
@@ -1,0 +1,1 @@
+package-lock.json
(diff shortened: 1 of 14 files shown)


aaron-prindle (approved):
LGTM!  I think if you can rebase and resubmit this, we can get all the tests passing and then merge

aaron-prindle:
I think there is one small error from our CI/CD:
```
Every code sample that is in ./examples should also be in ./integration/examples
The following are in examples but not in integration/examples:
buildpacks-node-tutorial
FAILED hack/check-samples.sh
```
Can you copy this sample to integration/examples?

PR #7986, opened 2022-10-26
fix: stderr of "helm template" not printed

michelesr:
Without stderr is hard to tell why template rendering fails


aaron-prindle (approved):
LGTM, thanks for this!

Issue #6637, opened 2021-09-27
Provide a gcr.io/k8s-skaffold/skaffold:slim image


zinderic:
### Expected behavior

Small skaffold docker image containing mostly the binary

### Actual behavior

Very big skaffold docker image which contains also its build env like the Go compiler

### Information

Check this image for example:

```
gcr.io/k8s-skaffold/skaffold@sha256:aab3c02ba37e1489669543e3d3d6511bce00ce329f5cec4a3d8b50971e8ce72e
```

It's very big, seems like the multistage build is not being used for the creation of this image which is at the time of writing tagged as latest.

### Steps to reproduce the behavior

```
docker pull gcr.io/k8s-skaffold/skaffold@sha256:aab3c02ba37e1489669543e3d3d6511bce00ce329f5cec4a3d8b50971e8ce72e
docker run -ti --rm gcr.io/k8s-skaffold/skaffold@sha256:aab3c02ba37e1489669543e3d3d6511bce00ce329f5cec4a3d8b50971e8ce72e bash
du -sh /*
```

Here's the output that I got:

```
root@a1e09d3928dc:/skaffold# du -sh /*
2.7M	/THIRD_PARTY_NOTICES
4.8M	/bin
4.0K	/boot
0	/dev
2.4M	/etc
450M	/google-cloud-sdk
4.0K	/home
13M	/lib
4.0K	/lib64
4.0K	/media
4.0K	/mnt
4.0K	/opt
du: cannot access '/proc/11/task/11/fd/3': No such file or directory
du: cannot access '/proc/11/task/11/fdinfo/3': No such file or directory
du: cannot access '/proc/11/fd/3': No such file or directory
du: cannot access '/proc/11/fdinfo/3': No such file or directory
0	/proc
155M	/root
12K	/run
3.8M	/sbin
4.0K	/skaffold
4.0K	/srv
0	/sys
8.0K	/tmp
1.6G	/usr
432M	/var
root@a1e09d3928dc:/skaffold#
```

briandealwis:
@zinderic you can easily build your own image and install [the latest version of Skaffold](https://storage.googleapis.com/skaffold/releases/latest/skaffold-linux-amd64).  You could even use nixery.dev (`nixery.dev/shell/skaffold/docker`).

The `gcr.io/k8s-skaffold/skaffold` image is very much a "batteries included" image.  It's heavily used by Cloud Build customers, not only for building images but also for managing deployments.

[...]

PR #7996, opened 2022-10-27
feat: new dockerfiles and cloudbuild config for slim image

renzodavid9:
Fixes: #6637

**Description**
This PR adds new Dockerfiles and a `cloudbuild-slim.yaml` configuration to trigger the creation of a skaffold slim image with every push to main branch.

The slim image is a copy of the `lts` version which has less dependencies that the `edge` version.


aaron-prindle (approved):
LGTM!

PR #7997, opened 2022-10-27
feat: Add support for Cloud Run services to stream logs

qwerjkl112:
**Description**
Add log streaming support for cloud run services

log streaming will automatically be triggered on skaffold if it is running in dev, debug, run --tail modes

**Follow-up Work**
Migrate gcloud command from alpha to beta once log streaming command is published to gcloud beta.

### DIFF
--- a/pkg/skaffold/deploy/cloudrun/accessor_test.go
+++ b/pkg/skaffold/deploy/cloudrun/accessor_test.go
@@ -39,6 +39,9 @@ func (t *testAccessConfig) PortForwardOptions() config.PortForwardOptions {
 func (t *testAccessConfig) Mode() config.RunMode {
 	return config.RunModes.Run
 }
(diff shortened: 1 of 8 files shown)


pkg/skaffold/deploy/cloudrun/accessor_test.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
gsquared94, line 8: you can add a new line after line 15. That'll fix the linting issue with this boilerplate.

Issue #7979, opened 2022-10-25
Skaffold v2.0.0 has an issue where error log messages are output when no error occurs


aaron-prindle:
Currently on a successful deployment, Skaffold will always output error log messages of the form:
```
time="2022-10-24T14:34:56Z" level=error subtask=1 task=Render
time="2022-10-24T14:34:56Z" level=error subtask=2 task=Render
time="2022-10-24T14:34:57Z" level=error subtask=4 task=Render
```
These messages do not actually state what the error was that occurred and are a bit confusing as the deploy succeeds.  These messages should  either be more verbose if they are actual errors or should not be output at all if they are not actual errors

renzodavid9:
Hey @aaron-prindle, do you have the `skaffold.yaml` file you used for this?

I found this https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/render/renderer/helm/helm.go#L154 where we log an error but the buffer could be empty, so we get the empty error message. Not sure if this is the one related with what you found. Thanks!

notnmeyer:
looks like a fix is being worked on already, here's a very basic skaffold.yaml i can consistently replicate with,

```
apiVersion: skaffold/v3
kind: Config
metadata:
  name: my-app
build:
  artifacts:
    - image: registry/my-app
deploy:
  helm:
    releases:
      - name: my-app
        namespace: default
        chartPath: k8s
        valuesFiles:
          - k8s/values.yaml
```

```
➜ skaffold render -i registry/my-app:whatever
ERRO[0003]                                               subtask=0 task=Render
...
```

PR #8005, opened 2022-10-28
fix: not print error message if it is empty

renzodavid9:
Fixes: #7979

**Description**
With this PR we check first if an error message has a value before printing it so we don't get empty errors

**Follow-up Work**
Check other places where this could be happening


ericzzzzzzz (approved):

aaron-prindle (approved):
LGTM, manually confirmed this is working now using `examples/helm-deployment`

Issue #7992, opened 2022-10-26
Images are not getting loaded on Kind Cluster


zeyaddeeb:
### Expected behavior

the deployment is live and port is forwarded

### Actual behavior

deployment always fails

When I inspect the images on the kind cluster it never gets loaded: 
```bash
docker exec -it $(kind get clusters | head -1)-control-plane crictl images
```

### Information

- Skaffold version: v2.0.0 
- Operating system:  macOS 13.0 (22A380)  - Darwin 22.1.0 
- Installed via: Homebrew 
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta20
kind: Config

build:
  artifacts:
    - image: apps-web
      docker:
        dockerfile: apps/web/Dockerfile
      context: .
  local:
    push: false

profiles:
  - name: dev
    build:
      local:
        useBuildkit: true
      artifacts:
        - image: apps-web
          docker:
            dockerfile: apps/web/Dockerfile
          context: .

    deploy:
      kubeContext: kind-local-dev
      helm:
        releases:
          - name: apps-web
            chartPath: deployments/helm/web
            upgradeOnChange: true

    portForward:
      - resourceType: deployment
        resourceName: apps-web
        port: 3000
        localPort: 3000
```

### Steps to reproduce the behavior

1. a clonable repository with the sample skaffold project
2. `skaffold dev --profile dev `

selfsk:
I did some preliminary poking around, seems like problem is coming from [`imagesToLoad()`](https://github.com/GoogleContainerTools/skaffold/blob/aa0bf348b1cb3151356715612ac3db5a63a0d4df/pkg/skaffold/kubernetes/loader/load.go#L63) function which returns empty list, instead of list of produced artifacts.

```
func imagesToLoad(localImages, deployerImages, images []graph.Artifact) []graph.Artifact {
	local := map[string]bool{}
	for _, image := range localImages {
		local[docker.SanitizeImageName(image.ImageName)] = true
	}

	tracked := map[string]bool{}
	for _, image := range deployerImages {
		if local[image.ImageName] {
			tracked[image.ImageName] = true
		}
	}

	var res []graph.Artifact
	for _, image := range images {
		if tracked[docker.SanitizeImageName(image.ImageName)] {
			res = append(res, image)
		}
	}
	return res
}
```

[...]

PR #8007, opened 2022-10-31
fix: container image push to local cluster

selfsk:
Starting with version 2.x, `skaffold` isn't pushing container images to kind cluster(s), as result - deployment fail due to lack of container image. Issue #7992

Verified changes on 3 [`examples`](/examples):
 * [getting-started](/examples/getting-started)
 * [kustomize](/examples/kustomize)
 * [helm-deployment](/examples/helm-deployment)

Run `skaffold --kube-context=kind-kind dev` in each to validate container images being pushed with this fix.

Fixes: #7992 

**Description**

It appeared that problem with skaffold not loading container images to kind cluster is due to the way how skaffold constructs the list of images that supposed to be pushed to cluster. Function [`imagesToLoad`](https://github.com/GoogleContainerTools/skaffold/blob/aa0bf348b1cb3151356715612ac3db5a63a0d4df/pkg/skaffold/kubernetes/loader/load.go#L63) iterating through different lists, and those lists kind of depend on each other.

Result list check for docker images that being produced, but in case of [`kubectl`](https://github.com/GoogleContainerTools/skaffold/blob/aa0bf348b1cb3151356715612ac3db5a63a0d4df/pkg/skaffold/deploy/kubectl/kubectl.go#L233) and [`helm`](https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/deploy/helm/helm.go#L240) deployers list of produced artifact hasn't been passed to this function at all, as result `skaffold` is treating as there were no images to push.

### DIFF
--- a/pkg/skaffold/deploy/helm/helm.go
+++ b/pkg/skaffold/deploy/helm/helm.go
@@ -237,7 +237,7 @@ func (h *Deployer) Deploy(ctx context.Context, out io.Writer, builds []graph.Art
 	}
 
 	childCtx, endTrace := instrumentation.StartTrace(ctx, "Deploy_LoadImages")
-	if err := h.imageLoader.LoadImages(childCtx, out, h.localImages, nil, builds); err != nil {
+	if err := h.imageLoader.LoadImages(childCtx, out, h.localImages, h.originalImages, builds); err != nil {
 		endTrace(instrumentation.TraceEndError(err))
 		return err
 	}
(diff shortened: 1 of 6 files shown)


KoltesDigital:
I suggest renaming this PR to `fix: container image push to *local* cluster`, as I have the same issue with a local k3d cluster.

aaron-prindle (approved):
LGTM, thanks for the PR here @selfsk!!

arikkfir:
FYI this still happens with Skaffold v2.0.4 (which I think should include #8007 iianm)
Only adding `--load-images=true` to the `deploy` command fixes it for now.

PR #8018, opened 2022-11-02
fix: properly add RemoteManifests support to skaffold v2

### LINKED ISSUE #7990: remoteManifests no longer works in v2.0.0
### Expected behavior
Remote manifests loads manifest from specified kubernetes resource

### Actual behavior
Manifests are looked for locally

### Information
- Skaffold version: working: v1.37.2, not working: v2.0.0
- Operating system: Ubuntu 20.04
- Installed via: skaffold.dev
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta8
kind: Config
build:
  artifacts:
  - image: test
deploy:
  kubectl:
    remoteManifests:
      - deploy/test
```

1. clone https://github.com/rishka/skaffold-manifests-bug
2. run `kubectl apply -f deploy.yaml`
3. run `skaffold run`
4. results in `nothing to deploy` in v2, correctly applies in v1

aaron-prindle:
fixes #7990 
 
MERGE_AFTER= #8034

### DIFF
--- a/docs-v2/content/en/schemas/v4beta1.json
+++ b/docs-v2/content/en/schemas/v4beta1.json
@@ -1,0 +1,4191 @@
+{
+  "type": "object",
+  "anyOf": [
+    {
+      "$ref": "#/definitions/SkaffoldConfig"
+    }
+  ],
+  "$schema": "http://json-schema-org/draft-07/schema#",
+  "definitions": {
+    "Activation": {
+      "properties": {
+        "command": {
+          "type": "string",
+          "description": "a Skaffold command for which the profile is auto-activated.",
+          "x-intellij-html-description": "a Skaffold command for which the profile is auto-activated.",
+          "examples": [
+            "dev"
+          ]
+        },
+        "env": {
+          "type": "string",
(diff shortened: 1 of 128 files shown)


ericzzzzzzz:
https://github.com/GoogleContainerTools/skaffold/issues/7990#issuecomment-1299651204 
We have some logic to deduce url manifests, not remoteManifests. 
In v1 remoteManifests are resources in cluster, not urls where store manifests.  implementation is here https://github.com/GoogleContainerTools/skaffold/blob/v1/pkg/skaffold/deploy/kubectl/kubectl.go#L379-L397 

we call ``kubectl get --namespace ${namespace} ${resource name} -o=yaml`` to retrieve the manifests from clusters.

So I don't think the current change is the right fix, we're currently just not supporting remoteManifests in v2, this change may help supporting reading get manifests from kubenetes resources in remote clusters, but the implementation is hard. 

like the reproduction project : https://github.com/rishka/skaffold-manifests-bug/blob/main/skaffold.yaml 

the path can just be ``deploy/test``, it's hard to tell if is a local path or some remote k8s resource. 

Also, we will need to specify cluster context for this kind of kubectl call, in v1 this readRemoteManifests is a function for deployer, we have a place to specify that kube context in config, but in v2 we cannot specify this kube context under ``rawK8``, we may need a separate config to support reading remote manifests from clusters. Then it's clear that the path is for k8s resources, maybe we don't even have to name it as path, also we're able to specify context. What do you think?

aaron-prindle:
@ericzzzzzzz thanks for the clarification, I misunderstood the `remoteManifests` field.  I will close this PR and add your comment to the original issue

Issue #7993, opened 2022-10-27
Config dependencies order does not behave like what the doc claims


duongcongtoaimanabie:
Using `skaffold dev --module cfg` still output invalid deployment order

### Expected behavior

According to the doc: https://skaffold.dev/docs/design/config/#configuration-dependencies, if cg1 is dependent on cf2, cfg2 should be deployed before cg1. But this behavior is not consistent.

I setup a sample example (multiple-file configs) in this branch (checked out from release 2.0.0): https://github.com/duongcongtoaimanabie/skaffold/tree/debug-cfg-dependencies

I also added some logs to check the order of deployment in that debugging branch
### Actual behavior

The deploy order usually is sk4->sk3->sk2->sk1, but not consistent, for example:

Log incuding the debug:

```
toaiduong@LAP00349:~/project/skaffold/examples/getting-started$ skaffold dev --module sk1 --iterative-status-check
getting deployer
deployer kubectl sk3 sk3
deployer kubectl sk2 sk2
deployer kubectl sk1 sk1
deployer kubectl sk4 sk4
Listing files to watch...
 - skaffold-example
Generating tags...
 - skaffold-example -> localhost:5001/skaffold-example:v2.0.0-1-gbfa0e2317
Checking cache...
 - skaffold-example: Found Remotely
Tags used in deployment:
 - skaffold-example -> localhost:5001/skaffold-example:v2.0.0-1-gbfa0e2317@sha256:8fa1bb80d45e64d7f694991ad90f046eea27fee53b79fc0c490f12c56f26b50c
Starting deploy...
deploying for sk3
 - pod/getting-started created
Waiting for deployments to stabilize...
 - pods is ready.
Deployments stabilized in 3.052 seconds
deploying for sk2
 - pod/getting-started unchanged
Waiting for deployments to stabilize...
Deployments stabilized in 4.462862ms
deploying for sk1
 - pod/getting-started unchanged
Waiting for deployments to stabilize...
Deployments stabilized in 3.311561ms
deploying for sk4
 - pod/getting-started unchanged

```

### Information

- Skaffold version: v2.0.0
- skaffold config: in the above branch

### Steps to reproduce the behavior

1. checkout from this commit https://github.com/GoogleContainerTools/skaffold/commit/c3440b8ca29b72682f3acf94961c018ee06099fc
2. build skaffold using the new source `go build ./cmd/skaffold`
3. `cd examples/getting-started`
4. `skaffold dev --module sk1`

[...]

PR #8028, opened 2022-11-02
fix: preserve configs order when creating renderers and deployers

renzodavid9:
Fixes: #7993

**Description**
This PR fixes the order in which every deployer and renderer should be created and invoked according to the dependencies created between different modules. So, if `module1` depends on/requires `module2`, skaffold should deploy first `module2` and then `module1`

Also, we now mock the `GetNamespace` from [deploy/helm/helm_test.go](https://github.com/GoogleContainerTools/skaffold/pull/8028/files#diff-00324bab5a92b65d9af9c88f4eb9db3aa71cad52423ca109a0b0b0ac2026ff70) to not depend on kubectl output to get the namespace.

**How to test**
From the code example created in this [commit](https://github.com/duongcongtoaimanabie/skaffold/commit/bfa0e231713ed042c253bb8838561ef8edd83611), run `skaffold dev`and you should be able to see the modules deployed in the proper dependency order all the time:

```
Starting deploy...
deploying for sk4
 - pod/getting-started created
deploying for sk3
 - pod/getting-started unchanged
deploying for sk2
 - pod/getting-started unchanged
deploying for sk1
 - pod/getting-started unchanged
```

### DIFF
--- a/integration/deploy_test.go
+++ b/integration/deploy_test.go
@@ -19,6 +19,7 @@ package integration
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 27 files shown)


aaron-prindle (approved):
LGTM.  If possible can you add a new integration/ test that test the config ordering? (Eg a test project that has a multi-config project that if ordered incorrect produces a different skaffold.yaml and then expect taht skaffold.yaml).  This way we can guarantee this property in the future

PR #8038, opened 2022-11-04
Avoid nil context error

justinsb:
If this code path is followed, we end up with a nil context.


aaron-prindle (approved):
LGTM, thanks for this!

Issue #8033, opened 2022-11-03
Cleanup is not working when helm deployer is used with skaffold dev


ericzzzzzzz:
### Expected behavior
 - skaffold should clean up running resource if helm deployer is used with skaffold dev

### Actual behavior
 - releases are still running. 

### Information

- Impacted Skaffold version: 2.*.*
 - in this example https://github.com/GoogleContainerTools/skaffold/tree/main/examples/helm-deployment , replace skaffold.yaml with the following ones 
 ```yaml
apiVersion: skaffold/v3
kind: Config
build:
  artifacts:
  - image: skaffold-helm
deploy:
  helm:
    releases:
    - name: skaffold-helm
      chartPath: charts
```
 - run ```skaffold dev```

the cause is that when only helm deployer being used,  https://github.com/GoogleContainerTools/skaffold/blob/main/cmd/skaffold/app/cmd/dev.go#L74 value is always empty, so cleanup never gets triggered.

### Potential solution

can we remove this condition and always run cleanup?

ericzzzzzzz:
We should also create an example for ``helm-deployment``, the current one is actually for helm render, not for helm-deployment,

PR #8040, opened 2022-11-04
fix: cleanup not called when using helm deployer alone

ericzzzzzzz:
Fixes: #8033  

**Description**
 -  deployment can happen without user specifies manifest then `deployedManifests` will be empty, so in this pr,  change dev command to always call cleanup.

### DIFF
--- a/cmd/skaffold/app/cmd/dev.go
+++ b/cmd/skaffold/app/cmd/dev.go
@@ -71,11 +71,9 @@ func runDev(ctx context.Context, out io.Writer) error {
 				err := r.Dev(ctx, out, artifacts)
 				manifestListByConfig := r.DeployManifests()
 
-				if manifestListByConfig.String() != "" {
-					cleanup = func() {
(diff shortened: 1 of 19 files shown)


examples/helm-deployment/skaffold.yaml
```diff
@@ -3,7 +3,7 @@ kind: Config
 build:
   artifacts:
   - image: skaffold-helm
-manifests:
+deploy:
```
tejal29, line 6: Is this required as part of this change? Is this for legacy helm deployer? If that is the case, should we add another example.

ericzzzzzzz: This is not required for the fix, but we should fix this. This actually uses helm render and kubectl deployer, not as the name suggested, this also was the cause that I didn't catch this issue when fixing cleanup/ delete issues. 

Added another example

tejal29 (approved):

tejal29:
just a nit. I think can use profiles to override the `skaffold.yaml` to use helm deployer. 

Thanks
Tejal

Issue #8023, opened 2022-11-02
Log duplication when using dependencies


Jul13nT:
### Expected behavior

Skaffold should display each application log exactly one time.

### Actual behavior

Since Skaffold 2.0, log tailing with multiple configuration dependencies is duplicating log lines. For instance, if I deploy 5 dependencies, each log line will be displayed 5 times.

### Information

- Skaffold version: 2.0.1
- Operating system: Ubuntu 22.04
- Installed via: skaffold.dev

### Steps to reproduce the behavior

1. Use [Skaffold configuration local dependencies example](https://skaffold.dev/docs/tutorials/config-dependencies/)
2. Follow the tutorial, without --tail=false `skaffold dev --port-forward` (the default skaffold version is 1.39.2). The result is the following logs:

![logs-skaffold-1](https://user-images.githubusercontent.com/2983260/199521043-5c331789-393d-4696-863c-5467f9cdcfc5.png)

3. Stop Skaffold, download and install version 2.0.1. Run `skaffold dev --port-forward` again. The result is the following logs:

![logs-skaffold-2](https://user-images.githubusercontent.com/2983260/199522685-c943a401-9940-471b-acde-98beb443f602.png)

This project has 10 dependencies so each log in printed 10 times.

PR #8042, opened 2022-11-04
fix: log duplication when using dependencies

ericzzzzzzz:
Fixes: #8023  

**Description**
 - The cause was that each deployer was adding all builds to podSelectors, the logger of a deployer will stream log from all pods deployed with runner built artifacts.
 - Ideally, each deployer should only stream from pods that it deploys artifacts to. 
 - The workable implementation in v1 only added shared artifacts between deployerArtifacts set and runner builds artifacts set to podSelector for streaming logs, port-forwarding, etc. With artifact image name as identifier for an artifact to filter common elements, but due to the difference between v1 and v2 we cannot use imageName as the key for the map to filter those, the reason was added in the comment, let me know if you have any concern for using artifact.tag as key to filter. Alternatively, we can just track deployed artifacts regardless they're built from runner or not. This should fix log duplication issue, but it will also stream logs for remote artifacts. 

**Manual Testing** 
 - use this project [ examples/multi-config-microservices](https://github.com/GoogleContainerTools/skaffold/tree/main/examples/multi-config-microservices)
 - run skaffold dev 
 - expect to one log from each pod
Screen Shot 2022-11-07 at 10 16 01 AM
 - without this pr, the log would be like below.

Screen Shot 2022-11-07 at 10 17 16 AM

### DIFF
--- a/integration/util.go
+++ b/integration/util.go
@@ -430,6 +430,7 @@ func WaitForLogs(t *testing.T, out io.Reader, firstMessage string, moreMessages 
 		case <-timer.C:
 			t.Fatal("timeout")
 		case line := <-lines:
+			t.Logf("Expecting %s, received %s \n", message, line)
 			if strings.Contains(line, message) {
 				if current >= len(moreMessages) {
 					return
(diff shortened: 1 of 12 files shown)


aaron-prindle (approved):
LGTM

PR #8044, opened 2022-11-04
chore(doc): note that filesync works for debug

briandealwis:
**Description**
File sync does work with `debug`, but is disabled by default.

We should document some of the limitations of `sync` since some users assume that it magically rebuilds their applications.  Some context in https://chat.google.com/room/AAAAuTCR_1s/srcFk70Ok_A


aaron-prindle (approved):
LGTM!

PR #8050, opened 2022-11-07
fix(sec): upgrade runc version v1.0.2 -> v1.1.2

aaron-prindle:


go.mod
```diff
@@ -95,7 +95,7 @@ require (
 	google.golang.org/grpc v1.49.0
 	google.golang.org/protobuf v1.28.1
 	gopkg.in/yaml.v2 v2.4.0
-	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
+	gopkg.in/yaml.v3 v3.0.0
```
gsquared94, line 98: I remember there was a reason we were pinned to a specific version of `yaml.v3` although I don't recall what it was.

gsquared94: aah here it is:
https://github.com/GoogleContainerTools/skaffold/blob/9802b0e1c285cfd79759f1bad466ad3ebbe5d703/go.mod#L11-L14

aaron-prindle: Hmm, I can hold off on this then and try to understand more why we pin atm.  There was a security bot that suggested this change which is the context here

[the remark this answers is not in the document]
aaron-prindle: Ah good point, I might separate these changes out as it seems there is some issue with bumping `gopkg.in/yaml.v3`.  There was a security bot that suggested this change which is the context here

aaron-prindle:
Changed this PR to only update runc

go.mod
aaron-prindle: Changed the PR to only update `runc` due to comments here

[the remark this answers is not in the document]
aaron-prindle: Changed this pr to only update `runc`

ericzzzzzzz (approved):

PR #8056, opened 2022-11-08
refactor: use exclude directives instead of replace directives for pinning

bcmills:
**Description**

The 'replace' directive is very heavy-handed: it replaces nodes in the dependency graph with different source code. That distorts their apparent transitive dependencies, and makes a module that uses 'exclude' directives much more difficult for downstream consumers to use even if they aren't using the parts of the module that depend on the replaced versions.

In contrast, the 'exclude' directive knocks out specific version requirements from the module graph without otherwise affecting the dependencies of any remaining module in the graph.

'exclude' directives also make it clearer why particular pins are necessary: normally only the 'require' directive is needed to select desired versions (since they are not upgraded implicitly), so having 'exclude' blocks for specific conflicting requirements more clearly shows which of the transitive dependencies are believed to be spurious.

(See https://go.dev/ref/mod#go-mod-file-exclude.)

**Follow-up Work**

In the long term, the code should be updated to work with dependencies selected using only `require` directives, which respect the requirements of dependencies, rather than pinning versions below those requirements.

**Testing**

The following is a diff of `go list -m all` before and after this change.
Note that the former `replace` targets are now the MVS-selected versions.

One additional module is now visible in the module graph (`github.com/docker/spdystream`), but it has no effect on the resulting artifacts because it does not provide any transitively-imported package.

```diff
~/src/github.com/GoogleContainerTools/skaffold$ diff before.txt after.txt
312a313
> github.com/docker/spdystream v0.0.0-20160310174837-449fdfce4d96
499c500
< github.com/googleapis/gnostic v0.4.1 => github.com/googleapis/gnostic v0.4.1
---
> github.com/googleapis/gnostic v0.4.1
702c703
< github.com/opencontainers/image-spec v1.0.3-0.20220114050600-8b9d41f48198 => github.com/opencontainers/image-spec v1.0.2-0.20210730191737-8e42a01fb1b7
---
> github.com/opencontainers/image-spec v1.0.2-0.20210730191737-8e42a01fb1b7
842c843
< github.com/tektoncd/pipeline v0.5.1-0.20190731183258-9d7e37e85bf8 => github.com/tektoncd/pipeline v0.5.1-0.20190731183258-9d7e37e85bf8
---
> github.com/tektoncd/pipeline v0.5.1-0.20190731183258-9d7e37e85bf8
1005c1006
< gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b => gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c
---
> gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c
1012c1013
< k8s.io/apimachinery v0.23.5 => k8s.io/apimachinery v0.21.6
---
> k8s.io/apimachinery v0.21.6

```


bcmills:
I think your `codecov` tool is broken: this PR makes no changes to the actual Go code in use, so the coverage delta should be 0%, not 4% — that seems like an awfully large noise floor. 😅

aaron-prindle (approved):
LGTM!  Thanks for this, I have created #8094 to track the suggested `Follow-up Work`.  We will be more tactful  in how we use `excludes` + `requires` going forward.

PR #8066, opened 2022-11-10
fix: correct issues with current upgrade logic for artifactOverrides with helm imageStrategy

aaron-prindle:
Initially artifactsOverrides and imageStrategy were removed in skaffold v2 as we now use skaffold as a helm --post-renderer.  In our upgrade.go for v2beta29 we had some logic added that would migrate people that used those fields to setValues and setValueTemplates.  The upgrade.go logic here though did not work properly for multi-config projects as setValueTemplates has unfriendly behaviour when used with multi-config images as noted in https://github.com/GoogleContainerTools/skaffold/issues/5317.  As such this new PR changes upgrade.go to use all setValueTemplates replacements and adds a helm multi-config example as an integration test.  This docs PR below might help to explain this a bit as well:
https://github.com/GoogleContainerTools/skaffold/pull/8093

### DIFF
--- a/pkg/skaffold/kubernetes/manifest/images.go
+++ b/pkg/skaffold/kubernetes/manifest/images.go
@@ -19,6 +19,7 @@ package manifest
 import (
 	"context"
 	"strconv"
+	"strings"
 
 	apimachinery "k8s.io/apimachinery/pkg/runtime/schema"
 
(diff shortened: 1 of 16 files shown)


pkg/skaffold/kubernetes/manifest/images.go
```diff
@@ -200,6 +200,12 @@ func (r *imageReplacer) Visit(gk apimachinery.GroupKind, navpath string, o map[s
 		log.Entry(context.TODO()).Debugf("Couldn't parse image [%s]: %s", image, err.Error())
 		return false
 	}
+	// this is a hack to properly support `imageStrategy=helm+explicitRegistry` from Skaffold v1.X.X
```
gsquared94, line 203: can you add a bit more detail about this hack? Like with an example of `image` value, and why we're setting `BaseName` to `Repo`.

gsquared94: is it possible that any other regular scenario will conflict with this conditional?

gsquared94 (approved):
lgtm, but please add some more text about the hack logic.

pkg/skaffold/kubernetes/manifest/images.go
aaron-prindle: I don't believe there is realistically any regular scenario where this will conflict.  I have tried with both artifact/artifactOverrides names of the format:
```
  artifacts:
  - image: gcr.io/aprindle-test-cluster/skaffold-helm
```
```
  artifacts:
  - image: skaffold-helm
```
and this works for both cases

aaron-prindle: Updated the comment to have more explanation:

```
	// this is a hack to properly support `imageStrategy=helm+explicitRegistry` from Skaffold v1.X.X which has the form:
	// image: "{{.Values.image.registry}}/{{.Values.image.repository}}:{{.Values.image.tag}}"
	// works by looking for intermediate helm replacement of the form image: <artifactName>/<artifactName>:<artifactName>
	// and treating that as just <artifact> by modifying the parsed representation.
```

PR #8080, opened 2022-11-12
chore: reduce gcp integration test time

ericzzzzzzz:
Fixes: #8091  

**Description**
 - change partition strategy to be the hash value of a test case name module the number of partition, this makes test distribute more balanced, and it's easier to maintain
 - refactor tests need to build cross-platform/multi-platform images to use example/multi-pods project, as cross-platform build is very slow when running, ``go build ...``, we should make project source code as simple as possible when using cross-platform/multi-platform builds.
 - comment out TestApplyFailure -- Config resource connector, the reason is in the comment. 

**Result**
 - cutting testing time from 50 mins to 25 mins. 

**Follow up work**
 - None-gpc test partitioning is also based on hash value now, but it's based on a test suit not per test case, as the unbalancing issue is not that bad, we probably also want to partition test based on test case, but I'd like to do it in a separate pr.
 - Gpc ci machine image change, may also reduce the test environment setup time.

### DIFF
--- a/integration/apply_test.go
+++ b/integration/apply_test.go
@@ -35,7 +35,7 @@ func TestDiagnoseRenderApply(t *testing.T) {
 		MarkIntegrationTest(t.T, NeedsGcp)
 		ns, client := SetupNamespace(t.T)
 
-		out := skaffold.Diagnose("--yaml-only").InDir("examples/multi-config-microservices").RunOrFailOutput(t.T)
+		out := skaffold.Diagnose("--yaml-only").InDir("testdata/multi-config-pods").RunOrFailOutput(t.T)
 
 		tmpDir := testutil.NewTempDir(t.T)
 		tmpDir.Chdir()
@@ -47,11 +47,10 @@ func TestDiagnoseRenderApply(t *testing.T) {
 
 		skaffold.Apply("render.yaml", "-f", "skaffold-diagnose.yaml").InNs(ns.Name).RunOrFail(t.T)
 
-		depApp := client.GetDeployment("leeroy-app")
-		t.CheckNotNil(depApp)
-
-		depWeb := client.GetDeployment("leeroy-web")
-		t.CheckNotNil(depWeb)
+		pod1 := client.GetPod("module1")
+		t.CheckNotNil(pod1)
+		pod2 := client.GetPod("module2")
+		t.CheckNotNil(pod2)
 	})
 }
 
@@ -91,10 +90,13 @@ func TestApplyStatusCheckFailure(t *testing.T) {
 			description: "status check for statefulset resources",
 			profile:     "statefulset",
 		},
-		{
-			description: "status check for config connector resources",
-			profile:     "configconnector",
-		},
+		//{
+		// config connector resource status doesn't distinguish between resource that is making progress towards reconciling from one that is doomed.
+		// This is tracked in b/187759279 internally. The test currently passes due to status check timeout, it's not what we want to test, hence
+		// commenting this out at the moment.
+		// description: "status check for config connector resources",
+		// profile:     "configconnector",
+		// },
 		{
 			description: "status check for standalone pods",
 			profile:     "pod",
--- a/integration/build_test.go
+++ b/integration/build_test.go
@@ -153,8 +153,6 @@ func TestBuildWithWithPlatform(t *testing.T) {
 }
 
 func TestBuildWithMultiPlatforms(t *testing.T) {
-	MarkIntegrationTest(t, NeedsGcp)
-
 	tests := []struct {
 		description       string
 		dir               string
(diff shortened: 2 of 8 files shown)


integration/apply_test.go
```diff
@@ -35,7 +35,7 @@ func TestDiagnoseRenderApply(t *testing.T) {
 		MarkIntegrationTest(t.T, NeedsGcp)
 		ns, client := SetupNamespace(t.T)
 
-		out := skaffold.Diagnose("--yaml-only").InDir("examples/multi-config-microservices").RunOrFailOutput(t.T)
+		out := skaffold.Diagnose("--yaml-only").InDir("testdata/multi-config-pods").RunOrFailOutput(t.T)
```
aaron-prindle, line 38: nit:  in the PR description you mention:
```
refactor tests builds cross-platform/multi-platform examples/microservice project to use example/multi-pods, as cross-platform build is very slow when running, go build ..., we should make project source code as simple as possible when using cross-platform/multi-platform builds.
```
example/multi-config-microservices doesn't actually use `cross-platform/multi-platform` builds, it only use multiple skaffold configs/modules.  I'm not sure this needs to be changed, was this actually slowing things down?  This doesn't actually use cross-platform/multi-platform builds

integration/deploy_test.go
```diff
@@ -38,43 +38,21 @@ func TestBuildDeploy(t *testing.T) {
 
 	ns, client := SetupNamespace(t)
 
-	outputBytes := skaffold.Build("--quiet", "--platform=linux/arm64,linux/amd64").InDir("examples/microservices").InNs(ns.Name).RunOrFailOutput(t)
+	outputBytes := skaffold.Build("--quiet", "--platform=linux/arm64,linux/amd64").InDir("examples/nodejs").InNs(ns.Name).RunOrFailOutput(t)
```
aaron-prindle, line 41: nit:  in the PR description you mention:
```
refactor tests builds cross-platform/multi-platform examples/microservice project to use example/multi-pods, as cross-platform build is very slow when running, go build ..., we should make project source code as simple as possible when using cross-platform/multi-platform builds.
```
example/multimicroservices doesn't actually use `cross-platform/multi-platform` builds.  I'm not sure this needs to be changed, was this actually slowing things down?  This doesn't actually use cross-platform/multi-platform builds

aaron-prindle:
Awesome!! `kokoro` ran in ~22 mins with this 🥇.  Left some nits related to if certain tests actually needed to be modified. I think it might be the case that changing the partitions + removing the `kpt live apply` + `configconnector` test might be the only necessary changes for this improvement

integration/apply_test.go
gsquared94: @aaron-prindle it should be alright since the multi-platform values are being provided on line 45 using the flag `--platform=linux/amd64,linux/arm64`

integration/build_test.go
```diff
@@ -207,6 +205,7 @@ func TestExpectedBuildFailures(t *testing.T) {
 	}
 	for _, test := range tests {
 		t.Run(test.description, func(t *testing.T) {
+			MarkIntegrationTest(t, NeedsGcp)
```
gsquared94, line 208: what effect does this change have?

ericzzzzzzz: > what effect does this change have?


This change uses the hash value of an individual test case name to assign partition instead of the whole test, that is 

``TestExpectedBuildFailures/jib_is_too_old`` vs  ``TestExpectedBuildFailures``, we can have many test cases per test, then we will have better granularity,  for example, a test suite with 10 test case, partitioning by test suite name, will assign all the 10 test cases in this test suite into one partition, partitioning by test case name(this change) will distribute those 10 test case among all partitions we have, the workload is more likely to be balanced

ericzzzzzzz: Though we currently only have one test case in this test, that doesn't make much difference for partitioning.. but we may have more test cases in the future..

integration/apply_test.go
ericzzzzzzz: @aaron-prindle We're running integration tests against 3 types of cluster, and our CI machine is amd architecture, so when testing against arm cluster, the machine has to build cross-platform image with binfmt interpreter, this can be 10X slow compared to building image with native platform. Reducing the project source code complexity helps a little bit in build speed. You can test this by running `` skaffold build -p gcb --platform="linux/arm64"  --default-repo=${}``, in skaffold/examples/microservice folder  it usually takes 10 minutes to finish. 

Then replace everything in leery-app/app.go , leery-web/app.go with very simple code like 
```go
package main

import (
	"fmt"
)

func main() {
	for {
		fmt.Println("...")
	}
}
```
should be able to reduce the building time to 3 mins, still very slow : (

aaron-prindle: +1 to the original logic/changes, sorry for the misunderstanding and thanks for the explanation!

aaron-prindle (approved):
LGTM!  Excited for this 💯

integration/build_test.go
gsquared94: got it. That makes perfect sense.

gsquared94 (approved):
lgtm

Issue #8045, opened 2022-11-05
Override protocols doesn't work with helm deployment


manslaughter03:
### Expected behavior

When i use args `--protocols=pydevd` in skaffold debug command, i expect that command and args of deployment will be override with pydevd.

### Actual behavior

Deployment use default debugger protocol: debugpy.

### Information

- Skaffold version: v2.0.1
- Operating system: Nixos 22.05
- Installed via: skaffold.dev
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v3
kind: Config
build:
  artifacts:
  - image: app
manifests:
  helm:
    releases:
    - name: app
      chartPath: charts
      setValues:
        imagePullPolicy: Never
deploy:
  helm: {}
```

### Steps to reproduce the behavior

1. git clone https://github.com/manslaughter03/skaffold-protocol-issue && cd skaffold-protocol-issue
2. `skaffold debug --protocols=pydevd -vdebug`
3. You will notice that the deployment use debugpy instead of pydevd

```yaml
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    app: app
    skaffold.dev/run-id: 64227770-b551-4221-9021-bfe01ae888cb
  name: app
spec:
  replicas: 1
  selector:
    matchLabels:
      app: app
  strategy: {}
  template:
    metadata:
      annotations:
        debug.cloud.google.com/config: '{"app":{"artifact":"app","runtime":"python","workingDir":"/usr/src","ports":{"dap":5678}}}'
      creationTimestamp: null
      labels:
        app: app
        skaffold.dev/run-id: 64227770-b551-4221-9021-bfe01ae888cb
    spec:
      containers:
      - args:
        - -m
        - app
        command:
        - /dbg/python/launcher
        - --mode
        - debugpy
        - --port
        - "5678"
        - --
        - python
        image: app:36f76829cf227397213ff89a4c5517a230ba1223724b59a67feb98713bcce8e6
        imagePullPolicy: Never
        name: app
        ports:
        - containerPort: 5678
          name: dap
        resources: {}
        volumeMounts:
        - mountPath: /dbg
          name: debugging-support-files
      initContainers:
      - image: gcr.io/k8s-skaffold/skaffold-debug-support/python
        name: install-python-debug-support
        resources: {}
        volumeMounts:
        - mountPath: /dbg
          name: debugging-support-files
      volumes:
      - emptyDir: {}
        name: debugging-support-files
```

[...]

PR #8083, opened 2022-11-13
fix: override protocols argument pass to helm post-renderer

manslaughter03:
Fixes: #8045 

**Description**
I add override protocols args on filter cmd use by helm post render.


aaron-prindle (approved):
LGTM,  thanks for repro project and fix here @manslaughter03!

PR #8096, opened 2022-11-15
feat: implement cleanup for Cloud Run Jobs

bskaplan:
**Description**
Implements Cleanup for Cloud Run Jobs, as well as support cleaning up multiple resources.

This is still fire-and-forget deleting so if Cloud Run encounters any errors asynchronously during the deletion, this won't catch them. Also, if multiple errors occur, it only returns the first one.

Support cleanup for Cloud Run Jobs.

Improve error handling, add integration test.


aaron-prindle (approved):
LGTM

PR #8106, opened 2022-11-17
chore: repartition integration test

ericzzzzzzz:
Fixes: #8115
 - re-partition non-gcp integration tests based on the hash value of each test case name module the number of partition
 - The non-gcp integration tests execution time should be under 30 mins now. 
 - removed unused code.


aaron-prindle (approved):
LGTM, nice!

Issue #8041, opened 2022-11-04
Skaffold build create kpt-pipeline folder even if not needed


cmeury:
### Expected behavior

No kpt folder is created.

### Actual behavior

`.kpt-pipeline` is created in the local directory.

### Information

- Skaffold version: 2.0.1
- Operating system: macOS 13
- Installed via: Homebrew
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v3
kind: Config
metadata:
  name: foo
```

### Steps to reproduce the behavior

1. Change into temporary directory
2. Drop above `skaffold.yaml` into it
3. `skaffold build`
4. `ls -la`

PR #8117, opened 2022-11-18
fix: new condition to create hydrate-dir only if a kpt renderer or deployer

renzodavid9:
Fixes: #8041

**Description**
This PR adds an extra condition before creating the `hydration-dir`, so it is created only when kpt is used (as renderer, deployer, or both).

### DIFF
--- a/pkg/skaffold/deploy/util/util.go
+++ b/pkg/skaffold/deploy/util/util.go
@@ -18,7 +18,6 @@ package util
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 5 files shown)


pkg/skaffold/deploy/util/util.go
```diff
@@ -164,37 +167,6 @@ func GroupVersionResource(disco discovery.DiscoveryInterface, gvk schema.GroupVe
 	return false, schema.GroupVersionResource{}, fmt.Errorf("could not find resource for %s", gvk.String())
 }
 
-func GetManifestsFromHydrationDir(ctx context.Context, opts config.SkaffoldOptions) (manifest.ManifestList, error) {
```
aaron-prindle, line 167: nice!

Issue #8116, opened 2022-11-18
deployments are not cleaned up in some integration tests with default namespace


ericzzzzzzz:
### Information

 - Some tests running skaffold.dev in background process leave their deployments after tests finish
 - Those leftover deployments can contaminate others tests. 
 - This only happens when tests deploy something to defaultNamesapce, as for tests using non-default namespace, the setupNameSpace method has already taken care of cleanup work.
 - In those tests with deployments left in the cluster, they currently use 
 ``defer skaffold.Delete().InDir("xxx").RunBackground(t)`` for cleaning up, this is not working,  the background process will be killed too soon before any cleanup work is done. We should replace it with ``defer skaffold.Delete().InDir("xxx").Run(t)`` for proper cleanup.

PR #8118, opened 2022-11-18
fix: integration-test cleanup

ericzzzzzzz:
Fixes: #8116 

**Description**
 - change cleanup(delete) call in test to be run in the same process as test itself instead of running in background process, to avoid background process being killed before skaffold delete gets called. 
 - refactor file sync test to use different pod for each test to avoid test contamination.


ericzzzzzzz:
hmm... This is a little awkward, TestDevSyncDefaultNamespace whole test was passing due to file-sync pod in the first test case not being deleted. We're seeing the second test case failing, this is caused by the second test case the background dev process sees the terminating status of the pod from the first test case and before that 	``client.WaitForPodsReady("test-file-sync")`` has been already executed, it saw a running pod, that's due to kubeclt delete command returns after it sends request to the cluster, it doesn't wait for the deletion. Perhaps, we shouldn't share the test resource for this whole test.

aaron-prindle (approved):
Nice, LGTM!

PR #8135, opened 2022-11-22
fix: resolve issue where skaffold would panic when StatusCheck was not set

aaron-prindle:


pkg/skaffold/build/app.tar
```diff
@@ -0,0 +1 @@
+bazel:app
```
renzodavid9, line 1: I think we shouldn't commit this file?

renzodavid9 (approved):

pkg/skaffold/build/app.tar
aaron-prindle: Sorry, this was added accidentally.  Removed now, thanks!

Issue #8121, opened 2022-11-19
Applying further patches fails after an `add` patch


ali-idrizi:
There seems to be an issue with modifying fields after running an `add` patch operation. For example, if we add an extra build for a specific profile, then we want to change a field inside `build`, it fails.

### Actual behavior

In this skaffold.yaml:

```yaml
apiVersion: skaffold/v4beta1
kind: Config
build:
  artifacts:
    - image: node-example
      context: backend
profiles:
  - name: prod
    patches:
      - op: add
        path: /build/artifacts/0
        value:
          image: node-2
          context: backend
      - op: replace
        path: /build/artifacts/1/image
        value: test
```

The second operation fails with the error:

```
parsing skaffold config: failed to apply profiles to config "" defined in file "/skaffold/examples/nodejs/skaffold.yaml": applying profile "prod": invalid path: /build/artifacts/1/image. There's an issue with one of the profiles defined in config "" in file "/skaffold/examples/nodejs/skaffold.yaml"; refer to the documentation on how to author valid profiles: https://skaffold.dev/docs/environment/profiles/.
```

When using index 0 (`/build/artifacts/0/image`, so change the image of the added build) it fails with `panic: runtime error: invalid memory address or nil pointer dereference`.

### Expected behavior

Using either index should succeed, it does when trying on this online tool: https://json-schema-validator.herokuapp.com/jsonpatch.jsp

When patching `/build/artifacts/1/image`, it should produce:

```yaml
apiVersion: skaffold/v4beta1
kind: Config
build:
  artifacts:
    # First patch adds the build
    - image: node-2 
      context: backend
    # Second path changes the image to `test`
    - image: test
      context: backend
```

When patching `/build/artifacts/0/image`, it should produce:

```yaml
apiVersion: skaffold/v4beta1
kind: Config
build:
  artifacts:
    # First patch adds the build
    # Second patch changes the image to test
    - image: test
      context: backend
    - image: node-example
      context: backend
```

### Information

- Skaffold version: 2.0.2
- Operating system: Ubuntu 22.04
- Installed via: skaffold.dev
- Contents of skaffold.yaml: see above

### Steps to reproduce the behavior

- Simply run `skaffold run -p prod` on the yaml file above.

PR #8138, opened 2022-11-22
chore: fix profile patch validation against original config

ericzzzzzzz:
Fixes: #8121 

**Description**

 - profile patch validation is now against an on-the-fly buffer from a previous patched operation instead of a fixed buffer
 - This enables user to use patched stanza for their following patches 
 - This is not breaking some stuff, it will also fail fast compared to the original implementation, as the original implementation will apply these patches one by one anyway, if applying patches would fail eventually, it's better to fail it when we just do try patch.


aaron-prindle (approved):
LGTM, nice!

PR #8145, opened 2022-11-23
chore: fix github workflow linux unit diagnostics tests run command

ericzzzzzzz:
**Description**
 - The run diagnostics tests command is not correct in linux unit test workflow, update to the right one.


aaron-prindle (approved):
LGTM!

Issue #7989, opened 2022-10-26
Helm hooks are not using the correct image tag


michelesr:
### Expected behavior

When specifying an image override for a helm deployment, hooks should use the correct image tag just the rest of the resources.

### Actual behavior

When specifying an image override for a helm deployment, the tag for hooks isn't changed, just the repository.

### Information

- Skaffold version: v2.0.0-5-gbef0268b8
- Operating system: Arch Linux (but is not relevant)
- Installed via: built from source
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta29
kind: Config
build:
  artifacts:
    - image: localhost/nginx
deploy:
  kubeContext: minikube
  helm:
    releases:
      - name: test
        namespace: test
        skipBuildDependencies: true
        createNamespace: true
        chartPath: ./chart
        artifactOverrides:
          image.repository: localhost/nginx
```

### Steps to reproduce the behavior

1.  clone https://github.com/michelesr/skaffold-test-repo
2. `skaffold dev`
3. observe the pods with `kubectl get pods -n test`

The foo pod (which represents the hook) fails with `ImagePullBackOff`.

Observe image tag set for the deployment

```
$ kubectl get deployment -n test test-chart -o=jsonpath={.spec.template.spec.containers[0].image}
localhost/nginx:0d5bf3701e4847eaa6faff0ff8c8e7c26b34dc32fc8bd50b7aceb04b174d81e6
```
Observe image tag for the hook

```
$ kubectl get job -n test foo -o=jsonpath={.spec.template.spec.containers[0].image}
localhost/nginx:1.16.0
```

### Additional information

In skaffold v1.39.2 it was possible to set the image strategy, giving more flexibility on how image tags propagated to helm. Right now, what happens is that the value is set to the repository only (without tag), and then `--post-renderer` replaces the image references in the template directly.

v1 (with default image strategy):

```
$ helm get values test -n test
USER-SUPPLIED VALUES:
image:
  repository: localhost/nginx:0d5bf3701e4847eaa6faff0ff8c8e7c26b34dc32fc8bd50b7aceb04b174d81e6
```

v2:

```
$ helm get values test -n test
USER-SUPPLIED VALUES:
image:
  repository: localhost/nginx
```
[...]

Issue #7967, opened 2022-10-21
.Values.imageKey is rendered differently in helm template


letmutx:
### Expected behavior
Using `{{ .Values.imageKey }}` resolves to the same values when used everywhere in the template with helm release.

```
$ skaffold run
$ kubectl get deployment test -o yaml | rg -aF binaries
          value: binaries:2001d978bc5729c0bfc3c5f9a1a3499c7afdeb3f45e4c785711f58d644a3aa2e
        image: binaries:2001d978bc5729c0bfc3c5f9a1a3499c7afdeb3f45e4c785711f58d644a3aa2e 
```

### Actual behavior
With v2.0.0, it when using the `{{ .Values.imageKey }}` in a k8s deployment template in `env` block and setting it in container `image` results in different values. 

```
$ skaffold run
$ kubectl get deployment test -o yaml | rg -aF binaries
          value: binaries
        image: binaries:2001d978bc5729c0bfc3c5f9a1a3499c7afdeb3f45e4c785711f58d644a3aa2e 
```

This was not the case in v1.34.1 

### Information

- Skaffold version: `v2.0.0`
- Operating system: Linux, Pop!_OS 22.04
- Installed via: skaffold.dev
- Contents of skaffold.yaml: Attached a minimal reproducible example to the PR

### Steps to reproduce the behavior

1. [sample skaffold project](https://github.com/letmutx/skaffold-bug-report/)
2. Steps here: https://github.com/letmutx/skaffold-bug-report/blob/main/README.md

logs: https://github.com/letmutx/skaffold-bug-report/blob/main/run-2.0.0.log

iosifnicolae2:
We're encountering something similar in https://github.com/GoogleContainerTools/skaffold/issues/8008

sanketsudake:
Facing similar issue https://github.com/fission/fission, unable to migrate to scaffold v2 because of the bug.

aaron-prindle:
[...]

Issue #6207, opened 2021-07-14
Helm template variable naming strategy neither intuitive nor stable


gsquared94:
Related: https://github.com/GoogleContainerTools/skaffold/issues/6173
Skaffold creates environment variables `IMAGE_NAME`, `IMAGE_TAG`, `IMAGE_REPO`, `DIGEST`, and `IMAGE` that can be templated in various places in the `helm` deploy stanza.

If there are multiple images being built then it names them `IMAGE_NAME`, `IMAGE_NAME2`, `IMAGE_NAME3`, etc.
These numbers are not assigned stably from one run to the next. Also it becomes quickly unintuitive for projects using multiple configs which artifact corresponds to which number.

gsquared94:
One approach could be to prefix template variables with the `artifact name`.
For the `skaffold.yaml` file:

```yaml
apiVersion: skaffold/vX
kind: Config
build:
  artifacts:
    - image: foo
    - image: bar
deploy:
  helm:
    releases:
    - name:bar
      setValueTemplates:
        tag: "{{foo.IMAGE_TAG}}"
        tag: "{{foo.DIGEST}}"
        tag: "{{foo.IMAGE_TAG}}"
```

This change for `helm` would need to be made here:
https://github.com/GoogleContainerTools/skaffold/blob/27271dfcaab56c38ecf41f4b7158101d44c454b6/pkg/skaffold/deploy/helm/args.go#L67-L76

kadern0:
Hi @gsquared94, I've been looking into this issue and I like your idea although there is something I might be missing. You refer to `artifact name` but then you use `artifact image`  in the example (I guess they're synonyms in this context, right?) . According to the documentation and also my use cases, the artifact image doesn't look like "foo" or "bar" but something more like:
```
- image:  gcr.io/k8s-skaffold/example
```
This would make the template variables too long and unusable (I think). It could be useful having a proper `artifact name` for this purpose.

iosifnicolae2:
The same problems is affecting https://github.com/GoogleContainerTools/skaffold/issues/8008

aaron-prindle:
This was recently added, the format is:
IMAGE_[NAME|TAG|REPO|DOMAIN|REPO_NO_DOMAIN]_<artifact-name>

PR #8147, opened 2022-11-23
fix: support alternative env var naming using support env variable + artifact-name (vs env variable + index)

aaron-prindle:
fixes: #7967 

fixes: #6207 

fixes: #7989 

This PR:
- adds .\<artifact-name\>.IMAGE_[REPO|TAG|DOMAIN|REPO_NO_DOMAIN] vars for each artifact
- updates v2beta29 upgrade.go to fix code to use `setValuesTemplates` and the above env vars vs `setValues`

Major reason using `setValueTemplates` is better here is that:
- This replacement is done with `helm` vs the skaffold post-renderer so that user expectations are in line with how helm replaces values vs how to skaffold postrenderer replaces values which is non-obvious/unexpected given the field names `setValues`, etc.
- no longer requires hacks in our post-renderer to get the helm replacements to work as expected for various image strategies

### DIFF
--- a/integration/testdata/helm-multi-config/skaffold/app1/skaffold.yml
+++ b/integration/testdata/helm-multi-config/skaffold/app1/skaffold.yml
@@ -14,6 +14,6 @@ manifests:
       - chartPath: ../../charts/app
         createNamespace: true
         name: app1
-        setValues:
-          image.repository: app1
-          image.tag: app1
+        setValueTemplates:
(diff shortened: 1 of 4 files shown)


gsquared94 (approved):
lgtm

michelesr:
https://github.com/GoogleContainerTools/skaffold/issues/7989#issuecomment-1331426303

Issue #7850, opened 2022-09-08
Skaffold setValues is getting missing helm values.


m18unet:
Skaffold `setValues` is getting missing helm values. 

Instead of `setValues`, when I save the relevant values in the `values.yml` file and use `valuesFiles`, there is no problem and the rendering is successful. 

I guess `setValues` doesn't recognize nested arrays. Please review the example below. 

Why the `ingress.tls[0].hosts` doesn't exist?

### skaffold.yaml

```yaml
apiVersion: skaffold/v2beta29
kind: Config
build:
  local:
    push: false
  tagPolicy:
    sha256: {}
  artifacts:
  - image: example
    jib: {}
    sync:
      auto: false
deploy:
  helm:
    releases:
    - name: example
      chartPath: backend-app
      artifactOverrides:
        image: example
      imageStrategy:
        helm: {}
      setValues:
        ingress:
          enabled: true
          className: nginx
          hosts:
            - host: example.minikube
              paths:
                - path: /
                  pathType: ImplementationSpecific
          tls:
            - secretName: example-tls
              hosts:
                - example.minikube
```

### skaffold run

```bash
skaffold run -v TRACE
```

```bash
# Output
[...]
[...]
[...]
DEBU[0106] Running command: [
helm --kube-context minikube install example backend-app --set-string image.repository=example,image.tag=6ad72230060e482fef963b295c0422e8d2f967183aeaca0229838daa7a1308c3 --set ingress.className=nginx --set --set ingress.enabled=true --set ingress.hosts[0].host=example.minikube --set ingress.hosts[0].paths[0].path=/ --set ingress.hosts[0].paths[0].pathType=ImplementationSpecific --set ingress.tls[0].secretName=example-tls]  subtask=0 task=Deploy
[...]
[...]
[...]
```

I want to ask again, Why the `ingress.tls[0].hosts` doesn't exist?

### Ingress Manifest

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: example
spec:
  ingressClassName: nginx
  tls:
    - hosts:
      secretName: example-tls
  rules:
    - host: "example.minikube"
      http:
        paths:
          - path: /
            pathType: ImplementationSpecific
            backend:
              service:
                name: example
                port:
                  number: 80
```

briandealwis:
I think this code needs to handle the `!ok` case:
[...]

PR #8152, opened 2022-11-23
fix: change to make buildFlatMap support arrays as YAML values

renzodavid9:
Fixes: #7850 

**Description**
Change in `FlatMap` unmarshal logic to support YAML files with arrays that have simple values (not maps) and nested arrays.


aaron-prindle (approved):
LGTM!

PR #8178, opened 2022-11-30
fix: allow transformer values to contain ':'

hampus77:
Fixes: #8177

**Description**
Changed the key/value split from strings.Split to strings.Index to find only the first ':' deliminator. Everything after that is treated as a value


ericzzzzzzz:
Thank for your contribution! LGTM!

ericzzzzzzz (approved):

PR #8187, opened 2022-12-03
chore: bump schema version to v2beta4

aaron-prindle:


renzodavid9 (approved):

aaron-prindle:
Good catch RE:

> overall looks good to me. just a nit for the commit message .. bump schema version to v2beta4 should be bump schema version to v4beta2

you can change the commit message on merge so I'll do that here to fix thiis

Issue #8194, opened 2022-12-07
schema latest label is not correct in v2 doc site


ericzzzzzzz:
when open this site, https://skaffold.dev/docs/references/yaml/   

expected:
 - v4beta1 schema is shown in the main frame with latest label, it should be the same as the one in sidebar
actual : 
 - v3 schema is shown in the main frame,

cause, there is a version param need to be updated in doc site config.

PR #8195, opened 2022-12-07
chore: fix latest label misplaced issue

ericzzzzzzz:
Fixes: #8194  

 - run ``make preview-docs-v2`` to verify, the schema should be v4beta2 in the main frame of  ``docs/references/yaml/``  page

### DIFF
--- a/.github/workflows/integration-linux.yml
+++ b/.github/workflows/integration-linux.yml
@@ -12,7 +12,7 @@ concurrency:
 jobs:
 
   build:
-    name: PR integration tests (linux) 
+    name: PR integration tests (linux)
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-linux.yml
```diff
@@ -12,7 +12,7 @@ concurrency:
 jobs:
 
   build:
-    name: PR integration tests (linux)
+    name: PR integration tests (linux)
```
aaron-prindle, line 15: nit: I don't think this change is necessary

aaron-prindle (approved):
LGTM! Left one nit - not a big issue though

.github/workflows/integration-linux.yml
ericzzzzzzz: ... I remember I reverted this.. doing it now

Issue #6784, opened 2021-10-31
Do not use KRM setters in rendering 


yuwenma:
### Information

- Skaffold version: v2.0.0 

Do no use setters KRM functions. this is not compatible with our general non-templating hydration design. 

https://github.com/GoogleContainerTools/skaffold/blob/v2/pkg/skaffold/render/transform/transform.go#L40

https://github.com/GoogleContainerTools/skaffold/blob/v2/pkg/skaffold/render/transform/transform.go#L44

yuwenma:
The rationale is explained here https://github.com/GoogleContainerTools/kpt/issues/3131

ericzzzzzzz:
After discussion, we decided to keep the setters in skaffold to support parameterization.  So closing this out.

PR #8206, opened 2022-12-08
chore: remove setters from kpt transformer allowlist

ericzzzzzzz:
Fixes: #6784  

**Description**


aaron-prindle:
Holding off on merging this as I am not sure if we want to restrict the kpt capabilities as I think users and Cloud Deploy team see an opportunity to use setters for easily modifying "rawYaml" or setting up a chain of renderers

aaron-prindle:
In working with BETS team and Cloud Deploy team they both want kpt's `apply-setters` functionality currently (but there might be other ways to deliver it).  Will leave this open but likely shouldn't be merged as this functionality might be desireable

Issue #8183, opened 2022-12-01
Skaffold not properly replacing helm env variables under `defaultNamespace` (at least HOST variable)


aaron-prindle:
Info from - https://github.com/GoogleContainerTools/skaffold/issues/7995#issuecomment-1334279390

A user attempted to use [templating](https://skaffold.dev/docs/environment/templating/) of an environment variable in deploy.kubectl.defaultNamespace found that while the following example was working in 2.0.1, it does not work with a HEAD build (v2.0.2-31-g4b9badab3). {{.HOST}} is not replaced with the value of the HOST environment variable. #8129 might be the culprit here
```yamll
deploy:
  kubectl:
    hooks:
      before:
        - host:
            command: ["sh", "-c", "./predeploy.sh"]
    defaultNamespace: 'test-{{.HOST}}'
```

aaron-prindle:
Could be related to:
https://github.com/GoogleContainerTools/skaffold/issues/8105

RB14:
Also happens with `{{.USER}}`  variable. I guess it will happen with just any of them...

PR #8222, opened 2022-12-12
fix: expand namespace with env variables

renzodavid9:
Fixes: #8183

**Description**
Now, before returning the namespace from the runcontext, we replace the env variables if any is present.


aaron-prindle (approved):
Awesome!! LGTM

Issue #8233, opened 2022-12-14
Make it possible to disable enforced namespace when rendering with `kubectl`


bsideup:
### Expected behavior

A flag that would allow skipping setting the namespace while rendering the manifest.

### Actual behavior

The namespace is always set, either to the current namespace from `kubectl` or the default value.

### Information

- Skaffold version: Latest (2.0.3)
- Operating system: Any
- Installed via: skaffold.dev

### Context

We want to use Skaffold to render a manifest that will later be installed into multiple namespaces (GitOps style).

Right now, Skaffold will force the namespace to be set **at the render stage**, making it impossible to use the rendered manifest without some post-processing that would remove the enforced namespace.

bsideup:
Submitted a draft PR to demonstrate one of possible solutions:
https://github.com/GoogleContainerTools/skaffold/pull/8234

aaron-prindle:
Related - https://github.com/GoogleContainerTools/skaffold/issues/8186

bsideup:
@aaron-prindle ohhh, thanks for sharing! I did not expect it to be a regression, but a feature. If previously it wasn't the behavior, I will be more than happy if, by default, Skaffold will not force any namespace unless specified 👍

bsideup:
Update: after reading the linked issue more carefully, I realized that `--offline` flag does exactly what I needed and skips adding the namespace (in the latest release).

I consider my use case solved, assuming that #8186 will be addressed before the next release, but I will leave the issue open in case the new behavior is desired and the flag will still be needed. Thank you!

aaron-prindle:
Going to close this, tracking changing the default behaviour to be the same as `--offline=true` in this issue: 
https://github.com/GoogleContainerTools/skaffold/issues/8318

Issue #8186, opened 2022-12-02
[regression] Skaffold @ HEAD (in no releases) currently adds a namespace to all rendered manifests


aaron-prindle:
With #8105 merged, currently skaffold will always add a namespace to rendered yamls:
```
aprindle@aprindle-ssd ~/skaffold/examples/getting-started  [main]$ skaffold render
apiVersion: v1
kind: Pod
metadata:
  name: getting-started
  namespace: default # <--- this is set
spec:
  containers:
  - image: skaffold-example:v2.0.2-45-g130df30ae
    name: getting-started
```

This is an issue as skaffold cannot then use rendered manifests to deploy to multiple namespaces as most/all deployers will not "overwrite" a namespace set in the yaml config (will error stating such).  

Behavior should  be:
When no namespace is explicitly set, no namespace should be added to allow for deploying to multiple namespaces

aaron-prindle:
I am still seeing this, not sure if there is something wrong in my local env or this is unresolved.  Re-opening

aaron-prindle:
Going to create a new issue that better describes this problem (vs reverting this issue which mentions reverting the PR)

jleonar:
@aaron-prindle Does this mean `default` doesn't get added to all manifests where the namespace isn't specified in the yamls? My use case is sorta what you mention, we build in one environment and publish artifacts.  Then use render -> deploy in our prod environment.   This default getting added will break for us since we use manifests that don't specify a namespace.

aaron-prindle:
@jleonar can you repost your comment in the new issue tracking this:
https://github.com/GoogleContainerTools/skaffold/issues/8318

I will answer there when you post there, thanks!

PR #8234, opened 2022-12-14
Allow skipping namespace hydration on render

bsideup:
Fixes: #8233

**Description**
As per the original issue, this PR adds a `--set-namespace` flag to `render` that defaults to `true` (old behavior) but allows skipping namespace hydration.

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -116,6 +116,15 @@ var flagRegistry = []Flag{
 		DefinedOn:     []string{"dev", "run", "debug", "deploy", "render", "build", "delete", "apply"},
 	},
 	{
+		Name:          "set-namespace",
+		Usage:         "Set the namespace in the kubernetes manifest files.",
+		Value:         &opts.SetNamespace,
+		DefValue:      true,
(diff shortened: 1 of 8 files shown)


aaron-prindle:
Related - https://github.com/GoogleContainerTools/skaffold/issues/8186

aaron-prindle:
Closing this PR, I think likely changing the default behavior (and possible adding a flag like this) is the correct approach.  We have prioritized doing this and have this issue tracking it:
https://github.com/GoogleContainerTools/skaffold/issues/8318

aaron-prindle:
Sorry, in looking at this again I think this is almost exactly what is needed w/ the change being the default should be false (we are ok w/ changing the default as this current behavior seems user hostile).

aaron-prindle:
I think I might submit a new PR with very similar changes, let me know if you want to get this across the finish line instead.  Thanks for flagging this!

bsideup:
Hi @aaron-prindle! Since more work will be needed (tests, docs, etc etc), I am totally fine if you take over this work, thank you!

JFTR My original issue was solved with the `--offline` flag and I am unblocked now 👍

PR #8236, opened 2022-12-15
feat: always override `skaffold.dev/run-id` label and allow setting it via `--label` flag

gsquared94:
Fixes: #5966 

**Description**

Manually specifying the `skaffold.dev/run-id` label in any deployment manifest causes the status-checker from omitting that resource from being monitored.

The `skaffold.dev/run-id` label is a reserved label and shouldn't be manually set by the user in the kubernetes manifests directly. However, making that an error case will not be backwards-compatible for users that do not care about this behavior. This PR mandates that this label is always overwritten with the current skaffold session run-id so that all the monitors work correctly. Additionally the flag `--label` is enabled for the `skaffold apply` command so that it can be set that way
```
skaffold apply --label skaffold.dev/run-id=blah
```

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -230,7 +230,7 @@ var flagRegistry = []Flag{
 		Value:         &opts.CustomLabels,
 		DefValue:      []string{},
 		FlagAddMethod: "StringSliceVar",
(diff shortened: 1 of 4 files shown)


aaron-prindle (approved):
LGTM, thanks for digging into this

Issue #8166, opened 2022-11-28
File watching for sync is very slow on Mac OS


corporealshift:
### Expected behavior
Syncing files previously would notice changes in a few seconds (or less).
### Actual behavior
Since upgrading to version v2.0.2 it takes 15 seconds or more to notice files to sync. During this time the skaffold process consumes an entire core of CPU power (100% in activity monitor). Once the "Syncing 1 file to..." message gets posted to the console the CPU usage goes down.

### Information

- Skaffold version: v2.0.2
- Operating system: MacOS Monterey 12.6.1 
- Installed via:Homebrew
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v4beta1
kind: Config
build:
  artifacts:
  # Creating a new image requires a new ECR repo by adding a file here and following the README instructions.
  # https://github.com/fulcrumapp/eks-flux2/tree/main/infrastructure/base/fulcrum-ops/ecr
  #
  #
  #   READ BEFORE CHANGING ANYTHING IN ARTIFACTS!!!!
  #
  #
  # Changing anything here needs to consider the other profiles lower in the file that might override
  # or make assumptions on the these lists of artifacts
  - image: snip 
    sync:
      manual:
      - src: app/**
        dest: .
      - src: public/**
        dest: .
      - src: vendor/**
        dest: .
      - src: test/**
        dest: .
      - src: spec/**
        dest: .
      - src: lib/tasks/**
        dest: .
      # this one is less about syncing and more making sure we don't rebuild
      # since we don't use this path in this container
      - src: lib/assets/fulcrum-components/**
        dest: .
      hooks:
        after:
        - container:
            # This should be safe to run on every pod in chaos, but it would be better to only run on sidekiq.
            command: ["sh", "-c", "ps auxww | grep sidekiq | grep app | grep -v grep | awk '{ print $2; }' | xargs -r kill -12"]
            # Currently unsupported by sync hooks, but hopefully soon.
            #containerName: sidekiq
            #podName: rails-fulcrum-sidekiq-all*
        - host:
            command: ["sh", "-c", "echo Restarting sidekiq."]
            os: [darwin, linux]
    docker:
      dockerfile: docker/ruby/Dockerfile
      buildArgs:
        ENV: production
        BASE_VERSION: stable
        ASSETS_VERSION: "dev-{{.USER}}"
  - image: snip 
    docker:
      dockerfile: docker/components/Dockerfile
    sync:
      manual:
[...]

PR #8249, opened 2022-12-16
fix: scanning ignored files in dev session

ericzzzzzzz:
Fixes: #8166 
**Related**: #6851

**Description**
 - This fixes skaffold dev slow session start and syncing files due to skaffold scans ignored files. 
 - The root cause is that WalkWorkspace method traverses docker ignored folder which can contain too many files even the method is not performing actions on the unintended traversal result,  too many system calls for querying files can be very time consuming.  All the unnecessary visiting ignored folder should be avoided. 
 - The keepfile method was not properly updated when adding support for copying empty folder to docker, that's probably because the behavior of walk.Do is not obvious, so I added description for that method.

**Test Plan**
 - in example/nodejs project do the followings
    - create a folder under backend dir, you can name it as ``test_folder``
    - Inside the folder, run ``touch ab{1..50000}`` , ``touch abc{1..50000}`` , ``touch azz{1..50000}`` to generate files
    - edit backend/.dockerignore file to include ``test_folder``
    - run ``skaffold dev`` in the project root folder.  The dev session should start very quick and update a file to test syncing it should be fast as well. 
    - if you do the same without this change, dev session start and syncing will be slow

### DIFF
--- a/pkg/skaffold/docker/dependencies.go
+++ b/pkg/skaffold/docker/dependencies.go
@@ -252,16 +252,14 @@ func WalkWorkspace(workspace string, excludes, deps []string) (map[string]bool, 
 		absFrom := filepath.Join(workspace, dep)
 
 		keepFile := func(path string, info walk.Dirent) (bool, error) {
-			// Ignore non empty dirs
-			if info.IsDir() && !util.IsEmptyDir(path) {
-				return false, nil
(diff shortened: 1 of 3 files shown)


aaron-prindle (approved):
LGTM, nice!

Issue #8185, opened 2022-12-02
skaffold render ignores namespace in .deploy.helm.releases[].namespace when kube context has namespace defined


bendikp:
### Expected behavior
When specifying `namespace` in `.deploy.helm.releases[].namespace` I expect that skaffold passes that namespace to helm. Regardless of the namespace defined in the current kube context. This was the behavior in `v1.39`.

### Actual behavior
In `v2.0.3` the `.deploy.helm.releases[].namespace` seems to be ignored and overwritten by the namespace in the current kube context.

### Information

- Skaffold version: v2.0.3
- Operating system: Ubuntu 22.04
- Installed via: https://storage.googleapis.com/skaffold/releases/v2.0.3/skaffold-linux-amd64
- Contents of skaffold.yaml:

```yaml
---
apiVersion: skaffold/v2beta21
kind: Config
metadata:
  name: monitoring-stack
deploy:
  helm:
    releases:
      - name: prom
        repo: https://prometheus-community.github.io/helm-charts
        remoteChart: kube-prometheus-stack
        version: 37.1.0
        namespace: monitoring
```

### Steps to reproduce the behavior

1. `kubectl config set-context --current --namespace=default`
2. `curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/v1.39.4/skaffold-linux-amd64 && chmod +x skaffold && sudo mv skaffold /usr/local/bin`
3. `skaffold render -v DEBUG 2>&1 | grep -e "--namespace"`
4. `curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/v2.0.3/skaffold-linux-amd64 && chmod +x skaffold && sudo mv skaffold /usr/local/bin`
5. `skaffold render -v DEBUG 2>&1 | grep -e "--namespace"`

```shell
$ kubectl config set-context --current --namespace=default
$ curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/v1.39.4/skaffold-linux-amd64 && chmod +x skaffold && sudo mv skaffold /usr/local/bin
$ skaffold version
v1.39.4
$ skaffold render -v DEBUG 2>&1 | grep -e "--namespace"
time="2022-12-02T08:35:53+01:00" level=debug msg="Running command: [helm --kube-context k3d-k3s-default template prom kube-prometheus-stack --version 37.1.0 --namespace monitoring --repo https://prometheus-community.github.io/helm-charts]" subtask=-1 task=DevLoop
[...]

PR #8259, opened 2022-12-19
fix: use release namespace in render when specified

renzodavid9:
Fixes: #8185 

**Description**
This PR changes the method called to get the namespace that is used during rendering, from `GetNamespace` to `GetKubeNamespace`; in skaffold v1 is read from this last one.

**Context:**
- `GetNamespace` returns the namespace set in the detected cluster (even if it is the "default" namespace) when no namespace is passed with the `--namespace` flag and no namespace is set in a Kubectl deployer.
- `GetKubeNamespace` returns the namespace set with the `--namespace` flag

Changing the logic to use `GetKubeNamespace` instead of `GetNamespace` (as it was in v1) makes the helm renderer to use the value set in the release when no value is detected in the `--namespace` flag, and don't care about what value is set as default in the active cluster.

**Difference between `manifests.helm.releases.namespace` and `deploy.helm.releases.namespace`**

Both values do not interfere between each other. In the followin scenario:

```
apiVersion: skaffold/v4beta1
kind: Config
...
manifests:
  helm:
    releases:
    - name: skaffold-helm
      chartPath: charts
      namespace: helm-test-1
      setValues:
        image: skaffold-helm

deploy:
  helm:
    releases:
    - name: skaffold-helm
      chartPath: charts
      namespace: helm-test-2
      setValues:
        image: skaffold-helm
```

The output from `skaffold render` will produce two different sets of manifests:
- one for the release defined in the `manifests.helm.releases` config with `namespace: helm-test-1` namespace set
- other for the release defined in `deploy.helm.releases.namespace` config with `namespace: helm-test-2` namespace set

Behind the scene, skaffold is creating two different renderers: one for the `manifests` and other for the `deploy`, each with its corresponding configuration


integration/testdata/helm-namespace/builds.out.json
```diff
@@ -0,0 +1 @@
+{}
```
gsquared94, line 1: should you delete this file?

aaron-prindle: +1, it seems builds.out.json isn't used in the integration test -> might make sense to delete

aaron-prindle:
Can you add some more context in the description on GetNamespace vs GetKubeNamespace as well as what config values are looked at currently as we have 2 location that releases.namespaces can be set:
- in manifests.helm.releases.namespace
- in deploy.helm.releases.namespace

The original issue mentioned that deploy.* config wasn't respected in a `render` command.  If someone sets both values, what is the resultant namespace currently? (which one has priority?)

integration/testdata/helm-namespace/builds.out.json
renzodavid9: Done! I changed the test logic to add the `--build-artifacts=builds.out.json` flag conditionally and not for all the cases.

aaron-prindle (approved):
LGTM!

Issue #4158, opened 2020-05-12
skaffold dev should not fail on first failed build / deploy


balopat:
I propose that starting `skaffold dev` should start the dev loop, even if the first build or deployment fails. 

I see no additional value in keeping the first build failed. In the end what will the developer do to improve the situation and "finally start 'dev'-ing"? They will fix the errors and rerun `skaffold dev`. Why can't they do that while it's still running and get feedback faster? 

I argue that we can remove this artificial need for an extra `skaffold dev` command. If the user wants to stop `skaffold dev`-ing as they realize that it will take longer for them to fix things up, they can always Ctrl+C out of `skaffold dev`.

balopat:
I intend to work on this this week for the fixit effort - please comment if you feel strongly against it.

balopat:
More context on this: #516 
Also, failure might be due to infrastructure issues, not just application related ones, for example Docker is not installed, or the user is pointed to the wrong kubecontext, etc. 
The original motivation was around that - we didn't stop, just warned at build errors and that created all sorts of messes, but today we do stop at failed builds.

balopat:
As I'm thinking about this, maybe it is a good idea to fail the first build on _infrastructure_ errors. This would require us to start to differentiate a bit more smarter around error types - which is actually something @tejal29 started already.

balopat:
I realized that this is a bit more subtle: one issue that we'll have to resolve if we go down this route: currently skaffold assumes that everything was built & deployed once at least as a _baseline_, then every file change triggers only an _incremental_ change on top of that. 

Now, if we want to change this behavior, that means that we'll need to keep track of which artifacts hasn't been built yet. Skaffold shouldn't deploy before all artifacts are built.

dfang:
> More context on this: #516
> Also, failure might be due to infrastructure issues, not just application related ones, for example Docker is not installed, or the user is pointed to the wrong kubecontext, etc.
> The original motivation was around that - we didn't stop, just warned at build errors and that created all sorts of messes, but today we do stop at failed builds.

@balopat  there might be `ImagePullBackOff` errors too.

i've to run `skaffold dev --no-prune=true --cleanup=false`  when developing on local.

in prototype stage, i intend to keep images on local, do not push to any registry. no need to prune and cleanup

maybe just set  default value for `--no-prune` and `--cleanup` to `true`, `false`

or save these configs to  `~/.skaffold/config` like this:

```
global:
  survey:
[...]

PR #8270, opened 2022-12-21
feat: add keep-running-on-failure-implementation

ericzzzzzzz:
fixes: #4158

## Description
 - added support to enable skaffold session keep running on failure in the first dev loop with a flag
 - put registering files logic after the first devLoop as 
   - When first devloop fails, users could make some changes on their project, if we register files to monitor first, the listener will detect changes made for fixing up the first devLoop, so the callback(devloop) in the listener will get triggered again
 - added v2 event test helper functions so we can listen v2 events for our integration tests.
 - Refactor integration tests that use API trigger as now registering file to monitor  happens after the first devLoop, we need to wait the registering finishes before ``updating any files``, so file changes can be captured. 
 - Split TestDevAPITrigger into TestBuildAPITrigger and TestDeployAPITrigger, the original one test was hanging after we upgraded skaffold to go1.19, I think this test was a flake even before we upgrade to go1.19, we might have seen this failing before, but I'm not too sure about it. The cause of hanging was that the first build api trigger triggers a rebuild in devloop -- this build can take very long, then the test sends a deploy trigger to the server, Deploy intent was set to true, then the rebuild finishes and reset all Dev Intents, so Deploy intent was reset to false even devLoop ``runs``, the deploy is not happening. I think the root problem might be that we set intent immediately after the skaffold server receives the intent, we could set  the intent in skaffold session when the listener gets intent from the intent channel, that might require large change, so not doing it in this pr. 

## Test Plan

 - use examples/getting-started project, make a change on main.go file to let compile fail. 
 - run ``skaffold dev --keep-running-on-failure`` 
   - expect to see ``WARN[0008] Failed to build artifacts, please fix the error and press any key to continue.  subtask=-1 task=DevLoop`` from console
 - fix the the compile error then press any key to continue the dev session, dev should continue

### DIFF
--- a/pkg/skaffold/runner/dev.go
+++ b/pkg/skaffold/runner/dev.go
@@ -229,7 +229,7 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
 	var err error
 	bRes, err := r.Build(ctx, out, artifacts)
 	for ; err != nil && r.runCtx.Opts.KeepRunningOnFailure; bRes, err = r.Build(ctx, out, artifacts) {
-		log.Entry(ctx).Warnf("Failed to build artifacts, please fix the error and press any key to continue.")
+		log.Entry(ctx).Warnf("Failed to build artifacts: %v, please fix the error and press any key to continue.", err)
 		errT := term.WaitForKeyPress()
 		if errT != nil {
 			return errT
@@ -246,7 +246,7 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
 	if r.runCtx.IsTestPhaseActive() {
 		err = r.Test(ctx, out, bRes)
 		for ; err != nil && r.runCtx.Opts.KeepRunningOnFailure; err = r.Test(ctx, out, bRes) {
-			log.Entry(ctx).Warnf("Failed to run tests, please fix the error and press any key to continue.")
+			log.Entry(ctx).Warnf("Failed to run tests :%v, please fix the error and press any key to continue.", err)
 			errT := term.WaitForKeyPress()
 			if errT != nil {
 				return errT
@@ -269,13 +269,12 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
 	// First render
 	manifests, err := r.Render(ctx, out, r.Builds, false)
 	for ; err != nil && r.runCtx.Opts.KeepRunningOnFailure; manifests, err = r.Render(ctx, out, r.Builds, false) {
-		log.Entry(ctx).Warnf("Failed to render, please fix the error and press any key to continue.")
+		log.Entry(ctx).Warnf("Failed to render :%v, please fix the error and press any key to continue.", err)
 		errT := term.WaitForKeyPress()
 		if errT != nil {
 			return errT
 		}
 	}
-	r.deployManifests = manifests
 	if err != nil {
 		event.DevLoopFailedInPhase(r.devIteration, constants.Render, err)
 		eventV2.TaskFailed(constants.DevLoop, err)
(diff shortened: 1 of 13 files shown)


aaron-prindle:
Just tried it out, seems to work well for most cases. I noticed though that the error is not necessarily always the same as what would be output w/o the --keep-running-on-failure flag.  For example if i have a remote kubecontext:
```
aprindle@aprindle-ssd ~/skaffold/examples/getting-started  [feat-keep-running-on-failure]$ kubectl config current-context
gke_aprindle-test-cluster_us-west1-a_cluster-3

```

And I attempt `skaffold dev` (w/ no `--default-repo` flag), I get this error:
```
...
Build Failed. No push access to specified image repository. Try running with `--default-repo` flag. Otherwise start a local kubernetes cluster like `minikube`.
```

If I run this with `skaffold dev --keep-running-on-failure` I get:
```
The push refers to repository [docker.io/library/skaffold-example]
9d6349f27163: Preparing
994393dc58e7: Preparing
WARN[0004] Failed to build artifacts, please fix the error and press any key to continue.  subtask=-1 task=DevLoop
```

aaron-prindle:
also might make sense to add a new feature-reques issue around having skaffold surface the compiler error (docker error logs) if that can be grabbed from the docker build  logs.

pkg/skaffold/runner/dev.go
```diff
@@ -224,6 +225,93 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
 		"devIteration": strconv.Itoa(r.devIteration),
 	})
 
+	// First build
+	var err error
+	bRes, err := r.Build(ctx, out, artifacts)
+	for ; err != nil && r.runCtx.Opts.KeepRunningOnFailure; bRes, err = r.Build(ctx, out, artifacts) {
+		log.Entry(ctx).Warnf("Failed to build artifacts, please fix the error and press any key to continue.")
```
gsquared94, line 232: you need to print the `err` as well

pkg/skaffold/runner/dev.go
```diff
@@ -224,6 +225,93 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
...
+		event.DevLoopFailedInPhase(r.devIteration, constants.Build, err)
+		eventV2.TaskFailed(constants.DevLoop, err)
+		endTrace()
+		return fmt.Errorf("exiting dev mode because first build failed: %w", err)
+	}
+	// First test
+	if r.runCtx.IsTestPhaseActive() {
+		err = r.Test(ctx, out, bRes)
+		for ; err != nil && r.runCtx.Opts.KeepRunningOnFailure; err = r.Test(ctx, out, bRes) {
+			log.Entry(ctx).Warnf("Failed to run tests, please fix the error and press any key to continue.")
```
gsquared94, line 249: same here, the `err` message should be outputted

gsquared94:
I have a case where log tailing stops.
Repro:
1. In the `examples/getting-started` project make a typo like renaming `containers` to `containers2` in the `skaffold.yaml` file.
2. Run `skaffold dev --keep-alive-on-error`
3. Should see the message:
   ```
   WARN[0002] Failed to deploy, please fix the error and press any key to continue.  subtask=-1 task=DevLoop
   ```
4. Fix the typo in `skaffold.yaml` file. The deployment should succeed.
   ```
   Deployments stabilized in 33.811792ms
   Listing files to watch...
    - skaffold-example
   Press Ctrl+C to exit
   Watching for changes...
   ```
5. **But the log streaming hasn't started**.
6. Make some other change to retrigger a dev loop. Now after deploy the log streaming starts.
   ```
   Deployments stabilized in 3.191 seconds
   Watching for changes...
   [getting-started3] Hello world!
   [getting-started3] Hello world!
   ```

ericzzzzzzz:
> I have a case where log tailing stops. Repro:
> 
> 1. In the `examples/getting-started` project make a typo like renaming `containers` to `containers2` in the `skaffold.yaml` file.
> 2. Run `skaffold dev --keep-alive-on-error`
> 3. Should see the message:
>    ```
>    WARN[0002] Failed to deploy, please fix the error and press any key to continue.  subtask=-1 task=DevLoop
>    ```
> 4. Fix the typo in `skaffold.yaml` file. The deployment should succeed.
>    ```
>    Deployments stabilized in 33.811792ms
>    Listing files to watch...
>     - skaffold-example
>    Press Ctrl+C to exit
>    Watching for changes...
>    ```
> 5. **But the log streaming hasn't started**.
> 6. Make some other change to retrigger a dev loop. Now after deploy the log streaming starts.
>    ```
>    Deployments stabilized in 3.191 seconds
>    Watching for changes...
>    [getting-started3] Hello world!
>    [getting-started3] Hello world!
>    ```

@gsquared94 great catch! Thank you!  
This is actually not log stop tailing..  
- The cause was after fixing the manifests, Deploy is was called without detecting those changes.. so the actually didn't happen(we have a logic to compare previous deployed manifests vs current manifests we want to deploy, the deployed manifests are tracked even deploy failed), even deploy happened, it should fail as manifests changes were not updated in the session, we should probably 
  - check if error is bad request from k8s server , if so do the re-render or..
  - re-render anyway???? doesn't seem too bad..

pkg/skaffold/runner/dev.go
ericzzzzzzz: ahhh.. you're right! I'll add err message.. Thanks!!

aaron-prindle (approved):
LGTM

Issue #7589, opened 2022-06-26
Status Check throws error when deploying a statefulset with OnDelete update strategy


pehlert:
### Expected behavior

For statefulsets with the `updateStrategy` set to `OnDelete` the status check should be skipped.

### Actual behavior

Skaffold attempts to perform a status check via the `rollout status statefulset` command which is not supported.

### Information

- Skaffold version: 1.38.0
- Operating system: Linux
- Installed via: skaffold.dev
- Deployer: Helm

### Steps to reproduce the behavior

1. Deploy any helm chart that contains a statefulset with OnDelete strategy

### Logs

```
Release "foo-backend" has been upgraded. Happy Helming!
NAME: foo-backend
LAST DEPLOYED: Sun Jun 26 09:29:49 2022
NAMESPACE: foo-dev
STATUS: deployed
REVISION: 537
TEST SUITE: None
Waiting for deployments to stabilize...
 - foo-dev:deployment/tsdb is ready. [10/11 deployment(s) still pending]
 - foo-dev:statefulset/jsreport is ready. [9/11 deployment(s) still pending]
 - foo-dev:deployment/foo-backend-jaeger-collector is ready. [8/11 deployment(s) still pending]
 - foo-dev:deployment/dataingress is ready. [7/11 deployment(s) still pending]
 - foo-dev:statefulset/foo-backend-cassandra: running [kubectl --context kstg-stg --namespace foo-dev rollout status statefulset foo-backend-cassandra --namespace foo-dev --watch=false]
 - stdout: ""
 - stderr: "error: rollout status is only available for RollingUpdate strategy type\n"
 - cause: exit status 1
 - foo-dev:statefulset/foo-backend-cassandra failed. Error: running [kubectl --context kstg-stg --namespace foo-dev rollout status statefulset foo-backend-cassandra --namespace foo-dev --watch=false]
 - stdout: ""
 - stderr: "error: rollout status is only available for RollingUpdate strategy type\n"
 - cause: exit status 1.
1/11 deployment(s) failed
make: *** [Makefile:33: dev] Error 1
```

tejal29:
@aaron-prindle  Feel free to re-assign to someone else.

dtsong:
Hello, I am using Skaffold `v2.3.0` and it seems like this issue is occurring for a StatefulSet where I have the `updateStrategy` set to `OnDelete`.

From running `skaffold run -v debug`:

```
[...]

PR #8292, opened 2023-01-03
Handle StatefulSets with an OnDelete update strategy

hach-que:
Fixes: #7589


aaron-prindle:
@hach-que thanks for the PR here!  Can you change you commit message to adhere to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) format - eg:
"Handle StatefulSets with an OnDelete update strategy" -> "fix: Handle StatefulSets with an OnDelete update strategy"

gsquared94 (approved):

Issue #8276, opened 2022-12-21
Skaffold metrics upload fails when there are more than 200 records offline


gsquared94:
### Expected behavior

Metrics upload successfully when there are more than 200 records in the `~/.skaffold/metrics` file
### Actual behavior

Metrics upload fails with the error:
```
DEBU[0229] error uploading metrics: rpc error: code = InvalidArgument desc = Field timeSeries had an invalid value: A maximum of 200 TimeSeries can be written in a single request.  subtask=-1 task=DevLoop
```

PR #8294, opened 2023-01-04
fix: write maximum of 200 metrics per session

gsquared94:
Fixes: #8276 

**Description**

Per run of skaffold limit the number of metrics recorded to 200 which is the limit in a single request (https://cloud.google.com/monitoring/quotas). To do multiple requests would require a 10 second pause which will not be a good experience to wait for minutes at the end of a skaffold session.


aaron-prindle (approved):
LGTM!

PR #8309, opened 2023-01-10
feat: add new inspect namespaces list command to skaffold

### LINKED ISSUE #8191: [FR] Add flag/config feature to allow users to get renderered manifest or some information about the resulting namespace of render + deploy config
Issues tracks adding flag/config feature to allow users get a renderered manifest or some information about the resulting namespace of render + deploy config.

Option 1 - (Skaffold) Render flag that populates the Namespace configured in deploy stanza
If there is a reason not to change the default behavior for render to support this, then a flag could be introduced on skaffold render to support it.

Option 2 - (Skaffold) Dry run
Introduce the ability to dry run skaffold apply where the output is the resources that would be applied to the cluster. Cloud Deploy could then inspect the dry run resource to extract the namespace that is being used at deploy time. Could be built on top of kubectl apply [dry-run](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#apply) capability.

aaron-prindle:
fixes #8191

### DIFF
--- a/cmd/skaffold/app/cmd/inspect_namespaces.go
+++ b/cmd/skaffold/app/cmd/inspect_namespaces.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2021 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"context"
+	"errors"
+	"io"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/inspect"
+	namespaces "github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/inspect/namespaces"
+)
+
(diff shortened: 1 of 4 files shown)


cmd/skaffold/app/cmd/inspect_namespaces.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/inspect"
+	namespaces "github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/inspect/namespaces"
+)
+
+func cmdNamespaces() *cobra.Command {
+	return NewCmd("namespaces").
+		WithDescription("View skaffold test information").
```
renzodavid9, line 33: I think this is the description from `skaffold inspect tests`, do we need to change it for `skaffold inspect namespaces`?

aaron-prindle: Good catch, updated now

pkg/skaffold/inspect/namespaces/list_test.go
```diff
@@ -0,0 +1,219 @@
...
+		profiles    []string
+		module      []string
+		err         error
+		expected    string
+	}{
+		{
+			description: "print all deployment namespaces where no namespace is set in manifest(s) or deploy config",
+			manifest:    manifest,
+			expected:    `{"resourceToInfoMap":{"apps/v1, Kind=Deployment":[{"name":"leeroy-app","namespace":"default"}]}}` + "\n",
+			// expected: `` + "\n",
```
renzodavid9, line 97: Can we remove this?

renzodavid9 (approved):

pkg/skaffold/inspect/namespaces/list_test.go
aaron-prindle: Ah good catch sorry.  Removed now

Issue #8322, opened 2023-01-14
Wrong log URL is shown


shuuji3:
### Expected behavior

Output the correct log URL like `https://storage.cloud.google.com/shuuji3_cloudbuild/o/log-7df9dc55-168d-48f9-8b9b-be9a6b2040b6.txt`

### Actual behavior

I got the old URL format `https://console.cloud.google.com/m/cloudstorage/b/shuuji3_cloudbuild/o/log-7df9dc55-168d-48f9-8b9b-be9a6b2040b6.txt`. This URL format has been deprecated so I cannot see any log text and the following message is shown instead:

```
This Google Cloud Storage object download format is no longer supported. If you rely on this format, modify your download URLs to use the storage.cloud.google.com API instead.

Example:
console.cloud.google.com/m/cloudstorage/b/my-bucket/o/my-object -> storage.cloud.google.com/my-bucket/my-object

If you exported maps with Earth Engine before 2022-07-18, you can re-export your maps to correct the link or follow the instructions above.

Contact us at gcs-object-download-migration-help@google.com if you need support with migrating.
```

### Information

- Skaffold version: v2.0.3
- Operating system: macOS
- Installed via: Cloud Code
- Contents of skaffold.yaml:

```yaml
---
apiVersion: skaffold/v4beta1
kind: Config
metadata:
  name: takahe
build:
  platforms: ["linux/arm64"]
  artifacts:
    - image: gcr.io/shuuji3/takahe-dev
      context: .
      docker:
        dockerfile: docker/Dockerfile
  googleCloudBuild:
    timeout: 3600s
deploy:
  statusCheckDeadlineSeconds: 300
manifests:
  rawYaml:
    - kubernetes/configmap.yaml
    - kubernetes/postgres.yaml
    - kubernetes/secret.yaml
    - kubernetes/stator.yaml
    - kubernetes/webserver.yaml
```

### Steps to reproduce the behavior

1. a clonable repository with the sample skaffold project (sorry there is no repository)
2. `skaffold dev`:
```shell
Validating Kubernetes dependencies...
Validating image repository settings...

/Users/shuuji3/Library/Application Support/cloud-code/bin/versions/0031dbdda2711c2792ffe249194dabbb8599cded7f7b28a0d9085f5455359ec5/skaffold dev --filename skaffold.yaml --label ide=cloudcodeintellij --rpc-port 50051 --port-forward=true --wait-for-deletions-max=2m0s --status-check=true --verbosity warn
[...]

PR #8323, opened 2023-01-14
fix: use new URL format for Google Cloud Build log

shuuji3:
Fixes: #8322 
**Related**: N/A
**Merge before/after**: N/A

**Description**

As written in #8322, URL format `https://console.cloud.google.com/m/cloudstorage/b/<bucket-name>/o/<object-name>` has been deprecated so we need to use the new format `https://storage.cloud.google.com/<bucket-name>/<object-name>` instead.

**User facing changes**

This will replace the invalid deprecated log URL with the valid log URL.


aaron-prindle (approved):
LGTM!  Thanks for the fix here!

PR #7887, opened 2022-09-25
feat: context copy retry. If we add a for loop and execute the kubectlcli.Run method for copying context to kaniko pod , this makes more reliable and less prune to network failures

santinoncs:
* If we add a for loop and execute the kubectlcli.Run method for copying context to kaniko pod , this makes more reliable and less prune to network failures

PR #8329, opened 2023-01-18
feat: add timeout for copying build context on kaniko

hiro-o918:
**Related**: https://github.com/GoogleContainerTools/skaffold/pull/7887

**Description**

I found that uploading build context to kaniko pods sometimes stacks when copying build context on same conditions.
I inspected a pod on stack and found three things.
- initContainer keeps running
- `/kaniko/buildcontext` is empty
- `/tmp/complete` file is not created

I did not figure out exact reasons, but it supposed to be caused by creating tarball and uploading to pods.

Current implementation does not has timeout on creating tar, so I implemented this by passing `context.Context`.

**Problem**
This implementation assumes that the copying to pod must finish in 5 minutes, otherwise fail.
So We need to pass timeout duration if required.

All copying steps to kaniko pod must be finished in 5 minutes.

### DIFF
--- a/pkg/skaffold/build/cluster/kaniko.go
+++ b/pkg/skaffold/build/cluster/kaniko.go
@@ -112,7 +112,7 @@ func (b *Builder) buildWithKaniko(ctx context.Context, out io.Writer, workspace 
 	return docker.RemoteDigest(tag, b.cfg, nil)
 }
 
-func (b *Builder) copyKanikoBuildContext(ctx context.Context, workspace string, artifactName string, artifact *latest.KanikoArtifact, pods corev1.PodInterface, podName string) error {
+func (b *Builder) copyKanikoBuildContext(ctx context.Context, workspace string, artifactName string, artifact *latest.KanikoArtifact, podName string) error {
 	ctx, cancel := context.WithTimeout(ctx, copyTimeout)
 	defer cancel()
 	errs := make(chan error, 1)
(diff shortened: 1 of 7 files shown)


hiro-o918:
- I fixed lint error
- modify configuration to check timeout of tar_test which failed on only windows
  - But I'm not sure the reason, do you have any idea?

aaron-prindle:
Thanks for the PR here @hiro-o918!  In looking at out CI/CD tests it seems the test `TestAddFileToTarTimeout` is failing:
```
--- FAIL: TestAddFileToTarTimeout (0.01s)
    --- FAIL: TestAddFileToTarTimeout/TestAddFileToTarTimeout (0.01s)
        tar_test.go:302: expected error, but returned none
```

hiro-o918:
@aaron-prindle 
Thank you for checking this PR.
The test fails only on windows, and I do not have windows machine now.
I'll investigate the why context timeout does not work on windows but if your team knows any clues, please tell me :pray:

hiro-o918:
## Report for Windows error

I found that some file is treated as `drwxrwxrwx` then I try to figure out the reason
image

https://github.com/hiro-o918/skaffold/actions/runs/3972782894/jobs/6810980276#step:6:434

hiro-o918:
Windows error is fixed by extending the duration of sleeping to wait timeout.
I suppose that It may be caused by difference among OSs to handle go routine.

aaron-prindle (approved):
LGTM!  Thanks for the PR here @hiro-o918!

PR #8335, opened 2023-01-19
fix: add proper artifactOverrides->setValueTemplates conversion when upgrading from v2beta29

### LINKED ISSUE #8321: [regression] Skaffold v1 -> v2 helm `artifactsOverride` usage not automatically updated -> failures
We have a report where a user using Skaffold in CI/CD noted that for their helm chart which does not set a namespace, Skaffold v2 is deploying the helm chart to a different namespace than Skaffold v1 (this actually errors for them in this case).  Still waiting for more repro information and more specific project/skaffold information, will update the issue with that when it is available

aaron-prindle:
fixes #8321

### DIFF
--- a/pkg/skaffold/schema/v2beta29/upgrade.go
+++ b/pkg/skaffold/schema/v2beta29/upgrade.go
@@ -20,15 +20,20 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
 
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/output/log"
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/schema/util"
 	next "github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/schema/v3alpha1"
 	pkgutil "github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/util"
 )
+
+var artifactOverridesRegexp = regexp.MustCompile("/deploy/helm/releases/[0-9]+/artifactOverrides/image")
 
 var migrations = map[string]string{
 	"/deploy/kubectl":             "/manifests/rawYaml",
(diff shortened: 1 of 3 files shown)


pkg/skaffold/schema/v2beta29/upgrade.go
```diff
@@ -151,12 +185,112 @@ func upgradeOnePipeline(oldPipeline, newPipeline interface{}) error {
...
+						}
+
+						newsMap[k] = append(newsMap[k], next.JSONPatch{
+							Op:    newsMap[k][i].Op,
+							Path:  newPathModified + ".tag",
+							From:  newsMap[k][i].From,
+							Value: n,
+						})
+
+						if oldPL.Deploy.HelmDeploy.Releases[i].ImageStrategy.HelmConventionConfig.ExplicitRegistry {
```
renzodavid9, line 225: I think here we should use `idx` instead of `i` as index. `idx` is the index that we get from the json patch path, `i` is the current patch that we are processing, right?

pkg/skaffold/schema/v2beta29/upgrade.go
```diff
@@ -151,12 +185,112 @@ func upgradeOnePipeline(oldPipeline, newPipeline interface{}) error {
...
+							newsMap[k][i] = next.JSONPatch{
+								Op:    newsMap[k][i].Op,
+								Path:  newPathModified + ".repository",
+								From:  newsMap[k][i].From,
+								Value: n,
+							}
+						}
+					} else {
+						// is 'fqn' imageStrategy
+						// replace commonly used image name chars that are illegal helm template chars "/" & "-" with "_"
```
renzodavid9, line 269: I think this comment shouldn't be here. It is already in line 208, before `validV := pkgutil.SanitizeHelmTemplateValue(old.Value.Node.Value().(string))`

pkg/skaffold/schema/v2beta29/upgrade.go
```diff
@@ -151,12 +185,112 @@ func upgradeOnePipeline(oldPipeline, newPipeline interface{}) error {
...
+				idx, err := strconv.Atoi(strings.Split(old.Path, "/")[4])
+				if err != nil {
+					return err
+				}
+				newPathModified := strings.ReplaceAll(strings.ReplaceAll(old.Path, "/deploy/helm", "/manifests/helm"),
+					"artifactOverrides", "setValueTemplates")
+				if oldPL.Deploy.HelmDeploy != nil {
+					// replace commonly used image name chars that are illegal helm template chars "/" & "-" with "_"
+					validV := pkgutil.SanitizeHelmTemplateValue(old.Value.Node.Value().(string))
+					n := &util.YamlpatchNode{}
```
gsquared94, line 210: should this be `n := util.YamlpatchNode{}`? Otherwise you seem to be unmarshalling to a double pointer.

pkg/skaffold/schema/v2beta29/upgrade.go
aaron-prindle: Ah good catch, this was a bad copy paste that happened to work in my testing.  Thanks!

pkg/skaffold/schema/v2beta29/upgrade.go
aaron-prindle: removed

pkg/skaffold/schema/v2beta29/upgrade.go
aaron-prindle: Fixed the double pointer usage - instead changed the unmarshalling code to use `n` instead of `&n`

aaron-prindle:
PTAL, addressed comments

renzodavid9 (approved):

gsquared94 (approved):

PR #8340, opened 2023-01-19
feat: add ingore-path kaniko flag support

simonoff:
**Related**:  https://github.com/GoogleContainerTools/kaniko/issues/2164

**Description**
For local development we are using skaffold. Also, as we are from Ukraine and there could be no elictricty or no internet we need to be able to working offline. For such purposes we  have created a script which ups the kind cluster and we can continue working on.
On mac based systems we don't have any issues with kind and kaniko. 
But unfortunately on linux machines we have a similar issue as in linked kaniko issue.
So we need to add such flag to kaniko.

**User facing changes**

Ability to add an ingore paths for kaniko.

### DIFF
--- a/docs-v2/content/en/schemas/v4beta2.json
+++ b/docs-v2/content/en/schemas/v4beta2.json
@@ -2523,6 +2523,15 @@
           "x-intellij-html-description": "building outside of a container.",
           "default": "false"
         },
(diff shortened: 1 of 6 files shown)


aaron-prindle (approved):
LGTM!  Thanks for this @simonoff!

PR #8350, opened 2023-01-23
chore: remove unnecessary code comments

aaron-prindle:


pkg/skaffold/inspect/namespaces/list.go
```diff
@@ -58,10 +57,6 @@ func PrintNamespacesList(ctx context.Context, out io.Writer, manifestFile string
 		if len(resourceYAML) == 0 {
 			continue
 		}
-		// - obj is the API object (e.g., Deployment)
```
gsquared94, line 61: why are we deleting this comment?

aaron-prindle: I thought it wasn't necessary initially.  I have added back now

Issue #8245, opened 2022-12-15
Using kpt `apply-setter` on k8s manifests w/ kpt comments not working due to comment erasure after parsing


aaron-prindle:
Issue tracks current skaffold bug where using kpt `apply-setter` on k8s manifests w/ kpt comments not working due to comment erasure after parsing.  Skaffold should use kyaml/yaml-preserving-parser to generate the intermediate manifest.yaml that kpt operates on

aaron-prindle:
Related: #7954

Issue #8306, opened 2023-01-09
Skaffold should not validate kpt version


ericzzzzzzz:
### Information
 - Skaffold currently has a validation against kpt version on kpt render, deployer https://github.com/GoogleContainerTools/skaffold/blob/285c6972f5c0204d0e3261b5c5e72e34a628aad5/pkg/skaffold/schema/validation/validation.go#L104 
 - The validation was added due to a change introduced by kpt where ``kpt fn source xxx``  include meta-resources(Kptfile) content in its output
 - If skaffold deploys the output with a non-kpt deployer, the deploy attempt will fail 
 - Even we don't want deployments to fail, we shouldn't set maxVersion limit for kpt, or we may just don't use kpt beta version at all . 
   - Setting the maxVersion can prevent users from new kpt, users have to make a choice between to use skaffold and using new version kpt for new features and bug fixes. This is a bit frustrating.
   -  Also, if even users want to sacrifice the new stuff in a higher version of kpt, downgrading tool usually is not very pleasant experience
 - We should remove the validation for kpt version and remove extra Kptfile content in the output manifest on skaffold end.

ericzzzzzzz:
-  Kpt has a deprecated flag to toggle whether including Kptfile content into output for ``kpt fn source ..`` command, this is deprecated and not working with ``false``

ericzzzzzzz:
Let's include "upgrading kpt version in skaffold image" into the issue scope as well .

Issue #7954, opened 2022-10-17
[Deployment Parameterization] - Allow support for passing in key:value pairs to skaffold via a flag and allow for replacements across all renderers


aaron-prindle:
As part of the work for `Deployment Parameterization`, we have a feature request to add a flag to skaffold - `--set-values` that takes as args an array of key:value pairs (eg: `label:important-label`) and then pipes them to the renderer the user specified.  Currently the plan is to have two pathways:
- if `helm` renderer used, add these values to `values.yaml` (or appropriate similar location in code - eg: `setValues`)
- if a non-helm renderer is used, use kpt to set/apply to add the values as needed (as a pre/post renderer before specified renderer is used)

Issue #8318, opened 2023-01-12
Skaffold by default currently adds a namespace to all rendered manifests (--offline=true must be used to change this).  We should change the default not to add a `namespace``


aaron-prindle:
Currently when rendering a manifest with skaffold (`skaffold render ...`) by default skaffold adds a namespace to all rendered objects:
```
aprindle@aprindle-ssd ~/skaffold/examples/getting-started  [main]$ skaffold render -a artifacts.json
apiVersion: v1
kind: Pod
metadata:
  name: getting-started
  namespace: default # <--- namespace is added
spec:
  containers:
  - image: gcr.io/aprindle-test-cluster/skaffold-example:v2.0.2-79-g8d655cdb2@sha256:f45a3b94c62ea0a2f3375745809601af3cbb1aa64ebe36f388927096947f26c1
    name: getting-started
```

This is likely not the best experience for skaffold users as having a hard coded namespace on a manifest when it is not explicity requested in some form can lead to issues where users might want to `render` and then `deploy` to a different namespace but the manifests won't be deployable to that namespace as it has a namespace set likely unintentionally. 

Currently to get skaffold not to add a namespace when doing `skaffold render`, the `offline=true` flag must be added:
```
aprindle@aprindle-ssd ~/skaffold/examples/getting-started  [main]$ skaffold render -a artifacts.json --offline=true
apiVersion: v1
kind: Pod
metadata:
  name: getting-started
spec:
  containers:
  - image: gcr.io/aprindle-test-cluster/skaffold-example:v2.0.2-79-g8d655cdb2@sha256:f45a3b94c62ea0a2f3375745809601af3cbb1aa64ebe36f388927096947f26c1
    name: getting-started
```

This is not obvious (`offline` controls namespacing) and the default here should likely mimic this case vs what it does currently

NOTE: 
We should do a git-bisect and identify what PRs or what skaffold versions introduced this behavior and if this is a regression or how it has always worked

aaron-prindle:
Likely the approach with the best UX would be something like:
- change the default `skaffold render` to not add namespaces
- add a new flag `-set-namespace` that is used to get skaffold to set the namespace (the old default)
- change --offline=true message to not mention namespace as it is a bit confusing and the related conditions checking this flag related to setting the namespace.  The new default should mean users using this flag even without these special cases have their workflows still working. This means `offline=true` in the future won't have any mention/functionality related to rendering w/ or w/o a namespace.

[...]

PR #8365, opened 2023-01-25
feat: parameterization 

ericzzzzzzz:
Fixes: #8245, #7954

 Future work:  #8306, #8406, #8410  add kpt examples

### Description: 
 - move transformers responsibility out of specific kpt render as
   - transformers and validators are KRM functions, those functions should not be tied to a specific KRM function executors(kpt) in this case. We may support running krm functions without kpt(kustomize, or internal implementation).
   - ``Kustomize`` removes comments when building manifests,  the previous approach for using transformers doesn't work for manifests with comments. 
   -  creating extra hidden directory is not necessary, also it's contaminating users' workspace
   -  ``kpt render `` command outputs Kptfile content as part of render result unless using kpt with version under 1.0.0-beta.15 which is less preferred, this actually breaks krm function specs, and will cause `kubectl apply -f -` fail if piped with the render result. 
   - support parameterizations for all renders
     - kustomize
       - as kustomize build erases comments in the source manifests and we don't want to change users' source manifests either, we mirror kustomize directory to temporary folder and do transformation(apply-setters)  in a imperative way, then run kustomize build against the transformed manifests.
     - helm 
       - using helm native support approach ``--set flag``
       - This is not working if the render config is not directly set for manifests.helm,
     - kubectl  
     - kpt 
       - no extra folder, use ``kpt render`` command and output unwrapped result into memory, then running transformers against the output.
       - This currently doesn't work if kpt version is above 1.0.0-beta.15, will fix that in another pr. 

### Manual Test 
   - add kpt setters comment in your manifests, run ``skaffold render --set key1=value1 --set key2=value2``, should be able to see values with kpt setter comments are replaced by provided values from command line.


pkg/skaffold/runner/new.go
```diff
@@ -260,7 +260,7 @@ func isKptRendererOrDeployerUsed(pipelines runcontext.Pipelines) bool {
 		renderConfig := pipeline.Render
 		deployConfig := pipeline.Deploy
 
-		if renderConfig.Validate != nil || renderConfig.Transform != nil || renderConfig.Kpt != nil {
+		if renderConfig.Kpt != nil {
```
aaron-prindle, line 263: Why are these checks different now?

pkg/skaffold/helm/args.go
```diff
@@ -98,6 +98,11 @@ func ConstructOverrideArgs(r *latest.HelmRelease, builds []graph.Artifact, args
 
 		args = append(args, "-f", exp)
 	}
+
+	for _, k := range maps.SortKeys(manifestOverrides) {
```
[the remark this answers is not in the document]
ericzzzzzzz: yes.. tested locally later write wins.

pkg/skaffold/render/renderer/kpt/kpt.go
```diff
@@ -65,40 +60,43 @@ const (
 
 var (
 	KptVersion                      = currentKptVersion
-	maxKptVersionAllowedForDeployer = "1.0.0-beta.13"
+	maxKptVersionAllowedForDeployer = "1.0.0-beta.24"
```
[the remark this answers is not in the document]
ericzzzzzzz: This is a temporary workaround as using 1.0.0-beta.15 above will break kpt ``render``, but using lower version kpt doesn't support running krm function to  modify yaml out-place.  I'm going to remove the validation and fix ``kpt render`` output Kptfile issue.

pkg/skaffold/render/renderer/kpt/kpt_test.go
```diff
@@ -176,90 +88,36 @@ pipeline:
...
-					AndRunOut(fmt.Sprintf("kpt fn source %v -o unwrap",
-						filepath.Join(tmpDirObj.Root(), ".kpt-pipeline")), ""))
-			var b bytes.Buffer
-			_, err = r.Render(context.Background(), &b, []graph.Artifact{},
-				true)
-			t.CheckNoError(err)
-			t.CheckFileExistAndContent(filepath.Join(tmpDirObj.Root(), constants.DefaultHydrationDir, kptfile.KptFileName),
-				[]byte(test.updatedKptfile))
-		})
+		t.Skip("todo:" + test.description)
```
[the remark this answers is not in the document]
ericzzzzzzz: ohh.. I was thinking to re-write the tests. This skip was meant to add comments.  I'll add a separate comment.

pkg/skaffold/runner/new.go
ericzzzzzzz: > Why is this conditional different now?

yeah.. the function was mainly used to determine if a kpt hydrationDir should be created , because the original implementation around transform/validate is to marshal related config into a kptfile and put the kptfile and intermediary manifests  into the hydrationDir and then call ``kpt render`` to do the actual transformation in-place in that direcotry,  then call ``kpt source`` to get the result into memory.  In the new flow, we run krm function against a manifest input stream with ``kpt fn eval``, we don't create any directories.  Also, ``kpt`` here for renderConfig.Transform and renderConfig.Validate  is a krm function executor, this can be replace by internal function or ``kustomize run fn``. This is not tied to ``kpt render`` or ``kpt live apply`` , as the method name suggested ``isKptRendererOrDeployerUsed``,  those conditions shouldn't be here.

pkg/skaffold/render/renderer/kpt/kpt_test.go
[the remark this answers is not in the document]
ericzzzzzzz: Just realized that related tests are added in this pr .. I'm removing the test here.

pkg/skaffold/render/renderer/util/util.go
```diff
@@ -51,23 +51,29 @@ func GenerateHydratedManifests(ctx context.Context, out io.Writer, builds []grap
...
 	}
 	rs := manifest.NewResourceSelectorLabels(opts.TransformAllowList, opts.TransformDenylist)
 
 	if manifests, err = manifests.SetLabels(labels, manifest.NewResourceSelectorLabels(opts.TransformAllowList, opts.TransformDenylist)); err != nil {
-		return nil, err
+		return manifests, err
 	}
-	// TODO(tejaldesai) consult with cloud deploy team if namespaces can be set in offline mode
-	// in case namespace is set on the skaffold render cli command.
+	// This is kind of strange, it seems to indicate namespace setting is controlled by this offline flag. If a user set namespace through command line, the command line setting should always be honored
```
aaron-prindle, line 71: nit: I'm a bit confused if this comment should be here, it seems more this statement should be tracked in an issue - I think it might already be here:
https://github.com/GoogleContainerTools/skaffold/issues/8318

I don't think this comment adds clarify to any of the surrounding code atm

ericzzzzzzz: great! I'll remove it.

pkg/skaffold/render/renderer/kustomize/kustomize.go
```diff
@@ -0,0 +1,538 @@
...
+		}
+	}
+	return nil
+}
+
+func (k Kustomize) mirrorFile(kusDir string, fs TmpFS, path string) error {
+	if sUtil.IsURL(path) {
+		return nil
+	}
+	// todo making everything absolute path
```
aaron-prindle, line 289: nit: todos usually follow a syntax like
// TODO(#issue-number) <explanation>
OR
// TODO(@gh-username) <explanation>

aaron-prindle (approved):
LGTM, just a few nits.  Exciting!

Issue #8367, opened 2023-01-26
Skaffold slow to start before first log


nickdapper:
Trying to migrate v2 to v3 with about 40+ microservices. If I comment out more microservices, the startup time gets faster. Definitely a correlation there. 

### Expected behavior
Should see logs almost immediately to see what's happening

### Actual behavior
No logs for about 1.5 minutes.  

### Information

- Skaffold version:` v2.0.4`
- Operating system: `mac 12.5.1`
- Installed via: `homebrew`

Top level config file with microservices in each command folder. For example: 
```yaml
apiVersion: skaffold/v3
kind: Config
requires:
  - path: cmd/service-1/skaffold.yaml
  - path: cmd/service-2/skaffold.yaml
  - path: cmd/service-3/skaffold.yaml
  - path: cmd/service-4/skaffold.yaml
  - path: cmd/service-5/skaffold.yaml
  
```

Microservice: 
```yaml
apiVersion: skaffold/v3
kind: Config
metadata:
  name: service-1
profiles:
  - name: dev
    manifests:
      rawYaml:
        - 'deployment.yml'
        - 'service.yml'
    build:
      tagPolicy:
        inputDigest: {}
      artifacts:
        - image: myregistry/myorg/service-1
          context: build
          docker:
            dockerfile: Dockerfile
    deploy:
      kubectl: {}
```

### Steps to reproduce the behavior
`skaffold build -f skaffold.yaml -p dev`

### Logs
While I can't paste all logs (security), I can paste where I think the slowdown is. I didn't know about the `-vdebug` option before.  

```
DEBU[0000] Running command: [kubectl config view --minify -o jsonpath='{..namespace}']  subtask=-1 task=DevLoop
...repeats for 102 seconds? 
DEBU[0102] Running command: [kubectl config view --minify -o jsonpath='{..namespace}']  subtask=-1 task=DevLoop
```

ericzzzzzzz:
Hi @nickdapper  I might be a regression in v2.0.4, I think the main branch has the fix.. could you try version v2.1.0 to see if this version fix this issue?

nickdapper:
Thanks for the quick reply! 

I upgraded to `v2.1.0` and the problem persists.  Also, one odd thing is that when I run `skaffold version` it prints a blank line.  Not sure if you see the same.  Installed via homebrew.  

I'll create a project where you can reproduce the issue.

ericzzzzzzz:
[...]

PR #8376, opened 2023-01-27
fix: prevent long startup time

nickdapper:
Caches RunContext's kubernetes namespace so subsequent calls do not invoke kubectl.

Fixes: #8367 

**Description**
Use my example project found here: https://github.com/nickdapper/skaffold-slow

This caches the output of the `kubectl config view` command which only fetches the namespace. 

Please verify the location of returning the cached value is correct. 

This might not be the best approach - Why is `GetNamespace()` called so many times?

### DIFF
--- a/pkg/skaffold/runner/runcontext/context.go
+++ b/pkg/skaffold/runner/runcontext/context.go
@@ -35,14 +35,15 @@ import (
 )
 
 type RunContext struct {
-	Opts               config.SkaffoldOptions
-	Pipelines          Pipelines
-	KubeContext        string
-	Namespaces         []string
-	WorkingDir         string
-	InsecureRegistries map[string]bool
-	Cluster            config.Cluster
-	RunID              string
+	Opts                config.SkaffoldOptions
+	Pipelines           Pipelines
+	KubeContext         string
+	Namespaces          []string
+	WorkingDir          string
+	InsecureRegistries  map[string]bool
+	Cluster             config.Cluster
+	RunID               string
+	kubeConfigNamespace string
 }
 
 // Pipelines encapsulates multiple config pipelines
@@ -261,6 +262,10 @@ func (rc *RunContext) GetNamespace() string {
 	if rc.Opts.Namespace != "" {
 		return rc.Opts.Namespace
 	}
+	if rc.kubeConfigNamespace != "" {
+		return rc.kubeConfigNamespace
+	}
+
 	var defaultNamespace string
 	for _, p := range rc.GetPipelines() {
 		if p.Deploy.KubectlDeploy != nil && p.Deploy.KubectlDeploy.DefaultNamespace != nil {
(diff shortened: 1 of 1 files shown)


ericzzzzzzz:
Hi @nickdapper , thanks for your contribution. 
For caching a function call, we usually use this pattern,  https://github.com/GoogleContainerTools/skaffold/blob/88e1c1734ea0f1bd79a0f24783f9565057a0e46a/pkg/skaffold/kubectl/version.go#L96-L116 

``This might not be the best approach - Why is GetNamespace() called so many times?``

 If you check the commit history around this part of code,
 - 745309520c9f733a66a18a34b1d1f96161c94482 
 - 54d7e1a2dfd85f7ad281046e1e5770ec15ad5a99
 - b2ba1408cc341ee947045f8e05a12af730f39cf9
 - ee13c6f9e674da8a64ffeb7db18e22288703bf95

the initial implementation was just simply return namespace passed from command line, using this method was fine at that time, but  when we migrated to v2, there was some knowledge gap, that caused GetNamespace() discrepancy in a lot of places, the code added was meant to fix stuff, agree we should revisit that.  At this moment, I think we can cache the result to fix the start slow issue.

nickdapper:
Thanks for the fast review @ericzzzzzzz !  I updated to use the same pattern. Let me know how this looks

nickdapper:
@ericzzzzzzz I just tested this and while the command doesn't take almost two minutes, there is still a slight delay around 20 seconds before the first log. It's still calling `kubectl config view` 41 times in my setup.

ericzzzzzzz:
@nickdapper yeah.. there is another place calls ``kubectl config view``,  kubectl when deployer is initialized,  https://github.com/GoogleContainerTools/skaffold/blob/bcb59450d04d99db50ae54babb0af70f7ba1b4e0/pkg/skaffold/deploy/kubectl/kubectl.go#L87-L89 

We can create a syncstore in this package 
https://github.com/GoogleContainerTools/skaffold/blob/88e1c1734ea0f1bd79a0f24783f9565057a0e46a/pkg/skaffold/util/cmd.go#L1-L106 

syncstore implementation is here 
https://github.com/GoogleContainerTools/skaffold/blob/aa0bf348b1cb3151356715612ac3db5a63a0d4df/pkg/skaffold/util/store.go#L1-L121

And create a method like runOutOnce with wraps the runout command and use command args as key for caching purpose,  something like 


```
func RunCmdOutOnce(ctx context.Context, cmd *exec.Cmd) ([]byte, error) {
	return store.Exec(cmd.String(), func() ([]byte, error) {
		return RunCmdOut(ctx, cmd)
	})
}

```

nickdapper:
@ericzzzzzzz Excellent suggestion!  This results in just a single call now. 🚀  

Let me know if there are other changes needed.

nickdapper:
Will commits be squashed on merge?  Not seeing that.

nickdapper:
@ericzzzzzzz Is there a way to force a re-run of the checks?

aaron-prindle:
@nickdapper Maintainers can re-run specific parts of the Github Actions but I don't think the PR contributor can.  I think the easiest way to re-run the checks would be to re-push your commit and that will kick the CI jobs off.  Also you can ping the thread here and we can kick things off again if necessary/test-flake (specific Github Actions, etc.)

The unit tests that are currently failing can actually be run locally via `make tests` in the `skaffold/` root directory.  That might be an easier way to iterate here (assuming you have access to the platform the test is failing on).

nickdapper:
Ok all tests are passing now locally

nickdapper:
@ericzzzzzzz The previous commit re-ran checks but not the current one. Are you able to run them again please?  Should be passing now.

ericzzzzzzz (approved):

Issue #7913, opened 2022-10-04
FR: Add support for a hard-coded or default `--default-repo`


dhodun:
tldr: we are looking for an easy way to set a GCR repo for a full profile that does not involved developers always having to add the `--default-repo` flag.

*************

We are hoping for either a value somewhere in the skaffold configs, or a skaffold-env file to automatically set basic repo or project-wide values, namely `--default-repo` to control which GCR repo is used for a particular profile. This allows us to remove hard-coded GCR repos.

Some potential options:
- skaffold-env file at the repo level
- setting this variable once based on kubeContext activation
- some sort of set-once global skaffold variable in the root yaml

The minikube to GKE developer experience right now is well served with the kubeContext activations - when a developer has minikube started and runs `skaffold dev`, everything is deployed locally. When they run `kubectx my_cluster` and then `skaffold dev` again, it will deploy to the new cluster. However, to achieve this simplicity, we end up hard-coding the GCR repo of our dev project in all of our `dev` profiles.

Minikube (base config):
https://github.com/GoogleCloudPlatform/solutions-template/blob/47bd57a97935277801eb28dfff630bfd63051b0c/microservices/sample_service/skaffold.yaml#L13-L17

GKE (dev profile):
https://github.com/GoogleCloudPlatform/solutions-template/blob/47bd57a97935277801eb28dfff630bfd63051b0c/microservices/sample_service/skaffold.yaml#L71-L78

We have avoided hard-coding our GCR repo in other profiles (prod etc), but ultimately it has caused issues because our most tested skaffold config is the `dev` profile and when we start moving new features into the `prod` profile, it often breaks. When we copy+paste into prod we often get non-obvious errors from remnants of the hard coding.

Thus far we have left in the hard-coding to help new developers, since they don't have to remember to add `--default-repo` every time they run `skaffold`.

### Docker build caveat

Another small caveat where (I think) we benefit from using hard-coding is feeding CloudBuild or Docker builds images to cache from:
https://github.com/GoogleCloudPlatform/solutions-template/blob/47bd57a97935277801eb28dfff630bfd63051b0c/microservices/sample_service/skaffold.yaml#L73-L75

I believe `skaffold` used to and may still treat these values as literals, so `--default-repo` wasn't respected here and the caching broke. It's possible, though, I may be thinking of the just the git tags.

### Information

- Skaffold version: v1.36
- Operating system: Mac / linux

dhodun:
[...]

PR #8395, opened 2023-02-01
feat: define `skaffold.env` file for loading environment variables

gsquared94:
Fixes: #1545, #7913 

**Description**

In Skaffold, a `skaffold.env` file can be defined in the project root directory to specify environment variables that Skaffold will load into the process. This provides a more organized and manageable way of setting environment variables, rather than passing them as command line arguments.

The `skaffold.env` file should be in the format of `KEY=value` pairs, with one pair per line. Skaffold will automatically load these variables into the environment before running any commands.

Here is an example `skaffold.env` file:

```txt
ENV_VAR_1=value1
ENV_VAR_2=value2
```

### Setting Skaffold Flags with Environment Variables

In addition to loading environment variables from the `skaffold.env` file, Skaffold also allows users to set flags using environment variables. To set a flag using an environment variable, use the `SKAFFOLD_` prefix and convert the flag name to uppercase.

For example, to set the `--cache-artifacts` flag to `true`, the equivalent environment variable would be `SKAFFOLD_CACHE_ARTIFACTS=true`.

Here is an example usage in the `skaffold.env` file:

```txt
SKAFFOLD_CACHE_ARTIFACTS=true
SKAFFOLD_NAMESPACE=mynamespace
```

### DIFF
--- a/docs-v2/content/en/docs/environment/env-file.md
+++ b/docs-v2/content/en/docs/environment/env-file.md
@@ -1,0 +1,38 @@
+---
+title: "Load environment variables from a file"
+linkTitle: "Load ENV from a file"
+weight: 50
+aliases: [/docs/concepts/env_file]
+---
+
+In Skaffold, a `skaffold.env` file can be defined in the project root directory to specify environment variables that Skaffold will load into the process. This provides a more organized and manageable way of setting environment variables, rather than passing them as command line arguments.
+
+The `skaffold.env` file should be in the format of `KEY=value` pairs, with one pair per line. Skaffold will automatically load these variables into the environment before running any commands.
+
+Here is an example `skaffold.env` file:
+
+```txt
+ENV_VAR_1=value1
+ENV_VAR_2=value2
+```
+
+{{< alert title="Note" >}}
+Values set in a `skaffold.env` file will not overwrite existing environment variables in the process.
+{{< /alert >}}
+
+### Setting Skaffold Flags with Environment Variables
+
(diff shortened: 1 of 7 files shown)


aaron-prindle:
Can we add 1 integration tests that has a skaffold.env file and a go tempate that uses it in the skaffold.yaml (validating that it was correctly set)?  Want to make sure we have a way of verifying this continues to work over time (if we bump the deps or we change our env expansion logic)

docs-v2/content/en/docs/environment/env-file.md
```diff
@@ -0,0 +1,38 @@
...
+For example, to set the `--cache-artifacts` flag to `true`, the equivalent environment variable would be `SKAFFOLD_CACHE_ARTIFACTS=true`.
+
+Here is an example usage in the `skaffold.env` file:
+
+```txt
+SKAFFOLD_CACHE_ARTIFACTS=true
+SKAFFOLD_NAMESPACE=mynamespace
+```
+
+{{< alert title="Note" >}}
```
aaron-prindle, line 36: Can you add an explicit example/skaffold.yaml that has a skaffold.env file, a skaffold.yaml file, and 1 profile which swaps the env var used?

aaron-prindle:
Additionally once this is merged, we might need to coordinate w/ parter teams, etc. as some bundle all relevant skaffold files for their pipeline's.  We might need to add this to set of files to be tracked

aaron-prindle (approved):
LGTM

Issue #8389, opened 2023-01-30
Make nav menu options collapsible/expandable


maggieneterval:
Docsy now supports collapsing/expanding menu options(https://www.docsy.dev/docs/adding-content/navigation/#section-menu-options), which will improve the navigability of skaffold.dev documentation. (See https://kubernetes.io/docs/home/ nav as an example.)

To implement this, we will need to:
* Update the hugo/docsy version.
* Add `sidebar_menu_foldable = true` to config.toml.

Issue #8382, opened 2023-01-30
Improve the discoverability of Cloud Build documentation


maggieneterval:
Improve the discoverability of Cloud Build documentation:
* Refactor the [Builders](https://skaffold.dev/docs/pipeline-stages/builders/) page to include only the top-level overview section and a link to a standalone page for each builder.
* Refactor each subsection ([Local Build](https://skaffold.dev/docs/pipeline-stages/builders/#local-build), [In Cluster Build](https://skaffold.dev/docs/pipeline-stages/builders/#in-cluster-build), [Remotely on Google Cloud Build](https://skaffold.dev/docs/pipeline-stages/builders/#remotely-on-google-cloud-build)) into a standalone page.

PR #8398, opened 2023-02-01
docs: Upgrade hugo and docsy versions to enable collapsible nav

maggieneterval:
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/8389

**Description**

Docsy now supports collapsing/expanding menu options (https://www.docsy.dev/docs/adding-content/navigation/#section-menu-options), which improves the navigability of skaffold.dev documentation. (See https://kubernetes.io/docs/home/ nav as an example.)

Before breaking up the Builders page into sub-pages (https://github.com/GoogleContainerTools/skaffold/issues/8382), I wanted to ensure that sections were collapsible because the Build nav menu already has many sub-pages. 😄 

This PR:
1. Updates the hugo and docsy versions. (Note that these are not the latest releases, as updating to the latest versions introduced some errors to the build that it would be easier to address separately before upgrading.)
2. Adds `sidebar_menu_foldable = true` to config.toml.

**User facing changes**

Before:

3vU7TYBSk5jiPF5

After:

6Sk3nhMMqp5XktF

### DIFF
--- a/deploy/webhook-v2/Dockerfile
+++ b/deploy/webhook-v2/Dockerfile
@@ -14,7 +14,8 @@
 
 # Download Docsy theme for Hugo
 FROM alpine:3.10 as download-docsy
-ENV DOCSY_VERSION 3c4280bae0db96cb272ddc4b4959b2beef9299af
+# v0.3.0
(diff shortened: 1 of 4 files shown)


aaron-prindle (approved):
Awesome, great idea to add this!  Just tried it locally works great.  LGTM!

PR #8408, opened 2023-02-04
refactor: replace 4d63.com/tz with time/tzdata

pmalek:
**Description**

Looking at https://github.com/leighmcculloch/go-tz ( `4d63.com/tz` import path https://github.com/GoogleContainerTools/skaffold/blob/aa6f5a30f95ddc2008e4513c40b484ac3fa7f91b/go.mod#L32 ) it seems that the repo is unmaintained:

> Deprecated: Use [time/tzdata](https://golang.org/pkg/time/tzdata/) available in Go 1.15. time/tzdata does not work exactly the same as it defaults to local tzdata when available where as this package always uses embedded data. This package will be maintained until Go 1.16 is released.

This PR replaces usage of that package with `time/tzdata` by include a build tag `timetzdata` as suggested at https://pkg.go.dev/time/tzdata. An alternative is to import `time/tzdata` as suggested:

> This package should normally be imported by a program's main package, not by a library. Libraries normally shouldn't decide whether to include the timezone database in a program.

**Open questions**

Tests at https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/tag/date_time_test.go seem to pass (at least locally). Do we want to include more for this change?


aaron-prindle (approved):
@pmalek thanks for the PR here, LGTM!  The tests you updated should be enough here.

PR #8425, opened 2023-02-09
chore: remove validation for kpt version

ericzzzzzzz:
Fixes: #8306 
Related: https://github.com/GoogleContainerTools/kpt/issues/3844

**Description**
 - remove kpt version validation, the reason is in the linked issue. 
 - upgrade kpt version in skaffold image to support parameterization, the new version kpt has already been uploaded to gcs https://storage.googleapis.com/skaffold/deps/kpt/v1.0.0-beta.24/kpt_linux_amd64 
 - ``wget https://storage.googleapis.com/skaffold/deps/kpt/v1.0.0-beta.24/kpt_linux_amd64 -O kpt && shasum -a 256 -c ${skaffold_repo}deploy/skaffold/digests/kpt.amd64.sha256 ``  should output ok 

### DIFF
--- a/pkg/skaffold/render/renderer/kpt/kpt.go
+++ b/pkg/skaffold/render/renderer/kpt/kpt.go
@@ -17,19 +17,22 @@ limitations under the License.
 package kpt
 
 import (
+	"bytes"
 	"context"
-	"fmt"
 	"io"
 	"os/exec"
 
-	"github.com/blang/semver"
 	apimachinery "k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
 
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/graph"
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/instrumentation"
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/kubernetes/manifest"
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/render"
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/render/generate"
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/render/kptfile"
 	rUtil "github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/render/renderer/util"
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/render/transform"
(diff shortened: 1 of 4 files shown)


aaron-prindle:
+ @renzodavid9 who worked on this a bit in the past
My understanding was that for `kpt render` we didn't have a max version but for `kpt deploy` we did.  Can you confirm that `kpt deploy` works for all kpt versions and/or explain that for that use case there still is a version limit?

pkg/skaffold/render/renderer/kpt/kpt.go
```diff
@@ -117,7 +111,20 @@ func (r *Kpt) Render(ctx context.Context, out io.Writer, builds []graph.Artifact
 		cmd := exec.Command("kpt", "fn", "render", p, "-o", "unwrap")
 
 		if buf, err := util.RunCmdOut(rCtx, cmd); err == nil {
-			manifestList.Append(buf)
+			reader := kio.ByteReader{Reader: bytes.NewBuffer(buf)}
+			b := bytes.NewBuffer([]byte{})
+			writer := kio.ByteWriter{Writer: b}
+			// Kpt fn render outputs Kptfile content in result, we don't want this in our manifestList as Kptfile resource cannot be deployed to k8s cluster.
+			pipeline := kio.Pipeline{Filters: []kio.Filter{framework.ResourceMatcherFunc(func(node *yaml.RNode) bool {
+				return node.GetKind() != kptfile.KptFileKind
```
renzodavid9, line 119: If I recall correctly, I think the issue is not only with the `Kptfile` but with all the yaml files during rendering: kpt will read all the yaml files in the folder and render them. So, in the case we have a folder structure like this: 
```
project/
  Kptfile
  deployment.yaml
  skaffold.yaml
```
We'll need to filter the `Kptfile` and the `skaffold.yaml` or a deployer like `kubectl` will fail; I think the kpt deployer is able to filter this by itself properly. 🤔

ericzzzzzzz: users can also ignore skaffold.yaml file  by  .krmignore file, but there might be another problem related to ``fn render `` output, the command currently also output configMap file which cannot be used for deployment either, I'm gonna create an issue with kpt maintainers.

ericzzzzzzz: We don't filter out skaffold.yaml ourselves,  any yamls with ``Apiversion``,  ``Kind``, ``metadata.name`` will be included regardless of which kpt version we use .   It is users responsibility to make sure they put the right files in their kpt package  that should include ``Kptfile``, ``deployments files``, ``data config files`` , we can filter deployments files from the ``kpt fn render``.  Every standard kpt Data config file is expected to have ``config.kubernetes.io/local-config: "true"`` annotation , we can inspect the annotation for each yaml resource from ``kpt fn render`` and then filter them out.

renzodavid9 (approved):

Issue #8418, opened 2023-02-08
Update skaffold go version and related deps to meet security posture


aaron-prindle:
Currently skaffold has some out of date go deps that are causing our scanners to fire.  These should be updated

ericzzzzzzz:
as https://github.com/GoogleContainerTools/skaffold/pull/8433 doesn't work as expected.. re-open this.

PR #8431, opened 2023-02-12
chore: upgrade dependencies

ericzzzzzzz:
Fixes: #8418

## Description: 
 - The purpose of this pr is to upgrade some vulnerable dependencies reported by dependency bot. The upgrade of those caused their prerequisites upgraded as well which led to some broken code. Actions are taken for fixing those: 
   - open telemetry migration, mainly in the instrument package 
   - goyaml upgrade causes that marsharlling array has different format from before, the current one has indentation before ``-``, the previous one was aligned with its parent. Our integration tests use a lot of hard code yaml raw string comparison,  the is not a good practice, we should compare the target unmarshalled object field instead, to not include a lot of changes in this pr, I only convert those raw string into object for comparison, but we can make it better. 
   - go-git doesn't support adding empty commit by default anymore, hence specifying it in the test config.
   - fix ko-tests
   - deleted code related to ``generate pipeline``, it was broken after upgrade dependencies and I didn't see we use this anywhere, let me know if this is still in use.


pkg/skaffold/instrumentation/export.go
```diff
@@ -398,15 +405,15 @@ func initIOWriterTracer(w io.Writer) (*sdktrace.TracerProvider, func(context.Con
...
-	return texporter.InstallNewPipeline(
-		[]texporter.Option{
-			texporter.WithProjectID(os.Getenv("GOOGLE_CLOUD_PROJECT")),
-			texporter.WithOnError(func(err error) {
-				log.Entry(context.TODO()).Debugf("Error with metrics: %v", err)
-			}),
-		},
-		sdktrace.WithSampler(sdktrace.AlwaysSample()),
-	)
+	exporter, err := texporter.New()
```
ericzzzzzzz, line 408: this method has already handled default credential fetching.

ericzzzzzzz:
> Linter tests failing related to `gci`:
> 
> ```
> pkg/skaffold/render/renderer/kubectl/kubectl_test.go:22: File is not `gci`-ed with --skip-generated -s standard,default,prefix(github.com/GoogleContainerTools/skaffold) (gci)
> 	"github.com/google/go-cmp/cmp"
> ```
> 
> You can download `gci` here: https://github.com/daixiang0/gci
> 
> This can fixed running something like:
> 
> ```
> $ gci print -s "standard,default,prefix(github.com/GoogleContainerTools/skaffold)" pkg/skaffold/render/renderer/kubectl/kubectl_test.go
> ```

Thank you! I usually run ``make linters`` locally and then fixing the issue, but this command seems broken on latest macOS, I'll try ``gci`` next time.

pkg/skaffold/instrumentation/trace.go
```diff
@@ -91,15 +91,15 @@ func StartTrace(ctx context.Context, name string, attributes ...map[string]strin
...
 }
 
 // TraceEndError adds an "error" attribute with value err.Error() to a span during it's end/shutdown callback
 // This fnx is intended to used with the StartTrace callback - "endTrace" when an error occurs during the code path
 // of trace, ex: endTrace(instrumentation.TraceEndError(err)); return nil, err
-func TraceEndError(err error) trace.SpanOption {
+func TraceEndError(err error) trace.SpanEndOption {
 	if traceEnabled {
-		return trace.WithAttributes(attribute.KeyValue{Key: "error", Value: attribute.StringValue(PII(err.Error()))})
+		return trace.WithStackTrace(true)
```
ericzzzzzzz, line 102: now span.End takes SpanEndOption as input, however, we cannot create a SpanEndOption WithAttribute, I'd prefer to leave the code here, we can revisit that when working on exporting metrics to firelog.

aaron-prindle: 👍

pkg/skaffold/instrumentation/export.go
aaron-prindle: Makes sense

aaron-prindle (approved):
LGTM!

PR #8455, opened 2023-02-21
fix: cant deploy with kubectl if kubectl and pod have diffrerent namespace

catusax:
**Description**

if kubectl context's namespace is not `default`,and its different from pod's metadata.namespace, deploy will fail.

for example: 
kubectl context's namespace is `myns`, and manifests yaml is:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: testmodule
  namespace: podns
  labels:
    app: testmodule
  annotations:
    secrets.doppler.com/reload: 'true'
spec:
```

```
Starting deploy...
 - the namespace from the provided object "podns" does not match the namespace "myns". You must pass '--namespace=podns' to perform this operation.
 - the namespace from the provided object "podns" does not match the namespace "myns". You must pass '--namespace=podns' to perform this operation.
 - the namespace from the provided object "podns" does not match the namespace "myns". You must pass '--namespace=podns' to perform this operation.
 - the namespace from the provided object "podns" does not match the namespace "myns". You must pass '--namespace=podns' to perform this operation.
kubectl apply: exit status 1
```

So,if the the defaultNamespace of Deployer is kubectl's default namespace, then kubectl CLI  flag `--namespace=` should be removed.

### DIFF
--- a/pkg/skaffold/deploy/kubectl/kubectl.go
+++ b/pkg/skaffold/deploy/kubectl/kubectl.go
@@ -20,8 +20,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"os/exec"
-	"strings"
 
 	"github.com/segmentio/textio"
 	"go.opentelemetry.io/otel/trace"
@@ -86,13 +84,6 @@ type Deployer struct {
 // with the needed configuration for `kubectl apply`
 func NewDeployer(cfg Config, labeller *label.DefaultLabeller, d *latest.KubectlDeploy, artifacts []*latest.Artifact, configName string) (*Deployer, error) {
 	defaultNamespace := ""
-	b, err := util.RunCmdOutOnce(context.TODO(), exec.Command("kubectl", "config", "view", "--minify", "-o", "jsonpath='{..namespace}'"))
-	if err == nil {
-		defaultNamespace = strings.Trim(string(b), "'")
-		if defaultNamespace == "default" {
-			defaultNamespace = ""
-		}
-	}
 	if d.DefaultNamespace != nil {
 		var err error
 		defaultNamespace, err = util.ExpandEnvTemplate(*d.DefaultNamespace, nil)


ericzzzzzzz:
Hi @catusax , Thank you for creating the pr.   We've been discussing about not passing ``--namespace`` with  ``kubectl apply``,  for this, we need to remove namespace fetching when creating deployer like you did in this pr. Also we have to remove ``go
	if d.DefaultNamespace != nil {
		var err error
		defaultNamespace, err = util.ExpandEnvTemplate(*d.DefaultNamespace, nil)
		if err != nil {
			return nil, err
		}
	}
`` as this code will also lead to injecting namespace configured Deplyer.kubectl stanza in skaffoldy.yaml  to Kube CLI, otherwise the deployment will fail, if namespace in skaffold config doesn't match with the one in the manifest, or we can just remove code about appending namespace , https://github.com/GoogleContainerTools/skaffold/blob/88e1c1734ea0f1bd79a0f24783f9565057a0e46a/pkg/skaffold/kubectl/cli.go#L120-L123 . along with those.    However, there is more work to do here.  We're also working on removing namespace injection in Render stage, https://github.com/GoogleContainerTools/skaffold/issues/8318 , after this pr is complete,  we will not be able to deploy manifests to the namespace defined in skaffold.yaml if the block I mentioned is removed. If we remove this, we have to re-inject namespace into manifest at deploy stage to make everything works right! Considering the scope of the change and the amount of the work it requires, we'd assign this to an internal team member, so I'm going to close this pr. Again, thank you for the contribution!

rpstreef:
This problem still exists and in my case i'm using Helm charts in the Kustomization.yaml.
At render this is compiled.
The at Deploy time the kubectl force sets a namespace, that i have not configured at all, to the name: property value 

```
DEBU[0027] Running command: [kubectl --context default --namespace insight-prose get -f - --ignore-not-found -ojson]  subtask=0 task=Deploy
```

it also ignores the kubectl apply flags for that matter:

```
deploy:
      kubectl:
        flags:
          apply: ["--server-side", "--force"]
        hooks:
```

then it tries to execute and I get dozens of these, which are essentially ALL the name spaces other than the docker image and application helm chart namespace:

```
 - the namespace from the provided object "kube-system" does not match the namespace "insight-prose". You must pass '--namespace=kube-system' to perform this operation.
 - the namespace from the provided object "monitoring" does not match the namespace "insight-prose". You must pass '--namespace=monitoring' to perform this operation.
```

Issue #8423, opened 2023-02-09
Upgrade logic for v1 kustomize deployer, is not adding a deployer stanza


renzodavid9:
In Skaffold v1, kustomize is configured as a deployer, in Skaffold v2 it is a renderer. When the upgrade logic tries to transform a v1 config that uses `deploy.kustomize` into a v2 config, it adds the `manifests.kustomize` stanza, but doesn't add a deployer. In multi config/module projects this behaviour causes that the manifests renderer with kustomize are not deployed.

### Expected behavior
The upgrade logic (`skaffold fix`) to output configs that can deploy v1 manifests that uses the kustomize deployer, with  Skaffold v2 (for the case of multi config/module projects).

### Actual behavior
The resulting config can not be use to deploy manifests generated with kustomize (for the case of multi config/module projects)

### Information

- Skaffold version: version from @ HEAD
- Operating system: any
- Installed via: Built from @ HEAD
- Contents of skaffold.yaml (this is a Skaffold v1 config):

```yaml
apiVersion: skaffold/v2beta28
kind: Config
deploy:
  kustomize: {}

```

### Steps to reproduce the behavior

1. Using the [examples/kustomize](https://github.com/GoogleContainerTools/skaffold/tree/v1/examples/kustomize) project from `v1` branch
2. Run `skaffold fix` using Skaffold v2
3. The resulting config does not have a deploy stanza, making it to not work in multi config/module projects

aaron-prindle:
Making triage bot happy

PR #8457, opened 2023-02-21
fix: add upgrade logic to inject a kubectl deployer when an old kustomize deployer is detected

renzodavid9:
Fixes: #8423

**Description**
New logic to create a kubectl deployer when an old kustomize deployer is present during upgrade. The logic creates a new kubectl deployer and merge the values from the old kustomize deployer into it. When a kubectl deployer is already present, we try to merge it with the configuration from the old kustomize deployer. Code will error if the `defaultNamespace` or `flags.disableValidation` are present in both and are different.


pkg/skaffold/schema/v2beta29/upgrade.go
```diff
@@ -118,6 +118,17 @@ func (opu *OnePipelineUpgrader) upgradeOnePipeline(oldPipeline, newPipeline inte
...
 		}
+
+		// Inject Kubectl deployer to deploy kustomize manifests
+		kubectlDeployerWasInjected := false
+		if newPL.Deploy.KubectlDeploy == nil {
+			newPL.Deploy.KubectlDeploy = &next.KubectlDeploy{}
+			kubectlDeployerWasInjected = true
+		}
+
+		if err := mergeKustomizeIntoKubectlDeployer(newPL.Deploy.KubectlDeploy, oldPL.Deploy.KustomizeDeploy, kubectlDeployerWasInjected); err != nil {
```
gsquared94, line 129: will doing `mergeKustomizeIntoKubectlDeployer(oldPL *Pipeline, newPL *next.Pipeline)` simplify this?

pkg/skaffold/schema/v2beta29/upgrade.go
```diff
@@ -300,3 +311,32 @@ func upgradePatches(olds []JSONPatch, news []next.JSONPatch) {
 		}
 	}
 }
+
+func mergeKustomizeIntoKubectlDeployer(newKubectlDeployer *next.KubectlDeploy, oldKustomizeDeployer *KustomizeDeploy, kubectlDeployerWasInjected bool) error {
```
gsquared94, line 315: a good reading about variable names [here](go/go-style/decisions#variable-names)

renzodavid9: Awesome! Thanks for the reference, I changed the variables name. Let me know what do you think.

pkg/skaffold/schema/v2beta29/upgrade.go
renzodavid9: Done! I moved all the logic to the new function I created, removing the third parameter used before. Let me know what do you think.

gsquared94 (approved):

BabuRajan002:
We upgraded our Skaffold verstion to latest version 2.3 and Max Schema version is v4beat4. 
After that it stopped working during the deployment. Here is the sample file we are using
```
apiVersion: skaffold/v4beta4
kind: Config
metadata:
  name: sample-app
manifests:
  kustomize:
    paths:
      - </path/base>
profiles:
  - name: test
    manifests:
      kustomize:
        paths:
          - <path/kustomize>
        defaultNamespace: <dummy-namespace>
``` 

Our Kustomization's max schema version is "kustomize.config.k8s.io/v1beta1". Please help us to fix this issue. The current version we are using is 1.39 which accepts the above skaffold file with the kustomization version.

renzodavid9:
Hey @BabuRajan002, could you please open a new issue in the [Skaffold repo](https://github.com/GoogleContainerTools/skaffold/issues) with the details of the error you are getting? Also, it will be great if you can include a repo with a minimal version that can be use to reproduce the problem. I think for this case the content of the Kustomization file will be good to understand what's happening. Also, we can use that new issue to continue the discussion. Thanks a lot!

BabuRajan002:
@renzodavid9 As you mentioned above, I've created a new issue here - https://github.com/GoogleContainerTools/skaffold/issues/8802. But I did not get any response.

Issue #8453, opened 2023-02-20
Not able to reference secrets path to home folder


griseau:
### Expected behavior
I want to reference a secret file in `build.artifacts.docker.secret.src` in the home directory. This works well if using the realpath but not ~.
### Actual behavior
Using the realpath /home/username/.npmrc works, but using ~/.npmrc doesn't work because skaffold is appending ~/.npmrc to the actual working directory making it search to the wrong place.
I also tried to reference the home directory using $HOME env variable, but it's not a templated field, so doesn't work
### Information

- Skaffold version:  skaffold 2.1.0
- Operating system:  MacOS Ventura 13
- Installed via:  brew
- Contents of skaffold.yaml:

```yaml
build:
  local:
    useBuildkit: true
  artifacts:
  - image: image-name
    context: .
    docker:
      secrets:
        - id:   npmrc
          src: ~/.npmrc
```

PR #8476, opened 2023-02-24
fix: expand home directory for docker secrets

griseau:
Fixes: #8453 

**Description**

As mentioned in the issue, when using ~/ for secret path, this wasn't expended to home directory but was happened after current directory. This change expends the path to the realpath of home directory if starting with ~/

No user facing changes other than have it working as intended

Should we test the util function ? It's only calling Go standard package

And thanks @smaftoul for your help on this one ;)


griseau:
@ericzzzzzzz I see the issue with the test failing but not sure to understand, any help for resolving this ?

ericzzzzzzz:
@griseau This seems like a flaky test.. just re-ran to see if the test pass.

ericzzzzzzz (approved):

PR #8514, opened 2023-03-08
fix: move verify schema changes to v4beta4 and remove it from already released v4beta3

aaron-prindle:


gsquared94 (approved):
Thanks for clarifying that failing tests are expected here. LGTM!

PR #8539, opened 2023-03-10
chore: set up skaffold binary scanning 

ericzzzzzzz:
**Related**: tracked internally 

**Description**
 - Set up monitor to scan skaffold go binary vulnerabilities and report to github
 - The monitor scans skaffold binary built from the HAED of main branch as well as latest LTS version binary, such v2.0.6, v1.39.6 at the moment. 
 - The monitor will create/close issues based on the scanning result
   - For edge binary(built from the HEAD of main)
     - If there is an open issue for edge binary vulns, and the scanning result shows binary vulns, then there is no action.
     -  If there is an open issue for edge binary vulns, and the scanning result shows no binary vulns, the script will close the issue. 
     - If there is no open issue for edge binary vulns, and the scanning result shows vulns, then script create an issue to track the problem.
  - For LTS binary, the scanner always scans the up-to-date LTS although there might be an open issue about vulns found in previous version. For example, there is an open issue for v2.0.x, and we release a new version v2.0.x+1 , the scanner will target for v2.0.x+1, if the new release fixes the issue, the script will just close the original one(v2.0.x), otherwise other than closing the old one, it will also create a new issue to track skaffold vuls issues. If we don't have a new release, then the behavior should be the same as for edge version. Note that, we uses the release time of x.y.1 as the LTS start time, we only scan the latest LTS release if its starting point is within ``one year``.
  - The monitor will ping skaffold-team for every issue it creates to  increase awareness.

**Test Plan**
- manually run ``test-bin-scanning`` build trigger in skaffold project,  I tested this morning and related vuls were already created by my ``run``, you can close some of those issues and trigger the build to see the result, this should work as the description. 
 

**Monitor Frenquency**
 - set up on demand scanning with pub/sub with ``us-east1-docker.pkg.dev/k8s-skaffold/scanning/skaffold`` image creation events, the images should only be pushed by ``/deploy/cloudbuild`` and  ``deploy/cloudbuild-release-lts``.
 - set up daily scanning job

### DIFF
--- a/deploy/binary-vuln-monitor/report.sh
+++ b/deploy/binary-vuln-monitor/report.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2023 The Skaffold Authors
+# Copyright 2021 The Skaffold Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/deploy/binary-vuln-monitor/scan.sh
+++ b/deploy/binary-vuln-monitor/scan.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2023 The Skaffold Authors
+# Copyright 2021 The Skaffold Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.


ericzzzzzzz:
may need some changes for re-using the same script to scan different images with different reporting templates.

PR #8543, opened 2023-03-10
fix: `kubectl create` with `dry-run` shouldn't pass the namespace flag

gsquared94:
Skaffold runs `kubectl create --dry-run ... --namespace=foo` when reading a manifest in: https://github.com/GoogleContainerTools/skaffold/blob/e74ef33d0c6c88fc24a901dab7f0044dc41116eb/pkg/skaffold/deploy/kubectl/cli.go#L197

This can be run without the `--namespace` flag as Skaffold sets the namespace as a post-render step anyways.


aaron-prindle (approved):
LGTM!

Currently it seems one kokoro test is failing though, just need to fix that and should be good

Issue #8456, opened 2023-02-21
Add --output flag to `skaffold diagnose` flag options


aaron-prindle:
Issue tracks add `--output` flag to `skaffold diagnose` flag options

PR #8546, opened 2023-03-13
chore: Add output flag for diagnose

ericzzzzzzz:
Fixes: #8456


aaron-prindle (approved):
LGTM!

aaron-prindle:
@ericzzzzzzz my only concern here is maybe we should talk to internal partners to make sure putting all of the output in in the `--output` file is ok vs only the yaml (right now I believe `diagnose` outputs some additional text in a header)

Issue #8432, opened 2023-02-13
Add Firelog exporter to export metrics


ericzzzzzzz:
- skaffold team would like to enhance the skaffold metric-collecting to better understanding how users use skaffold
 - The metric will help us decide projects priorities
 - Currently we're using CloudMonitorMetricsExporter to export our metrics to gcp cloud monitoring platform
 - We're moving metric platform from cloud monitoring platfom to firelog one platform, so we need another exporter to send metrics data to the platform 

The ask in this issue, is to implement firelog metrics exporter.

ericzzzzzzz:
#8555  is reverted due to release timing, re-open this issue.

PR #8555, opened 2023-03-16
chore: firelog exporter 

ericzzzzzzz:
Fixes: #8432 

### Description
 - This pr switches the open telemetry exporter from cloudMonitorExport to fireLogExporter , so we can push skaffold metrics to firelog one-platform
 - Adds ci-platform data into metrics 
 - By default, the firelog apiKey is empty, so non-exporter is enabled at the moment. 
### Future work
 - The current implementation is sending metrics data sequentially, we may send metrics data in parallel.
 - We might also need to add a flag or global config variable to toggle enabling/disabling exporter
 - More metrics data may be added.

### Test Plan
 - add apiKey and build the binary, the run ``skaffold dev -vtrace`` in  examples/getting-started project
 - Should be able to see ``metrics uploading complete in xxxx seconds`` from the log


gsquared94 (approved):
lgtm

PR #8561, opened 2023-03-17
feat: standarize renders to inject namespace only if --namespace or render specific config is set

renzodavid9:
Fixes: #8318

**Description**
This PR will make the existing renderers to only inject a `metadata.namespace` field if a `--namespace` flag is set or if the render itself is configured for that.

**Pending**
Validate helm behavior


aaron-prindle (approved):
LGTM!

Issue #8522, opened 2023-03-09
[Custom Actions] Create actions runner to manage the actions execution


renzodavid9:
- It should manage the fail strategy (fail-fast or fail-safe)
- It should trigger the execution of all the containers in the action, in parallel
- It should report back the state of the execution"

PR #8563, opened 2023-03-21
feat: custom actions interfaces and actions runner implementation

renzodavid9:
Fixes: #8522

**Description**
This PR adds the set of interfaces and types used by the Custom Actions, and implements the ActionsRunner in charge of executing a custom action according to the configuration.

### DIFF
--- a/pkg/skaffold/actions/runner.go
+++ b/pkg/skaffold/actions/runner.go
@@ -27,12 +27,20 @@ import (
 
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/graph"
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/output/log"
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/runner"
 )
 
+var _ runner.ActionsRunner = Runner{}
+
 type Runner struct {
+	// Map to access the associated Exec environment of a given action.
 	execEnvByAction map[string]ExecEnv
+
+	// List with all the created exec environments; this helps to execute all the actions in the same order always.
 	orderedExecEnvs []ExecEnv
-	acsByExecEnv    map[ExecEnv][]string
+
+	// Map to access the list of associated actions of a given Execution environment.
+	acsByExecEnv map[ExecEnv][]string
 }
 
 func NewRunner(execEnvByAction map[string]ExecEnv, orderedExecEnvs []ExecEnv, acsByExecEnv map[ExecEnv][]string) Runner {
@@ -145,8 +153,8 @@ func execWithFailingSafe(ctx context.Context, out io.Writer, ts []Task) error {
 }
 
 func execAndLog(ctx context.Context, out io.Writer, t Task) error {
-	log.Entry(context.TODO()).Debugf("Starting execution for %v", t.Name())
+	log.Entry(ctx).Debugf("Starting execution for %v", t.Name())
 	err := t.Exec(ctx, out)
-	log.Entry(context.TODO()).Debugf("Execution finished for %v", t.Name())
+	log.Entry(ctx).Debugf("Execution finished for %v", t.Name())
 	return err
 }
--- a/pkg/skaffold/actions/runner.go
+++ b/pkg/skaffold/actions/runner.go
@@ -43,7 +43,15 @@ func (r Runner) ExecAll(ctx context.Context, out io.Writer, allbuilds []graph.Ar
 		return err
 	}
 
-	return nil
+	acs := r.allActions()
+	var ts []Task
+	for _, a := range acs {
+		execF := r.getExecFunc(a)
+		a.SetTasksExecFunc(execF)
+		ts = append(ts, a)
+	}
+
+	return r.execParallelFailingSafe(ctx, ts)
 }
 
 func (r Runner) prepareExecEnvs(ctx context.Context, out io.Writer, allbuilds []graph.Artifact) error {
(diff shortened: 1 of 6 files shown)


pkg/skaffold/actions/runner.go
```diff
@@ -0,0 +1,167 @@
...
+		close(execCh)
+	}()
+
+	select {
+	case err = <-execCh:
+		log.Entry(ctx).Debugf("Finishing execution for %v", t.Name())
+
+	case <-time.After(t.Timeout()):
+		msg := fmt.Sprintf("timing out %v", t.Name())
+		log.Entry(ctx).Debugf("Finishihin execution:%v", msg)
```
gsquared94, line 143: nit: typo

pkg/skaffold/actions/runner.go
```diff
@@ -0,0 +1,167 @@
...
+
+	return g.Wait()
+}
+
+func (r Runner) execute(ctx context.Context, t Task, out io.Writer) error {
+	var err error
+	execCh := make(chan error)
+
+	go func() {
+		execCh <- t.Exec(ctx, out)
```
gsquared94, line 133: since `Task` defines a `Timeout` shouldn't it be the responsibility of the `Task` implementation to ensure that it gets cancelled on timeout. In which case, you should be able to just call `t.Exec` directly without the separate goroutine wrapping and the `select` statements since this method is essentially synchronous. WDYT?

pkg/skaffold/actions/runner.go
```diff
@@ -0,0 +1,167 @@
...
+	return err
+}
+
+func (r Runner) cleanup(ctx context.Context, out io.Writer, ts []Task, execEnvs []ExecEnv) {
+	for _, t := range ts {
+		if t == nil {
+			continue
+		}
+		if err := t.Cleanup(ctx, out); err != nil {
+			log.Entry(ctx).Debugf("%v cleanup error:%v", t.Name(), err)
```
gsquared94, line 156: should these be level `Warn` instead of `Debug`?

pkg/skaffold/actions/types.go
```diff
@@ -0,0 +1,67 @@
...
+	// Exec triggers the execution of the task.
+	Exec(ctx context.Context, out io.Writer) error
+
+	// Cleanup frees the resources created by the task to execute itself.
+	Cleanup(ctx context.Context, out io.Writer) error
+}
+
+// Action represents a single custom action.
+// It is composed by Tasks, and itself is a Task too.
+type Action interface {
```
gsquared94, line 47: `Action` interface seems a bit redundant since it only encapsulates a `Task`. If the `Action` is fail fast then its `ExecFunc` is already known to be `execWithFailingFast`, all of which will be known when instantiating an `Action` implementation struct. Maybe all you need are `FailFastTaskRunner(ctx context.Context, t []Task)` and `GroupTaskRunner(ctx context.Context, t []Task)` functions.

pkg/skaffold/actions/runner.go
renzodavid9: You are right. From the config point of view, the timeout is defined at the action level. With the latest changes I created a new `Action` struct type that handles its tasks timeout. This way, the logic to interrupt the execution due to a time out is always the same regardless if we are executing docker or k8s tasks. Let me know what do you think!

pkg/skaffold/actions/types.go
renzodavid9: Cool! I changed this. Now we don't have the action interface, instead we have an Action struct that is the same regardless of the specific Tasks implementation it holds.

pkg/skaffold/actions/runner.go
```diff
@@ -0,0 +1,152 @@
...
+	"golang.org/x/sync/errgroup"
+
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/graph"
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/output/log"
+)
+
+type Runner struct {
+	execEnvByAction map[string]ExecEnv
+	orderedExecEnvs []ExecEnv
+	acsByExecEnv    map[ExecEnv][]string
```
gsquared94, line 35: add a description of what each of these fields mean

pkg/skaffold/actions/runner.go
```diff
@@ -0,0 +1,152 @@
...
+		g.Go(func() error {
+			return execAndLog(ctx, out, t)
+		})
+	}
+
+	return g.Wait()
+}
+
+func execAndLog(ctx context.Context, out io.Writer, t Task) error {
+	log.Entry(context.TODO()).Debugf("Starting execution for %v", t.Name())
```
gsquared94, line 148: can't `log.Entry` use `ctx` instead of `context.TODO`?

pkg/skaffold/actions/runner.go
```diff
@@ -0,0 +1,152 @@
...
+	"math"
+
+	"github.com/fatih/semgroup"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/graph"
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/output/log"
+)
+
+type Runner struct {
```
gsquared94, line 32: add a line:
```
var _ ActionsRunner = (Runner)(nil)
```
so that `Runner` cannot break the `ActionsRunner` interface accidentally

gsquared94 (approved):
lgtm

pkg/skaffold/actions/runner.go
renzodavid9: Done.

pkg/skaffold/actions/runner.go
renzodavid9: I was thinking about a case when one action fails, canceling the context, but we should be fine using `ctx` here. Done, thanks!

pkg/skaffold/actions/runner.go
renzodavid9: Oh! Cool! it's a way to have a visible, easy to understand error in the same file, right? I just added it, done!

PR #8575, opened 2023-03-23
feat: add 'skaffold inspect jobManifestPath' and 'skaffold transform-schema jobManifestPath' commands

### LINKED ISSUE #8573: Add abillity to modify `skaffold verify` jobManifestPath
Issue tracks adding the abillity to modify `skaffold verify` jobManifestPath by adding new commands to support this use case

aaron-prindle:
fixes #8573

### DIFF
--- a/pkg/skaffold/inspect/jobManifestPaths/modify.go
+++ b/pkg/skaffold/inspect/jobManifestPaths/modify.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2021 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package inspect
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/config"
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/output/log"
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/parser"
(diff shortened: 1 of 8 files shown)


pkg/skaffold/inspect/jobManifestPaths/modify.go
```diff
@@ -0,0 +1,84 @@
...
+
+var (
+	getCfgs = parser.GetAllConfigs
+)
+
+func Modify(ctx context.Context, out io.Writer, opts config.SkaffoldOptions, inputFile, outputFile string) error {
+	// Open our jsonFile
+	jsonFile, err := os.Open(inputFile)
+	if err != nil {
+		fmt.Println(err)
```
renzodavid9, line 42: Shouldn't we return the err here? Or log it with `log.Entry(context.TODO()).Error`?

pkg/skaffold/inspect/jobManifestPaths/modify.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	// defer the closing of our jsonFile so that we can parse it later on
+	defer jsonFile.Close()
+
+	byteValue, _ := ioutil.ReadAll(jsonFile)
+
+	var result jobManifestPathList
+	json.Unmarshal(byteValue, &result)
```
renzodavid9, line 50: This operation can return an error, if the `inputFile` is not a json file. I noticed that we fail silently, and the generated output is the same skaffold.yaml without the modifications. Should we also log the error from the unmarshal?

pkg/skaffold/inspect/jobManifestPaths/modify.go
aaron-prindle: Yep, good catch.  This now properly handles errors and has a test case related to this

pkg/skaffold/inspect/jobManifestPaths/modify.go
aaron-prindle: Yep, good catch.  This now properly handles errors and has a test case related to this

pkg/skaffold/inspect/jobManifestPaths/modify.go
aaron-prindle: Also logs the error as well to stdout for users to know what happened

pkg/skaffold/inspect/jobManifestPaths/modify.go
aaron-prindle: Also logs the error as well to stdout for users to know what happened

renzodavid9 (approved):

PR #8582, opened 2023-03-24
chore: fix transformer share config-map issues

ericzzzzzzz:
Fixes: #8580 

**Description**
 - kpt Transformers with the same name are currently sharing configmap which can produce unexpected behavior 
   -  uses could define create-setter, apply-setter, create-setter, apply-setter in a row in skaffold transform stanza 
      each entry should only be able to access its own datamap
  - kpt set-namespace image doesn't have a latest version, we need to specify the version, so added one. 
  - change cmd run to use util cmd run to make debugging life easier


aaron-prindle (approved):
LGTM!

Issue #6996, opened 2022-01-06
Incorrect use of google.CredentialsFromJSON?


feliperyan:
`google.CredentialsFromJSON` is being called with the equivalent of 

```
{
    "token": "some_token_value",
    "type": "authorized_user"
}
```

Although no error is returned, it's constructing an _empty_ `google.Credentials` struct. I think this results on application default credentials always being called.

https://github.com/GoogleContainerTools/skaffold/blob/e1f52f45eed7de5708576a097eca0bb619d7c1ff/pkg/skaffold/gcp/auth.go#L64

MarlonGamez:
hey @feliperyan , thanks for taking the time to open this. Do you have any more information into the issue here, e.g. debug logs/skaffold's output that's being caused by this issue?

feliperyan:
Hey Marlon, using `-vdebug` I get:

```
NFO[0001] unable to retrieve token: oauth2: token expired and refresh token is not set  subtask=-1 task=DevLoop
INFO[0001] falling back to application default credentials  subtask=-1 task=DevLoop
DEBU[0001] project id set to fryan-secure-1              subtask=gcr.io/fryan-secure-1/cloudonar-deleteme task=Build
DEBU[0003] Running command: [tput colors]                subtask=-1 task=DevLoop
DEBU[0003] Command output: [256
]                        subtask=-1 task=DevLoop
creating bucket fryan-secure-1_cloudbuild in project fryan-secure-1 if not exists: googleapi: Error 400: User project specified in the request is invalid., invalid
```

I think the json value being passed to `c, err := google.CredentialsFromJSON(context.Background(), body)` on line 77 is not sufficient to create a valid `google.Credentials` struct. 

Later when `_, err = c.TokenSource.Token()` is called, I believe it's always failing and falling back to ADC when it's later used at https://github.com/GoogleContainerTools/skaffold/blob/e1f52f45eed7de5708576a097eca0bb619d7c1ff/pkg/skaffold/build/gcb/cloud_build.go#L76 to create a new Cloud Build client

feliperyan:
Hi @marlon-gamez I believe this issue is also causing https://github.com/GoogleContainerTools/skaffold/issues/6743

PR #8584, opened 2023-03-24
fix: use active gcp account

ericzzzzzzz:
Fixes: #6996 
**Description**
 - update the method to `get` gcp active account credential 
 - The original method to retrieve account credential always produces error when calling credential.tokenSource.token method. This might have not been verified when it was implemented, seems a little strange to `create` credential with an access token. My understanding is that the credential is used to retrieve an access token and the token can be used to initiate other requests to other services. oauth.google package can use adc(application default credential) 
```json
 "client_id": "..."
 "client_secret" :".."
 "refresh_token":"...."
 "type": ".."
 ``` 
to construct tokenSource which is used retrieve access token , we can also implement our own token source and provide it to google.Credentials struct , then the tokenSource.token() will be called when needed to request gcp services. 

## Test Plan
 - create a service account from your project 
 - create keyfile for your account
 - run ``gcloud auth activate-service-account {account_name} --key-file={file}`` to activate your account
 - run ``skaffold build -vdebug`` in ``integration/testdata/gcb-explicit-repo`` project, the run should fail if you didn't give your service account k8s-skaffold storage access and artifact registry, this is good to prove that we're using the active account to issue request
 - example message ``creating bucket if not exists: getting bucket "k8s-skaffold_cloudbuild": googleapi: Error 403: {{service-account-name}} does not have storage.buckets.get access to the Google Cloud Storage bucket. Permission 'storage.buckets.get' denied on resource (or it may not exist)., forbidden``

### DIFF
--- a/pkg/skaffold/gcp/auth.go
+++ b/pkg/skaffold/gcp/auth.go
@@ -17,13 +17,16 @@ limitations under the License.
 package gcp
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"os/exec"
 	"sync"
+	"time"
 
 	"github.com/docker/cli/cli/config/configfile"
+	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/output/log"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/skaffold/gcp/auth.go [hunk 1]
```diff
@@ -59,35 +62,51 @@ func AutoConfigureGCRCredentialHelper(cf *configfile.ConfigFile) {
...
+}
+
+func (ts tokenSource) Token() (*oauth2.Token, error) {
+	// the command return a json object containing id_token, access_token, token_expiry
+	cmd := exec.Command("gcloud", "auth", "print-identity-token", "--format=json")
+	var body bytes.Buffer
+	cmd.Stdout = &body
+	err := util.RunCmd(context.TODO(), cmd)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get access token %v", err)
```
aaron-prindle, line 80: nit: might make sense to differentiate the error messages here so it is clearer which one of these 2 steps failed

PR #8604, opened 2023-03-29
chore: upgrade deps in v1.37

ericzzzzzzz:
Fixes: #8568 

**Description**
 - The goal of this pr is to fix skaffold bin vulns from go stdlib  and some dependent packages , such as containerd.
 - To do that I have to upgrade go sdk from 1.17 to 1.19.6 in gcloud pipeline and github action pipeline
 - Also I need to make code changes due to transitive dependencies upgrade from direct dependencies(the transitives are also directly used in our code base), changes include
   - yaml3 upgrade causes object marshaling in a different format, array nodes now need to have indentation under  their parents, so tests were comparing yaml marshaling string  with hard-coded raw string are broken, now we need to compare yaml objects from those strings. 
   - new git lib doesn't allow empty message in commit by default now. 
   - bazel examples are changed due to ci doesn't have python 2.7, hence we also need to change how skaffold process bezel docker rules.
   - hack/check_schema was pointing to the wrong branch, it should be pointing to a previous base branch not main . 
   - golangci_lint lib need to upgrade as well for proper linting, rules also have to be updated. 
   - upgrade hack/tools deps as well
   - re-run gofmt 
   - deleted generate_pipeline command, this feature requires a lot of code change to make the build happy, and this feature probably should not be in skaffold anyway, so delete this feature in this pr.


ericzzzzzzz:
Let's discuss the changes before merging this

Issue #8626, opened 2023-04-03
Support templating in the `chartPath` field


aniskhan001:
It would be very nice to have templating support in the `chartPath` field.

## Use case
I have a repo containing multiple projects which need to render separately.
Separate render is important here because I'm trying to the rendered manifests for different ArgoCD Applications.
Using a matrix in GitHub Action gives me the flexibility of changing the name of each project.

Here is an example pipeline in GitHub Action:
```yaml
name: 'GitOps'
on:
  push:
    branches:
      - 'my-branch'

jobs:
  skaffold-gitops:
    name: 'GitOps'
    runs-on: ['ubuntu-latest']

    strategy:
      matrix:
        app-name:
          - 'app-name-1'
          - 'app-name-2'
          - 'app-name-3'

    steps:
      - name: 'Checkout'
        id: 'checkout'
        uses: 'actions/checkout@3.0.0'

      - name: 'Render Manifests'
        uses: 'example-workflow-skaffold@1.0.0'
        env:
          PROJECT_NAME: ${{ matrix.app-name }}
```

And in the Skaffold config file, I would like to use that `PROJECT_NAME` so that I don't have to keep multiple files for each project.

```yaml
apiVersion: skaffold/v3
kind: Config
metadata:
  name: {{.PROJECT_NAME}}

manifests:
  helm:
    releases:
      - name: {{.PROJECT_NAME}}
        # chartPath does not support templating at the moment, sadly 😢
        chartPath: kubernetes/helm/charts/{{.PROJECT_NAME}}
        valuesFiles:
          - kubernetes/helm/values/values-global.yaml
```

### Steps to reproduce the behavior

Right now, if I run `skaffold render` command with above configuration, it would give me the following error:

```bash
std out err: %!(EXTRA *errors.errorString=Error: repo kubernetes not found
exit status 1
)
```

### Information

- Skaffold version: v2.3.0
- Operating system: macOS 13.3
- Installed via: Homebrew

aniskhan001:
@aaron-prindle I created [a PR](#8645) for this. But it seems I don't have the option to request reviews there.
Can you please take a look when you have time? Thanks! 🙂

sakirma:
I am reopening this issue again, since the result in the unit test does not show the correct render on the latest tag of skaffold. 

Let me know if I have to write a new issue for this. 

[...]

PR #8645, opened 2023-04-05
feat: add template support for `chartPath`

aniskhan001:
Fixes: #8626 

**Description**
The `skaffold render` command doesn't support templating on `chartPath` field at the moment.
This PR attempts to enable templating on `chartPath` field.

I have added a basic unit test, please review it and let me know if I need to add/adjust something there.

**User facing changes**
We would get the following error when there is a template/environment variable
```bash
std out err: %!(EXTRA *errors.errorString=Error: repo kubernetes not found
exit status 1
)
```

With this PR, template/environment variables should render successfully.

### DIFF
--- a/pkg/skaffold/deploy/helm/helm_test.go
+++ b/pkg/skaffold/deploy/helm/helm_test.go
@@ -89,6 +89,17 @@ var testDeployEnvTemplateNamespacedConfig = latest.LegacyHelmDeploy{
 	}},
 }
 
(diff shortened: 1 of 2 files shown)


aaron-prindle (approved):
@aniskhan001 thanks for the PR here!  LGTM

I will update the page here:
https://skaffold.dev/docs/environment/templating/

as well to include this information in a separate PR.  Thanks again!

PR #8649, opened 2023-04-06
fix: create new docker network only when the --docker-network flag is not set

renzodavid9:
**Description**
Change behaviour so we don't create a new docker network when `skaffold verify` is run with the `--docker-network` flag, instead we assume the docker network with the given name already exists.

### DIFF
--- a/cmd/skaffold/app/cmd/flags.go
+++ b/cmd/skaffold/app/cmd/flags.go
@@ -675,7 +675,7 @@ The build result from a previous 'skaffold build --file-output' run can be used 
 	{
 		Name:          "docker-network",
 		Shorthand:     "",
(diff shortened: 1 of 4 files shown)


aaron-prindle:
Can you also update the flag help description here to indicate that this docker-network will not be created and what happens by default

https://github.com/GoogleContainerTools/skaffold/blob/main/cmd/skaffold/app/cmd/flags.go#L678

renzodavid9:
> Can you also update the flag help description here to indicate that this docker-network will not be created and what happens by default
> 
> https://github.com/GoogleContainerTools/skaffold/blob/main/cmd/skaffold/app/cmd/flags.go#L678

Done! I changed the flag description.

aaron-prindle (approved):
LGTM!

PR #8660, opened 2023-04-10
fix: update skaffold verify to respect deploy default namespace field

### LINKED ISSUE #8629: `skaffold verify` should use the `defaultNamespace` specified in `deploy` as it's own default namespace
Issue tracks the idea that `skaffold verify` should use the `defaultNamespace` specified in `deploy` as it's own default namespace.  Currently `skaffold verify` deploys to `default` by default which is not expected for most users and leads to confusion, perms issues,  etc.

Slack thread that brought up this issue w/ more context:
https://kubernetes.slack.com/archives/CABQMSZA6/p1680520998478329

aaron-prindle:
fixes #8629


pkg/skaffold/runner/verifier.go
```diff
@@ -52,8 +52,16 @@ func GetVerifier(ctx context.Context, runCtx *runcontext.RunContext, labeller *l
 		}
 	}
 
+	var defaultDeployNS *string
+	if runCtx.DefaultPipeline().Deploy.KptDeploy != nil {
+		defaultDeployNS = runCtx.DefaultPipeline().Deploy.KptDeploy.DefaultNamespace
+	}
+	if runCtx.DefaultPipeline().Deploy.KubectlDeploy != nil {
+		defaultDeployNS = runCtx.DefaultPipeline().Deploy.KubectlDeploy.DefaultNamespace
+	}
```
renzodavid9, line 61: I was thinking, does it make sense to use the `runCtx.GetNamespace` method here? I know we are not using the `DefaultPipeline` there, instead we are going through all the pipelines to calculate the namespace, but maybe is ok to have the same behaviour, what do you think?

renzodavid9 (approved):

PR #8689, opened 2023-04-19
chore: Upgrade gcloud sdk version 1.39

ericzzzzzzz:
Fixes: #nnn 

**Description**


ericzzzzzzz:
The codecov key is missing causes ci fail .. should not block this pr.

Issue #8672, opened 2023-04-14
skaffold verify when using executionMode: kubernetesCluster {} can have duplicate log output


aaron-prindle:
Currently seeing an bug in `skaffold verify` in which duplicate log outputs can occur:

```
aprindle@aprindle-ssd ~/verify-demo/microservices-demo  [main]skaffold verify -a artifacts.json -f skaffold-fail.yaml
Tags used in verification:
 - integration-test-container -> gcr.io/aprindle-test-cluster/integration-test-container:v0.5.0-238-g1691e12-dirty@sha256:67bf59cf1ed43b8b71c834f0107abf15ada239a95f2ee8b59469ac59f183ba6b
 - index-page-test-container -> gcr.io/aprindle-test-cluster/index-page-test-container:v0.5.0-238-g1691e12-dirty@sha256:29340891b6165ac358e027eaef868c5e39897061d8a3fb92c1db20b4fa0c086d
[index-page-test-container]             <h3>Hot Products</h3>
[index-page-test-container] Index Page Test 1/1 Passed - index.html contained expected text: "Hot Products" 
[alpine-fail-2] this test will fail
[alpine-fail-1] this test will fail
[integration-test-container] Integration Test 1/4 Running ...
[integration-test-container] Integration Test 1/4 Passed!
[alpine-fail-1] this test will fail # <------------DUPLICATE-LOG-OUTPUT, only should print once for alpine-fail-1
[index-page-test-container]             <h3>Hot Products</h3>
[index-page-test-container] Index Page Test 1/1 Passed - index.html contained expected text: "Hot Products" 
[integration-test-container] Integration Test 1/4 Passed!
[integration-test-container] Integration Test 2/4 Running...!
[integration-test-container] Integration Test 2/4 Passed!
[integration-test-container] Integration Test 3/4 Running...!
[integration-test-container] Integration Test 3/4 Passed!
[integration-test-container] Integration Test 4/4 Running...!
[integration-test-container] Integration Test 4/4 Passed!
2 error(s) occurred:
* verify test failed: "alpine-fail-2" running job "alpine-fail-2" errored during run
* verify test failed: "alpine-fail-1" running job "alpine-fail-1" errored during run

```

PR #8695, opened 2023-04-20
feat: add wait group in docker and k8s job logger to avoid race condition

renzodavid9:
Fixes: #8672

**Description**
This PR adds a wait group in the k8s and docker logger to avoid a race condition when:
1. We have the loggers running, and
2. The execution of the containers/jobs finish, and 
3. We trigger the loggers stop to start draining their content
 
Before, we were draining the logs in the stop step without waiting for the logs triggered during execution to finish.


aaron-prindle (approved):
LGTM!

Issue #8628, opened 2023-04-03
[Custom Actions] Create new exec command


renzodavid9:
- New command to trigger a custom action: `exec <action-name>`
- We need to support the following flags:
  - Common flags
  - docker network
  - env file

PR #8696, opened 2023-04-20
Ca feature 8628

renzodavid9:
Fixes: #8628 

**Description**
This PR adds the logic to include the new `exec` command to execute a Custom Action together with the new v2 events to communicate the state of an `exec` run.


integration/exec_test.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func TestExec_ActionsEvents(t *testing.T) {
+	tests := []struct {
+		description  string
+		action       string
+		shouldErr    bool
+		expectedLogs []string
+	}{
+		// {
```
aaron-prindle, line 146: Should this be commented out?

renzodavid9: No, it shouldn't be commented. I just changed it, thanks!

cmd/skaffold/app/cmd/exec.go
```diff
@@ -0,0 +1,116 @@
...
+			}
+			return nil
+		}, doExec)
+}
+
+func doExec(ctx context.Context, out io.Writer, args []string) error {
+	return withRunner(ctx, out, func(r runner.Runner, configs []util.VersionedConfig) error {
+		buildArtifacts, err := getBuildArtifactsAndSetTagsForAction(configs, r.ApplyDefaultRepo, args[0])
+		if err != nil {
+			tips.PrintUseRunVsDeploy(out)
```
aaron-prindle, line 55: this tip message "run vs deploy" is likely not correct for the exec command

pkg/skaffold/actions/runner.go
```diff
@@ -72,8 +77,10 @@ func (r Runner) Exec(ctx context.Context, out io.Writer, allbuilds []graph.Artif
 
 	a := acs[0]
 
-	defer r.cleanup(context.TODO(), out, []Task{a}, []ExecEnv{execEnv})
-	return a.Exec(ctx, out)
+	err = a.Exec(ctx, out)
+	log.Entry(ctx).Debugf("Finished execution for %v", a.name)
+	r.cleanup(context.TODO(), out, []Task{a}, []ExecEnv{execEnv})
```
aaron-prindle, line 82: nit: should use the current `ctx` instead of `context.TODO()`?

aaron-prindle:
nit: I believe everytime 2+ actions are run together there will be a warning in stdout that liely shouldn't be there:
```
WARN[0009] action-succeeded cleanup error:removing stopped container: Error: No such container: 3d3a3398da733805ae819738f9fbce90ef04906ad50e16966681bafe531af372  subtask=-1 task=DevLoop
```

I think this is related to the docker purge issue we mentioned in the past, forget the status of that

Doesn't have to be addressed here, just wanted to point out this UX issue

aaron-prindle:
Just tried it out, works great!

pkg/skaffold/actions/runner.go
renzodavid9: You are right, fixed. Thanks!

renzodavid9:
> nit: I believe everytime 2+ actions are run together there will be a warning in stdout that liely shouldn't be there:
> 
> ```
> WARN[0009] action-succeeded cleanup error:removing stopped container: Error: No such container: 3d3a3398da733805ae819738f9fbce90ef04906ad50e16966681bafe531af372  subtask=-1 task=DevLoop
> ```
> 
> I think this is related to the docker purge issue we mentioned in the past, forget the status of that
> 
> Doesn't have to be addressed here, just wanted to point out this UX issue

Yes! That's exactly what's happening. I have a branch with the changes to improve this for the custom actions case (https://github.com/renzodavid9/skaffold/compare/ca-feature-8628...renzodavid9:skaffold:ca-issue-prune-err) where I split the docker client logic in another method so we don't run prune. This is to improve the custom actions case, but for a long term solution the idea is to use docker labels to know what to prune. My idea is to create the PR with the changes once this one gets merged.

aaron-prindle (approved):
LGTM!

Issue #8625, opened 2023-04-03
Currently `skaffold delete` does not have build artifacts info (& thus `envMap` vals) populated -> unexpected deletion failures


aaron-prindle:
The issue tracks a bug with `skaffold delete` where currently the command does not have all correct env values populated that it uses for generating the intermediate set of manifests it then cleans up:

From the `delete` command code:
```
func doDelete(ctx context.Context, out io.Writer) error {
	return withRunner(ctx, out, func(r runner.Runner, configs []util.VersionedConfig) error {
		manifestListByConfig, err := r.Render(ctx, io.Discard, []graph.Artifact{}, false)
		if err != nil {
			return err
		}
		return r.Cleanup(ctx, out, dryRun, manifestListByConfig, opts.Command)
	})
}
```
We can see that the `r.Render` call passes in an empty set of artifacts (`[]graph.Artifact{}`).  This is why currently the `envMap` used to populate variables in manifests (eg: when using `helm`) are empty for `skaffold delete`.  We should fix this by having `skaffold delete` correctly generate the same manifests that `skaffold` generating during creation

Bug was flagged from the information in the thread here:
https://kubernetes.slack.com/archives/CABQMSZA6/p1677781526359139

PR #8702, opened 2023-04-24
fix: delete does not working properly

ericzzzzzzz:
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/8625

 ## Description 
  - Skaffold delete command fails when using helm renderer and some templated fields are required, as skaffold delete calls render internally to get the targeted manifests. 
  - We need to call build method from runner to get the build artifacts then pass them to render method, otherwise the internal render call can fail if the templated fields are required.
  - A test added to guard this change.


aaron-prindle (approved):
LGTM!

PR #8710, opened 2023-04-26
fix: new `remove` method in docker client to use in custom actions and don't get the error related with prune

renzodavid9:
**Description**
This PR is to remove the warning message we get when running a custom action with more than one container, caused by a docker prune executed for each container.

### DIFF
--- a/pkg/skaffold/actions/docker/task.go
+++ b/pkg/skaffold/actions/docker/task.go
@@ -127,7 +127,8 @@ func (t Task) Cleanup(ctx context.Context, out io.Writer) error {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/skaffold/actions/docker/task.go
```diff
@@ -127,7 +127,8 @@ func (t Task) Cleanup(ctx context.Context, out io.Writer) error {
 		return nil
 	}
 
-	if err := t.client.Delete(ctx, out, id); err != nil {
+	t.client.Stop(ctx, id, nil)
```
aaron-prindle, line 130: Should this also be used in `verify` and possibly the docker deployer as well?

aaron-prindle (approved):
LGTM!

PR #8717, opened 2023-04-27
fix: resolve issue where skaffold logger could hang indefinitely if k8s job pod wasn't created

### LINKED ISSUE #8682: `skaffold verify` can hang indefinitely in deploying for certain errors (ex: deploying into target namspace that doesn't exist)
Issue tracks a bug where if `skaffold verify` is attempting to create a K8s job in a namespace that the user has specified but does not exist, the command will hang indefinitely (vs error stating the ns doesn't exist or possibly creating the ns)

EDIT:  Seems this can also occur when passing in `--rpc-port=0` & `--event-log-file=/dir/not/exist/foo.json`

aaron-prindle:
fixes #8682

### DIFF
--- a/pkg/skaffold/k8sjob/logger/log.go
+++ b/pkg/skaffold/k8sjob/logger/log.go
@@ -18,6 +18,7 @@ package logger
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"sync"
 	"sync/atomic"
@@ -154,7 +155,6 @@ func (l *Logger) streamLogsFromKubernetesJob(ctx context.Context, id, namespace 
 					return true, nil
 				}
 			}
-
 			var podName string
 			w, err := clientset.CoreV1().Pods(namespace).Watch(context.TODO(),
 				metav1.ListOptions{
@@ -164,13 +164,26 @@ func (l *Logger) streamLogsFromKubernetesJob(ctx context.Context, id, namespace 
 				return false, nil
 			}
 
-			for event := range w.ResultChan() {
-				pod, ok := event.Object.(*corev1.Pod)
-				if ok {
-					podName = pod.Name
-					// TODO(aaron-prindle) add support for jobs w/ multiple pods in the future
-					break
+			done := make(chan bool)
+			go func() {
+				for event := range w.ResultChan() {
+					pod, ok := event.Object.(*corev1.Pod)
+					if ok {
+						podName = pod.Name
+						done <- true
+						break
+					}
 				}
+			}()
+
+			select {
+			case <-ctx.Done():
+				fmt.Println("receive loop ends")
+				return false, fmt.Errorf("context cancelled for k8s job logging of pod of kubernetes job: %s", "id")
+			case <-done:
+				// Continue
+			case <-time.After(15 * time.Second): // Timeout after 15 seconds
+				return false, fmt.Errorf("timeout waiting for event from pod of kubernetes job: %s", id)
 			}
 
 			podLogOptions := &corev1.PodLogOptions{


pkg/skaffold/k8sjob/logger/log.go
```diff
@@ -164,13 +164,26 @@ func (l *Logger) streamLogsFromKubernetesJob(ctx context.Context, id, namespace
...
+						podName = pod.Name
+						done <- true
+						break
+					}
 				}
+			}()
+
+			select {
+			case <-ctx.Done():
+				fmt.Println("receive loop ends")
```
renzodavid9, line 181: Do we want a log message here? Can we use `log.Entry(ctx).Debug` or any of its variations instead?

pkg/skaffold/k8sjob/logger/log.go
```diff
@@ -164,13 +164,26 @@ func (l *Logger) streamLogsFromKubernetesJob(ctx context.Context, id, namespace
...
+					if ok {
+						podName = pod.Name
+						done <- true
+						break
+					}
 				}
+			}()
+
+			select {
+			case <-ctx.Done():
```
renzodavid9, line 180: I'm a little confuse with this `ctx.Done()`, is the intention here to cancel the context after the `for event := range w.ResultChan()` (line 169)? Or is to listen a cancellation  from other place (e.g, Ctrl+C)? If is the first case, we are missing to call a `cancel` function after the for loop.

renzodavid9:
I was trying the following example:
```
apiVersion: skaffold/v4beta5
kind: Config
verify:
  - name: alpine-1
    container:
      name: alpine-1
      image: alpine:3.15.4
      command: ["/bin/sh"]
      args: ["-c", "echo $FOO; sleep 10; echo bye"]
    executionMode:
      kubernetesCluster:
        overrides: '{ "metadata": {"namespace": "not-created-ns"}}'
```

`not-created-ns` is a namespace that is not created in my minikube cluster. Running `skaffold verify` with the previous configuration makes the program to never end. Not sure if this is another case not related with the one mentioned in the related issue (https://github.com/GoogleContainerTools/skaffold/issues/8682) or if I'm missing something 🤔 .

aaron-prindle:
I just attempted what you suggested and the call does timeout after the 30 second window added in the code here.  Did you wait 30s here?:

```
aprindle@aprindle-ssd ~/repro-8682 skaffold verify
Tags used in verification:
WARN[0032] <nil>                                         subtask=-1 task=DevLoop

WARN[0063] <nil>                                         subtask=-1 task=DevLoop
WARN[0063] verifier cleanup:cleaning up deployed job: jobs.batch "alpine-1" not found  subtask=-1 task=DevLoop
1 error(s) occurred:
* creating verify job in cluster: namespaces "not-created-ns" not found

```

pkg/skaffold/k8sjob/logger/log.go
aaron-prindle: Sorry, this was added unintentionally.  Deleted now

pkg/skaffold/k8sjob/logger/log.go
aaron-prindle: The intention is to listen for cancellation (the 2nd case)

renzodavid9:
> I just attempted the repro you suggested with the same skaffold.yaml + minikube and the call does timeout after the ~ 1 minute window added in the code here (working as intended). Did you wait ~1 min in your testing? The test that you did there is identical to the test I used for manual testing:
> 
> ```
> aprindle@aprindle-ssd ~/repro-8682 time skaffold verify
> Tags used in verification:
> WARN[0032] <nil>                                         subtask=-1 task=DevLoop
> WARN[0062] <nil>                                         subtask=-1 task=DevLoop
> WARN[0062] verifier cleanup:cleaning up deployed job: jobs.batch "alpine-1" not found  subtask=-1 task=DevLoop
> 1 error(s) occurred:
> * creating verify job in cluster: namespaces "not-created-ns" not found
> 
> real	1m3.037s
> user	0m0.611s
> sys	0m0.467s
> ```

Yes! I just tried again, I was not waiting long enough for the timeout. Thanks for checking!

renzodavid9 (approved):

PR #8727, opened 2023-05-01
fix: scanning filter not working properly due to version sorting

ericzzzzzzz:
**Description**

 - the script was using wrong flag for sort versions , this causes vul monitor misreporting, e.g. it's supposed to scan v2.0.10 instead of v2.0.9, 
 - We should use -V flag to compare version directly instead of comparing numeric values with -n flag.


aaron-prindle (approved):
LGTM!

PR #8737, opened 2023-05-02
fix: missing Sessionable ID in exported metrics

gsquared94:
This PR:
* adds a sessionable ID (`client_install_id`) set to a random GUID
* removes event fields not required or incorrectly set in current iteration

### DIFF
--- a/pkg/skaffold/instrumentation/firelog/exporter.go
+++ b/pkg/skaffold/instrumentation/firelog/exporter.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"os"
 
+	"github.com/google/uuid"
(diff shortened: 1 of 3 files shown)


aaron-prindle (approved):
LGTM!  Thanks for the quick fix here @gsquared94!

aaron-prindle:
Seems gcloud might have some breaking change it in, leading to kokoro failures.  From the kokoro error logs:
```
ERROR: gcloud crashed (AttributeError): 'bool' object has no attribute 'lower'
```

Recent thread regarding this error:
https://stackoverflow.com/questions/76159439/job-failing-with-error-gcloud-crashed-attributeerror-bool-object-has-no-at

Issue #8521, opened 2023-03-09
[Custom Actions] Create remote action


renzodavid9:
- Create k8s job action implementation

Issue #8560, opened 2023-03-17
Add `--custom-action-env-file`  flag to custom actions once work there is complete


aaron-prindle:
Issue tracks adding a `--custom-action-env-file`  flag to custom actions once work there is complete . 
 This flag should have near identical implementation as the `verify-env-file` flag added in the PR here https://github.com/GoogleContainerTools/skaffold/pull/8557

renzodavid9:
This issue is resolved with:
https://github.com/GoogleContainerTools/skaffold/pull/8755 -> For k8s actions
https://github.com/GoogleContainerTools/skaffold/pull/8662 -> For docker actions

PR #8755, opened 2023-05-08
feat: new k8s task and exec env for custom actions

renzodavid9:
Fixes: #8521 #8560
**Description**
This PR adds the logic to execute k8s custom actions as jobs inside the cluster.

### DIFF
--- a/pkg/skaffold/actions/runner.go
+++ b/pkg/skaffold/actions/runner.go
@@ -46,8 +46,8 @@ func NewRunner(execEnvByAction map[string]ExecEnv, orderedExecEnvs []ExecEnv, ac
 	return Runner{execEnvByAction, orderedExecEnvs, acsByExecEnv}
 }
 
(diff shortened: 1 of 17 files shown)


pkg/skaffold/actions/runner.go
```diff
@@ -79,7 +79,7 @@ func (r Runner) Exec(ctx context.Context, out io.Writer, allbuilds []graph.Artif
 
 	err = a.Exec(ctx, out)
 	log.Entry(ctx).Debugf("Finished execution for %v", a.name)
-	r.cleanup(ctx, out, []Task{a}, []ExecEnv{execEnv})
+	r.cleanup(context.TODO(), out, []Task{a}, []ExecEnv{execEnv})
```
renzodavid9, line 82: When the `ctx` is cancelled due to an event not related with the actions runner, e.g, a CTRL+C event, the cleanup fails, so, we need to send a `context.TODO()` to avoid this case.

aaron-prindle (approved):
LGTM!

PR #8842, opened 2023-06-01
chore: cherry-pick scanning marker change to release/v2.3

ericzzzzzzz:
**Related**
 - similar to #8834 for v2.3
 - lint checks were failing due to tests were against with wrong branch and there are schema version lags in projects in examples folder

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,6 +166,7 @@ release-build:
 		-f deploy/skaffold/Dockerfile \
 		--target release \
 		-t gcr.io/$(GCP_PROJECT)/skaffold:edge \
+		-t gcr.io/$(GCP_PROJECT)/skaffold:public-image-edge \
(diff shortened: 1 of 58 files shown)


plumpy (approved):

ericzzzzzzz:
> I was checking why does the `release/v2.0` branch tests were not failing with the same error about the schema version (`hack/check-schema-changes.sh` and `hack/check-samples.sh`). Looks like we just disabled those scripts, so the check is not run (Commit: [147129a](https://github.com/GoogleContainerTools/skaffold/commit/147129a3a98005d207beae996ffd9515424d7019)). Not sure if it make sense to do the same for this branch so we have the same solution for both branches, what do you think?
> 
> Also, it looks like `skaffold v2.0.x` uses the `v4beta1` 🤔 schema

oh I'm fine with either way.

Issue #8545, opened 2023-03-11
skaffold builds multiple images (one for each change) 😢


cheburakshu:
skaffold build generates multiple images, one for each commit (or with every change), is there a way to prevent this behavior? One time it built images worth 60 GB due to multiple builds/rebuilds. I just want one build (image:latest)

```
⟩ docker images my-service
REPOSITORY     TAG                                                                IMAGE ID       CREATED      SIZE
my-service   ef5fa6514a6c0ebdb16d04063c1da831f3f469647a146ba589108c68b96df693   ef5fa6514a6c   2 days ago   1.39GB
my-service   latest                                                             ef5fa6514a6c   2 days ago   1.39GB
my-service   e77a7046e44df826b930552e3a35c6046a98472e69abdcfc4b05c49902971c32   e77a7046e44d   2 days ago   1.39GB
my-service   cdec8e40e1d2c045a317a710cd3d29a3fa7b4b4f82f2b200264b249e1ef1bb45   cdec8e40e1d2   4 days ago   1.39GB
my-service   33b7444e0f0a077f31db8becb7308d1d609ab5b5b60cd349c0b1d0646a7ccd21   33b7444e0f0a   4 days ago   1.39GB
my-service   6765a190b64805606987ed38ec2bd4946209c3561e0119e04cf792c38d69e8b2   6765a190b648   4 days ago   1.39GB
my-service   d3892392900d518c4699f4af57e2269c02736eba030275b43b72685811eba62b   d3892392900d   4 days ago   1.39GB
my-service   0cc898c05689fbce8e183f6f62ce70622bc10bb78de23f2b9d0ff8b9829f866d   0cc898c05689   4 days ago   1.39GB
my-service   4ac1ffc4241f4148e9a637faba7ee5e3068d20cb4611a20c8445eba222370197   4ac1ffc4241f   4 days ago   1.39GB
my-service   b4b984de04c808110c9de8fed29643a990cc673a75720856cc0de08fe5282a9f   b4b984de04c8   6 days ago   1.38GB
my-service   ef86fb7e0726dd023e8f37d8224b0f88431cac3751b2f5e0392a2f1d2149ba2b   ef86fb7e0726   6 days ago   391MB
my-service   e76caba733c6b21a2b8a6234e748ca573f530e87df6a6f5446db8013355941e4   e76caba733c6   6 days ago   391MB
[...]

PR #8897, opened 2023-06-16
fix: Clean up dev images except the last

ericzzzzzzz:
Fixes: #8545 

**Description**
 - Clean up all images created during skaffold dev loop except the last one in local environment.  

**Test Plan**
 - Run `skaffold dev` then modify some file to trigger a few rebuilds, check your built images with `docker image ls` or `minikube image ls` if you're using minikube

### DIFF
--- a/pkg/skaffold/build/local/local.go
+++ b/pkg/skaffold/build/local/local.go
@@ -89,6 +89,11 @@ func (b *Builder) buildArtifact(ctx context.Context, out io.Writer, a *latest.Ar
 	}
 
 	imageID := digestOrImageID
+	artifacts, err := b.artifactStore.GetArtifacts([]*latest.Artifact{a})
+	// delete previous built images asynchronously
+	if b.mode == config.RunModes.Dev && len(artifacts) > 0 {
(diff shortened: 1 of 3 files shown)


pkg/skaffold/build/local/local.go
```diff
@@ -57,6 +57,21 @@ func (b *Builder) PostBuild(ctx context.Context, _ io.Writer) error {
 	return nil
 }
 
+// Prune uses the docker API client to remove all images built with Skaffold
+func (b *Builder) Prune(ctx context.Context, _ io.Writer) error {
```
renzodavid9, line 61: I noticed that running `skaffold dev --cache-artifacts=false --no-prune=false` triggers a prune logic that calls this function, that goes through all the images built by Skaffold and prune them, due to the new prune logic in line 107, it doesn't find the images (were already deleted) and now it shows a warning message doing the clean up. Do you think that message is something we should avoid to print now?

ericzzzzzzz: ahhhhh! Great catch! updated! Now we use localpruner to do the cleanup job which should track deleted images, so when the devloop finishes, we won't see the warnings.

renzodavid9 (approved):

Issue #8901, opened 2023-06-20
import apply-setter function to skaffold to replace kpt command call for parameterization 


ericzzzzzzz:
- https://github.com/GoogleContainerTools/skaffold/pull/8365 implemented initial parameterization with kpt native approach by applying krm functions.
 - The downside is that approach requires pulling images with external network, CD(continuous deployment) pipelines may not have that configured.
 - The apply-setter is no longer under maintained.  
 - Also, pulling images can be slow  as to the normal render phase 
 - In addition, if we import that function directly to skaffold, people who use paramaterization for kustomize, kubectl don't need to have kpt installed. 
 - We're also able to rename the commenter syntax, 

## Ask
 - The Ask here is to change the way we parameterize manifests when using kustomize render or raw manifests by using apply-setter function in skaffold directly. 
 - We should also use the new name for replacement commenter, so users won't be under the impression that they need ``kpt`` for parameterization.

ericzzzzzzz:
Here is the apply-setter function:  https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/functions/go/apply-setters

PR #8902, opened 2023-06-21
chore: port apply-setter krm function over to skaffold

ericzzzzzzz:
Fixes: #8901  

 - port over  https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/functions/go/apply-setters to skaffold to avoid external calls for parameterization
 - rename kpt-set to from-param
 - the change only support for rawmanifest and kustomizer render, template commenter doesn't work for helm for now it will be done in a separate pr.

### DIFF
--- a/integration/render_test.go
+++ b/integration/render_test.go
@@ -1151,41 +1151,6 @@ spec:
     image: skaffold-example
 `,
 		},
-		{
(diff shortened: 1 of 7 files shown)


ericzzzzzzz:
it seems that kokoro job is completely broken now.

renzodavid9 (approved):

PR #8905, opened 2023-06-22
fix: add docker to the LTS container images

plumpy:
This fixes an issue where kpt functions wouldn't work in rendering since they depend on docker to run.


ericzzzzzzz (approved):

ericzzzzzzz:
kokoro job is broken, we can ignore that at the moment.

PR #8911, opened 2023-06-23
feat: support tempalte parameterization for helm

ericzzzzzzz:
Fixes: #nnn 

**Description**


renzodavid9 (approved):

adriantr:
@ericzzzzzzz can you confirm that this should work with `skaffold render --set <param>=<value>` with the latest skaffold version?

I'm having troubles with getting it to work:
```
- name: 'ADRIAN_TEST'
  value: {{ .Values.backendTimeout | quote }} # from-param: ${adrian}
```

still ends up with templated value from .Values instead of the one passed through --set

ericzzzzzzz:
hi @adriantr 

 To use template parameterization for helm 
 - the render command must be passed with --build-artifacts flag to provide build results for post render stage,  as the replacement happens on helm post render stage and if no build artifacts are provided the stage won't happen.
 - in skaffold config, manifests stanza must be used for helm  instead of relying on skaffold to populate information under deploy sanza to manifests stanza -- The population is a historial problem.

PR #8917, opened 2023-06-26
chore: push scanning marker tag to remote

ericzzzzzzz:
- tracked internal 
 - We need use a tag with public-image prefix for skaffold images so vulnerabilities scanner can scan those.
 - That tag is added to the image but never gets pushed to remote repo.
 - so configuring it to push the tag to remote.

### DIFF
--- a/deploy/cloudbuild-release-lts.yaml
+++ b/deploy/cloudbuild-release-lts.yaml
@@ -61,6 +61,7 @@ steps:
 
 images:
 - 'gcr.io/$PROJECT_ID/skaffold:$TAG_NAME-lts'
+- 'gcr.io/$PROJECT_ID/skaffold:$_SCANNING_MARKER-lts'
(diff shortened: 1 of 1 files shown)


ericzzzzzzz:
test fails due to writeFilesToTar timeouts, the method write each file in a separate go-routine , when timeouts happen, the go-routine continues to write and tar.Close() also gets called in the main go-routine, that method also writes 'end marker' to the underlying buffer, which causes concurrent write with the regular file-writing go-routine. This should not block this pr.

renzodavid9:
> test fails due to writeFilesToTar timeouts, the method write each file in a separate go-routine , when timeouts happen, the go-routine continues to write and tar.Close() also gets called in the main go-routine, that method also writes 'end marker' to the underlying buffer, which causes concurrent write with the regular file-writing go-routine. This should not block this pr.

Awesome! I've seen that flaky test before, now we know what's happening 😄 , I'll create an issue with this explanation to follow-up the work there. Thanks for checking what's happening.

renzodavid9 (approved):

PR #8942, opened 2023-07-13
chore: disable edge image scanning

ericzzzzzzz:
-  As team discussed, we would like to disable edge image vulnerabilities scanning due to the scope is too large and proper scope has already been covered by lts images.


ericzzzzzzz:
merged this as the change should not affect integration tests.

PR #8944, opened 2023-07-13
fix: Fix typo in Cloud Run log tailing

bskaplan:
Switch & to % so we properly rotate through all of the colors. This matches the behavior in https://github.com/GoogleContainerTools/skaffold/blob/453b101d70ccbf75c98f945516589950c87f63a0/pkg/skaffold/output/colorpicker.go#L52


ericzzzzzzz:
the ci is broken,  the change is very small and should not be blocked by broken ci tests.

PR #8953, opened 2023-07-17
chore: update the way the LTS images are built

plumpy:
I moved everything from `Dockerfile.deps.lts` into `Dockerfile.lts` directly. Unforunately, this makes this diff pretty useless, but basically the only changes were the `download-skaffold` section at the top and the removal of the Go installation (the last two lines).


deploy/cloudbuild-release-lts.yaml
```diff
@@ -7,52 +7,15 @@ steps:
...
-  - name: gcr.io/cloud-builders/gcloud
-    entrypoint: 'bash'
-    args: ['deploy/setup-secret.sh','-p', $PROJECT_ID]
-
-# Build and tag skaffold builder
-  - name: 'gcr.io/cloud-builders/docker'
-    args:
-    - 'build'
-    - '--cache-from'
-    - 'gcr.io/$PROJECT_ID/skaffold-builder:latest'
```
ericzzzzzzz, line 27: I think we need to add 
```yaml
 - '--build-arg'
 - 'SKAFFOLD_VERSION=$TAG_NAME'
```

to inject version to the build context

deploy/cloudbuild-release-lts.yaml
```diff
@@ -7,52 +7,15 @@ steps:
...
+    - 'gcr.io/$PROJECT_ID/skaffold:$TAG_NAME-lts'
     - '-t'
-    - 'gcr.io/$PROJECT_ID/skaffold-builder:latest'
+    - 'gcr.io/$PROJECT_ID/skaffold:$_SCANNING_MARKER-lts'
+    - '-t'
+    - 'us-east1-docker.pkg.dev/$PROJECT_ID/scanning/skaffold:$TAG_NAME-lts'
```
ericzzzzzzz, line 14: This means, artifact registry scanner will scan all binary deps, probably we want to do it anyway. Just double check this is intended behavior?

ericzzzzzzz (commented):
Maybe test it with a person project by connecting gcb with your fork ?

deploy/cloudbuild-release-lts.yaml
ericzzzzzzz: oh.. seems that our internal scanner is already doing this may be we don't need to push the image to ar anymore

deploy/cloudbuild-release-lts.yaml
plumpy: Ah, sorry, I was testing this with a modified version of this file (mostly because I didn't want to push to AR too) and then I forgot to copy my fixes back into this file. Fixed!

deploy/cloudbuild-release-lts.yaml
plumpy: Okay, I removed AR. I wasn't sure why that was there but didn't want to change it...

plumpy:
> Maybe test it with a person project by connecting gcb with your fork ?

I testing it by doing this:
`gcloud builds submit --project=$MYPROJECT --substitutions=TAG_NAME=v2.6.1,_SCANNING_MARKER=public-image-2.6 --config=deploy/cloudbuild-lts.yaml`

Do you think that's good enough or should I wire up a trigger?

ericzzzzzzz (approved):

Issue #7613, opened 2022-06-30
Skaffold verify hangs when using --event-log-file w/ dir that doesn't exist 


aaron-prindle:
Skaffold verify hangs when using --event-log-file w/ dir that doesn't exist.  Other commands (eg: render, etc) don't hang out display an error

```
open /ccd-start-project/somerandom/output: no such file or directory
```

PR #8961, opened 2023-07-19
fix: verify hangs if event-logs-file does not exist (#7613)

Darien-Lin:
Fixes: #7613 

**Description**

When running `skaffold verify` with rpc-port and --event-log-file, if the event-log-file references a path to a file where the directory doesn't exist, the command hangs indefinitely. The changes I am proposing ensure that the directories to event-log-file exists before attempting to create the file and dump the logs. The relevant unit tests are also adjusted with the assumption that the directory does not exist. 

In addition, SaveLastLog seems to follow the same issue, which is also fixed in this PR.   

0700 is used since 'execute' permission is needed on the directories. Execute permissions on directory are so that files can be accessed in those nested directories.

### DIFF
--- a/pkg/skaffold/event/event.go
+++ b/pkg/skaffold/event/event.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"path/filepath"
 	"sync"
 
 	//nolint:golint,staticcheck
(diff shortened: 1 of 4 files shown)


Darien-Lin:
hm, looking into the permissions error. Weird that the file cannot be created in the directory that we have created.

Darien-Lin:
https://unix.stackexchange.com/questions/21251/execute-vs-read-bit-how-do-directory-permissions-in-linux-work/21252#21252 

changed to use 0700 instead of 0600 for the directory

ericzzzzzzz (approved):

Issue #7687, opened 2022-07-26
skaffold render does not support overrides?


mcg1969:
### Expected behavior

`skaffold render` should take into account not just the `setValues` entries but the `overrides` as well, just as `skaffold run` does.

### Actual behavior

The `overrides` section is not included in the generated `helm template` command.

### Information

- Skaffold version: v1.39.1
- Operating system: macOS Monterey / Minikube
- Installed via: skaffold.dev

I'm afraid I cannot share the contents of my skaffold.yaml. What I can share however is that the generated helm command looks as follows:

```
helm --kube-context minikube template <CHART_DIR> \
    <--set-string image.[NAME]=[TAG] commands for each image>
    <lots of --set commands for my setValues section>
    --namespace default
```

In contrast, the `skaffold run` command generates a temporarily file called `skaffold-overrides.yaml` and generates a helm command structured as follows:

```
helm --kube-context minikube install <RELEASE_NAME> <CHART_DIR> \
    --namespace default \
    <--set-string image.[NAME]=[TAG] commands for each image> \
    <lots of --set commands for my setValues section> \
     -f skaffold-overrides.yaml
```

gsquared94:
Can you provide the `skaffold render` and `skaffold run` commands  along with the flags that you're executing?

mcg1969:
Yep, just `skaffold run --namespace default` and `skaffold render --namespace default`.

mcg1969:
I have found that `skaffold run --namespace default --render-only`, which is of course deprecated, does not supply the overrides either.

jkleckner:
I was scripting some render operations and was frustrated that the `--kube-context` argument doesn't work with render.

I don't really want to modify my kube context globally just to perform a render.

skrech:
I've encountered the same issue with Skaffold 2.1.0 as well. When using `skaffold render` `overrides` section in `deploy/helm/releases/<release-item>` doesn't seem to be applied, while using skaffold dev/run works as expected.

PR #8966, opened 2023-07-21
feat: support overrides in helm renderer

jackwilsdon:
Fixes: #7687 

**Description**
Adds support for the `manifests.helm.releases[].overrides` field. Previously this was only supported when using Helm as a deployer.

The majority of this change is moving the body of the `range h.config.Releases` loop into a separate function so that the `defer` is run after each release. Viewing the changes with whitespace hidden makes it a bit easier to read: https://github.com/GoogleContainerTools/skaffold/pull/8966/files?w=1

### DIFF
--- a/pkg/skaffold/deploy/helm/helm_test.go
+++ b/pkg/skaffold/deploy/helm/helm_test.go
@@ -270,6 +270,16 @@ var testDeployCreateNamespaceConfig = latest.LegacyHelmDeploy{
 	}},
 }
 
+var testDeployNoOverridesConfig = latest.LegacyHelmDeploy{
+	Releases: []latest.HelmRelease{{
+		Name:      "skaffold-helm",
+		ChartPath: "examples/test",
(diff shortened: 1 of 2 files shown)


jackwilsdon:
I'm not sure that failing integration test is related to the changes I made here.

renzodavid9 (approved):

renzodavid9:
> I'm not sure that [failing integration test](https://github.com/GoogleContainerTools/skaffold/actions/runs/5624027420/job/15879119243?pr=8966#step:17:2414) is related to the changes I made here.

Hey @jackwilsdon, I triggered the tests again and they worked, so looks like we should be fine.

Thanks for the PR! It looks good. I think we can add a new integration test that uses `override` with `skaffold render`; I'll merge this and will create an issue to follow-up with the integration test. Thanks! 😄

mcg1969:
Tremendous! Thank you very much @jackwilsdon

PR #8984, opened 2023-08-03
chore: remove latest tagging from release process v1.39

ericzzzzzzz:
Fixes: #nnn 

**Description**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,6 @@ release: $(BUILD_DIR)/VERSION
 		--build-arg VERSION=$(VERSION) \
 		-f deploy/skaffold/Dockerfile \
 		--target release \
-		-t gcr.io/$(GCP_PROJECT)/skaffold:latest \
 		-t gcr.io/$(GCP_PROJECT)/skaffold:$(VERSION) \
 		.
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -149,7 +149,6 @@ release: $(BUILD_DIR)/VERSION
 		--build-arg VERSION=$(VERSION) \
 		-f deploy/skaffold/Dockerfile \
 		--target release \
-		-t gcr.io/$(GCP_PROJECT)/skaffold:latest \
```
plumpy, line 152: I think you also need to remove it from `cloudbuild-release.yaml` or else it will complain that the image with that tag doesn't exist.

ericzzzzzzz: I did this when upgrading v1 deps in this commit https://github.com/GoogleContainerTools/skaffold/pull/8621/commits/1d71b41505b1129cb98796b1a09dd8c74798d0a5#diff-6547f8c009c546989c6d4d936f8eecddf7815e8a40e1924e9a67fec8b24d7132 , 
but I forgot why I did that 🤦

ericzzzzzzz: full pr : https://github.com/GoogleContainerTools/skaffold/pull/8621

plumpy (approved):

Issue #9006, opened 2023-08-10
Multiple containers in docker deploy result in extra ports being forwarded


jcscottiii:
I am seeing the same/similar issue reported in https://stackoverflow.com/q/73533579.
Look there for more details.

### Expected behavior

Given the following config, I expect two ports to be forwarded. One for bert and one for ernie.

```yaml
...
portForward:
- resourceType: container
  resourceName: bert
  port: 8080
  localPort: 9000
- resourceType: container
  resourceName: ernie
  port: 8081
  localPort: 9001
```

Expected output

```sh
[bert] Forwarding container port 8080 -> local port http://127.0.0.1:9000
[ernie] Forwarding container port 8081 -> local port http://127.0.0.1:9001
```

### Actual behavior

In the output below, there are four ports exposed. Two for each container. Also the second container is exposed correctly because the first container used the port. As I add more services that are port forwarded, this problem increases.

```sh
$ skaffold dev
WARN[0000] failed to detect active kubernetes cluster node platform. Specify the correct build platform in the `skaffold.yaml` file or using the `--platform` flag  subtask=-1 task=DevLoop
Generating tags...
 - bert -> bert:bc4100080
 - ernie -> ernie:bc4100080
Checking cache...
 - bert: Found Locally
 - ernie: Found Locally
Tags used in deployment:
 - bert -> bert:ccb36787e044aaf4fc46cc0b3282771c73724e7405da67e4794e91cf25317644
 - ernie -> ernie:a85231d544a06b0bbdbe8c7603b3b14488f8611f952148a4d6b37d1ee46830a2
Starting deploy...
[bert] Forwarding container port 8080 -> local port http://127.0.0.1:9000
[bert] Forwarding container port 8081 -> local port http://127.0.0.1:9001
[ernie] Forwarding container port 8080 -> local port http://127.0.0.1:9002
[ernie] Forwarding container port 8081 -> local port http://127.0.0.1:9003
Listing files to watch...
 - bert
 - ernie
Press Ctrl+C to exit
Watching for changes...
[ernie] Hey there Bert!
[bert] Oh hey Ernie.
```

### Information

- Skaffold version: v2.6.0
- Operating system: Linux
- Installed via: skaffold.dev 
- Contents of skaffold.yaml: 

https://github.com/jcscottiii/skaffold/blob/1a747b1b32c7a70d3f0f733374cd1526dbf34828/examples/docker-deploy/skaffold.yaml

[...]

PR #9008, opened 2023-08-10
fix: #9006 - Filter port forwarding resources for docker deploy

jcscottiii:
Fixes: #9006

**Description**

Add a filter function to only send port forwarding resources for that particular container into AllocatePort


pkg/skaffold/deploy/docker/deploy_test.go
```diff
@@ -113,6 +119,44 @@ func TestDebugBindings(t *testing.T) {
 					},
 				},
 			},
+			portForwardResources: nil,
+		},
+		{
+			name: "two artifacts two bindings and two port forward resources",
```
jcscottiii, line 125: Initially, I added a test case to the TestDebugBindings function since it was the only function here. However, I did end up adding a separate function `TestFilterPortForwardingResources` to test specifically that. Let me know if I should revert the changes in this TestDebugBindings function.

Issue #8462, opened 2023-02-23
Add support for user supplied CRDs to be added to our `validations.go` logic (similar to bespoke ConfigController support currently but make it dynamic)


aaron-prindle:
Issue tracks adding support for user supplied CRDs to be added to our `validations.go` logic (similar to bespoke ConfigController support currently but make it dynamic).  With this, users could  supply an allowlist of CRDs that skaffold would watch dynamically as long as they support the `kstatus` standard:
https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md

ericzzzzzzz:
Hi @aaron-prindle  could you help me understand the ask here? Is this asking to provide a way for CRDs statuscheck like ConfigConnector here https://github.com/GoogleContainerTools/skaffold/blob/88e1c1734ea0f1bd79a0f24783f9565057a0e46a/pkg/diag/validator/config_connector.go#L1-L87 in `validator` package but make it dynamic like providing CRDs through command line or in a skaffold config stanza?   Thank you :)

PR #9016, opened 2023-08-11
feat: crd status check

ericzzzzzzz:
Fixes: #8462
**Description**
 - users can provide resource selectors through file for their crds status check 

**Test Plan**
  
 - I don't know any other crd controllers than ConfigConnector that implements kstatus, you can test other crds if you know those 
 - checkout this branch , comment out this line https://github.com/GoogleContainerTools/skaffold/blob/700fd81d5ee28977c1bc0da6c34336925cdc205b/pkg/skaffold/kubernetes/manifest/selector.go#L62 (otherwise, the code will not use customize resource validator) and build skaffold
 - follow this guide to set up Config Connector https://cloud.google.com/config-connector/docs/how-to/getting-started
 -  download this tar
[test-crds-status-check.tar.gz](https://github.com/GoogleContainerTools/skaffold/files/12323816/test-crds-status-check.tar.gz) and untar it 
 - enter the result folder and run `skaffold dev --status-check-selectors selectors.json` should see status check against the crd, the status check against ConfigConnector will hang until check success, this is expected as config controller returns IN_PROGRESS for both resource reconciling and failing.


renzodavid9 (approved):

renzodavid9:
LGTM! I was just wondering if there is a way to have the CRDs logic to manage the ConfigConnector case, I saw we are using different errors codes, but I guess we can discuss it if make sense. Thanks!

Issue #8948, opened 2023-07-14
Vulnerability of dependency "github.com/moby/buildkit"


Silence-worker-02:
Hello, we are a team researching the dependency management mechanism of Golang. During our analysis, we came across your project and noticed that it contains a vulnerability (CVE-2023-26054). In your project, the github.com/moby/buildkit package is being used at version github.com/moby/buildkit v0.8.0, but the patched version is v0.11.4. To fix the vulnerability, we recommend modifying the go.mod file to update the version to v0.11.4 or higher. Thank you for your attention to this matter.

PR #9043, opened 2023-08-23
chore: upgrade ko

ericzzzzzzz:
Fixes: #8948 

**Description**
 - upgrade ko and moby/buildkit 
 - Resolve conflicts: 
    - the AST returned from parsing dockerfile with docker/parser.Parse uses exact string for token command now, originally the AST converts command token to lower case, so we need to convert it to lowercase manually now when comparing tokens. 
    - Other struct breaking changes:
      - [buildpack lifecycycle realted](https://github.com/buildpacks/lifecycle/pull/1076/files)
      - [docker api related ](https://github.com/moby/moby/pull/43530/files)
      - [buildpack process struct](https://github.com/buildpacks/lifecycle/pull/920/files#diff-571aefdf18ffa8fd370aeed80f357f99a92a3e78f1949b5e236132a3af511ea8 )
 - clean up go.mod , some exclude clauses don't make sense anymore, as we're already on a later version.


pkg/skaffold/docker/image.go
```diff
@@ -719,7 +719,12 @@ func (l *localDaemon) Prune(ctx context.Context, images []string, pruneChildren
 }
 
 func (l *localDaemon) Stop(ctx context.Context, id string, stopTimeout *time.Duration) error {
-	if err := l.apiClient.ContainerStop(ctx, id, stopTimeout); err != nil {
+	// - Use '-1' to wait indefinitely.
```
ericzzzzzzz, line 722: https://github.com/moby/moby/blob/24.0/api/types/container/config.go#L28

joeyslalom:
Thank you! As a user of ko with skaffold, this will help me update my project to Go 1.20 <3

pkg/skaffold/docker/image.go
```diff
@@ -719,7 +719,12 @@ func (l *localDaemon) Prune(ctx context.Context, images []string, pruneChildren
 }
 
 func (l *localDaemon) Stop(ctx context.Context, id string, stopTimeout *time.Duration) error {
-	if err := l.apiClient.ContainerStop(ctx, id, stopTimeout); err != nil {
+	var so container.StopOptions
+	if stopTimeout != nil {
+		seconds := int(stopTimeout.Seconds())
+		so.Timeout = &seconds
```
renzodavid9, line 725: Can we use the util function here instead?:
```
so.Timeout = util.Ptr[int](int(stopTimeout.Seconds()))
```

renzodavid9 (approved):

Issue #9056, opened 2023-08-30
[FR] Allow specifying a remote config dependency from Google Cloud Storage


mattsanta:
Expand [remote config dependency feature](https://skaffold.dev/docs/design/config/#remote-config-dependency) to allow specifying a config stored in Google Cloud Storage.

PR #9057, opened 2023-08-30
feat: allow specifying a remote config dependency from Google Cloud Storage

mattsanta:
**Related**:  https://github.com/GoogleContainerTools/skaffold/issues/9056

**Description**
Introduces the ability to specify a remote config dependency that is stored in Google Cloud Storage (GCS). The current feature only supports remote configs stored in git repositories.

Skaffold config has new fields to support configuring the GCS remote config dependency. Also updated the docs on the feature at https://skaffold.dev/docs/design/config/#remote-config-dependency.


pkg/skaffold/gcs/gsutil.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+	}
+	// If sync property is false then skip fetching latest object from remote storage.
+	if g.Sync != nil && !*g.Sync {
+		return cacheDir, nil
+	}
+	// If sync is turned off via flag `--sync-remote-cache` then skip fetching latest object from remote storage.
+	if opts.SyncRemoteCache.FetchDisabled() {
+		return cacheDir, nil
+	}
```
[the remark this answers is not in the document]
mattsanta: Good find - thanks! It turns out I was checking if sync was disabled even if the cache didn't exist but should only be checking if it does exist. Updated the PR.

pkg/skaffold/parser/config.go
```diff
@@ -231,7 +235,9 @@ func processEachConfig(ctx context.Context, config *latest.SkaffoldConfig, cfgOp
 				}
 			}
 		}
-		newOpts := configOpts{file: cfgOpts.file, profiles: depProfiles, isRequired: required, isDependency: cfgOpts.isDependency}
+		isRemoteDependency := d.GitRepo != nil || d.GoogleCloudStorage != nil
```
renzodavid9, line 238: Not sure but, is this needed? I think we are always using the value from the flag in line 282 (isRemoteCfg), and we are overwriting the value from here(?)

renzodavid9:
Thanks @mattsanta! Just one more small check. Also I was thinking that these remote dependencies implementations (Git and GCS) are sharing some logic, (e.g, if we should download or not the dependency according to the different flags values and current state of the remote cache folder), do you think we should find a way to unify everything? I think is ok if we don't do it in this PR, but just curious about your opinion 🤔

renzodavid9 (approved):

pkg/skaffold/parser/config.go
mattsanta: You're right, removed. Thanks!

mattsanta:
> Thanks @mattsanta! Just one more small check. Also I was thinking that these remote dependencies implementations (Git and GCS) are sharing some logic, (e.g, if we should download or not the dependency according to the different flags values and current state of the remote cache folder), do you think we should find a way to unify everything? I think is ok if we don't do it in this PR, but just curious about your opinion 🤔

Yeah, I think this would be valuable especially as more remote dependencies are added as well. To be honest, I wanted to limit the amount of changes that I made in a new code base that I'm unfamiliar with and felt like this already had a good amount of changes.

renzodavid9:
> > Thanks @mattsanta! Just one more small check. Also I was thinking that these remote dependencies implementations (Git and GCS) are sharing some logic, (e.g, if we should download or not the dependency according to the different flags values and current state of the remote cache folder), do you think we should find a way to unify everything? I think is ok if we don't do it in this PR, but just curious about your opinion 🤔
> 
> Yeah, I think this would be valuable especially as more remote dependencies are added as well. To be honest, I wanted to limit the amount of changes that I made in a new code base that I'm unfamiliar with and felt like this already had a good amount of changes.

Yeap, I agree, this could be a improvement for the future then. Thanks!

Issue #9028, opened 2023-08-16
useDockerCli flag should enforce all operations in scope of docker builder to use command line


ericzzzzzzz:
### Expected behavior
 - when useDockerCli is set to true in build.local stanza, skaffold should use cli for all operations that fall into builder's scope, that includes, build, getImageID, push, purge images etc,

### Actual behavior
 - only `build` is used with cli when the flag is specified. 

### Information
 - By default, skaffold uses golang docker RESTapi bindings to talk to docker engine, building with docker might not be successful if DOCKER_HOST is not set to default value `unix:///var/run/docker.sock` , using docker command line should be a way to fix this, however, due to other operations such as getImageID from docker engine also uses the bindings, the skaffold Build phase will fail after building is done. 
 - Due to the how DockerApiClient in skaffold -- a singleton is set when constructing skaffold runner,  a hook configured for registry authentization with docker login will not work, using cli should be the way to fix it. However, due to related operations are not using cli when useDockerCli is set to true, the hook will not work as expected.

ericzzzzzzz:
related #7078

ericzzzzzzz:
All these problems are caused by golang dockerClient uses "unix:///var/run/docker.sock" if HOST_DOCKER env is not set, the command line behaves in different way, if  HOST_DOCKER is not set the docker command line tool will use `the connection from current context` as the host. I think we may take a different approach other than re-implementing golang dockerClient apis with command line, we may just get the right host when initiating docker client in skaffold.

ericzzzzzzz:
eval echo $(docker context inspect --format='{{json .Endpoints.docker.Host}}')   to get the docker host 

this command may fail, if a user uses docker as podman alias, in that case, we may do a second try with `docker machine inspect --format '{{.ConnectionInfo.PodmanSocket.Path}}'` to get the host, if neither of them works, goes back to the original behavior.

ericzzzzzzz:
`podman` will be handled by different logic

PR #8349, opened 2023-01-22
feat: add support for remote docker over ssh

catusax:
Fixes: #5703

**Description**
this pr adds support for remote docker over ssh using [connhelper package](https://github.com/docker/cli/tree/master/cli/connhelper), the feat was introduced by docker cli by this commit:  https://github.com/docker/cli/pull/1014
 

### TestCase

open google cloud shell: 

https://ssh.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https://github.com/catusax/skaffold&cloudshell_git_branch=docker-ssh-support&cloudshell_open_in_editor=examples/cross-platform-builds/skaffold.yaml

```shell

# setup ssh
ssh-keygen
cat ~/.ssh/id_rsa.pub > ~/.ssh/authorized_keys
chmod 640 ~/.ssh/authorized_keys
sudo service ssh restart
ssh-copy-id 127.0.0.1
export DOCKER_HOST=ssh://127.0.0.1

# setup docker
export DOCKER_USER=<your dockerhub userid>
docker login -u $DOCKER_USER

export ROOT=$PWD && export PROJECT=$ROOT/examples/cross-platform-builds
cd $PROJECT

sed -i 's/skaffold-example/${DOCKER_USER}\/skaffold-example/g' skaffold.yaml
skaffold build # this will fail
# Cannot connect to the Docker daemon at ssh://127.0.0.1. Is the docker daemon running?. Docker build ran into internal error. Please retry.
# If this keeps happening, please open an issue..

# build new patch
cd $ROOT && LOCAL=true make 
sudo cp out/skaffold /usr/bin/skaffold

# build with ssh support
cd $PROJECT
skaffold build # success

# without ssh
unset DOCKER_HOST
skaffold build # success

```

PR #9094, opened 2023-09-14
chore: add set docker host by current context

ericzzzzzzz:
Fixes: #9028

 - The original ask from #9028 is to enforce all operations under a builder scope to use cli when useCli is set to true when using docker, but after revisiting several issues linked in #9028, the core problem we want to fix is that when DOCKER_HOST env is not set,  `golang docker rest api client` use a hard coded value as docker engine host for example, ``unix:///var/run/docker.sock`` on mac, ``but docker command line client`` use the socket from `docker current context` as host. 
 - This pr changes the hard coded value by inspecting the host of `docker current context` with docker client, if the DOCKER_HOST is not set, we should use the value from `docker current context`.  If the value is set, both docker command line and golang rest client will honor the env variable. 
 - This approach should solve the linked issues in #9028, and much easier to maintain than duplicating similar logics with command line client.

### DIFF
--- a/examples/getting-started/skaffold.yaml
+++ b/examples/getting-started/skaffold.yaml
@@ -3,6 +3,8 @@ kind: Config
 build:
   artifacts:
   - image: skaffold-example
+  local:
+    push: false
 manifests:
   rawYaml:
   - k8s-pod.yaml
--- a/examples/getting-started/skaffold.yaml
+++ b/examples/getting-started/skaffold.yaml
@@ -3,8 +3,6 @@ kind: Config
 build:
   artifacts:
   - image: skaffold-example
-  local:
-    push: false
 manifests:
   rawYaml:
   - k8s-pod.yaml
--- a/pkg/skaffold/docker/client.go
+++ b/pkg/skaffold/docker/client.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"os/exec"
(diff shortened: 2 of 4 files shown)


renzodavid9 (approved):

ericzzzzzzz:
I was looking into the alternative way to simplify the whole `newEnvAPIClient()` method with `	"github.com/docker/cli/cli/command"` ,  https://pkg.go.dev/github.com/docker/cli/cli/command#DockerCli.CurrentContext seems to have everything we need, it also helps set up ssh connection helper implemented in #8349, but it seems that the tls config has different behavior from the current setup, maybe we can revisit this later. 
```golang
	var opts = []client.Opt{client.WithHTTPHeaders(getUserAgentHeader())}

	cli, err := command.NewDockerCli()
	if err != nil {
		return nil, nil, fmt.Errorf("fail to create dockerCli instance %w", err)
	}
	err = cli.Initialize(flags.NewClientOptions())
	if err != nil {
		return nil, nil, fmt.Errorf("fail to initialize dockerCli instance %w", err)
	}
	endpoint := cli.DockerEndpoint()
	clientOpts, err := endpoint.ClientOpts()
	if err != nil {
		return nil, nil, fmt.Errorf("fail to get dockerCli endpoint options %w", err)
	}
	opts = append(opts, clientOpts...)
	c, err := client.NewClientWithOpts(opts...)
```

PR #9109, opened 2023-10-03
docs: update changelog for v2.8.0

renzodavid9:
Changelog and schema changes for v2.8.0


renzodavid9:
Failing test is a known problem: when we do an schema upgrade the test fails due to it expects the examples to have the schema version from the latest version found in GCS, which doesn't correspond to the version we want to launch

plumpy (approved):

PR #9136, opened 2023-10-16
fix: custom crd not printing streams logs

beastpu:
**Description**
my company uses a custom CRD (openkruise https://openkruise.io/) to manage workloads. 
During local development, we discovered that aggregated logs were not being printed.  Aggregated logs are crucial for microservices architecture.

dce3f3fb-d041-4a1d-8235-35be90b628a6

**skaffold.yaml**
```
apiVersion: skaffold/v4beta7
kind: Config
build:
  artifacts:
  - image: skaffold-example
  local:
    push: false
manifests:
  rawYaml:
  - sts.yaml

resourceSelector:
  allow:
    - groupKind: CloneSet.apps.kruise.io
      image:
        - .*
      labels:
        - .*
    - groupKind: StatefulSet.apps.kruise.io
      image:
        - .*
      labels:
        - .*
```

**sts.yaml**
```
apiVersion: apps.kruise.io/v1beta1
kind: StatefulSet
metadata:
  name: getting-started
spec:
  replicas: 1
  serviceName: getting-started
  selector:
    matchLabels:
      app: getting-started
  template:
    metadata:
      labels:
        app: getting-started
    spec:
      containers:
      - name: getting-started
        image: skaffold-example
        imagePullPolicy: IfNotPresent
```

I found that the reason was the absence of corresponding GVK (GroupVersionKind) in `manifest.TransformAllowlist` and `manifest.TransformDenylist`. 

My solution is to pass ConsolidateTransformConfiguration into the NewResourceSelectorImages method because we can configure resourceSelector in the skaffold.yaml file.

### DIFF
--- a/pkg/skaffold/deploy/helm/helm.go
+++ b/pkg/skaffold/deploy/helm/helm.go
@@ -303,7 +303,7 @@ func (h *Deployer) Deploy(ctx context.Context, out io.Writer, builds []graph.Art
 	for ns := range nsMap {
 		namespaces = append(namespaces, ns)
 	}
-	deployedImages, _ := manifests.GetImages(manifest.NewResourceSelectorImages(manifest.TransformAllowlist, manifest.TransformDenylist))
+	deployedImages, _ := manifests.GetImages(manifest.NewResourceSelectorImages(h.transformableAllowlist, h.transformableDenylist))
 
 	h.TrackBuildArtifacts(builds, deployedImages)
 	h.trackNamespaces(namespaces)
--- a/pkg/skaffold/deploy/helm/helm.go
+++ b/pkg/skaffold/deploy/helm/helm.go
@@ -295,7 +295,7 @@ func (h *Deployer) Deploy(ctx context.Context, out io.Writer, builds []graph.Art
 	// Otherwise, templates have no way to use the images that were built.
 	// Skip warning for multi-config projects as there can be artifacts without any usage in the current deployer.
 	if !h.isMultiConfig {
-		warnAboutUnusedImages(builds, manifests)
+		h.warnAboutUnusedImages(builds, manifests)
 	}
 
 	// Collect namespaces in a string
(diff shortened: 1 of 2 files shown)


pkg/skaffold/deploy/helm/helm.go
```diff
@@ -303,7 +303,7 @@ func (h *Deployer) Deploy(ctx context.Context, out io.Writer, builds []graph.Art
 	for ns := range nsMap {
 		namespaces = append(namespaces, ns)
 	}
-	deployedImages, _ := manifests.GetImages(manifest.NewResourceSelectorImages(manifest.TransformAllowlist, manifest.TransformDenylist))
+	deployedImages, _ := manifests.GetImages(manifest.NewResourceSelectorImages(h.transformableAllowlist, h.transformableDenylist))
```
renzodavid9, line 306: Hey @beastpu. This part looks good. Could you please also update https://github.com/GoogleContainerTools/skaffold/blob/1f591bafe367868e61eedc38f427acaae61e0a31/pkg/skaffold/deploy/helm/helm.go#L627 to use `h.transformableAllowlist` and `h.transformableDenylist`? We'll need to send those to the `warnAboutUnusedImages` function; it will remove a warning when using the Helm deployer with a CRD. Thanks a lot 😄

renzodavid9 (commented):
Hey @beastpu! Thanks for opening this PR. I left a comment for one more place we need to update the resource lists.

Checking the change, I think I found a case where still we'll not be able to get the logs: when the CRD uses a field different than `image` to define the image to deploy. When we are going through the manifests to get the images, Skaffold is looking for the specific `image` field:
https://github.com/GoogleContainerTools/skaffold/blob/1f591bafe367868e61eedc38f427acaae61e0a31/pkg/skaffold/kubernetes/manifest/images.go#L114-L117

I think the proper way here would be to use the selectors defined in the `resourceSelector` from the `skaffold.yaml`.

With that said, I'll create another issue to track that case and tackle that in another PR; I think we can merge this PR once we have the last change from the comment I left. Thanks a lot for your help! 😄

renzodavid9 (approved):

Issue #9139, opened 2023-10-16
[Custom Actions] New config option to check images in local host before pull


renzodavid9:
### Actual behavior
Right now, when we run a [local custom action](https://skaffold.dev/docs/custom-actions/#local-docker), Skaffold will trigger a pull of the image every single time, e.g:

With this `skaffold.yaml`
```yaml
apiVersion: skaffold/v4beta7
kind: Config

customActions:
  - name: action
    containers:
      - name: alpine-task
        image: alpine
        command: ["/bin/sh"]
        args: ["-c", "echo From action"]
```

Executing `skaffold exec action` will produce the following logs:
```
Starting execution for action
latest: Pulling from library/alpine
Digest: sha256:eece025e432126ce23f223450a0326fbebde39cdf496a85d8c016293fc851978
Status: Image is up to date for alpine:latest
[alpine-task] From action
```
The previous log indicates that Skaffold is triggering a pull of the `alpine` image, even when the image is already in the local host.

### Expected behavior
We want to add a new config option to prevent Skaffold to trigger the pull of the images every single time for a given [local custom action](https://skaffold.dev/docs/custom-actions/#local-docker). Per default (no config set explicitly), Skaffold should keep its current behavior and trigger the pull every time an action is run. With the new config option set in the `skaffold.yaml` Skaffold should first check if the images of the custom action exist in the local host before triggering a pull.

PR #9147, opened 2023-10-23
chore: change custom actions pull logic, to check if an image exists locally first before triggering a pull

renzodavid9:
Fixes: #9139

**Description**
This PR adds a new config field for local custom actions to force the use of a local image if one exists, if it doesn't exists it will trigger a pull. 

**How to test the change**
With the following `skaffold.yaml` file:
```
apiVersion: skaffold/v4beta8
kind: Config

customActions:
  - name: my-action
    executionMode:
      local:
        useLocalImages: true
    containers:
      - name: gcloud-task
        image: gcr.io/cloud-builders/gcloud:latest
        command: ["/bin/sh"]
        args: ["-c", "gcloud version"]
```

Execute `skaffold exec my-action`. You'll notice that the image is pull from the registry before running the container. Now if you run `skaffold exec my-action` again, you'll see that Skaffold is not trying to pull the image again from the registry.

**Follow-up work**
https://github.com/GoogleContainerTools/skaffold/issues/9148: We'll be adding the same logic for `skaffold verify`, the schema change affects both, `customActions`, and `verify`.

### DIFF
--- a/pkg/skaffold/actions/docker/exec_env_test.go
+++ b/pkg/skaffold/actions/docker/exec_env_test.go
@@ -52,7 +52,7 @@ func (fd *fakeDockerDaemon) ImageID(ctx context.Context, ref string) (string, er
 	return img, nil
 }
 
-func getActionCfg(aName string, containers []latest.VerifyContainer) latest.Action {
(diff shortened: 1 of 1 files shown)


aaron-prindle (approved):
LGTM!

Issue #9177, opened 2023-11-15
When the image is remote and cached, skaffold tries to pull it locally and ignores the cache.


3droj7:
Hey,
We're using helm to deploy our images to the EKS and the images themselves are saved in the gitlab registry.
When we're trying to use skaffold, it will always try to download the image locally when it shouldn't because the deployment doesn't use the local images.

There's seems to be a bug in https://github.com/GoogleContainerTools/skaffold/blob/1f591bafe367868e61eedc38f427acaae61e0a31/pkg/skaffold/build/cache/lookup.go#L83
It tries to import the image when the image can be remote and doesn't need to be imported.
https://github.com/GoogleContainerTools/skaffold/blob/1f591bafe367868e61eedc38f427acaae61e0a31/pkg/skaffold/runner/new.go#L229

Also I would suggest to put some debug logs to understand if skaffold decided that the image is remote or local.

### Expected behavior
Checking cache...
 - manager_microservice: Found
### Actual behavior
Checking cache...
 - manager_microservice: Not found. Building
### Information
- Skaffold version: 2.9.0
- Operating system: ubuntu 22.04
- Installed via: skaffold.dev
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v4beta8
kind: Config
metadata:
  name: manager-microservice
build:
  cluster: {}
  artifacts:
    - image: manager_microservice
      context: ../../
      sync: {}
      custom:
        buildCommand: Build/buildx.sh "Manager/DtmMicroservice/Dockerfile" development
        dependencies:
          dockerfile:
            path: Manager/DtmMicroservice/Dockerfile
  tagPolicy:
    inputDigest: {}
```
skaffold build --default-repo registry.gitlab.com/.... --platform=linux/amd64 -v debug

ericzzzzzzz:
@3droj7 Hi, I'm trying to understand this issue. 

You provided a `build` command but you also mentioned about deployment with a remote image. This sounds you don't need `build` and just want to use the image on the remote registry?  

Or you want skaffold to compute the inputDigest then check cache, if not available, check local, check remote, tryImport happen in sequence, if any found then (store and) return result?

3droj7:
@ericzzzzzzz I actually meant "skafold run". What I expect to happen is the following - 
1. It will compute the inputDigest
2. Check remote cache  
3. If it found the image inside the cache then it moves on to the deployment stage
[...]

PR #9181, opened 2023-11-18
fix: puling images when working with a remote repository (#9177)

3droj7:
Fixes: #9177

**Description**
If the skaffold is configured to use the cluster and not the local images, it will not pull the images locally. 
If everything is cached it will move to deployment stage instead of unnecessarily downloading the images.

**Follow-up Work**
The code could be probably improved so there will be less duplicated code.

### DIFF
--- a/pkg/skaffold/build/cache/lookup.go
+++ b/pkg/skaffold/build/cache/lookup.go
@@ -88,7 +88,7 @@ func (c *cache) lookup(ctx context.Context, a *latest.Artifact, tag string, plat
 				log.Entry(ctx).Debugf("Could not import artifact from Docker, building instead (%s)", err)
 				return needsBuilding{hash: hash}
 			}
-	}
+		}
 		return c.lookupLocal(ctx, hash, tag, entry)
 	}
 	if !cacheHit {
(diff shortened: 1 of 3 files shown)


ericzzzzzzz:
Hi @3droj7 , Thanks for the contribution, could you please fix the failing tests,  you can run `make unit-tests` to verify your change.

3droj7:
@ericzzzzzzz It appears that there are some issues in the tests themselves- 
In the `TestCacheFindMissing` test, the function `isLocalImage` is mocked to return False. However, this seems contradictory. Given that `latest.LocalBuild{TryImportMissing: true}` implies the image should be local, not remote, this mock setup doesn't align with the expected behavior as per the definition. [retrieve_test.go#L342
](https://github.com/GoogleContainerTools/skaffold/blob/d49be7a831f893ec6d209f84f3ba3dbcf9afaeb6/pkg/skaffold/build/cache/retrieve_test.go#L342)

There's also an issue with `ImageID` returning an empty string which leads to a failure in the test, as the logic then defaults to building the image, as seen here [lookup.go#L98](https://github.com/GoogleContainerTools/skaffold/blob/d49be7a831f893ec6d209f84f3ba3dbcf9afaeb6/pkg/skaffold/build/cache/lookup.go#L98).

Do you agree that there's an issue or maybe I missed something? 
If there's a real problem I can fix the test but would like some help with mocking the `ImageID`. It would be really helpful if you can provide an example that already does that.

3droj7:
@ericzzzzzzz Hey, I fixed the tests. Can you please review it

PR #9218, opened 2023-12-12
chore: change bazel example

ericzzzzzzz:
- github actions CI is broken due to upstream docker rule changes, as it's also deprecated now https://github.com/bazelbuild/rules_docker, we should move our bazel example to use oci rule


ar3s3ru:
@ericzzzzzzz @renzodavid9 would it also make sense to remove this limitation on the target name? It was mostly due to `rules_docker` convention if I'm not mistaken: https://github.com/GoogleContainerTools/skaffold/blob/c4af00e97e7ffaf164655bf5a953f78fade99533/pkg/skaffold/build/bazel/build.go#L62-L64

ericzzzzzzz:
@ar3s3ru agree to remove the validation.

renzodavid9 (approved):

Issue #8936, opened 2023-07-06
Skaffold "runtimeType" not working


anthonyalayo:
There is something wrong with "runtimeType". Despite going through the Skaffold source code to verify the schema of the yaml file, my IDE setup with Cloud Code is still complaining that `runtimeType: jvm` is not allowed.

![image](https://github.com/GoogleContainerTools/skaffold/assets/17555572/f2b4b584-d5ef-4f2e-a1c0-4d938ee529fd)

I have seen this for months now, and I had thought that this was just a bug. However today I added `dumb-init` in front of my java process to fix other issues related to running as PID 1 in Linux. That's when I noticed that `runtimeType: jvm` is most definitely not working, because now `JAVA_TOOL_OPTIONS` is no longer set.

### Expected behavior

When specifying the following in a skaffold file:

```yaml
build:
  artifacts:
    - image: backend-java-app
      runtimeType: jvm
      docker:
        dockerfile: ci/Dockerfile
```

Regardless of what my entrypoint/command is for my Dockerfile, `JAVA_TOOL_OPTIONS` will be set.

### Actual behavior

`JAVA_TOOL_OPTIONS` is not set unless `java` is the command entry point.

### Information

- Skaffold version: v2.6.1
- Operating system: macOS Venture 13.2
- Installed via: Homebrew
- Contents of skaffold.yaml:

```yaml
#nonk8s
apiVersion: skaffold/v4beta5
kind: Config
build:
  artifacts:
    - image: backend-java-app
      runtimeType: jvm
      docker:
        dockerfile: ci/Dockerfile
  tagPolicy:
    inputDigest: { }
  local:
    push: false
    tryImportMissing: true
deploy:
  helm:
    releases:
      - name: backend-java-app
        chartPath: ./charts/backend-java-app
        valuesFiles:
          - ./charts/backend-java-app/values.yaml
          - ./charts/backend-java-app/values/dev/values.yaml
        setValueTemplates:
          fullnameOverride: backend-java-app
          image.repository: '{{.IMAGE_REPO}}'
          image.tag: '{{.IMAGE_TAG}}'
```

PR #9246, opened 2024-01-08
fix: debug-runtime-priority-issue

ericzzzzzzz:
Fixes: #8936

### DIFF
--- a/integration/testdata/debug/specified-runtime/Procfile
+++ b/integration/testdata/debug/specified-runtime/Procfile
@@ -1,0 +1,1 @@
+web: node src/index.js
--- a/pkg/skaffold/debug/transform_go_test.go
+++ b/pkg/skaffold/debug/transform_go_test.go
@@ -60,6 +60,33 @@ func TestExtractDlvArg(t *testing.T) {
 	}
 }
 
+func TestDlvTransformer_MatchRuntime(t *testing.T) {
+	tests := []struct {
+		description string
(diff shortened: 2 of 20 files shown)


integration/testdata/debug/specified-runtime/Procfile
```diff
@@ -0,0 +1 @@
+web: node src/index.js
```
renzodavid9, line 1: Not completely sure, but, does it look like we can remove this file? 🤔

pkg/skaffold/debug/transform_go_test.go
```diff
@@ -60,18 +60,40 @@ func TestExtractDlvArg(t *testing.T) {
 	}
 }
 
-func TestDlvTransformer_IsApplicable(t *testing.T) {
+func TestDlvTransformer_MatchRuntime(t *testing.T) {
 	tests := []struct {
 		description string
 		source      ImageConfiguration
 		launcher    string
```
renzodavid9, line 67: Looks like we are not setting any value for the `launcher` property in the `TestDlvTransformer_MatchRuntime` tests, is it safe to remove it? In that case, this will apply for all the `transform_<runtime>_test.go` files modified. Thanks! 😄

ericzzzzzzz: done

renzodavid9 (approved):

Issue #9248, opened 2024-01-10
lookupRemote function may always return found when the same tag is reused


yudppp:
Hi, I'm @yudppp.

This is my first time submitting an issue to this project.

If the format is incorrect, please let me know.

### Expected behavior

The lookupRemote function should return found if and only if the image with the given tag and digest exists in the cache. If the digests are different, even if the tags are the same, the function should return needsRemoteTagging.

This ensures that the user always builds the image they intend to build, even if the same tag is reused with a different digest.

### Actual behavior

The actual behavior of the lookupRemote function is that it will always return found if the same tag is reused, even if the digests are different.

### Information

- Skaffold version: v2.10.0
- Operating system: Ubuntu  22.0.3  LTS
- Installed via: `curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/latest/skaffold-linux-amd64`

I think this comment may be relevant to the issue.
https://github.com/GoogleContainerTools/skaffold/pull/9211/files#r1447231997

Thank you.

PR #9254, opened 2024-01-12
fix: digest different return needsRemote

ervitis:
Fixes: #9248 

**Description**

Check digests before and returns `found` type or `needsRemoteTagging`


ericzzzzzzz:
fixed in #9278

PR #9261, opened 2024-01-18
feat: added retry on files sync error

idsulik:
**Description**

Hi! I've been using skaffold for the last 2 years and from time to time I see error - "Skipping deploy due to sync error" and you can't do nothing with it in it was executed in auto sync mode.  
Within this PR I added retry mechanism if sync returns error

### DIFF
--- a/pkg/skaffold/runner/dev.go
+++ b/pkg/skaffold/runner/dev.go
@@ -23,6 +23,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
+
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/constants"
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/event"
 	eventV2 "github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/event/v2"
@@ -36,7 +38,6 @@ import (
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/util/term"
 	timeutil "github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/util/time"
 	"github.com/GoogleContainerTools/skaffold/v2/proto/v1"
-	"github.com/cenkalti/backoff/v4"
 )
 
 var (
--- a/pkg/skaffold/runner/dev.go
+++ b/pkg/skaffold/runner/dev.go
@@ -75,11 +75,9 @@ func (r *SkaffoldRunner) doDev(ctx context.Context, out io.Writer) error {
 	eventV2.TaskInProgress(constants.DevLoop, "")
 	defer func() { r.devIteration++ }()
 	eventV2.LogMetaEvent()
-	ctx, endTrace := instrumentation.StartTrace(
-		ctx, "doDev_DevLoopInProgress", map[string]string{
-			"devIteration": strconv.Itoa(r.devIteration),
-		},
-	)
+	ctx, endTrace := instrumentation.StartTrace(ctx, "doDev_DevLoopInProgress", map[string]string{
+		"devIteration": strconv.Itoa(r.devIteration),
+	})
 
 	meterUpdated := false
 	if needsSync {
@@ -241,11 +239,9 @@ func (r *SkaffoldRunner) Dev(ctx context.Context, out io.Writer, artifacts []*la
 	eventV2.TaskInProgress(constants.DevLoop, "")
 	defer func() { r.devIteration++ }()
 	eventV2.LogMetaEvent()
-	ctx, endTrace := instrumentation.StartTrace(
-		ctx, "Dev", map[string]string{
-			"devIteration": strconv.Itoa(r.devIteration),
-		},
-	)
+	ctx, endTrace := instrumentation.StartTrace(ctx, "Dev", map[string]string{
+		"devIteration": strconv.Itoa(r.devIteration),
+	})
 
 	// First build
 	var err error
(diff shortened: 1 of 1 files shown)


idsulik:
Hi!

idsulik:
@ericzzzzzzz Hi!

ericzzzzzzz:
Hi @idsulik  I think this needs rebase, also do we know what errors are retryable and what not?  I'm thinking we should distinguish those if we implement a retry mechanism for sync error

idsulik:
@ericzzzzzzz I think we should retry sync regardless of the error because in the case of auto-sync mode, you can't force it to sync again. there are several cases:  

- pre/post hooks error. 
- copying/deleting docker files. 


So if something goes wrong it can retry it because otherwise I have to use a workaround - checkout a different branch, checkout back to force it to sync changes

idsulik:
do you know any error related to the sync when we shouldn't retry ?

ericzzzzzzz:
@idsulik  You can find a few errors which are not recoverable without manual work using the query in `"sync error" in:title`.

idsulik:
@ericzzzzzzz what do you mean?
Search for "sync error" in the project gives me only 3 lines:  
image

I've already specified these errors above.  Could you please specify 1 sync error that is not recoverable without manual action?

pkg/skaffold/runner/dev.go
```diff
@@ -99,10 +104,23 @@ func (r *SkaffoldRunner) doDev(ctx context.Context, out io.Writer) error {
...
+			return nil
+		}
+		for _, s := range r.changeSet.NeedsResync() {
+			err := backoff.Retry(
+				func() error {
+					return syncHandler(s)
+				}, backoff.WithContext(opts, childCtx),
+			)
+
+			if err != nil {
```
ericzzzzzzz, line 121: why returning nil here?

pkg/skaffold/runner/dev.go
```diff
@@ -85,9 +86,13 @@ func (r *SkaffoldRunner) doDev(ctx context.Context, out io.Writer) error {
 			r.changeSet.ResetSync()
 			r.intents.ResetSync()
 		}()
+
+		// todo: make this configurable
```
ericzzzzzzz, line 90: Is this something you left here while you're coding and wanted to add before committing your changes ?

pkg/skaffold/runner/dev.go
idsulik: it's old logic, I left it as is. 
image

pkg/skaffold/runner/dev.go
idsulik: No, not in this PR, I left it in case if anyone wants to do it, maybe I'll do it by myself, but not now

ericzzzzzzz:
> @ericzzzzzzz I think we should retry sync regardless of the error because in the case of auto-sync mode, you can't force it to sync again. there are several cases:
> 
> * pre/post hooks error.
> * copying/deleting docker files.
> 
> So if something goes wrong it can retry it because otherwise I have to use a workaround - checkout a different branch, checkout back to force it to sync changes

> * pre/post hooks error.
 Do you mean hooks error will cause sync fail and we should re-try sync in this case? 
 Cloud you give me a specific example why re-try sync would help ?
 
 > * copying/deleting docker files.
  Wouldn't this trigger a rebuild?

pkg/skaffold/runner/dev.go
ericzzzzzzz: I'd suggest to implement configurable re-try for sync and default it to `0` -- not re-try to keep the user experience the same as before, if we want to have a sync retry, but I am still not convinced that we need this. 

While `// todo` comments might seem like helpful reminders, they often create more problems than they solve:

`Limited Visibility`: Only those who stumble upon the specific code section see the comment, potentially leaving others unaware of the task.
`Unclear Ownership and Status`: There's no indication of who's responsible for the task, whether it's assigned, in progress, or forgotten.
`Lack of Discussion`: Meaningful conversations are difficult without a dedicated issue or pull request, hindering collaboration.
`Prioritization Challenges`: Maintainers struggle to assess the urgency and importance of tasks scattered throughout the codebase.
`Ambiguity in Pull Requests`: It's unclear if // todo marks unfinished work or instructions for future changes, causing confusion during reviews.
These issues often lead to // todo items being overlooked or forgotten.

A more effective approach is to create an issue for each task:

- Issues provide a centralized location for tracking, assigning, and discussing tasks.
- They allow for clear prioritization and status updates.
- Discussions within issues create better communication and collaboration.

idsulik:
> * Cloud you give me a specific example why re-try sync would help ?

while skaffold is trying to sync files something happened with the connection, skaffold will fail and when the connection become stable, you have to somehow trigger skaffold to start syncing again. if we have retry, on second/third retry it can sync the files successfully .  

I have such cases a lot(VPN issue, quick checkout between branches lead to sync error etc.)

ericzzzzzzz:
> > * Cloud you give me a specific example why re-try sync would help ?
> 
> while skaffold is trying to sync files something happened with the connection, skaffold will fail and when the connection become stable, you have to somehow trigger skaffold to start syncing again. if we have retry, on second/third retry it can sync the files successfully .
> 
> I have such cases a lot(VPN issue, quick checkout between branches lead to sync error etc.)

@idsulik thank you for the explanation! Could you please fix the ci issue. I'll approve the change.

idsulik:
@ericzzzzzzz pushed the fix. thank you!

ericzzzzzzz (approved):

PR #9267, opened 2024-01-19
feat: extracted kaniko copyTimeout and copyMaxRetries into config

idsulik:
**Description**
I use skaffold for big project and sometimes it takes more than 5 minutes to create tar(it tries to add more than 80k files) and fails because of timeout, but 5m was hardcoded so I decided to extract it to config.

**Changes previews**
image

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -42,7 +42,7 @@ jobs:
 
       # Upload the results as artifacts (optional).
       - name: "Upload artifact"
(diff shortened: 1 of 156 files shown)


idsulik:
Hi!

idsulik:
@MarlonGamez , hi! Could you please review the PR? The size is big because of generated files(2 new config keys for kaniko), the changes themselves aren't that big.
I have to build skaffold with increased kaniko timeout for my colleagues because the default timeout of 5m is too low for us, it'll be convenient to use specify timeout in the config file

ericzzzzzzz (approved):

PR #9269, opened 2024-01-20
fix(kaniko): resolve blocking on Ctrl+C during Docker context creation

idsulik:
**Description**

This commit resolves a deadlock issue in Skaffold where sending a `Ctrl+C` interrupt during the Docker context creation process, specifically while executing `tw.WriteHeader(header)` in `pkg/skaffold/util/tar.go`, would cause an indefinite block. The solution implemented adds a `select` statement to handle the context cancellation, allowing the process to exit gracefully.

---

**User facing changes**

**Before**: Previously, when a user pressed `Ctrl+C` during the Docker context creation with Kaniko, Skaffold would block indefinitely, requiring manual termination of the process.

**After**: With the fix, Skaffold now responds to `Ctrl+C` interrupts promptly during the Docker context creation, terminating the process gracefully and avoiding a deadlock.


idsulik:
the bug also happens when you try to call Ctrl + C on kaniko build retry:
image

As you can see on the screen, I tried to close the process, but it didn't respond me, I had to kill it

ericzzzzzzz (approved):

Issue #8800, opened 2023-05-18
skaffold build: skaffold doesn't delete kaniko pod if build is interrupted with ctrl-c on Kaniko's filesystem snapshot stage


CarpathianUA:
For in-cluster builds with `kaniko`, `skaffold build` doesn't delete `kaniko-<>` pod if `build` is interrupted with `ctrl-c` on filesystem snapshot stage:

`skaffold build -vdebug`

```
...
INFO[2023-05-18T16:35:55Z] Taking snapshot of full filesystem...        
^CERRO[0032] deleting pod: client rate limiter Wait returned an error: context canceled  subtask=<REDACTED>.dkr.ecr.us-east-1.amazonaws.com/<REDACTED>task=Build
Build [<REDACTED>.dkr.ecr.us-east-1.amazonaws.com/<REDACTED>] was canceled
```

```
kubectl get po 

kaniko-d7bxc                                1/1     Running   0          17s
```

### Expected behavior
skaffold trigger a deletion of kaniko pod if build is interrupted with ctrl-c on Kaniko's filesystem snapshot stage

### Actual behavior
skaffold doesn't delete kaniko pod if build is interrupted with ctrl-c on Kaniko's filesystem snapshot stage

### Information

- Skaffold version: v2.4.1
- Operating system: MacOS Ventura 13.3.1 (a) , 22.4.0 Darwin Kernel arm64
- Installed via: Homebrew
- k8s: v1.25.6-eks-48e63af
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v4beta5
kind: Config
build:
  tagPolicy:
    envTemplate:
      template: "{{.CUSTOM_IMAGE_TAG}}"
  artifacts:
    # build.artifacts.*.image is not templateable.
    # see https://github.com/GoogleContainerTools/skaffold/issues/7455
    # it will be overriden by profile name, see profiles section below. 
    # {{.SKAFFOLD_PROFILE}} is just a placeholder here, at least for now, untill Skaffold allows to template image field.
    - image: "<REDACTED>.dkr.ecr.us-east-1.amazonaws.com/{{.SKAFFOLD_PROFILE}}"
      kaniko:
        dockerfile: Dockerfile
        snapshotMode: redo
        pushRetry: "2"
        cleanup: true
        logTimestamp: true
        useNewRun: true
        whitelistVarRun: true
        # see skaffold.env
        label:
          builder: kaniko
[...]

PR #9270, opened 2024-01-21
fix(kaniko): delete kaniko pod on graceful shutdown

idsulik:
Fixes: #8800 

**Description**
Fixed kaniko pod delete in case of graceful shutdown

**User facing changes**
**Before**: skaffold doesn't delete kaniko pod if build is interrupted with ctrl-c after creating the pod

**After**: skaffold triggers deletion of kaniko pod if build is interrupted with ctrl-c after creating the pod

### DIFF
--- a/pkg/skaffold/build/cluster/kaniko.go
+++ b/pkg/skaffold/build/cluster/kaniko.go
@@ -83,6 +83,8 @@ func (b *Builder) buildWithKaniko(ctx context.Context, out io.Writer, workspace 
 	}
 
 	defer func() {
+		// if build interrupted the original context is cancelled
+		// and pod deletion will not be called, so we need a new ctx
 		ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
 		defer cancel()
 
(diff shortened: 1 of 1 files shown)


idsulik:
@ericzzzzzzz Hi!

idsulik:
@ericzzzzzzz hi! Check please the PR, it's very important because all my colleagues have this issue which leads to 100500 undelete kaniko pods. 

image

ericzzzzzzz (approved):

ericzzzzzzz:
Thanks for the fix! There's something wrong with integration tests,  but it doesn't seems related to the change, I'm still investigating, after the issue resolved, I'll merge this change. Sorry for the delay.

pkg/skaffold/build/cluster/kaniko.go
```diff
@@ -77,10 +77,15 @@ func (b *Builder) buildWithKaniko(ctx context.Context, out io.Writer, workspace
 	}
 
 	pod, err := pods.Create(ctx, podSpec, metav1.CreateOptions{})
+
 	if err != nil {
 		return "", fmt.Errorf("creating kaniko pod: %w", err)
 	}
+
 	defer func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
```
ericzzzzzzz, line 86: Hi, can you add a comment to explains that if build interrupted the original context is cancelled and pod deletion will not be called, so we need a new ctx?

idsulik: Hi, yea, good idea.  Pushed changes

PR #9274, opened 2024-01-24
Support `FORCE_COLOR` env variable

omus:
**Related**: #2033

**Description**
Adds support for the common [`FORCE_COLOR`](https://force-color.org/) environmental variable to force the use of color output. I also decided to rename `--force-colors` to `--force-color` for consistency with the environmental variable and have made the option no longer hidden as it makes it hard to discover.

**User facing changes**

Makes discovery of the global option and the associated environmental variable easier to discover.

### DIFF
--- a/cmd/skaffold/app/cmd/cmd.go
+++ b/cmd/skaffold/app/cmd/cmd.go
@@ -48,7 +48,7 @@ var (
 	opts              config.SkaffoldOptions
 	v                 string
 	defaultColor      int
-	forceColors       bool
+	forceColor        bool
 	overwrite         bool
 	interactive       bool
 	timestamps        bool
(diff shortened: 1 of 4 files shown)


cmd/skaffold/app/cmd/cmd.go
```diff
@@ -205,7 +205,8 @@ func NewSkaffoldCommand(out, errOut io.Writer) *cobra.Command {
 	templates.ActsAsRootCommand(rootCmd, nil, groups...)
 	rootCmd.PersistentFlags().StringVarP(&v, "verbosity", "v", log.DefaultLogLevel.String(), fmt.Sprintf("Log level: one of %v", log.AllLevels))
 	rootCmd.PersistentFlags().IntVar(&defaultColor, "color", int(output.DefaultColorCode), "Specify the default output color in ANSI escape codes")
-	rootCmd.PersistentFlags().BoolVar(&forceColors, "force-colors", false, "Always print color codes (hidden)")
+	rootCmd.PersistentFlags().BoolVar(&forceColor, "force-colors", false, "Always print color codes (hidden)")
```
omus, line 208: Is there a way I should deprecate this option?

alphanota: Changing the name of the flag is a breaking change, we shouldn't do this.

alphanota, line 208: The flag is being initialized twice, in line 208 and 209.

alphanota (changes requested):

cmd/skaffold/app/cmd/cmd.go
omus: We can keep both `--force-colors` and `--force-color` as flags to avoid any breakage. How does Skaffold deal with informing users about upcoming breaking changes?

Issue #9236, opened 2024-01-02
[Feature Request] Support remote config dependencies with v2 Cloud Build repositories


mattsanta:
Currently [remote config dependencies](https://skaffold.dev/docs/design/config/#remote-config-dependency) supports Skaffold configs stored in a git repository and Google Cloud Storage. Unfortunately, the git repository option requires the credentials to already be set up in the local workspace in order to access private git repositories. 

As an alternative it would be nice if a user could configure a [Cloud Build v2 repository](https://cloud.google.com/build/docs/repositories#repositories_2nd_gen) where the Skaffold config is stored and Skaffold handles setting up the credentials using CB repositories [`accessReadToken`](https://cloud.google.com/build/docs/api/reference/rest/v2/projects.locations.connections.repositories/accessReadToken) and then pulls the repository to resolve the config dependencies.

PR #9283, opened 2024-01-29
feat: new repo resolver logic to fetch info from a gcbrepov2

renzodavid9:
**Related**: #9236

**Description**

- New logic to fetch the information related with a given Cloud Build repository v2
- Includes new `cloud.google.com/go/cloudbuild` library to manage connection with GCB repos v2

- Update the Skaffold schema to add the new fields to map a GCB Repository V2 dependency
- Update the [config parser](https://github.com/GoogleContainerTools/skaffold/blob/e4f19756b95020bff442938044a4a03adc01a702/pkg/skaffold/parser/config.go#L278) to retrieve the repo + clone URI from `gcbrepov2.GetRepoInfo` when a GCB Repository V2 is configured in the `skaffold.yaml` file
- We'll be using the logic in https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/git/gitutil.go#L92 to clone the repository once we have the proper oauth URI from the `gcbrepov2.GetRepoInfo` function; we'll need to update the parameters to receive a new type of object that supports both, the repo + clone repo URI


pkg/skaffold/gcbreposv2/repo_resolver.go
```diff
@@ -0,0 +1,81 @@
...
+
+type cloudBuildRepoClient interface {
+	GetRepository(ctx context.Context, req *cloudbuildpb.GetRepositoryRequest, opts ...gax.CallOption) (*cloudbuildpb.Repository, error)
+	FetchReadToken(ctx context.Context, req *cloudbuildpb.FetchReadTokenRequest, opts ...gax.CallOption) (*cloudbuildpb.FetchReadTokenResponse, error)
+	Close() error
+}
+
+var RepositoryManagerClient = repositoryManagerClient
+
+func GetRepoInfo(ctx context.Context, gcpProject, gcpRegion, gcpConnectionName, gcpRepoName string) (repoURI string, readAccessToken string, err error) {
```
ericzzzzzzz, line 36: I'm finding the current design of this function a bit difficult to follow. For a feature requiring multiple PRs, a top-down approach might make it easier to visualize how everything fits together.  Though bottom-up approaches ensure solid unit tests, top-down can be more conducive to getting everyone on the same page – but I understand this preference can be subjective.

pkg/skaffold/gcbreposv2/repo_resolver_test.go
```diff
@@ -0,0 +1,135 @@
...
+		return &cloudbuildpb.FetchReadTokenResponse{
+			Token: c.ReadToken,
+		}, nil
+	}
+	return nil, fmt.Errorf("invalid request")
+}
+
+func (c GCBReposClientMock) Close() error { return nil }
+
+func TestGetRepoURI(t *testing.T) {
```
ericzzzzzzz, line 71: The test is named as TestGetRepoURI, but we have an assertion on RepoToken

pkg/skaffold/gcbreposv2/repo_resolver_test.go
```diff
@@ -0,0 +1,135 @@
...
+	tests := []struct {
+		description         string
+		gcpRegion           string
+		gcpProject          string
+		gcpConnection       string
+		gcpRepo             string
+		expectedGitRepo     string
+		expectedToken       string
```
ericzzzzzzz, line 79: no non-zero values for test

pkg/skaffold/gcbreposv2/repo_resolver.go
ericzzzzzzz, line 36: can we use a struct for (repoURI string, readAccessToken string) , or we can also construct the uri with token here 

https://oauth2:<token>@<repo>

pkg/skaffold/gcbreposv2/repo_resolver.go
```diff
@@ -0,0 +1,81 @@
...
+	cbRepoRef := fmt.Sprintf("projects/%v/locations/%v/connections/%v/repositories/%v", gcpProject, gcpRegion, gcpConnectionName, gcpRepoName)
+	cbClient, err := RepositoryManagerClient(ctx)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to create repository manager client: %w", err)
+	}
+	defer cbClient.Close()
+
+	repoURI, err = getrepoURI(ctx, cbClient, cbRepoRef)
```
ericzzzzzzz, line 44: just want to call out, the returned uri has protocol, when we construct the url for `git` we need to remove `https://`

pkg/skaffold/gcbreposv2/repo_resolver.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	return
+}
+
+func repositoryManagerClient(ctx context.Context) (cloudBuildRepoClient, error) {
+	return cloudbuild.NewRepositoryManagerClient(ctx)
+}
+
+func getrepoURI(ctx context.Context, cbClient cloudBuildRepoClient, cbRepoRef string) (string, error) {
```
ericzzzzzzz, line 61: getrepoURI -> getRepoURI

pkg/skaffold/gcbreposv2/repo_resolver_test.go
renzodavid9: You are right, I changed the test and its name. Thanks!

pkg/skaffold/gcbreposv2/repo_resolver_test.go
renzodavid9: I changed the structure of the unit tests, also taking into account that now this package has the logic to build the oauth2 format. Let me know how it looks. Thanks!

pkg/skaffold/gcbreposv2/repo_resolver.go
renzodavid9: Got it! I changed it to return a struct as you suggested. The struct contains two fields, the `CloneRepoURI` with the oauth2 format, and the `URI` without the token. The reason to return both is to use the CloneRepoURI to execute all the needed git commands (clone, fetch, etc), and the URI for the errors we print so we don't show the token in the execution logs. The idea is to re use the logic that we already have in [pkg/skaffold/git/gitutil.go](https://github.com/GoogleContainerTools/skaffold/blob/e4f19756b95020bff442938044a4a03adc01a702/pkg/skaffold/git/gitutil.go#L92) to trigger the repo download/sync. Let me know what do you think.

pkg/skaffold/gcbreposv2/repo_resolver.go
renzodavid9: Yep! Thanks for head-up. I moved the logic that builds the git clone URI to this file so we have all in the same place.

pkg/skaffold/gcbreposv2/repo_resolver.go
renzodavid9: Done!

ericzzzzzzz (approved):

Issue #9284, opened 2024-01-29
skaffold build with docker fails to set BUILDPLATFORM variable 


aizam6066:
We have a step in our Github workflow, where we use skaffold to build a Docker image to run some unit-tests. Recently, the following command:

`skaffold build --push=false --platform=linux/amd64 -b $APP`

fails with this error:

```docker build failure: failed to parse platform : "" is an invalid component of "": platform specifier component must match "^[A-Za-z0-9_-]+$": invalid argument. Please fix the Dockerfile and try again..```

In our Dockerfile, we have:
`FROM --platform=$BUILDPLATFORM golang:1.19 AS builder`

and it seems that $BUILDPLATFORM is empty.

### Expected behavior
$BUILDPLATFORM should be set with flag `--platform=linux/amd64`

### Actual behavior
$BUILDPLATFORM is ""

### Information

- Skaffold version: 2.10.0
- Operating system: 
```
Current runner version: '2.312.0'
Operating System
  Ubuntu
  LTS
Runner Image
  Image: ubuntu-22.0
  Version: 20240122.3.0
  Included Software: https://github.com/actions/runner-images/blob/ubuntu22/20240122.3/images/ubuntu/Ubuntu2204-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20240122.3
```
- Contents of skaffold.yaml:
```yaml

apiVersion: skaffold/v3
kind: Config
build:
  artifacts:
    - image: app1
      sync:
        infer:
          - internal/**
          - pkg/**
          - cmd/kafka/**
          - main.go
          - go.mod
      platforms: ["linux/amd64", "linux/arm64"]
      docker:
        buildArgs:
          app_name: kafka
    - image: app2
      sync:
        infer:
          - internal/**
          - pkg/**
          - cmd/grpc/**
          - main.go
          - go.mod
      platforms: ["linux/amd64", "linux/arm64"]
      docker:
        buildArgs:
          app_name: grpc

```
Any help with this would be greatly appreciated! Thanks!

aizam6066:
Thank you very much for the quick response there, @ericzzzzzzz! Adding `useDockerCLI` did fix the issue! I just have couple follow-up questions. 
1) Why do you think it started breaking now after a year or so of us not having this key in our skaffold.yaml. The only thing I can think of is GH runners install the latest skaffold version and somewhere in the later version updates, this started to break?
[...]

PR #9291, opened 2024-02-01
fix(docker): pass BUILDPLATFORM env when --platform argument is used

idsulik:
Fixes: #9284 

**Description**
I'm not sure if I fixed the bug correctly, as I couldn't reproduce the bug for myself

**before**: if you use `--platform={platform}` the BUILDPLATFORM argument was not passed to the docker
**now**: it'll pass `BUILDPLATFORM`


idsulik:
@ericzzzzzzz Hi!

ericzzzzzzz:
The original has already been fixed. see https://github.com/GoogleContainerTools/skaffold/issues/9284#issuecomment-1927689782 . So I'm closing this one.

PR #9294, opened 2024-02-03
feat(profile): Added profile list command to output available profile names

idsulik:
**Description**
Hi! What do you think about the new command `skaffold profile list` which will output a list of profile names? In big projects it's quite hard to see all the profiles you can use

**User facing changes**
*Before*: You should open a config file to see available profiles
*After*: You can execute `skaffold profile list` to see available profiles


idsulik:
@ericzzzzzzz Hi!

Issue #7651, opened 2022-07-15
useHelmSecrets not decrypting on render


panbanda:
### Expected behavior

`skaffold render` should produce the same output as `skaffold up` for secrets that are used with `useHelmSecrets: true`

### Actual behavior

`skaffold render` does not decrypt the secret, `skaffold up` does decrypt the secret

### Information

Demo repo here:

https://github.com/panbanda/skaffold-helm-secrets-issue

### Steps to reproduce the behavior

```
# Install SOPS sample gpg key
gpg --import pgp/sops_test_key.asc

# Encrypted with 
sops -e -i secrets.yaml

# Decrypted with 
sops -d -i secrets.yaml

# Skaffold dev puts the secret in container decrypted
# ie: test
skaffold dev

# Skaffold render creates a kube secret manifest with encrypted string
# ie: ENC[AES256_GCM,data:SyyH6w==,iv:CbNC...
skaffold render
```

tejal29:
Thanks @panbanda. I am not aware or `skaffold up` do yo mean `skaffold dev`?

anthonyalayo:
I have the same issue. Skaffold render should all the secrets as encrypted. Can this be fixed? Ideally `useHelmSecrets` would have had them decrypted.

anthonyalayo:
Can anyone from the skaffold team chime in?

anthonyalayo:
@aaron-prindle @tejal29 If someone could point me to where the change needs to happen, I could also give it a shot. Is there anyone around that would know?

anthonyalayo:
Thanks @ericzzzzzzz !

Issue #9119, opened 2023-10-08
useHelmSecrets does not append `secrets` to `helm template` command


aimichelle:
### Expected behavior
The `helm-secrets` plugin allows users to pass encrypted secrets/values files to Helm. The plugin decrypts the secrets and passes the decrypted files for use in the Helm command, and is initiated by running `helm secrets <usual command>`. The `useHelmSecrets` config in Skaffold instructs skaffold to run helm with the `secrets` plugin. 
When running `skaffold run` with a Helm release, I would expect skaffold to run the `helm secrets template` command.

### Actual behavior
When running `skaffold run -v DEBUG` to deploy a Helm release with `useHelmSecrets: true`, the Helm command clearly does not contain the `secrets` call.

### Information

- Skaffold version: v.2.8.0
- Operating system: Ubuntu 22.04.3 LTS
- Installed via: Github releases binary

### Steps to reproduce the behavior

1. Add `useHelmSecrets: true` to a Helm release in the skaffold.yaml
2. `skaffold run -f skaffold.yaml -v DEBUG`
3. Inspect the `helm template` command, notice it does not begin with `helm secrets`.

Possibly relevant/related issues: https://github.com/GoogleContainerTools/skaffold/issues/7651

PR #9295, opened 2024-02-04
fix(helm): use secrets helm plugin to render when useHelmSecrets is true

idsulik:
Fixes: #7651 and possibly #9119

**Description**
`skaffold render` used to ignore `useHelmSecrets` flag

**User facing changes**
*Before*: `skaffold render` didn't use helm secrets plugin
*After*: `skaffold render` uses helm secrets plugin


ericzzzzzzz (approved):

ericzzzzzzz:
thank you for the fix!

PR #9278, opened 2024-01-26
fix(lookupRemote): fixed lookup.go lookupRemote to compare remote and cached digests

idsulik:
Fixes: #9248 

**Description**
It is used to lookup remote images by tag and overwrite a cached item even if the digests are different, now it compares the digests

PR #9211, opened 2023-12-06
fix: Scope Issue with the 'entry' variable when looking up remote images and tests additions

3droj7:
**Related**: #9177

**Description**
The variable "entry" was not overridden before triggering the lookupRemote method.
The reason is because the entry variable was redeclared within an if block, resulting in a new, locally scoped variable. This redeclaration prevented the updates made to entry within this block from being reflected outside of it.

The bug was introduced in my last PR (9181)... I had no idea that in go, the "if" statement can define a whole new scope

I've also added some tests, so it would test the scenario where we have tags in the remote which shouldn't be built and also fixed another test where it uses missingPull functionality on the remote

PR #9306, opened 2024-02-09
fix(lookupRemote): fixed lookup.go lookupRemote caching mechanism

idsulik:
Fixes: #9248  

**Description**
Hi! Within this PR https://github.com/GoogleContainerTools/skaffold/pull/9278 I've fixed lookupRemote and it turned out that I've missed a case when the cache was missed but the image was found remotely (it'll be better to check the test case to understand it)

**Before**: when a tag wasn't found in the cache, but the image was found remotely by the tag, then we use **needsRemoteTagging**
**After**: with the same conditions it now uses **needsBuilding** instead

### DIFF
--- a/pkg/skaffold/build/cache/lookup.go
+++ b/pkg/skaffold/build/cache/lookup.go
@@ -128,13 +128,6 @@ func (c *cache) lookupRemote(ctx context.Context, hash, tag string, platforms []
 			log.Entry(ctx).Debugf("Found %s remote with the same digest", tag)
 			return found{hash: hash}
 		}
-
-		if !cacheHit {
-			log.Entry(ctx).Debugf("Added digest for %s to cache entry", tag)
-			c.cacheMutex.Lock()
-			c.artifactCache[hash] = ImageDetails{Digest: remoteDigest}
-			c.cacheMutex.Unlock()
-		}
 	}
 
 	if cacheHit {
--- a/pkg/skaffold/build/cache/lookup.go
+++ b/pkg/skaffold/build/cache/lookup.go
@@ -131,28 +131,29 @@ func (c *cache) lookupRemote(ctx context.Context, hash, tag string, platforms []
 
 		if !cacheHit {
 			log.Entry(ctx).Debugf("Added digest for %s to cache entry", tag)
-			cachedEntry.Digest = remoteDigest
 			c.cacheMutex.Lock()
-			c.artifactCache[hash] = cachedEntry
+			c.artifactCache[hash] = ImageDetails{Digest: remoteDigest}
 			c.cacheMutex.Unlock()
 		}
 	}
 
-	if cachedEntry.HasDigest() {
-		// Image exists remotely with a different tag
-		fqn := tag + "@" + cachedEntry.Digest // Actual tag will be ignored but we need the registry and the digest part of it.
-		log.Entry(ctx).Debugf("Looking up %s tag with the full fqn %s", tag, cachedEntry.Digest)
-		if remoteDigest, err := docker.RemoteDigest(fqn, c.cfg, nil); err == nil {
-			log.Entry(ctx).Debugf("Found %s with the full fqn", tag)
-			if remoteDigest == cachedEntry.Digest {
-				return needsRemoteTagging{hash: hash, tag: tag, digest: cachedEntry.Digest, platforms: platforms}
+	if cacheHit {
+		if cachedEntry.HasDigest() {
+			// Image exists remotely with a different tag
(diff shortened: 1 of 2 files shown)


idsulik:
@ericzzzzzzz hi! Check it please, I think it'll be better to create patch release with these changes

ericzzzzzzz:
Hi, @idsulik  thank you for the fix, but I think our best option may be just roll back all changes related to cache lookup to the previous good version v2.9.0, and revisit https://github.com/GoogleContainerTools/skaffold/issues/9177 to see if the request is valid, we definitely need to document this about how skaffold cache works and what are the expected behaviors, I'll put a short summary later and create another issue to document skaffold caching

idsulik:
@ericzzzzzzz I've double-checked these changes and I use it daily, it works perfectly, so I think it's safe to merge and use it

idsulik:
I missed one case in the previous PR and fixed it here + added a test for it.

pkg/skaffold/build/cache/lookup.go
```diff
@@ -131,28 +131,29 @@ func (c *cache) lookupRemote(ctx context.Context, hash, tag string, platforms []
 
 		if !cacheHit {
 			log.Entry(ctx).Debugf("Added digest for %s to cache entry", tag)
-			cachedEntry.Digest = remoteDigest
 			c.cacheMutex.Lock()
-			c.artifactCache[hash] = cachedEntry
+			c.artifactCache[hash] = ImageDetails{Digest: remoteDigest}
```
ericzzzzzzz, line 135: I think skaffold saves digest to cache by checking the remote image after build, the code here might not be necessary and if this works as the code indicates it's confusing
 - we find a remoteDigest with the target tag 
 - in cache miss case, we save the digest, skaffold will rebuild and re-push,  then the digest changes on remote but the digest on the cache is old one??

ericzzzzzzz:
> @ericzzzzzzz I've double-checked these changes and I use it daily, it works perfectly, so I think it's safe to merge and use it

if you believe you've made some improvements on the good working version -- v2.9.0, please them on the description. Thanks 

I think there maybe some misunderstanding on the original issue, 
[When the image is remote and cached, skaffold tries to pull it locally and ignores the cache. #9177](https://github.com/GoogleContainerTools/skaffold/issues/9177) 

 - if an remote image is cached on previous build and digests match, skaffold should not build
 - The pr https://github.com/GoogleContainerTools/skaffold/pull/9181/files and https://github.com/GoogleContainerTools/skaffold/pull/9211/files#diff-c29561cd5f488e560760ff508c85b0a485100b79f6971be9a1229278fb76b482 look for remote image with the target tag, if found it skips the build, regardless cache miss or found,  it seems like @3droj7  just wants to use the image(tag) specified in the skaffold.yaml . It seems to me that we're trying to solve a different issue. 
 - So I'm still thinking going back to the old state.

pkg/skaffold/build/cache/lookup.go
idsulik: I've just checked it and found that we don't need to save the digest here because it'll be overwritten here `pkg/skaffold/build/cache/details.go:121`.  
removed this part and pushed.

idsulik:
@ericzzzzzzz 

> if an remote image is cached on previous build and digests match, skaffold should not build
in this PR works in this way

ericzzzzzzz:
> @ericzzzzzzz
> 
> > if an remote image is cached on previous build and digests match, skaffold should not build.
> 
> in this PR works in this way

thank you so much! I'll test it and give approve if everything works as expected.

ericzzzzzzz:
> I don't know about v2.9.0, if it works the same, here we have separated methods for/local remote at least, so we can look at it as refactoring

I admit this is all my fault for not carefully reviewing the #9181 and #9211  for the requirements in #9177,
 - the implementation in #9211 ignores the cache and just check if a image tag exists the remote, if exists use the remote image as the build result
 - that caused the problem reported in #9248 -- remote image lookup always returns found even cache misses or digest mismatch
 - #9278 made an attempt to fix #9248, but is misses the case where cache is not hit.
 - This pr is making another patch to fix #9248, but we have to stop here. Every pr after #9211 should have two goals 
  - fix the issue #9248 
  - persist the functionality introduced in #9211, if the ask in #9177 is valid,
 - we took #9177 as valid requirement but we missed the second goal in #9278 and end up here, lookup() method seems to behave the same as v2.9.0 after the pr merges, we should not merge a pr if we think #9177 is valid and the pr is not meant to solve it as we don't want to work on something then get a result of something else. If we believe #9177 is not valid, then all attempts for #9177, #9248 should be reverted.
 - As we're still having the remote cache lookup problem in #9248, going back to the original good state is the best option we have.
 
 Sorry, I have to close this and start to revert #9181, #9211, #9278

PR #9330, opened 2024-02-29
Update skaffold.yaml

panwarharsh15:
change apiVersion

**Description**
with previous apiversion, it gives the error of older API version " error parsing skaffold configuration file: unknown skaffold config API version "skaffold/v4beta9" "
I change its apiVersion from "v4beta9" to "v2beta15"
 Before : apVersion: skaffold/v4beta9
 After : apiVerion: skaffold/v2beta15


ericzzzzzzz:
Please upgrade your skaffold binary version, that should fix the issue

Issue #9326, opened 2024-02-26
Manually triggered sync with manual mapping not working when multiple files are changed of which the first one is not mapped


simonstratmann:
### Expected behavior
When two files are modified in one session and only the second is included in a manual sync mapping then a sync should be executed.

### Actual behavior
When two files are modified in one session and only the second is included in a manual sync mapping then no sync is being executed. If only the second file is modified then sync is executed. In other words, for any number of changed files if the first in the list does not match the pattern then no sync is being executed.

### Information

- Skaffold version: 2.10.1
- Operating system: Windows 10
- Installed via: Google Cloud code
- Contents of skaffold.yaml (excerpt):
```yaml
sync:
    manual:
        - src: "project-rest/target/classes/**/*.class"
        dest: "/app/classes/"
        strip: "project-rest/target/classes/"
```

When .java and .class files are modified (by changing the java file in IntelliJ and compiling):
```
files modified: [project-rest\src\main\java\package\WebController.java project-rest\target\classes\package\WebController.class]
files modified: [project-rest\src\main\java\package\WebController.java project-rest\target\classes\package\WebController.class]
Changed file project-rest\src\main\java\package\WebController.java does not match any sync pattern. Skipping sync
dev intents: build false, sync true, deploy false
```

When only the .class file is modified ((by changing the java file in IntelliJ, changing the window for IntelliJ to write the java file and then compiling):
```
files modified: [project-rest\target\classes\package\WebController.class]
files modified: [.idea\workspace.xml project-rest\target\classes\package\WebController.class]
dev intents: build false, sync true, deploy false

 devloop: build false, sync true, deploy false

Syncing 1 files for localhost:5010/project-application:jbo_temp_20240220-9-gff455bb-dirty@sha256:0a3adc1d601c8097a7a4ea6665de500422afa7bb1807776a3e651a51fd8ee5c5
```

I'm working on a sample project to reproduce but I'm having issues with docker registry authentication.

This problem occurs with any number of modified files:
```
files modified: [project-application\target\classes\banner.txt project-application\target\classes\bootstrap.yml project-rest\target\classes\de\ppi\pcy\gen\rest\WebController.class]
[...]

PR #9344, opened 2024-03-14
fix: Allow partial match for manual sync

simonstratmann:
Fixes: #9326 

**Description**
Current behavior: If files `someFile.txt` and `someFile.html` are changed and a manual sync pattern ` *.html` exists then no sync is made because the first file does not match. 

With this MR partial matches should be correctly recognized.

Not sure if this approach is correct or if the returned values should be interpreted differently instead of saving the mismatch inside the function.

### DIFF
--- a/pkg/skaffold/sync/sync.go
+++ b/pkg/skaffold/sync/sync.go
@@ -217,9 +217,9 @@ func autoSyncItem(ctx context.Context, a *latest.Artifact, tag string, e filemon
 }
 
 func latestTag(image string, builds []graph.Artifact) string {
-	for _, build := range builds {
-		if build.ImageName == image {
-			return build.Tag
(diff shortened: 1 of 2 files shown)


simonstratmann:
I've updated the code and the expectations in the test. To me it looks like the behavior was wrong before but as some of the test code has existed since 2019 I may be misunderstanding the use case.

Please see the first commit that keeps the old expectations mostly intact and the newer commit that updates the tests according to what I would expect to be correct.

theotherp:
I would love for somebody to review this. The bug is preventing me from using hot deploy in a meaningful manner.

simonstratmann:
I would love for somebody to review this. The bug is preventing me from using hot deploy in a meaningful manner.

PR #9376, opened 2024-04-02
release: v2.11.0

ericzzzzzzz:
update release schema and changelog


ericzzzzzzz:
Linter failing is caused by sample check against the latest release version on gcs but the version on gcs got updated after the release, this something we should fix in the future but this should not block the release

docs-v2/content/en/schemas/version-mappings/v4beta10-version.json
```diff
@@ -0,0 +1 @@
+{"binVersion":"2.11.0","releaseNoteLink":"https://github.com/GoogleContainerTools/skaffold/releases/tag/2.11.0"}
```
renzodavid9, line 1: The release script is not generating the `releaseNoteLink` properly under some parameters, for this case the last part of the link should be `/v2.11.0` instead of `/2.11.0`, could you change it manually? We'll have to fix the release script to not have this problem in future releases. Thanks!

renzodavid9 (approved):

docs-v2/content/en/schemas/version-mappings/v4beta10-version.json
ericzzzzzzz: ahhh, thanks

PR #9387, opened 2024-04-13
fix: add --load flag for local buildkit

Crystalix007:
**Description**

Adds the --load flag when building images using buildkit, and not pushing to a remote registry.

This is required, to get them loaded into the local docker context.

Before:

Running `dev` complained with:

```
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
...
Build [docker.io/...] failed: could not push image "docker.io/...:4002597-dirty": An image does not exist locally with the tag: ...
```

After:

```
Tags used in deployment:
 - docker.io/... -> docker.io/crystalix007/...:e1958765598cf4713ec0f5c9d9afa702fccc6098d509d84d272a14b373f422dd
```


ericzzzzzzz:
Hi @Crystalix007 Thank you for the contribution! The change looks good, but skaffold integration test infra seems broken at the moment, we'll merge the change after the issue is fixed.

ericzzzzzzz (approved):

Issue #9389, opened 2024-04-16
Allow any network to be provided to docker builder


edwardthiele:
Under the Skaffold build configuration for docker, you can provide a network to connect the docker container (https://skaffold.dev/docs/references/yaml/#build-artifacts-docker-network), however you are restricted in terms of what network options can be set.

In particular, you cannot set network=cloudbuild, which can be helpful for auth reasons when running inside Google Cloud Build.

### Expected behavior
You should be able to set network=cloudbuild and ideally any other network

### Actual behavior
Validation fails when network=cloudbuild is set

idsulik:
Removed validation within this PR https://github.com/GoogleContainerTools/skaffold/pull/9390, now it supports any network

PR #9352, opened 2024-03-20
chore: updating google api and opentelemetry version

bskaplan:
**Description**

Update the Google Cloud API Client and OpenTelemetry to the latest versions.

**User facing changes (remove if N/A)**

The latest Cloud Run features will now be supported, notably built-in support for mounting gcsfuse and NFS volumes. See https://cloud.google.com/run/docs/configuring/services/cloud-storage-volume-mounts#yaml and https://cloud.google.com/run/docs/configuring/services/nfs-volume-mounts#yaml for example YAMLs.

PR #9390, opened 2024-04-17
feat(docker-network): docker.network now supports any value

idsulik:
Fixes: #9389  

**Description**
Removed docker network validation, now you can pass any network

*before*: docker.network used to support only predefined values. 
*before*: docker.network supports any value.

### DIFF
--- a/fs/assets/credits_generated/github.com/hashicorp/hcl/go.mod
+++ b/fs/assets/credits_generated/github.com/hashicorp/hcl/go.mod
@@ -1,3 +1,5 @@
 module github.com/hashicorp/hcl
 
+go 1.22
+
 require github.com/davecgh/go-spew v1.1.1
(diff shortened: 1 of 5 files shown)


fs/assets/credits_generated/github.com/hashicorp/hcl/go.mod
```diff
@@ -1,3 +1,5 @@
 module github.com/hashicorp/hcl
 
+go 1.22
```
idsulik, line 3: to fix this issue https://github.com/GoogleContainerTools/skaffold/pull/9352#issuecomment-2053899682

ericzzzzzzz:
Hi @idsulik , thank you for the pr, I think we might not need a new schema version.

idsulik:
@ericzzzzzzz , agree. replaced the commit(it was easy to replace rather than update the current)

ericzzzzzzz (approved):

idsulik:
@ericzzzzzzz hi! When the release is scheduled?

ericzzzzzzz:
Hi, @idsulik We haven't set a specific release date yet, but we expect it to be available sometime next month.

Issue #8758, opened 2023-05-09
Q: Should `skaffold diagnose` replace env vars in output?


dsiebel:
### Description

We use `skaffold diagnose --yaml-only` in CI to do some feature-detection on the resulting YAML output to identify required dependencies, like Java/Jib, Qemu+Buildkit, Cloud Run / gcloud SDK.
This allows us to only install what is needed and reduce build times.

A recent use-case uses _one_ `skaffold.yaml` to build multiple artifacts and render multiple Helm Charts (for GitOps with ArgoCD) using env vars to change what needs to change for each Chart.

In the example `skaffold.yaml` below the variable `APP_NAME` is being passed via CI env.
This works well for `skaffold build` and `skaffold render`, but when running `skaffold diagnose` the output contains `{{ .APP_NAME }}`.

Our pipelines use `metadata.name` from `skaffold diagnose --yaml-only` to determine the filename for `skaffold render --out ${filename}.yaml` in a later step:

```yaml
effective_skaffold_config="$(skaffold diagnose --yaml-only --filename "${SKAFFOLD_FILE}" --profile "${profile}")"
filename="$(yq eval '.metadata.name' - <<< "${effective_skaffold_config}")"
```

resulting in filename `{{ .APP_NAME }}.yaml`, causing `skaffold render` to overwrite the same file multiple times.

Seeing that profiles and config dependencies are evaluated before output, should env vars be replaced with their value in the output as well?
Is there another way to achieve this, that I am not aware of?

### Information

- Skaffold version: `2.4.0`
- Operating system: `macos 12.6.5`
- Installed via: Homebrew
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v3alpha1
kind: Config
metadata:
  name: {{ .APP_NAME }}
build:
  artifacts:
    - jib: {}
      # rest omitted for readability
manifests:
  helm:
    releases:
      - name: {{ .APP_NAME }}
        chartPath: chart/app
        setValues:
          nameOverride: {{ .APP_NAME }}
```

### Steps to reproduce the behavior

1. a clonable repository with the sample skaffold project
2. `skaffold diagnose`
3. Check output

dsiebel:
Awesome, thanks! 😃

PR #9393, opened 2024-04-19
feat: support templating in diagnose command

ericzzzzzzz:
Fixes: #8758  

**Description**
 - add support templating in diagnose command when using --yaml-only
    -  add skaffold:template tag to label templated fields, so we can use `go template` engine to render the specified fields, supported fields are the same as other commands see [templated fields](https://skaffold.dev/docs/environment/templating/) , 
    - the feature is guarded by a flag `enable-templating` in diagnose command, as this feature is a breaking change for users are keeping template place holder when using diagnose command. 
    - The implementation doesn't have special handling for templated fields for `ko builder` as other command which replace `{{.ENV` with `{{`, that was for ko-skaffold templates compatibilities , but we've documented that for more than one year https://skaffold.dev/docs/builders/builder-types/ko/#build-flags `.ENV` is not needed, so treating it as normal templated fields should be fine. 
    - The implementation will  error out if  key is missing when rendering templates. 

**Schema change**
 - #9401

**Follow-up Work**
 - the skaffold:template may be used for generating docs to show which fields are templated fields, we may add an issue if the approach in this pr is the way we want to implement the feature . 

**Alternative Solution**
 - unmarshal  the whole config and then render with go template. 
   - Pros:
      - easy to implement 
   - Cons:
     - The behavior is not consistent with other commands and the whole skaffold.yaml might be taken as a template file, misuse can lead to unexpected result when skaffold marshal `skaffold.yaml` -> `Skaffold.Configuration` struct

### DIFF
--- a/pkg/skaffold/tags/templates.go
+++ b/pkg/skaffold/tags/templates.go
@@ -1,0 +1,164 @@
+/*
+Copyright 2024 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tags
+
+import (
+	"fmt"
+	"reflect"
+	"slices"
+	"strings"
+
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/util"
+)
+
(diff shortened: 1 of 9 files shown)


pkg/skaffold/tags/templates.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	case reflect.Map:
+		for _, key := range v.MapKeys() {
+			vv := v.MapIndex(key)
+			if vv.Kind() == reflect.Ptr {
+				if err := expandTemplate(vv); err != nil {
+					return err
+				}
+			} else if vv.Kind() == reflect.String {
+				updated, err := util.ExpandEnvTemplate(vv.String(), nil)
```
renzodavid9, line 144: I think we are missing another `if strings.Contains(updated, "<no value>")` check here?

ericzzzzzzz: right! good catch! I added the check.

renzodavid9 (approved):

Issue #4955, opened 2020-10-27
Skaffold reloads unchanged, existing image again into the cluster


ghost:
### Expected behavior

Skaffold should not load already loaded image. In my case that's the image `eu.gcr.io/tools-org-veroo/party-service:d46578239dd325395e15e8ab16a2c499616a15e662ed196c18549e8d2bea38b6`.

### Actual behavior

Skaffold does load image `eu.gcr.io/tools-org-veroo/party-service:d46578239dd325395e15e8ab16a2c499616a15e662ed196c18549e8d2bea38b6` anew.

### Information

* Skaffold version: v1.15.0
* operating system: macOS Catalina 10.15.7

I am loading two images into my `k3d` cluster.

This is the log output:
```
Checking cache...
 - eu.gcr.io/tools-org-veroo/party-service: Found Locally
 - eu.gcr.io/tools-org-veroo/party-service-flyway: Found Locally
Tags used in deployment:
 - eu.gcr.io/tools-org-veroo/party-service -> eu.gcr.io/tools-org-veroo/party-service:d46578239dd325395e15e8ab16a2c499616a15e662ed196c18549e8d2bea38b6
 - eu.gcr.io/tools-org-veroo/party-service-flyway -> eu.gcr.io/tools-org-veroo/party-service-flyway:65572b687b0f4a65e31fe7ef32545a4fa2902a05473701a5cc9dc45e2c707a69
Loading images into k3d cluster nodes...
 - eu.gcr.io/tools-org-veroo/party-service:d46578239dd325395e15e8ab16a2c499616a15e662ed196c18549e8d2bea38b6 -> Loaded
 - eu.gcr.io/tools-org-veroo/party-service-flyway:65572b687b0f4a65e31fe7ef32545a4fa2902a05473701a5cc9dc45e2c707a69 -> Found
```

However, both images are already in my cluster. So both images should have been found.

When I execute the command `kubectl get nodes -o jsonpath='{@.items[*].status.images[*].names[*]}'` I get this output:

[...]

PR #9408, opened 2024-05-02
fix: Correctly call kubectl to get all known images of the node

ComaVN:
Fixes: #4955 

**Description**

When checking if images are already present on the k8s node, `os.exec` is used, which doesn't go through any shell expansion, so it doesn't require quoting any arguments (quotes are passed through as-is).
This means the first image returned had a `'` prepended to the name, and the last image had a `'` appended, which causes issue #4955

A few less unnecessary image reloads :)

There are a few more places where `jsonpath='(...)'` shows up in an `os.exec` call, someone might want to check those. I noticed at least some of them were "fixed" by trimming the quotes from the output.


ericzzzzzzz:
Hi @ComaVN  Thanks for the fix!!

ComaVN:
made redundant by https://github.com/GoogleContainerTools/skaffold/pull/9449

PR #9433, opened 2024-06-04
feat: Optimized fs walker and util.IsEmptyDir

idsulik:
**Description**
1. Optimized file system walker, it used to make 80k(in my case) IO operations, because in the condition `util.IsEmptyDir` was before `!info.IsDir()`, though there is no reason to make IO operation to check is the directory empty if it's not a directory. 
2. Optimized  utils.IsEmptyDir, it used to fetch all information about the directory while we just need to check if it has anything in it, so now it fetches only name

**User facing changes**
*before* - It makes N extra IO operations, where N - is docker context files count
*after* - It doesn't make extra IO operations. 

p.s. this changes optimized fs walker by 50% on my laptop

### DIFF
--- a/pkg/skaffold/docker/dependencies.go
+++ b/pkg/skaffold/docker/dependencies.go
@@ -268,7 +268,8 @@ func WalkWorkspace(workspace string, excludes, deps []string) (map[string]bool, 
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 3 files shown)


ericzzzzzzz:
@idsulik Thanks for the optimization!

ericzzzzzzz (approved):

PR #9450, opened 2024-06-19
fix: make `IMAGE_TAG` available in buildArgs when used in docker `FROM`

abe-winter:
**Related**:
- related fix for kaniko https://github.com/GoogleContainerTools/skaffold/issues/4295
- related fix  https://github.com/GoogleContainerTools/skaffold/pull/7788

**Description**
Apologies if I am simply misunderstanding something and this can be fixed w/out a code change. I am certainly not fully wrapping my head around this codebase.

This PR adds the ability to use `IMAGE_TAG` (IMAGE_REPO etc) in the docker buildArgs section, **specifically** when the build arg is then consumed in the Dockerfile `FROM`. On current stable and main, this case fails in the depList / SingleArtifactDependencies phase, with logs like:
> Checking cache...
> - base: Found Remotely
> - child: Error checking cache.
>
> getting hash for artifact "child": getting dependencies for "child": parsing ONBUILD instructions: retrieving image "gcr.io/headsdown/base:<no value>": parsing reference "gcr.io/headsdown/base:<no value>": could not parse reference: gcr.io/headsdown/base:<no value>

Note that when used with dependencies (as below), this seems to only work reliably if I run `skaffold build && skaffold run`. skaffold run by itself will sometimes fail with 'image not found remotely'.

**Repro**

Stripped-down skaffold.yaml:

```yml
build:
  local:
    useBuildkit: true
  artifacts:
    - image: base
      docker:
        dockerfile: Dockerfile
    - image: child
      requires:
      - image: base
      docker:
        dockerfile: child/Dockerfile
        buildArgs:
          IMAGE_TAG: '{{.IMAGE_TAG}}'
```

Top of `child/Dockerfile`:
```Dockerfile
# syntax=docker/dockerfile:1
ARG IMAGE_TAG
FROM gcr.io/PROJECTNAME/base:$IMAGE_TAG
```

tldr the second artifact depends on the first and then uses it as a base image

**User facing changes**
I think the docs already imply this is possible

**Follow-up Work**
No idea if I'm conforming to house style -- if accepted, guessing someone on the team will need to clean up or at least assign me cleanup. I would normally have submitted a bug rather than a PR for something like this, but I didn't understand why my template param was missing until I'd already solved the problem.

If this is something useful, happy to add tests etc.


alphanota (approved):

alphanota:
Hi @abe-winter , I cherry-picked your changes on a new PR with conflicts removed (#9664 ) So i'm going to close this PR. Thanks for your contribution!

abe-winter:
thank you! sorry I didn't get to it sooner

PR #9455, opened 2024-06-28
feat(profile): auto-activated support match exactly

beastpu:
Description
we often use env to automatically trigger the profile, the env represent different business environments and I need to implement precise matching.

For example, there are two environments called cbt and cbtcn, I want to accurately match one of them based on environment variables. However, when the environment variable is set to cbtcn, it matches both environments, which is not what I expect.

Changes
I found that auto-activated by env is fuzzy matched.

so I want to add a new boolean field named exactMatch in the activation struct, which when set to true, would implement an exact match.

### DIFF
--- a/docs-v2/config.toml
+++ b/docs-v2/config.toml
@@ -82,7 +82,7 @@ weight = 1
 copyright = "Skaffold Authors"
 privacy_policy = "https://policies.google.com/privacy"
 github_repo = "https://github.com/GoogleContainerTools/skaffold"
-skaffold_version = "skaffold/v4beta11"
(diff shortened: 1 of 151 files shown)


alphanota:
Hi @beastpu thank you for this PR. Would you be able to fix the merge conflicts and failed unit tests? We can review the PR after that. Thanks!

alphanota:
Hi @beastpu thank you for your contribution. At this time we will not be able to review your PR for the following reasons:
> Too many files changed
> The objective of the PR is not clear from the PR description and doesn't appear to be related to the changes in the PR.

Regards.

PR #9493, opened 2024-08-09
feat(sync): Add pod filter using FieldSelector

idsulik:
**Description**
Added pods filter.  
`pkg/skaffold/sync/sync.go:326`: There is no need to fetch all pods and filter based on `status.phase`, we can fetch filtered pods


ChrisGe4 (approved):
LGTM

ChrisGe4 (approved):

Issue #6992, opened 2022-01-05
Skaffold should rewrite imagePullPolicy when using a local cluster


briandealwis:
When using a local cluster like minikube, where the built images are loaded directly into the cluster's container runtime, a user will encounter see a deployment failure should they specify an [`imagePullPolicy: Always`](https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy).  Skaffold should transform the manifests and either remove such pull-policies or possibly transform them to `imagePullPolicy: Never`.

tejal29:
I think it make complete sense. I would be in to transform the manifests in memory since skaffold already parses these.

th0ger:
Fully agree with the proposal.

A workaround for me is to use `imagePullPolicy: IfNotPresent`, since I use immutable image tags. (In which case `imagePullPolicy` can also be omitted, since `IfNotPresent` is [default](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting).)

lucasrod16:
Hi @briandealwis @tejal29,

I’ve submitted PR #9495 for this feature. Your feedback would be greatly appreciated. Thanks!

PR #9495, opened 2024-08-12
feat: transform imagePullPolicy when using local cluster

lucasrod16:
Relates to #6992

**Description**

Introduces a `ReplaceImagePullPolicy` function to update the `imagePullPolicy` field in Kubernetes manifests to `Never` when running on a local cluster. This prevents deployment errors that occur when `imagePullPolicy` is set to `Always`, as detailed in #6992.

**User facing changes**

Files to reproduce:

- Dockerfile

  ```Dockerfile
  FROM nginx:alpine
  ```
- pod.yaml

  ```yaml
  apiVersion: v1
  kind: Pod
  metadata:
    name: nginx
  spec:
    containers:
    - name: nginx
      image: nginx:alpine
      imagePullPolicy: Always
  ```
- skaffold.yaml

  ```yaml
  apiVersion: skaffold/v4beta11
  kind: Config
  metadata:
    name: nginx
  build:
    artifacts:
      - image: nginx
        context: .
        docker:
          dockerfile: Dockerfile
  manifests:
    rawYaml:
     - ./pod.yaml
  ```

Before:
failed

After:
success

### DIFF
--- a/pkg/skaffold/kubernetes/manifest/image_pull_policy.go
+++ b/pkg/skaffold/kubernetes/manifest/image_pull_policy.go
@@ -1,0 +1,88 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package manifest
+
+import apimachinery "k8s.io/apimachinery/pkg/runtime/schema"
+
+// resourceSelectorImagePullPolicy selects PodSpecs for transforming the imagePullPolicy field
+// based on allowlist and denylist rules for their GroupKind and navigation path.
+type resourceSelectorImagePullPolicy struct{}
+
+func NewResourceSelectorImagePullPolicy() ResourceSelector {
+	return &resourceSelectorImagePullPolicy{}
+}
+
+// allowByGroupKind checks if a GroupKind is allowed for transformation.
+// It allows GroupKinds in the allowlist unless they are in the denylist with ".*" in PodSpec paths, which blocks all PodSpecs.
+func (rs *resourceSelectorImagePullPolicy) allowByGroupKind(gk apimachinery.GroupKind) bool {
+	if _, allowed := TransformAllowlist[gk]; allowed {
+		if rf, disallowed := TransformDenylist[gk]; disallowed {
+			for _, s := range rf.PodSpec {
+				if s == ".*" {
+					return false
+				}
+			}
+		}
+		return true
+	}
+	return false
+}
+
+// allowByNavpath checks if a GroupKind's PodSpec path (navpath) allows transformation.
+// It blocks transformation if the path matches a denylist entry or ".*".
+// If not denied, it permits transformation if the path matches an allowlist entry or ".*".
+func (rs *resourceSelectorImagePullPolicy) allowByNavpath(gk apimachinery.GroupKind, navpath string, k string) (string, bool) {
+	if rf, ok := TransformDenylist[gk]; ok {
+		for _, denypath := range rf.PodSpec {
+			if denypath == ".*" || navpath == denypath {
+				return "", false
+			}
+		}
+	}
+	if rf, ok := TransformAllowlist[gk]; ok {
+		for _, allowpath := range rf.PodSpec {
+			if allowpath == ".*" || navpath == allowpath {
+				return "", true
+			}
+		}
+	}
+	return "", false
+}
+
+func (l *ManifestList) ReplaceImagePullPolicy(rs ResourceSelector) (ManifestList, error) {
+	r := &imagePullPolicyReplacer{}
+	return l.Visit(r, rs)
+}
+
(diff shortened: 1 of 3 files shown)


lucasrod16:
Hi @ericzzzzzzz,

I noticed that this pull request has been pending for a while. Is there anything needed from me? Please let me know if there are any updates required.

Thanks!

ericzzzzzzz:
Hi, @lucasrod16  Thank you for the pr! 

I think we should only do this transformation when using `skaffold dev`, `skaffold run`, `skaffold debug` as we know that the transformed manifests will be applied to the connected local cluster, but for `skaffold render` command, we're not sure how the manifests will be used, the rendered results may be used in other places even without `skaffold`. 


Also, this pr does the requested change when kubectl deployer is used, it needs to do the similar thing when helm deployer is used.

lucasrod16:
> Hi, @lucasrod16 Thank you for the pr!
> 
> I think we should only do this transformation when using `skaffold dev`, `skaffold run`, `skaffold debug` as we know that the transformed manifests will be applied to the connected local cluster, but for `skaffold render` command, we're not sure how the manifests will be used, the rendered results may be used in other places even without `skaffold`.
> 
> Also, this pr does the requested change when kubectl deployer is used, it needs to do the similar thing when helm deployer is used.

@ericzzzzzzz That makes sense, thanks for the feedback!

I moved the transform logic from `BaseTransform` to kubectl `Deploy` and verified it works locally with `dev`, `run`, and `debug`.

I will work on adding the transform logic to the helm deployer as well

pkg/skaffold/kubernetes/manifest/image_pull_policy.go
```diff
@@ -0,0 +1,79 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package manifest
+
+import apimachinery "k8s.io/apimachinery/pkg/runtime/schema"
+
+type ResourceSelectorImagePullPolicy struct{}
```
alphanota, line 21: This should be unexported, since this new struct is only used as an implementor of the ResourceSelector interface. 

I would also add a comment about what this ResourceSelector does/ selects. For example
resourceSelectorImagePullPolicy is an implementation of ResourceSelector to select X and do Z.

pkg/skaffold/kubernetes/manifest/image_pull_policy.go
```diff
@@ -0,0 +1,79 @@
...
+
+func NewResourceSelectorImagePullPolicy() *ResourceSelectorImagePullPolicy {
```
alphanota, line 23: Should return the interface ResourceSelector instead of *ResourceSelectorImagePullPolicy.

pkg/skaffold/kubernetes/manifest/image_pull_policy.go
```diff
@@ -0,0 +1,79 @@
...
+	return &ResourceSelectorImagePullPolicy{}
+}
+
+func (rs *ResourceSelectorImagePullPolicy) allowByGroupKind(gk apimachinery.GroupKind) bool {
```
alphanota, line 27: A short comment explaining the goal of the method and the criteria used would be helpful.

pkg/skaffold/kubernetes/manifest/image_pull_policy.go
```diff
@@ -0,0 +1,79 @@
...
+					return false
+				}
+			}
+		}
+		return true
+	}
+	return false
+}
+
+func (rs *ResourceSelectorImagePullPolicy) allowByNavpath(gk apimachinery.GroupKind, navpath string, k string) (string, bool) {
```
alphanota, line 41: Here too, a description of the goal would help.

pkg/skaffold/kubernetes/manifest/image_pull_policy.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	return "", false
+}
+
+func (l *ManifestList) ReplaceImagePullPolicy(rs ResourceSelector) (ManifestList, error) {
+	r := &imagePullPolicyReplacer{}
+	return l.Visit(r, rs)
+}
+
+type imagePullPolicyReplacer struct{}
```
alphanota, line 64: A comment explaining the purpose of this struct. ie imagePullPolicyReplacer implements FieldVisitor and is used to edit an entry in a YAML document.

pkg/skaffold/kubernetes/manifest/image_pull_policy.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (i *imagePullPolicyReplacer) Visit(gk apimachinery.GroupKind, navpath string, o map[string]interface{}, k string, v interface{}, rs ResourceSelector) bool {
```
alphanota, line 66: Please add a comment explaining this method also. For example Visit sets the value of the "imagePullPolicy" field in a Kubernetes manifest to "Never".

alphanota (changes requested):
Hello! Thank  you for sending this PR! Please see my comments on it, thanks!.

Also for the PR description, since this PR doesn't have the helm changes needed to fully address the issue, I don't think we should close the bug when this PR is merged.

mepi262:
@lucasrod16 
The reviewer has commented to your pull request. How about check it?

lucasrod16:
Hey @alphanota, thanks for the PR review! I've updated the PR description to ensure the related issue isn't closed when this PR is merged and have addressed all your feedback. I currently don't have the bandwidth to implement the helm functionality, and I appreciate your understanding. Thanks again for the feedback!

alphanota (approved):

plumpy:
Given the issue reported in #9687, this appears to cause a breaking change. I'm going to go ahead and revert this PR.

I _suspect_ what needs to happen is you only want to replace the pull policy on images that were actually built by Skaffold as part of a `skaffold dev` command. But it might need more discussion.

Issue #9496, opened 2024-08-13
skaffold.yaml with only `portForward` errors out with "nothing to deploy"


johannespfrang:
We have a utility skaffold file called `skaffold.forward.yaml` which developers can use to forward some usually non-exposed ports. The file works with skaffold v1, but does not work with skaffold v2.

### Expected behavior

```bash
$ skaffold1 run -f ./skaffold.forward.yaml --port-forward=user
Listing files to watch...
WARN[0000] k8s/*.yaml did not match any file             subtask=-1 task=DevLoop
Generating tags...
Checking cache...
Tags used in deployment:
Starting deploy...
WARN[0000] k8s/*.yaml did not match any file             subtask=-1 task=DevLoop
Waiting for deployments to stabilize...
Deployments stabilized in 11.509247ms
Port forwarding Service/ingress-nginx-controller in namespace ingress-nginx, remote port 443 -> http://127.0.0.1:443
Port forwarding Service/ingress-nginx-controller in namespace ingress-nginx, remote port 80 -> http://127.0.0.1:80
```

### Actual behavior

```bash
$ skaffold2 run -f ./skaffold.forward.yaml --port-forward=user
No tags generated
Starting test...
WARN[0000] k8s/*.yaml did not match any file             subtask=-1 task=DevLoop
Starting deploy...
nothing to deploy
```

### Information

- Skaffold version:  v2.13.1
- Operating system: Ubuntu 24.04
- Installed via: skaffold.dev
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v2beta28
kind: Config
portForward:
  - resourceType: Service
    resourceName: ingress-nginx-controller
    namespace: ingress-nginx
    port: 80
    address: 127.0.0.1
    localPort: 80
  - resourceType: Service
    resourceName: ingress-nginx-controller
    namespace: ingress-nginx
    port: 443
    address: 127.0.0.1
    localPort: 443
```

(simplified example, we have many other forwards)

### Steps to reproduce the behavior

1. Install ingress-nginx
2. Run the above yaml with skaffold v1 vs skaffold v2

PR #9497, opened 2024-08-14
fix(kubectl): Allow port-forwarding if there is nothing to deploy

idsulik:
Fixes: #9496

**Description**
If there is nothing to deploy, we can use skaffold for port-forwarding so there is no need to return error if there is nothing to deploy


menahyouyeah:
This would unfortunately for the majority of users mean that they may not realize their configs are misconfigured and they aren't actually deploying anything. I would suggest 2 options instead 1) a shell script to run kubectl for port forwarding or 2) setting up a separate command in Skaffold that would only set up port forwarding

PR #9513, opened 2024-08-29
chore: remove unused taggers field

cuishuang:
Fixes: #nnn 

**Description**

remove no-used taggers field

### DIFF
--- a/pkg/skaffold/tag/tagger_mux.go
+++ b/pkg/skaffold/tag/tagger_mux.go
@@ -26,7 +26,6 @@ import (
 )
 
 type TaggerMux struct {
-	taggers     []Tagger
 	byImageName map[string]Tagger
 }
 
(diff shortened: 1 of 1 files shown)


plumpy (changes requested):
Interesting, I was wondering why this wasn't a problem before, and it looks like it's because the `taggers` field is never actually used. Can you just delete it instead?

Thanks for the PR!

cuishuang:
> Interesting, I was wondering why this wasn't a problem before, and it looks like it's because the `taggers` field is never actually used. Can you just delete it instead?
> 
> Thanks for the PR!

Of course. Please review again.

pkg/skaffold/tag/tagger_mux.go
```diff
@@ -41,18 +41,16 @@ func (t *TaggerMux) GenerateTag(ctx context.Context, image latest.Artifact) (str
...
 		if err != nil {
 			return nil, fmt.Errorf("creating tagger: %w", err)
 		}
-		sl = append(sl, t)
 		for _, a := range p.Build.Artifacts {
 			m[a.ImageName] = t
 		}
 	}
-	return &TaggerMux{taggers: sl, byImageName: m}, nil
+	return &TaggerMux{byImageName: m}, nil
```
plumpy, line 53: Can you also remove the field from the TaggerMux struct (on line 29)?

Thanks!

cuishuang: My pleasure.

Please review again.

plumpy:
Thanks for sending the PR, we appreciate it!

cuishuang:
> Thanks for sending the PR, we appreciate it!

I am very happy to do this.

PR #9451, opened 2024-06-25
feat: Optimize helm deploy by using goroutines

idsulik:
**Description**
1. Optimized helm install by using goroutines which reduced deploy time by more than 50% 
2. Added config option to control the concurrency, by default it keeps current logic - sequential installation
3. Minor fixes

in my case:
before: _Deploy completed in **3 minutes 52.351** seconds_
after: _Deploy completed in **1 minute 57.148** seconds_

colleague:
before: _Deploy completed in _4 minutes 4.594_ seconds_
after: _Deploy completed in **53.423** seconds_

PR #9520, opened 2024-09-06
feat(bin): Add graceful shutdown for helm command

idsulik:
**Description**
Added graceful shutdown for helm command

the CommandContext kills the process when we cancel the context  
image

it leads to an inconsistent state in helm charts, the helm left chart in "pending-install" state.  
image

How to reproduce the issue:
1. start a deployment using the helm 
2. in the middle of the installation stop the process using `Ctrl/CMD + C`
3. check the release status using `helm status {releaseName}`, the status will be "pending-install"
4. try to deploy again and you won't be able to do it, because helm will return `helm another operation (install/upgrade/rollback) is in progress`

This fix gives the helm 15 seconds to rollback(if you pass `--atomic`) and/or save the actual(`failed`) state of the release

### DIFF
--- a/pkg/skaffold/helm/bin.go
+++ b/pkg/skaffold/helm/bin.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"time"
 
 	"github.com/blang/semver"
 	shell "github.com/kballard/go-shellquote"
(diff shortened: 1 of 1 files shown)


idsulik:
Hello @plumpy, just a gentle reminder about this PR created months ago. If possible, could it be reviewed before the next Helm release? I’d appreciate any feedback to help get it ready in time. Thank you

idsulik:
Hi @plumpy, please check the PR, because it's very important, especially with these changes https://github.com/GoogleContainerTools/skaffold/pull/9451, because if a user executes parallel package installation and interrupts it, all the packages will remain in pending state, because skaffold kills the helm command execution on context cancellation.
Thank you!

plumpy:
Okay, I played around with this some. For testing, I [made my own little binary](https://gist.github.com/plumpy/990c330ddba06f7230d02bb71b759079) that waits for a `SIGINT`, then waits around another `n` seconds before exiting.

From playing around with that, it seems to me like you can just use the built-in support in the `cmd` package, rather than rolling your own goroutine:

```
cmd := exec.CommandContext(ctx, "helm", args...)
cmd.Cancel = func() error {
  fmt.Println("Terminating helm, giving it 2 minutes to clean up...")
  return cmd.Process.Signal(os.Interrupt)
}
cmd.WaitDelay = 120 * time.Second
```

I tried a variety of combinations, where the process is still running after `WaitDelay` (it [gets sent a `SIGKILL`](https://github.com/golang/go/blob/b68f8ca89a2fb6b33d1e78fadc33d1c35693f6b8/src/os/exec/exec.go#L828)) or where the process exits before the `WaitDelay` (Skaffold exits as soon as it does, not waiting around the full `WaitDelay` time), so it seems like it does what you want. Can you give that a shot and see if it works for you?

idsulik:
@plumpy you're right, it's a more concise way to do what I wanted, good job, I didn't know about the WaitDelay. 
I've just tried it several times with concurrent installation(https://github.com/GoogleContainerTools/skaffold/pull/9451) and it looks like it works. 

thank you!

p.s. Is there an approximate time when the next release will be published?

plumpy:
> p.s. Is there an approximate time when the next release will be published?

Hopefully relatively soon, but we need to figure out why the internal presubmits are failing.

idsulik:
@plumpy if you mean this error
```
2024/11/22 16:25:23 error signing scorecard results: getting key from Fulcio: retrieving cert: error obtaining token: expired_token
2024/11/22 16:25:23 retrying in 10s...
2024/11/22 16:25:33 error signing scorecard json results: error signing payload: getting key from Fulcio: retrieving cert: error obtaining token: expired_token
```
(c) https://github.com/GoogleContainerTools/skaffold/actions/runs/11976028911/job/33390752067

then it looks like the secret `SCORECARD_READ_TOKEN` should be updated with a new token

plumpy (approved):

PR #9521, opened 2024-09-08
fix: Set the client DialContext to the connhelper dialer DOCKER_HOST is present

### LINKED ISSUE #9484: Wrong host for docker API calls when using an ssh DOCKER_HOST

aran:
Summary: Fixes #9484

Fixes: #9484 

**Description**
This was tested with and without #9512 in place.

Test Plan: Tested skaffold over a ssh connection to minikube.

### DIFF
--- a/pkg/skaffold/docker/client.go
+++ b/pkg/skaffold/docker/client.go
@@ -104,7 +104,11 @@ func newEnvAPIClient() ([]string, client.CommonAPIClient, error) {
 					DialContext: helper.Dialer,
 				},
 			}
-			opts = append(opts, client.WithHTTPClient(httpClient), client.WithHost(helper.Host))
+			opts = append(opts,
+				client.WithHTTPClient(httpClient),
+				client.WithHost(helper.Host),
+				client.WithDialContext(helper.Dialer),
+			)
 		} else {
 			opts = append(opts, client.FromEnv)
 		}


plumpy (approved):
Thanks!

Issue #9522, opened 2024-09-09
Support disabling Cloud Run URL - populate CloudRunServiceReady even if URL is not there


menahyouyeah:
### Expected behavior
Skaffold populates the `CloudRunReadyEvent` struct when a deployment is successful to Cloud Run when the following annotations have been added to the Cloud Run manifest:

run.googleapis.com/default-url-disabled:true
run.googleapis.com/launch-stage: BETA

This is using a [new feature](https://cloud.google.com/run/docs/securing/ingress#disable-url) from Cloud Run.

### Actual behavior
Because of [this line](https://github.com/GoogleContainerTools/skaffold/blob/1d83c098041dc58b9da309941480a34b706fe408/pkg/skaffold/deploy/cloudrun/status.go#L348), if the URL is not populated, `reportSuccess()` does not populate `CloudRunServiceReady`.

### Information

```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
name: deploy-run-service-dev
annotations:
run.googleapis.com/launch-stage: 'BETA'
run.googleapis.com/default-url-disabled: 'true'
spec:
template:
spec:
containers:
- image: my-app-image
```

### Steps to reproduce the behavior

Try to deploy a Cloud Run service using the above manifest. To repro, use this [quick start](https://cloud.google.com/deploy/docs/deploy-app-run) with the service yaml modified per the above.

PR #9523, opened 2024-09-10
fix: emit CloudRunServiceReady event even if default url is disabled

bskaplan:
Fixes: #9522 

**Description**
Only emit CloudRunServiceReady event if the status is success, and emit it even if no URL is present which could be the case if default-url-disabled is set to true

Before:
A creation or update of a service with annotation run.googleapis.com/default-url-disabled: true would not emit a CloudRunServiceReady event.
If an update of an existing service failed, CloudRunServiceReady event would be emitted.

After:
A CloudRunServiceReady event is only populated if the Ready condition is successful, and is populated even if the default URL is disabled.

### DIFF
--- a/pkg/skaffold/deploy/cloudrun/status.go
+++ b/pkg/skaffold/deploy/cloudrun/status.go
@@ -350,7 +350,7 @@ func (r *runServiceResource) reportSuccess() {
 		// event status
 		url = "-"
 	}
-		eventV2.CloudRunServiceReady(r.path, url, r.latestRevision)
+	eventV2.CloudRunServiceReady(r.path, url, r.latestRevision)
 }
 
 type runJobResource struct {
--- a/pkg/skaffold/deploy/cloudrun/status.go
+++ b/pkg/skaffold/deploy/cloudrun/status.go
@@ -302,10 +302,10 @@ func (s *Monitor) printStatusCheckSummary(out io.Writer, c *counter, r *runResou
 		return
 	}
 	eventV2.ResourceStatusCheckEventCompleted(r.resource.String(), curStatus.ae)
-	r.sub.reportSuccess()
 	if curStatus.ae.ErrCode != proto.StatusCode_STATUSCHECK_SUCCESS {
 		output.Default.Fprintln(out, fmt.Sprintf("Cloud Run %s %s failed with error: %s", r.resource.Type(), r.resource.Name(), curStatus.ae.Message))
 	} else {
+		r.sub.reportSuccess()
 		output.Default.Fprintln(out, fmt.Sprintf("Cloud Run %s %s finished: %s. %s", r.resource.Type(), r.resource.Name(), curStatus.ae.Message, c.remaining()))
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/skaffold/deploy/cloudrun/status.go
```diff
@@ -344,9 +344,13 @@ func (r *runServiceResource) getTerminalStatus(crClient *run.APIService) (*run.G
 	return ready, nil
 }
 func (r *runServiceResource) reportSuccess() {
-	if r.url != "" {
-		eventV2.CloudRunServiceReady(r.path, r.url, r.latestRevision)
+	url := r.url
+	if url == "" {
```
idsulik, line 348: I think it'll be better to check if the service definition contains the feature flags:  
```
run.googleapis.com/launch-stage: 'BETA'
run.googleapis.com/default-url-disabled: 'true'
```

pkg/skaffold/deploy/cloudrun/status.go
```diff
@@ -344,9 +344,13 @@ func (r *runServiceResource) getTerminalStatus(crClient *run.APIService) (*run.G
...
+		// a URL may not be present if the default URL is disabled. Use - instead of empty in case anyone is parsing the
+		// event status
+		url = "-"
```
idsulik, line 351: It's not obvious that the URL can be not URL, but "dash". what if someone uses it like this:
```
if r.url != "" {
   ....
}
```
the condition will be true, because the URL is not empty, but contains "-"

idsulik:
@bskaplan , thank you for the contribution, I'm not the maintainer, but left couple of comments

pkg/skaffold/deploy/cloudrun/status.go
bskaplan: I don't want to hard-code this to that specific feature (I also definitely don't want to base it on the launch stage because that'll be removed when the feature goes GA). I think it makes more sense to trigger the event any time the service goes ready regardless of whether there's a URL or not, and as this feature shows a URL being present is not a requirement for a service going ready.

pkg/skaffold/deploy/cloudrun/status.go
bskaplan: ok, keeping it as empty

pkg/skaffold/deploy/cloudrun/status.go
```diff
@@ -344,9 +344,8 @@ func (r *runServiceResource) getTerminalStatus(crClient *run.APIService) (*run.G
 	return ready, nil
 }
 func (r *runServiceResource) reportSuccess() {
-	if r.url != "" {
-		eventV2.CloudRunServiceReady(r.path, r.url, r.latestRevision)
-	}
+	url := r.url
```
renzodavid9, line 347: nit: looks like we don't need this `url` variable anymore

renzodavid9:
Hey @bskaplan, thanks for helping us with this one 😄. Just a small nit, but it looks good. Just waiting for the integration test to finish.

renzodavid9 (approved):

Issue #9409, opened 2024-05-03
Custom actions ignores container volumes


Isakdl:
When I run the custom action my container settings in my job manifest are overridden by the container specification in the custom actions. When I use the `overrides` keyword it does not seem to apply these overrides to the final yaml file. 

### Expected behavior

That the container mounts the volumes.

### Actual behavior

The volume mounts are overridden and ignored. 

### Information

Similar issue: https://github.com/GoogleContainerTools/skaffold/issues/8974

- Skaffold version: 2.11.1
- Operating system: MacOS Sonoma 14.4.1
- Installed via:  Homebrew
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v4beta7
kind: Config
metadata:
  name: gke-automation
manifests:
  rawYaml:
    - deployment.yaml
    - service.yaml
    - http-route.yaml
    - secret.yaml
    - service-account.yaml
deploy:
  kubectl: {}
customActions:
  - name: custom-job
    executionMode:
      kubernetesCluster:
        overrides: '{"spec":{"template":{"spec":{"containers":[{"volumeMounts":[{"name":"config","mountPath":"/config/config.yaml","subPath":"config.yaml"},{"name":"secrets","mountPath":"/config/secrets.yaml","subPath":"secrets.yaml"}]}]}}}}'
        jobManifestPath: ./job.yaml
    containers:
      - name: migration
        image: image
```

Job.yaml
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: migration
  namespace: app-space
  labels:
    app: migration
spec:
  backoffLimit: 0
  template:
    metadata:
      annotations:
        gke-gcsfuse/volumes: "true"
      labels:
        app: migration
    spec:
      serviceAccountName: service-account
      image: image
      containers:
        - name: migration
          volumeMounts:
            - name: config
              mountPath: /config/config.yaml
              subPath: config.yaml
            - name: secrets
              mountPath: /config/secrets.yaml
              subPath: secrets.yaml
      volumes:
        - name: config
          csi:
            driver: gcsfuse.csi.storage.gke.io
            readOnly: true
            volumeAttributes:
              bucketName: config-bucket
              mountOptions: "implicit-dirs"
              gcsfuseLoggingSeverity: warning
        - name: secrets
          csi:
            driver: secrets-store-gke.csi.k8s.io
            readOnly: true
            volumeAttributes:
              secretProviderClass: secrets
```

[...]

PR #9532, opened 2024-09-26
fix: change k8s job container creation logic to retain existing value…

cmanzi-bestow:
…s from job manifest

Fixes GoogleContainerTools/skaffold#9409

Fixes: #9409  

**Description**

This changes the logic that applies the provided container properties for executing a custom action (when using a custom job manifest) so that only the provided values (if not empty) are overwritten, and the container is not replaced entirely. I have done an initial test and this does now allow the preservation of volumeMounts and other properties that are currently erased. 

Using the overrides config does not work here, as the overrides are applied prior to the logic that replaces the container.

This is a proposed solution for the referenced issue. Having looked under the hood while trying to debug this, I'm not exactly sure what the intended behavior was when providing a custom manifest, so I am open to other possible solutions - perhaps if a custom manifest is provided, the `containers` block should be unnecessary, since theoretically this should already be defined in the manifest?

This is currently requiring us to do a pretty extensive workaround to get this working with Cloud Deploy, as we need to run pre-deploy migrations for most of our services, and we have existing manifests that we need to use (the cut-down Container interface isn't sufficient for our purposes). 

This would change the current behavior (which appears to be a bug, at least according to what one would expect) when providing a custom job manifest for a Kubernetes custom action. For example: 
```yaml
customActions:
- name: db-migration
  containers:
  - name: db-migrate
    image: db-migration-image:latest
    command:
    - /bin/sh
    - -c
    - run-db-migration
  executionMode:
    kubernetesCluster:
      jobManifestPath: migration-job.yaml
```

### DIFF
--- a/pkg/skaffold/actions/k8sjob/task.go
+++ b/pkg/skaffold/actions/k8sjob/task.go
@@ -146,7 +146,26 @@ func (t Task) getK8SEnvVars(envVars []latest.VerifyEnvVar) (k8sEnvVar []corev1.E
 }
 
 func (t *Task) setManifestValues(c corev1.Container) {
-	t.jobManifest.Spec.Template.Spec.Containers = []corev1.Container{c}
+	if len(t.jobManifest.Spec.Template.Spec.Containers) == 0 {
+		t.jobManifest.Spec.Template.Spec.Containers = []corev1.Container{c}
+	} else {
+		if c.Name != "" {
+			t.jobManifest.Spec.Template.Spec.Containers[0].Name = c.Name
+		}
+		if c.Image != "" {
+			t.jobManifest.Spec.Template.Spec.Containers[0].Image = c.Image
+		}
+		if len(c.Command) > 0 {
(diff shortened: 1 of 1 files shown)


cmanzi-bestow:
I am working on some tests right now, I just wanted some initial feedback on the solution I'm proposing.

It would help if I could better understand what the intended behavior is when using a custom K8s job manifest, as how it's set up right now really doesn't make sense to me.

mattsanta:
Thanks for taking the time to look into this.

So turns out that the `verify` implementation, which also supports providing a Job manifest properly handles this by [**only** replacing the container fields that can be configured in the Skaffold Config](https://github.com/GoogleContainerTools/skaffold/blob/414ab983b1c775cd09a1cdfd22c9919e8e759afc/pkg/skaffold/verify/k8sjob/verify.go#L385) (i.e. image, command, args, name, and env (appends to this)).

To fix the issue for custom actions we should mimic the behavior that verify currently implements. Could you update the PR with these changes and add some tests?

PR #9565, opened 2024-11-13
feat: Add network option to BuildpackArtifact

idsulik:
Fixes: https://kubernetes.slack.com/archives/CABQMSZA6/p1731095108637869

**Description** 
Add `network` to the `BuildpackArtifact` so the user can configure network settings of the build containers.

**User facing changes**
**Before**: no way to specify network for the buildpacks
**After**: there is a way to specify network for the buildpacks

### DIFF
--- a/docs-v2/content/en/schemas/v4beta12.json
+++ b/docs-v2/content/en/schemas/v4beta12.json
@@ -1031,6 +1031,11 @@
             "[\"key1=value1\", \"key2=value2\", \"key3={{.ENV_VARIABLE}}\"]"
           ]
         },
+        "network": {
+          "type": "string",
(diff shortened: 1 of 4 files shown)


menahyouyeah:
I was unable to reproduce the issue from the slack kubernetes channel. When running the quickstart, I eventually see a line `Status: Downloaded newer image for gcr.io/buildpacks/gcp/run:v1`. Can you give a little more context on how you were able to repro the issue and how this change fixes it?

idsulik:
@menahyouyeah hi! I haven't reproduced it, decided that the feature will be useful anyway

bogdannazarenko:
#9745 and #9744

PR #9572, opened 2024-11-15
chore: upgrade buildpacks (and transitive dependencies)

plumpy:
I started with just buildpacks and all of its transitive dependencies, since it seemed to be the most troublesome.

More to come in future PRs.


mattsanta:
Looks like there's also issues with the CI build.

PR #9580, opened 2024-11-21
fix: Handle `StandalonePods` `Succeeded` case when checking status

Y--:
**Description**
When Skaffold checks a standalone pod's status (method `checkStandalonePodsStatus`) , it currently only considers `Failed` and `Running` state, any other status will lead the pod to be systematically added to the `pendingPods` collection.

According to [the documentation](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/), a pod can be in 5 different states: `Pending`, `Running`, `Succeeded`, `Failed` or `Unknown`.

While it make sense for `Pending` or `Unknown` pods to be added to the `pendingPods` collection, I think it does not for the `Succeeded` ones (pods that terminated successfully)

This PR introduces a tiny change which just skips `Succeeded` pods.

**Tests**

I tried to look at other test and could only find:
* unit tests on `CheckStatus` that don't involve `StandalonePods` resource
* `TestPollServiceStatus`, `TestPollJobStatus`, `TestPollDeployment`, none of which involve `StandalonePods` resource

Please let me know if I've missed it, I'd be happy to add test on it.

Thanks!

### DIFF
--- a/pkg/skaffold/kubernetes/status/resource/deployment.go
+++ b/pkg/skaffold/kubernetes/status/resource/deployment.go
@@ -166,6 +166,8 @@ func (r *Resource) checkStandalonePodsStatus(ctx context.Context, cfg kubectl.Co
 			if podReady, _ := strconv.ParseBool(string(b)); !podReady {
 				pendingPods = append(pendingPods, pod.Name())
 			}
+		case "Succeeded":
+			log.Entry(ctx).Debugf("pod '%s' succeeded - ignoring", pod.Name())
 		default:
 			pendingPods = append(pendingPods, pod.Name())
 		}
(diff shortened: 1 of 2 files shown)


mattsanta:
Thanks for the PR!

So the current behavior is that status checking does not work if the Pod has completed and has a `Succeeded` phase, skaffold will continue to poll the Pod status until it times out. I also went ahead and tested your fix by building skaffold locally and it looks like it works.

I know the current tests don't ever use `StandalonePods`, but could you be the first one to add a test case that does for this scenario?

Y--:
Thanks for the review @mattsanta ! I have added a few tests.
I also had to skip event publishing in `CheckStatus` for `Succeeded` pods, which was failing for these pods.

mattsanta (approved):

mattsanta:
> Thanks for the review @mattsanta ! I have added a few tests. I also had to skip event publishing in `CheckStatus` for `Succeeded` pods, which was failing for these pods.
> 
> Also, I'm not sure why the conventional commit linter isn't happy with me... I must be missing something big, but I think I have a subject & type for each commit message?

It's referring to the title of the PR. I went ahead and added `fix: ` prefix to it :)

dntosas:
@mattsanta hey! are you aiming to release this anytime soon? :)

Issue #9581, opened 2024-11-23
Feature: Support TemplatedField for build.artifacts.docker.cliFlags


kitagry:
### Expected behavior

Upon installing Docker Buildx, the docker build command becomes an [alias for docker buildx build](https://docs.docker.com/reference/cli/docker/buildx/build/). This means that any subsequent docker build commands will be executed using the more advanced features provided by Buildx. Specifically, we intend to leverage the `--cache-to=type=registry,ref=repo/image:tag` flag within these commands. In order to several registry, I want to TemplateField for `build.artifacts.docker.cliFlags` as following.

```yaml
build:
  artifacts:
    - image: repo/image
      context: .
      docker:
        dockerfile: Dockerfile
        cliFlags:
          - "--cache-to={{ .IMAGE_REPO }}/{{ .IMAGE_NAME }}:latest-cache"
```

### Actual behavior

We can't use TemplateField with cliFlags.

PR #9582, opened 2024-11-23
Support TemplateField for build.artifacts.docker.cliFlags

kitagry:
Fixes: #9581

**Description**

Upon installing Docker Buildx, the docker build command becomes an [alias for docker buildx build](https://docs.docker.com/reference/cli/docker/buildx/build/). This means that any subsequent docker build commands will be executed using the more advanced features provided by Buildx. Specifically, we intend to leverage the `--cache-to=type=registry,ref=repo/image:tag` flag within these commands. In order to several registry, I want to TemplateField for build.artifacts.docker.cliFlags as following.

```yaml
build:
  artifacts:
    - image: repo/image
      context: .
      docker:
        dockerfile: Dockerfile
        cliFlags:
          - "--cache-to={{ .IMAGE_REPO }}/{{ .IMAGE_NAME }}:latest-cache"
```

User can use TemplateField in cliFlags.


pkg/skaffold/build/docker/docker.go
```diff
@@ -106,6 +106,12 @@ func (b *Builder) dockerCLIBuild(ctx context.Context, out io.Writer, name string
 	if err != nil {
 		return "", fmt.Errorf("unable to evaluate build args: %w", err)
 	}
+	for i, cliFlag := range a.CliFlags {
```
louisjimenez, line 109: Is this something that can be moved to where the other cli args are built: https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/docker/image.go#L629? I believe that would also add support for the GCB case.

kitagry: Sorry for the late reply. I fixed it!

louisjimenez (approved):

kitagry:
Are there any plans to release a new version?

PR #9584, opened 2024-11-25
fix: Update the k8s Job container logic for custom actions to match v…

mattsanta:
…erify

Fixes https://github.com/GoogleContainerTools/skaffold/issues/9409

**Description**

This PR updates the logic when it comes to custom actions with a provided Job manifest to be similar to [verify](https://github.com/GoogleContainerTools/skaffold/blob/414ab983b1c775cd09a1cdfd22c9919e8e759afc/pkg/skaffold/verify/k8sjob/verify.go#L385). What this means is that any container fields defined in the Job manifest that are not configurable in the customActions stanza are kept, e.g. `volumes`.  It also appends env vars rather than replacing to be consistent with verify as well.

Frankly, I'm not sure why the Job execution for verify and custom actions were implemented in two different places which caused these deviations. I suspect there are other ways in which they differ, just from a quick glance at the two implementations.

Unfortunately, it also looks like executing the k8s Jobs for verify and custom actions lacks unit testing. In order to get this in I've just gone ahead and added a *very* basic test for the container patching logic, same as what exists for verify [here](https://github.com/GoogleContainerTools/skaffold/blob/414ab983b1c775cd09a1cdfd22c9919e8e759afc/pkg/skaffold/verify/k8sjob/verify_test.go#L28).

Any fields that are defined in the Job container manifest that are not configurable in the customActions container stanza will be kept. In other words, any fields that are **not** name, image, command, or args.

### DIFF
--- a/pkg/skaffold/actions/k8sjob/task.go
+++ b/pkg/skaffold/actions/k8sjob/task.go
@@ -87,7 +87,17 @@ func (t Task) Exec(ctx context.Context, out io.Writer) error {
 	}
 
 	c := t.getContainerToDeploy()
-	t.setManifestValues(c)
+	var original corev1.Container
+	for _, ec := range t.jobManifest.Spec.Template.Spec.Containers {
+		if ec.Name == c.Name {
+			original = ec
+			break
+		}
+	}
+	patchToK8sContainer(c, &original)
+
+	t.jobManifest.Spec.Template.Spec.Containers = []corev1.Container{original}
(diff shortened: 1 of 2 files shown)


cmanzi-bestow:
@mattsanta Thanks for addressing this! Has this change been rolled out to GCP (Cloud Deploy)?

plumpy:
We're working to release a v2.14 soon, and then we'll add it to Cloud Deploy within a few weeks after that. Sorry for the delay.

Issue #5417, opened 2021-02-19
Run helm concurrently


satazor:
This a feature request to add `concurrency` (default being `1`) to the helm deployer, just like we have for some builders.

Currently, helm releases are deployed sequentially , which is slow if a given `skaffold.yaml` has many releases. It would be awesome if we could run these concurrently, to speed up the process.

tejal29:
sounds like a good idea. Can you give us more details on 
1) how many helm charts do you have?
2) If one fails, do you envision the already deployed ones being cleaned up?

tejal29:
relates to #5442

satazor:
1. We have a mono-repository with currently 7 micro-services a two gateways (graphql and downloads). So it's a total of 9 helm releases, but it will likely increase in the future, as the product we are developing grows.

2. Good question. I would say that to not cleanup / revert, but it could maybe be an option? Currently, if a helm installation/upgrade fails, the deployment process stops.

bartmanus:
I wholeheartedly second this proposal. We also have a monorepo but with about 50 independent charts.

Deploying them with run on skaffold 1.19.0 takes just over 2 minutes, while cleanup/delete takes over 1 minute on an i9 2018 MBP. It is obvious from the sequential output that concurrency would help greatly in both use cases. Our current dev stack orchestrator is docker-compose and, while admittedly much more limited, is also much faster at about 20 seconds for bringing the stack up and down.

Regarding what to do on deploy failure - just leave the existing releases be. This is also the current behaviour of the sequential deployer.

tejal29:
@aaron-prindle something to look at for performance.

iosifnicolae2:
Also, we need this as our deployments are pretty slow because of this issue..

More context here: https://github.com/GoogleContainerTools/skaffold/issues/8363

davidgwps:
This feature would be greatly appreciated, we have a monorepo with 10+ workloads that we are deploying together to aid local development and avoid deploying ephemeral environments into GKE. The lack of concurrency is a major headache as deployments being sequential results in us waiting 10+ minutes for initial deployment by Skaffold into Minikube and 5+ minutes for iteration.

iosifnicolae2:
@davidgwps as a temporary solution, we've splitted the deployments into multiple skaffold files and then we've used `parallel` to execute them:
```bash
cd parallel_deployment
[...]

PR #9588, opened 2024-11-28
feat: helm concurrent deploys with cross dependency

orlandoburli:
Fixes: #5417

**Description**
This PR is adding a option for concurrent helm releases to wait for other helm charts.

```yaml
deploy:
  helm:
    concurrency: 3
    releases:
      - name: service1
        chartPath: charts/service1
      - name: service2
        chartPath: charts/service2
      - name: service3
        chartPath: charts/service3
        dependsOn:
          - service1
          - service2
```

On the following example, the `service3` will wait until helm deploy for `service1` and `service2` to finish. If some of those services fails, the `service3` deployment won't start.

**User facing changes**

User can add the `dependsOn` option to a helm release template.

### DIFF
--- a/pkg/skaffold/deploy/helm/helm.go
+++ b/pkg/skaffold/deploy/helm/helm.go
@@ -261,6 +261,8 @@ func (h *Deployer) Deploy(ctx context.Context, out io.Writer, builds []graph.Art
 
 	olog.Entry(ctx).Infof("Deploying with helm v%s ...", h.bV)
 
+	releaseChannels := make(map[string]chan bool)
+
 	nsMap := map[string]struct{}{}
 	manifests := manifest.ManifestList{}
 	g, ctx := errgroup.WithContext(ctx)
@@ -271,6 +273,15 @@ func (h *Deployer) Deploy(ctx context.Context, out io.Writer, builds []graph.Art
 	} else {
 		g.SetLimit(*h.Concurrency)
 		olog.Entry(ctx).Infof("Installing %d releases concurrently", len(h.Releases))
+
+		for _, r := range h.Releases {
+			// Check if there is any process that depends on this deployment
+			if h.hasDependentReleases(r.Name) {
+				if _, ok := releaseChannels[r.Name]; !ok {
+					releaseChannels[r.Name] = make(chan bool)
+				}
+			}
+		}
 	}
 
 	var mu sync2.Mutex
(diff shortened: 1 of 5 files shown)


pkg/skaffold/deploy/helm/helm.go
```diff
@@ -295,8 +307,33 @@ func (h *Deployer) Deploy(ctx context.Context, out io.Writer, builds []graph.Art
 				return helm.UserErr(fmt.Sprintf("cannot expand chart path %q", r.ChartPath), err)
 			}
 
+			for _, dependency := range r.DependsOn {
```
idsulik, line 310: what about nested dependencies?  
ServiceA depends on ServiceB and ServiceB depends on Service C

idsulik:
@orlandoburli welcome! good feature! 
I think you can use topological sorting to achieve this or a simpler way:  
1. put releases in some data structure to extract them level by level.  
2. lock each level using semaphore.   

flow:
1. start installing deepest level dependencies
2. lock until all the releases will be installed
3. start installing next level dependencies
4. lock
5. etc.

pkg/skaffold/deploy/helm/helm.go
orlandoburli: I have not created a test for this case but, in theory, it should work the same way. Here's why:

* The first action is to check all services that has dependents, then create the channels to use as wait controllers;
* So, In the case you're suggesting:
  * If ServiceC starts first, it will hit the signal break with channels; It's gonna wait for Service B to continue;
  * If ServiceB starts in second, it will hit the signal break too, and wait for ServiceA
  * ServiceA does not depends on anyone, so it deploys and send the "all clear" signal to his dependents
  * ServiceB get's the "all clear" signal, and starts this deployment. When finished, it will also sends his signal for ServiceC
  * Finally, ServiceC get's his all clear and do this deployment.

One scenario that **may** happen is a deadlock. I did not think about it on this PR. 

Another thing that I'm thinking is that this deserve a integration testing. I'm still learning about this repo and I will create a separated PR for that. This PR already has the label "size/L" 😱

idsulik: I suggest you to check this approach https://github.com/GoogleContainerTools/skaffold/pull/9588#issuecomment-2505671935, because the current approach will lead to deadlock(concurrency = 3 and 3 dependencies are waiting, for example).    
don't worry about the "size/L", it's ok)

orlandoburli: Ok thanks for the feedback @idsulik . I'll work on that. Now I see the point, it's easy to get to a deadlock situation because of the lack of resources.

mepi262:
@orlandoburli 
It seems that some checks are not successful.
Would you confirm it?

orlandoburli:
> @orlandoburli It seems that some checks are not successful. Would you confirm it?

Hey @mepi262 yes, I saw that. I will implement first the suggestion made by @idsulik before I jump into this. But thanks for the heads up.

mepi262:
@orlandoburli 
I'm looking forward to this feature and I hope this project will be success and be released.
Thank you for your reply!

idsulik:
I've implemented this idea https://github.com/GoogleContainerTools/skaffold/pull/9588#issuecomment-2505671935 within this PR https://github.com/GoogleContainerTools/skaffold/pull/9624

@orlandoburli I'll close my PR if you'd prefer to implement it yourself, just let me know! because you started it first)

orlandoburli:
> I've implemented this idea [#9588 (comment)](https://github.com/GoogleContainerTools/skaffold/pull/9588#issuecomment-2505671935) within this PR #9624
> 
> @orlandoburli I'll close my PR if you'd prefer to implement it yourself, just let me know! because you started it first) If you want to keep my PR, then could you please check the implementation, if it works for you or not

Hey @idsulik let's go with your's PR, I didn't had the time to do it yet.

PR #9591, opened 2024-11-30
Add support for OpenTofu

abergmeier:
**Description**
Very simple initial config by introducing a new Deployment `tofu`. Simply uses an initialized Workspace to apply.

Since this needs an extension to _4beta11_ schema introduces a new version _4beta12_.

**User facing changes**
Has new `tofu` deploy struct with an obligatory `workspace` parameter.

Also introduces an integration test to ensure this works.


oprudkyi:
@abergmeier thank you for effort, 
I haven't checked deeply 
but it seems pkg/skaffold/tofu/version_test.go expects version 0.16 
```
			commands:        testutil.CmdRunOut("tofu version --json", `{"terraform_version": "0.16.0-alpha2"}`),
			expectedVersion: "0.16.0-alpha2",
		},
```
but there is already opentofu 1.8.6 
https://github.com/opentofu/opentofu/releases

abergmeier:
> @abergmeier thank you for effort, I haven't checked deeply but it seems pkg/skaffold/tofu/version_test.go expects version 0.16
> 
> ```
> 			commands:        testutil.CmdRunOut("tofu version --json", `{"terraform_version": "0.16.0-alpha2"}`),
> 			expectedVersion: "0.16.0-alpha2",
> 		},
> ```
> 
> but there is already opentofu 1.8.6 https://github.com/opentofu/opentofu/releases

Is there any reason why you think testing for a particular version is necessary?
My reasoning was that the test should be fine with handling an "obscure" case.
The code so far does not really care about which `tofu` version is installed.

Issue #9598, opened 2024-12-05
Can't sync files


lifeBalance:
### Expected behavior

Changing a typescript file should sync it in the running container

### Actual behavior

I get the error:
```
Syncing 1 files for ticket-app/auth:3d6686b1a49c42b910358112eca8d8d38b9c1cca5269f27d64c360bcc7b7db4b
INFO[1194] Copying files:map[auth/src/index2.ts:[/app/src/index2.ts]]toticket-app/auth:3d6686b1a49c42b910358112eca8d8d38b9c1cca5269f27d64c360bcc7b7db4b  subtask=-1 task=DevLoop
DEBU[1194] getting client config for kubeContext: docker-desktop  subtask=-1 task=DevLoop
WARN[1194] sync failed for artifact "ticket-app/auth:3d6686b1a49c42b910358112eca8d8d38b9c1cca5269f27d64c360bcc7b7db4b"  subtask=-1 task=DevLoop
WARN[1194] Skipping deploy due to sync error:copying files: didn't sync any files: sync failed for artifact "ticket-app/auth:3d6686b1a49c42b910358112eca8d8d38b9c1cca5269f27d64c360bcc7b7db4b"  subtask=-1 task=DevLoop
Watching for changes...
```

### Information

- Skaffold version: v2.13.2
- Operating system: macos Sonoma
- Installed via: Homebrew
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v4beta11
kind: Config
metadata:
  name: ticket-app
build:
  local:
    push: false
  artifacts:
    - image: ticket-app/auth
      context: auth
      docker:
        dockerfile: Dockerfile
      sync:
        manual:
          - src: "src/**/*.ts"
            dest: .
manifests:
  rawYaml:
    - https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.12.0-beta.0/deploy/static/provider/cloud/deploy.yaml
    - infra/k8s/*
```

### Steps to reproduce the behavior

1. a clonable repository with the sample skaffold project: https://github.com/lifeBalance/micro
2. `cd ticket-app && skaffold dev -v debug`
3. I save changes to some file, open a shell into the container, and the changes are not there.

idsulik:
@lifeBalance try to update the skaffold to the latest version

lifeBalance:
@idsulik isn't `v2.13.2` the latest version?

idsulik:
sorry, you're right.

[...]

PR #9599, opened 2024-12-06
fix(sync): add validation for namespaces and handle empty pods

idsulik:
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/9598

**Description**
- Return an error if no namespaces are provided for syncing.
- Log a warning if no running pods are found in a namespace during the sync process.


pkg/skaffold/sync/sync_test.go
```diff
@@ -921,58 +921,64 @@ var pod = &v1.Pod{
 }
 
 func TestPerform(t *testing.T) {
-	tests := []struct {
-		description string
```
menahyouyeah, line 925: nit: I would revert and keep description as a field instead of making this a map. It is easier to read the tests that way and is the general best practice.

pkg/skaffold/sync/sync_test.go
```diff
@@ -921,58 +921,64 @@ var pod = &v1.Pod{
...
-			description: "no copy",
-			image:       "gcr.io/different-pod:123",
-			files:       syncMap{"test.go": {"/test.go"}},
-			pod:         pod,
-			cmdFn:       fakeCmd,
-			shouldErr:   true,
+		"pod not running": {
+			image:     "gcr.io/k8s-skaffold:123",
+			files:     syncMap{"test.go": {"/test.go"}},
+			pod:       nil,
```
menahyouyeah, line 960: in the code you're only logging a warning, but here the test should fail?

pkg/skaffold/sync/sync.go
```diff
@@ -328,6 +328,10 @@ func Perform(ctx context.Context, image string, files syncMap, cmdFn func(contex
 		return nil
 	}
 
+	if len(namespaces) == 0 {
```
menahyouyeah, line 331: is there any way to add a test for this? not sure if that's possible with the current unit test setup

menahyouyeah, line 331: were you able to reproduce the issue and confirm this fixes it? In the bug they filed, I didn't see anything in the logs to indicate what the issue was

pkg/skaffold/sync/sync_test.go
idsulik: it fails because no files were synced

pkg/skaffold/sync/sync.go
idsulik: it's just an assumption, because according to the log if something went wrong, it returns error, but it doesn't return any specific error if there is no namespaces OR nor running pod was found

idsulik: I think these logs should be merged regardless of the issue, because in these cases(empty namespaces or no running pods) there is no way to know what went wrong

pkg/skaffold/sync/sync_test.go
idsulik: done

pkg/skaffold/sync/sync.go
idsulik: done

pkg/skaffold/sync/sync_test.go
```diff
@@ -973,6 +981,10 @@ func TestPerform(t *testing.T) {
 
 			t.Override(&util.DefaultExecCommand, cmdRecord)
 			t.Override(&client.Client, func(string) (kubernetes.Interface, error) {
+				if test.pod == nil {
+					return fake.NewSimpleClientset(), nil
+				}
+
 				return fake.NewSimpleClientset(test.pod), test.clientErr
 			})
```
menahyouyeah, line 990: there are no namespaces being given below in Perform the arg is `[]string{""} - why does the first test case still work "no error"? Is there a default namespace being used? 

instead of adding a new test (i.e. TestPerform_WithoutNamespaces) could you just add another test case here? First add a new value to the test struct "namespaces", add to existing tests and then pass that along when calling Perform. And in that test case you'd have an empty namespace.

idsulik: because it's `[]string{""}` slice of 1 item and the item is an empty string, this is why the `len` returns a non-zero value and it works.

idsulik: I was thinking about adding one more test case to the old test, but it'll complicate the test, so decided to create a separate test

pkg/skaffold/sync/sync.go
menahyouyeah: So I would delete this line as it'll never be hit. Specifically here: https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/deploy/util/namespaces.go#L35C4-L35C26 if the namespace is empty, it gets the kube context's namespace. If nothing was set, the default namespace is used.

pkg/skaffold/sync/sync_test.go
menahyouyeah: See above comment it turns out we don't need to check for empty namespaces and can remove the separate test. Appreciate the explanation though.

pkg/skaffold/sync/sync.go
idsulik: done

menahyouyeah (approved):

PR #9600, opened 2024-12-07
fix: Helm deploy was not working with variable templatinging chart path

travishein:
I believe the chartPath property is supposed to be a templatable field.

In my use I have helm charts in a separate git repo from the application where the skaffold.yaml is. We defined a skaffold.env.template file, that we add to git and then on checkout we copy this to skaffold.env and modify as needed, so the path to the location to the other repo with a helm chart is a variable. and this skaffold.env file is added to the .gitignore

We noticed that we were getting
listing files: issue walking releases: lstat {{.HELM_REPO}}/helm/app/chart/path: no such file or directory

This PR fixes this issue.

Fixes: #nnn 

**Description**

### DIFF
--- a/pkg/skaffold/deploy/helm/helm.go
+++ b/pkg/skaffold/deploy/helm/helm.go
@@ -398,11 +398,11 @@ func (h *Deployer) Dependencies() ([]string, error) {
 			return true, nil
 		}
 
-		expanded_path, e := util.ExpandEnvTemplateOrFail(release.ChartPath, nil)
+		expandedPath, e := util.ExpandEnvTemplateOrFail(release.ChartPath, nil)
 		if e != nil {
 			return deps, helm.UserErr("issue expanding variable", e)
 		}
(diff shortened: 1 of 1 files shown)


pkg/skaffold/deploy/helm/helm.go
```diff
@@ -398,7 +398,11 @@ func (h *Deployer) Dependencies() ([]string, error) {
 			return true, nil
 		}
 
-		if err := walk.From(release.ChartPath).When(isDep).AppendPaths(&deps); err != nil {
+		expanded_path, e := util.ExpandEnvTemplateOrFail(release.ChartPath, nil)
```
alphanota, line 401: Please fix this lint issue, `expanded_path` should be `expandedPath`

alphanota (changes requested):

pkg/skaffold/deploy/helm/helm.go
travishein: appologies. I did not notice the lint check before.
Fixed.

Issue #9197, opened 2023-11-29
Support additional docker buildkit CLI options


dsiebel:
Currently BuildKit / buildx support is Skaffold seems to be rudimentary:
When `useBuildKit:true` it simply sets env `DOCKER_BUILDKIT=1`, effectively enabling BuildKit
but not allowing to specify any of the command line options that `docker buildx build` support, for example.
This means that `--cache-from` with, for example `type=registry` can not be used without a custom build command like this one:

```sh
apiVersion: 'skaffold/v3'
kind: 'Config'
metadata:
  name: 'buildx-inline-caching'
build:
  artifacts:
    - image: us-docker.pkg.dev/skaffold-examples/buildx-inline-caching
      context: docker
      custom:
        buildCommand: |
          echo
          echo "Building image '${IMAGE}' for platforms '${PLATFORMS}' from context '${BUILD_CONTEXT}'"
          echo "Image is expected to exist remotely: ${PUSH_IMAGE}"
          echo
          args=()
          if [[ "${PUSH_IMAGE}" == true ]]; then
            args+=("--push")
          elif [[ "${PLATFORMS}" != *","* ]]; then
            args+=("--load")
          fi
          docker buildx build \
            --file "${BUILD_CONTEXT}/Dockerfile" \
            --tag "${IMAGE}" \
            --cache-to "type=inline" \
            --cache-from "type=registry,ref=us-docker.pkg.dev/skaffold-examples/buildx-inline-caching:cache" \
            --platform "${PLATFORMS}" \
            "${args[@]}" \
            "${BUILD_CONTEXT}"
```

This is what we currently use.
Skaffold is being invoked with `--tag cache` after a successful build, which re-tags and pushes the image with the `:cache` tag so it can be used as a cache source in subsequent builds.
However this approach doesn't really scale well, having to implement this custom script across potentially hundreds of `skaffold.yaml` files and repositories.
It would be way nicer if the `buildx build` arguments would be supported by Skaffold natively.

As a start, I imagine using the combination of the `useBuildKit` and `cacheFrom` docker builder config options to construct the `--cache-from` option for example and adding a `cacheTo` config option for the docker builder to do the same for the `--cache-to` option. 
This way one could leverage the advanced caching mechanisms of BuildKit _through_ Skaffold.
Replacing the call to `docker build` with `docker buildx build` should already be transparent once you installed buildx (it is on my machine), but I need to double check.

Let me know what you think!
[...]

Issue #2110, opened 2019-05-10
feature request: more control over the docker build context


majelbstoat:
I have a monorepo with shared code. Because docker doesn't follow symlinks and never will, I have to copy that shared code into the right place before building the image for each service. There's currently no facility for doing that with skaffold. I'm using VSCode's Cloud Code extension, so running a wrapper script isn't straightforward, but in any case, I'd like the build to be fully described in the definition, not be dependent on some external glue.

#1441 describes one approach to solving this, which would be great, but I don't need the full flexibility of a custom script.

What I'd like to see is the ability to control which parts of the context are put into the tarball that is sent to Docker. Being able to specify paths, and mappings would be sufficient for my use-case. For example:

```yaml
context:
  root: .
  paths:
    - mapRef:
        from: services/mercury/src
        to: .
    - mapRef:
        from: shared/nodejs/src
        to: ./lib
```

This would create a tarball that had files from mercury at the root, and files from the shared source at `./lib`. Ideally, it would do this without actually copying the shared files into the mercury service in the local file system.

The absence of the path key would cause behaviour as currently is, i.e. everything beneath the context.

corneliusweig:
Would it be an option to create hard links instead of soft links?

majelbstoat:
Hard links cannot be created for directories, so unfortunately not.

corneliusweig:
> Yeah, hard links cannot be created for directories, so unfortunately not.

So recursively creating hard links is also not an option?
```
cp -rl from to
```

majelbstoat:
The code is shared to multiple different locations in the monorepo, and the benefit of symlinks is whenever a new shared library is added, it is immediately available to all the services. Running a script each time a new file is added complicates the developer workflow. I already have a hacky script to glue bits and pieces together, replacing it with another isn't an improvement :)

balopat:
I think it would be interesting to provide support for a flexible Docker context creation.
Skaffold could assemble the tar file based on the context spec and pipe it to `docker build`. 
It is big enough though that I think a good design proposal would be beneficial before someone jumps in to implement it.

tstromberg:
Thank you for the idea!

I'm closing this issue as it's been open a while, and it is not clear if it's still an open issue. No one has recently stated an interest in addressing this, but if you feel strongly about it, please feel free to add a comment or send us a PR!

majelbstoat:
[...]

PR #9648, opened 2025-01-11
feat: transparent buildx support for remote buildkit (#6732)

### LINKED ISSUE #6732: Support building securely against remote buildkitd

reingart:
Fixes: #6732 #9197 
**Related**: #8172 (inspired by by a prior [ebekebe fork](https://github.com/ebekebe/skaffold/commit/1c1fdeb18f4d2847e65e283fba498a14745039af), with a more general approach)

**Description**

`docker buildx` is an enhanced builder using BuildKit that can replace traditional `docker build` command, with almost the same syntax (transparently).

This adds several distributed features and advanced capabilities:
* local & remote build-kit builders, either running standalone, docker or Kubernetes clusters
* improved cache support (registry destination, pushing full metadatada and multi-stage layers)
* improved multi-platform image building support (eg. x86_64 and arm64 combined, with emulation or cross-compilation)

This features are very useful for corporate CI/CD, for example when using GitLab, where the use case requires:
* using a ephemeral remote buildkit instance (rootless)
* using a remote docker registry for cache
* using a different cache destination tag

The remote BuildKit support is useful in cases with a privileged docker daemon is not possible or desirable due security policies (eg. untrusted images or review pipelines).
Daemon-less mode is also useful to offload container building from developers notebooks, sharing and reusing remote caches more effectively.

The `cache-tag` is useful to point to latest or a generic cache tag, instead of the generated one for this build (via `tagPolicy` that would be useless in most distributed cases, as it can be invalidated by minor changes, specially if using `inputDigest`).
Having it as an option could allow different branches to have different cache destinations (production vs development cache)

**User facing changes**

To use BuildKit transparently, BuildX should be installed as default builder (this creates an alias in docker config) and create a remote instance:
```
docker buildx install
docker buildx create --name remote --driver remote tcp://buildkitd:2375
```
Then Skaffold should be configured to detect buildx and set a default cache tag:
```
skaffold config set -g buildx-builder remote
skaffold config set -g cache-tag cache
```

Optionally, in skaffold.yaml the cache source and destination can be configured with `cacheFrom` and `cacheTo` with the image name:
```
build:
  artifacts:
  - image: my-app-image
    context: my-app-image
    docker:
      dockerfile: Dockerfile
      cacheFrom:
      - "my-app-image"
  tagPolicy:
    inputDigest: {}
  local:
    useBuildkit: true
    useDockerCLI: true
    tryImportMissing: true
```
* If no tag is specified for cache, the configured `cache-tag` will be used (in this example `my-app-image:cache`)
* If cache to destination is not specified, the cache-from image name and tag will be adjusted, adding `type=registry,mode=max` (only if push images is enabled)

This cache behavior is intended to provide transparent user experience, with similar functionality compared to traditional local docker builds, without additional boilerplate or different configuration / patches for a CI workflow.

Finally, the user can run `skaffold dev` or `skaffold build` as usual.

**Follow-up Work**

Future work: #2110 to include more BuildKit advanced features support like multiple build contexts, for an improved mono-repo experience for large code-bases.

TODO:

- [x] Design document
- [x] Unit Tests
- [x] Integration tests

**Changes summary:**

* adds a new global config `buildx-builder` to enable buildx detection and support for different builders
* implements logic to detect buildx is the default builder (via docker config alias)
* removes dependency on docker daemon if using BuildX (still supported to load images if using minikube or similar)
* adds a new global config `cache-tag` to override default cache tagging (instead of generated tag)
* adds a `cacheTo` to sakffold schema for the cache destination (optional, default is to use new `cacheFrom` + cache tag)
* adds `--cache-to` support in Docker CLI build if using BuildX CLI 
* adds multiplatform images building support for buildkit under buildx

All changes are backward compatible, so if the developer is not using buildx, traditional docker builder will be run.

Signed-off-by: reingart@gmail.com


reingart:
Hi team! ping :-) 

I've updated the new schema changes, following the new version 23 generated by @menahyouyeah in https://github.com/GoogleContainerTools/skaffold/pull/9741, so now the schema test pass.

In my personal fork all checks are passing: https://github.com/reingart/skaffold/pull/1/checks including:
* New **unit test cases** fully covering buildx local and remote scenarios, including detection and new cache-to.
* New **integration test cases** fully covering basic buildx image building, and advanced ones for amd64/arm64 architectures and multi-platform images.
Integration tests runs with a buildx buildkit builder and a registry deployed in containers, so no additional changes to testing infra is required (can run locally, do not uses GCP or any other special cloud infra)

Also I'd written a tentative design proposal: [buildx-transparent-support.md](https://github.com/GoogleContainerTools/skaffold/blob/c5eb4e50b91563c357eb08358efa2779f895a2f1/docs-v2/design_proposals/buildx-transparent-support.md), trying to address request in related PR like https://github.com/GoogleContainerTools/skaffold/pull/8172#issuecomment-1332620175 by @gsquared94 regarding multiple buildx instances, actionable errors and multi-platform images.

 Let me know if you need anything else to evaluate the PR, any recommendation is welcomed

mikegore1000:
Hi,

I'm currently looking to use buildx with Skaffold. One of the use cases I have is being able to authenticate with a package feed during the build process.

I have spiked out an approach using custom build scripts so I can use buildx commands. One of the features I've used is (Docker secret mounts)[https://docs.docker.com/build/building/secrets/#secret-mounts] as this allows me to pass in the credentials in a way that doesn't invalidate the Docker cache if the secret changes.

Our build process passes a unique secret per run, so without this support we effectively have to rebuild the image from scratch every single time which adds minutes onto our builds.

Would it be possible to add support for build secrets as part of this PR?

docs-v2/content/en/schemas/v4beta13.json
```diff
@@ -1766,6 +1766,18 @@
             "[\"golang:1.10.1-alpine3.7\", \"alpine:3.7\"]"
           ]
         },
+        "cacheTo": {
+          "items": {
+            "type": "string"
+          },
+          "type": "array",
+          "description": "the Docker images used as cache destination. If omitted, cacheFrom is used with max mode to export all layers.",
```
snickell, line 1774: Not terribly important, since most people playing with this (today at least) will already know what buildx `--cache-to` accepts, but I wouldn't mind a link to the docker docs here.

docs-v2/design_proposals/buildx-transparent-support.md
```diff
@@ -0,0 +1,284 @@
...
+
+`docker buildx build` can be configured to execute against a [remote buildkit instance](https://docs.docker.com/build/builders/drivers/remote/).
+No docker daemon is necessary for this to work, but also that is supported by default using the [docker driver](https://docs.docker.com/build/builders/drivers/docker/).
+Additionally, [docker container driver](https://docs.docker.com/build/builders/drivers/docker-container)
+or [kubernetes driver](https://docs.docker.com/build/builders/drivers/kubernetes/) are available too for advanced use cases.
+
+Since `docker build` and `docker buildx build` essentially share the same options, no major modifications are needed to Skaffold for backward compatibility.
+
+This proposal implements the logic to detect if buildx is the default builder (looking for an alias in the docker config).
+To [set buildx as the default builder](https://github.com/docker/buildx?tab=readme-ov-file#set-buildx-as-the-default-builder), the command `docker buildx install` should be used.
```
snickell, line 80: Not a must-have by any means, but it'd be nice to have an explicit way at the skaffold.yml level, like `useBuildkit: true`, to specify that buildx *should* be used, and therefore error out if its not available.

docs-v2/design_proposals/buildx-transparent-support.md
```diff
@@ -0,0 +1,284 @@
...
+
+Then, additional buildx features are availables, like multi-platform support and different cache destinations.
+
+Cache adjustment is extended via the the `cache-tag`, useful to point to latest or a generic cache tag, instead of the generated one for this build
+(via `tagPolicy` that would be useless in most distributed cases, as it can be invalidated by minor changes, specially if using `inputDigest`).
```
snickell, line 85: This is a great feature, appreciate the thoughtfulness here.

docs-v2/design_proposals/buildx-transparent-support.md
```diff
@@ -0,0 +1,284 @@
...
+
+### User experience
+
+To use BuildKit transparently, BuildX should be installed as default builder (this creates an alias in docker config):
+
+```
+docker buildx install
+```
+
+Then Skaffold should be configured to detect buildx (default builder) and set a generic cache tag:
```
snickell, line 97: I believe, per below in the doc, these are optional. If my understanding is correct, it'd be great if this doc lists these as "can be" rather than "should be".

pkg/skaffold/build/local/local.go
```diff
@@ -123,8 +126,8 @@ func (b *Builder) buildArtifact(ctx context.Context, out io.Writer, a *latest.Ar
 }
 
 func (b *Builder) runBuildForArtifact(ctx context.Context, out io.Writer, a *latest.Artifact, tag string, platforms platform.Matcher) (string, error) {
-	if !b.pushImages {
-		// All of the builders will rely on a local Docker:
+	if !b.buildx && !b.pushImages {
+		// Most builders will rely on a local Docker (except when using a remote buildkit via buildx):
```
snickell, line 130: How does this relate to using a local buildkit? Will these have to basically dump to local docker?

pkg/skaffold/build/builder_mux.go
```diff
@@ -106,7 +108,8 @@ func (b *BuilderMux) Build(ctx context.Context, out io.Writer, tags tag.ImageTag
 		}
 		var built string
 
-		if platforms.IsMultiPlatform() && !SupportsMultiPlatformBuild(*artifact) {
+		// buildx creates multiplatform images via buildkit directly
+		if platforms.IsMultiPlatform() && !SupportsMultiPlatformBuild(*artifact) && !b.buildx {
 			built, err = CreateMultiPlatformImage(ctx, out, artifact, tag, platforms, artifactBuilder)
```
snickell, line 113: I appreciate the simplicity of this approach, and this is really a curiosity more than a comment. Is there a reason not to instead extend CreateMultiPlatformImage to run individual buildx instances for each platform, and then stitch them into a single manfiest using the built-in skaffold stitching?

In particular, in my use case, I would want to run (via GH actions) separate /native/ builders (our build is too slow on arm to want to use qemu), define them as multiple remotes, and then stitch them in the main skaffold build command.

snickell (approved):
This PR is a huge leap forward, we (code.org) are evaluating skaffold, and our primary reservation in selecting skaffold is the complexity of maintaining our own custom builder scripts to use buildx. I hope the skaffold team has the bandwidth to work with this PR all the way to merge.

My primary question, which may best be addressed in a follow-on PR in order to build momentum: how can we use multiple buildx instances in order to build with native builders, and then stitch them into a multi-platform image?

reingart:
> I have spiked out an approach using custom build scripts so I can use buildx commands. One of the features I've used is [Docker secret mounts](https://docs.docker.com/build/building/secrets/#secret-mounts) as this allows me to pass in the credentials in a way that doesn't invalidate the Docker cache if a secret changes.

Yes, I use that too, including SSH key forwarding. 
Beside cache invalidation, it also helps to avoid leaking secrets in the built image.

If you can share the custom script and skaffold config I could take a look and tell you how to adapt to use this PR.

> Our build process passes a unique secret per run to authenticate with the package feed, so without this support we effectively have to rebuild the image from scratch every single time which adds minutes onto our builds.
> 
> Would it be possible to add support for build secrets as part of this PR?

AFAIK build secrets are already supported by BuildKit in skaffold (natively), and this PR supports them even with remote builders (either docker secrets and ssh mounts).
Other BuildKit features also work remotely with this PR, like cache mounts, multi-stage builds and cache import/export (some will be limited by the type o builder and shared storage, for example only a registry can be used as cache destination if the builder run in containers IIRC).

I will try to implement a more advanced example with these features (secrets and ssh mounts + caching), in a future stage once the code stabilizes (proposed in the design document).

reingart:
Hi @mikegore1000 , thanks for the comments and review!

> My primary question, which may best be addressed in a follow-on PR in order to build momentum: how can we use multiple buildx instances in order to build with native builders, and then stitch them into a multi-platform image?

The current PR already support multi-platforms builds without needing to stitch different images (see "build multiplatform images with buildx" integration test for both amd64 & arm64).
In fact, the individual build pipeline and stitching was modified to not be used in buildx, and the integration test is using the emulation.

I think that this could even work for native builders, according the [buildx documentation](https://github.com/docker/buildx#building-multi-platform-images) this could be implemented just adding the nodes:
> Using multiple native nodes provide better support for more complicated cases that are not handled by QEMU and generally have better performance. You can add additional nodes to the builder instance using the --append flag.
```
$ docker buildx create --use --name mybuild node-amd64
$ docker buildx create --append --name mybuild node-arm64
$ docker buildx build --platform linux/amd64,linux/arm64 .
```
I didn't tested this scenario, but it should work as it is just a configuration to buildx outside skaffold.

docs-v2/content/en/schemas/v4beta13.json
reingart: Ok, will add a link to the docs: https://docs.docker.com/reference/cli/docker/buildx/build/#cache-to  (I see ko and other builders have similar links)

docs-v2/design_proposals/buildx-transparent-support.md
reingart: A more actionable error would be great if buildx is not installed properly (that is included in the design doc for further phases), but I explicitly would like to avoid a configuration in skaffold.yaml because:
* users that doesn't have buildx installed would still be able to build
* no modification of skaffold.yaml is needed to switch between CI and local envs

This way, buildx support is transparent and the user can decide to use it or not, just setting the config parameter with `skaffold config set buildx-builder ...`

Yes, some advanced features would not work (like exporting the cache), but this would not be critical for local development (and wont work either if proper registry push credentials are not given).

pkg/skaffold/build/local/local.go
reingart: If there is a local docker daemon, it should have a "default" buildkit builder.
If I understand correctly, it is currently being used when the skaffold.yaml has `useBuildkit: true`.
It can be used with buildx too, executing `skaffold config set buildx-builder default`.
Let me know if I understood ok the question or more info is needed.

pkg/skaffold/build/builder_mux.go
reingart: Using the native buildkit multi-platform can offload the build parallelization and image stitching from skaffold, and it would support more strategies: 
https://github.com/docker/buildx?tab=readme-ov-file#building-multi-platform-images

> 1. Using the QEMU emulation support in the kernel
> 2. Building on multiple native nodes using the same builder instance
> 3. Using a stage in Dockerfile to cross-compile to different architectures

I addressed more about using native nodes in my other reply, I think it should be possible and match your use case without needing separate builds or stitching: https://github.com/GoogleContainerTools/skaffold/pull/9648#issuecomment-2810512129

docs-v2/design_proposals/buildx-transparent-support.md
snickell: That thinking makes sense to me, I do like your idea of providing graceful "fallback". That said, iIn the cases of very big projects, if cache-from in particular isn't working, I'd personally prefer my "app developers" to get an error that they _really_ need to configure buildx, at the very least a noisy warning. Many of them will silently not configure buildx, and will later complain that builds take 30 minutes, when if cache-from was working, it'd be like 5 minutes.

`skaffold config` setting the builder system-wide rather than per-project feels like it has small issues in some cases: for example, in my case I will have one project where I need multi-platform, and another where I don't. But I agree that hardcoding the builder names in skaffold.yaml would be much worse.

pkg/skaffold/build/builder_mux.go
snickell: That makes great sense to me, thanks for explaining, and its really cool that buildx has support for "fanning out" to native builders built-in. Somehow I had never noticed that despite using buildx for a couple years.

snickell:
Thank you @reingart for linking me to the use of "native build nodes" from buildx. Really cool. I'll do a test of using this from within skaffold (using this branch) via gh actions.

In case it wasn't clear to everyone: I really support this PR, and for our use case, its a very good fit. I would love to see this PR merged as-is.

mikegore1000:
Hi @reingart,

I'm pretty new to Skaffold so must have missed the built-in secret support. That plus your PR means we can do everything without any build scripts and it's clear how I'd go about setting that up.

Thanks for the help. Really looking forward to this getting merged in!

rajatvig:
Really looking forward to having this change merged in.

rajatvig:
@reingart Are you still working on this PR?

reingart:
@rajatvig yes, mostly testing it in development.
I will try to set-up a fork to provide a compiled release to ease reviews/QA and keep in sync (and to include probably other minor improvements).
Let me know if you have any question (sorry for the late reply, I was on vacation)

dsiebel:
@reingart Are you, by any chance still working on this? What are the chances of this landing in `main` at some point in the near future?
We're experiencing more and more issues with the current BuildKit implementation in Skaffold. Especially since Docker v29 (default attestations on multi-platform builds)..

tschoonj:
Any progress here? It would be great to have better `buildx` support in skaffold.

reingart:
Hi @tschoonj we have a fork with these buildkit and other changes / improvements / bugfixes in https://github.com/Onapsis/skaffold

It has been updated with the latest skaffold changes, and it is being used internally. 
The idea is to maintain it synced and keep testing, to open additional PR here if is needed.

@dsiebel see this https://github.com/GoogleContainerTools/skaffold/commit/daf2b09e405f02a14f17b319e8f46198d3895cb8 for an attestation workaround; let me know if you have more details (like if you´re using remote builders), I could investigate.

PR #9671, opened 2025-01-17
ci: requiring two org members to approve for PR merge; preventing in org sock puppet accounts

Darien-Lin:
**Description**

This adds multi-approver workflow following this: https://github.com/abcxyz/pkg?tab=readme-ov-file#multi-approversyml.
This prevents in-org sock puppet accounts.

### DIFF
--- a/.github/workflows/members.json
+++ b/.github/workflows/members.json
@@ -1,0 +1,233 @@
+[
+  {
+    "login": "aaron-prindle"
+  },
+  {
+    "login": "akulkapoor-google"
+  },
+  {
+    "login": "alicejli"
+  },
+  {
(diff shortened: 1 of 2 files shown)


.github/workflows/members.json
```diff
@@ -0,0 +1,233 @@
+[
```
plumpy, line 1: Wait, this is the approved workflow? We're supposed to manually maintain this file?

Is there a way to just add `skaffold-team` to this file? I don't even know who most of these people are

plumpy: (And skaffold-team has the advantage of being automatically synced to our internal systems)

Issue #9681, opened 2025-01-24
`skaffold render --debug` suggested but missing


th0ger:
The `skaffold render` command suggests `--debug` to render invalid YAML, but the option is missing in the CLI, in `--help`, and latest [docs](https://skaffold.dev/docs/references/cli/#skaffold-render).

```bash
$ skaffold render
std out err: [blah blah] error converting YAML to JSON: yaml: [blah blah]
Use --debug flag to render out invalid YAML
)

$ skaffold render --debug
unknown flag: --debug
See 'skaffold render --help' for usage.

$ skaffold render --help | grep "\-\-debug"

$skaffold --debug render
unknown flag: --debug
See 'skaffold --help' for usage.

$ skaffold --help | grep "\-\-debug"

$ skaffold version
v2.12.0
```

I browsed for open issues.

idsulik:
@th0ger thank you for the issue, after this fix https://github.com/GoogleContainerTools/skaffold/pull/9682, you'll be able to specify [helm.flags.global](https://skaffold.dev/docs/references/yaml/?version=v4beta12#manifests-helm-flags-global) --debug and the flag will be passed to helm command, example output with invalid template and _debug_ flag:

<img width="988" alt="Image" src="https://github.com/user-attachments/assets/8500a758-1b6d-4f09-a4dd-4d4fb911411f" />

th0ger:
Fast simple fix, thanks.

katiexzhang:
Hi @th0ger , could you please point to where in the docs you see this: "The skaffold render command suggests --debug to render invalid YAML"?

idsulik:
@katiexzhang he has already pointed it out:
```
$ skaffold render
std out err: [blah blah] error converting YAML to JSON: yaml: [blah blah]
Use --debug flag to render out invalid YAML
```
and, he has already written that the output was from the helm command which was executed by the scaffold:
> I just found out that, the error is thrown by helm, not skaffold:

th0ger:
@idsulik (katiexzhang could be a bot...)

idsulik:
oh, it's interesting)

katiexzhang:
Ah my bad, I didn't see the second comment and was referring to our public docs, which does not support `skaffold render --debug` https://skaffold.dev/docs/references/cli/#skaffold-render.

Could you please update the issue title / description to reflect the actual issue?

th0ger:
> Could you please update the issue title / description to reflect the actual issue?

[...]

PR #9682, opened 2025-01-25
fix(rendered): Pass helm legacy deploy flags to render config

idsulik:
Fixes: #9681  

**Description**
The helm's flag 'global' says _additional flags passed on every command_ https://skaffold.dev/docs/references/yaml/?version=v4beta12#manifests-helm-flags-global, but skaffold doesn't pass while rendering, fixed it


katiexzhang (changes requested):
Could you add a test case covering this?

idsulik:
> Could you add a test case covering this?

refactored a bit and added test

katiexzhang (approved):

PR #9686, opened 2025-01-27
cloud-build.md replace container registry with artifact registry

ersin-ertan:
**Description**
Update product name and link. Container Registry is deprecated and the existing link redirects to Artifact Registry.


katiexzhang (changes requested):
Please follow the conventional commits format:
❌ The following linting errors found:

subject may not be empty
type may not be empty
for the following input:
"cloud-build.md replace container registry with artifact registry"

PR #9688, opened 2025-01-28
chore: Fix BuildContextCompressionLevel description, output the level

idsulik:
**Description**
1. Fixed `BuildContextCompressionLevel` description
2. Add logging for the compression level - `using gzip compression level {level}`

### DIFF
--- a/docs-v2/content/en/schemas/v4beta12.json
+++ b/docs-v2/content/en/schemas/v4beta12.json
@@ -2672,8 +2672,9 @@
         },
         "buildContextCompressionLevel": {
           "type": "integer",
-          "description": "gzip compression level(0-9) for the build context. Defaults to 1 (BestSpeed). 0: NoCompression. 1: BestSpeed. 9: BestCompression. -1: DefaultCompression. -2: HuffmanOnly.",
-          "x-intellij-html-description": "gzip compression level(0-9) for the build context. Defaults to 1 (BestSpeed). 0: NoCompression. 1: BestSpeed. 9: BestCompression. -1: DefaultCompression. -2: HuffmanOnly."
+          "description": "gzip compression level(0-9) for the build context. 0: NoCompression. 1: BestSpeed. 9: BestCompression. -1: DefaultCompression. -2: HuffmanOnly.",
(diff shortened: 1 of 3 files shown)


katiexzhang:
Thanks for the change @idsulik, although since there is no explicit request for this, I don't think it's necessary.

idsulik:
@katiexzhang , what do you mean? the description isn't full, how does the user should know what value they can use and what does it mean?

idsulik:
> gzip compression level for the build context. 

https://skaffold.dev/docs/references/yaml/#build-artifacts-kaniko-buildContextCompressionLevel

who knows what compression levels gzip has? I didn't even know until I made this feature.
Originally, I wrote all the levels in the description just for this purpose, so that the documentation would show it and the developers would know what value they could substitute, but because of a bug in the documentation generation, not the full description was published, and I didn't notice it.

katiexzhang (changes requested):
Where are you getting these descriptions from, specifically the -2? If possible, could you use the description from public docs, like https://docs.python.org/3/library/gzip.html.

Also I think you accidentally removed the default value.

idsulik:
1. from here https://pkg.go.dev/compress/flate#pkg-constants. 
2. the changes contain default value as well

idsulik:
@plumpy hi! Could you please check the PR? @katiexzhang is doing a weird code review, [saying](https://github.com/GoogleContainerTools/skaffold/pull/9688#issuecomment-2620158722) full documentation is unnecessary without explicitly asking for it and [asking](https://github.com/GoogleContainerTools/skaffold/pull/9688#pullrequestreview-2581530077) to put a link to gzip's page while the constants from the go' package

docs-v2/content/en/schemas/v4beta12.json
```diff
@@ -2672,9 +2672,8 @@
         },
         "buildContextCompressionLevel": {
           "type": "integer",
-          "description": "gzip compression level for the build context.",
-          "x-intellij-html-description": "gzip compression level for the build context.",
-          "default": "1"
```
katiexzhang, line 2677: This line was removed: "default": "1"

idsulik: oh, sorry, haven't noticed it. pushed fix

PR #9689, opened 2025-01-28
fix(helm): Add expand env template for dependsOn, fix concurrent installation

idsulik:
**Description**
1. Fixed the releases installation order to keep the original order for each level, it needs at least to keep the old behavior, because the v2.14.0 changed the order. 
2. Fixed an issue with the [dependsOn](https://skaffold.dev/docs/references/yaml/#manifests-helm-releases-dependsOn), we don't expand the chart's name. Now we use env in the dependsOn chart's name, example:
```
    deploy:
      helm:
        concurrency: 0 #no-limit
        releases:
          - name: my-service-{{.MYUSER}}
            remoteChart: repo/php-nginx
            upgradeOnChange: true
            version: ~4.30.0
            setValues:
              image.pullPolicy: IfNotPresent
            wait: true
            dependsOn:
              - my-cron-{{.MYUSER}}
          - name: my-cron-{{.MYUSER}}
            remoteChart: repo/php-nginx
            upgradeOnChange: true
            version: ~4.30.0
            setValues:
              image.pullPolicy: IfNotPresent
            wait: true
```

### DIFF
--- a/pkg/skaffold/deploy/helm/helm.go
+++ b/pkg/skaffold/deploy/helm/helm.go
@@ -285,40 +285,42 @@ func (h *Deployer) Deploy(ctx context.Context, out io.Writer, builds []graph.Art
 	// Group releases by their dependency level to deploy them in the correct order.
 	levelGroups := groupReleasesByLevel(deploymentOrder, dependencyGraph)
 
-	g, levelCtx := errgroup.WithContext(ctx)
-
+	var concurrency int
 	if h.Concurrency == nil || *h.Concurrency == 1 {
-		g.SetLimit(1)
+		concurrency = 1
 		olog.Entry(ctx).Infof("Installing %d releases sequentially", len(h.Releases))
 	} else {
-		g.SetLimit(*h.Concurrency)
+		if *h.Concurrency == 0 {
+			concurrency = -1
+		} else {
+			concurrency = *h.Concurrency
+		}
 		olog.Entry(ctx).Infof("Installing %d releases concurrently", len(h.Releases))
 	}
 
 	releaseNameToRelease := make(map[string]latest.HelmRelease)
 	for _, r := range h.Releases {
-		releaseName, err := util.ExpandEnvTemplateOrFail(r.Name, nil)
-		if err != nil {
-			return fmt.Errorf("cannot parse the release name template: %w", err)
-		}
-		releaseNameToRelease[releaseName] = r
+		releaseNameToRelease[r.Name] = r
 	}
 
 	// Process each level in order
 	for level, releases := range levelGroups {
 		if len(levelGroups) > 1 {
-			olog.Entry(ctx).Infof("Installing level %d/%d releases (%d releases)", level, len(levelGroups), len(releases))
+			olog.Entry(ctx).Infof("Installing level %d/%d releases (%d releases)", level+1, len(levelGroups), len(releases))
 		} else {
 			olog.Entry(ctx).Infof("Installing releases (%d releases)", len(releases))
 		}
+
+		g, levelCtx := errgroup.WithContext(ctx)
+		g.SetLimit(concurrency)
 
 		// sort releases in the same level, this is merely to ensure that the series of helm commands are in order for
 		// consistency in unit testing.
 		sort.Strings(releases)
 
 		// Deploy releases in current level
-		for _, releaseName := range releases {
-			release := releaseNameToRelease[releaseName]
+		for _, name := range releases {
+			release := releaseNameToRelease[name]
 
 			g.Go(func() error {
 				chartVersion, err := util.ExpandEnvTemplateOrFail(release.Version, nil)
@@ -334,6 +336,11 @@ func (h *Deployer) Deploy(ctx context.Context, out io.Writer, builds []graph.Art
 				release.ChartPath, err = util.ExpandEnvTemplateOrFail(release.ChartPath, nil)
 				if err != nil {
 					return helm.UserErr(fmt.Sprintf("cannot expand chart path %q", release.ChartPath), err)
+				}
+
+				releaseName, err := util.ExpandEnvTemplateOrFail(release.Name, nil)
+				if err != nil {
+					return helm.UserErr(fmt.Sprintf("cannot expand release name %q", release.Name), err)
 				}
 
 				m, results, err := h.deployRelease(levelCtx, out, releaseName, release, builds, h.bV, chartVersion, repo)
--- a/pkg/skaffold/deploy/helm/helm_test.go
+++ b/pkg/skaffold/deploy/helm/helm_test.go
@@ -66,6 +66,21 @@ var testDeployConfig = latest.LegacyHelmDeploy{
 	}},
 }
 
+var testDeployWithDependsOnConfig = latest.LegacyHelmDeploy{
+	Releases: []latest.HelmRelease{{
+		Name:      "skaffold-helm-{{.FOO}}",
+		ChartPath: "examples/test",
+		Overrides: schemautil.HelmOverrides{Values: map[string]interface{}{"foo": "bar"}},
+		SetValues: map[string]string{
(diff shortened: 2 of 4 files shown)


idsulik:
Pushed one more fix, if the installation happens level by level(in case of dependsOn), then the `errgroup` cancels the `levelCtx` after the first iteration, moved it inside the loop so it'll be created for each iteration

idsulik:
@alphanota  could you please review the PR and consider releasing a patch version if appropriate

idsulik:
@alphanota hi! Could you please check the PR and publish the patch release? the v2.14.0 has some issues which I fixed in this PR. Tested it locally with/without dependencies, as I see it works correctly

idsulik:
@plumpy hi! check pls the PR, it's very important because the v2.14.0 breaks the release order and it can lead to unexpected behavior

alphanota:
@idsulik Can you explain what you mean by "keep the original order for each level, " as far as I can tell, internally dependencies that were on the same level were stored on a hashmap so by definition they had no order.

idsulik:
@alphanota check the tests, the releases are split level by level , if there is no "dependsOn" in the config file, then all releases will be at the 0 level and if we have dependsOn, then there will be more levels. inside each level, the releases keep the original sorting(as they appear in the config file).  the important part is to keep the order if there is no "dependsOn", otherwise the installation won't be stable(in our case with the new skaffold deploy failed because we had specific order in the config file, but the new version doesn't keep the order from the config file)

pkg/skaffold/deploy/helm/dependencygraph.go
```diff
@@ -0,0 +1,198 @@
...
+import (
+	"fmt"
+
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/schema/latest"
+)
+
+// DependencyGraph represents a graph of helm release dependencies
+type DependencyGraph struct {
+	graph           map[string][]string
+	releases        []latest.HelmRelease
```
idsulik, line 28: @alphanota here, we save releases as they're passed into the constructor

pkg/skaffold/deploy/helm/dependencygraph.go
```diff
@@ -0,0 +1,198 @@
...
+			if err := visit(dep); err != nil {
+				return err
+			}
+		}
+		order = append(order, node)
+		return nil
+	}
+
+	// Process releases in their original order
+	for _, release := range g.releases {
```
idsulik, line 165: @alphanota here, we process the releases by their original order

pkg/skaffold/deploy/helm/dependencygraph.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}
+
+	return order, nil
+}
+
+// groupReleasesByLevel groups releases by their dependency level while preserving
+// the original order within each level
+func (g *DependencyGraph) groupReleasesByLevel(order []string) map[int][]string {
+	levels := make(map[int][]string)
```
idsulik, line 177: @alphanota and, here we have a map, but the value is slice, and the slice will have the  original order

pkg/skaffold/deploy/helm/dependencygraph_test.go
```diff
@@ -0,0 +1,375 @@
...
+						t.Errorf("order not preserved within level: %s (original position %d) should come after %s (original position %d)",
+							prevRelease, originalOrder[prevRelease], currRelease, originalOrder[currRelease])
+					}
+				}
+			}
+		})
+	}
+}
+
+func TestOrderPreservationWithinLevels(t *testing.T) {
```
idsulik, line 321: @alphanota here, I wrote separate tests to check the order

idsulik:
@alphanota, you're right, I missed one place: the level-by-level iteration isn't correct because it iterates the map. I've just pushed fix for it, it extracts keys(levels), sorts them and iterates

idsulik:
@alphanota we have several helm releases and they ordered in the config file, in the first release we create kubernetes' roles and other releases use these roles, but the new version of the skaffold installs helm releases in random order and it lead to an issue, when we install helm package that needs roles, but the roles wasn't installed

alphanota:
> the rolebinding release should be installed first, but v1.24.0 installs in random order not in the order they described in the config file
> 
> ```
>     deploy:
>       helm:
>         releases:
>         - name: rolebinding-{{.BRANCH}}
>           remoteChart: repo/config-helper
>           upgradeOnChange: true
>           valuesFiles:
>           - ./branch/values.rolebinding.yaml
>           version: ~4.1.0
>         - name: redis-{{.BRANCH}}
>           repo: https://charts.helm.sh/stable
>           remoteChart: redis
>           upgradeOnChange: true
>           valuesFiles:
>           - ./branch/values.redis.yaml
>           version: ~10.4
>           setValues:
>             image.pullPolicy: IfNotPresent
>             fullnameOverride: redis
>           wait: true
> ```

This seems fine to me. rolebinding doesn't appear to be a dependency of redis so there's no need for it to be installed first? Keep in mind also the concurrency factor, they might even be installed in parallel. From what I could tell from the code pre-2.14 release, order was never guaranteed in the first place.

idsulik:
> rolebinding doesn't appear to be a dependency of redis

you're wrong, the order matters, and redis depends on rolebinding, but not explicitly(I mean, we didn't use `dependsOn` because it was added only in v2.14.0)

>Keep in mind also the concurrency factor, they might even be installed in parallel. From what I could tell from the code pre-2.14 release, order was never guaranteed in the first place.

There wasn't concurrency in the helm installation part before https://github.com/GoogleContainerTools/skaffold/blob/v2.13.2/pkg/skaffold/deploy/helm/helm.go#L265, it was added in v2.14.0

pkg/skaffold/deploy/helm/dependencygraph.go
idsulik: > All the releases in the same level do not depend on each other, so the order in which they are installed should not matter

Why do you think so? They could depend on each other implicitly, and in my case, they depend on each other implicitly - first we have to install the release that comes first in the config order, then second, then third and so on

idsulik:
@alphanota Let’s assume the order is correct, but in **v2.14.0**, there’s another bug: when `dependsOn` is present, the first level will be set, but the second one won’t. This happens because `errgroup` is created outside the `for` loop, and after the first successful `g.Wait()`, the context gets canceled, preventing the second group of releases from being set. In this PR, I moved the `errgroup` creation inside the loop so that a new one is created for each group.

idsulik:
@alphanota I’ll try to explain the installation order again. Before version **2.14.0**, there was no concurrency—each release was installed one after another in the order they appeared in the config. However, in **v2.14.0**, this order was broken. To enforce the correct order, `dependsOn` needed to be added, but the old behavior was disrupted, even though this was just a patch release.

In our project, the installation order of releases is crucial, and due to this change, deployments started failing after upgrading to **v2.14.0**. To fix this, I added the `dependsOn` option, but even that didn’t work because, on the second iteration (second level), `errgroup` canceled the `context`, preventing the next releases from being installed. We had to roll back to the previous version.

alphanota:
@idsulik Can you limit this PR to just the fix for the `dependsOn` bug? Lets adress the order thing in a separate PR. Thanks!

idsulik:
@alphanota created a separate PR for the order fix https://github.com/GoogleContainerTools/skaffold/pull/9693

alphanota:
@idsulik Thanks for breaking this up, can you add unit test / and or integration test for this behavior? Thanks!

idsulik:
@alphanota pushed tests

alphanota (approved):

idsulik:
@plumpy hi! Could you please release a patch version? because the latest version has an issue with dependsOn. Or, you're waiting for this PR https://github.com/GoogleContainerTools/skaffold/pull/9693 ?

idsulik:
@plumpy great, it has just been approved

Issue #8413, opened 2023-02-07
[Feature request] support for `--skip-refresh` when updating helm chart dependencies


muzalevs:
In our project we use a separate Helm library chart for re-usable templates. Every other chart has it defined as a dependency, e.g.:

```yaml
dependencies:
  - name: template-lib
    version: 0.1.0
    repository: file://../template-lib
```

In many our charts it is the only dependency.  It enforces us to have `skipDependencyUpdate: false`, but also updates cache of remote helm repositories every time we run deploy, which sometimes is time-consuming.

We would like skaffold to be able to pass `--skip-refresh` option to the `helm dependency update` [command](https://helm.sh/docs/helm/helm_dependency_update/#options). This way it would be possible for users to skip refreshing cache of every repo from `helm repo list`.

PR #9696, opened 2025-01-31
feat(helm): add depBuild and template flags to HelmDeployFlags schema

art-shutter:
Fixes: #8413 

**Description**
This adds new configuration options to `HelmDeployFlags` struct, `DepBuild` and `Template`

**User facing changes**
Users can now supply new option in the helm config to propagate additional cli flags to `helm template` and `helm dep build` commands when using helm as a renderer.


pkg/skaffold/render/renderer/helm/args.go
```diff
@@ -0,0 +1,58 @@
+/*
+Copyright 2022 The Skaffold Authors
```
idsulik, line 2: 2025

pkg/skaffold/render/renderer/helm/args_test.go
```diff
@@ -0,0 +1,151 @@
+/*
+Copyright 2022 The Skaffold Authors
```
idsulik, line 2: 2025

docs-v2/content/en/schemas/v4beta12.json
```diff
@@ -2318,6 +2327,15 @@
           "x-intellij-html-description": "additional flags passed to (<code>helm install</code>).",
           "default": "[]"
         },
+        "template": {
```
idsulik, line 2330: do you need this, or did you add it just in case? if the second one, then it'll be better to remove it, because the more things you add - the more things that should be support, it is better to add as needed

art-shutter: it's added here for consitency. If we don't add this for template then it means the commands `helm upgrade` and `helm install` behave differenly from `helm template`. The latter is used primarily in gitops workflows. Leaving out `template` would mean that we don't have feature parity when we use skaffold to ask helm to apply the changes directly and when we use helm to template only.

pkg/skaffold/render/renderer/helm/helm.go
```diff
@@ -193,27 +178,25 @@ func (h Helm) generateHelmManifest(ctx context.Context, builds []graph.Artifact,
...
 
 	outBuffer := new(bytes.Buffer)
 	errBuffer := new(bytes.Buffer)
 
 	// Build Chart dependencies, but allow a user to skip it.
 	if !release.SkipBuildDependencies && release.ChartPath != "" {
 		log.Entry(ctx).Info("Building helm dependencies...")
-		if err := helm.ExecWithStdoutAndStderr(ctx, h, io.Discard, errBuffer, false, env, "dep", "build", release.ChartPath); err != nil {
+		args := h.depBuildArgs(release.ChartPath)
+		if err := helm.ExecWithStdoutAndStderr(ctx, h, io.Discard, errBuffer, false, env, args...); err != nil {
```
idsulik, line 189: before: it can exit before `dep build` on lines https://github.com/GoogleContainerTools/skaffold/pull/9696/files#diff-85ea74a2a849b3a19f3a1d48895e2b217f8b1105400c66cae3463582329d07faL165, 171, 175 
after: it can exit after `dep build`, because you moved all the errors inside the `templateArgs` and call it after `dep build`.  

I'm not sure if it's ok or not, but to preserve the old behavior you can move the `templateArgs` call before the `dep build`.

pkg/skaffold/render/renderer/helm/helm.go
```diff
@@ -154,17 +154,6 @@ func (h Helm) generateHelmManifest(ctx context.Context, builds []graph.Artifact,
 		return nil, helm.UserErr(fmt.Sprintf("cannot expand chart path %q", release.ChartPath), err)
 	}
```
idsulik, line 156: it'll be better to write tests for the `generateHelmManifest` and refactor it after

pkg/skaffold/render/renderer/helm/helm.go
art-shutter: done

docs-v2/content/en/schemas/v4beta12.json
idsulik: if you need this - keep it) I meant that there is no need to add something if you don't it, because less code - easy to maintain)

pkg/skaffold/render/renderer/helm/helm.go
idsulik: yes, it was a good idea to extract them and test independently, but at the same time, you've changed the logic as I've explained here https://github.com/GoogleContainerTools/skaffold/pull/9696#discussion_r1937391395
let's wait for the mainainers)

art-shutter:
@plumpy @menahyouyeah could I get a review from maintainers please? is there a process to request it?

pkg/skaffold/render/renderer/helm/args_test.go
```diff
@@ -0,0 +1,151 @@
...
+	tests := []struct {
+		description    string
+		releaseName    string
+		release        latest.HelmRelease
+		builds         []graph.Artifact
+		namespace      string
+		additionalArgs []string
+		flags          latest.HelmDeployFlags
+		expected       []string
+		shouldErr      bool
```
menahyouyeah, line 80: nit: remove - this isnt being used

pkg/skaffold/render/renderer/helm/args_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	for _, test := range tests {
+		testutil.Run(t, test.description, func(t *testutil.T) {
+			h := Helm{
+				config: &latest.Helm{
+					Flags: test.flags,
+				},
+			}
+			args, err := h.templateArgs(test.releaseName, test.release, test.builds, test.namespace, test.additionalArgs)
+			if test.shouldErr {
```
menahyouyeah, line 143: doesn't look like this is being used - can be removed

menahyouyeah (commented):
Can you rebase? I just merged https://github.com/GoogleContainerTools/skaffold/pull/9741 which creates a new schema. Then I can merge this PR.

art-shutter:
@menahyouyeah done

art-shutter:
> Can you fix the failed unit tests? It just looks like the order of the helm render args changed so that's why the tests are failing

fixed the order of args in helm commands, since args are positional and expected to be at the positions they were (overrides must always come last)

menahyouyeah:
> > Can you fix the failed unit tests? It just looks like the order of the helm render args changed so that's why the tests are failing
> 
> fixed the order of args in helm commands, since args are positional and expected to be at the positions they were (overrides must always come last)

Just reran the tests and there are some failing - can you take a look?

pkg/skaffold/render/renderer/helm/args.go
```diff
@@ -0,0 +1,91 @@
...
+		defer func() {
+			os.Remove(constants.HelmOverridesFilename)
+		}()
+
+		return constants.HelmOverridesFilename, nil
+	}
+	return "", nil
+}
+
+func (h Helm) templateArgs(releaseName string, release latest.HelmRelease, builds []graph.Artifact, namespace string, additionalArgs []string) ([]string, error) {
```
tirimia, line 56: The test suite for this function should include a case where the overrideArgs kick in

pkg/skaffold/render/renderer/helm/args.go
```diff
@@ -34,23 +34,24 @@ func (h Helm) depBuildArgs(chartPath string) []string {
...
 
-		return constants.HelmOverridesFilename, nil
+	if err := os.WriteFile(constants.HelmOverridesFilename, overrides, 0o666); err != nil {
+		return "", helm.UserErr(fmt.Sprintf("cannot create file %q", constants.HelmOverridesFilename), err)
 	}
-	return "", nil
+
+	defer func() {
+		os.Remove(constants.HelmOverridesFilename)
+	}()
```
tirimia, line 52: 😱 just realized this will now remove the file as the location is returned, this will fail when run

art-shutter: thank you! reverted this back to inlining

pkg/skaffold/render/renderer/helm/args.go
art-shutter: added

art-shutter:
> Just reran the tests and there are some failing - can you take a look?

thanks, I did overlook an issue. It is now fixed and locally `make unit-tests` is not failing for me. Could you please re-run the tests?

menahyouyeah:
> > Just reran the tests and there are some failing - can you take a look?
> 
> thanks, I did overlook an issue. It is now fixed and locally `make unit-tests` is not failing for me. Could you please re-run the tests?

reran and there are still some broken tests

art-shutter:
> reran and there are still some broken tests

from what I can tell based on the logs, this an auth/rate-limiting problem around docker hub, not the code itself
can someone help out and rerun the tests?

menahyouyeah (approved):

vihangm:
Looks like this flag is used by the helm renderer but not the helm deploy.
See https://github.com/GoogleContainerTools/skaffold/blob/a65fe4b/pkg/skaffold/deploy/helm/helm.go#L565-L567 vs https://github.com/GoogleContainerTools/skaffold/blob/759fa38/pkg/skaffold/render/renderer/helm/args.go#L32

However the flag is configurable in both sections of `skaffold.yaml`, see https://skaffold.dev/docs/references/yaml/#deploy-helm-flags-depBuild

Is this intentional or just an oversight?

Issue #9710, opened 2025-02-06
"failed to port forward, port is taken, retrying..." error with v2.14.1


kpotter-m2:
![Image](https://github.com/user-attachments/assets/7d4bd0c3-9ed7-405d-991b-1e858c4aeb6a)

upgraded from version v2.13.2 to v2.14.1 and started getting the error above when port forwarding. downgraded to v2.13.2 fixed the issue, but am unaware of any changes that v2.14.X made that would cause this.

any help to point in the right direction for why this is happening is appreciated.

idsulik:
@kpotter-m2 could you please try again with `--verbosity trace`?

foriequal0:
I'm having the problem too.

```
time="2025-02-21T16:24:41+09:00" level=debug msg="found open port: 8080" subtask=-1 task=DevLoop
time="2025-02-21T16:24:41+09:00" level=debug msg="found open port: 6650" subtask=-1 task=DevLoop
time="2025-02-21T16:24:41+09:00" level=debug msg="found open port: 5432" subtask=-1 task=DevLoop
time="2025-02-21T16:24:41+09:00" level=debug msg="found open port: 6379" subtask=-1 task=DevLoop
time="2025-02-21T16:24:41+09:00" level=debug msg="getting client config for kubeContext: `kind-kind`" subtask=-1 task=DevLoop
time="2025-02-21T16:24:41+09:00" level=debug msg="getting client config for kubeContext: `kind-kind`" subtask=-1 task=DevLoop
time="2025-02-21T16:24:41+09:00" level=debug msg="getting client config for kubeContext: `kind-kind`" subtask=-1 task=DevLoop
time="2025-02-21T16:24:41+09:00" level=debug msg="getting client config for kubeContext: `kind-kind`" subtask=-1 task=DevLoop
time="2025-02-21T16:24:41+09:00" level=debug msg="Forwarding service pulsar/8080 to pod pulsar-standalone-0/8080" subtask=service/pulsar task=PortForward
time="2025-02-21T16:24:41+09:00" level=debug msg="Forwarding service redis-master/6379 to pod redis-master-0/6379" subtask=service/redis-master task=PortForward
time="2025-02-21T16:24:41+09:00" level=debug msg="Running command: [kubectl --context kind-kind port-forward --pod-running-timeout 1s --namespace default pod/pulsar-standalone-0 8080:8080]" subtask=service/pulsar task=PortForward
[...]

PR #9728, opened 2025-02-21
fix: port-forward error logs `failed to port forward`

foriequal0:
Fixes: #9710

**Description**
Golang has changed os.Process structure since 1.23
[cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/os/exec.go;l=21](https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/os/exec.go;l=21)
[cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/os/exec.go;l=58](https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/os/exec.go;l=58)

so this unpack doesn't work, lead to `failed to port-forward` message loop.

### DIFF
--- a/pkg/skaffold/kubectl/exec_windows.go
+++ b/pkg/skaffold/kubectl/exec_windows.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
+	"sync"
+	"sync/atomic"
 	"unsafe"
 
 	"golang.org/x/sys/windows"
@@ -65,6 +67,9 @@ func (c *Cmd) Start() error {
 	// Use `unsafe` to extract the process handle.
 	processHandle := (*struct {
 		Pid    int
+		mode   uint8 // processMode
+		state  atomic.Uint64
+		sigMu  sync.RWMutex
 		handle windows.Handle
 	})(unsafe.Pointer(c.Process)).handle
 
--- a/pkg/skaffold/kubectl/exec_windows.go
+++ b/pkg/skaffold/kubectl/exec_windows.go
@@ -64,7 +64,11 @@ func (c *Cmd) Start() error {
 		return fmt.Errorf("could not start the command: %w", err)
 	}
 
-	processHandle := getHandleFromProcess(c.Process)
+	processHandle, err := getHandleFromProcess(c.Process)
+	if err != nil {
+		return fmt.Errorf("could not get handle from process: %w", err)
+	}
+
 	if err := windows.AssignProcessToJobObject(handle, processHandle); err != nil {
 		return fmt.Errorf("could not assign job object: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/skaffold/kubectl/exec_windows.go
```diff
@@ -81,6 +78,13 @@ func (c *Cmd) Start() error {
 	return nil
 }
 
+func getHandleFromProcess(p *os.Process) windows.Handle {
+	// `os.Process` don't expose `handle uintptr` field.
+	v := reflect.ValueOf(p)
+	f := reflect.Indirect(v).FieldByName("handle")
```
Darien-Lin, line 84: There is a case that the struct changes and valueOf / indirect returns a zero value. If that is the case, FieldByName would panic if the type isn't a struct. Can you ensure the type before attempting to fieldbyname?

pkg/skaffold/kubectl/exec_windows.go
```diff
@@ -81,6 +78,13 @@ func (c *Cmd) Start() error {
 	return nil
 }
 
+func getHandleFromProcess(p *os.Process) windows.Handle {
```
Darien-Lin, line 81: Can you add a unit test for this function?

Darien-Lin (changes requested):
A few things:

pkg/skaffold/kubectl/exec_windows.go
foriequal0: The documentation says `ValueOf(p)` returns a zero value when `p` is `nil`, and `Indirect(v)` returns a zero value when `v` is a nil pointer.
Is it correct that you mean that I should handle `p == nil`?

Darien-Lin: more of a safeguard when you call `FieldByName`. Just make sure that the type when you finish the indirect call is a struct.  `FieldByName` panics if the kind isn't a struct from the documentation:  https://pkg.go.dev/reflect#Value.FieldByName

foriequal0: I've handled some panic cases. can you check them if it is okay?

pkg/skaffold/kubectl/exec_windows.go
foriequal0: I've added a basic test that executes `kubectl` and checks the result.
I've chosen it because it is sure to exist for the project, and is known to be side-effect-free

pkg/skaffold/kubectl/exec_windows.go
Darien-Lin: Looks good to me, thanks for making the change.

pkg/skaffold/kubectl/exec_windows_test.go
```diff
@@ -0,0 +1,44 @@
+/*
+Copyright 2019 The Skaffold Authors
```
Darien-Lin, line 2: nit: 2025

Darien-Lin (approved):

pkg/skaffold/kubectl/exec_windows.go
Darien-Lin: Thanks for adding the test case, looks good to me

pkg/skaffold/kubectl/exec_windows_test.go
foriequal0: I've fixed it. could you approve it again?

Darien-Lin (approved):
Looks good to me. Thanks for making the changes and contributing to the code base!

foriequal0:
Thank you for the review!

PR #9738, opened 2025-02-26
Downgrade godotenv.

plumpy:
godotenv 1.5 only allows environment variable names containing `[A-Za-z0-9_.]`. This is different than 1.4, which allowed almost any character.


Darien-Lin:
Was this bumped up b/c of vulnerabilities? if so, we should probably open an issue on that package's github about the restriction?

Darien-Lin (dismissed):

Issue #9747, opened 2025-03-07
Verify is not working for tests that share the same container image


katiexzhang:
Given a setup with multiple tests in a verify like:

```
verify:
  - name: test-1
    container:
      name: test-1
      image: 'gcr.io/cloud-builders/gcloud'
      command: ['bash']
      args:
      - '-ec'
      - |-
        echo "TESTING 1";
  - name: test-2
    container:
      name: test-2
      image: 'gcr.io/cloud-builders/gcloud'
      command: ['bash']
      args:
      - '-ec'
      - |-
        echo "TESTING 2";
```

`skaffold verify` fails with an error like:
```
* creating container in local docker: Error response from daemon: Conflict. The container name "/gcloud" is already in use by container "8a53722a6c1d1954b17eb92344a5f7f1e51876632a511b7ed8e865c3a4539ed7". You have to remove (or rename) that container to be able to reuse that name.
```
even though the two container names *are* different.

katiexzhang:
Note that this kind of setup works for custom actions:
```
customActions:
- name: postdeploy-action
  containers:
  - name: test-1
    image: 'gcr.io/cloud-builders/gcloud'
    command: ['bash']
    args:
    - '-ec'
    - |-
      echo "TESTING 1";
  - name: test-2
    image: 'gcr.io/cloud-builders/gcloud'
    command: ['bash']
    args:
    - '-ec'
    - |-
      echo "TESTING 2";
```

Verify and custom actions are using different logic it seems, so for verify we should mirror what custom actions is doing.

PR #9753, opened 2025-03-13
fix(verify): use container name from configuration in verify tests

idsulik:
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/9747

**Description**
Add using container name from configuration in verify tests

### DIFF
--- a/pkg/skaffold/verify/docker/verify.go
+++ b/pkg/skaffold/verify/docker/verify.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"math"
 	"path"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -48,6 +49,8 @@ import (
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/status"
 	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/util"
 )
+
+var validContainerNameRegex = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9_.-]*$`)
 
 // Verifier verifies deployments using Docker libs/CLI.
 type Verifier struct {
@@ -308,7 +311,7 @@ func (v *Verifier) getContainerName(ctx context.Context, imageName string, conta
 	name := containerName
 
 	// If no container name is provided, derive it from the image name
-	if name == "" {
+	if name == "" || !validContainerNameRegex.MatchString(name) {
 		name = path.Base(strings.Split(imageName, ":")[0])
 	}
 
--- a/pkg/skaffold/verify/docker/verify.go
+++ b/pkg/skaffold/verify/docker/verify.go
@@ -221,7 +221,10 @@ func (v *Verifier) createAndRunContainer(ctx context.Context, out io.Writer, art
 	if len(tc.Container.Args) != 0 {
 		containerCfg.Cmd = tc.Container.Args
 	}
-	containerName := v.getContainerName(ctx, artifact.ImageName)
+
+	// Use container name from test case if available, otherwise derive from image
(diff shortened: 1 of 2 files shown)


katiexzhang:
Thanks for the PR, I tested it locally and it seems to be working.

It looks like an integration test is failing however.

pkg/skaffold/verify/docker/verify.go
```diff
@@ -301,9 +304,14 @@ func (v *Verifier) containerConfigFromImage(ctx context.Context, taggedImage str
 	return config.Config, err
 }
 
-func (v *Verifier) getContainerName(ctx context.Context, name string) string {
-	// this is done to fix the for naming convention of non-skaffold built images which verify supports
-	name = path.Base(strings.Split(name, ":")[0])
+func (v *Verifier) getContainerName(ctx context.Context, imageName string, containerName string) string {
+	name := containerName
```
katiexzhang, line 311: I'm not sure replacing the image name ("/gcloud") with the container name ("testing1") is the way to go.

From this func it seems like a UUID should be getting attached to the containerName if the container exists already (which in the example of a double verify, it does). But the error message "The container name "/gcloud" is already in use..." suggests this is not happening.

idsulik: @katiexzhang I think it's a concurrency issue and it happens because of the `errgroup`, this will help if the container names are diff and they should be different while the one image we can use for multiple containers

idsulik: 1. A goroutine checks if the image exists -> FALSE
2. B goroutine checks if the image exists -> FALSE
3. A goroutine creates the image
4. B goroutine creates the image - FAILS

idsulik: another way is to handle the container exists error and try with a new name)

katiexzhang: Hmm custom actions has a similar implementation as verify for docker and is working as intended (same container image but different name): pkg/skaffold/actions/docker/task.go

idsulik: it's a concurrency issue, so most of the time, it can work properly, and only sometimes it fails(when this conditions are met https://github.com/GoogleContainerTools/skaffold/pull/9753#discussion_r1999421907)

katiexzhang (commented):
It looks like container name is required for verify actually, so this would solve the problem completely. 

Something else to consider - wdyt about extracting the name checking logic in getContainerName to run in the s.Go()? If we check the container name existence every time before createAndRunContainer is called, then it should be creating the new ones with diff UUIDs.

idsulik:
> It looks like container name is required for verify actually, so this would solve the problem completely.
> 
> Something else to consider - wdyt about extracting the name checking logic in getContainerName to run in the s.Go()? If we check the container name existence every time before createAndRunContainer is called, then it should be creating the new ones with diff UUIDs.

I think it's also a good solution, it'll reduce a bit the execution speed, because you have to call the docker client in sync mode multiple times, but I think it should be okay, because it shouldn't take a lot of time and it'll make sense if a user has tons of containers. And, I think it'll be better to extract the `getContainerName` func into separate package and start reusing the logic instead of having two almost similar functions in task/verify (for some reason the `containerName` in the `task.go ` doesn't generate name until it finds free name, it just makes 2 attempts - one with image name and with image name + uuid's part)

katiexzhang:
I'm going to merge this PR for now + create another issue to unify the verify + custom actions logic. I'm still not sure exactly why everything is working for custom actions (likely has to do with how it gets called + the threads start running), so refactoring might be more complicated than anticipated.

katiexzhang (approved):

PR #9759, opened 2025-03-19
fix: port-forward error logs `failed to port forward` (#9728)

mattsanta:
* fix: port-forward error logs `failed to port forward`

Reflect changes in `os.Process` introduced in Go 1.23

* use `reflect` to make it future proof(?)

* Add safeguards to call FieldByName

* Add basic test to getHandleFromProcess

* Fix copyright year

### DIFF
--- a/pkg/skaffold/kubectl/exec_windows.go
+++ b/pkg/skaffold/kubectl/exec_windows.go
@@ -19,7 +19,9 @@ package kubectl
 import (
 	"context"
 	"fmt"
+	"os"
 	"os/exec"
+	"reflect"
 	"unsafe"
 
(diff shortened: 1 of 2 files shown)


mattsanta:
Linter is failing because a new schema version was added after this commit was merged to main and the linter is looking for the files generated when that new schema version was added. Since this commit we're cherry picking doesn't have any config changes I think we should be OK to ignore the failure.

Error in linter:
```
Error: time="2025-03-19T13:11:55Z" level=fatal msg="reading \"pkg/skaffold/schema/v4beta12/config.go\": open pkg/skaffold/schema/v4beta12/config.go: no such file or directory"
```

Issue #9251, opened 2024-01-11
The `--cache-artifacts=false` flag provoking race-condition and leads to prune of the wrong images.


AlexzAK:
### Expected behavior

The `--cache-artifacts=false` flag works as specified in the docs.

### Actual behavior

Some percent of the `skaffold run --cache-artifcats=false` end up with error `ErrImagePull`. For our project broken build happens in 10% of cases. Depends on the project.

### Information

- Skaffold version: v2.7.1
- Operating system: Linux, Ubuntu 20.04.6
- Installed via: GitHub Releases
- Contents of skaffold.yaml: Not provided

### Steps to reproduce the behavior

1. Slow down image prune in the code. 
2. Run `skaffold run --cache-artifcats=false` many times.

In the file `pkg/skaffold/build/local/local.go` there is a code:
```
func (b *Builder) Build(ctx context.Context, out io.Writer, a *latest.Artifact) build.ArtifactBuilder {
	if b.prune {
		b.localPruner.asynchronousCleanupOldImages(ctx, []string{a.ImageName})
 	}
 	builder := build.WithLogFile(b.buildArtifact, b.muted)
 	return builder
```

It eventually calls `cleanup` with `sync = false`:
```
func (p *pruner) cleanup(ctx context.Context, sync bool, artifacts []string) {
	toPrune := p.collectImagesToPrune(ctx, artifacts)
	if len(toPrune) == 0 {
		return
	}

	if sync {
		err := p.runPrune(ctx, toPrune)
		if err != nil {
			log.Entry(ctx).Debugf("Failed to prune: %v", err)
		}
	} else {
		go func() {
			err := p.runPrune(ctx, toPrune)
			if err != nil {
				log.Entry(ctx).Debugf("Failed to prune: %v", err)
			}
		}()
	}
}
```

If there are a lot of fat docker images, then the `runPrune` will take some time to delete them.

Meanwhile the rebuild process is running concurrently and might to finish image rebuild _before_ the `runPrune`  call completes. If that happens you might see events in wrong order:
- First rebuild the docker image by id
- Second delete the docker image by id

If there were no changes in the docker file then the id of the image have not been changed. And the Skaffold deleted it. Due to that the project will not be able to start due to missing image ( `ErrImagePull`).

It is possible to run the prune with `sync = true` to fix that:
```
diff --git a/pkg/skaffold/build/local/local.go b/pkg/skaffold/build/local/local.go
index 368894b36..768a5d570 100644
[...]

PR #9787, opened 2025-04-11
Bugfix: Only prune images if changed

mthaak:
I hope I understood this right...

**Description**
I observed that when using `skaffold dev`, when images were re-built because the dependencies defined in Skaffold changed (I use ko), but didn't result in a new image ID (because the contents hadn't changed - due to dependencies being too broad), then the previous image would be pruned. This means that images which aren't currently in use, like those used by k8s Jobs and CronJobs are removed, causing the deployment to fail.

**Solution**
Now we check whether the image ID of the existing artifact changed before removing it. Because if it's the same then we should keep it.

I tested this solution locally and seems to fix my problem.

Related: #9251

### DIFF
--- a/pkg/skaffold/build/local/local.go
+++ b/pkg/skaffold/build/local/local.go
@@ -113,8 +113,15 @@ func (b *Builder) buildArtifact(ctx context.Context, out io.Writer, a *latest.Ar
 		go func() {
 			if len(artifacts) > 0 {
 				bgCtx := context.Background()
-				id, _ := b.getImageIDForTag(bgCtx, artifacts[0].Tag)
-				b.localPruner.runPrune(bgCtx, []string{id})
+				id, err := b.getImageIDForTag(bgCtx, artifacts[0].Tag)
+				if err != nil {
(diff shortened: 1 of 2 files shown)


mthaak:
How do I get this approved? CLA is signed btw

katiexzhang (changes requested):
LGTM, could you please add a test for this?

mthaak:
> LGTM, could you please add a test for this?

Sure! How does this [look](https://github.com/GoogleContainerTools/skaffold/pull/9787/commits/2b6bfddea0b76920a9b5f5e640552c15913bcb41)? I'm not at all sure confident of this change so definitely let me know if I misunderstood the issue

katiexzhang (changes requested):
I think the change makes sense.

Thanks for the tests, could you please address the PR checks?

katiexzhang (changes requested):
Thanks for the updates, there are still a couple of tests failing

mthaak:
@katiexzhang do you agree with the change though? I'm worried I don't understand skaffold enough because this seems quite an obvious oversight

katiexzhang (approved):

katiexzhang:
Yes, the change LGTM and probably should've been caught earlier on.

PR #9808, opened 2025-05-03
[bug]: Support helm depBuild flags for helm deploys

vihangm:
Related: #8413, #9696

**Description**
The helm depBuild flags were added in the most recent version of skaffold, but
they are currently only supported when using helm to render and not supported
when using helm to deploy. This ensures that the flags are applied on deploy
too.

The flag is already a part of the schema, it was just ignored before this.


vihangm:
Sorry for the delay here, I am waiting on the authorized signatory to complete the CLA!

vihangm:
@alphanota This is good to go now!

alphanota (approved):

vihangm:
@alphanota It's been almost 2 months since the last release. Any update on future release schedules? I'm waiting for this fix to make it in.

PR #9825, opened 2025-06-04
Update PrintNamespacesList to break on YAML file separator '---' followed by a new line.

katiexzhang:
Multiple manifests may be defined in a single YAML file. `---` is used to separate the different manifests.

Before:
PrintNamespacesList breaks on `---` to get each manifest individually and then find the namespace in it.
If any string in the manifest (say the deployment name is "hello---world") then it will break on that as well and try to read the namespace of a jumbled manifest.

After:
PrintNamespacesList breaks on `---\n`.

### DIFF
--- a/pkg/skaffold/inspect/namespaces/list_test.go
+++ b/pkg/skaffold/inspect/namespaces/list_test.go
@@ -61,7 +61,7 @@ kind: Deployment
 metadata:
   labels:
     app: leeroy-app
-  name: leeroy-app
+  name: leeroy-app---rel01
   namespace: manifest-namespace
 spec:
   replicas: 1
@@ -119,7 +119,7 @@ func TestPrintTestsList(t *testing.T) {
 		{
 			description: "print all deployment namespaces where the manifest has a namespace set but it is also set via the kubectl flag deploy config",
 			manifest:    manifestWithNamespace,
-			expected:    `{"resourceToInfoMap":{"apps/v1, Kind=Deployment":[{"name":"leeroy-app","namespace":"manifest-namespace"}]}}` + "\n",
+			expected:    `{"resourceToInfoMap":{"apps/v1, Kind=Deployment":[{"name":"leeroy-app---rel01","namespace":"manifest-namespace"}]}}` + "\n",
 			profiles:    []string{"baz-flag-ns"},
 			module:      []string{"cfg-with-default-namespace"},
 		},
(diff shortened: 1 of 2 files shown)


plumpy:
What about

```
yaml: "
  ---
"
```

plumpy:
Also YAML explicitly allows `\r\n` as a valid line break (for Windows compatibility): https://yaml.org/spec/1.2.2/#54-line-break-characters

katiexzhang:
> What about
> 
> ```
> yaml: "
>   ---
> "
> ```

Right now this returns an error like:
`2025/06/04 18:42:37 couldn't get version/kind; json parse error: json: cannot unmarshal string into Go value of type struct { APIVersion string "json:\"apiVersion,omitempty\""; Kind string "json:\"kind,omitempty\"" }`

A file with just `---` is technically empty so according to the comment "skip empty documents, `Decode` will fail on them" we should just skip it (and return nothing / no error). But, it is weird to have a manifest with just `---` right?

plumpy:
Sorry, that was a confusing example, but my point was that this is technically a valid YAML file:

```
foo: "
   ---
"
```

[If you parse it](https://yaml-online-parser.appspot.com/?yaml=foo%3A+%22%0A+++---%0A%22&type=json), you get a map: `{"foo": " --- "}`. But your code is just going to split on `---\n` and rip it into two different documents (neither of which is valid YAML).

My larger point is that trying to parse any structured format with a basic text-processing tools (like `strings.Split`) is not going to work very well. Just quickly perusing the spec, I can find all kinds of other ways to make valid YAML that won't work with your code. (You're allowed to have whitespace and comments and a variety of other things after the `---`, for example; it doesn't need to be immediately followed by a newline.)

In short: if you want to parse YAML, use a YAML parser.

katiexzhang:
Ok, updated to use https://pkg.go.dev/gopkg.in/yaml.v3 for parsing the multiple files, but kept the kubernetes file parser.

So for the file with just "---" --> should that be valid then?

plumpy:
Yeah, I think this looks great, thanks.

pkg/skaffold/inspect/namespaces/list_test.go
```diff
@@ -56,12 +56,108 @@ spec:
...
+      labels:
+        app: leeroy-app
+    spec:
+      containers:
+      - image: leeroy-app:1d38c165eada98acbbf9f8869b92bf32f4f9c4e80bdea23d20c7020db3ace2da
+        name: leeroy-app
+        ports:
+        - containerPort: 50051
+          name: http
+---
```
ChrisGe4, line 81: Does it make sense to add a case 

```
---

---
```

Issue #5216, opened 2021-01-11
remoteManifests don't replace image on deployment if digest specified on existing deployment


spmason:
### Expected behavior

When using `remoteManifests`, if the deployed manifest contains an image digest then the image is not replaced with the one built by skaffold

### Actual behavior

The image is not replaced and a warning `WARN[0002] image [foo] is not used by the deployment` is printed

### Information

- Skaffold version: v1.17.2
- Operating system: MacOS Mojave
- Contents of skaffold.yaml:

This is a basic reproduction I've cooked up

```yaml
---
apiVersion: skaffold/v2beta10
kind: Config
metadata:
  name: nginx
build:
  # Abuse the fact that the image is already pushed with `:latest` to docker.io
  tagPolicy:
    sha256: {}
  artifacts:
  - image: nginx
    custom:
      # Skaffold will simply pick up the version of `nginx` with the `:latest` tag from docker.io
      buildCommand: true

deploy:
  kubectl:
    manifests: [""]
    remoteManifests:
    - skaffold-digest-bug:deployment/nginx
```

And deployment.yaml:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
  namespace: skaffold-digest-bug
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2@sha256:4cf620a5c81390ee209398ecc18e5fb9dd0f5155cd82adcbae532fec94006fb9
        ports:
        - containerPort: 80
```

### Steps to reproduce the behavior

1. Use the above `skaffold.yaml` and `deployment.yaml` files
2. `kubectl create ns skaffold-digest-bug`
3. `kubectl apply -f deployment.yaml`
4. `skaffold run`
5. You'll see the warning `WARN[0002] image [nginx] is not used by the deployment` as described
6. Edit `deployment.yaml` to remove the sha from the image (ie `image: nginx:1.14.2`)
7. `kubectl apply -f deployment.yaml`
8. `skaffold run`
9. The warning will not appear

It looks like the problem is here: https://github.com/GoogleContainerTools/skaffold/blame/v1.17.2/pkg/skaffold/kubernetes/manifest/images.go#L108

[...]

PR #9828, opened 2025-06-07
demonstrate that remoteManifests rendering is broken

ctrahey:
**Related**:
 #5855
 #5216

**Description**

The method `ReplaceRemoteManifestImages` is no longer used anywhere except the test case modified in this PR. This is a hint that all manifests, even those from remote, go through `ReplaceImages`. So this one-line PR exists to demonstrate that a regression has taken place for remoteManifests by showing that it breaks a test case; but is *correct* to be broken, because the without the change in this PR the test is exercising an orphaned code path. 
The expectations outlined in the test case are real expectations that the community relies on for remote manifests. 

If I were a better Go programmer, I would love to also PR a fix for this, but I hope this PR make the issue easy to understand. 

TL;DR: If this test case were an integration test rather than a unit test, I think it would have started failing 2.5 years ago.

**User facing changes**

Before:
All tests pass, so we think there is no issue
After:
kubernetes/manifest unit tests fail, revealing the truth about how remoteManifests no longer work

The handling of remoteManifests needs some tune-up. 
Commits in #7603 and #7658 fully removed any _actual_ code paths to [ReplaceRemoteManifestImages](https://github.com/GoogleContainerTools/skaffold/blob/main/pkg/skaffold/kubernetes/manifest/images.go#L144).

### DIFF
--- a/pkg/skaffold/kubernetes/manifest/images_test.go
+++ b/pkg/skaffold/kubernetes/manifest/images_test.go
@@ -125,7 +125,7 @@ spec:
 		fakeWarner := &warnings.Collect{}
 		t.Override(&warnings.Printf, fakeWarner.Warnf)
 
-		resultManifest, err := manifests.ReplaceRemoteManifestImages(context.TODO(), builds, NewResourceSelectorImages(TransformAllowlist, TransformDenylist))
+		resultManifest, err := manifests.ReplaceImages(context.TODO(), builds, NewResourceSelectorImages(TransformAllowlist, TransformDenylist))
 
 		t.CheckNoError(err)
 		t.CheckDeepEqual(expected.String(), resultManifest.String(), testutil.YamlObj(t.T))


ChrisGe4:
Are you trying to use this PR to notify us the `ReplaceRemoteManifestImages` or `ReplaceImages` is not working properly? Or is there a specific test case in `TestReplaceRemoteManifestImages` should be in `TestReplaceImages`? 

 I'm having a hard time to understand the change, and the change doesn't seem to be reasonable to me.

ctrahey:
Its a bit more nuanced than just a notify-via-pr...
The test case I modified here was testing a code path that is not longer reached from anywhere. So the test case was a false positive, but even worse, the expectation that the test case exists to validate is still important - so the fact that my change makes it fail is, in my suggestion, to be taken the same way it should have been taken if the test had started failing back when #7603 and #7658 were open. If they had failed during that work, a fix would likely have been supplied.

ctrahey:
If it helps clarify, this is my super ugly hack, but this is an absolute pain to maintain across many services: 
```
manifests:
  hooks:
    # This is a hack to work around a defect in Skaffold's rendering of remoteManifests.
    # We're just removing sha digest from the existing deployment because skaffold
    # is accidentally treating remoteManifests the same way as local, where images
    # with digests specified are intentionally skipped.
    before:
      - host:
          command: ["sh", "-c", 'kubectl get deployment -n my-ns -o yaml my-deployment | sed -E "s/^([[:space:]]*image: ${SKAFFOLD_IMAGE_REPO}[^@]*)@sha256.*$/\1/" | kubectl apply -n my-ns --server-side=true --overwrite=true -f -; sleep 5']
          os: [darwin, linux]
```

Issue #7008, opened 2022-01-12
Skaffold + Kaniko - Not error when kaniko.noPush is true


vikas027:
### Expected behavior
skaffold should not fail when `noPush` is true under kaniko

### Actual behavior
skaffold errors out with this error

```
INFO[0176] Skipping push to container registry due to --no-push flag 
GET https://1111111.dkr.ecr.ap-southeast-2.amazonaws.com/v2/my-image/manifests/<tag>: MANIFEST_UNKNOWN: Requested image not found
Error: Process completed with exit code 1.
```
Kaniko/Skaffold should not be checking whether the image has been pushed or not if `.build.artifacts.[0].kaniko.noPush` is `true`

### Information

- Skaffold version: 
```
$ skaffold version
v1.35.1
```
- Operating system: 
```
Ubuntu 20.04.3 LTS (Focal Fossa)"
```

- Installed via: 
GitHub Releases

- Contents of skaffold.yaml:

```yaml
---
apiVersion: skaffold/v2beta26
kind: Config
metadata:
  name: build-image
build:
  cluster:
<redacted>
  artifacts:
  - image: 1111111.dkr.ecr.ap-southeast-2.amazonaws.com/my-image
    kaniko:
      dockerfile: ./Dockerfile_vikas027
      noPush: true
```

### Steps to reproduce the behavior

1. `skaffold build`

tejal29:
Unfortunately the way skaffold is designed, skaffold needs to know the sha of the image built to substitute. 
Hence `--no-push` is not supported in skaffold. 

We should definitely document this case.  I am keeping this ticket open to address the doc issue.

vikas027:
> Unfortunately the way skaffold is designed, skaffold needs to know the sha of the image built to substitute.

Hey @tejal29 ,
Is there an alternative to just build the image locally and not push it?

tejal29:
no. You could add functionality to Kaniko to programmatically report the sha of the built image. 
It would need changes both in kaniko and skaffold.

tejal29:
re-opening. Closed by mistake. 


Unfortunately at this time, this is "wont-fix" for skaffold team. Keep this bug open, since we can definitely work on adding a validation when `--no-push` is specified in the kaniko flags.

vikas027:
Hey @tejal29 ,

[...]

PR #9860, opened 2025-08-28
Fixed no-push flag and added compressedCaching option when building skaffold kaniko

rayshoo:
Fixes: #7008
Fixed kaniko's no-push option.

Fixes: [compressedCaching](https://github.com/GoogleContainerTools/kaniko?tab=readme-ov-file#flag---compressed-caching)
Added to support kaniko's compressedCaching=false flag.

**Description**
I used kaniko when building our company's CICD, but while trying to migrate to skaffold for multi-architecture builds, I discovered that the no-push option and `compressedCaching=false` that I had been using were not possible with skaffold. While testing kaniko when building CICD, I discovered that the no-push option could only be used if there was no destination. Also, when there are a lot of images and the like during build, turning off the compressed caching option was a necessary option because it reduced excessive memory usage, although the build time increased.

skaffold enables the use of kaniko's no-push and compressedCaching=false options.

To add the new flag feature, I looked into reverting the commit that removed the existing no-push feature and added the necessary code.

### DIFF
--- a/pkg/skaffold/build/platform.go
+++ b/pkg/skaffold/build/platform.go
@@ -46,7 +46,9 @@ func CreateMultiPlatformImage(ctx context.Context, out io.Writer, a *latest.Arti
 	if err != nil {
 		return "", err
 	}
-
+	if len(images) == 0 {
+		return tag, nil
+	}
 	return docker.CreateManifestList(ctx, images, tag)
 }
 
@@ -63,6 +65,10 @@ func buildImageForPlatforms(ctx context.Context, out io.Writer, a *latest.Artifa
 
 		if err != nil {
 			return nil, err
+		}
+
+		if a.KanikoArtifact != nil && a.KanikoArtifact.NoPush {
+			continue
 		}
 
 		pl := util.ConvertToV1Platform(p)


rayshoo:
I apologize for the immaturity of my initial PR due to my limited contribution experience. I found that the test code is sensitive to the order of arguments, which made it difficult to pass the tests with the code I had previously modified.
Although I am not fully satisfied with this approach, I revised the code as follows in order to minimize changes while ensuring the tests pass.
In addition, since the skaffold schema config has been updated and the previous version (v4beta13) has already been released, I added a commit that runs ./hack/new-version.sh as described in the documentation. Please let me know if a revert is necessary. I would appreciate your review.
The commit history is a bit messy, so if this gets merged, I would kindly ask for it to be squashed. Thank you.

rayshoo:
I sincerely apologize for the immaturity of my initial PR, as my contribution experience is still limited. Since there were config changes and I noticed that the previous version had already been released, I created a new version and added a commit with the new feature. However, the commit history became cluttered, so I consolidated it into two final commits to keep things cleaner.  

I also later discovered that the test code is quite sensitive to the order of Kaniko arguments. Adapting the test code to the main code changes turned out to be more challenging than expected, so I minimized the modifications while still ensuring the tests pass.  

Please let me know if there are any additional changes or adjustments you would like me to make. Thank you very much for your time and review.

rayshoo:
The modified Kaniko no-push works as expected for single-platform builds. However, the issue described in the main text occurs in multi-platform builds. The problem seems to arise when a new image is generated after each platform build by checking the remote registry metadata. This patch updates Kaniko no-push to skip that step. While it may not be the most optimal solution, it has been tested and works reliably in practice.

rayshoo:
The modified Kaniko no-push works correctly for single-platform builds. However, the issue described in the main text was confirmed to occur in multi-platform builds. The problem appears to arise during the process of generating a new image after each platform build by checking the remote registry metadata. Accordingly, Kaniko no-push was modified to skip this step. While it may not be the optimal solution, tests have shown that it works reliably, including layer caching after the build.

rayshoo:
I believe it would be more appropriate to split this PR. Therefore, I am reverting `b903841` commit and will submit a separate PR specifically addressing the #9861 issue.

Issue #9861, opened 2025-08-31
Kaniko multi-platform build actually runs as single-platform


rayshoo:
This issue was discovered while working on PR #9860.
The tests were conducted using the official Skaffold image:
skaffold:latest (gcr.io/k8s-skaffold/skaffold:latest).

When building with Kaniko through Skaffold, the platforms array is effectively locked to the first index only, due to the nodeSelector (kubernetes.io/arch) being fixed. As a result, subsequent platforms are not built for their intended architecture but instead fall back to the first platform.

For example, with the following configuration:

```
platforms:
  - linux/amd64
  - linux/arm64
```

After the first linux/amd64 build, the second build should be scheduled on a linux/arm64 node. However, it is still scheduled on the linux/amd64 node (nodeSelector: kubernetes.io/arch=amd64) and thus produces another amd64 image. The image is tagged and pushed as `IMAGE_NAME:IMAGE_TAG_arm_64`, and grouped together as if it were a valid multi-platform image, but in reality the supposed arm64 image is also recognized as amd64.
On the other hand, if linux/arm64 is the first index in the platforms array, all remaining images are also built as arm64.

It seems this might be fixed by adjusting the nodeSelector, but the codebase is quite large and difficult for me to patch alone. I’m filing this issue in case anyone knows a solution or can help address it.

```yaml
# skaffold.yaml used for testing
---
apiVersion: skaffold/v4beta13
kind: Config
build:
  tagPolicy:
    envTemplate:
      template: multi-arch-test
  artifacts:
  - image: harbor.deleo.co.kr/chan/test
    context: .
    platforms:
    - linux/amd64
    - linux/arm64
    kaniko:
      cleanup: true
      useNewRun: true
      snapshotMode: redo
  cluster:
    namespace: test
    dockerConfig:
      secretName: test
    tolerations:
    - effect: NoSchedule
      key: karpenter/node.build
      operator: Exists
    nodeSelector:
      node: build
    resources: {}
```

![image](https://dragonz.dev/assets/images/multi-platforms-bug.JPG)

```sh
ubuntu@docker-0:~/test/skaffold$ docker pull harbor.deleo.co.kr/chan/test:test-2-origin
test-2-origin: Pulling from chan/test
6e174226ea69: Pull complete
4a821b521af1: Pull complete
Digest: sha256:651a8739815fcd8dabf13964f80e3ce3901ca554c38e616e3d58b65708629ff3
[...]

PR #9862, opened 2025-09-01
fix: Switched to multi-node builds to support Kaniko multi-platform builds

rayshoo:
Fixes: #9861

**Description**
Kaniko cannot generate multi-platform images from a single node.
- amd64 builds must run on amd64 nodes.
- arm64 builds must run on arm64 nodes.

Although the skaffold.yaml config schema suggests multi-platform support, the resulting images actually only contained a single platform.
If the first build was amd64, subsequent images (even marked as arm64) were still amd64.
This behavior is essentially a bug.
The fix allows building images separately on the correct architecture nodes and then combining them into a genuine multi-platform image. Verified by testing.

1. Modified the loop in pkg/skaffold/build/platform.go#L59 to reflect each architecture as a node selector, since the architectures are already being passed individually.
2. Fixed an issue where the IMAGE_TAG environment variable was always suffixed with the first build architecture.

### DIFF
--- a/pkg/skaffold/build/cluster/pod.go
+++ b/pkg/skaffold/build/cluster/pod.go
@@ -43,6 +43,9 @@ func (b *Builder) kanikoPodSpec(artifact *latest.KanikoArtifact, tag string, pla
 	if err != nil {
 		return nil, fmt.Errorf("building args list: %w", err)
 	}
+	if len(platforms.Platforms) == 1 {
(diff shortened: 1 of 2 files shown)


rayshoo:
When building an arm64 image with Kaniko, if the base image is a multi-platform image (e.g., FROM baseimage), Kaniko would pull the amd64 image instead of the arm64 one. Although the build completed successfully, the resulting image failed to run properly. To resolve this, I added Kaniko’s --custom-platform flag, which ensured that the correct architecture’s image was pulled during the build. After this change, the image built and ran successfully on the intended architecture.

Issue #3039, opened 2019-10-12
Improve support for registries with self-signed certs (x509: certificate signed by unknown authority)


balchua:
### Expected behavior
Skaffold should allow the use of private registries using self signed certificates.

### Actual behavior
Skaffold complains about the certificate being signed by unknown authority.

### Information

- Skaffold version: tested with bleeding edge (the version with `skipTLS` option)
- Operating system: RedHat 7.5.
- Contents of skaffold.yaml:

```yaml
apiVersion: skaffold/v1beta15
kind: Config
profiles:
 - name: sub
    build:
      artifacts:
      - image: my-private-registry.com/balchu/gonuts-sub
        context: sub
        kaniko:
          dockerfile: Dockerfile
          skipTLS: true
          buildContext:
            localDir: {}
          cache:
            repo: my-private-registry.com/balchu/gonuts-sub          
      cluster:
        dockerConfig: 
          secretName: regcred
        namespace: gonuts
      insecureRegistries: #Use this for local registry.  such as microk8s registry.
      - my-private-registry.com
    deploy:
      helm:
        releases:
          - name: gonuts-sub
            chartPath: k8s-manifest/sub
            namespace: gonuts
            wait: true
            values:
              image.repository: my-private-registry.com/balchu/gonuts-sub
```

### Steps to reproduce the behavior

1. Find a private registry with a self signed certificate
2. Make sure your Dockerfile's base images are also using the private registry.

For example:

```
# Use base golang image from Docker Hub
FROM my-private-registry.com/golang:1.12.10 as build

WORKDIR /src/github.com/balchua/gonuts

# Copy go.mod and go.sum 
ADD ./go.mod /src/github.com/balchua/gonuts/
ADD ./go.sum /src/github.com/balchua/gonuts/
# Install dependencies in go.mod and go.sum
RUN go mod download

# Copy application source code
COPY ./main.go /src/github.com/balchua/gonuts

# Compile the application to /app.
RUN go build -o /app -v .

# Now create separate deployment image
FROM my-private-registry.com/distroless/base
COPY --from=build /app /app
# Cause full tracebacks; also serves to identify this image as a Go image for `skaffold debug`
ENV GOTRACEBACK=all
ENTRYPOINT ["/app"]
```
Skaffold complains about the private registry's certificates and then it fails.  It does not even start the Kaniko pod.

Thanks!

raul1991:
I am getting the same error as well. Is there any workaround yet ?

[...]

PR #9863, opened 2025-09-01
fix: Enable registries with self-signed certificates

halvards:
This change enables Skaffold to push to container image registries that use self-signed or other untrusted certificates, when Skaffold runs on macOS or Windows. It also removes the need for the `SSL_CERT_FILE` environment variable workaround on Linux.

Prior to this fix, Skaffold would fail to retrieve the digest from the registry after the image was built, even if the registry was configured as an insecure registry in Skaffold configuration:

```
getting image: Get "https://localhost:8443/v2/": tls: failed to verify certificate: x509: certificate signed by unknown authority; GET http://localhost:8443/v2/: unexpected status code 400 Bad Request: Client sent an HTTP request to an HTTPS server.
```

On Linux environments only, a possible workaround was to set the `SSL_CERT_FILE` environment variable. However, this workaround cannot be used on macOS or Windows.

This change updates `getRemoteIndex()` and `getRemoteImage()` in `pkg/skaffold/docker/remote.go`, adding the `InsecureSkipVerify` TLS config field to the transport if the registry from the image name matches one of the insecure registries configured in Skaffold.

Fixes: #3039 #3116
Related: https://github.com/google/go-containerregistry/pull/1559

### DIFF
--- a/pkg/skaffold/docker/remote.go
+++ b/pkg/skaffold/docker/remote.go
@@ -18,7 +18,9 @@ package docker
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
+	"net/http"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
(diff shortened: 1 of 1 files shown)


plumpy:
Can  you rebase this so the tests will run again? Sorry for the delay.

halvards:
> Can you rebase this so the tests will run again? Sorry for the delay.

Thanks for looking, done!

plumpy:
Sorry for doing this _again_ but there was another issue with the integration tests (some third-party containers we used were removed). Could you merge to head one more time (to get #9904) so that the tests will hopefully all pass?

Thanks!

halvards:
Done.

Issue #9878, opened 2025-10-12
refs for external skaffold dependencies


maxplumley:
Hi folks, 

Loving Skaffold, thanks for all the hard work on this! 

I've been messing around with external configurations hosted in git repositories and wondering if there were more options for pining a dependency to something like a tag, rather than just the git branch. I noticed that there are a number of closed issues such as #7018 and #9430, which suggest that it's possible to pass a commit SHA to `git.ref` - I've given this a shot and it doesn't appear to be working. Just wondering if I'm missing something here - this seems like a really great enhancement.

Things I've tried:

I've created a simple configuration that reproduces the issue [here](https://github.com/maxplumley/skaffold-minimal)

setting `ref` to a commit SHA

```yaml
requires:
  - git:
      repo: https://github.com/maxplumley/skaffold-external.git
      path: skaffold.yaml
      ref: e919f57aae48bdb69c262dd26093a2debd433518
```

and running `skaffold dev`, results in this error

```txt
parsing skaffold config: error parsing skaffold configuration file: caching remote dependency https://github.com/maxplumley/skaffold-external.git: failed to clone repo: running [/opt/homebrew/bin/git clone https://github.com/maxplumley/skaffold-external.git ./eMqefSnXMmhKzHNfWBPPEaZGg22-V7Bo --branch e919f57aae48bdb69c262dd26093a2debd433518 --depth 1]
 - stdout: ""
 - stderr: "Cloning into './eMqefSnXMmhKzHNfWBPPEaZGg22-V7Bo'...\nfatal: Remote branch e919f57aae48bdb69c262dd26093a2debd433518 not found in upstream origin\n"
 - cause: exit status 128
```

using Skaffold v2.16.1

PR #9891, opened 2025-10-18
fix: streamline repo cloning logic in gitutil.go

idsulik:
Fixes: https://github.com/GoogleContainerTools/skaffold/issues/9878

**Description**
Fixed error handling to properly fall back from branch-based cloning to commit SHA fetching when needed.
Previously, the code parsed error messages to determine fallback behavior, which is fragile and can break if git changes its error messages


idsulik:
@plumpy hi! could you please check it?

Issue #9871, opened 2025-09-29
Helm v4 breaking change in `helm template --post-renderer`, Skaffold failing


AndiDog:
You'll get an error like this with Skaffold + Helm v4 pre-release:

```
Error: invalid argument "/opt/homebrew/bin/skaffold" for "--post-renderer" flag: plugin: {Name:/opt/homebrew/bin/skaffold Type:postrenderer/v1} not found
```

I'm trying to find out a good implementation path in the upstream issue https://github.com/helm/helm/issues/31340.

scottrigby:
hi @AndiDog 👋 

I'm a Helm maintainer who proposed the new plugin system in Helm 4, which changed the post-renderer behavior. Skaffold should still work as a post-renderer in Helm 4, with some minimal changes on the Skaffold side. I'd like to work with you on fixing this in Skaffold as users test out the Helm 4 pre-releases leading up to the full 4.0.0 release scheduled for November (full Hem 4 development timeline is here: https://github.com/helm/community/blob/main/hips/hip-0012.md#the-process--timelines).

I reached out to you in the #skaffold k8s slack channel too, so we can have more real-time collaboration on this if you'd like to, especially given the timeline.

I read https://skaffold.dev/docs/deployers/helm/ to understand how Skaffold handles post-rendering with Helm 3.

You hit the nail on the head with how to do this in Helm 4, in your comment here https://github.com/helm/helm/issues/31340#issuecomment-3345781124.

To explain for everyone else – Helm 4 post-renderers have been moved to a new plugin system. Docs for Helm 4 are underway, but the implementation is pretty faithful to the proposal, which you can read here: https://github.com/helm/community/blob/main/hips/hip-0026.md. So yes, you can still use the `skaffold` binary for post-rendering in Helm 4, using the new subprocess plugin runtime and `postrenderer/v1` plugin type. I replied with general information about this for any binary (sed as an example) on the upstream issue too, because I anticipate questions from other people about using other binaries.

As a follow-up, you may also want to look at building a Skaffold plugin for the Wasm runtime too – which is a big reason why we upgraded the plugin system in Helm 4 – but I'd suggest starting with the subprocess runtime because it's essentially a drop-in replacement for the way you do post-rendering with Helm 3. This can help get it out the door for folks to test the Helm 4 pre-release and before the 4.0.0 release. But I'd be happy to look at a Wasm plugin with you too.

brutus333:
Hello! 
[...]

PR #9892, opened 2025-10-20
feat(helm): Support Helm v4 for rendering and deployment, drop Helm v2 support for rendering since `helm version --client` parameter is not supported anymore in v4

AndiDog:
Fixes: #9871
**Related**: n/a
**Merge before/after**: n/a

**Description**

Using the Helm v4 prerelease builds doesn't work right now due to breaking changes in the post-renderer support. Instead of passing `--post-renderer /path/to/skaffold`, a plugin must be registered (see #9871). This PR supports Helm v4 by adding v3-vs-v4 distinction code. Along the way, since deployer code already needs a minimum of Helm v3.1, I drop support for Helm v2 in the render code because `helm version` doesn't support `--client` as parameter anymore.

Since 1) multiple Skaffold binaries could be installed on a system, particularly development builds, and 2) package managers like homebrew won't require helm and install a post-renderer plugin manifest, I opted to instead have Skaffold do the needed `helm plugin install UNIQUE_PLUGIN_NAME` call on the fly (and clean up after every use).

Minor stuff: consistent variable naming `helmVersion`, tiny docs typo fix

- Helm v2 won't work anymore
- Helm v4 works now

- I didn't change documentation yet. The PR is already large, so I'd like to check any Helm version hints in Skaffold docs later.
- On arm64, I couldn't run `make integration-in-kind` and certain other targets. Maybe an integration smoke test for Helm v4 makes sense?

### DIFF
--- a/pkg/skaffold/render/renderer/helm/helm.go
+++ b/pkg/skaffold/render/renderer/helm/helm.go
@@ -44,6 +44,11 @@ import (
 	sUtil "github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/util"
 )
 
+var (
+	// RendererHelmVersionOverride allows replacing the Helm version for testing purposes (to avoid calling installed `helm` binary in tests)
+	RendererHelmVersionOverride *semver.Version = nil
+)
+
 type Helm struct {
 	configName string
 	generate.Generator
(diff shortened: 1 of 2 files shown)


mattsanta:
Could you address the failing checks?

AndiDog:
@mattsanta Tests look green now. I had to introduce another test-related override variable to ensure that no real `helm` command is attempted. Lints also fixed.

AndiDog:
@mattsanta now that [Helm v4 is officially released](https://github.com/helm/helm/releases/tag/v4.0.0), can Skaffold also be released?

mattsanta:
Yep, working on it. @AndiDog

Issue #9901, opened 2025-11-04
image load into kind is not working when kind is running with kind


Blaimi:
Because the image loader is using `kind load docker-image` which is infected by https://github.com/kubernetes-sigs/kind/issues/2038 the image loading doesn't work when kind is started with podman. The builder can easily be utilised to use podman by setting `DOCKER_HOST` to a podman socket like described in https://github.com/GoogleContainerTools/skaffold/issues/8430#issuecomment-1612221929.

This can be fixed by using the concept described in https://github.com/kubernetes-sigs/kind/issues/2038#issuecomment-2889329040.

PR #9902, opened 2025-11-04
Draft: Use a pipeline to load images into kind

Blaimi:
This fixes the issue when kind runs in podman instead of docker

See: https://github.com/kubernetes-sigs/kind/issues/2038
Fixes: #9901

**Description**
Because `kind load docker-image`  doesn't work if kind is running with podman and the image to load is created with podman as well, the concept to load the image has to be changed. This problem is described in https://github.com/kubernetes-sigs/kind/issues/2038. An "easy fix" for this is to save the image with the corresponding container engine as a docker-image to stdout and pass it to `kind image-load /dev/stdin`. By using `localDocker` like the builder does, podman can be selected by setting `DOCKER_HOST` to a podman-socket.

This implementation is more like a proof of concept than anything else. As everyone easily can see, I'm not a go-developer ;-)

This was tested as in "works on my machine". My machine at the moment:
* Fedora 34
* podman 5.6.2
* `systemctl`s `--user podman.socket` is enabled and started
* `DOCKER_HOST=unix:///var/run/user/$(id -u)/podman/podman.sock`
* kind 0.30.0
* kubectl v1.28.15
* cluster version v1.34.0
* application used for testing: https://gitlab.com/biletado/backend/go/-/tree/feature/skaffold

If implemented correctly, nothing changes for docker users

Do a real implementation ;-)

### DIFF
--- a/pkg/skaffold/deploy/component/kubernetes/component.go
+++ b/pkg/skaffold/deploy/component/kubernetes/component.go
@@ -84,7 +84,7 @@ func newDebugger(mode config.RunMode, podSelector kubernetes.PodSelector, namesp
 
 func newImageLoader(cfg k8sloader.Config, cli *kubectl.CLI) loader.ImageLoader {
 	if cfg.LoadImages() {
-		return k8sloader.NewImageLoader(cfg.GetKubeContext(), cli)
+		return k8sloader.NewImageLoader(cfg, cli)
 	}
 	return &loader.NoopImageLoader{}
 }
--- a/pkg/skaffold/docker/image.go
+++ b/pkg/skaffold/docker/image.go
@@ -92,6 +92,7 @@ type LocalDaemon interface {
 	Push(ctx context.Context, out io.Writer, ref string) (string, error)
 	Pull(ctx context.Context, out io.Writer, ref string, platform v1.Platform) error
 	Load(ctx context.Context, out io.Writer, input io.Reader, ref string) (string, error)
+	Save(ctx context.Context, out io.Writer, ref string) (io.ReadCloser, error)
 	Run(ctx context.Context, out io.Writer, opts ContainerCreateOpts) (<-chan container.WaitResponse, <-chan error, string, error)
 	Delete(ctx context.Context, out io.Writer, id string) error
 	Tag(ctx context.Context, image, ref string) error
@@ -541,6 +542,11 @@ func (l *localDaemon) Load(ctx context.Context, out io.Writer, input io.Reader, 
 	}
 
 	return l.ImageID(ctx, ref)
+}
+
+// Save saves an image to a tar file. Returns a ReadCloser for the tar file.
+func (l *localDaemon) Save(ctx context.Context, out io.Writer, ref string) (io.ReadCloser, error) {
+	return l.apiClient.ImageSave(ctx, []string{ref})
 }
 
 // Tag adds a tag to an image.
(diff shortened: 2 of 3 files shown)


polarathene:
I may be mistaken, but care may want to be considered when dealing with large images (eg: 10GB+).

On WSL2 at least, I recall a similar issue with a tool `wslcompact` that performed the separate write/read operations for exporting a distro VM and then importing it for replacement (resolving a disk usage issue), and it did so via piping like proposed here?

That ended up requiring sufficient RAM if the VM disk image was large enough that it was problematic (_it also involved implicit tar archive/unarchive of the VHDX file IIRC_), which caused the operation to fail depending on the system memory available. Meanwhile manually doing the commands individually without a pipe was successful as actual disk was usable.

Blaimi:
The implementation of `kind load docker-image` is actually using the same approach: https://github.com/kubernetes-sigs/kind/blob/91cc28782d7343dc1547c3d40e7136fd24ed5ddc/pkg/cmd/kind/load/docker-image/docker-image.go#L186

polarathene:
> The implementation of `kind load docker-image` is actually using the same approach

Sorry? I don't see piping there, just support for it. [Here `kind` uses `docker save` to export `.tar` to a temp directory and then load that in](https://github.com/kubernetes-sigs/kind/blob/91cc28782d7343dc1547c3d40e7136fd24ed5ddc/pkg/cmd/kind/load/docker-image/docker-image.go#L149-L166) which is not equivalent to piping.

I'm not that familiar with Go, so I may be mistaken but it looks like this PR is not writing the tar to disk, just piping the export into a reader. IIRC, if the operation doesn't block on the export itself completing and can instead maintain a smaller buffer during the operation to stream the export/import in smaller chunks, then memory overhead shouldn't balloon like I've observed with other tools like `wslcompact`.

---

## Context

Import/export separately (_required if using `nerdctl` for example, rather than implicitly supported by `kind`_):

```bash
docker save localhost/example -o example.tar
kind load image-archive example.tar
```

What [you proposed at `kind`](https://github.com/kubernetes-sigs/kind/issues/2038#issuecomment-2889329040) to implement in Go via piping:

```bash
kind load image-archive -n kind-cluster <(podman save "$TAG" --format oci-archive)
```

That creates a temporary file descriptor, but I can't recall if that has that same implications that [`wslcompact` has](https://github.com/okibcn/wslcompact/issues/26) which does the following (PowerShell):

```powershell
cmd /c "wsl --export ""$wsl_distro"" - | wsl --import wslcompact ""$tmp_folder"" -" 
```

Using an actual pipe operator `|` with stdout piped into stdin. AFAIK your use of _process substitution_ `<(...)` is roughly equivalent with the file descriptor still being treated as an in-memory buffer by the kernel, nothing is written to disk.

I'm pretty sure that would run into the same piping issue, and so should the approach in Go then if the whole image is buffered into memory vs read from disk.

Blaimi:
Podman desktop is also writing the image to the disk first: https://github.com/podman-desktop/podman-desktop/blob/main/extensions/kind/src/image-handler.ts#L80.

Maybe this is the saver approach, since two tools already are using this concept. Piping itself should not increase the memory usage, but I don't know, how this works, if you mix the systems where the image is coming from with the target system. E.g. if the image is loaded from docker (and therefore the docker wsl-vm) targeting a kind cluster running in podman (and therefore the podman wsl-vm)

Issue #9913, opened 2025-11-14
skaffold build enforces Helm with 2.17.0


AndreasBergmeier6176:
After 2.17.0 `skaffold build` now enforces Helm. It would be preferably to be able to run it without helm.

mattsanta:
Could you provide more information to reproduce the issue? Logs could also be helpful if you run the command with `--verbosity=debug`.

I'm able to run a basic local `skaffold build` without Helm in my environment so there must be some case that's causing this issue.

AndreasBergmeier6176:
Well we call `skaffold build -f skaffold_build.yaml --tag 80116a54bc6a474898dc5f2a8c0f1923a8373b55`

which then errors with: `creating runner: creating renderer: Helm not found. Please install helm via https://helm.sh/docs/intro/install.`

Just before it fails it says:
> time="2025-11-14T19:33:00Z" level=debug msg="Running command: [helm version]" subtask=-1 task=DevLoop
time="2025-11-14T19:33:00Z" level=debug msg="Running command: [tput colors]" subtask=-1 task=DevLoop

It may be that the problem got introduced in https://github.com/GoogleContainerTools/skaffold/commit/2b280721870552e9bf8b3f846172d65bec7d3434.

Previously a new instance of Helm support was instantiated without testing whether helm was actually present. Now in the new code it explicitly calls helm in `New` which then enforces a present `helm` binary as soon as some Helm rendering gets detected.

Whether it makes sense to instantiate Helm support in the first place when calling `skaffold build` is a different question - but I fear that only instantiating what is necessary might need a wider refactoring.

AndreasBergmeier6176:
Anyone?

PR #9914, opened 2025-11-14
fix: late evaluate helm version

AndreasBergmeier6176:
This restores the behavior that you can execute skaffold build without needing helm.

Fixes: #9913


mattsanta:
Could you address the unit test issues? Thanks!

AndreasBergmeier6176:
> Could you address the unit test issues? Thanks!

`make coverage` should now work.

PR #10054, opened 2026-04-18
fix: add --set and --set-value-file flag support to delete command and unit tests

### LINKED ISSUE #10053: `skaffold delete` does not support `--set` flag for manifest overrides from Google Cloud Deploy deploy parameters
## Problem

When using Cloud Deploy's [deploy parameters](https://cloud.google.com/deploy/docs/parameters) with `# from-param:` annotations in manifests, `skaffold delete` cannot resolve the correct resource name to delete because the `--set` or `--set-value-file` flag is not available on the `delete` subcommand.

This affects any deployment where the resource name or namespace is parameterized which is common in ephemeral/preview environments where each PR gets a uniquely-named Cloud Run service or Kubernetes deployment.

## Example

A Cloud Run service manifest using deploy parameters:

```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: my-service # from-param: my-service-${environment}
```

`skaffold render` and `skaffold deploy` both support `--set environment=pr-42`, which substitutes the name to `my-service-pr-42` via the `ApplySetters` render step. However, `skaffold delete` does not accept `--set` or `--set-value-file`:

```
$ skaffold delete --set environment=pr-42
Error: unknown flag: --set
```

Without `--set`, the `delete` command renders the manifest with the un-substituted base name (`my-service`) and either:

- Deletes the **wrong resource** (the base service instead of the ephemeral one), or
- Fails with a 404 if the base service doesn't exist

## Root Cause

In [`cmd/skaffold/app/cmd/flags.go`](https://github.com/GoogleContainerTools/skaffold/blob/main/cmd/skaffold/app/cmd/flags.go), the `--set` flag is defined with:

```go
DefinedOn: []string{"render", "filter"},
```

The `delete` command is missing from this list, even though `delete` already executes the full render pipeline internally (`doDelete` → `Build` → `Render` → `Cleanup`), and the render step already consumes `opts.ManifestsOverrides` through `ApplySetters`.

## Proposed Fix

Add `"delete"` to the `DefinedOn` list for the `--set` flag:

```go
DefinedOn: []string{"render", "filter", "delete"},
```

No other changes are needed — the render pipeline in the delete flow already processes `ManifestsOverrides` correctly.

## Use Cases

- CI/CD cleanup jobs for ephemeral environments (e.g., when a PR is closed):
- Simulating Cloud Deploy deployment locally and then cleaning up.

```bash
skaffold delete --set environment=pr-${PR_NUMBER}
skaffold delete --set environment=test123
```

This applies to both Cloud Run and Kubernetes/GKE deployers when resource names or namespaces are parameterized via `# from-param:` annotations.

bogdannazarenko:
**Description**
This PR adds support for the `--set` and `--set-value-file` flag to the `skaffold delete` command, allowing users to override templated manifest fields with key-value pairs during deletion operations.

Fixes #10053 

**Changes made:**
1. Updated `cmd/skaffold/app/cmd/flags.go` line 764 to include "delete" in the `DefinedOn` list for the "set" and "set-value-file" flags
2. Created `cmd/skaffold/app/cmd/delete_test.go` with unit tests to verify that the delete command accepts `--set`, `--set-value-file` and `--dry-run` flags
3. Verified the built skaffold binary correctly accepts the `--set` and `--set-value-file` flag for the delete command

**User facing changes**
Users can now use the `--set` and `--set-value-file` flags with `skaffold delete` to override templated manifest fields:

**Before:** `skaffold delete` did not accept `--set` or `--set-value-file` flags
**After:** `skaffold delete --set key=value --set set-value-file=file` is now supported

This allows for more flexible deletion operations when working with templated manifests.

**Testing**
- Added unit test `TestNewCmdDelete` in `delete_test.go` that verifies both `--set`, `--set-value-file` and `--dry-run` flags are available
- All existing tests pass
- All linters pass

### DIFF
--- a/cmd/skaffold/app/cmd/delete_test.go
+++ b/cmd/skaffold/app/cmd/delete_test.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2019 The Skaffold Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"testing"
+
+	"github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/config"
+	"github.com/GoogleContainerTools/skaffold/v2/testutil"
+)
+
+func TestNewCmdDelete(t *testing.T) {
+	testutil.Run(t, "", func(t *testutil.T) {
+		t.NewTempDir().Chdir()
+		t.Override(&opts, config.SkaffoldOptions{})
+		t.Override(&dryRun, false)
+
+		cmd := NewCmdDelete()
+		cmd.SilenceUsage = true
+
+		// Check that the command accepts the --set flag and it's correctly bound
+		err := cmd.Flags().Parse([]string{"--set", "key1=val1", "--set", "key2=val2"})
+		t.CheckNoError(err)
+		t.CheckDeepEqual([]string{"key1=val1", "key2=val2"}, opts.ManifestsOverrides)
+
+		// Check that the command accepts the --set-value-file flag and it's correctly bound
+		err = cmd.Flags().Parse([]string{"--set-value-file", "values.env"})
+		t.CheckNoError(err)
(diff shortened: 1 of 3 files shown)


Darien-Lin:
Seems that the unit test require you to run: ./hack/generate-man.sh

bogdannazarenko:
@Darien-Lin the failed integration tests are not related to these changes. Looks like they are pre-existing failures
https://github.com/GoogleContainerTools/skaffold/actions/runs/24786039920/job/72530549490?pr=10054#step:19:3587
https://github.com/GoogleContainerTools/skaffold/actions/runs/24786039920/job/72530549490?pr=10054#step:19:3529

Darien-Lin (approved):

PR #10066, opened 2026-04-24
feat: local runArgs whitelist for custom actions and verify (skaffold/v4beta15)

bogdannazarenko:
## Local `runArgs` for custom actions **and** verify

Closes #10130.

Lets local (Docker) custom-action and `verify` containers forward a small, conservative whitelist of `docker run` flags via `executionMode.local.runArgs`. The primary motivation is **local/Cloud Deploy parity**: Cloud Deploy's execution environment already runs as a service account (ADC works out of the box), but local `skaffold exec` / `skaffold verify` containers have no access to the host's Google credentials. A bind-mount via `runArgs` closes the gap.

```yaml
customActions:
  - name: reuse-local-adc
    executionMode:
      local:
        runArgs:
          - "-v=/root/.config/gcloud:/root/.config/gcloud:ro"
          - "--network=host"
    containers:
      - name: gcloud
        image: google/cloud-sdk:slim
        command: ["gcloud"]
        args: ["auth", "list"]
```

### Whitelist (all map to the Docker `HostConfig`)

| Flag | Effect |
| --- | --- |
| `--network=<mode>` | `HostConfig.NetworkMode` |
| `-v`/`--volume=<src>:<dst>[:opts]` | `HostConfig.Binds` |
| `--add-host=<host>:<ip>` | `HostConfig.ExtraHosts` |
| `--tmpfs=<path>[:opts]` | `HostConfig.Tmpfs` |

Only the `--flag=value` form is accepted; unknown flags are rejected at load time. (Reduced from an earlier, broader set — `-e`/`--user`/`--privileged`/`--cap-add`/`--cap-drop` were dropped for a tighter security posture, which also let `ApplyToContainerConfig` go away entirely.)

### DRY

The parser + `HostConfig` overlay live once in the neutral `pkg/skaffold/docker` package (next to `ContainerCreateOpts`); `actions/docker` and `verify/docker` both consume it through the existing `ContainerCreateOpts.HostConfigApply` hook. The two container-creation runtimes stay separate — no rewrite.

### New schema version `skaffold/v4beta15`

`skaffold/v4beta14` is released ("DO NOT MODIFY"), so adding `LocalVerifier.runArgs` required cutting a new version via `hack/new-version.sh`.

**Note to reviewer**: I only committed the new `config.go` schema v4beta15 as @Darien-Lin mentioned this is usually done by the maintainers. If you would like me to push all changes generated by cutting the schema version, let me know. Feel free to push to this branch yourself as well and merge.

### Tests

[...]

PR #10068, opened 2026-04-24
feat(verify/docker): apply runArgs whitelist to verify test containers

bogdannazarenko:
## Summary

Follow-up to #10066 (custom-actions runArgs). Extends the `LocalVerifier` schema with an optional `runArgs` list and plumbs it through `pkg/skaffold/verify/docker` so `skaffold verify` containers get the same conservative whitelist already supported on `customActions.*.executionMode.local.runArgs`.

## Why

The roadmap note on PR #10066 explicitly called this out as a deferred follow-up:
> `skaffold verify` shares the `LocalVerifier` struct but has its own run path in `pkg/skaffold/verify/docker` that does not yet consume `RunArgs`. Wiring that into verify is a straightforward follow-up but kept separate to keep this diff focused on custom actions.

This PR is that follow-up. It's intentionally small — one new schema field and one call site.

## Changes

- **schema**: `LocalVerifier` gains `RunArgs []string` with the same godoc whitelist (`--network`, `-v`/`--volume`, `-e`/`--env`, `--user`, `--add-host`, `--tmpfs`, `--privileged`, `--cap-add`, `--cap-drop`).
- **verify/docker/verify.go**: `createAndRunContainer` parses the per-test-case `runArgs` via `actionsdocker.ParseRunArgs`. Unknown flags fail the test case instead of being silently dropped. The parsed result overlays the container config (User, extra Env) after the schema-sourced env and `--verify-env-file` values, and exposes `ApplyToHostConfig` via the existing `ContainerCreateOpts.HostConfigApply` hook.
- **tests**: `fakeDockerDaemon` gains a `RunOpts` capture slice; new `Test_RunArgs` covers the nil, happy-path overlay, and unknown-flag cases.

No runtime change occurs when `runArgs` is absent — overlays are no-ops on a nil `*RunArgs` receiver.

## Stacked on #10066

This branch is based on #10066's head, so the PR diff includes the three phase-2 commits. Once #10066 merges, I'll rebase onto `main` and the diff will collapse to the two commits introduced here:

```
d2005318e feat(verify/docker): apply runArgs whitelist to verify test containers
9a0bf21ed test(verify/docker): unit coverage for runArgs overlay + unknown-flag error
```

## No schema version bump

Keeps `latest` at `skaffold/v4beta14` per the maintainer-driven `hack/new-version.sh` workflow.

## Tests

```
go test -count=1 -short -timeout=60s ./pkg/skaffold/verify/docker/...
ok  github.com/GoogleContainerTools/skaffold/v2/pkg/skaffold/verify/docker
```

`make quicktest` is clean aside from the pre-existing `TestNewEnvClient/DOCKER_HOST_not_set` and `TestNewRunner` failures (environment-dependent, reproduce on `main`) and `TestParseExamples/custom-actions-runargs` (pre-existing on the phase-2 branch this stacks on; the example YAML declares `skaffold/v4beta15` which doesn't exist on HEAD yet — will resolve with the maintainer's `hack/new-version.sh` bump).

### DIFF
--- a/docs-v2/content/en/docs/custom-actions.md
+++ b/docs-v2/content/en/docs/custom-actions.md
@@ -109,6 +109,53 @@ A Custom Action has an execution mode associated with it that indicates Skaffold
 
 This is the default configuration when no [`customActions[].executionMode`]({{< relref "/docs/references/yaml/#customActions-executionMode" >}}) is specified. With this execution mode, Skaffold will run every container associated to a given Custom Action with a Docker daemon.
 
+##### Passing Docker run flags with `runArgs`
+
+When an action needs to reach host resources (for example your local
+`~/.config/gcloud` credentials) or to tighten its runtime (dropping
+capabilities, pinning a non-root user), use
+[`customActions[].executionMode.local.runArgs`]({{< relref
+"/docs/references/yaml/#customActions-executionMode-local-runArgs" >}}).
+Skaffold parses each entry with a whitelist and overlays the result on
+the Docker `HostConfig`/`Config` used to start the container:
+
+| Flag | Effect |
+| --- | --- |
+| `--network=<mode>` | Sets `HostConfig.NetworkMode`. Accepts `host`, `bridge`, `none`, or a named network. |
+| `-v=<src>:<dst>[:opts]`, `--volume=<src>:<dst>[:opts]` | Appends to `HostConfig.Binds`. |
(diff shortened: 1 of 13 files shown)


bogdannazarenko:
Superseded by #10066, which now folds in the verify wiring (and adds the new skaffold/v4beta15 schema cut + reduced whitelist). Closing in favor of that combined PR.

PR #10108, opened 2026-06-11
fix: use correct module to handle windows path separator

foriequal0:
Fixes: #9988

**Description**
On Windows, Skaffold doesn't work with helm 4 since beginning.

e.g., for the following `skaffold.yaml`:

```yaml
apiVersion: skaffold/v4beta13
kind: Config
deploy:
  helm:
    releases:
      - name: postgresql
        remoteChart: oci://registry-1.docker.io/bitnamicharts/postgresql
        valuesFiles:
          - postgresql.values.yaml
        version: 16.3.1
```

`skaffold run` fails with the following cryptic error message:
```
deploying "postgresql": Failed to install Helm plugin: Installing plugin from local directory (development mode)
Error: plugin is installed but unusable: failed to load plugin "C:\\Users\\user\\AppData\\Roaming\\helm\\plugins\\skaffold-render4198907898": failed to peek plugin.yaml API version: yaml: line 2: did not find expected hexdecimal number
```

The previous code `helmPluginName := path.Base(helmPluginDir)` couldn't handle Windows' path separator `\` properly.
`helmPluginName` becomes just `C:\Users\user\AppData\Local\Temp\skaffold-render1414094366` (instead of intended `skaffold-render1414094366`), and it is `Sprintf`'d into `PostRendererTemplate`. `plugin.yaml` ends up containing an unescaped `\`, which confuses the helm's YAML parser with the escape sequence, so it makes helm emit a seemingly unrelated confusing error message `did not find expected hexdecimal number` (typical golang library/toolchain error message).

So, fix it by using `path/filepath` instead of `path` which can handle both `\\` and `/`.

skaffold with helm deploy works correctly for helm v4.

N/A

### DIFF
--- a/pkg/skaffold/helm/bin.go
+++ b/pkg/skaffold/helm/bin.go
@@ -24,7 +24,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
-	"path"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


aniskhan001 (approved):

xDeyan:
It would be really nice if we can get this fix into the next release!

PR #10127, opened 2026-06-26
feat(diagnose): default --enable-templating to true

### LINKED ISSUE #10126: diagnose: --enable-templating defaults to false, breaking Cloud Deploy deploy-parameter interpolation
## Description

Google Cloud Deploy's render phase generates the canonical `config.yaml` for a release internally via `skaffold diagnose`, but invokes it without `--enable-templating=true`. As a result, any Skaffold config field that supports Go-template interpolation (tagged `skaffold:"...,template"`, e.g. `deploy.kubectl.defaultNamespace`) is left as a literal, unexpanded template string instead of being populated from the deploy parameters / environment variables that Cloud Deploy injects for the release.

This further cascades into downstream steps — e.g. the `verify` job — that rely on the resolved `deploy.kubectl.defaultNamespace`.

## Relevant log (Cloud Deploy render phase, trimmed to the relevant `skaffold` invocations)

```
Running the following command: skaffold [render --filename=/workspace/stable/config.yaml --build-artifacts=/workspace/artifacts.json --output=/workspace/stable/manifest.yaml --offline=true --digest-source=none --set="environment=pr-88" --set="git_sha=<redacted>"]
--digest-source set to 'none', tags listed in Kubernetes manifests will be used for render
Running the following command: skaffold [inspect namespaces list /workspace/stable/manifest.yaml --filename=/workspace/stable/config.yaml]
{"resourceToInfoMap":{"apps/v1, Kind=Deployment":[{"name":"myapp-gke","namespace":"myapp-gke-{{.environment}}"}]}}
```

Note `"namespace":"myapp-gke-{{.environment}}"` — the `environment` deploy parameter (`pr-88` in this release) was never substituted into the template, because `/workspace/stable/config.yaml` (the effective config Cloud Deploy generated via `skaffold diagnose`) was produced without `--enable-templating`.

## Root cause

`skaffold diagnose --enable-templating` defaults to `false` (`cmd/skaffold/app/cmd/diagnose.go`). Cloud Deploy never passes `--enable-templating=true` when it calls `diagnose` to produce its internal `config.yaml`, so fields tagged `skaffold:"...,template"` are never expanded.

## Reproduction

Minimal `skaffold.yaml` that templates `deploy.kubectl.defaultNamespace` from a deploy parameter named `environment`:

```yaml
apiVersion: skaffold/v4beta14
kind: Config
metadata:
  name: myapp
build:
  artifacts:
    - image: myapp
      docker:
        dockerfile: Dockerfile
manifests:
  rawYaml:
    - k8s/*.yaml
deploy:
  kubectl:
    defaultNamespace: "myapp-{{.environment}}"
```

```shell
# simulates how Cloud Deploy invokes diagnose internally (deploy parameter as env var, no --enable-templating)
$ environment=pr-88 skaffold diagnose --yaml-only -f skaffold.yaml
...
deploy:
  kubectl:
    defaultNamespace: myapp-{{.environment}}   # BUG: not interpolated

# manual workaround, passing the flag explicitly
$ environment=pr-88 skaffold diagnose --yaml-only -f skaffold.yaml --enable-templating
...
deploy:
  kubectl:
    defaultNamespace: myapp-pr-88              # expected
```

## Expected behavior

Deploy parameters supplied as environment variables (as Google Cloud Deploy does) should be interpolated into templated fields by default, without requiring Cloud Deploy (or any other caller of `diagnose`) to explicitly pass `--enable-templating=true`.

## Proposed fix

Default `--enable-templating` to `true` on `skaffold diagnose`. Templating is opt-out (`--enable-templating=false`) rather than opt-in, and `expandTemplate` already leaves a field untouched if expansion produces `<no value>` (e.g. when the underlying env var isn't set), so this does not change behavior for configs that aren't using templated fields.

bogdannazarenko:
## Summary

- Fixes #10126
- `skaffold diagnose` generates the effective Skaffold config used by Google Cloud Deploy's render phase, but the `--enable-templating` flag defaulted to `false`. Cloud Deploy never passes this flag explicitly, so deploy parameters (injected as environment variables) were never interpolated into templated fields such as `deploy.kubectl.defaultNamespace`, which then cascaded into downstream steps (e.g. the `verify` job) relying on the resolved namespace.
- Flips the default to `true` so templating happens out of the box; users can still opt out with `--enable-templating=false`. `expandTemplate` already preserves the original template string when expansion yields `<no value>` (e.g. env var unset), so configs that don't rely on templated fields are unaffected.
- Regenerated the CLI reference docs (`docs-v2/content/en/docs/references/cli/_index.md`) via `./hack/generate-man.sh` to reflect the new default.

## Test plan

- [x] `make` builds successfully
- [x] `skaffold diagnose --help` shows `--enable-templating=true`
- [x] Manual repro: with a config templating `deploy.kubectl.defaultNamespace: "myapp-{{.environment}}"`, running `environment=pr-88 skaffold diagnose --yaml-only` (no flag) now resolves to `defaultNamespace: myapp-pr-88` instead of leaving the literal template
- [x] `go test ./cmd/skaffold/app/cmd/ -run TestDoDiagnose` passes
- [x] Existing `integration/diagnose_test.go` cases that assert on non-templated testdata are unaffected (no `{{ }}` syntax in those fixtures)

### DIFF
--- a/cmd/skaffold/app/cmd/diagnose.go
+++ b/cmd/skaffold/app/cmd/diagnose.go
@@ -57,7 +57,7 @@ func NewCmdDiagnose() *cobra.Command {
 		WithCommonFlags().
 		WithFlags([]*Flag{
 			{Value: &yamlOnly, Name: "yaml-only", DefValue: false, Usage: "Only prints the effective skaffold.yaml configuration"},
-			{Value: &enableTemplating, Name: "enable-templating", DefValue: false, Usage: "Render supported templated fields with golang template engine"},
+			{Value: &enableTemplating, Name: "enable-templating", DefValue: true, Usage: "Render supported templated fields with golang template engine"},
 			{Value: &outputFile, Name: "output", Shorthand: "o", DefValue: "", Usage: "File to write diagnose result"},
 		}).
 		NoArgs(doDiagnose)
--- a/docs-v2/content/en/docs/references/cli/_index.md
+++ b/docs-v2/content/en/docs/references/cli/_index.md
@@ -1330,7 +1330,7 @@ Options:
     -c, --config='':
 	File for global configurations (defaults to $HOME/.skaffold/config)
 
-    --enable-templating=false:
+    --enable-templating=true:
 	Render supported templated fields with golang template engine
 
     -f, --filename='skaffold.yaml':


bogdannazarenko:
Hi @Darien-Lin @menahyouyeah could you please take a look at this PR?

